PR #335, opened 2019-11-26
Implement client API for `ApplicationClient`

### LINKED ISSUE #333: Implement client API for `ApplicationClient`
The new API should include all the parameters that affect the root of trust of the TCB when performing remote attestation.

`ApplicationClient` should be parametrised in order to include at least the following:

- Intel public key used to verify remote attestation claims
- list of one or more MRENCLAVE values that are considered valid during the remote attestation process
  + in the future this may be replaced with a pointer to a verifiable log of some sort including such list of valid MRENCLAVE values
- minimum required Intel SGX version / CPU configuration
- other relevant attributes to check in the SGX attestation report

Blocks #119

ipetr0v:
This change:
- Adds a new client API implemented in `SgxApplicationClient`
- Implements parameter initialization for the remote attestation
- Adds bazel rule for enclave metadata generation

Fixes #333

### DIFF
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -63,8 +63,16 @@ cc_library(
         "//oak/common:nonce_generator",
         "//oak/proto:application_cc_grpc",
         "@com_github_grpc_grpc//:grpc++",
+        "@com_google_asylo//asylo/grpc/auth:grpc++_security_enclave",
+        "@com_google_asylo//asylo/grpc/auth:enclave_credentials_options",
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
+        # TODO: Uncomment when asylo will release remote attestation
+        #"@com_google_asylo//asylo/grpc/auth:sgx_remote_credentials_options",
+        "@com_google_asylo//asylo/identity:enclave_assertion_authority_configs",
+        "@com_google_asylo//asylo/identity:descriptions",
         "@com_google_asylo//asylo/identity:init",
+        "@com_google_asylo//asylo/identity/sgx:sgx_identity_util",
+        "@com_google_asylo//asylo/identity/util:sha256_hash_util",
         "@com_google_asylo//asylo/util:logging",
     ],
 )
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -45,9 +45,6 @@ cc_library(
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
         "@com_google_asylo//asylo/identity:init",
         "@com_google_asylo//asylo/util:logging",
-        # Dependency is added for automated tests
-        # TODO: Merge application_client and sgx_application_client
-        ":sgx_application_client"
     ],
 )
 
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -45,6 +45,27 @@ cc_library(
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
         "@com_google_asylo//asylo/identity:init",
         "@com_google_asylo//asylo/util:logging",
+        # Dependency is added for automated tests
+        # TODO: Merge application_client and secure_application_client
+        ":secure_application_client"
+    ],
+)
+
+cc_library(
+    name = "secure_application_client",
+    srcs = ["secure_application_client.cc"],
+    hdrs = ["secure_application_client.h"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":authorization_bearer_token_metadata",
+        ":policy_metadata",
+        "//oak/common:app_config",
+        "//oak/common:nonce_generator",
+        "//oak/proto:application_cc_grpc",
+        "@com_github_grpc_grpc//:grpc++",
+        "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
+        "@com_google_asylo//asylo/identity:init",
+        "@com_google_asylo//asylo/util:logging",
     ],
 )
 
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -63,6 +63,7 @@ cc_library(
         "@com_google_asylo//asylo/grpc/auth:grpc++_security_enclave",
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
         # TODO: Uncomment when Asylo will release remote attestation.
+        # https://github.com/project-oak/oak/issues/119
         #"@com_google_asylo//asylo/grpc/auth:sgx_remote_credentials_options",
         "@com_google_asylo//asylo/identity:enclave_assertion_authority_configs",
         "@com_google_asylo//asylo/identity:descriptions",
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -46,15 +46,15 @@ cc_library(
         "@com_google_asylo//asylo/identity:init",
         "@com_google_asylo//asylo/util:logging",
         # Dependency is added for automated tests
-        # TODO: Merge application_client and secure_application_client
-        ":secure_application_client"
+        # TODO: Merge application_client and sgx_application_client
+        ":sgx_application_client"
     ],
 )
 
 cc_library(
-    name = "secure_application_client",
-    srcs = ["secure_application_client.cc"],
-    hdrs = ["secure_application_client.h"],
+    name = "sgx_application_client",
+    srcs = ["sgx_application_client.cc"],
+    hdrs = ["sgx_application_client.h"],
     visibility = ["//visibility:public"],
     deps = [
         ":authorization_bearer_token_metadata",
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -64,7 +64,6 @@ cc_library(
         "//oak/proto:application_cc_grpc",
         "@com_github_grpc_grpc//:grpc++",
         "@com_google_asylo//asylo/grpc/auth:grpc++_security_enclave",
-        "@com_google_asylo//asylo/grpc/auth:enclave_credentials_options",
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
         # TODO: Uncomment when asylo will release remote attestation
         #"@com_google_asylo//asylo/grpc/auth:sgx_remote_credentials_options",
@@ -72,7 +71,7 @@ cc_library(
         "@com_google_asylo//asylo/identity:descriptions",
         "@com_google_asylo//asylo/identity:init",
         "@com_google_asylo//asylo/identity/sgx:sgx_identity_util",
-        "@com_google_asylo//asylo/identity/util:sha256_hash_util",
+        "@com_google_asylo//asylo/identity/util:sha256_hash_cc_proto",
         "@com_google_asylo//asylo/util:logging",
     ],
 )
--- a/oak/client/BUILD
+++ b/oak/client/BUILD
@@ -62,7 +62,7 @@ cc_library(
         "@com_github_grpc_grpc//:grpc++",
         "@com_google_asylo//asylo/grpc/auth:grpc++_security_enclave",
         "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
-        # TODO: Uncomment when asylo will release remote attestation
+        # TODO: Uncomment when Asylo will release remote attestation.
         #"@com_google_asylo//asylo/grpc/auth:sgx_remote_credentials_options",
         "@com_google_asylo//asylo/identity:enclave_assertion_authority_configs",
         "@com_google_asylo//asylo/identity:descriptions",
@@ -70,6 +70,7 @@ cc_library(
         "@com_google_asylo//asylo/identity/sgx:sgx_identity_util",
         "@com_google_asylo//asylo/identity/util:sha256_hash_cc_proto",
         "@com_google_asylo//asylo/util:logging",
+        "@com_google_asylo//asylo/util:status",
     ],
 )
 
--- a/oak/client/sgx_application_client.cc
+++ b/oak/client/sgx_application_client.cc
@@ -24,8 +24,8 @@
 #include "asylo/grpc/auth/enclave_credentials_options.h"
 #include "asylo/grpc/auth/null_credentials_options.h"
 //#include "asylo/grpc/auth/sgx_remote_credentials_options.h"
-#include "asylo/identity/enclave_assertion_authority_configs.h"
 #include "asylo/identity/descriptions.h"
+#include "asylo/identity/enclave_assertion_authority_configs.h"
 #include "asylo/identity/init.h"
 #include "asylo/identity/sgx/sgx_identity_util.h"
 #include "asylo/identity/util/sha256_hash.pb.h"
@@ -64,17 +64,14 @@ SgxApplicationClient::SgxApplicationClient(
   // Initialize credentials
   auto channel_credentials = this->CreateChannelCredentials(mrenclave_strings);
   auto call_credentials = this->CreateCallCredentials();
-  this->credentials_ =
-      grpc::CompositeChannelCredentials(channel_credentials, call_credentials);
+  this->credentials_ = grpc::CompositeChannelCredentials(channel_credentials, call_credentials);
 }
 
-std::shared_ptr<grpc::Channel>
-SgxApplicationClient::CreateChannel(std::string address) {
+std::shared_ptr<grpc::Channel> SgxApplicationClient::CreateChannel(std::string address) {
   return grpc::CreateChannel(address, this->credentials_);
-}  
- 
-asylo::EnclaveAssertionAuthorityConfig
-SgxApplicationClient::GetNullAssertionAuthorityConfig() {
+}
+
+asylo::EnclaveAssertionAuthorityConfig SgxApplicationClient::GetNullAssertionAuthorityConfig() {
   asylo::EnclaveAssertionAuthorityConfig test_config;
   asylo::SetNullAssertionDescription(test_config.mutable_description());
   return test_config;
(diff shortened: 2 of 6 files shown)


oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+#include "oak/client/sgx_application_client.h"
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "asylo/grpc/auth/enclave_channel_credentials.h"
+#include "asylo/grpc/auth/enclave_credentials_options.h"
+#include "asylo/grpc/auth/null_credentials_options.h"
+//#include "asylo/grpc/auth/sgx_remote_credentials_options.h"
```
tiziano88: Also add a TODO here.

oak/client/BUILD
```diff
@@ -48,6 +48,31 @@ cc_library(
...
+    deps = [
+        ":authorization_bearer_token_metadata",
+        ":policy_metadata",
+        "//oak/common:app_config",
+        "//oak/common:nonce_generator",
+        "//oak/proto:application_cc_grpc",
+        "@com_github_grpc_grpc//:grpc++",
+        "@com_google_asylo//asylo/grpc/auth:grpc++_security_enclave",
+        "@com_google_asylo//asylo/grpc/auth:null_credentials_options",
+        # TODO: Uncomment when asylo will release remote attestation
```
tiziano88:
```suggestion
        # TODO: Uncomment when Asylo will release remote attestation.
```

(and below)

oak/client/sgx_application_client.h
```diff
@@ -0,0 +1,56 @@
...
+#include <string>
+#include <vector>
+
+#include "asylo/grpc/auth/enclave_credentials_options.h"
+#include "asylo/identity/enclave_assertion_authority_configs.h"
+#include "include/grpcpp/grpcpp.h"
+
+namespace oak {
+
+class SgxApplicationClient {
```
tiziano88: Could you add some documentation on what this does, and where the settings come from (e.g. from Asylo).

oak/client/sgx_application_client.h
```diff
@@ -0,0 +1,56 @@
...
+ public:
+  SgxApplicationClient(
+      /*TODO: intel_public_key,*/ std::vector<std::string> mrenclave_strings);
```
tiziano88: I would move this TODO up just before the constructor, like:

`TODO: Add parameter to specify Intel public key to use for verification`

or similar.

oak/client/sgx_application_client.h
```diff
@@ -0,0 +1,56 @@
...
+
+  std::shared_ptr<grpc::Channel> CreateChannel(std::string address);
+
+ private:
+  // TODO: Use assertion helper from asylo, when it becomes available.
```
tiziano88:
```suggestion
  // TODO: Use assertion helper from Asylo, when it becomes available.
```

oak/server/asylo/BUILD
```diff
@@ -93,6 +93,7 @@ sgx.debug_enclave(
     config = ":grpc_enclave_config",
 )
 
+# Pulls the MRENCLAVE value from an enclave signing material
```
tiziano88:
```suggestion
# Pulls the MRENCLAVE value from an enclave signing material.
```

oak/server/asylo/BUILD
```diff
@@ -109,6 +110,15 @@ genrule(
     cmd = "xxd -p -s 188 -l 32 -c 32 $< > $@",
 )
 
+# Generates an enclave dump that contains the MRSIGNER value
```
tiziano88:
```suggestion
# Generates an enclave dump that contains the MRSIGNER value.
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+#include "oak/client/authorization_bearer_token_metadata.h"
+#include "oak/client/policy_metadata.h"
+#include "oak/common/nonce_generator.h"
+#include "oak/common/policy.h"
+
+constexpr size_t kPerChannelNonceSizeBytes = 32;
+// Debug MRSIGNER value derived from the SGX test key (will change after changing SVN and PRODID).
+const char* kDebugMrSigner = "83d719e77deaca1470f6baf62a4d774303c899db69020f9c70ee1dfc08c7ce9e";
+
+// TODO: Use same function from asylo, when it will become public.
```
tiziano88:
```suggestion
// TODO: Use same function from Asylo, when it becomes public.
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+bool Sha256HashFromHexString(const std::string& hex, asylo::Sha256HashProto* h) {
```
tiziano88: Please clarify at least what the return value means. Even better, return a `Status` object with an appropriate error message instead.

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+  if (hex.size() != 64) {
+    return false;
+  }
+  for (auto ch : hex) {
+    if (std::isxdigit(ch) == 0) {
+      return false;
+    }
+  }
+  h->set_hash(absl::HexStringToBytes(hex).c_str(), hex.size() / 2);
```
tiziano88: Maybe safer to use the size from the resulting string, even if it means declaring an extra variable. Also I'm surprised there is no `set_hash` method that accepts an `std::string`?

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+  return true;
+}
+
+namespace oak {
+
+SgxApplicationClient::SgxApplicationClient(
+    /*intel_public_key,*/
+    std::vector<std::string> mrenclave_strings) {
+  // Initialize assertion authorities
```
tiziano88:
```suggestion
  // Initialize assertion authorities.
```

oak/client/sgx_application_client.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include "asylo/grpc/auth/enclave_credentials_options.h"
+#include "asylo/identity/enclave_assertion_authority_configs.h"
+#include "include/grpcpp/grpcpp.h"
+
+namespace oak {
+
+class SgxApplicationClient {
+ public:
+  SgxApplicationClient(
```
tiziano88: Document what `mrenclave_strings` is used for, i.e. that it verifies the remote attestation and then only accepts an attestation in which MRENCLAVE is one of the values provided here.

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+
+// TODO: Add CPUSVN as a parameter.
+asylo::EnclaveIdentityExpectation SgxApplicationClient::CreateSgxIdentityExpectation(
+    std::string& mrenclave_string) const {
+  asylo::SgxIdentity sgx_identity;
+
+  // Assign code identity.
+  auto code_identity = sgx_identity.mutable_code_identity();
+  if (!Sha256HashFromHexString(mrenclave_string, code_identity->mutable_mrenclave())) {
+    LOG(QFATAL) << "Bad MRENCLAVE";
```
tiziano88: Maybe return `StatusOr<asylo::EnclaveIdentityExpectation>` from the function instead of this failure?

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  auto signer_assigned_identity = code_identity->mutable_signer_assigned_identity();
+  if (!Sha256HashFromHexString(kDebugMrSigner, signer_assigned_identity->mutable_mrsigner())) {
+    LOG(QFATAL) << "Bad MRSIGNER";
+  }
+  // TODO: Consider assigning prodid and svn even without MRSIGNER.
+  signer_assigned_identity->set_isvprodid(0);
+  signer_assigned_identity->set_isvsvn(0);
+
+  // TODO: Use asylo misc attributes util when they will become available.
```
tiziano88:
```suggestion
  // TODO: Use Asylo misc attributes util when they become available.
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+  code_identity->set_miscselect(0);
+  auto attributes = code_identity->mutable_attributes();
+  attributes->set_flags(0);
+  attributes->set_xfrm(0);
+
+  // Assign machine configuration.
+  auto machine_configuration = sgx_identity.mutable_machine_configuration();
+  machine_configuration->mutable_cpu_svn()->set_value("0000000000000000");
```
tiziano88: Where does this come from?

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+                             .ValueOrDie();
+  return asylo::SerializeSgxIdentityExpectation(sgx_expectation).ValueOrDie();
+}
+
+asylo::IdentityAclPredicate SgxApplicationClient::CreateSgxIdentityAcl(
+    std::vector<std::string>& mrenclave_strings) const {
+  asylo::IdentityAclPredicate acl;
+  auto acl_predicates = acl.mutable_acl_group();
+  acl_predicates->set_type(asylo::IdentityAclGroup::OR);
+  for (auto mrenclave_str : mrenclave_strings) {
```
tiziano88: Add a comment explaining what is being done here.

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,167 @@
...
+    asylo::IdentityAclPredicate acl_predicate;
+    *acl_predicate.mutable_expectation() = CreateSgxIdentityExpectation(mrenclave_str);
+    *acl_predicates->add_predicates() = acl_predicate;
+  }
+  return acl;
+}
+
+std::shared_ptr<grpc::ChannelCredentials> SgxApplicationClient::CreateChannelCredentials(
+    std::vector<std::string>& mrenclave_strings) const {
+  // TODO: Use remote attestation when it will become available.
```
tiziano88:
```suggestion
  // TODO: Use remote attestation when it becomes available.
```

oak/client/sgx_application_client.cc
ipetr0v: Done

oak/client/BUILD
ipetr0v: Done

oak/client/sgx_application_client.h
ipetr0v: Done

oak/client/sgx_application_client.h
ipetr0v: Done

oak/server/asylo/BUILD
ipetr0v: Done

oak/server/asylo/BUILD
ipetr0v: Done

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,197 @@
...
+    if (std::isxdigit(ch) == 0) {
+      return Status(asylo::error::GoogleError::INVALID_ARGUMENT,
+                    "Hash contains non-hexademical charachters: " + hash_string);
+    }
+  }
+  hash->set_hash(absl::HexStringToBytes(hash_string));
+  return Status::OkStatus();
+}
+
+namespace oak {
```
tiziano88: This should probably be in an unnamed namespace? Also the constants, if they are only used in this file.

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,197 @@
...
+
+constexpr size_t kPerChannelNonceSizeBytes = 32;
+// Debug MRSIGNER value derived from the SGX test key (will change after changing SVN and PRODID).
+// https://github.com/google/asylo/blob/088ea3490dd4579655bd5b65b0e31fe18de7f6dd/asylo/distrib/sgx_x86_64/linux_sgx_2_6.patch#L5481
+const char* kDebugMrSigner = "83d719e77deaca1470f6baf62a4d774303c899db69020f9c70ee1dfc08c7ce9e";
+// Debug CPUSVN value.
+// This value should be provided by EREPORT and it reflects the microcode update version.
+const char* kDebugCpuSvn = "0000000000000000";
+
+// Parses a hexademical hash string from `hash_string` into `Sha256HashProto` in `hash`.
```
tiziano88:
```suggestion
// Parses a hexadecimal hash string from `hash_string` into `Sha256HashProto` in `hash`.
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,197 @@
...
+// TODO: Use same function from Asylo, when it becomes public.
+Status Sha256HashFromHexString(const std::string& hash_string, asylo::Sha256HashProto* hash) {
```
tiziano88: The comment should mention that it expects a string of 64 bytes. Although it seems odd to couple this check to the conversion. Personally I would let it convert any hex string to bytes (like `absl::HexStringToBytes`, but with validity checks and returning `StatusOr<string>` or similar), then have the caller check the length of that output bytes to be 32 (or whatever), then finally put that in the proto. You can definitely have a function `SetValueInProto` that handles all this logic, if you don't want to have separate checks in `CreateSgxIdentityExpectation`. WDYT?

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,196 @@
...
+  }
+  hash->set_hash(absl::HexStringToBytes(hash_string));
+  return asylo::Status::OkStatus();
+}
+
+namespace oak {
+
+SgxApplicationClient::SgxApplicationClient(std::vector<std::string> mrenclave_strings) {
+  // Initialize assertion authorities.
+  this->InitializeAssertionAuthorities();
```
anghelcovici: Nit: do you need `this`?

oak/client/sgx_application_client.h
```diff
@@ -0,0 +1,76 @@
...
+// an attestation in which the Oak enclave hash (MRENCLAVE value) is one of the
+// hashes provided in `mrenclave_strings`.
+// Current implementation uses debug parameters such as zero flags and an SGX test key:
+// https://github.com/intel/linux-sgx/tree/master/psw/ae/ref_le/ref_keys
+// TODO: Change to real values, when Asylo releases a client API.
+class SgxApplicationClient {
+ public:
+  // Initializes an SGX application client.
+  //
+  // `mrenclave_strings` contains all valid Oak enclave hashes (hexademical strings).
```
tiziano88:
```suggestion
  // `mrenclave_strings` contains all valid Oak enclave hashes (hexadecimal strings).
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,196 @@
...
+  // Create a group of ACL values connected with `OR` statements.
+  acl_predicates->set_type(asylo::IdentityAclGroup::OR);
+  for (auto mrenclave_string : mrenclave_strings) {
+    asylo::IdentityAclPredicate acl_predicate;
+
+    // Add a predicate that verifies a single MRENCLAVE value.
+    auto expectation = CreateSgxIdentityExpectation(mrenclave_string);
+    if (expectation.ok()) {
+      *acl_predicate.mutable_expectation() = expectation.ValueOrDie();
+      *acl_predicates->add_predicates() = acl_predicate;
```
anghelcovici: This syntax is funny :)

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,196 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "oak/client/sgx_application_client.h"
+
+#include <memory>
```
anghelcovici: why?

oak/client/sgx_application_client.cc
ipetr0v: Done

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,214 @@
...
+const char* kDebugCpuSvn = "0000000000000000";
+
+// Converts a hexadecimal string `hash_string` to the string of bytes.
+// This function is a wrapper for an absl::HexStringToBytes with a validity check.
+// TODO: Use same function from Asylo, when it becomes public.
+StatusOr<std::string> HexStringToBytes(const std::string& hash_string) {
+  for (auto ch : hash_string) {
+    if (std::isxdigit(ch) == 0) {
+      return Status(asylo::error::GoogleError::INVALID_ARGUMENT,
+                    "String contains non-hexademical charachters: " + hash_string);
```
tiziano88:
```suggestion
                    "String contains non-hexadecimal characters: " + hash_string);
```

oak/client/sgx_application_client.cc
```diff
@@ -0,0 +1,214 @@
...
+    }
+  }
+  return absl::HexStringToBytes(hash_string);
+}
+
+// Parses a hexadecimal hash string from `hash_string` into `Sha256HashProto` in `hash`.
+// `hash_bytes` should be a hexadecimal string that represents a Sha256 hash (64 bytes long).
```
tiziano88: 64 or 32 bytes? The wording is ambiguous, whether it's referring to the string or the hash :)

oak/client/sgx_application_client.cc
ipetr0v: Done

oak/client/sgx_application_client.cc
ipetr0v: Added clarification

ipetr0v:
@anghelcovici currently Asylo doesn't support remote credentials, and null credentials do not use expectations that I initialize in this file.
They said, that I will be able to test this only when they will release remote attestation:
https://groups.google.com/a/google.com/g/enclave-discuss/c/xzoBL2q96iI

PR #338, opened 2019-11-28
Maintain wabt patches in oak tree

blaxill:
- Use http_archive https://github.com/WebAssembly/wabt
- Keep wabt patches in oak tree at third_party/wabt


blaxill:
If the CI looks good I'll remove the commented-out patches

blaxill:
0007-Include-opcode-code-table.c.patch could do with a better description as pointed out by @daviddrysdale. Any ideas beyond "comment out config.h and WABT_STATIC_ASSERT in opcode-code-table.c" ?

WORKSPACE
```diff
@@ -90,11 +90,21 @@ http_archive(
...
-# WebAssembly Binary Toolkit (forked by tiziano88).
-git_repository(
+# WebAssembly Binary Toolkit
+http_archive(
     name = "wabt",
-    commit = "30e914b1630db13080cc054b591ab5822b9b4768",
-    remote = "https://github.com/daviddrysdale/wabt",
+    build_file = "@//third_party/wabt:wabt.BUILD",
+    patches = [
+        "@//third_party/wabt:0003-Add-static-config-file.patch",
```
tiziano88: I think you can remove all the leading `@`.

```suggestion
        "//third_party/wabt:0003-Add-static-config-file.patch",
```

(BTW please do not apply these suggestions from the GitHub UI directly, the UI is quite confusing :) )

third_party/wabt/0003-Add-static-config-file.patch
```diff
@@ -0,0 +1,356 @@
+From 5fcc5f6acaa35422b600efa1c6f42e89012a99a9 Mon Sep 17 00:00:00 2001
+From: Tiziano Santoro <tiziano88@gmail.com>
+Date: Wed, 31 Oct 2018 10:46:27 +0000
+Subject: [PATCH 03/10] Add static config file
+
+---
+ config.h | 337 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
tiziano88: Did you manage to rebuild this using `cmake` as described in https://github.com/project-oak/oak/pull/336#issuecomment-559485683? If so, it may be useful to add those instructions (even at high level) here, so we can rerun it if and when needed.

third_party/wabt/0009-Only-warn-on-NDEBUG-build.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git config.h config.h
+index df9a34e2..d9c68558 100644
+--- config.h
++++ config.h
+@@ -3,7 +3,7 @@
+ 
+ // BEGIN GOOGLE3 MODIFICATIONS
+ #ifdef NDEBUG
+-#error This library uses assert()
++#warning This library uses assert()
```
tiziano88: This patch can probably just be merged with the other one that it applies to.

third_party/wabt/0010-Use-std-stringstream-to-build-unique-name.patch
```diff
@@ -0,0 +1,35 @@
...
++#include <sstream>
+ #include <vector>
+ 
+ #include "src/binary-reader-nop.h"
+@@ -1060,7 +1061,9 @@ Result BinaryReaderIR::OnFunctionName(Index index, string_view name) {
+   int counter = 1;
+   std::string orig_name = dollar_name;
+   while (module_->func_bindings.count(dollar_name) != 0) {
+-    dollar_name = orig_name + "." + std::__cxx11::to_string(counter++);
++    std::stringstream ss;
```
tiziano88: I think this patch overrides patch 006 and makes that obsolete?

third_party/wabt/0003-Add-static-config-file.patch
blaxill: This patch is not my config.h- My config.h was building correctly, but I'm apprehensive to add the instructions as I'm not sure if the `run_examples` failures are related.

tiziano88: Ah I see, makes sense, thanks for clarifying.

tiziano88:
This seems ready for merging, do you need any help with that? If not, feel free to go ahead, probably you should "squash and merge", and make sure the final commit message respects [standard guidelines](https://chris.beams.io/posts/git-commit/) (by default GitHub just concatenates all the messages from the various commits, I think).

Also, as a side note, in the future please create your own fork of the project and use that for development, we try and not create additional branches on the main repo (except sometimes it is useful to do that to test CI or other integrations).

PR #348, opened 2019-12-04
Add clang-tidy to the CI

### LINKED ISSUE #64: Run clang-tidy as part of CI
I found an other GitHub project who were able to get clang-tidy work with Bazel and run it in the CI pipeline: https://github.com/envoyproxy/envoy/issues/921.

We could adopt this too.

ipetr0v:
This change:
- Adds dependency on Kythe
- Adds scripts for compilation database generation
- Adds a `clang-tidy` script

Fixes #64


WORKSPACE
```diff
@@ -104,6 +104,42 @@ http_archive(
...
+    name = "io_kythe",
+    sha256 = "85dac12cdeea46f9e369ff109934aa98967bed1cd7c4c4afdc11577c3f99d31a",
+    strip_prefix = "kythe-4814f9f3fcc05c49fbe11f62f1e58a428048da27",
+    urls = [
+        # Head commit on 2019-12-03.
+        "https://github.com/kythe/kythe/archive/4814f9f3fcc05c49fbe11f62f1e58a428048da27.tar.gz",
+    ],
+)
+
+# Kythe tool dependency.
```
tiziano88: Does `kythe_dependencies` work here, apart from pulling in extra deps? My understanding is that these deps are not actually downloaded unless they are actually needed anyways, or unless we run `bazel sync`, which we probably shouldn't do very often anyways. WDYT?

scripts/run_clang_tidy
```diff
@@ -0,0 +1,29 @@
...
+source "$SCRIPTS_DIR/common"
+
+bazel_build_flags+=(
+    '--experimental_action_listener=@io_kythe//kythe/cxx/tools/generate_compile_commands:extract_json'
+    '--noshow_progress'
+    '--noshow_loading_progress'
+)
+
+# Generate compilation database.
+bazel build "${bazel_build_flags[@]}" -- //oak/...:all -//oak/server/asylo:all
```
tiziano88: Mention in a comment why the asylo package is excluded.

scripts/run_clang_tidy
```diff
@@ -0,0 +1,29 @@
...
+./third_party/kythe/generate_compilation_database.sh
+
+# Run clang-tidy.
+SOURCE_FILES=$(find oak -name "*.cc" -printf '%h/%f\n' | grep -v "oak/server/asylo*")
```
tiziano88:
```suggestion
readonly SOURCE_FILES="$(find oak -name "*.cc" -printf '%h/%f\n' | grep -v "oak/server/asylo*")"
```

scripts/run_clang_tidy
```diff
@@ -0,0 +1,29 @@
...
+clang-tidy -p "$(bazel info execution_root)" -header-filter=.* $SOURCE_FILES
```
tiziano88:
```suggestion
clang-tidy -p "$(bazel info execution_root)" -header-filter='.*' "$SOURCE_FILES"
```

(if you don't quote `.*` doesn't the shell try to expand it?)

third_party/kythe/generate_compilation_database.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Generates a compile_commands.json file at $(bazel info execution_root) for
+# your Clang tooling needs.
+
+BAZEL_ROOT="$(bazel info execution_root)"
```
tiziano88:
```suggestion
readonly BAZEL_ROOT="$(bazel info execution_root)"
```

third_party/kythe/generate_compilation_database.sh
```diff
@@ -0,0 +1,19 @@
...
+pushd "$BAZEL_ROOT" > /dev/null
```
tiziano88: This and `popd` do not seem necessary? Just `cd "$BAZEL_ROOT"` should be enough?

third_party/kythe/generate_compilation_database.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "[" > compile_commands.json
```
tiziano88: Use a readonly variable to hold the file name.

tiziano88:
```suggestion
echo '[' > compile_commands.json
```

third_party/kythe/generate_compilation_database.sh
```diff
@@ -0,0 +1,19 @@
...
+COUNT=0
+find . -name '*.compile_command.json' -print0 | while read -r -d '' fname; do
+  if ((COUNT++)); then
+    echo ',' >> compile_commands.json
+  fi
+  sed -e "s|@BAZEL_ROOT@|$BAZEL_ROOT|g" < "$fname" | \
+  # Clang doesn't recognize '-fno-canonical-system-headers'.
+  sed -e "s|-fno-canonical-system-headers|-no-canonical-prefixes|g" >> compile_commands.json
+done
+echo "]" >> compile_commands.json
```
tiziano88:
```suggestion
echo ']' >> compile_commands.json
```

third_party/kythe/generate_compilation_database.sh
ipetr0v: This script was just copied from Kythe. I'll change to `cd`.

tiziano88: If it was just copied, feel free to leave it as it is (but leave a comment somewhere to say that it is copied). I thought you already modified it a bit, so you may as well modify it even more. TBH I think at this point you can just take it out of third_party and remove the license, it can just become one of our own scripts.

WORKSPACE
ipetr0v: `add_kythe_dependencies` contains `maybe` functions:
https://github.com/kythe/kythe/blob/25c5f5a88520250f9fd3605eb60a420605b1d458/setup.bzl#L30-L36
that are not executed without `bazel sync`, and the attempt to create a compilation database results in:
```
ERROR: error loading package '': in /root/.cache/bazel/_bazel_root/b6bc0828d5b43c7085b650e0698bf388/external/io_kythe/external.bzl: Unable to load file '@rule
s_java//java:repositories.bzl': file doesn't exist
```
So I explicitly added libraries that a necessary for CDB creation.

scripts/generate_compilation_database
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
tiziano88: Follow the same convention as our other scripts if possible (including sourcing the common script).

PR #367, opened 2019-12-10
Update wabt to 1.0.12

blaxill:
- Updates wabt to the most current release 1.0.12
- Removes the remaining patches which no longer seem necessary (pending on CI)
- Updated wabt specific `config.h` by following the wabt build instructions to generate a new `config.h.patch`:
   ~~~
   $ mkdir build
   $ cd build
   $ cmake ..
   $ cmake --build .
   $ diff -Naur /dev/null config.h > config.h.patch
   ~~~
  
#276


tiziano88:
Neat! Could you add the instructions you mentioned here next to the wabt section in the WORKSPACE file?

tiziano88:
BTW I hope this would also fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18296 , let's keep an eye on it.

PR #370, opened 2019-12-12
Add licence in google/rpc

ipetr0v:
This change adds a `licenses` parameter to `package` in `third_party/google/rpc/BUILD`.

### DIFF
--- a/third_party/google/rpc/BUILD
+++ b/third_party/google/rpc/BUILD
@@ -22,6 +22,8 @@ package(
     licenses = ["notice"],
 )
 
+exports_files(["LICENSE"])
+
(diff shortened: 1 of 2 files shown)


ipetr0v:
Added a `LICENSE` file and `exports_files(["LICENSE"])`.

tiziano88:
I thought we have a check for this kind of things already, does that exclude third party?

ipetr0v:
Yes, it looks like it excludes:
https://github.com/project-oak/oak/blob/cc940ccc685c09c54bc3522d6909ef50f4401af9/scripts/check_formatting#L22-L30

PR #376, opened 2019-12-16
Add Android hello-world client

### LINKED ISSUE #353: Implement Android Oak client
Something like

https://github.com/project-oak/oak/blob/39838c8b71dd42731615ebdb09f4d15924e53ec9/oak/client/application_client.h#L47

but usable by Android applications.

I am not sure whether this means it should be implemented natively in Java, or have a thin Java wrapper around the C++ client, or whether we can just offer the C++ version and assume that Android applications will wrap it.

In any case, it requires establishing a channel using [EKEP](https://asylo.dev/docs/concepts/ekep.html).

I note though that there is already a separate [Android gRPC stack](https://developer.android.com/guide/topics/connectivity/grpc), but it does not seem to be as flexible as the regular gRPC stack. In particular, I am not sure it offers any way of using different channel credentials.

@ipetr0v could you start looking into it, and seeing what else is missing? e.g., can we use bazel with a different configuration to check whether some of our targets can be built for Android?

cc @annasapek @KeithMoyer

ipetr0v:
This change adds:
- hello-world Android client that uses JNI with C++ gRPC
- android.Dockerfile with Android SDK and NDK
- GCB command to compile Android client for x86_64
- patch for gRPC that fixes Ares compilation for Android

Fixes #353


examples/hello_world/proto/BUILD
```diff
@@ -33,10 +35,22 @@ cc_proto_library(
...
     grpc_only = True,
     well_known_protos = True,
     deps = ["hello_world_cc_proto"],
 )
+
+java_grpc_library(
+    name = "hello_world_java_grpc",
+    srcs = ["hello_world_proto"],
+    flavor = "lite",  # Lite version is optimized for mobile devices.
+    deps = ["hello_world_java_proto"],
```
conradgrobler: This should include constraints = ["android"]. This ensures that the gRPC library does not depend on any non-lite java proto libraries. This should also use the lite versions to be added above.

examples/hello_world/proto/BUILD
```diff
@@ -33,10 +35,22 @@ cc_proto_library(
     deps = [":hello_world_proto"],
 )
 
+java_proto_library(
```
conradgrobler: This should have a corresponding java_lite_proto_library that can be referenced by the java_grpc_library.

examples/hello_world/proto/BUILD
```diff
@@ -33,10 +35,22 @@ cc_proto_library(
...
+    name = "hello_world_java_proto",
+    deps = [":hello_world_proto"],
+)
+
 cc_grpc_library(
```
conradgrobler: As far as I can tell, there is no option for a 'lite' flavor or adding 'android' as a constraint, so probably not usable for an Android example.

WORKSPACE
```diff
@@ -40,6 +40,21 @@ http_archive(
     ],
 )
 
+# Patch gRPC ares BUILD file.
+# TODO: Remove when Asylo will update gRPC dependency.
```
anghelcovici: Link to an issue? :)

android.Dockerfile
```diff
@@ -0,0 +1,157 @@
+# Download Debian-9 (Stretch) with openjdk-8 that is required for Android.
```
anghelcovici: Wow... this is a long Dockerfile. Is there anything here we can inherit from other files, to void duplication?

examples/hello_world/client/android/BUILD
```diff
@@ -0,0 +1,56 @@
...
+load("@rules_android//android:rules.bzl", "android_binary", "android_library")
+
+package(
+    licenses = ["notice"],
+)
+
+cc_library(
+    name = "native_client",
+    srcs = ["cpp/client.cc"],
+    copts = ["-std=c++11"],
```
anghelcovici: Why do you need this?

examples/hello_world/client/android/cpp/client.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#define APPNAME "Oak JNI"
+#define JNI_LOG(MSG) __android_log_print(ANDROID_LOG_INFO, APPNAME, MSG);
+
+using ::oak::examples::hello_world::HelloRequest;
+using ::oak::examples::hello_world::HelloResponse;
+using ::oak::examples::hello_world::HelloWorld;
+
+// Global channel storage.
+uint CHANNEL_COUNT = 1;
```
anghelcovici: can we use the sized types? (uint32_t and all)

examples/hello_world/client/android/cpp/client.cc
```diff
@@ -0,0 +1,97 @@
...
+std::unordered_map<jint, std::unique_ptr<HelloWorld::Stub>> CHANNEL_MAP;
```
anghelcovici: Is this the coding standard for global variables? (uppercase?)

examples/hello_world/client/android/cpp/client.cc
anghelcovici: Also, can we make this const? :)

android.Dockerfile
```diff
@@ -0,0 +1,157 @@
...
+
+# Install rustup.
+ARG RUSTUP_DIR=/usr/local/cargo
+ENV RUSTUP_HOME $RUSTUP_DIR
+ENV CARGO_HOME $RUSTUP_DIR
+RUN curl https://sh.rustup.rs -sSf > /tmp/rustup
+RUN sh /tmp/rustup -y --default-toolchain=none
+ENV PATH "$RUSTUP_DIR/bin:$PATH"
+RUN rustup --version
+RUN chmod a+rwx $RUSTUP_DIR
```
tiziano88: Why do we need all those deps here?  was expecting this image is only used to build the android targets.

android.Dockerfile
ipetr0v: I have added a separate Dockerfile for Android only because Asylo's Dockerfile depends on Debian-10, that does not have Java-8 (only has Java-11). And Android requires Java-8 and does not support higher versions.

examples/hello_world/client/android/BUILD
ipetr0v: I only used this for `auto`.

examples/hello_world/client/android/cpp/client.cc
ipetr0v: Done

examples/hello_world/client/android/cpp/client.cc
ipetr0v: Deleted

ipetr0v:
Deleted unnecessary tools in `Dockerfile`.

android.Dockerfile
ipetr0v: Deleted unnecessary tools in Dockerfile.

examples/hello_world/client/android/java/com/google/oak/hello_world/res/layout/activity_main.xml
```diff
@@ -0,0 +1,42 @@
...
+        android:layout_height="wrap_content"
+        android:hint="@string/ip_address">
+    </EditText>
+
+    <EditText
+        android:id="@+id/tcpInput"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:hint="@string/tcp_port">
+    </EditText>
```
tiziano88: Not a big deal, but how about just having a single text field for address:port? That's all you are doing anyways with these two text fields it seems. Anyways up to you, the current version is fine.

PR #380, opened 2019-12-17
Update Bazel cache in GCB

### LINKED ISSUE #357: GCB does not update Bazel cache

ipetr0v:
This change adds `BAZEL_GOOGLE_CREDENTIALS` environment variable in `cloudbuild` to update Bazel cache after merges.

Fixes #357

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -96,7 +96,7 @@ timeout: 2h
 options:
   env: 
   # This variable is only defined in the `merge` GCB trigger,
-  # and contains GCB credentials for updating Bazel cache.
+  # and contains GCB credentials for updating the Bazel cache.
   - 'BAZEL_GOOGLE_CREDENTIALS=$_BAZEL_GOOGLE_CREDENTIALS'
   # See: https://cloud.google.com/cloud-build/docs/api/reference/rest/Shared.Types/MachineType
   machineType: 'N1_HIGHCPU_32'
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -17,6 +17,14 @@ steps:
   # See: https://cloud.google.com/cloud-build/docs/create-custom-build-steps
   # Init .git repository used by check_generated
   # Workaround for https://github.com/GoogleCloudPlatform/cloud-builders/issues/236
+  - name: 'gcr.io/oak-ci/oak:latest'
+    id: env_test
+    entrypoint: 'bash'
+    waitFor: ['build_image']
+    timeout: 5m
+    args: ['./scripts/env_test']
+    env:
```
tiziano88: FYI https://cloud.google.com/cloud-build/docs/build-config#options

tiziano88:
I just noticed that the variable value actually shows up in the build configuration: https://pantheon.corp.google.com/cloud-build/builds/2645c947-4116-4f45-b3e4-a039670513e4;tab=detail?project=oak-ci . I think it's fine for now, since we are the only ones with access to it (it's not public), but it seems that that feature is not actually meant to store any kind of secret data.

PR #393, opened 2019-12-19
Add policy tag for modules

michael-kernel-sanders:


oak/proto/policy.proto
```diff
@@ -47,3 +48,11 @@ message GrpcTag {
   // https://github.com/google/asylo/blob/3158887cb768112516424b3e65046f3946eb4465/asylo/identity/identity.proto#L40.
   bytes authorization_bearer_token = 1;
 }
+
+// Policies related to modules, referring to the native WebAssembly node within the TCB.
+message ModuleTag {
```
tiziano88: I think the name should make it clear that this is related to WebAssembly modules specifically. How about `WebAssemblyModuleTag`?

oak/proto/policy.proto
```diff
@@ -47,3 +48,11 @@ message GrpcTag {
...
+  // The attestation for a single WebAssembly module.
```
tiziano88: It is unclear what an "attestation for a module is". Perhaps mention something like it is the sha256 of the module in binary format, or something like that? It's fine if we change our mind later, but currently it seems very abstract.

oak/proto/policy.proto
michael-kernel-sanders: Done.

oak/proto/policy.proto
michael-kernel-sanders: Done.

PR #395, opened 2019-12-19
Attach policy to storage operations

michael-kernel-sanders:


oak/proto/policy.proto
```diff
@@ -47,3 +48,11 @@ message GrpcTag {
   // https://github.com/google/asylo/blob/3158887cb768112516424b3e65046f3946eb4465/asylo/identity/identity.proto#L40.
   bytes authorization_bearer_token = 1;
 }
+
+// Policies related to modules, referring to the native WebAssembly node within the TCB.
+message ModuleTag {
```
tiziano88: How about `WebAssemblyModuleTag`, with a field `web_assembly_module_sha256` or similar?

oak/common/policy.h
```diff
@@ -47,6 +47,10 @@ oak::policy::Label DeserializePolicy(const std::string& serialized_policy);
 // provided authorization bearer token.
 oak::policy::Label AuthorizationBearerTokenPolicy(const std::string& authorization_token);
 
+// Creates a policy that only allows declassifying data for modules that can present the
```
tiziano88: I don't think modules "present" the attestation, rather the Oak Runtime knows the hash of the module making a request, and will allow a module with sha256 `123` to declassify tags `wasm_sha256: 123`. Also note that any node will be able to read the data, it's just the declassification that is gated by the module identity.

michael-kernel-sanders: Revised to indicate matching the attestation instead.

oak/proto/policy.proto
michael-kernel-sanders: Renamed.  It seems like we might want the attestation to be a serialized proto for extensibility?

oak/proto/storage_channel.proto
```diff
@@ -18,26 +18,33 @@ syntax = "proto3";
...
 package oak;
 
+import "oak/proto/policy.proto";
+
 // In all the request messages below, the storage_name field is used to partition data within an Oak
 // Module and is not exposed to the StorageProvider. The StorageChannel generates a unique storage
 // identifier from storage_name and the Oak Module identifier to prevent name collisions among Oak
 // Modules.
 
+message StorageDatum {
```
tiziano88: Maybe `Item` (or `Entry`, or some variation) would be a more familiar term for developers?Vaguely inspired by https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem , which at some point I would like to investigate more (i.e. if we could emulate storage with that in a browser when running client-side... anyway, that's a separate discussion)

oak/server/storage/storage_node.cc
```diff
@@ -78,26 +78,32 @@ void StorageNode::Run(Handle request_handle) {
...
-      storage_processor_.Read(
-          channel_read_request.storage_name(), channel_read_request.datum_name(),
-          channel_read_request.transaction_id(), channel_read_response.mutable_datum_value());
-
+      std::string value;
+      storage_processor_.Read(channel_read_request.storage_name(),
+                              channel_read_request.datum().name(),
+                              channel_read_request.transaction_id(), &value);
+      channel_read_response.mutable_datum()->ParseFromString(value);
+      // TODO: Check security policy for datum.
```
tiziano88: Could you create an issue for this and link it as `TODO(#xxx)` here?

oak/proto/storage_channel.proto
michael-kernel-sanders: Item SGTM.

oak/server/storage/storage_node.cc
michael-kernel-sanders: Done.

PR #397, opened 2019-12-19
Pull Android Docker in GCB

ipetr0v:
This change:
- Adds `docker_pull` for Android SDK in the GCB
- Updates the text field in the hello-world Android application

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -79,13 +79,6 @@ steps:
     timeout: 5m
     entrypoint: 'cp'
     args: ['./bazel-bin/oak/server/asylo/liboak_enclave_unsigned.so', './liboak_enclave_unsigned.so']
-  # Build Android Hello-World application.
-  - name: 'gcr.io/oak-ci/oak-android:latest'
-    id: build_examples_android
-    waitFor: ['build_android_image']
-    timeout: 30m
-    entrypoint: 'bash'
-    args: ['./scripts/build_examples_android']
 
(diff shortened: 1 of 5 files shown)


cloudbuild.yaml
```diff
@@ -7,12 +7,24 @@ steps:
     waitFor: ['-']
     timeout: 10m
     args: ['pull', 'gcr.io/oak-ci/oak:latest']
+  # Pull Android Docker image.
```
tiziano88: I would keep the Android stuff at the end, in fact perhaps adds an extra line between the current steps and the android steps. I'm thinking we may want to split the triggers into two separate cloudbuild files in the short future. WDYT?

examples/hello_world/client/android/java/com/google/oak/hello_world/MainActivity.java
```diff
@@ -46,8 +46,8 @@ public void onCreate(Bundle savedInstanceState) {
 
     // Set default address.
     // Android emulator forwards `10.0.2.2` to the host machine.
```
tiziano88: Where does this `10.0.2.2` come from? Could you link to some documentation?

cloudbuild.yaml

examples/hello_world/client/android/java/com/google/oak/hello_world/MainActivity.java
ipetr0v: Done

PR #399, opened 2019-12-19
Add Oak server Docker image

### LINKED ISSUE #382: Create deployable Docker image
The image should be as minimal as possible, and only contain the main [Oak server binary](https://github.com/project-oak/oak/blob/ad485a84f4095f69bd4e599f4cf30b31980761f3/oak/server/asylo/BUILD#L130-L144) and the enclave binary, plus any necessary parts of the intel SGX toolkit. It should be able to run both the simulated and the real SGX (if available on the physical machine). Note that the base Asylo image that we use currently is not ideal for this, as it has too many additional tools and dependencies installed, but maybe we can reuse the first few Docker steps from it to install the Intel SGX on an empty image.

I am not sure whether we should use Bazel to build the image (via https://github.com/bazelbuild/rules_docker#container_image-1), or do it via a regular Dockerfile.

There are interesting questions about signing the enclave binary: even though we do not rely on `MRSIGNER` at all in Oak, it seems that SGX requires the enclave binary to be signed. We may sign it with a key that we keep internal in Google, or, if we are sure we will never actually rely on it, we should consider creating a private key and checking it in this repository, and use it to sign enclave binaries just for the sake of making SGX happy. @annasapek @keithmoyer or is this a bad idea?

ipetr0v:
This change adds Bazel rules to create a deployable Docker image with Oak server.

Fixes #382

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -46,7 +46,9 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_pull")
 
 container_pull(
     name = "cc_image",
-    digest = "sha256:86f16733f25964c40dcd34edf14339ddbb2287af2f7c9dfad88f0366723c00d7",
+    # Image uploaded on 2019-10-28.
+    # https://pantheon.corp.google.com/gcr/images/distroless/GLOBAL/cc
+    digest = "sha256:f81e5db8287d66b012d874a6f7fea8da5b96d9cc509aa5a9b5d095a604d4bca1",
     registry = "gcr.io",
     repository = "distroless/cc",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -19,6 +19,34 @@ workspace(name = "oak")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# Docker rules should be loaded in the beginning of the WORKSPACE file
+# to avoid diamond dependencies:
+# https://github.com/bazelbuild/rules_docker/blob/bb8da501955e5f7c1f704c50c0e4fce0193b2b2e/README.md#known-issues
+http_archive(
+    name = "io_bazel_rules_docker",
+    sha256 = "df13123c44b4a4ff2c2f337b906763879d94871d16411bf82dcfeba892b58607",
+    strip_prefix = "rules_docker-0.13.0",
+    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.13.0/rules_docker-v0.13.0.tar.gz"],
+)
+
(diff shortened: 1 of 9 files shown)


WORKSPACE
```diff
@@ -19,6 +19,38 @@ workspace(name = "oak")
...
+load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
+
+container_deps()
+
+# Load a Distroless-CC Docker image.
+load("@io_bazel_rules_docker//container:container.bzl", "container_pull")
+
+container_pull(
+    name = "cc_image",
+    digest = "sha256:86f16733f25964c40dcd34edf14339ddbb2287af2f7c9dfad88f0366723c00d7",
```
tiziano88: I guess these images don't really use any meaningful tags, right? https://pantheon.corp.google.com/gcr/images/distroless/GLOBAL/cc

Could you add a comment saying that this is the latest as of 2019-12-19? But actually, is it? It seems there are a few more recent ones.

oak/server/asylo/BUILD
```diff
@@ -142,3 +143,22 @@ enclave_loader(
...
+        "--enclave_path=oak_enclave_debug.so",
+        "--grpc_port=8888",
+    ],
+    files = [
+        "oak_enclave_debug.so",
+        "oak_host_loader",
+    ],
+    ports = [
+        "8080/tcp",  # Application port.
+        "8888/tcp",  # gRPC port.
```
tiziano88:
```suggestion
        "8888/tcp",  # Oak Manager gRPC port.
```

oak/server/asylo/BUILD
```diff
@@ -142,3 +143,22 @@ enclave_loader(
...
+        "./oak_host_loader",
+        "--enclave_path=oak_enclave_debug.so",
+        "--grpc_port=8888",
+    ],
+    files = [
+        "oak_enclave_debug.so",
+        "oak_host_loader",
+    ],
+    ports = [
+        "8080/tcp",  # Application port.
```
tiziano88:
```suggestion
        "8080/tcp",  # Oak Application gRPC port.
```

tiziano88: It would be good to add a comment pointing to where these ports are defined in the code currently.

WORKSPACE
ipetr0v: Done

oak/server/asylo/BUILD
ipetr0v: Done

oak/server/asylo/BUILD
ipetr0v: Done

oak/server/asylo/BUILD
ipetr0v: Done

PR #403, opened 2019-12-19
Remove Cargo.lock files for libraries

blaxill:
- Removes existing Cargo.lock files
- Adds a gitignore filter

In the future we might want to add inclusion rules for Cargo.lock's
associated with binaries


tiziano88:
I actually believe we should keep all the lock files, even for libraries. If we don't, then we may end up pulling in a different version of a dependency between test runs of otherwise unmodified code, leading to spurious failures. Instead we should explicitly control when to upgrade dependencies (via `cargo update`) and create a new commit for this. https://dependabot.com/ is an example of an external service that submits PRs when it notices more recent versions of dependencies, and makes sure that tests are passing, etc., I use it for my personal projects, and we may consider integrating it here perhaps.

blaxill:
Yeah I immediately ran into this problem

PR #410, opened 2019-12-23
Implement alternative API for chat example

tiziano88:
Alternative API for the chat example.

The main goal of this PR is to prototype a different (ideally more ergonomic) way for developers to write Oak applications.

- Change the structure of the example so that the room worker node uses an event-loop style pattern, with the main node sending commands to it. The commands are serialized as bytes + handles using hand-written code, but this could easily be generated automatically, either by writing our own implementation, or using something like FIDL.
- Remove the `#[cfg(target_arch = "wasm32")]` annotation on `oak_main`, since otherwise rust clippy would complain about everything in that file being unused.
- Add `Encode` and `Decode` traits to convert from Rust to bytes + handles and vice versa.
- Remove room names, as they do not contribute to the security of the underlying model.
- Rename `admin_id` to `admin_token`, and require it to be passed in addition to `room_id` for admin operations.
- Unify `ReceivedMessage` and `SentMessage` into a single `Message` type.
- Rename `JoinRoom` method to `Subscribe`.
- Add serializer and deserializer for `Handle` type, converting to and from invalid handles to ensure that unusable handles cannot be serialized (they will be serialized, but obviously unusable when deserialized).


sdk/rust/oak/src/lib.rs
```diff
@@ -42,15 +43,15 @@ pub type Handle = u64;
 /// Wrapper for a handle to the read half of a channel.
 ///
 /// For use when the underlying [`Handle`] is known to be for a receive half.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, PartialEq, Serialize, Deserialize)]
```
daviddrysdale: I think this is a bad idea; if you make it possible to serialize handles, people will be encouraged to serialize handles and then be confused when the deserialized handle doesn't work (because it has crossed a Node boundary and so is no longer valid).

examples/chat/proto/chat.proto
```diff
@@ -21,47 +21,51 @@ package oak.examples.chat;
 import "google/protobuf/empty.proto";
 
 message CreateRoomRequest {
-  // Name of the room; this is only used for UI purposes (and so does not need
-  // to be unique).
+  // Name of the room; this is only used for UI purposes (and so does not need to be unique).
```
daviddrysdale: I couldn't immediately see where the room name is now used &ndash; does it get lost?

examples/chat/proto/chat.proto
```diff
@@ -21,47 +21,51 @@ package oak.examples.chat;
...
+message Message {
+  // Handle for the user; this is only used for UI purposes (and so does not need to be unique nor
+  // persistent).
   string user_handle = 2;
-  string text = 3;
+  string text = 1;
 }
 
-message SentMessage {
+message SubscribeRequest {
```
daviddrysdale: super-nit: diffs would make more sense if `SubscribeRequest` and `SendMessageRequest` were re-ordered to match the messages they're replacing (`JoinRoomRequest` and `SentMessage` respectively).

sdk/rust/oak/src/lib.rs
tiziano88: Good point, here I added these in order to be able to use `serde` for most of the serialisation, and then manually pull out the handles and send them separately. Something like [FIDL](https://fuchsia.dev/fuchsia-src/development/languages/fidl/reference/wire-format) may replace (in fact, combine) the serialisation and sending steps. I will add a comment to clarify this.

examples/chat/proto/chat.proto
tiziano88: I guess it does get lost now. I don't see a strong reason for it over `room_id` now though.

We could (in descending order of my personal preference):

- remove it entirely
- add back proper support for it, adding a new method `GetRoomInfo` or similar that returns the name of the room given the `room_id` (not sure how to actually use it on the client though), and possibly a new type of message to change the room name

@daviddrysdale WDYT?

sdk/rust/oak/src/lib.rs
tiziano88: Added some comments (and also on `Handle`).

tiziano88:
This is not yet working (something hangs when running the examples on the actual Oak server), but I'd like to get people to start reviewing it. High-level comments are also welcome, in particular whether it feels a step in the right direction or not, in terms of dev ux.

Thanks

examples/chat/common/src/lib.rs
```diff
@@ -14,4 +14,39 @@
...
 pub mod proto;
+
+use anyhow::{anyhow, Context, Result};
+
+/// A trait for objects that can be encoded as bytes + handles.
+pub trait Encodable {
+    fn encode(&self) -> Result<(Vec<u8>, Vec<oak::Handle>)>;
+}
+
+/// A trait for objects that can be decoded from bytes + handles.
```
daviddrysdale: This is nice &ndash; I think it should live under `sdk/rust/oak/` somewhere for more general use (but that can be a separate PR).

sdk/rust/oak/src/lib.rs
daviddrysdale: I still think this is a big footgun, so maybe add an issue to go with the TODO?

examples/chat/proto/chat.proto
```diff
@@ -21,47 +21,51 @@ package oak.examples.chat;
...
+  bytes room_id = 1;
+  bytes admin_token = 2;
 }
 
-message JoinRoomRequest {
+message SubscribeRequest {
   bytes room_id = 1;
-  // Handle for the user; this is only used for UI purposes (and so does not
-  // need to be unique nor persistent).
-  string user_handle = 2;
```
daviddrysdale: Removing this removes the ability for the client UI to emit messages like "Oak Discussion: tiziano88 has joined the room", so again seems to be a reduction in functionality in exchange for a more minimal API.

(Again, modulo the fact that user names are entirely spoofable like room names are.)

examples/chat/proto/chat.proto
daviddrysdale: The room name was aimed at giving humans a way to distinguish between rooms; "NKsceNlg69UbcvryfzmFGnMv9qnZ0DYh6u6gJxujnPPxvHsxMehoD368sumKawVaq9WaSkzrcStoNYLvVNdzhA==" isn't very friendly but "Oak Discussion" is more comprehensible (albeit spoofable).

So removing the room name feels like we're dropping functionality from what is already a fairly limited demo.  I was kind of expecting that the chat app would get extended (e.g. with an actual UI) rather than stripped down, so the latter option makes more sense to me.

examples/chat/proto/chat.proto
tiziano88: I think eventually the streaming API must be complemented by a REST API that lets a user retrieve all messages (perhaps with a limit) for a room that they have permission to access. What controls the set of recipients of a chat room should (eventually) be the policies associated with each message, not the set of clients with a currently open gRPC connection. The streaming API may be a nice optimisation to allow receiving new messages more quickly, but I don't think it can be the only (or even main) way of interacting with the app. 

In fact perhaps the API may be:

```protobuf
service Chat {
  // Create a new room with the provided fixed set of participants.
  rpc CreateRoom(CreateRoomRequest) returns (CreateRoomResponse);
  // Retrieve rooms that the current user has access to. Note this does not actually retrieve the messages in the rooms, just their metadata, including the list of participants (which is public).
  rpc GetRooms(GetRoomsRequest) returns (GetRoomsResponse);
  // Retrieve all messages for the specified room. It requires the client to have permission to read such messages, via some authentication mechanism (e.g. public key, OpenID connect token)
  rpc GetMessages(GetMessagesRequest) returns (GetMessagesResponse);
  // Send a message to the specified room. The client must attach a label to the message corresponding to the list of participants with which the room was created.
  rpc SendMessage(SendMessageRequest) returns (SendMessageResponse);
}
```

WDYT?

sdk/rust/oak/src/lib.rs
tiziano88: Sure, though I think the main issue is that `Handle` itself is just a u64 and therefore already trivially serializable (even if we don't specify `derive(Serialize)` on that type).

Created https://github.com/project-oak/oak/issues/427 to track and updated TODOs.

examples/chat/proto/chat.proto
tiziano88: /cc @ipetr0v as he has been looking into implementing the actual Android UI for the hypothetical chat app

examples/chat/module2/rust/src/lib.rs
```diff
@@ -14,76 +14,70 @@
...
+#[derive(Default)]
+struct Room {
+    messages: Vec<Message>,
+    clients: Vec<oak::grpc::ChannelResponseWriter>,
+}
 
-    let mut rsp_writers: Vec<oak::grpc::ChannelResponseWriter> = Vec::with_capacity(1);
-    loop {
+impl Room {
+    fn event_loop(mut self, in_handle: u64) -> i32 {
```
conradgrobler: I think the new approach is a step in the right direction, but an event loop is still not a very natural or intuitive approach for most developers. In a future version we should look at also automatically generating this code so the developer only has to write the appropriate event handlers and not have to think about the mechanics of how they are triggered.

examples/chat/common/src/msg.rs
```diff
@@ -0,0 +1,45 @@
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+pub enum Msg {
```
conradgrobler: It might be worth considering a different name for the enum (perhaps Request or Command or something similar), as the name confused me a bit when I read the code for the first time due to overloading of the term 'message'.

sdk/rust/oak/src/lib.rs
blaxill: Ref for tracking: #429 will implement opaque & non-serializable `Handle`s

examples/chat/common/src/msg.rs
tiziano88: Agreed, renamed to `Command`.

examples/chat/proto/chat.proto
tiziano88: Note: I am not suggesting we change the API in this PR, but wanted to give a heads-up on what I think the API may look like in order for it to be used by a realistic looking Android app that we can put together as a demo.

@daviddrysdale as we discussed yesterday, maybe the API should keep the `Subscribe` method in addition to `GetMessages`, since polling would be inconvenient in practice, and also it's nice to demonstrate streaming capabilities of the application.

examples/chat/proto/chat.proto
tiziano88: At this point the room name (and also the user handle) are effectively just part of the payload, it does not seem to matter much whether they are separate fields or just have the client format messages as "user_handle@room > text". The advantage of the current version is that we don't need separate types for `SentMessage` and `ReceivedMessage`, each user just sends a `Message`, and it is delivered as it is to the other users. Even the user handle seems a bit confusing, since as you say it does not actually correspond to anything (and I would be in favour of removing it at some point, in favour of something label-based that can be trusted, if we can come up with it -- perhaps integrity labels could play a role here).

tiziano88:
Reviewers PTAL and approve if you are happy with this, I have other changes in the pipeline that I want your opinion on (e.g. changing the gRPC API) but I'll propose them in a separate PR.

docs/programming-oak.md
```diff
@@ -379,20 +379,26 @@ each chat room and saves off the write handle that will be used to send messages
 to the room:
 
 <!-- prettier-ignore-start -->
-[embedmd]:# (../examples/chat/module_0/rust/src/lib.rs Rust /.*Create a new Node/ /^ +\);$/)
+[embedmd]:# (../examples/chat/module_0/rust/src/lib.rs Rust /.*create_chat_room()/ /\}$/)
```
daviddrysdale: This comment is showing in the HTML output &ndash; I think the parentheses need escaping.

examples/chat/module_0/rust/src/lib.rs
```diff
@@ -14,159 +14,133 @@
...
     fn new() -> Self {
         oak_log::init_default();
-        Node {
-            room_to_admin: HashMap::new(),
-            admin_to_room: HashMap::new(),
-            rooms: HashMap::new(),
-        }
+        Node::default()
     }
     fn invoke(&mut self, method: &str, req: &[u8], writer: grpc::ChannelResponseWriter) {
```
blaxill: If the standard method for sdk usage is to have invoke calling dispatch, perhaps the `#[derive(OakExports)]` should also generate an invoke method, completely hiding this from the user. At that point we are in quite a good place in terms of hiding complexity to the developer. If we require `#[derive(OakExports,Default)]` for the SDK as per this example we could derive `new` as well. This isn't too crazy as we expect Nodes to configure themselves after reading values from their initial channel.

examples/chat/common/src/proto/chat.rs
```diff
@@ -48,33 +47,7 @@ impl CreateRoomRequest {
         ::std::default::Default::default()
     }
```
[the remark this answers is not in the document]
tiziano88: That's a good question, and I think the answer is "no", since for real applications (e.g. for Android) the client will still be protobuf / gRPC (or JSON / REST, I guess). Which is why although I like FIDL I am still a bit hesitant to push forward with it, since we would still need to support gRPC.

examples/chat/module_0/rust/src/lib.rs
tiziano88: Yes we should consider that. Feel free to try out new APIs / patterns if you have any ideas, and we can discuss in a PR or issue.

examples/chat/module_1/rust/src/lib.rs
```diff
@@ -14,79 +14,76 @@
...
+    std::panic::catch_unwind(|| {
+        let room = Room::default();
+        room.event_loop(handle)
+    })
+    .unwrap_or_else(|_| oak::OakStatus::ERR_INTERNAL.value())
 }
 
-pub fn main(in_handle: u64) -> i32 {
-    oak_log::init_default();
-    oak::set_panic_hook();
```
daviddrysdale: Have these two lines been lost along the way (and with it the ability to log from the chat frontend Node)?

sdk/rust/oak/src/lib.rs
```diff
@@ -43,6 +44,44 @@ pub struct Handle {
     id: u64,
 }
 
+/// Serialize a `Handle` as the invalid handle value.
```
daviddrysdale: I don't think I really understand what this is for, but on the face of it it appears to be a dreadful footgun.

Will this allow users to serde-serialize structures holding `oak::Handle`s but will silently drop those handles?  So the type system tells the user that everything is hunky-dory and type-safe but is actually throwing things away under the covers?

(Or am I completely misunderstanding what's going on?)

examples/chat/proto/chat.proto
daviddrysdale: Adding more functionality (like `GetMessages`) seems sensible, in line with making this closer to a real chat app, to confirm that such a thing works fine in Oak.  

I guess there's a tension between that, and making the example app as simple as possible so it can be used as a testbed for exploring how the policy system might work. 

The latter is very much in line with what the example has been used for so far, as a vehicle for confirming that various parts of the runtime (dynamic node creation, persistent server-side streaming, multi-client parallel use, multi-Application parallel use) work OK.

So I guess removing room names and joining announcements is fine given that aim, but at some point we might have to reinstate equivalent functionality.

examples/chat/client/chat.cc
```diff
@@ -273,16 +269,12 @@ int main(int argc, char** argv) {
...
   std::unique_ptr<Room> room;
-  if (!room_name.empty()) {
-    room = absl::make_unique<Room>(stub.get(), room_name);
+  if (room_id.empty()) {
+    room = absl::make_unique<Room>(stub.get());
     room_id = room->Id();
     LOG(INFO) << "Join this room with --app_address=" << addr
               << " --room_id=" << absl::Base64Escape(room_id);
-  } else if (room_id.empty()) {
-    LOG(WARNING) << "Neither room name nor room ID specified, exiting";
```
daviddrysdale: This seems to work with CI, but I don't quite follow *how* &ndash; with this removed, I've have expected the `run_examples` script to hang on this example.

sdk/rust/oak/src/lib.rs
tiziano88: Yes, this allows structs containing `Handle`s to be serialized via serde, and it will silently convert them to invalid handles. Which is exactly what I do now in the example, in which I encode the command as bytes via bincode (using serde), and then manually pull out handles from the original struct, and put them back together on the other side.

FTR, I don't think this is any less type safe than creating `Handle` from a raw number?

I cannot think of a better solution that has the flexibility of serializing large structs with arbitrary handles using existing formats (bincode, json), but I'm happy to hear suggestions.

Obviously FIDL solves this problem (except that it uses its own type for handles, and we would still have to manually convert to and from it, or fork FIDL to use our type).

examples/chat/proto/chat.proto
tiziano88: Thanks for your thoughts, I'll follow up with another PR to discuss the exact API that we may use for a "real" chat app.

examples/chat/module_1/rust/src/lib.rs
tiziano88: Good catch, added back to `oak_main`.

examples/chat/common/src/lib.rs
```diff
@@ -14,4 +14,39 @@
...
+    fn decode(bytes: &[u8], handles: &[oak::Handle]) -> Result<Self>;
+}
+
+pub fn send<M: Encodable>(write_handle: oak::WriteHandle, msg: M) -> Result<()> {
+    let (bytes, handles) = msg.encode().context("could not encode object")?;
+    let status = oak::channel_write(write_handle, &bytes, &handles);
+    if status == oak::OakStatus::OK {
+        Ok(())
+    } else {
+        Err(anyhow!("could not write to channel: {:?}", status))
```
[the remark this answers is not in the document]
tiziano88: Indeed, added TODOs.

examples/chat/client/chat.cc
tiziano88: Ah, I think https://github.com/project-oak/oak/pull/433 actually broke this expectation, since Rust modules are not called `module` any more. I will follow up with a separate PR (see https://github.com/project-oak/oak/issues/458), though I am not sure what to do with this example. Maybe add a `--dry-run` flag?

examples/chat/README.md
```diff
@@ -19,15 +19,15 @@ The first client to run typically creates the Oak Application and a first chat
 room inside it:
 
 ```bash
-$ ./examples/chat/run -C --room_name=MyRoom
+$ ./examples/chat/run
```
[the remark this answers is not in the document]
tiziano88: See other comment, I think we should either add a `--dry-run` flag, or take this out of the `run_examples` script, as it's really more of an interactive thing anyways.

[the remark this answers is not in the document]
daviddrysdale: Yeah, the invocation via `run_examples` was pretty minimal, but it did at least ensure that the Node code was still building and the frontend Node was able to be loaded and its `oak_main` function executed &ndash; I'd be loathe to lose that. 

How would a `--dry_run` flag work with `run_examples`?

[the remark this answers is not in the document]
tiziano88: `--dry_run` (not the best name admittedly) would load the code to the oak runtime, and then either not interact with it at all, or interact minimally from the client point of view. 

Note that when #431 is done we could just load the application and check that it starts up fine (i.e. the server does not immediately crash), which would bring us back on par to what we have now.

Anyway let's discuss that in #458 to avoid this PR becoming even larger?

daviddrysdale:
BTW, I was just re-reading the PR description and I noticed a couple of minor things.

> It changes the structure of the example so that incoming connections end up on an "entry" node, which in turn then creates worker nodes on demand, one per chat room. Users subscribe to a specific chat room, and receive new messages in streaming when messages are posted to that room.

This part seems like a description of #379 not this PR?

> I also had to remove the #[cfg(target_arch = "wasm32")] annotation on oak_main, since otherwise rust clippy would complain about everything in that file being unused.

That was there so that a (potential future) test-only run of the application wouldn't end up with duplicate `oak_main` symbols when 2 Nodes are running in the same application (cf. the ["Testing Multi-Node Applications"](https://github.com/project-oak/oak/blob/master/docs/programming-oak.md#testing-multi-node-applications) section of the 'Programming Oak' doc). 

As there isn't yet a test, removing it is fine but if / (hopefully) when we add multi-Node tests for the chat app we'll need to revisit (and also figure out why clippy is complaining now when it wasn't before).

tiziano88:
> That was there so that a (potential future) test-only run of the application wouldn't end up with duplicate `oak_main` symbols when 2 Nodes are running in the same application (cf. the ["Testing Multi-Node Applications"](https://github.com/project-oak/oak/blob/master/docs/programming-oak.md#testing-multi-node-applications) section of the 'Programming Oak' doc).

I think we will have to revisit that at some point, probably we won't be able to have multiple nodes.
 
> As there isn't yet a test, removing it is fine but if / (hopefully) when we add multi-Node tests for the chat app we'll need to revisit (and also figure out why clippy is complaining now when it wasn't before).

My understanding is that there was no struct in that crate that clippy could complain about unused methods. And in `module_0`, `oak_main` is generated by `OakExports`, which marks it as `extern` unconditionally, so I guess clippy will assume it may be invoked from the outside.

examples/chat/README.md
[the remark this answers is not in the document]
daviddrysdale: Sorry, I wasn't clear about my question about `run_examples`: how would `run_examples` pass `--dry_run` to *just* this example, without also passing an unknown flag to all the other examples?  Are you expecting to add a special case in `run_examples` that watches out for this specific example?

BTW I've fixed #458, so I think something is needed here before this PR would be mergeable.

sdk/rust/oak/src/lib.rs
daviddrysdale: (From discussion).

So I wondered if we could have a trait `Encodable` with a method `encode` to bytes and handles (like in `chat/common/src/lib.rs`) that re-uses `Serialize` where available.  So POD types become automatically `Encodable` (because they're `Serialize`able) and Oak Handle structs are explicitly `Encodable`, meaning that `struct`s that combine the two are `Encodable` &ndash; but are **not** `Serialize`able.  But I don't know enough about trait inheritance/borrowing to immediately know if this is feasible or not&hellip;

examples/chat/README.md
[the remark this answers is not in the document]
tiziano88: Merged and temporarily removed the chat example from `run_examples`, and created #462 to track migration of examples to use the single-app server, which should make it easier to at least make sure that they all compile and start up correctly (even if they cannot be interacted with).

sdk/rust/oak/src/lib.rs
tiziano88: Created #463 to track, thanks for writing this down!

tiziano88:
@daviddrysdale @anghelcovici can I have approval?

docs/programming-oak.md
```diff
@@ -379,20 +379,15 @@ each chat room and saves off the write handle that will be used to send messages
...
-            req.admin_id.clone(),
-            Room {
-                name: req.name,
-                channel: wh,
-            },
-        );
+        Room {
+            channel: wh,
+            admin_token,
+        }
```
anghelcovici: I still find this code a bit cumbersome :(
- I do get why we use halfs, but maybe we need to encapsulate that better? I don't find it at all natural that the function called channel_create returns a tuple and that I have to reason about members separately. 
- If the pattern is to create node with rh and then close it, why not encapsulate that?

examples/chat/client/chat.cc
```diff
@@ -75,75 +74,73 @@ class Safe {
   T val_ GUARDED_BY(mu_);
 };
 
-void Prompt(std::shared_ptr<Safe<std::string>> room_name, const std::string& user_handle) {
-  std::cout << room_name->get() << ":" << user_handle << "> ";
-}
+void Prompt(const std::string& user_handle) { std::cout << user_handle << "> "; }
 
 void ListenLoop(Chat::Stub* stub, const RoomId& room_id, const std::string& user_handle,
```
anghelcovici: Nit: Is this a Loop because we have a while inside? I would feel more natural to call this OnReceive, OnSend or something.

docs/programming-oak.md
daviddrysdale: - Yes, I think we're expecting higher layers of abstraction to appear in the SDK to hide these details.  But writing a few examples helps us figure out the patterns that are best to abstract.
 - The pattern of {create channel, create node with read half, drop local read half} is definitely a candidate for one of those higher level helpers.
 - "room-config" is the name of the config that describes a per-Room Wasm Node (the config currently holds the Wasm module bytes for the Node, and [may soon](https://github.com/project-oak/oak/pull/464) also hold the name of the main entrypoint)

examples/chat/client/chat.cc
daviddrysdale: Yes, it's got `Loop` in the name because it does (and is run in a thread of its very own as a result). 

(For me, a function named `OnReceive` would imply something that's called by some external scheduler when a message is received.)

tiziano88:
> Rather grudging approval, I'm afraid. The gain of having a new way to encapsulate IPC between the Chat Nodes doesn't seem to be worth what's lost along the way (room name functionality, person joining announcements, CI builds / dry-runs of the application).

Thanks, I appreciate the time and effort that was put into writing and reviewing the original example and this PR.

A couple of follow-ups:

- room name functionality and person joining announcements: the point of this (and other) example IMO is not to make a production-ready chat app that can be used by actual users, but rather a testbed on which to prototype things like IPC and policies, so I think it is achieving this goal by generating discussions exactly like this (and I'm sure more to come), for which I am thankful to you and other reviewers. We could have kept the original chat app, and came up with yet another example to try out different IPC approaches, and ended up with each example using its own version of it; it may have been less controversial, but I think we (at least, I) would not have learned so much from it.

- CI builds / dry-runs: this is going to be re-instated as part of #462 .

> However, it's just an example application after all, and I may have spent more time reviewing this than I did writing the chat application in the first place!

Obviously even this PR was meant as a very incremental change, not something we should consider final.

> When this is merged, please could you re-open #427?

I'm fine re-opening, but TBH I am not sure we will be able to do much about it, without reinventing / forking `serde` or `FIDL`. But let's keep discussing there.

docs/programming-oak.md
anghelcovici: > "room-config" is the name of the config that describes a per-Room Wasm Node 

Should we have this strongly typed? I don't like passing strings :)

anghelcovici: > higher layers of abstraction to appear in the SDK 

I don't really understand the need for this PR in that case :)

examples/chat/module_0/rust/src/lib.rs
```diff
@@ -14,159 +14,133 @@
...
-        self.admin_to_room
-            .insert(req.admin_id.clone(), req.room_id.clone());
         Ok(Empty::new())
     }
 
     fn destroy_room(&mut self, req: DestroyRoomRequest) -> grpc::Result<Empty> {
-        if !self.rooms.contains_key(&req.admin_id) {
-            return unknown_id_err("Admin");
+        info!("destroying room");
+        match self.rooms.entry(req.room_id.clone()) {
```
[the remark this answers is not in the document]
tiziano88: Removed.

examples/chat/module_0/rust/src/lib.rs
```diff
@@ -14,159 +14,133 @@
...
+            Entry::Occupied(e) => {
+                if e.get().admin_token == req.admin_token {
+                    // Close the only input channel that reaches the per-room Node, which
+                    // will trigger it to terminate.
+                    oak::channel_close(e.get().channel.handle);
```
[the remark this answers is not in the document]
tiziano88: I don't think we support terminating nodes "from the outside" yet. They have to spontaneously decide to terminate.

examples/chat/module_0/rust/src/lib.rs
```diff
@@ -14,159 +14,133 @@
...
-        let user_id = self.admin_to_room.get(&req.admin_id).unwrap().clone();
-        self.admin_to_room.remove(&req.admin_id);
-        self.room_to_admin.remove(&user_id);
-        self.rooms.remove(&req.admin_id);
-        Ok(Empty::new())
     }
 
-    fn join_room(&mut self, req: JoinRoomRequest, mut writer: grpc::ChannelResponseWriter) {
-        let admin_id = match self.room_to_admin.get(&req.room_id) {
+    fn subscribe(&mut self, req: SubscribeRequest, mut writer: grpc::ChannelResponseWriter) {
```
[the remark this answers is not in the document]
tiziano88: The intention here is to clarify that this method is just a convenience to asynchronously receive future messages.

It will not cause policies to change or anything that other participants may care about: participants must have set labels on the messages accordingly from the start, so that all the intended readers are able to get them and subscribe to updates.

e.g. A wants to send a message to B and C, they will tag it as "A or B or C", and from that point on any of A B or C may retrieve it (either in batch, or asynchronously using the subscribe stream). D will not be able to "join" the room; if A B and C want to communicate with D, they will have to create a new room, and attach the label "A or B or C or D" to future messages.

Does it make more sense in this context?

docs/programming-oak.md
tiziano88: > > higher layers of abstraction to appear in the SDK
> 
> I don't really understand the need for this PR in that case :)

It's a prototype for how we want those higher layer to look like. We could have done it in a separate example, but I thought it would be more useful to operate on an existing example. It sounds like I was wrong :) I'll try and do experiments in separate, perhaps more focused examples in the future, if it helps, e.g. create an IPC example and only hack on IPC stuff there. Would this help?

blaxill: > Should we have this strongly typed? I don't like passing strings :)

IIRC this is passed into the oak runtime via protobuf'ed ApplicationConfiguration which specifies the names as strings. Of course the SDK can provide a strongly typed interface here, for the pseudo nodes it would be easy, for developer created nodes it would require boiler plate generation, but may be possible.

blaxill:
@anghelcovici 

> Based on @daviddrysdale comment, this might be the first step and we need another layer of abstraction, but I do not understand why there are two layers :)

I think David was just commenting that we intend the SDK to evolve to be "higher level" and hide these details rather than additional layers within the SDK. There are however two layers that a developer can use, the low level Wasm bindings and the higher level SDK.

examples/chat/module_1/rust/src/lib.rs
```diff
@@ -14,78 +14,78 @@
...
+                        .expect("could not parse message from bytes");
+                    self.messages.push(message.clone());
+                    info!("fan out message to {} clients", self.clients.len());
+                    for writer in &mut self.clients {
+                        // TODO: Improve error handling.
+                        writer
+                            .write(&message, oak::grpc::WriteMode::KeepOpen)
+                            .expect("could not write to channel");
+                    }
+                }
```
tiziano88: @anghelcovici @daviddrysdale I am not sure it was apparent, but I think the main contribution of this PR is to encapsulate the core logic of the interaction between nodes in this `match` expression, and have algebraic data types (serializable enums containing handles in this case) drive the state changes. This is somewhat similar to https://guide.elm-lang.org/architecture/ and Redux, and also not too far from what happens behind the scenes in gRPC or FIDL when interacting using services or protocols. The point is that now almost all the rest of the code could be at least in principle generated, and developers would only write this command dispatching logic. This of course may be abstracted even more, but a layering is emerging that hopefully should make that easier in the future.

PR #414, opened 2019-12-24
Export license in BUILD

ipetr0v:
This change exports the `LICENSE` file in the root `BUILD`.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -21,3 +21,5 @@
 package(
     licenses = ["notice"],
 )
+
+exports_files(["LICENSE"])


ipetr0v:
I just merged an internal and an external root `BUILD` files.

tiziano88:
That does not really explain much to be honest :) I can kind of guess what this is for, but my point (and general advice) is that it should be explained in the commit message so that it's more obvious to people looking at why this line was added in the future. Or even better, add a comment next to it so that future readers don't even need to look at the history. Makes sense?

ipetr0v:
Done in https://github.com/project-oak/oak/pull/415

Issue #426, opened 2020-01-06
evaluate FIDL for inter-node communication


tiziano88:
Our messaging system relies on sending bytes and handles separately, but we don't yet have a good mechanism for serialising and deserialising data structures from Rust or C++ into these components, so we currently do this by hand and inconsistently in our examples.

[FIDL](https://fuchsia.dev/fuchsia-src/development/languages/fidl/reference/wire-format) seems to address this by providing a mechanism for serialising and also for performing RPCs and message based communication, so it may be useful to look into it and see whether we can integrate it in Oak (both in the runtime, and also as an option for developers developing applications for Oak).

related to #420
blocks #389

tiziano88:
@wildarch I wonder whether this is something you would be interested in looking into? I can provide more details in person the coming week, but I think this is potentially a good issue because:

- it is interesting / non trivial
- it is pretty Rust-specific
- it is not currently blocking other parts of the project (so it would work well as an ongoing 20% kind of task)

For now perhaps you can read up on [FIDL](https://fuchsia.dev/fuchsia-src/development/languages/fidl/reference/wire-format) and take a look at some of the previous related PRs (some merged, some abandoned), e.g.: #442 #444 #445

wildarch:
Sounds interesting! I'll take a look this week 😄

wildarch:
After some discussion with @tiziano88 I have decided to experiment with passing channels over GRPC/Protobuf first. Tracking that work in #673.

I will come back to working on this issue depending on the outcome of the experiment.

The main reason I'm hesitating to bet on using FIDL is that it is currently very much tied to the Fuchsia ecosystem, if this changes then FIDL might become a much more appealing alternative.

wildarch:
Since #1108 we support passing handles between nodes as fields in protocol buffers, so I don't think there is a need to still look into FIDL.

tiziano88:
Well, FIDL would still be much more expressive than protobuf I think, having generics, and the ability to pass references to other services in a type safe way. I would say we would need similar features either in protobuf, or for Rust structs, at some point.

PR #444, opened 2020-01-08
Install Fuchsia core SDK in Docker image

tiziano88:
We will need the `fidlc` and `fidlgen` binaries from it.

ref #426


tiziano88:

/cc @conradgrobler

PR #445, opened 2020-01-08
Import fidl Rust crate and dependent crates

tiziano88:
The packages are all unmodified, apart from adding a `README.md` file
each to clarify provenance, and a `Cargo.toml` file when necessary.

ref #426


tiziano88:

PR #469, opened 2020-01-14
Use `runner` in cloud build

tiziano88:


tiziano88:
I will follow up with a more up-to-date PR for this when `runner` supports more workflows (e.g. running tests, running examples, etc.)

PR #474, opened 2020-01-15
Use single app server in examples

### LINKED ISSUE #462: Refactor examples to use single-app Oak Server
The examples should now use the single-app server implemented as part of #383. This will allow all examples to have the (server-side) application code loaded, and then in some cases a client may also be used to interact with it.

Even better, I think some examples may be better suited to be rewritten as tests, running against an already-provisioned oak application.

For instance, the following may be rewritten as a test that actually asserts on the produced output:

https://github.com/project-oak/oak/blob/a637f9721562f88c51b1762a2308588f86d42165/examples/running_average/client/running_average.cc#L84-L100

ipetr0v:
This change refactors examples to use a single application server.

Fixes #462

Documentation is updated in #590


examples/abitest/client/abitest.cc
```diff
@@ -69,81 +57,21 @@ static bool run_tests(OakABITestService::Stub* stub, const std::string& include,
...
-  std::stringstream addr;
-  addr << "127.0.0.1:" << create_rsp.grpc_port();
-  std::string application_id(create_rsp.application_id());
-  LOG(INFO) << "Connecting to Oak Application id=" << application_id << ": " << addr.str();
+  std::string address = absl::GetFlag(FLAGS_address);
+  LOG(INFO) << "Connecting to Oak Application: " << address;
 
   oak::ApplicationClient::InitializeAssertionAuthorities();
 
   // Connect to the newly created Oak Application.
```
tiziano88: Update or remove doc (the application is not newly created any more).

examples/hello_world/config/BUILD
```diff
@@ -0,0 +1,48 @@
...
+serialized_config(
+    name = "config_cpp",
+    modules = {
+        "app": "//examples/hello_world/module/cpp:hello_world.wasm",
+    },
+    textproto = ":config.textproto",
+)
+
+serialized_config(
+    name = "storage_config",
```
tiziano88: `translator_config`? or rather, `config_translator` for consistency with the other rules? Also rename the textproto file to be `config_*`.

examples/tensorflow/module/cpp/tensorflow.cc
```diff
@@ -91,36 +91,45 @@ uint32_t channel_write(uint64_t handle, uint8_t* buff, size_t usize, uint8_t* ha
...
 
   while (true) {
     int32_t result = wait_on_channels(handle_space, 1);
     if (result != oak::OakStatus::OK) {
       return;
     }
 
-    uint64_t rsp_handle;
-    channel_read(grpc_in_handle, _buf, sizeof(_buf), &actual_size, &rsp_handle, 1, &handle_count);
+    // Reading from main channel should give no data and a (read, write) pair of handles.
```
tiziano88:
```suggestion
    // Reading from main channel should return no bytes and a (read, write) pair of handles.
```

scripts/build_example
```diff
@@ -41,28 +41,46 @@ else
...
+        cargo build --release --target=wasm32-unknown-unknown --manifest-path="$module"
     done
-fi
 
-bazel_build_flags+=(
-    '--symlink_prefix=bazel-client-'
-    "--compilation_mode=$COMPILATION_MODE"
-)
+    # Serialize application configuration for Rust module.
+    if [[ "hello_world" != "$TARGET" ]]; then
```
tiziano88: Could you invert this condition, instead of having to consider a double negative in the else case? Also below in other scripts.

scripts/build_example
```diff
@@ -41,28 +41,46 @@ else
...
+        bazel --output_base="$CACHE_DIR/client" build "${bazel_build_flags[@]}" "//examples/$TARGET/config:config"
+    else
+        # Build translator.
```
tiziano88: Maybe explain a bit more why this is needed? Does not need to describe the entire architecture in detail, but at least justify why the build looks different in this case.

scripts/build_example
```diff
@@ -41,28 +41,46 @@ else
...
+            "//examples/$TARGET/config:storage_config"
+    fi
+fi
 
 readonly CPP_FOLDER="$(find examples/"$TARGET" -type d -name cpp)"
 if [[ -n "$CPP_FOLDER" ]]; then
+    # `config_cpp` depends on a Wasm module, so it should be built with `wasm32` or `emscripten`.
+    readonly BUILD_RULES=(
+        "//examples/$TARGET/module/cpp:all"
+        "//examples/$TARGET/config:config_cpp")
```
tiziano88: (nit) Move `)` to new line.

scripts/run_examples_dev
```diff
@@ -7,37 +7,56 @@ source "$SCRIPTS_DIR/common"
...
-SERVER_PID=$!
 # TODO: Ensure that background processes are killed with something like `trap cleanup_fn EXIT`.
 
 # Run Rust-based Oak examples against a common Oak server instance.
-# Note that the chat example cannot be run because the client is interactive.
-# TODO(#462): Re-enable chat example, even if it may be just loading the server-side code and not
-# interact with it.
-readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq | grep -v chat)"
+readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq)"
+
```
tiziano88: Is this script mostly a copy of `run_examples`, but with a different runner? If that's literally the only difference, I guess they should be refactored to share most of the logic. Perhaps create a separate issue if it's not practical to do in this one?

examples/abitest/client/abitest.cc
ipetr0v: Done

examples/hello_world/config/BUILD
ipetr0v: Done

examples/tensorflow/module/cpp/tensorflow.cc
ipetr0v: Done

scripts/build_example
ipetr0v: Done

scripts/build_example
ipetr0v: Done

scripts/build_example
ipetr0v: Done

scripts/run_examples_dev
ipetr0v: Created an issue: https://github.com/project-oak/oak/issues/595

scripts/build_example
```diff
@@ -41,28 +41,47 @@ else
...
+    if [[ "hello_world" == "$TARGET" ]]; then
+        # `hello_world` example has an optional additional `translator` module.
+        cargo build --release --target=wasm32-unknown-unknown --manifest-path="examples/translator/module/rust/Cargo.toml"
 
-# Use a different output_base so that we don't lose incremental state.
-# See https://docs.bazel.build/versions/master/command-line-reference.html#flag--output_base.
-bazel --output_base="$CACHE_DIR/client" build "${bazel_build_flags[@]}" "//examples/$TARGET/client"
+        bazel --output_base="$CACHE_DIR/client" build "${bazel_build_flags[@]}" \
+            "//examples/$TARGET/config:config" \
+            "//examples/$TARGET/config:config_translator"
```
tiziano88: Maybe it's clearer to replace `$TARGET` with `hello_world` here, since it's very specific to that already.

scripts/run_examples_dev
```diff
@@ -7,38 +7,56 @@ source "$SCRIPTS_DIR/common"
...
 # TODO(#594): Re-enable rustfmt when upstream rustc internal error is fixed.
-readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq | grep -v chat | grep -v rustfmt)"
+readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq | grep -v rustfmt)"
 for example in ${RUST_EXAMPLES}; do
-  "${SCRIPTS_DIR}/run_example" "${example}"
+  # Running `build_example` to compile application configurations before running the server.
+  "${SCRIPTS_DIR}/build_example" "${example}"
+  ./bazel-clang-bin/oak/server/dev/dev_oak_runner \
+    --config="$PWD/bazel-client-bin/examples/${example}/config/config.bin" &
+  SERVER_PID=$!
```
tiziano88: If this is not a constant it should probably be lowercase (I think). Also maybe it should be `local` too.

scripts/run_examples_dev
```diff
@@ -7,38 +7,56 @@ source "$SCRIPTS_DIR/common"
...
-# interact with it.
 # TODO(#594): Re-enable rustfmt when upstream rustc internal error is fixed.
-readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq | grep -v chat | grep -v rustfmt)"
+readonly RUST_EXAMPLES="$(find examples -mindepth 2 -maxdepth 4 -type d -regex '.*/module.*/rust$' | cut -d'/' -f2 | uniq | grep -v rustfmt)"
 for example in ${RUST_EXAMPLES}; do
-  "${SCRIPTS_DIR}/run_example" "${example}"
+  # Running `build_example` to compile application configurations before running the server.
+  "${SCRIPTS_DIR}/build_example" "${example}"
+  ./bazel-clang-bin/oak/server/dev/dev_oak_runner \
+    --config="$PWD/bazel-client-bin/examples/${example}/config/config.bin" &
```
tiziano88: (nit) use `${VAR}` everywhere (see https://google.github.io/styleguide/shell.xml?showone=Variable_expansion#Variable_expansion )

examples/chat/client/chat.cc
```diff
@@ -224,55 +173,43 @@ class Room {
...
-    room_id = room->Id();
-    LOG(INFO) << "Join this room with --app_address=" << addr
-              << " --room_id=" << absl::Base64Escape(room_id);
-  }
+  if (!absl::GetFlag(FLAGS_no_main_loop)) {
+    RoomId room_id;
+    if (!absl::Base64Unescape(absl::GetFlag(FLAGS_room_id), &room_id)) {
+      LOG(QFATAL) << "Failed to parse --room_id as base 64";
+    }
+    std::unique_ptr<Room> room;
```
anghelcovici: Can we add a comment here to explain why room is declared outside of the if block (even if it's used only inside the if block)? cc: @daviddrysdale

examples/chat/client/chat.cc
```diff
@@ -224,55 +173,43 @@ class Room {
...
-    LOG(QFATAL) << "Failed to parse --room_id as base 64";
-  }
-  std::unique_ptr<Room> room;
-  if (room_id.empty()) {
-    room = absl::make_unique<Room>(stub.get());
-    room_id = room->Id();
-    LOG(INFO) << "Join this room with --app_address=" << addr
-              << " --room_id=" << absl::Base64Escape(room_id);
-  }
+  if (!absl::GetFlag(FLAGS_no_main_loop)) {
```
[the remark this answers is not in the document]
anghelcovici: Also, I don't really see what this has to do with the main description of the PR. Just talked to @tiziano88  about this... again, no strong feelings, but ideally it would be nice to have these separated.

scripts/run_examples_dev
```diff
@@ -7,38 +7,56 @@ source "$SCRIPTS_DIR/common"
...
+  SERVER_PID=$!
+  sleep 2  # Wait for an application to start.
```
anghelcovici: Oh my.. there must be a better way :)

examples/chat/client/chat.cc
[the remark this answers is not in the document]
tiziano88: I also find this flag name confusing, both because of the negation, but also because of the mention of "main loop", which is not very clear. How about a `--test` flag instead? We could also add it to the other examples if necessary if we wanted to make them do something in a non-interactive way as part of the test.

scripts/run_examples_dev
ipetr0v: Done

docs/programming-oak.md
```diff
@@ -177,6 +177,8 @@ as a gRPC server:
 
 ## Using an Oak Application from a client
 
+TODO(#431) Update documentation to use a single app oak server.
```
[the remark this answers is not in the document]
ipetr0v: Finishing docs in https://github.com/project-oak/oak/pull/590.
Will merge both PRs simultaneously.

examples/chat/client/chat.cc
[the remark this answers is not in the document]
ipetr0v: Have moved this logic to a separate PR: https://github.com/project-oak/oak/pull/600

scripts/run_examples_dev
ipetr0v: I think it will require rewriting this script in Python, so we could parse the output of the `oak_runner` in order to understand that the module was loaded.

tiziano88:
Note that CI time has increased by 50% after this PR was merged: https://pantheon.corp.google.com/cloud-build/builds?query=trigger_id%3D%2216e67b67-cc42-4799-abf0-ad22b6c8559b%22&project=oak-ci . @ipetr0v do you think it will be fixed once the bazel cache is populated again?

ipetr0v:
In this PR I also re-enabled `tensorflow` and `chat` examples.
Previously they have not been built and run.

PR #485, opened 2020-01-17
Support additional panic hook payload types

conradgrobler:
Code is inspired by the default panic hook.

https://github.com/rust-lang/rust/blob/7d761fe0462ba0f671a237d0bb35e3579b8ba0e8/src/libstd/panicking.rs#L171-L177

### DIFF
--- a/sdk/rust/oak/src/lib.rs
+++ b/sdk/rust/oak/src/lib.rs
@@ -308,10 +308,14 @@ pub fn random_get(buf: &mut [u8]) -> OakStatus {
 /// [panic information]: std::panic::PanicInfo
 pub fn set_panic_hook() {
     std::panic::set_hook(Box::new(|panic_info| {
-        let msg = panic_info
-            .payload()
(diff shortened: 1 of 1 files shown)


sdk/rust/oak/src/lib.rs
```diff
@@ -308,10 +308,14 @@ pub fn random_get(buf: &mut [u8]) -> OakStatus {
 /// [panic information]: std::panic::PanicInfo
 pub fn set_panic_hook() {
     std::panic::set_hook(Box::new(|panic_info| {
-        let msg = panic_info
-            .payload()
-            .downcast_ref::<&str>()
-            .unwrap_or(&"<UNKNOWN MESSAGE>");
+        let payload = panic_info.payload();
```
tiziano88: Could you add a link in a comment to the upstream one? Would be useful as a reference if things change in the future.

conradgrobler: Done

PR #516, opened 2020-01-22
Improve `runner` script

tiziano88:
- add command-line flag parsing
- add `dry_run` command-line flag, which does not run commands
- add `commands` command-line flag, which prints commands
- add `logs` command-line flag, which prints logs
- introduce sequences of commands


anghelcovici:
Just a drive by comment: doesn't look at all more readable than bash.. just saying :)

tiziano88:
> Just a drive by comment: doesn't look at all more readable than bash.. just saying :)

It's supposed to be more maintainable and more scalable, not necessarily more readable (though to me personally it also looks more readable, but that's mostly because I'm bad at Bash). Also most of it is internal stuff that should disappear into a separate crate, all we would need to maintain is the list of commands that we actually want to run. But we should decide together whether this is something we want to commit to, so far I'm still playing around with it (and using it as an excuse to learn about rust async / futures, which I want to also use in the Oak SDK at some point).

anghelcovici:
Just had a chat with @blaxill -  The previous comment I had wasn't suppose to mean that Bash is more readable than Rust :) 
I think Rust would provide a better way to run these, I was just trying to point out that, if possible, we can maybe experiment with making it more readable. So moving things to a crate might be what is missing.

tiziano88:
> * The individual runners such as `run_buildifier` are somewhat obscured by the sequencer logic. Would it be possible to simplify this? Perhaps by removing sequencing/steps from the runners, but keeping it in calling code [...]:

This seems equivalent to using `Fn()` (i.e. an anonymous lambda) instead of `Runnable`. If it weren't for the `context` parameter, the two types would be isomorphic, and it would be fine to just collect some lambdas and run them when traversing the hierarchy, but I think having the explicit trait would be useful for more things in the future, e.g. reporting errors, or allowing killing stuck processes, etc. Also, it would be problematic to pass down a context object in that scenario, unless we changed the type of the functions to `Fn(Context)` or similar, but this would impose a dependency on the `Context` type, which I now removed.

> * There is a lot of unwrapping `PathBuf`'s to `&str`, it might be possible to keep more variables as `Path`s. In particular step could receive `PathBuf` for `entry` instead of a string as `cmd.arg` accepts any `AsRef<OsStr>` and there exists `impl AsRef<OsStr> for PathBuf`. To avoid placing it directly in `Step::args` you could perhaps return the `Command` directly:

I changed things a bit to make it easier, by adding an explicit unwrapping step so that things are already converted to `String` by the time they get to the step builder. I think this could be abstracted even more, but I have not experimented with it yet.

PR #551, opened 2020-02-04
Refactor `oak_utils`

### LINKED ISSUE #548: Refactor `oak_utils`
Update `oak_utils` logic based on the following comments:
https://github.com/project-oak/oak/pull/533#discussion_r374329609
https://github.com/project-oak/oak/pull/533#discussion_r374330722

Issue was created to solve #510 in multiple PRs.

ipetr0v:
This change refactors `oak_utils` crate.

Fixes #548

### Checklist

      construction.


sdk/rust/oak_utils/src/lib.rs
```diff
@@ -41,31 +42,114 @@ pub fn run_protoc_rust(args: protoc_rust::Args) -> io::Result<()> {
...
+                    FileAction::Update {
+                        new_content: new_content.to_string(),
+                    }
+                } else {
+                    FileAction::Ignore
+                }
+            }
+            (Some(_), None) => {
+                // `mod.rs` file is not generated by `protoc_rust` and `protoc_rust_grpc`.
+                if "mod.rs" != self.filename {
```
tiziano88: These conditions are hard to follow because they are all negated, I think they would read more nicely if they were switched to be positive (e.g. `if old_content == new_content` and `if self.filename == "mod.rs"`)

sdk/rust/oak_utils/src/lib.rs
```diff
@@ -41,31 +42,114 @@ pub fn run_protoc_rust(args: protoc_rust::Args) -> io::Result<()> {
...
+                new_content: new_content.to_string(),
+            },
+            _ => FileAction::Ignore,
+        }
+    }
+}
+
+// Copies changed files from `temp_path` to `path` and removes files from `path`
+// that are not present in `temp_path`.
+fn update_files(path: &Path, temp_path: &Path) -> io::Result<()> {
```
tiziano88: Consider swapping the arguments, it seems more natural to have from -> to, also it would mirror the comment on this function.

sdk/rust/oak_utils/src/lib.rs
```diff
@@ -41,31 +42,114 @@ pub fn run_protoc_rust(args: protoc_rust::Args) -> io::Result<()> {
...
+    let temp_files = get_files(temp_path);
+    get_file_diffs(&files, &temp_files)
+        .iter()
+        .try_for_each(|diff| {
+            let filepath = path.join(&diff.filename);
+            match diff.action() {
+                FileAction::Ignore => Ok(()),
+                FileAction::Remove => fs::remove_file(filepath),
+                FileAction::Update { new_content } => {
+                    let mut file = fs::File::create(filepath).unwrap();
```
tiziano88: Add a comment that if the file is already there, it will truncate it (does it?)?

sdk/rust/oak_utils/src/tests.rs
```diff
@@ -22,72 +22,143 @@ use std::fs::File;
...
+    assert_eq!(
+        FileDiff {
+            filename: "test".to_string(),
+            old_content: Some("1".to_string()),
+            new_content: None,
+        }
+        .action(),
+        FileAction::Remove
+    );
+    assert_eq!(
```
tiziano88: Could you add a comment here that "mod.rs" is treated as an exception? Otherwise it looks odd to someone reading this test.

sdk/rust/oak_utils/src/tests.rs
```diff
@@ -22,72 +22,143 @@ use std::fs::File;
...
+        FileDiff {
+            filename: "test".to_string(),
+            old_content: None,
+            new_content: Some("2".to_string()),
+        }
+        .action(),
+        FileAction::Update {
+            new_content: "2".to_string()
+        }
+    );
```
tiziano88: Could you add a test case for a file with empty string content (either old or new)?

sdk/rust/oak_utils/src/lib.rs
ipetr0v: Yes, it truncates it: https://github.com/rust-lang/rust/blob/f564c4db0d97eabd7fdd72e589d3e415790ee2a4/src/libstd/fs.rs#L378-L400
Done

sdk/rust/oak_utils/src/tests.rs
ipetr0v: Done

sdk/rust/oak_utils/src/tests.rs
ipetr0v: Done

sdk/rust/oak_utils/src/tests.rs
```diff
@@ -22,72 +22,165 @@ use std::fs::File;
...
+            old_content: None,
+            new_content: Some("string4".to_string()),
+        },
+    ];
+    expected_file_diffs.sort();
 
-    let old_temp_dir = tempfile::tempdir().unwrap();
-    let new_temp_dir = tempfile::tempdir().unwrap();
+    let mut file_diffs = get_file_diffs(&old_files, &new_files);
+    file_diffs.sort();
```
tiziano88: Alternatively could use a hash set, and avoid sorting (though probably need to implement / derive hashing instead of ord). Up to you.

ipetr0v: Done

PR #559, opened 2020-02-06
Add a `serialized_config` Bazel rule

ipetr0v:
This change adds a Bazel rule for creating a serialized config binary.

Necessary for #462

### Checklist

      construction.

### DIFF
--- a/oak/common/app_config.bzl
+++ b/oak/common/app_config.bzl
@@ -1,0 +1,40 @@
+#
+# Copyright 2020 The Project Oak Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Rule for serializing an Oak application configuration."""
+
+def serialized_config(name, textproto, modules, output):
+    """Serializes an Oak application configuration in a binary file.
+
+    Args:
+        name: Name of the build rule.
+        textproto: Textproto file with application configuration.
+        modules: A dictionary with module names as keys and module paths as values.
+        output: Output file.
+    """
+    srcs = [textproto] + modules.values()
+    apps = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
+    cmd = "$(location //oak/common:app_config_serializer)" + \
+          " --textproto=$(location " + textproto + ")" + \
+          " --modules=" + apps + \
+          " --output_file=$@"
+    native.genrule(
+        name = name,
+        srcs = srcs,
+        outs = [output],
+        cmd = cmd,
+        tools = ["//oak/common:app_config_serializer"],
+    )
--- a/oak/common/app_config.bzl
+++ b/oak/common/app_config.bzl
@@ -26,10 +26,10 @@ def serialized_config(name, textproto, modules, output):
         output: Output file.
     """
     srcs = [textproto] + modules.values()
-    apps = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
+    module_list = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
     cmd = "$(location //oak/common:app_config_serializer)" + \
           " --textproto=$(location " + textproto + ")" + \
-          " --modules=" + apps + \
+          " --modules=" + module_list + \
           " --output_file=$@"
     native.genrule(
         name = name,
--- a/oak/common/app_config.bzl
+++ b/oak/common/app_config.bzl
@@ -20,13 +20,12 @@ def serialized_config(name, textproto, modules):
     """Serializes an Oak application configuration in a binary file.
 
     Args:
-        name: Name of the build rule.
+        name: Name of the generated binary file.
         textproto: Textproto file with application configuration.
         modules: A dictionary with module names as keys and module paths as values.
     """
     srcs = [textproto] + modules.values()
-    output = name + ".bin"
-    module_list = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
+    module_list = ",".join(["{}:$(location {})".format(name, path) for (name, path) in modules.items()])
     cmd = "$(location //oak/common:app_config_serializer)" + \
           " --textproto=$(location {})".format(textproto) + \
           " --modules={}".format(module_list) + \
@@ -34,7 +33,7 @@ def serialized_config(name, textproto, modules):
     native.genrule(
         name = name,
         srcs = srcs,
-        outs = [output],
+        outs = [name],
         cmd = cmd,
         tools = ["//oak/common:app_config_serializer"],
     )
--- a/oak/common/app_config.bzl
+++ b/oak/common/app_config.bzl
@@ -16,16 +16,16 @@
 
 """Rule for serializing an Oak application configuration."""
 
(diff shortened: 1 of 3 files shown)


oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+"""Rule for serializing an Oak application configuration."""
+
+def serialized_config(name, textproto, modules, output):
+    """Serializes an Oak application configuration in a binary file.
+
+    Args:
+        name: Name of the build rule.
+        textproto: Textproto file with application configuration.
+        modules: A dictionary with module names as keys and module paths as values.
+        output: Output file.
```
tiziano88: Maybe just use the name of the rule as the name of an implicit output?

oak/common/app_config_serializer.cc
```diff
@@ -88,6 +89,12 @@ int main(int argc, char* argv[]) {
     }
   }
 
+  // Check application configuration validity.
+  if (!oak::ValidApplicationConfig(*config.get())) {
+    LOG(INFO) << "Application config is not valid";
```
tiziano88: Why is this INFO but the other one is QFATAL?

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+    """
+    srcs = [textproto] + modules.values()
+    apps = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
```
tiziano88: The name `apps` seems misleading. Perhaps `modules_expanded` or something like that?

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+    cmd = "$(location //oak/common:app_config_serializer)" + \
+          " --textproto=$(location " + textproto + ")" + \
```
tiziano88: Is it not possible to use variable interpolation inside a string, e.g. `$(textproto)` or something like that?

oak/common/app_config_serializer.cc
ipetr0v: Done

oak/common/app_config.bzl
ipetr0v: `textproto` that is passed to a function looks like `//examples/hello_world/module/cpp:hello_world.wasm`, and after `$(location ...)` it will become a full fs path.

oak/common/app_config.bzl
ipetr0v: Should it become `serialized_config_file` ?

oak/common/app_config.bzl
ipetr0v: Done

oak/common/app_config.bzl
tiziano88: I meant, use the `name` variable itself.

oak/common/app_config.bzl
tiziano88: I understand why we need `location`, but not why we need to manually connect strings together. Would `$(location $(textproto))` not work?

ipetr0v: Done.
Using `format` instead.

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+
+    Args:
+        name: Name of the build rule.
+        textproto: Textproto file with application configuration.
+        modules: A dictionary with module names as keys and module paths as values.
+    """
+    srcs = [textproto] + modules.values()
+    output = name + ".bin"
+    module_list = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
+    cmd = "$(location //oak/common:app_config_serializer)" + \
```
tiziano88: Is the trailing `\` needed? I don't see it in other similar files: https://github.com/google/asylo/blob/ebdff40ca061a6f733812cc50a88e403759c6a14/asylo/bazel/asylo.bzl#L265-L269

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+          " --textproto=$(location {})".format(textproto) + \
```
tiziano88: Would this work?

```suggestion
          " --textproto=$(location {textproto})"
```

From https://www.programiz.com/python-programming/string-interpolation

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Rule for serializing an Oak application configuration."""
+
+def serialized_config(name, textproto, modules):
+    """Serializes an Oak application configuration in a binary file.
+
```
tiziano88: Can you mention what the implicit output name is? Otherwise I think it's impossible for clients to figure it out, apart from looking at this implementation.

See https://docs.bazel.build/versions/2.0.0/be/c-cpp.html#cc_binary_implicit_outputs

oak/common/app_config.bzl
```diff
@@ -0,0 +1,40 @@
...
+    Args:
+        name: Name of the build rule.
+        textproto: Textproto file with application configuration.
+        modules: A dictionary with module names as keys and module paths as values.
+    """
+    srcs = [textproto] + modules.values()
+    output = name + ".bin"
+    module_list = ",".join([n + ":$(location " + path + ")" for (n, path) in modules.items()])
```
tiziano88: Use format or interpolation here too.

oak/common/app_config.bzl
ipetr0v: IIUC it's only possible in `f-strings` in python3.6, and it's probably not implemented in Starlark.
Doesn't seem to work - `f' --modules={module_list}'` gives:
```
ERROR: /opt/my-project/oak/common/app_config.bzl:31:11: name 'f' is not defined
```
and `" --modules={module_list}"` gives:
```
label '//examples/hello_world/config:{textproto}' in $(location) expression is not a declared prerequisite of this rule.
```

oak/common/app_config.bzl
ipetr0v: Done

oak/common/app_config.bzl
ipetr0v: Strangely enough, but it only works, if I use it as a parameter to a rule.
For example, this will compile:
```python
native.genrule(
        name = name,
        srcs = srcs,
        outs = [output],
        cmd = "$(location //oak/common:app_config_serializer)" + 
          " --textproto=$(location {})".format(textproto) + 
          " --modules={}".format(module_list) + 
          " --output_file=$@",
        tools = ["//oak/common:app_config_serializer"],
    )
```
And this will not:
```python
cmd = "$(location //oak/common:app_config_serializer)" + 
          " --textproto=$(location {})".format(textproto) + 
          " --modules={}".format(module_list) + 
          " --output_file=$@"
    native.genrule(
        name = name,
        srcs = srcs,
        outs = [output],
        cmd = cmd,
        tools = ["//oak/common:app_config_serializer"],
    )
```
```
ERROR: /opt/my-project/oak/common/app_config.bzl:30:62: syntax error at 'newline': expected expression
ERROR: /opt/my-project/oak/common/app_config.bzl:40:5: Trailing comma is allowed only in parenthesized tuples.
ERROR: /opt/my-project/oak/common/app_config.bzl:30:62: contains syntax error(s)
ERROR: /opt/my-project/oak/common/app_config.bzl:31:61: contains syntax error(s)
ERROR: /opt/my-project/oak/common/app_config.bzl:37:9: name 'outs' is not defined
ERROR: /opt/my-project/oak/common/app_config.bzl:32:47: cannot assign to '+$error$'
ERROR: /opt/my-project/oak/common/app_config.bzl:39:9: name 'tools' is not defined (did you mean 'bool'?)
ERROR: /opt/my-project/oak/common/app_config.bzl:40:6: contains syntax error(s)
```

oak/common/app_config.bzl
ipetr0v: Done, using the `name` argument as the name of the output file (without `.bin`)

Issue #564, opened 2020-02-06
oak_runtime `load_wasm` should not panic


blaxill:
~~~rust
pub fn load_wasm(wasm_bytes: &[u8]) -> NodeConfiguration {
    let module = wasmi::Module::from_buffer(wasm_bytes).expect("failed to load wasm");
~~~

This should not panic and instead should return a result or option.

https://github.com/project-oak/oak/pull/472#discussion_r375777837

~~Blocked on #472~~

PR #579, opened 2020-02-07
oak_runtime: Return a `Result` from load_wasm

blaxill:
#564

### Checklist

  - [X] I have updated [documentation](docs/) accordingly.
      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/mod.rs
+++ b/oak/server/rust/oak_runtime/src/node/mod.rs
@@ -41,6 +41,7 @@ pub enum NodeConfiguration {
     WasmNode { module: Arc<wasmi::Module> },
 }
 
+/// A enumeration for errors occuring when building `NodeConfiguration` from protobuf types.
 pub enum ConfigurationError {
     WasmiModuleInializationError(wasmi::Error),
 }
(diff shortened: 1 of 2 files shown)


oak/server/rust/oak_runtime/src/runtime.rs
```diff
@@ -67,7 +67,10 @@ impl Runtime {
                     }
                     Some(NodeConfiguration_oneof_config_type::wasm_config(
                         proto::manager::WebAssemblyConfiguration { module_bytes, .. },
-                    )) => load_wasm(&module_bytes),
+                    )) => load_wasm(&module_bytes).map_err(|e| {
+                        error!("Error loading wasm: {}", e);
```
tiziano88:
```suggestion
                        error!("Error loading Wasm module: {}", e);
```

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -41,15 +41,31 @@ pub enum NodeConfiguration {
...
+            ConfigurationError::WasmiModuleInializationError(e) => {
+                write!(f, "Failed to initialize wasmi::Module: {}", e)
+            }
+        }
     }
 }
 
+/// Loads a Wasm module into a node configuration, returning an error if `wasmi` failed to load the
+/// module.
+pub fn load_wasm(wasm_bytes: &[u8]) -> Result<NodeConfiguration, ConfigurationError> {
```
tiziano88: Why not return `Result<NodeConfiguration, wasmi::Error>` instead, which is more precise?

blaxill: I think we will want to catch more configuration errors in the future, when there are more nodes. Perhaps `Runtime::configure_and_run` returning `Result<(RuntimeRef, ChannelWriter), ConfigurationError>`, although maybe adding the enum now is a bit premature.

oak/server/rust/oak_runtime/src/runtime.rs
blaxill: Done

Issue #569, opened 2020-02-07
Change single app server terminology


ipetr0v:
We need to change binary and library names for the single app servers (both `dev` and `asylo`), because currently `loader` refers both to a binary and a library.
https://github.com/project-oak/oak/blob/0526d782a4b2b9854579a3b28368710ad74d64cc/oak/server/asylo/BUILD#L52-L56
https://github.com/project-oak/oak/blob/0526d782a4b2b9854579a3b28368710ad74d64cc/oak/server/asylo/BUILD#L172-L174

Proposed names:
- change library name to `oak_loader_dev` and `oak_loader_asylo`
- change binary name to `oak_runner_dev` and `oak_runner_asylo`

Issue is based on the following comment https://github.com/project-oak/oak/pull/553/files#r376103473

PR #583, opened 2020-02-09
Rename loaders (binary, library)

ipetr0v:
This change resolves single app server name ambiguity by introducing `oak_loader` library and `oak_runner` binary.

Fixes https://github.com/project-oak/oak/issues/569

### Checklist

      construction.


oak/server/asylo/BUILD
```diff
@@ -170,8 +170,8 @@ enclave_loader(
 )
 
 enclave_loader(
-    name = "oak_app_loader",
-    srcs = ["asylo_oak_app_loader_main.cc"],
+    name = "oak_runner",
```
tiziano88: What made you decide against `oak_runner_asylo` and `oak_runner_dev` in the end, which is what #569 references? Not necessarily against it either way, but I'm just curious why the change.

In any case, I think the binary name should match the source file name, or at least follow a similar structure.

ipetr0v: I just thought that the `asylo` is already implied by the directory.
Changed to `asylo_oak_runner`.
Setting `asylo_` at the beginning since all our files follow this convention.

PR #584, opened 2020-02-09
Fix name duplication in `app_config` Bazel rule

ipetr0v:
Use different `name` for `genrule` in `app_config.bzl`.

### Checklist

      construction.

### DIFF
--- a/oak/common/app_config.bzl
+++ b/oak/common/app_config.bzl
@@ -31,7 +31,8 @@ def serialized_config(name, textproto, modules):
           " --modules={}".format(module_list) + \
           " --output_file=$@"
     native.genrule(
-        name = name,
+        # Name of the rule cannot be the same as the output file.
+        name = "serialized_{}".format(name),
         srcs = srcs,
         outs = [name],
         cmd = cmd,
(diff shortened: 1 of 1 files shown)


tiziano88:
Also please document the actual output name in the comments.

ipetr0v:
It showed the following error:
```
target 'config' is both a rule and a file; please choose another name for the rule
```
But it's not described in the documentation.
I think this only applies to `genrule`.

oak/common/app_config.bzl
```diff
@@ -20,7 +20,7 @@ def serialized_config(name, textproto, modules):
     """Serializes an Oak application configuration in a binary file.
 
     Args:
-        name: Name of the generated binary file.
+        name: Name of the generated binary file (the output file will have a `.bin` extension, e.g. `name.bin`).
```
tiziano88: Please add an explicit comment before the args section. If someone is looking for the name of the file generated by this rule, they shouldn't have to look at the args section.

See https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary_implicit_outputs

ipetr0v: Done

PR #585, opened 2020-02-09
Remove Oak Manager

ipetr0v:
This change:
- Removes `OakManager` in favor of `OakLoader`
- Deprecates `{dev|asylo}/oak` in favor of `{dev|asylo}/oak_runner` (single app server)

Fixes https://github.com/project-oak/oak/issues/431
Documentation is updated in #590

To be reviewed after https://github.com/project-oak/oak/pull/474 is merged

### Checklist

      construction.


oak/proto/manager.proto
```diff
@@ -72,34 +72,9 @@ message StorageProxyConfiguration {
...
-  // to verify that its configuration corresponds to what the client expects.
-  rpc CreateApplication(CreateApplicationRequest) returns (CreateApplicationResponse);
-
-  // Request that an Oak Application terminate.
-  rpc TerminateApplication(TerminateApplicationRequest) returns (TerminateApplicationResponse);
+// ApplicationCreationStatus describes the configuration of a successfully
+// loaded Oak application.
+message ApplicationCreationStatus {
+  // The port on which the application is listening over gRPC.
+  int32 grpc_port = 1;
```
tiziano88: Since we now expect exactly one instance of a gRPC server to be running for each invocation, I think it's reasonable to assume that a specific port will be passed in -- also it is necessary for it to work with docker, and I don't think we have any use case in which we parse the text output of the loader to figure out what the actual port is. I think we should validate that the requested port in the app configuration is valid (i.e. strictly > 0) and not even bother with the logic for dynamic port assignment, and also remove this field (and perhaps the entire message, if you think we won't need it for anything else anytime soon). WDYT?

oak/server/dev/dev_oak_runner_main.cc
```diff
@@ -36,21 +36,20 @@ int main(int argc, char* argv[]) {
...
+  if (!status.ok()) {
     LOG(ERROR) << "Failed to create application";
     return 1;
   }
-  oak::CreateApplicationResponse response = result.ValueOrDie();
   std::stringstream address;
-  address << "0.0.0.0:" << response.grpc_port();
-  std::string application_id(response.application_id());
-  LOG(INFO) << "Oak Application id=" << application_id << ": " << address.str();
+  address << "0.0.0.0:" << application_config->grpc_port();
```
tiziano88: Should we check for the port number to be >0 here?

ipetr0v: It's already checked:
https://github.com/project-oak/oak/blob/7978a6eda87df6be569505c2bab9b4f1d0a0aa1e/oak/server/oak_runtime.cc#L65-L70

tiziano88: Is that check correct though? AFAICT it will allow ports like -1, 0, and 1025, is that the intention?

ipetr0v: Will change it to `grpc_port <= 1023`

PR #600, opened 2020-02-12
Add a `--test` flag to chat

ipetr0v:
This change:
- Adds a non-interactive version of Chat application with `--test` flag
- Enables Chat application in `run_examples`

### Checklist

      construction.

### DIFF
--- a/examples/chat/client/chat.cc
+++ b/examples/chat/client/chat.cc
@@ -252,32 +252,33 @@ int main(int argc, char** argv) {
   }
 
   if (absl::GetFlag(FLAGS_test)) {
-    return EXIT_SUCCESS;
-  } else {
-    RoomId room_id;
-    if (!absl::Base64Unescape(absl::GetFlag(FLAGS_room_id), &room_id)) {
-      LOG(QFATAL) << "Failed to parse --room_id as base 64";
-    }
-    std::unique_ptr<Room> room;
-    if (room_id.empty()) {
-      room = absl::make_unique<Room>(stub.get());
-      room_id = room->Id();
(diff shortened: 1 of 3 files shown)


examples/chat/client/chat.cc
```diff
@@ -250,29 +251,31 @@ int main(int argc, char** argv) {
...
-    LOG(QFATAL) << "Failed to parse --room_id as base 64";
-  }
-  std::unique_ptr<Room> room;
-  if (room_id.empty()) {
-    room = absl::make_unique<Room>(stub.get());
-    room_id = room->Id();
-    LOG(INFO) << "Join this room with --app_address=" << addr
-              << " --room_id=" << absl::Base64Escape(room_id);
-  }
+  if (!absl::GetFlag(FLAGS_test)) {
```
tiziano88: Could you reverse the condition, so that it reads something like

```cpp
if (test) {
  return EXIT_SUCCESS;
}
```

ipetr0v: Done

examples/chat/client/chat.cc
```diff
@@ -251,7 +251,9 @@ int main(int argc, char** argv) {
     LOG(QFATAL) << "Failed to create application stub";
   }
 
-  if (!absl::GetFlag(FLAGS_test)) {
+  if (absl::GetFlag(FLAGS_test)) {
+    return EXIT_SUCCESS;
+  } else {
```
tiziano88: I think you can remove the else, an early return is clear enough, and you can remove a layer of indentation, but up to you.

ipetr0v: Done

PR #615, opened 2020-02-17
Support structured output in runner script

tiziano88:
The downside is that it now requires all the steps to have finished
before printing anything at all, which makes it hard to figure out at
what point of the overall execution it is.

On the other hand this will make it easier to support future-based
return value in the future, at which point we will reimplement the
streaming-line output functionality.

# Checklist

      construction.

### DIFF
--- a/runner/src/internal.rs
+++ b/runner/src/internal.rs
@@ -36,9 +36,9 @@ pub struct Context {
 }
 
 impl Context {
-    pub fn root(opt: Opt) -> Self {
+    pub fn root(opt: &Opt) -> Self {
         Context {
-            opt,
+            opt: opt.clone(),
             prefix: "".to_string(),
         }
     }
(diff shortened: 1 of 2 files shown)


runner/src/internal.rs
```diff
@@ -51,15 +51,90 @@ impl Context {
...
+                }
+                StatusResultValue::Skipped => {
+                    eprintln!("{} ⊢ {}", context.prefix, "SKIPPED".bold().bright_yellow())
+                }
+            };
+            if context.opt.commands || context.opt.dry_run {
+                eprintln!("{} ⊢ [{}]", context.prefix, command);
+            };
+            if value == &StatusResultValue::Error || context.opt.logs {
+                eprintln!("{} {}", context.prefix, "╔════════════════════════".blue());
```
ipetr0v: Does this table need `╗` ?

runner/src/internal.rs
```diff
@@ -51,15 +51,90 @@ impl Context {
...
 pub type R = Box<dyn Runnable>;
 
-pub fn run(runnable: R, opt: Opt) {
-    runnable.run(&Context::root(opt))
+/// Run the provided runnable, and print out a summary of the final status (including subtasks).
+pub fn run_all(runnable: R, opt: &Opt) {
+    let root = Context::root(&opt);
+    let status = runnable.run(&opt);
+    print_status(&root, &status);
+}
```
ipetr0v: Does `runner` exit with `1`, so GCB would stop with an error, if something is wrong ?

runner/src/internal.rs
tiziano88: It's only a "half" table, to separate logs from different command, since I don't know how long the log lines actually are.

runner/src/internal.rs
tiziano88: Good point, it does not do that now, but I'll add it in a future PR (this is not run on GCB yet).

PR #617, opened 2020-02-17
Parallelize build steps

tiziano88:
Reduces CI build time from ~35 minutes to ~21 minutes (see
https://pantheon.corp.google.com/cloud-build/builds?project=oak-ci).

Some of the build steps may be run in parallel; the main rule of thumb
is that they can if they don't share the same Bazel `output_base`,
because otherwise they would contend for the same lock.

Ensure that Bazel caching for wasm32 and emscripten works properly.

Ensure that Bazel is properly initialized in the Dockerfile.

Reorder steps in Dockerfile so that less frequently changed steps appear
first (namely, Emscripten, which is also quite large).

Fix various other minor things in the Dockerfile (env variables).

# Checklist

      construction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,52 @@ RUN shellcheck --version
 RUN node --version
 RUN npm --version
 
+# Make sure Bazel is correctly initialized.
+RUN bazel version
+
+# Install the necessary binaries and SDKs, ordering them from the less frequently changed to the
+# more frequently changed.
+# See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache.
+
+# Install Emscripten.
+ARG EMSCRIPTEN_VERSION=1.39.6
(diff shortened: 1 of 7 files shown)


tiziano88:
There seems to be a race condition when starting multiple instances of `bazel`, despite the distinct `output_base` values:

```log
Extracting Bazel installation...
FATAL: Failed to extract embedded binaries: Failed to write zipped file '/builder/home/.cache/bazel/_bazel_root/install/84defa6eb1e9416bf92d6f89ab2d4f31.tmp.9/embedded_tools/jdk/lib/security/public_suffix_list.dat': (error: 2): No such file or directory
```

From https://pantheon.corp.google.com/cloud-build/builds/27a8485c-e860-4940-a98d-68d7828adedc;step=5?project=oak-ci

ipetr0v:
Does this PR intersect with https://github.com/project-oak/oak/pull/492 ?

PR #628, opened 2020-02-21
Add `Dispatcher`.

### LINKED ISSUE #577: Unify generated gRPC traits with `oak::grpc::OakNode`

wildarch:
This is a convenient wrapper around a gRPC service implementation to
implement `OakNode` for it.

Fixes #577. Unfortunately there is no way to automatically implement `OakNode` for anything that implements the autogenerated gRPC service traits, so this is probably the most ergonomic solution.

Consider: We might even want to remove `OakNode` entirely, I'm not sure it really pulls its weight. We could implement `Node` for `Dispatcher` directly.

# Checklist

      construction.


tiziano88:
> Consider: We might even want to remove OakNode entirely, I'm not sure it really pulls its weight. We could implement Node for Dispatcher directly.

I think we should consider generating a custom `Dispatcher` struct in the gRPC codegen phase, alongside the server trait, instead of (or in addition to) the `dispatch` method.

When used it would look something like:

```rust
oak::entrypoint!(oak_main => {
    oak_log::init_default();
    protobuf::chat_grpc::Dispatcher::new(Node::default()),
});
```

and remove the `oak::grpc::Dispatcher` struct you have introduced here. Basically monomorphize (or "reify") `Dispatcher` in the context of the generated trait.

@wildarch WDYT?

wildarch:
That would absolutely work! But I'm not sure there is a big advantage in generating a dispatcher for every struct, the approach in this PR means that there is less code we need to generate, which I would say is generally a good thing. It's also easy to find documentation for the type in the main SDK crate. 
I guess the one downside is that there is some indirection because we are calling a reference to a function, but given that we are already sending network packets etc. in this context I don't think the performance hit should be a problem.

@tiziano88 could you elaborate a little bit on why you'd like to have the dispatcher generated instead?

tiziano88:
The generated code would not be much more I think? Just the 4-5 lines of the struct and the `dispatch` method invocation.

The problem with the current approach is that there is coupling between the node implementation and the dispatch method, i.e. there is only one possible valid combination that can possibly work, so my suggestion is to abstract that away and only have one thing that needs to be imported from the generated code and takes care of everything.

Also the `dispatch` method seems like the wrong level of abstraction, it would never be used outside of this particular use case.

wildarch:
Aah then I see what you mean, I do agree the dispatch function is a weird thing to expose.

I think we can get the best of both worlds: Have one generic `Dispatcher` in a familiar place with good docs and an example (which would be hard to auto-generate), while hiding the `dispatch` function. We can do this by adding a new method to the generated traits:
```rust
pub trait ExampleService {
    // ...
    fn into_dispatcher(self) -> grpc::Dispatcher<Self> {
        grpc::Dispatcher::new(self, dispatch)
    }
}
```

We can then make the `dispatch` function private.

What do you think of this idea?

tiziano88:
I did think about adding methods with default impl to the generated trait, but I discarded the idea as it seems not very elegant, and also introduces issues with naming of gRPC methods: what if the gRPC service defines an RPC method called `into_dispatcher`? then you would need to escape / rename it somehow.

wildarch:
You comment about `into_dispatcher` possibly causing name clashes convinced me that this is indeed not a great solution, I've followed your advice and have gone ahead with auto-generating dispatcher structs for every service trait.

PR #643, opened 2020-02-24
rust/oak_abi: Make oak_abi

blaxill:
This is a temporary fix to allow oak_abi `no_std` compilation. See #638, a follow up PR will be done to resolve this long term. I would like to add this as a temporary fix as it blocks work on making the oak_runtime `no_std` compatible. 

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_abi/src/lib.rs
+++ b/oak/server/rust/oak_abi/src/lib.rs
@@ -21,8 +21,7 @@
 pub mod proto;
 #[cfg(feature = "std")]
 mod inner {
-    pub use super::proto::oak_api::ChannelReadStatus;
-    pub use super::proto::oak_api::OakStatus;
+    pub use super::proto::oak_api::{ChannelReadStatus, OakStatus};
 }
 
 #[cfg(feature = "no_std")]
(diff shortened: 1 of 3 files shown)


oak/server/rust/oak_abi/src/lib.rs
```diff
@@ -16,11 +16,51 @@
...
 pub use proto::oak_api::ChannelReadStatus;
+#[cfg(feature = "std")]
 pub use proto::oak_api::OakStatus;
 
+#[cfg(feature = "std")]
 pub mod proto;
 
+// TODO(#638): Generate from protobuf in a no_std compatible way
+#[cfg(feature = "no_std")]
+pub mod proto {
```
tiziano88: Would something like this work?

```rust
#[cfg(feature = "no_std")]
mod inner {
  pub enum ChannelReadStatus { /* ... */ }
  pub enum OakStatus { /* ... */ }
}

#[cfg(feature = "std")]
mod inner {
  mod proto;
  pub use proto::oak_api::{ChannelReadStatus, OakStatus};
}

pub use inner::*;
```

blaxill: I can't put the `mod proto;` inside `inner`, but I can do the rest, WDYT?

tiziano88: LG! Are nested `mod`s not supported then?

blaxill: They are supported, but as that one in particular is a file import, rustc looks for the definition in `src/inner/proto` and using super/crate doesn't seem to work for this.

Issue #630, opened 2020-02-24
Implement labels and information flow control in Rust Oak Runtime


tiziano88:
Details TBD, but the implementation must support the functionality sketched out in #515 , at a minimum.

This issue is specifically about implementing this for the Rust Oak Runtime (which is not used yet) as opposed to the C++ one.

PR #648, opened 2020-02-25
[wip] rust runtime channel ownership

blaxill:
Very rough draft, rust_runtime compiles. Code isn't supposed to be good yet, just an idea of whats structs/how ownership might work

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/wasm.rs
+++ b/oak/server/rust/oak_runtime/src/node/wasm.rs
@@ -28,9 +28,9 @@ use wasmi::ValueType;
 
 use oak_abi::{ChannelReadStatus, OakStatus};
 
-use crate::channel::{ChannelEither, ChannelReader, ChannelWriter, ReadStatus};
+use crate::channel::ReadStatus;
 use crate::platform;
-use crate::{Message, RuntimeRef};
+use crate::{Message, RuntimeRef, ChannelWriter, ChannelReader, ChannelEither, ChannelEitherOwned};
 
 /// These number mappings are not exposed to the Wasm client, and are only used by `wasmi` to map
 /// import names to host functions. See https://docs.rs/wasmi/0.6.2/wasmi/trait.Externals.html
@@ -85,7 +85,7 @@ struct WasmInterface {
 impl WasmInterface {
     /// Generate a randomized handle. Handles are random to prevent accidental dependency on
     /// particular values or sequence. See https://github.com/project-oak/oak/pull/347
-    fn allocate_new_handle(&mut self, channel: ChannelEither) -> AbiHandle {
+    fn allocate_new_handle(&mut self, channel: ChannelEitherOwned) -> AbiHandle {
         loop {
             let handle = rand::thread_rng().next_u64();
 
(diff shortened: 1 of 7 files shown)


oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -85,7 +85,7 @@ struct WasmInterface {
 impl WasmInterface {
     /// Generate a randomized handle. Handles are random to prevent accidental dependency on
     /// particular values or sequence. See https://github.com/project-oak/oak/pull/347
-    fn allocate_new_handle(&mut self, channel: ChannelEither) -> AbiHandle {
+    fn allocate_new_handle(&mut self, channel: ChannelEitherOwned) -> AbiHandle {
```
tiziano88: Would it make sense to introduce this "randomized" handle concept to the ABI interface? I don't think it's specific to Wasm.

oak/server/rust/oak_runtime/src/runtime.rs
```diff
@@ -16,34 +16,68 @@
...
-use crate::channel::{ChannelReader, ChannelWriter};
+use crate::message::Message;
+use crate::channel::Channel;
 use crate::node::{load_wasm, NodeConfiguration};
 use crate::platform;
 use crate::proto;
 use crate::proto::application::NodeConfiguration_oneof_config_type;
 
-/// Runtime structure for configuring and running a set of Oak nodes.
+#[derive(PartialEq, Eq, Hash)]
```
tiziano88: How about the following (trying to simplify things a bit):

```rust
struct NodeId(usize);
struct Node {
  handles: HashSet<Handle>;
  /* ... */
};

struct ChannelId(usize);
struct Channel {
  /* ... */
}

/// All handles have the same representation.
/// Ideally this would also be already randomized (as per my other comment).
struct Handle(usize);

// Possibly come up with a better name...
enum Object {
  ChannelWriter { channel_id: ChannelId },
  ChannelReader { channel_id: ChannelId },
}

struct Runtime {
  nodes: HashMap<NodeId, Node>,
  channels: HashMap<ChannelId, Channel>,

  objects: HashMap<Handle, Object>,
}
```

Alternatively, we could use [`Weak`](https://doc.rust-lang.org/std/rc/struct.Weak.html) as in

```rust
enum Object {
  ChannelWriter { channel: Weak<Channel> }
  ChannelReader{ channel: Weak<Channel> }
}
```

and avoid the intermediate `ChannelId` (basically use the memory address as the id), but I feel that having explicit, printable ids for everything (including nodes and channels) would be useful anyways.

Thoughts?

Also would like to hear @daviddrysdale 's opinion.

tiziano88: FTR

I realised that effectively there are already cases in which just having a node talk to its direct channel(reader)s is not sufficient. For instance:

https://github.com/project-oak/oak/blob/ca6eb8608e826640957540f3984fe3fa5c3008c4/oak/server/rust/oak_runtime/src/node/logger.rs#L35

in which in order to wait on channels we need to go back to the Runtime.

That line may as well read

```rust
runtime.wait_on_channels([handle])
```

and shows how the API I am suggesting may look like if we extended to other channel operations, e.g.

```rust
runtime.read(handle)
```

tiziano88:
BTW this is starting to block #630 . @blaxill are you intending to go ahead with it, or should I assume the current version of the runtime (with node-owned channels) going forward?

blaxill:
Yes, I'll make this more presentable and incorporate some of the ideas we've had

blaxill:
Closing in favour of #680

PR #652, opened 2020-02-26
Rename `oak::grpc::OakNode` to `oak::grpc::ServerNode` 

rbehjati:
# Checklist

      construction.

# Notes
- A few additional files were updated after I ran the `format` script, and they are included in this PR. 
- The rust documentation needs to be regenerated after this PR is merged. I did not create a new issue for it. Do we need one?


tiziano88:
FYI I think this conflicts with #628 , which was already merged earlier today, so you would need to rebase on top of that (or rather, on top on master, usually).

tiziano88:
> A few additional files were updated after I ran the format script, and they are included in this PR.

I'm not sure how that could have happened. Maybe you ran the format script from outside Docker?

> The rust documentation needs to be regenerated after this PR is merged. I did not create a new issue for it. Do we need one?

No need for the issue, just run https://github.com/project-oak/oak/blob/master/scripts/build_gh_pages after this PR is merged.

tiziano88:
Could you use rebase instead of merge so that it's clearer what commits belong to this PR, and merging is hopefully more predictable? Thanks

sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -119,14 +119,14 @@ impl ChannelResponseWriter {
 
 /// Trait for Oak Nodes that act as a gRPC services.
 ///
-/// An `OakNode` instance is normally passed to [`oak::run_event_loop`], to
+/// An `ServerNode` instance is normally passed to [`oak::run_event_loop`], to
```
tiziano88:
```suggestion
/// A `ServerNode` instance is normally passed to [`oak::run_event_loop`], to
```

rbehjati:
> thanks! Between the PR message and the various commit, I am not sure what the final commit message you are intending to use is. Perhaps squash your own commits together and give the final commit the message you would like to use for merging? Unless there is a nicer way of doing it that I'm not aware of :)

Sure! My intention was to squash when merging. I can squash first :)

PR #653, opened 2020-02-26
Refactor runner script

tiziano88:
The data types are much clearer now, and it's getting closer to being
able to support async steps via Tokio (in a future PR).

Add support for timing individual commands.

# Checklist

      construction.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -20,7 +20,7 @@
 //! To invoke, run the following command from the root of the repository:
 //!
 //! ```
-//! cargo run --manifest-path=./runner/Cargo.toml
+//! cargo run --package=runner
 //! ```
 
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


ipetr0v:
Why do we need async steps in `runner` if they are already provided by `cloudbuild`?

runner/src/main.rs
```diff
@@ -125,125 +124,153 @@ fn is_cargo_workspace_file(path: &PathBuf) -> bool {
...
+                        "--",
+                        "--deny=warnings",
+                    ],
+                ),
+            })
+            .collect(),
+    }
+}
+
+fn run_bazel_build() -> Step {
```
ipetr0v: Should it be called `run_bazel_build_all`?

runner/src/main.rs
```diff
@@ -33,9 +33,9 @@ use internal::*;
 fn main() -> Result<(), Box<dyn std::error::Error>> {
```
ipetr0v: If it possible to run a server or examples using the `runner`, or it only runs Cargo related tasks and Bazel builds?

tiziano88: I am going to add that functionality in the next PR.

tiziano88:
> Why do we need async steps in `runner` if they are already provided by `cloudbuild`?

- cloudbuild parallelism is very coarse grain (only scripts can be run in parallel in different steps)
- it is not easy to run locally the same steps that cloudbuild runs, apart from looking at `.cloudbuild.yaml` and trying to run them in the same order; `runner` is supposed to work both locally and in cloudbuild

runner/src/main.rs
tiziano88: I'll rename later when I add more precise rules anyways.

PR #656, opened 2020-02-27
oak_runtime: Move platform module to eal

blaxill:
Isolates platform specific things not provided by `no_std` from the runtime into an eal crate. The intention so far is to keep these items std compatible for now. This brings the rust runtime closer to being able to run under Asylo. 

# Checklist

      construction.


Cargo.toml
```diff
@@ -28,7 +28,8 @@ members = [
 # > explicit about the dependency relationships between the crates.
 [patch.crates-io]
 # Oak.
-asylo_rust = { path = "oak/asylo_rust" }
+eal = { path = "oak/server/rust/eal" }
```
tiziano88: Not sure if it's too late for this, but I'm not a big fan of EAL, especially since in some cases the "enclave" may be a VM, or there may be no enclave at all (e.g. dev server).

How about `oak_platform` and `oak_platform_asylo`, or something like that?

Cargo.toml
```diff
@@ -53,4 +54,4 @@ translator_common = { path = "examples/translator/common" }
 expect = { path = "third_party/expect" }
 grpc-compiler = { path = "third_party/grpc-rust/grpc-compiler" }
 protoc-rust-grpc = { path = "third_party/grpc-rust/protoc-rust-grpc" }
-rust = { path = "third_party/rust/src/libstd" }
+rust_asylo = { path = "third_party/rust/src/libstd" }
```
tiziano88: I'm a bit confused as to what should go in `eal_asylo` and what in `rust_asylo` now, could you clarify? I guess that's part of #656, so perhaps it makes sense to reply there?

oak/server/rust/eal/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#![cfg_attr(feature="no_std", no_std)]
+
+#[cfg(feature = "asylo")]
+extern crate eal_asylo;
```
tiziano88: I thought this is not necessary in Rust 2018 -- or is it still needed for optional deps?

oak/server/rust/eal/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+#[cfg(feature = "asylo")]
+pub use eal_asylo::{Mutex, spawn, JoinHandle, RwLock, thread_id};
```
tiziano88: I'd imagine `eal_asylo` to expose the same things as `imp` below, so presumably it should be fine to `pub use eal_asylo::*` here?

Cargo.toml
blaxill: I'm happy with whatever name we go with, but it makes sense to decide before this PR. @daviddrysdale do you have a preference over `oak_platform` vs `eal` vs something else?

tiziano88: Other alternatives:

- oak_env (similar to https://doc.rust-lang.org/reference/conditional-compilation.html#target_env)
- oak_sys

tiziano88: cc @project-oak/core if anyone has any suggestion / preference.

blaxill: Done, switched to oak_platform.

Cargo.toml
blaxill: - rust_asylo is in third_party as its derived from rust_lang/rust, and intends to be as close to standard as possible
- oak_platform_rust is an implementation of oak_platform for asylo that uses rust_asylo

I agree there is overlap here. oak_platform_rust could be merged into either rust_asylo or as oak_platform. Currently `oak_platform_rust` is mostly just what was previously in `oak/asylo_rust`. This is worth revisiting when the implementation of "eal like things" becomes clearer, I've added `TODO(#662)`.

#656 is this PR did you mean to link something else?

oak/server/rust/eal/src/lib.rs
blaxill: You're correct, it's not needed even for optional dependencies.

oak/server/rust/eal/src/lib.rs
blaxill: Done

Cargo.toml
tiziano88: Uhmm I don't remember which issue I meant to point to, anyways I think #662 covers it now, thanks!

oak/server/rust/oak_platform/Cargo.lock
```diff
@@ -0,0 +1,5 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+[[package]]
+name = "eal"
```
tiziano88: Is this still needed?

oak/server/rust/oak_platform_asylo/src/lib.rs
```diff
@@ -29,69 +30,38 @@
...
-pub fn spawn<F>(f: F) -> io::Result<thread::Thread>
+pub fn spawn<F>(f: F) -> Thread
 where
     F: FnOnce() -> (),
     F: Send + 'static,
 {
-    unsafe {
-      thread::Thread::new(Box::new(f))
-    }
+    unsafe { Thread::new(Box::new(f)) }.expect("could not spawn thread!?")
```
tiziano88:
```suggestion
    unsafe { Thread::new(Box::new(f)) }.expect("could not spawn thread‽")
```

... (just kidding)

blaxill:
> Do you think we could [merge the crates]

Yeah definitely an option in the future. Having them separate helps me do more experimental work currently though.

Issue #627, opened 2020-02-21
Implement experimental split gRPC Rust server


conradgrobler:
The idea is to be able to support networking in an untrusted service VM outside of the enclave. The TLS termination and gRPC processing should happen inside the enclave.

A first step towards this goal is to split the gRPC server across two processes, with one handling networking and the second terminating TLS and processing the requests. The IPC mechanims between the two is TBD.

conradgrobler:
- Use [Tokio](https://github.com/tokio-rs/tokio) as the front-end to handle incoming network connections
  - Run in a separate process for now (not yet a separate VM)
  - Proxy all calls to gRPC via a custom communication channel
- Use [Tonic](https://github.com/hyperium/tonic) as a gRPC server
  - It seems to be the most actively developed and standards-compliant Rust native gRPC library
  - Supports manually setting up low-level connections
- Create new modules to encapsulate communication channels between the front-end and the gRPC server
  - Version 1 will be based on [UDS](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html)
    - An initial connection will be set up between the front-end and the gRPC server as a control channel which will be used to set up future connections
    - For each incoming connection the front-end and gRPC processes will set up a new UDS connection
    - Number of concurrent connections will be limited by maximum number of open file descriptors allowed in OS
  - Version 2 will use shared memory to facilitate communication between the front-end and the gRPC server
    - Design TBD
    - Might use https://docs.rs/crate/shared_memory/ or something similar
    - Will probably use a pair of ring buffers for each connection
    - Not too important to optimise yet, as specific implementations for different hypervisor environments will probably be required when this moves to using inter-VM communication

tiziano88:
Does this boil down to having a UDS pseudo-node or similar in the Oak Runtime? Then it does not even matter what language we write the thing on the other side, FWIW it may even be C++ (and perhaps it should).

I guess TLS is a good starting point, but we would still need to figure out how to implement remote attestation on it. Alternatively we could go directly with [EKEP](https://asylo.dev/docs/concepts/ekep.html) instead.

conradgrobler:
The plan is not for a UDS psuedo node. The UDS will only be used as a temporary transport mechanism and will transport encrypted data. The encryption (TLS or otherwise) will terminate in the second process (to allow termination in a VM-based enclave in future).

The ideas for now is that this is a stand-alone proof of concept as a first step. Once proven and tested, it can be incorporated into the Rust runtime.

PR #658, opened 2020-02-28
Implement naive proxy using new TCP connections

conradgrobler:
This initial version of the proxy makes a new TCP connection to the server for each incoming request from the client and copies the bytes directly. 

This is just a test to ensure that the TLS still works when the connections are being proxied before implementing a UDS-based proxy as described in #627.

# Checklist

      construction.

### DIFF
--- a/experimental/split_grpc/proxy/Cargo.toml
+++ b/experimental/split_grpc/proxy/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "split_grpc_proxy"
+version = "0.1.0"
+authors = ["Conrad Grobler <grobler@google.com>"]
+edition = "2018"
+
+[dependencies]
+env_logger = "0.7"
+futures = "0.3"
+log = "0.4"
+tokio = { version = "0.2", features = ["full"] }
--- a/experimental/split_grpc/proxy/Cargo.toml
+++ b/experimental/split_grpc/proxy/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Conrad Grobler <grobler@google.com>"]
 edition = "2018"
 
 [dependencies]
-env_logger = "0.7"
-futures = "0.3"
-log = "0.4"
-tokio = { version = "0.2", features = ["full"] }
+env_logger = "*"
+futures = "*"
+log = "*"
+tokio = { version = "*", features = ["full"] }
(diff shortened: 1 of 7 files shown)


experimental/split_grpc/proxy/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "split_grpc_proxy"
+version = "0.1.0"
+authors = ["Conrad Grobler <grobler@google.com>"]
+edition = "2018"
+
+[dependencies]
+env_logger = "0.7"
+futures = "0.3"
+log = "0.4"
+tokio = { version = "0.2", features = ["full"] }
```
tiziano88: I personally don't believe in semver, so I have been using "*" everywhere, up to you if you think it makes sense to specify anything specific here.

experimental/split_grpc/proxy/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+            let down = spawn(async move { io::copy(&mut reader_down, &mut writer_down).await });
+
+            match up.await? {
+                Err(e) => error!("Error copying to server: {:?}", e),
+                Ok(count) => info!("Copied {:?} bytes to server", count),
+            }
+            match down.await? {
+                Err(e) => error!("Error copying to client: {:?}", e),
+                Ok(count) => info!("Copied {:?} bytes to client", count),
+            }
```
tiziano88: Should this be a [`select`](https://docs.rs/tokio/0.2.12/tokio/macro.select.html) so that `up` and `down` will be executed in parallel? My understanding is that otherwise `down` will only start running after `up` has finished, or am I missing something?

experimental/split_grpc/proxy/Cargo.toml
conradgrobler: I have changed it to *. I have been following the examples for tonic and tokio, and all the samples were using specific versions. I assumed it was because of API incompatibility between different versions, but I guess if specific versions are locked down in the cargo.lock file it should not matter.

experimental/split_grpc/proxy/src/main.rs
conradgrobler: The calls to spawn should schedule these as two tasks in tokio and these should run independently. The match on the await should just mean it won't log the result of down before up has completed. These are running concurrently at the moment, because the TLS handshake before the request means that there up and down copies are being interleaved.

blaxill:
LGTM

PR #666, opened 2020-03-03
oak_abi: Switch to Prost from protobuf-rust

blaxill:
Prost appears close to achieving `no_std` support see https://github.com/danburkert/prost/pull/215. This switches to using prost for oak_abi. We could then continue to use protobuf for labels(except using prost vs protobuf-rust), and can reenable protobuf configuration for rust (instead of feature gating it).  

# Checklist

      construction.


oak/server/rust/oak_abi/Cargo.toml
```diff
@@ -5,14 +5,9 @@ authors = ["David Drysdale <drysdale@google.com>"]
...
 license = "Apache-2.0"
 
-[features]
-std = ["protobuf"]
-no_std = []
-default = ["std"]
-
 [dependencies]
-protobuf = { version = "*", optional = true }
+prost = "0.6"
```
tiziano88: Could we use "*" here so it is more likely that all our crates will always depend on the same version?

oak/server/rust/oak_abi/build.rs
```diff
@@ -15,11 +15,17 @@
...
-        input: &["../../../../oak/proto/oak_api.proto"],
-        includes: &["../../../../oak/proto"],
-        customize: protoc_rust::Customize::default(),
-    })
-    .expect("protoc");
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    println!("cargo:rerun-if-changed=../../../../oak/proto/oak_api.proto");
+
+    let mut config = prost_build::Config::new();
+    config.out_dir("src/proto");
```
tiziano88: Should we use "OUT_DIR" here, as per https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script ? I know we were already not following that rule, but this may be a good opportunity to fix that.

oak/server/rust/oak_abi/build.rs
```diff
@@ -15,11 +15,17 @@
...
 fn main() {
-    oak_utils::run_protoc_rust(protoc_rust::Args {
-        out_dir: "src/proto",
-        input: &["../../../../oak/proto/oak_api.proto"],
-        includes: &["../../../../oak/proto"],
-        customize: protoc_rust::Customize::default(),
-    })
-    .expect("protoc");
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    println!("cargo:rerun-if-changed=../../../../oak/proto/oak_api.proto");
```
tiziano88: Add a reference to https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed.

tiziano88: Make the proto path a const and reuse here and below.

tiziano88: I'm not sure it makes a difference, but I would move this line after the actual code generation, so if things go wrong it will not have printed something incorrect.

oak/server/rust/oak_abi/build.rs
```diff
@@ -15,11 +15,17 @@
...
+
+    let mut config = prost_build::Config::new();
+    config.out_dir("src/proto");
+    config
```
tiziano88: I think `out_dir` and `compile_protos` can be chained all together (and avoid the `config` mut binding altogether).

oak/server/rust/oak_abi/build.rs
blaxill: Done. It does mean we have to include the generated file from OUT_DIR see new `proto.rs`.

oak/server/rust/oak_abi/build.rs
```diff
@@ -14,12 +14,15 @@
...
-        includes: &["../../../../oak/proto"],
-        customize: protoc_rust::Customize::default(),
-    })
-    .expect("protoc");
+    let proto_dir = path::Path::new("../../../../oak/proto/");
+    let oak_api_path = proto_dir.join("oak_api.proto");
+
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    // https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed
+    println!("cargo:rerun-if-changed={:?}", oak_api_path);
```
tiziano88: I think this may be:

```suggestion
    println!("cargo:rerun-if-changed={}", oak_api_path.to_str().unwrap());
```

oak/server/rust/oak_abi/build.rs
```diff
@@ -14,12 +14,15 @@
...
+
+    prost_build::compile_protos(&[&*oak_api_path], &[proto_dir]).unwrap();
```
tiziano88: Just curious, does it not compile without `&*`?

oak/server/rust/oak_abi/build.rs
blaxill: Done

oak/server/rust/oak_abi/build.rs
blaxill: In that case rust sees the arrays as `[PathBuf]` and `[Path]` but the definition of `compile_protos` requires them to be the same (for no real reason). I have moved the cast up to the definition line as it looks a little cleaner.

oak/server/rust/oak_abi/build.rs
tiziano88: Ah cool, thanks for the heads-up, fine to keep at the top then, let's make sure this is consistent across the various build scripts anyways (cc @conradgrobler ).

PR #667, opened 2020-03-04
Rerun build.rs only if proto file changed

### LINKED ISSUE #661: `split_grpc_*` crates are rebuilt every time
It seems that these targets are the only one that are not cached, and in fact are rebuilt every time. Perhaps something caused by the `build.rs` files.

Not a big deal, as those are pretty small crates, but it would be good to figure out what's going on.

```log
16:37:16 [tzn@box:~/src/oak] tzn_label_runtime± cargo check
   Compiling split_grpc_client v0.1.0 (/usr/local/google/home/tzn/src/oak/experimental/split_grpc/client)
   Compiling split_grpc_server v0.1.0 (/usr/local/google/home/tzn/src/oak/experimental/split_grpc/server)
    Finished dev [unoptimized + debuginfo] target(s) in 0.50s
16:37:17 [tzn@box:~/src/oak] tzn_label_runtime± cargo check
   Compiling split_grpc_server v0.1.0 (/usr/local/google/home/tzn/src/oak/experimental/split_grpc/server)
   Compiling split_grpc_client v0.1.0 (/usr/local/google/home/tzn/src/oak/experimental/split_grpc/client)
    Finished dev [unoptimized + debuginfo] target(s) in 0.38s
```

conradgrobler:
Fixes #661

# Checklist

      construction.

### DIFF
--- a/experimental/split_grpc/client/build.rs
+++ b/experimental/split_grpc/client/build.rs
@@ -16,20 +16,19 @@
 
 use std::path::Path;
 
-const PROTO_PATH: &'static str = "../../../examples/hello_world/proto/";
-const SOURCE_FILE: &'static str = "hello_world.proto";
-
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let file_path = std::format!("{}{}", PROTO_PATH, SOURCE_FILE);
+    let proto_path = Path::new("../../../examples/hello_world/proto");
+    let file_name = Path::new("hello_world.proto");
+    let file_path = proto_path.join(file_name);
     tonic_build::configure()
         .build_client(true)
         .build_server(false)
         .out_dir("src/proto")
-        .compile(&[Path::new(&file_path)], &[Path::new(PROTO_PATH)])?;
+        .compile(&[file_path.as_path()], &[proto_path])?;
 
     // Tell cargo to not rerun this script unless the proto file has changed.
     // This is required because the proto compiler is outputting the file into the source tree.
     // https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath
-    println!("cargo:rerun-if-changed={}", file_path);
+    println!("cargo:rerun-if-changed={}", file_path.display());
     Ok(())
 }
--- a/experimental/split_grpc/client/build.rs
+++ b/experimental/split_grpc/client/build.rs
@@ -14,15 +14,22 @@
 // limitations under the License.
 //
 
+use std::path::Path;
+
+const PROTO_PATH: &'static str = "../../../examples/hello_world/proto/";
+const SOURCE_FILE: &'static str = "hello_world.proto";
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let file_path = std::format!("{}{}", PROTO_PATH, SOURCE_FILE);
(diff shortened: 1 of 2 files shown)


experimental/split_grpc/client/build.rs
```diff
@@ -23,5 +23,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             &["../../../examples/hello_world/proto/hello_world.proto"],
             &["../../../examples/hello_world/proto/"],
         )?;
+    println!("cargo:rerun-if-changed=../../../examples/hello_world/proto/hello_world.proto");
```
tiziano88: According to https://doc.rust-lang.org/cargo/reference/build-scripts.html#life-cycle-of-a-build-script , 

> The build script will be rebuilt if any of its source files or dependencies change.

So why is this being re-run (before this PR)? AFAICT no source files or dependencies have changed.

blaxill: Good point, I think this is because running the scripts _does_ update the `mtime` of the newly generated `*.rs` file, which then causes cargo to think it's sources have changed. (i.e. it doesn't know that the generated file is an output rather than input.)

`cargo:rerun-if-changed=PATH` is then working (I haven't checked) because its actually removing these generated files from the "watched" path

experimental/split_grpc/client/build.rs
```diff
@@ -14,14 +14,22 @@
 // limitations under the License.
 //
 
+use std::path::Path;
+
+const PROTO_PATH: &'static str = "../../../examples/hello_world/proto/";
+const SOURCE_FILE: &'static str = "hello_world.proto";
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let file_path = std::format!("{}{}", PROTO_PATH, SOURCE_FILE);
```
tiziano88: Alternatively:

```rust
let proto_path = Path::new("../../../examples/hello_world/proto");
let file_name = Path::new("hello_world.proto");
let file_path = proto_path.join(file_name);
...
```

experimental/split_grpc/client/build.rs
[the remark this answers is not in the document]
tiziano88: Actually the recommended way would be to use `OUT_DIR`, does that not work for us? https://docs.rs/prost-build/0.6.1/prost_build/index.html

Also I think we should reorganise things, moving protos under the folder where they belong (runtime, ABI, SDK), and compiling them once only. Then the compiled artefacts may be re-esported so that they be referenced from other crates, but at least we don't need to deal with proto generation all over the place. WDYT?

cc @daviddrysdale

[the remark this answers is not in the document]
conradgrobler: Done.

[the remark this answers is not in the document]
conradgrobler: I think this is only a problem because I am compiling the proto files into the source tree. Tonic normally generates the file into the output directory, so this should not generally happen.

conradgrobler: I agree that this is because the generated .rs file is in the source tree.

experimental/split_grpc/client/build.rs
conradgrobler: Done.

PR #690, opened 2020-03-09
oak_runtime: Track Nodes with u64 NodeRefs

blaxill:
# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/mod.rs
+++ b/oak/server/rust/oak_runtime/src/node/mod.rs
@@ -21,8 +21,7 @@ use std::sync::Arc;
 
 use oak_abi::OakStatus;
 
-use crate::runtime::ChannelReader;
-use crate::RuntimeRef;
+use crate::{ChannelReader, NodeRef, RuntimeRef};
 
 mod logger;
 mod wasm;
@@ -73,6 +72,7 @@ impl Configuration {
         &self,
         config_name: &str, // Used for pretty debugging
         runtime: RuntimeRef,
+        _noderef: NodeRef,
         entrypoint: String,
         initial_reader: ChannelReader,
     ) -> Result<oak_platform::JoinHandle, OakStatus> {
--- a/oak/server/rust/oak_runtime/src/runtime/mod.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/mod.rs
@@ -27,7 +27,6 @@ use oak_abi::{ChannelReadStatus, OakStatus};
 use oak_platform::{JoinHandle, Mutex};
 
 use log::debug;
-use rand::RngCore;
 
 use crate::message::Message;
 use crate::node;
@@ -40,7 +39,7 @@ type Channels = Vec<Weak<channel::Channel>>;
 #[derive(Debug)]
 struct Node {
     reference: NodeRef,
-    join_handle: Option<JoinHandle>,
+    join_handle: JoinHandle,
 }
 
 #[derive(PartialEq, Eq, Hash, Clone, Copy, Debug)]
@@ -102,15 +101,11 @@ impl Runtime {
 
         // Unpark any threads that are blocked waiting on channels.
         for thread in threads.values() {
-            thread.join_handle.as_ref().unwrap().thread().unpark();
+            thread.join_handle.thread().unpark();
         }
 
         for (_, thread) in threads.drain() {
-            thread
-                .join_handle
-                .unwrap()
-                .join()
-                .expect("Failed to join handle");
+            thread.join_handle.join().expect("Failed to join handle");
         }
     }
 
@@ -345,7 +340,13 @@ impl RuntimeRef {
                 )
             }) {
             Ok(join_handle) => {
-                nodes.get_mut(&reference).unwrap().join_handle = Some(join_handle);
+                nodes.insert(
+                    reference,
+                    Node {
+                        reference,
+                        join_handle,
+                    },
+                );
             }
             Err(e) => {
                 nodes.remove(&reference);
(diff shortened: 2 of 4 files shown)


oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -36,6 +37,15 @@ pub use channel::{ChannelEither, ChannelReader, ChannelWriter, ReadStatus};
 
 type Channels = Vec<Weak<channel::Channel>>;
 
+#[derive(Debug)]
+struct Node {
+    reference: NodeRef,
+    join_handle: Option<JoinHandle>,
```
tiziano88: I guess this may become `HashSet<JoinHandle>` in the future, if / when we support multiple threads per node?

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -81,20 +91,28 @@ impl Runtime {
     /// Thread safe method for signaling termination to a `Runtime` and waiting for its node
     /// threads to terminate.
     pub fn stop(&self) {
-        let handles = {
-            let mut node_threads = self.node_threads.lock().unwrap();
+        {
+            let nodes = self.nodes.lock().unwrap();
+            debug!("nodes: {:?}", nodes);
```
tiziano88: Is this intended? Why not just move this after line 99?

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -73,6 +72,7 @@ impl Configuration {
         &self,
         config_name: &str, // Used for pretty debugging
         runtime: RuntimeRef,
+        _noderef: NodeRef,
```
tiziano88: What is this used for?

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -289,6 +307,31 @@ impl Runtime {
             }
         }
     }
+
+    /// Assigns a randomized reference to a [`Node`].
+    fn new_node_handle(&self) -> NodeRef {
+        let mut handle;
+
+        loop {
```
tiziano88: I feel it would be sufficient / simpler to just use an incrementing (atomic) counter, similar to how Process IDs work in OSs.

oak/server/rust/oak_runtime/src/node/mod.rs
blaxill: When a Node is created and the runtime generates a new reference/id, i pass it here into the newly created threads so it knows its own reference. Currently unused as the runtime/channel methods don't use the references.

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Yeah, this can be expanded

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Removed: left over from checking things were working as intended :)

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Done. Argh, I had randomized it as I think thats the direct you want to go with Channels.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -47,7 +56,8 @@ pub struct Runtime {
     configurations: HashMap<String, node::Configuration>,
     terminating: AtomicBool,
     channels: Mutex<Channels>,
-    node_threads: Mutex<Vec<JoinHandle>>,
+    nodes: Mutex<HashMap<NodeRef, Node>>,
+    next_node: AtomicU64,
```
tiziano88: (optional)

```suggestion
    next_node_reference: AtomicU64,
```

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -312,20 +323,36 @@ impl RuntimeRef {
             return Err(OakStatus::ErrTerminated);
         }
 
-        let mut node_threads = self.node_threads.lock().unwrap();
+        let reference = NodeRef(self.next_node.fetch_add(1, SeqCst));
```
tiziano88: (optional) Could have a method `next_node_reference` or similar on `self`.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -312,20 +323,36 @@ impl RuntimeRef {
...
+                    self.clone(),
+                    reference,
+                    entrypoint.to_owned(),
+                    reader,
+                )
+            }) {
+            Ok(join_handle) => {
+                nodes.insert(
+                    reference,
+                    Node {
```
tiziano88: Maybe `Configuration::new_instance` should return a `Node` struct directly?

oak/server/rust/oak_runtime/src/runtime/mod.rs
tiziano88: I think channels will also not be randomized, but channel **handles** (read or write) will be, since they are the ones used as capabilities.

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Done

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Done `new_node_reference`

oak/server/rust/oak_runtime/src/runtime/mod.rs
blaxill: Currently the implementation of `Node` is private to the runtime, and would require being visible to the crate for this. I think the encapsulation to only the runtime is preferable, but we can change it in future PRs as necessary.

PR #702, opened 2020-03-13
Support passing handles inline in protobuf messages

### LINKED ISSUE #673: Experiment with passing handles between nodes in Protobufs
This would allow nodes to share channels with each other simply by passing them over an existing channel, encoding them in an ordinary protocol buffer.

The approach I am proposing for now is that we introduce methods in the Oak ABI to generate a kind of 'invite token' for a combination of a channel handle to send and a channel handle over which to send the token. This token can then be transmitted as part of an ordinary protobuf message, and the receiver can call upon the Oak ABI to exchange the invite token for an actual handle.

The experiment should answer the following questions:
- [ ] Can we make this work in a way that is ergonomic and safe for users?
- [x] Can we ensure handles are passed in a way that is type-safe?

Related to #426 and #420.

wildarch:
# Checklist

      construction.

Fixes #673.

These changes implement a way of passing channels inline in protobuf messages using channel invite tokens.

This is just an experiment for now. TODO list:
- [x] Add ABI calls to create and use invites
- [x] Support new ABI calls in Rust runtime
- [ ] Support new ABI calls in C++ runtime
- [ ] Define Protobuf format for handle invites
- [ ] Annotate handles in protobuf with the correct type (Current top priority)
- [ ] SDK changes to ergonomically convert to and from channel invites

### DIFF
--- a/examples/abitest/module_0/rust/src/lib.rs
+++ b/examples/abitest/module_0/rust/src/lib.rs
@@ -158,6 +158,8 @@ impl OakABITestService for FrontendNode {
         );
         tests.insert("AbsentGrpcClient", FrontendNode::test_absent_grpc_client);
 
+        tests.insert("ChannelInvite", FrontendNode::test_channel_invite);
+
         for (&name, &testfn) in &tests {
             if !include.is_match(name) {
                 debug!(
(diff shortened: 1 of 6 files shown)


tiziano88:
Thanks @wildarch . I thought about this some more, and I think this approach is a regression compared to the current bytes + handles serialization mechanism. Adding the concept of invite basically bypasses the introspection that the runtime should have on handles being sent around, because now the invites themselves are bearer tokens that the runtime has no visibility into; at that point the concept of separate handle space loses most of its value. Or am I missing something?

wildarch:
I think there are a few ways in which these invite tokens help us keep control over handles:
- Invite tokens can only be used once, so they can't be used to clone handles without asking the runtime for explicit permissions to do so (after obtaining the handle)
- The number of invite tokens for a particular handle is always known, so we also know when there is an unclaimed invite token to a particular handle

As far as I can see the only regression here is that nodes are free to pass invite tokens on to other nodes, and the runtime can only known this once the token is turned into a handle. If this is a concern we could restrict an invite token to only be claimable by a node that can read from a certain channel (i.e. linking invite tokens to the channel they are passed over).

Could you provide some more details on how you think this is a regression over the current approach?

I suppose I do not know enough about the requirements we have for this introspection capability of the runtime, maybe we should discuss this a bit?

Issue #646, opened 2020-02-25
Rust Oak Runtime ownership of Channels 


blaxill:
Move ownership of Channels to the Runtime

Part of #633

blaxill:
Fixed in #728

PR #728, opened 2020-03-20
Channel ownership

blaxill:
Move channel ownership (data structure and creation/deletion) into runtime #646

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/config.rs
+++ b/oak/server/rust/oak_runtime/src/config.rs
@@ -106,7 +106,11 @@ pub fn from_protobuf(
     Ok(config)
 }
 
-/// Configure a `Runtime` from the given protobuf `ApplicationConfiguration` and begin execution.
+/// Configure a [`Runtime`] from the given protobuf [`ApplicationConfiguration`] and begin
+/// execution. This returns a [`RuntimeRef`] reference to the created runtime, and a
+/// writeable [`Handle`] to send messages into the runtime. Creating a new
+/// channel and passing the write [`Handle`] into the runtime will enable messages to be
+/// read back out from the [`Runtime`].
 pub fn configure_and_run(
     app_config: ApplicationConfiguration,
 ) -> Result<(RuntimeRef, Handle), OakStatus> {
--- a/oak/server/rust/oak_runtime/src/config.rs
+++ b/oak/server/rust/oak_runtime/src/config.rs
@@ -28,7 +28,7 @@ use oak_abi::OakStatus;
 use crate::node;
 use crate::node::load_wasm;
 use crate::runtime;
-use crate::runtime::{ChannelWriter, Runtime, RuntimeRef};
+use crate::runtime::{ChannelRef, Runtime, RuntimeRef};
 
 /// Create an application configuration.
 ///
@@ -109,7 +109,7 @@ pub fn from_protobuf(
 /// Configure a `Runtime` from the given protobuf `ApplicationConfiguration` and begin execution.
 pub fn configure_and_run(
     app_config: ApplicationConfiguration,
-) -> Result<(RuntimeRef, ChannelWriter), OakStatus> {
+) -> Result<(RuntimeRef, ChannelRef), OakStatus> {
     let config = from_protobuf(app_config)?;
     Runtime::configure_and_run(config)
 }
--- a/oak/server/rust/oak_runtime/src/config.rs
+++ b/oak/server/rust/oak_runtime/src/config.rs
@@ -28,7 +28,7 @@ use oak_abi::OakStatus;
 use crate::node;
 use crate::node::load_wasm;
 use crate::runtime;
-use crate::runtime::{ChannelRef, Runtime, RuntimeRef};
+use crate::runtime::{Handle, Runtime, RuntimeRef};
 
 /// Create an application configuration.
 ///
@@ -109,7 +109,7 @@ pub fn from_protobuf(
 /// Configure a `Runtime` from the given protobuf `ApplicationConfiguration` and begin execution.
 pub fn configure_and_run(
     app_config: ApplicationConfiguration,
-) -> Result<(RuntimeRef, ChannelRef), OakStatus> {
+) -> Result<(RuntimeRef, Handle), OakStatus> {
     let config = from_protobuf(app_config)?;
     Runtime::configure_and_run(config)
 }
--- a/oak/server/rust/oak_runtime/src/node/wasm.rs
+++ b/oak/server/rust/oak_runtime/src/node/wasm.rs
@@ -449,7 +449,7 @@ impl WasmInterface {
             .map(|bytes| {
                 let handle = LittleEndian::read_u64(bytes);
 
-                self.readers.get(&handle).map(|x| *x)
+                self.readers.get(&handle).copied()
             })
             .collect();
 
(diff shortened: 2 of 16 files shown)


blaxill:
@tiziano88 how do you feel this is looking? I need to add a few doc comments but

- Channel references are u64
- ChannelEither is removed 
- Runtime holds all channels and channel references
- Follow up PR will add channel tracking to Nodes

tiziano88:
To clarify what I am thinking in terms of interface between runtime and nodes:

- nodes are identified by a unique id
- operations on the runtime (e.g. channel_read, etc.) get an additional parameter to identify the caller node by its id
- we introduce a new type (e.g. `RuntimeProxy`) that binds all these functions to a specific node id, and we pass this object to the node or pseudo-node instance.

e.g.

```rust
struct Runtime {
  fn channel_read(node_id: NodeId, channel_handle: Handle, /* ... */) { /* ... */ }
  fn channel_write(node_id: NodeId, channel_handle: Handle, /* ... */) { /* ... */ }
  /* ... */
}

struct RuntimeProxy {
  node_id: NodeId,
  runtime: Arc<Runtime>, // Or similar.
}

impl RuntimeProxy {
  fn channel_read(&self, channel_handle: Handle, /* ... */) {
    self.runtime.channel_read(self.node_id, channel_handle, /* ... */)
  }
  fn channel_write(&self, channel_handle: Handle, /* ... */) {
    self.runtime.channel_write(self.node_id, channel_handle, /* ... */)
  }
  /* ... */
}
```

And of course we do not pass a direct reference to the runtime to the node.

WDYT about this approach?

blaxill:
@tiziano88 

>channel ID, then channel Reference, and then ABI handle as concepts to refer to the same underlying object.

- Channel ID and channel reference are different- you're right I should have called ChannelRef Handle and I'll update that in this PR. Channel ID should also probably be also renamed but is specific to the implementation and is only needed to track channel allocation. That is, channel id is basically a reference but there isn't an appropriate lifetime. An alternative would be to use Arcs/Weak again:

~~~rust
pub readers: RwLock<HashMap<ChannelRef, Weak<Channel>>>, 
~~~

>I think instead we should provide an interface for nodes and pseudo-nodes that is as "safe" as possible, i.e. have the runtime directly generate channel handles, 

- This PR moves in that direction, I'll randomize handles, and then after adding tracking of channels in node in a follow up PR, we can remove specialised handling of handles can be removed from nodes/Wasm.rs

> To clarify what I am thinking in terms of interface between runtime and nodes:

I think I am moving things in this direction- with the following steps:

1. Add this PR, having all datastructures owned by the runtime.
2. Next PR, add tracking of Handles (currently ChannelRef here but Ill fix that this PR) to nodes and enforce nodes only access their own handles.
3. We should now be at your code example, and `RuntimeProxy` should be simply to add on top.

WDYT? 

> fn channel_read(node_id: NodeId, channel_handle: Handle,

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -115,22 +122,22 @@ impl Runtime {
...
     /// [`ChannelReadStatus::InvalidChannel`] is set for `None` readers in the slice. For `Some(_)`
     /// readers, the result is set from a call to `has_message`.
-    fn readers_statuses(&self, readers: &[Option<&ChannelReader>]) -> Vec<ChannelReadStatus> {
+    fn readers_statuses(&self, readers: &[Option<ChannelRef>]) -> Vec<ChannelReadStatus> {
         readers
             .iter()
             .map(|chan| {
                 chan.map_or(ChannelReadStatus::InvalidChannel, |chan| {
-                    self.channel_status(chan)
+                    match self.channel_status(chan) {
```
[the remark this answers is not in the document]
blaxill: Done, with `unwrap_or`

oak/server/rust/oak_runtime/src/config.rs
```diff
@@ -109,7 +109,7 @@ pub fn from_protobuf(
 /// Configure a `Runtime` from the given protobuf `ApplicationConfiguration` and begin execution.
```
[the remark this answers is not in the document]
blaxill: Argh I clarified on the `Runtime` and not in config. I will update here too…

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -86,21 +86,18 @@ struct WasmInterface {
 impl WasmInterface {
     /// Generate a randomized handle. Handles are random to prevent accidental dependency on
     /// particular values or sequence. See https://github.com/project-oak/oak/pull/347
-    fn allocate_new_handle(&mut self, channel: ChannelEither) -> AbiHandle {
+    fn allocate_new_handle(&mut self, channel: Handle, direction: HandleDirection) -> AbiHandle {
```
tiziano88: I'm still confused by `Handle` vs `AbiHandle`. Why do they both need to exist, and are both randomized? We should be able to just use the underlying `Handle` here?

blaxill: Apart from `oak_abi::Handle` there is
- `channel::Handle` which is `pub struct Handle(oak_abi::Handle);`, I was thinking it might be useful to have this completely opaque, but since the handles are randomized maybe there isn't much reason WDYT?
- `Wasm::AbiHandle` is `type AbiHandle = u64` and will be removed/replaced with oak_abi::handle. I was holding off on changing this until the next PR with tracking channels in nodes, so that I can remove all of the "handle" generation from Wasm and only use the runtime handles (and also enforce that nodes only access handles they own).

tiziano88: > Apart from `oak_abi::Handle` there is
> 
> * `channel::Handle` which is `pub struct Handle(oak_abi::Handle);`, I was thinking it might be useful to have this completely opaque, but since the handles are randomized maybe there isn't much reason WDYT?

That's a good question. On one hand, it would be nice for this to be type-safe in Rust (modulo unsafe code), in the sense of not being forgeable out of thin air. On the other hand, a Wasm node does rely on the fact that the Wasm module it is executing will provide handles as numbers, and the runtime around the Wasm interpreter will turn those numbers into Handles in order to pass them to the runtime.

How about the following proposal:

- ids (e.g. node_id, channel_id) are generated via incrementing counters, but are tracked and protected by the Rust type system (i.e. unforgeable, even if someone guessed the underlying number, they would not be able to wrap it in the correct type). We have to assume this at least for node_id, since if we are not trusting the Rust type system to provide this safety, then all bets are off (a malicious node may just impersonate another node when interacting with the runtime).
- handles only exist in the context of a Wasm node instance, and the Wasm node itself generates them randomly (**not** the Runtime), and keeps track of the mapping from each random / unguessable handle to a corresponding typesafe id object locally.

In this model, the `channel_read` ABI call would be defined in terms of ids (caller node id, target channel id) on the Runtime itself, but the Wasm node would instead:

- not expose the caller node id (since it is implicitly bound to the current node instance, which is enforced by a RuntimeProxy object or similar in the type system)
- accept a channel handle as parameter, which is then looked up by the Wasm node to figure out the underlying channel id (if any), which is then passed to the underlying call on the RuntimeProxy object. The Runtime never sees the handle, only the id.

WDYT?

@daviddrysdale thoughts?

> * `Wasm::AbiHandle` is `type AbiHandle = u64` and will be removed/replaced with oak_abi::handle. I was holding off on changing this until the next PR with tracking channels in nodes, so that I can remove all of the "handle" generation from Wasm and only use the runtime handles (and also enforce that nodes only access handles they own).

blaxill: > WDYT?

I like this idea, although pseudo nodes not having handles means that tracking when channels become orphaned is more complex.

Regarding this PR, WDYT can I merge and then work on tracking nodes with channels in a follow up PR ?

PR #761, opened 2020-03-24
Unpark actual threads waiting on channels

### LINKED ISSUE #760: Potential inconsistency in Rust Oak Runtime thread handling
When calling `Runtime::wait_on_channel` from a Node

https://github.com/project-oak/oak/blob/70da1899692a7ceb9eb0875fb3ee001645adb04e/oak/server/rust/oak_runtime/src/runtime/mod.rs#L160

the "current thread" is looked up using `thread::current`

https://github.com/project-oak/oak/blob/70da1899692a7ceb9eb0875fb3ee001645adb04e/oak/server/rust/oak_runtime/src/runtime/mod.rs#L164

and eventually parked, if no message is readily available on the specified channel(s)

https://github.com/project-oak/oak/blob/70da1899692a7ceb9eb0875fb3ee001645adb04e/oak/server/rust/oak_runtime/src/runtime/mod.rs#L206

Then, when calling `Runtime::stop`, the thread corresponding to the Node instance is unparked

https://github.com/project-oak/oak/blob/70da1899692a7ceb9eb0875fb3ee001645adb04e/oak/server/rust/oak_runtime/src/runtime/mod.rs#L117-L120

I think the assumption here is that this would be the same single thread that is associated with the Node instance, which is the case for the current implementation.

The problem arises if (or rather, when) we introduce Nodes that may spawn additional threads, since it may happen that the call to `Runtime::wait_on_channel` does not come from the initial thread, and therefore unparking the initial thread would not actually have the desired effect.

I think the correct (future-proof) thing to do would be for `Runtime::wait_on_channel` to keep track of the actual waiting thread itself, perhaps in a separate field (perhaps of type `HashMap<ThreadId, JoinHandle>`), which should also be able to deal with the fact that multiple threads from the same node may be waiting at the same time.

@blaxill WDYT?

tiziano88:
Fixes #760

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/runtime/mod.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/mod.rs
@@ -116,20 +116,43 @@ impl Runtime {
     /// Thread safe method for signaling termination to a [`Runtime`] and waiting for its node
     /// threads to terminate.
     pub fn stop(&self) {
-        let mut threads = {
+        // Take the list of nodes out of the runtime instance, and set the terminating flag; this
+        // will prevent additional nodes from starting to wait again, because `wait_on_channels`
+        // will return immediately with `OakStatus::ErrTerminated`.
+        let mut nodes = {
             let mut nodes = self.nodes.lock().unwrap();
             self.terminating.store(true, SeqCst);
 
             std::mem::replace(&mut *nodes, HashMap::new())
         };
 
-        // Unpark any threads that are blocked waiting on channels.
-        for thread in threads.values() {
-            thread.join_handle.thread().unpark();
+        // Unpark any threads that are blocked waiting on any channels.
+        for channel in self
+            .channels
+            .channels
+            .read()
+            .expect("could not acquire channel mapping")
+            .values()
+        {
+            for thread in channel
+                .waiting_threads
+                .lock()
+                .expect("could not acquire waiting threads for channel")
+                .values()
+            {
+                if let Some(thread) = thread.upgrade() {
+                    thread.unpark();
+                }
+            }
         }
 
-        for (_, thread) in threads.drain() {
-            thread.join_handle.join().expect("Failed to join handle");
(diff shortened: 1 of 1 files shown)


tiziano88:
It would be nice to have at least a unit test to verify that this is actually solving the issue, but I cannot think of an easy way to write one.

Also there is some similarity with 

https://github.com/project-oak/oak/blob/70da1899692a7ceb9eb0875fb3ee001645adb04e/oak/server/rust/oak_runtime/src/runtime/channel.rs#L39

but they are not quite the same. Do you think we should be using `Weak<Thread>` for this PR too?

tiziano88:
@blaxill actually changed things around now, PTAL, I am just reusing the existing map, and iterate over channels and then inside that over waiting threads, and unparking everything that is still upgradeable.

blaxill:
This is a valid solution, another one would be for nodes to add all `JoinHandles` to `runtime::Node` and iterate over them when stopping.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -107,20 +107,35 @@ impl Runtime {
...
+            {
+                if let Some(thread) = thread.upgrade() {
+                    thread.unpark();
+                }
+            }
+        }
+
+        let mut nodes = {
             let mut nodes = self.nodes.lock().unwrap();
             self.terminating.store(true, SeqCst);
```
blaxill: I think swapping the nodes out and setting terminating to true needs to occur before unparking the threads as the wait_on_channels loop will cause them to park again if runtime is not terminating.

blaxill: This is probably why CI is timing out.

tiziano88: Good point, fixed, and added some comments, PTAL.

PR #772, opened 2020-03-26
Rust gRPC server pseudo-node

### LINKED ISSUE #751: gRPC server pseudo-node in Rust Oak Runtime
Currently Rust Oak Runtime doesn't support TLS gRPC credentials that is supported by the C++ version:
https://github.com/project-oak/oak/blob/09c7d6d86dd3a6c3808f3157884dc702c5352a50/oak/server/dev/dev_oak_loader.h#L33-L34
Necessary, since we are moving towards using a Rust Oak Loader binary in the future #749.

ipetr0v:
This change adds a Rust gRPC server pseudo-node.

Fixes #751

TLS support is tracked by https://github.com/project-oak/oak/issues/806
Unit tests are tracked by https://github.com/project-oak/oak/issues/807

# Checklist
      construction.


oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -630,6 +631,7 @@ impl RuntimeRef {
         entrypoint: &str,
         label: &oak_abi::label::Label,
         reader: Handle,
+        writer: Handle,
```
ipetr0v: If I add a `writer` handle as a parameter to this function, it means that I will need to pass some `writer` here:
https://github.com/project-oak/oak/blob/589b2d20bbd9e0d8388f58e8b50b50522842013c/oak/server/rust/oak_runtime/src/node/wasm.rs#L198-L211
And IIUC we don't have `None` handles, right?
cc @tiziano88 @daviddrysdale @blaxill

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,195 @@
...
+            ChannelReadStatus::ReadReady => http::StatusCode::OK,
+            ChannelReadStatus::InvalidChannel => http::StatusCode::INTERNAL_SERVER_ERROR,
+            ChannelReadStatus::Orphaned => http::StatusCode::INTERNAL_SERVER_ERROR,
+        }
+    }
+}
+
+impl HttpStatusCode for OakStatus {
+    fn status_code(&self) -> http::StatusCode {
+        match self {
```
ipetr0v: WDYH about HTTP status codes that we need to return based on the `ChannelReadStatus` and `OakStatus` ?
cc @tiziano88 @daviddrysdale

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,195 @@
...
+    ) -> Result<hyper::Response<hyper::Body>, hyper::Error> {
+        let body = hyper::body::aggregate(req)
+            .await
+            .expect("Couldn't get request body");
+
+        let client_message = crate::message::Message {
+            // TODO: Is it copying the data?
+            data: body.bytes().to_vec(),
+            channels: vec![],
+        };
```
ipetr0v: I'm passing received data to the Runtime as it is (just copy bytes from the HTTP body).
Is this correct?
cc @tiziano88 @daviddrysdale

tiziano88: I don't think this is what the C++ runtime does (and what the SDK expects). Each incoming request should create a gRPC invocation message, which is a pair of channels (and no bytes), over which the actual request(s) can be read, and response(s) can be written. See https://github.com/project-oak/oak/blob/master/oak/server/invocation.h . Also, in your PR, how would clients reply to a particular request?

ipetr0v: Ok, will implement an invocation

tiziano88:
It may be worth hooking up some grpc tests (possibly the existing c++ ones) before proceeding too much with this.

ipetr0v:
Do you mean tests from `abitest`?
Because there a no unit tests for the C++ of gRPC server.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: IIUC `grpc_request` that is implemented in `oak_tests` just passes bytes to the `writer`:
https://github.com/project-oak/oak/blob/589b2d20bbd9e0d8388f58e8b50b50522842013c/sdk/rust/oak_tests/src/lib.rs#L105-L124

tiziano88:
It's fine to merge something that does not work yet, but yes, we should be in a better position to test everything once the rust `oak_loader` is in place

oak/server/rust/oak_runtime/src/node/grpc_server.rs
tiziano88: I think this is what gets sent to the node under test

https://github.com/project-oak/oak/blob/589b2d20bbd9e0d8388f58e8b50b50522842013c/sdk/rust/oak_tests/src/lib.rs#L133-L142

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+            data: vec![],
+            channels: vec![request_reader, response_writer],
+        };
+        
+        // Serialize gRPC request into a message.
+        let message = crate::Message { data: vec![], channels: vec![] };
+        if let Err(protobuf_error) = request.write_to_writer(&mut message.data) {
+            let error = format!("Failed to serialize GrpcRequest message: {}", protobuf_error);
+            return Err(error);
+        }
```
tiziano88: It would be nice to have more idiomatic error handling using `?`, it would make things more readable.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        Ok(response_reader)
+    }
+        
+    /// Processes a gRPC response from the Oak node.
+    /// Returns a HTTP status and the response body or an error.
+    fn process_response(&self, response_reader: Handle) -> Result<Vec<u8>, String> {
+        match self.runtime.wait_on_channels(self.node_id, &[Some(response_reader)]) {
```
tiziano88: Use `?` here too.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+                oak_status
+            );
+            return Err(error);
+        }
+
+        Ok(response_reader)
+    }
+        
+    /// Processes a gRPC response from the Oak node.
+    /// Returns a HTTP status and the response body or an error.
```
tiziano88: Where is the status returned?

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+use std::{
+    thread,
+    thread::{spawn, JoinHandle},
+};
+use futures_executor::block_on;
+use protobuf::Message;
+use protobuf::well_known_types::Any;
+
+use oak_abi::{ChannelReadStatus, OakStatus};
+use oak::grpc::GrpcRequest;
```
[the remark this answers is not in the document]
ipetr0v: Agree, but currently it's in the `oak/src/proto` - so it's a generated Rust Protobuf file.
Is is ok to move it to `oak_abi`?

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+                    vec![],
+                ));
+            }
+        };
+
+        // Send response back to the gRPC client.
+        Ok(Self::http_response(http::StatusCode::OK, response_body))
+    }
+    
+    fn decode_grpc_request(grpc_method: &str, http_body: &dyn hyper::body::Buf) -> Result<Box<GrpcRequest>, String> {
```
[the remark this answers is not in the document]
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: Old comment, deleted.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+            self.node_id,
+            &oak_abi::label::Label::public_trusted()
+        );
+        let (response_writer, response_reader) = self.runtime.new_channel(
+            self.node_id,
+            &oak_abi::label::Label::public_trusted()
+        );
+        
+        // Create a notification message and attach the method-invocation specific channels to it.
+        let notification = crate::Message {
```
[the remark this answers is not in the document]
ipetr0v: Done.
This class exists but it's not a message:
https://github.com/project-oak/oak/blob/158ff848068d7c81ebb56dbca2ebeea87fea3a3a/sdk/rust/oak/src/grpc/invocation.rs#L19-L22

oak/server/rust/oak_runtime/src/node/grpc_server.rs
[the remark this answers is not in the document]
ipetr0v: Created an issue https://github.com/project-oak/oak/issues/799

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+
+#[derive(Clone)]
+struct GrpcService {
+    config_name: String,
+    runtime: RuntimeRef,
+    node_id: NodeId,
+    writer: Handle,
+}
+
+impl GrpcService {
```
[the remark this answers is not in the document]
ipetr0v: Created node thread uses `Service` inside it for a Hyper server.
And `GrpcServerNode` contains a thread reference.
So it would require creating a custom (non-`derive`) `clone` function that doesn't copy this thread reference and sets it to `None`.
Not sure if it will be readable.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    // Creates an HTTP response message.
+    fn http_response(status: http::StatusCode, body: Vec<u8>) -> hyper::Response<hyper::Body> {
+        let mut response = hyper::Response::new(hyper::Body::from(body));
+        *response.status_mut() = status;
+        response
+    }
+
+    /// Processes a gRPC request from the client and forwards it to the Oak node.
```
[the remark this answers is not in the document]
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: Done

oak/server/rust/oak_runtime/src/runtime/mod.rs
ipetr0v: I have added `writer` as a parameter to the `node_create`:
https://github.com/project-oak/oak/blob/c2b2ff4b666de2a4f5fe7ff7cbfadbfc00af69f2/oak/server/rust/oak_runtime/src/runtime/mod.rs#L773-L780
But this function is called inside a Wasm node, when `oak_abi` funciton `node_create` is called (and it doesn't provide a `writer`).

Is it ok to search for this `writer` in the list of channels and pass to the `node_create` in the Wasm node, or it may affect the Control Flow?

oak/server/rust/oak_runtime/src/node/grpc_server.rs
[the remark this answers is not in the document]
tiziano88: Makes sense, but perhaps at least create the `GrpcService` instance as part of `GrpcServerNode` itself, rather than repeat all the fields?

Also I am confused by the naming, why is one called `service` and the other `server`? I think we should stick to one of them, and I think `server` would be appropriate.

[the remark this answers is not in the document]
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,255 @@
...
+                            })
+                    })
+                    .or_else(|error| {
+                        error!("Couldn't process gRPC request: {}", error);
+                        Ok(Self::http_response(
+                            http::StatusCode::INTERNAL_SERVER_ERROR,
+                            vec![],
+                        ))
+                    })
+            })
```
tiziano88, line 125: Could this be encapsulated by a `Result<hyper::Response<hyper::Body>, Error>`, where `Error` is a newly defined enum, and just convert `Error` to the appropriate HTTP response here? Then you can use `?` throughout.

ipetr0v: Thanks, now it looks much nicer!
Done

ipetr0v:
@conradgrobler I was planning to add TLS and tests in separate PRs - not to clutter this one.

oak/server/rust/oak_runtime/src/runtime/mod.rs
tiziano88: I think perhaps we should have a pre-requisite to this PR to make gRPC node creation more similar to the creation of other pseudo-nodes, i.e. let the wasm node create them. This is what we do for logging, for instance.

ipetr0v: I already done this in this PR:
https://github.com/project-oak/oak/pull/772/files#diff-ce2a63846754333df981075c4b07ec96R93-R101

tiziano88: e.g.:

https://github.com/project-oak/oak/blob/006b9210e02e6483d298d24ec2940cff394906ee/sdk/rust/oak/src/logger/mod.rs#L80-L99

this is what creates a log pseudo-node, and we also have a default config name for it. We could have the same for gRPC.

ipetr0v: So pseudo-nodes should be created by the Oak Node, right?
But Runtime also creates a pseudo-node (and passes it a reader):
https://github.com/project-oak/oak/pull/772/files#diff-6c581d36a36131023ea8060843bb2dffR126

tiziano88: Ah, makes sense, but then I think we should also flip the direction of the channel, so that the gRPC server pseudo-node would have an incoming *read* channel, like every other node, over which the caller passes a single write end of a channel to receive invocations. Would that work?

ipetr0v: Yes, I think it will work.
So before starting the server, gRPC pseudo-node should wait on the channel to receive the writer.

tiziano88: I think the server may as well start as soon as possible, but then at the first incoming gRPC request it would block (`wait_on_channels`) to receive the channel to which to write the invocation, store it somewhere in its local state, and then use that channel for all subsequent invocations.

ipetr0v: Done
Created an `init` function:
https://github.com/project-oak/oak/pull/772/files#diff-1211ded35ba967a10825322cdf0cbd20R362-R378

It returns a **read** handle to read gRPC invocations from. This handle should be used by the Runtime, but I haven't found a loop/server inside the Runtime that reads invocations from a channel and dispatches them (runs corresponding gRPC method).

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,334 @@
...
+        } else {
+            error!("Couldn't read channel: {:?}", read_status[0]);
+            Err(GrpcServerError::ResponseProcessingError)
+        }
+    }
+}
+
+/// Oak Node implementation for the gRPC server.
+impl super::Node for GrpcServerNode {
+    fn start(&mut self) -> Result<(), OakStatus> {
```
conradgrobler, line 286: The gRPC seems to be combining two different Hyper concepts into one: The server, which is listening for new requests and the service, which handles each request. I think this is why there are so many clones statements needed.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -0,0 +1,334 @@
...
+                    async move {
+                        Ok::<_, hyper::Error>(hyper::service::service_fn(move |req| {
+                            let request_server = connection_server.clone();
+                            async move { request_server.serve(req).await }
+                        }))
+                    }
+                });
+
+                // Start the HTTP server.
+                let result = block_on(hyper::Server::bind(&server.address).serve(service));
```
conradgrobler, line 315: I think this will block as long as the server is listening on the port, which means that `start` will never return.

Also, I think this should not be using `block_on`, which uses a threadpool executor, but should rather be using Tokio executors as Hyper is based on Tokio. (https://docs.rs/tokio/0.2.16/tokio/runtime/struct.Runtime.html)

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: This logic was merged into the Node based on this thread: https://github.com/project-oak/oak/pull/772#discussion_r401123844
But even if we will have a separate `service` object, we'll still need to clone it, since it should live inside a Future.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: This is happening inside a newly created thread: https://github.com/project-oak/oak/pull/772/files/327276097e2c3593e513d92efa4f8ce713ab6a0a#diff-475373f331232762d978c36eb8027376R288-R290
Same is happening in Logger node:
https://github.com/project-oak/oak/blob/006b9210e02e6483d298d24ec2940cff394906ee/oak/server/rust/oak_runtime/src/node/logger.rs#L61-L65
Where `logger` is an infinite loop.

ipetr0v: Done.
Added a `tokio` runtime for executing a Future.
https://github.com/project-oak/oak/pull/772/files#diff-475373f331232762d978c36eb8027376R313-R317

oak/server/rust/oak_runtime/src/node/grpc_server.rs
conradgrobler: As part of the cleanup in #813, it might be worth looking at using a [Tower service](https://github.com/hyperium/hyper/blob/master/examples/tower_server.rs#L14) (implemented as a gRPC service node), which could encapsulate the needed state so that it does not have to be cloned so many times. It would also reduce the need for so many nested `move`s and `async move`s.

sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -350,3 +350,29 @@ where
...
+
+/// Default name for predefined node configuration that corresponds to a gRPC pseudo-Node.
+pub const DEFAULT_CONFIG_NAME: &str = "grpc_server";
+
+/// Initialize a gRPC pseudo-node with the default configuration.
+pub fn init_default() {
+    init(DEFAULT_CONFIG_NAME).unwrap();
+}
+
+/// Initialize a gRPC pseudo-node and pass it a handle to write invocations to.
```
tiziano88, line 362: I thought this method *returns* a handle on which to read invocations?

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: They are necessary in the Node thread, but not in the Future, yes

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -85,6 +97,22 @@ pub fn load_wasm(wasm_bytes: &[u8]) -> Result<Configuration, ConfigurationError>
     })
 }
 
+/// Parses an `address` from `String` to `SocketAddr`.
+pub fn parse_address(address: &str) -> Result<SocketAddr, ConfigurationError> {
+    address
+        .parse()
+        .map_err(|error| ConfigurationError::AddressParsingError(error))
```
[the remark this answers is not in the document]
ipetr0v: Done

PR #775, opened 2020-03-27
Wrap Runtime in Mutex

tiziano88:
Avoid relying on interior mutability.

# Checklist

      construction.


blaxill:
I'm curious the motivation for going in this direction, is it for simplicity? A global lock will not be performant if we want to support a larger number of threads.

tiziano88:
It was just to play around and learn a bit more about the Runtime and how Mutex and Arcs work, I don't think I'll even go ahead with any of this in the end. :)

PR #782, opened 2020-03-27
oak_tests: Don't pass RUSTFLAGS to `compile_rust_wasm`

blaxill:
# Checklist

      construction.

### DIFF
--- a/sdk/rust/oak_tests/src/lib.rs
+++ b/sdk/rust/oak_tests/src/lib.rs
@@ -36,6 +36,7 @@ pub fn compile_rust_wasm(cargo_path: &str, module_name: &str) -> std::io::Result
             "--target=wasm32-unknown-unknown",
             &format!("--manifest-path={}", cargo_path),
         ])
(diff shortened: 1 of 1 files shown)


sdk/rust/oak_tests/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub fn compile_rust_wasm(cargo_path: &str, module_name: &str) -> std::io::Result
             "--target=wasm32-unknown-unknown",
             &format!("--manifest-path={}", cargo_path),
         ])
+        .env_remove("RUSTFLAGS")
```
tiziano88: Perhaps we should just call [`env_clear`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env_clear).

blaxill: `env_clear` causes wasm builds to fail, I'm not sure what entries are needed exactly though.

PR #785, opened 2020-03-28
Add unit tests for Rust wasm node (#726)

### LINKED ISSUE #726: Implement Rust Wasm node tests
Similar to https://github.com/project-oak/oak/blob/master/oak/server/wasm_node_test.cc, but for https://github.com/project-oak/oak/blob/master/oak/server/rust/oak_runtime/src/node/wasm.rs.

Assigning to @rbehjati ,  perhaps @blaxill may help reviewing this too when ready?

rbehjati:
Adds unit-tests for the Rust wasm node, similar to those for the cpp wasm node.

- Since the Rust WasmNode takes a wasmi::Module as input (instead of a string as in cpp WasmNode), some scenarios in https://github.com/project-oak/oak/blob/master/oak/server/wasm_node_test.cc are not relevant here. In particular:
  - Empty string
  - Module with wrong version
- wat is used instead of wabt, since it is implemented completely in Rust.

Fixes #726

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/Cargo.toml
+++ b/oak/server/rust/oak_runtime/Cargo.toml
@@ -23,6 +23,7 @@ wasmi = { version = "*", default-features = false, features = ["core"] }
 
 [dev-dependencies]
 maplit = "*"
+wat = "*"
 
 [build-dependencies]
 oak_utils = "*"
--- a/oak/server/rust/oak_runtime/src/node/wasm.rs
+++ b/oak/server/rust/oak_runtime/src/node/wasm.rs
@@ -877,35 +877,112 @@ impl super::Node for WasmNode {
     }
 }
 
-#[test]
-fn wasm_invalid_args() {
-    use super::Node;
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::node::Node;
     use crate::runtime::TEST_NODE_ID;
-    let configuration = crate::runtime::Configuration {
-        nodes: HashMap::new(),
-        entry_module: "test_module".to_string(),
-        entrypoint: "test_function".to_string(),
-    };
-    let runtime_ref = crate::runtime::Runtime::create(configuration).into_ref();
-    let (_, reader_handle) =
-        runtime_ref.new_channel(TEST_NODE_ID, &oak_abi::label::Label::public_trusted());
+    use wat::{parse_file, parse_str};
 
-    // From https://docs.rs/wasmi/0.6.2/wasmi/struct.Module.html#method.from_buffer:
-    // Minimal module:
-    // \0asm - magic
-    // 0x01 - version (in little-endian)
-    let module =
-        wasmi::Module::from_buffer(vec![0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]).unwrap();
+    fn setup_node<S: AsRef<[u8]>>(buffer: S, entrypoint: &str) -> Box<dyn Node> {
+        let configuration = crate::runtime::Configuration {
+            nodes: HashMap::new(),
+            entry_module: "test_module".to_string(),
+            entrypoint: entrypoint.to_string(),
+        };
+        let runtime_ref = crate::runtime::Runtime::create(configuration).into_ref();
+        let (_, reader_handle) =
+            runtime_ref.new_channel(TEST_NODE_ID, &oak_abi::label::Label::public_trusted());
 
-    let mut node = WasmNode::new(
-        "test",
-        runtime_ref,
-        TEST_NODE_ID,
-        Arc::new(module),
-        "test_function".to_string(),
-        reader_handle,
-    );
+        let module = wasmi::Module::from_buffer(buffer).unwrap();
 
-    // The minimal module does not have the necessary entry point, so it should fail immediately.
-    assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+        Box::new(WasmNode::new(
+            "test",
+            runtime_ref,
+            TEST_NODE_ID,
+            Arc::new(module),
+            entrypoint.to_string(),
+            reader_handle,
+        ))
+    }
+
+    #[test]
+    fn wasm_starting_module_without_content_fails() {
+        // Loads an empty module that does not have the necessary entry point, so it should fail
+        // immediately.
+        let binary = parse_file("../../testdata/empty.wat").unwrap();
+
+        // An empty module is equivalent to: [0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]
+        // From https://docs.rs/wasmi/0.6.2/wasmi/struct.Module.html#method.from_buffer:
+        // Minimal module:
+        // \0asm - magic
+        // 0x01 - version (in little-endian)
+        assert_eq!(binary, vec![0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]);
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+    }
+
+    #[test]
+    fn wasm_starting_minimal_module_succeeds() {
+        let binary = parse_file("../../testdata/minimal.wat").unwrap();
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Ok(()), node.start());
+        node.stop();
+    }
+
+    #[test]
+    fn wasm_starting_module_missing_an_export_fails() {
+        let binary = parse_file("../../testdata/missing.wat").unwrap();
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+    }
+
+    #[test]
+    fn wasm_starting_module_with_wrong_export_fails() {
+        let binary = parse_file("../../testdata/minimal.wat").unwrap();
+        let mut node = setup_node(binary, "oak_other_main");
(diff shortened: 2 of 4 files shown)


oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+    }
+
+    #[test]
+    fn wasm_starting_minimal_module_succeeds() {
+        let binary = from_testfile("minimal.wat");
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Ok(()), node.start());
+        node.stop();
```
rbehjati: Does this make sense here? It causes the test to panic:
`thread '<unnamed>' panicked at 'remove_node_id: No such node_id', oak/server/rust/oak_runtime/src/runtime/mod.rs:585:28`
And why does not the test fail!

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
 }
 
-#[test]
-fn wasm_invalid_args() {
-    use super::Node;
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::node::Node;
     use crate::runtime::TEST_NODE_ID;
```
rbehjati: Is it necessary to have TEST_NODE_ID declared in runtime? Or can I move it to this module?

oak/server/rust/oak_runtime/Cargo.toml
```diff
@@ -20,6 +20,7 @@ oak_abi = "=0.1.0"
 prost = "*"
 rand = { version = "*", default-features = false, features = ["alloc"] }
 wasmi = { version = "*", default-features = false, features = ["core"] }
+wat = "*"
```
blaxill: Can this be moved into `dev-dependencies`?

oak/server/rust/oak_runtime/src/node/wasm.rs
blaxill: `TEST_NODE_ID` is used by `sdk/rust/oak_tests` so I think the declaration in runtime is the best place. Future tests in other parts of `oak_runtime` may also need to use `TEST_NODE_ID`

oak/server/rust/oak_runtime/src/node/wasm.rs
blaxill: Ahh the panic is because `remove_node_id` is called by the node when stopping, but `remove_node_id` doesn't filter for RUNTIME_NODE_ID/TEST_NODE_ID, which it should do. Can you add it is similar to https://github.com/project-oak/oak/blob/a005757bd99fe6642e3813108283d28e7fc42f61/oak/server/rust/oak_runtime/src/runtime/mod.rs#L254 ? Note that RUNTIME_NODE_ID/TEST_NODE_ID are the same value.

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
+            entrypoint: entrypoint.to_string(),
+        };
+        let runtime_ref = crate::runtime::Runtime::create(configuration).into_ref();
+        let (_, reader_handle) =
+            runtime_ref.new_channel(TEST_NODE_ID, &oak_abi::label::Label::public_trusted());
+
+        // From https://docs.rs/wasmi/0.6.2/wasmi/struct.Module.html#method.from_buffer:
+        // Minimal module:
+        // \0asm - magic
+        // 0x01 - version (in little-endian)
```
tiziano88: Not sure this comment belongs here, it was meant to explain where `vec![0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]` comes from in the original example.

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
+        path.pop();
+        path.pop();
+        path.push("testdata");
+        path.push(file_name);
+        parse_file(path).unwrap()
+    }
+
+    #[test]
+    fn wasm_with_invalid_args_fails() {
+        let buffer = vec![0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00];
```
tiziano88: This could use the comment above :)

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
+
+        // Wrong magic: The minimal module does not have the necessary entry point, so it should
```
tiziano88: Is this because of wrong magic or because of missing entry point? AFAICT the magic number is fine here.

Just to make sure we are on the same page, I mean this magic number: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,92 @@ impl super::Node for WasmNode {
...
+    #[test]
+    fn wasm_starting_module_without_content_fails() {
+        let binary = from_testfile("empty.wat");
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+    }
+
+    #[test]
+    fn wasm_starting_minimal_module_succeeds() {
+        let binary = from_testfile("minimal.wat");
```
tiziano88: I think these tests would read better if the wat files were inlined here as strings, since they are pretty minimal, anyways we can do that in the future if we want. Or perhaps for now you could come up with an additional test case in which the module is specified as a string, in addition to the current ones that we can keep as loading the wasm from a file. WDYT?

oak/server/rust/oak_runtime/src/node/wasm.rs
rbehjati: I have moved it to `wasm_starting_module_without_content_fails` now.

oak/server/rust/oak_runtime/src/node/wasm.rs
rbehjati: Done.

oak/server/rust/oak_runtime/src/node/wasm.rs
rbehjati: You are right. I just noticed that this test is the same as `wasm_starting_module_without_content_fails`, so I've merged the two (sort of)!

oak/server/rust/oak_runtime/src/node/wasm.rs
rbehjati: Are you thinking of creating the wasmi module from something like the following?
```
// copied from https://crates.io/crates/wat

let wat = r#"
    (module
        (func $foo)

        (func (export "bar")
            call $foo
        )
    )
"#;
```

tiziano88: Exactly.

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,94 @@ impl super::Node for WasmNode {
...
+        ))
+    }
+
+    fn from_testfile(path: &str) -> Vec<u8> {
+        parse_file(path).unwrap()
+    }
+
+    #[test]
+    fn wasm_starting_module_without_content_fails() {
+        // Loads an empty module, equivalent to: vec![0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00,
```
tiziano88: May as well leave around an actual test case with that vec, and here just mention that this tests corresponds to that. Or perhaps compare binary with that vec. Basically prove what you are stating in the comment, but as code :)

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,94 @@ impl super::Node for WasmNode {
...
+
+    #[test]
+    fn wasm_starting_module_with_wrong_signature_fails() {
+        let binary = from_testfile("../../testdata/wrong.wat");
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Err(OakStatus::ErrInvalidArgs), node.start());
+    }
+
+    #[test]
+    fn wasm_starting_module_with_no_channels_succeeds() {
```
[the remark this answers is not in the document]
tiziano88: Yes I'd like to keep just one around, just in case anyone in the future feels inspired to do something along those lines, it would be nice to have this as a starting point. Thanks!

oak/server/rust/oak_runtime/src/node/wasm.rs
```diff
@@ -877,35 +877,94 @@ impl super::Node for WasmNode {
...
+        (module
+            (type (;0;) (func (param i64)))
+            (func $oak_main (type 0))
+            (memory (;0;) 18)
+            (export "memory" (memory 0))
+            (export "oak_main" (func $oak_main)))
+        "#;
+        let binary = parse_str(wat).unwrap();
+        let mut node = setup_node(binary, "oak_main");
+        assert_eq!(Ok(()), node.start());
```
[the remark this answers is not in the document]
blaxill: I'm not familiar with reading the wat format but It looks to me as though the type signature of oak_main here is `func (param i64)`, so although it's not using the handle it still has the correct type signature which I would expect to succeed. If the type signature was different I would expect it to fail.

PR #794, opened 2020-03-31
Use named threads for nodes and loggers

### LINKED ISSUE #770: Use "named" threads in Rust runtime
See https://doc.rust-lang.org/std/thread/struct.Builder.html#method.name

matled:
Using named threads should help identifying threads during debugging.

https://doc.rust-lang.org/std/thread/struct.Builder.html#method.name

Example log:

2020-03-31 13:49:42,080 DEBUG [oak_runtime::runtime] wait_on_channels: \
    thread Thread { id: ThreadId(4), name: Some("node_logger=log") } re-woken
2020-03-31 13:49:42,080 DEBUG [oak_runtime::runtime] wait_on_channels: \
    thread Thread { id: ThreadId(3), name: Some("node=translator") } re-woken

Before it would show: Thread { id: ThreadId(4), name: None }

Fixes #770

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/logger.rs
+++ b/oak/server/rust/oak_runtime/src/node/logger.rs
@@ -14,16 +14,15 @@
 // limitations under the License.
 //
 
+use std::fmt::{self, Display, Formatter};
 use std::string::String;
 
 use log::{error, info};
 
 use oak_abi::OakStatus;
-use std::{
-    thread,
-    thread::{spawn, JoinHandle},
-};
+use std::thread::{self, JoinHandle};
 
+use crate::pretty_name_for_thread;
 use crate::proto::log::{Level, LogMessage};
 use crate::runtime::Handle;
 use crate::runtime::RuntimeProxy;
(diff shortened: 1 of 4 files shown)


oak/server/rust/oak_runtime/src/node/logger.rs
```diff
@@ -58,13 +54,15 @@ impl super::Node for LogNode {
...
         let reader = self.reader;
-        // TODO(#770): Use `std::thread::Builder` and give a name to this thread.
-        let thread_handle = spawn(move || {
-            let pretty_name = format!("{}-{:?}:", config_name, thread::current());
-            let result = logger(&pretty_name, runtime.clone(), node_id, reader);
-            info!("{} LOG: exiting log thread {:?}", pretty_name, result);
-            runtime.remove_node_id(node_id);
-        });
+        let thread_builder = thread::Builder::new().name(format!("node_logger={}", config_name));
+        let thread_handle = thread_builder
```
tiziano88: Could this just be inlined? Not sure the extra binding is adding much value here :)

oak/server/rust/oak_runtime/src/node/logger.rs
```diff
@@ -58,13 +54,15 @@ impl super::Node for LogNode {
...
+            .spawn(move || {
+                let pretty_name = format!("{}-{:?}:", config_name, thread::current());
```
tiziano88: The `Debug` implementation of `Thread` is not great, how about making it a bit better by introducing a separate function that can make it prettier? It may show the thread id and optional thread name in a more compact name than the current one. Feel free to create a separate issue and we can follow up there perhaps.

oak/server/rust/oak_runtime/src/node/logger.rs
```diff
@@ -58,13 +54,15 @@ impl super::Node for LogNode {
...
         let node_id = self.node_id;
         let reader = self.reader;
-        // TODO(#770): Use `std::thread::Builder` and give a name to this thread.
-        let thread_handle = spawn(move || {
-            let pretty_name = format!("{}-{:?}:", config_name, thread::current());
-            let result = logger(&pretty_name, runtime.clone(), node_id, reader);
-            info!("{} LOG: exiting log thread {:?}", pretty_name, result);
-            runtime.remove_node_id(node_id);
-        });
+        let thread_builder = thread::Builder::new().name(format!("node_logger={}", config_name));
```
tiziano88, line 57: Maybe `impl Display for LogNode` and then you can just print `self` here? And similarly for the other node.

oak/server/rust/oak_runtime/src/node/logger.rs
matled: Done.

I wasn't sure what the preferred style is here. I was also wondering if this makes sense to be a helper method, e.g. `spawn_with_name(name, f)`.

oak/server/rust/oak_runtime/src/node/logger.rs
matled: Done.

oak/server/rust/oak_runtime/src/node/logger.rs
matled: Done.

Also used in the runtime, see the example in the commit message.

PR #795, opened 2020-03-31
Add link to Slack instance in Readme

tiziano88:


sarahdeh:
Can you add a link to the external google group I created for discussions as well?https://groups.google.com/g/project-oak-discuss

PR #802, opened 2020-04-01
Move node thread creation to Runtime

tiziano88:
# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/logger.rs
+++ b/oak/server/rust/oak_runtime/src/node/logger.rs
@@ -16,11 +16,11 @@
 
 use std::fmt::{self, Display, Formatter};
 use std::string::String;
+use std::thread;
 
 use log::{error, info};
 
 use oak_abi::OakStatus;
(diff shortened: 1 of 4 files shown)


tiziano88:
This is still WIP, but the issue with moving thread creation into the Runtime is that the Wasm node now cannot validate the config upfront.

In particular, now this test case fails:

https://github.com/project-oak/oak/blob/c2b2ff4b666de2a4f5fe7ff7cbfadbfc00af69f2/examples/abitest/module_0/rust/src/lib.rs#L1034-L1041

@blaxill do you have any ideas?

@ipetr0v I hope this would help with #772

tiziano88:
@blaxill I think at some point you suggested a two-phase initialization process, which effectively I think boils down to a state machine for a node moving from "initial" to "validated" and eventually to "started"; the return value of the `node_create` ABI call would be the outcome of the transition from "initial" to "validated" in this model.

This all seems more complicated than it should be though.

tiziano88:
Done in #847

PR #811, opened 2020-04-02
Allow passing label to `node_create`

tiziano88:
# Checklist


tiziano88:
Yes, the tests pass but the examples do not work because the C++ runtime does not implement the corresponding ABI functionality now. I am happy to wait until we switch to the Rust version of the Runtime in the C++ loader (#801), that should allow the examples to run successfully with this PR.

PR #823, opened 2020-04-08
Use k8s secrets to provision TLS private key

### LINKED ISSUE #696: Investigate options for provisioning secrets to VM instances in GCP
Also understand how service accounts are provisioned onto VM instances, and whether the mechanism is reusable/extensible.

conradgrobler:
# Checklist

      construction.

Fixes #696

### DIFF
--- a/examples/aggregator/gcp/pod.yaml
+++ b/examples/aggregator/gcp/pod.yaml
@@ -14,14 +14,20 @@ spec:
         - name: tls-secret
           mountPath: '/etc/oak-secrets'
           readOnly: true
-      env:
-        - name: OAK_TLS_PRIVATE_KEY
-          value: '/etc/oak-secrets/docker.key'
+      command: ['./oak_runner']
+      args:
+        [
+          '--application=config.bin',
+          '--ca_cert=ca.pem',
+          '--cert_chain=docker.pem',
+          '--private_key=/etc/oak-secrets/docker.key',
+        ]
     - name: backend
       image: gcr.io/oak-ci/oak-aggregator-backend:latest
       env:
         - name: RUST_LOG
           value: 'info'
+      command: ['./aggregator_backend']
   volumes:
     - name: tls-secret
       secret:
--- a/examples/aggregator/gcp/pod.yaml
+++ b/examples/aggregator/gcp/pod.yaml
@@ -10,11 +10,22 @@ spec:
       image: gcr.io/oak-ci/oak-aggregator:latest
       ports:
         - containerPort: 8080
+      volumeMounts:
+        - name: tls-secret
+          mountPath: '/etc/oak-secrets'
+          readOnly: true
+      env:
+        - name: OAK_TLS_PRIVATE_KEY
+          value: '/etc/oak-secrets/docker.key'
     - name: backend
       image: gcr.io/oak-ci/oak-aggregator-backend:latest
       env:
         - name: RUST_LOG
           value: 'info'
+  volumes:
+    - name: tls-secret
+      secret:
+        secretName: tls-private-key
 ---
 apiVersion: v1
 kind: Service
--- a/examples/aggregator/gcp/pod.yaml
+++ b/examples/aggregator/gcp/pod.yaml
@@ -14,7 +14,6 @@ spec:
         - name: tls-secret
           mountPath: '/etc/oak-secrets'
           readOnly: true
-      command: ['./oak_runner']
       args:
         [
           '--application=config.bin',
@@ -27,7 +26,6 @@ spec:
       env:
         - name: RUST_LOG
           value: 'info'
-      command: ['./aggregator_backend']
   volumes:
     - name: tls-secret
       secret:
--- a/examples/aggregator/gcp/pod.yaml
+++ b/examples/aggregator/gcp/pod.yaml
@@ -17,8 +17,8 @@ spec:
       args:
         [
           '--application=config.bin',
-          '--ca_cert=ca.pem',
-          '--cert_chain=docker.pem',
+          '--ca_cert=/etc/oak-secrets/ca.pem',
+          '--cert_chain=/etc/oak-secrets/docker.pem',
           '--private_key=/etc/oak-secrets/docker.key',
         ]
     - name: backend
--- a/examples/aggregator/scripts/deploy
+++ b/examples/aggregator/scripts/deploy
@@ -10,5 +10,7 @@ source "${GLOBAL_SCRIPTS_DIR}/common"
 
 gcloud container clusters get-credentials aggregator
 kubectl create secret generic tls-private-key \
-  --from-file="${GLOBAL_ROOT_DIR}/examples/certs/docker/docker.key"
+  --from-file="${GLOBAL_ROOT_DIR}/examples/certs/docker/docker.key" \
+  --from-file="${GLOBAL_ROOT_DIR}/examples/certs/docker/docker.pem" \
+  --from-file="${GLOBAL_ROOT_DIR}/examples/certs/docker/ca.pem"
 kubectl apply --filename="${GLOBAL_ROOT_DIR}/examples/aggregator/gcp/pod.yaml"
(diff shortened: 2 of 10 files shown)


oak/server/loader/oak_runner_main.cc
```diff
@@ -28,6 +29,8 @@
 ABSL_FLAG(std::string, application, "", "Application configuration file");
 ABSL_FLAG(std::string, ca_cert, "", "Path to the PEM-encoded CA root certificate");
 ABSL_FLAG(std::string, private_key, "", "Path to the private key");
+ABSL_FLAG(std::string, private_key_env, "OAK_TLS_PRIVATE_KEY",
+          "Name of environment variable providing the path to the private key");
```
tiziano88, line 33: This seems... a bit convoluted :) I think the problem with the current config is that we don't pass flags to the binary from the k8s config, but I think we should, rather than setting them in the docker image. In fact I am thinking that the docker image should probably just be for the oak_loader binary only (i.e. the same for any application), and even the application configuration and any other files and flags should just be specified in `pod.yaml`. WDYT? cc @ipetr0v

conradgrobler: On the first point: it is convoluted because `entrypoint` to a docker image does not expand environment variables. I will test specifying the command in `pod.yaml` and removing the entry point form the docker image.

Another option is to always specify a fixed path and then mount different volumes depending on where it is run (eg mount secrets for k8s or local directory for local docker).

Regarding the second part, I am not sure what you mean with 'any other files'. `pod.yaml` does not provision files into the image. It just sets configuration. If the app config and the certificates are not already inside the container image it would need to be downloaded from somewhere.

tiziano88: Ah yes, you are right. Perhaps we should have a Docker image with the loader + app config at some known location, and then specify all the flags from `pod.yaml` (which will always include the flag to specify the app config from the known location, plus possibly other flags).

examples/aggregator/gcp/pod.yaml
```diff
@@ -10,11 +10,28 @@ spec:
       image: gcr.io/oak-ci/oak-aggregator:latest
       ports:
         - containerPort: 8080
+      volumeMounts:
+        - name: tls-secret
+          mountPath: '/etc/oak-secrets'
+          readOnly: true
+      command: ['./oak_runner']
```
tiziano88, line 17: FWIW I think it's fine to have the command in the Docker image (for both of them), we should still be able to override the rest of the args from here. I don't remember if it only works with ENTRYPOINT or CMD, or both, though.

oak/server/loader/oak_runner_main.cc
conradgrobler: Removed `entrypoint`s from docker images and added `command`s and `args` to `pod.yaml`. Reverted c++ changes.

examples/aggregator/scripts/deploy
```diff
@@ -1,11 +1,14 @@
...
+readonly GLOBAL_SCRIPTS_DIR="${GLOBAL_ROOT_DIR}/scripts/"
 # shellcheck source=scripts/common
 source "${GLOBAL_SCRIPTS_DIR}/common"
 
 "${SCRIPTS_DIR}/undeploy" || true
 
 gcloud container clusters get-credentials aggregator
-kubectl apply --filename=examples/aggregator/gcp/pod.yaml
+kubectl create secret generic tls-private-key \
+  --from-file="${GLOBAL_ROOT_DIR}/examples/certs/docker/docker.key"
```
tiziano88, line 13: We should rename this key to something like "gcp" rather than "docker", at some point?

conradgrobler: Created #826 to do this.

examples/aggregator/gcp/pod.yaml
conradgrobler: Yes, according to the [documentation](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes) it should override `entrypoint` if `command` is specified, but I don't think having it in the docker image is useful at the moment. The secret key is no longer provisioned, so the entrypoint will alwyas fail to run unless the right volume is mounted and the parameters updated to include `--private_key`.

tiziano88: I think it's useful to have the binary path baked in the image, it's one less thing to have to keep in sync between the image and this file, right? Then we only deal with the flags here (we never change the actual binary that we run in an image anyways)

conradgrobler: I don't have a strong preference, so I added it back in.

examples/aggregator/gcp/pod.yaml
```diff
@@ -10,11 +10,26 @@ spec:
       image: gcr.io/oak-ci/oak-aggregator:latest
       ports:
         - containerPort: 8080
+      volumeMounts:
+        - name: tls-secret
+          mountPath: '/etc/oak-secrets'
+          readOnly: true
+      args:
+        [
+          '--application=config.bin',
```
ipetr0v, line 19: Configuration file may not always be called `config.bin`
E.x.:
https://github.com/project-oak/oak/blob/3619cab6406ed292960a1e5371d0d5b11a378909/examples/hello_world/config/BUILD#L26-L49

examples/aggregator/gcp/pod.yaml
```diff
@@ -10,11 +10,26 @@ spec:
...
+          '--ca_cert=ca.pem',
+          '--cert_chain=docker.pem',
```
ipetr0v, line 21: We can store all TLS related files in the `/etc/oak-secrets/` (just like `.ssh`), so they will not be scattered in different places

examples/aggregator/scripts/docker_push
```diff
@@ -9,8 +9,8 @@ readonly AGGREGATOR_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-aggregator'
 readonly BACKEND_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-aggregator-backend'
 
 declare -a image backend_image
-image="$(find ./bazel-cache -name aggregator_image.tar)"
-backend_image="$(find ./bazel-cache -name aggregator_backend_image.tar)"
+image="$(find -L ./bazel-bin -name aggregator_image.tar)"
```
ipetr0v, line 12: This search doesn't work outside the Docker container.
And it's currently not possible to push Docker images from within the container (you need to set up GCP).

examples/aggregator/gcp/pod.yaml
conradgrobler: I agree, but this `pod.yaml` file is for this specific example where it will be called that. With the decision to move the arguments to `pod.taml` this must be specified in a file that is currently not autogenerated. Is there a bazel build rule to build pod files like these?

oak/server/loader/BUILD
```diff
@@ -66,7 +66,6 @@ container_image(
     files = [
         "oak_runner",
         "//examples/certs/docker:ca.pem",
-        "//examples/certs/docker:docker.key",
```
rbehjati, line 69: Is it possible to remove the `.pem` files too? It does not seem quite right to include example certificates in the runner image.

examples/aggregator/gcp/pod.yaml
conradgrobler: I don't think it is possible to mount a secret to an existing directory that contains the other files. It would be possible to provision all certificate files (even the non-secret ones) using k8s secrets as @rbehjati suggested.

examples/aggregator/scripts/docker_push
conradgrobler: I will revert it. The original search did not work on my machine for some reason.

oak/server/loader/BUILD
conradgrobler: For consistency I will also provision them as secrets, even though they are not considered secret.

examples/aggregator/gcp/pod.yaml
ipetr0v: No, currently there is only a rule to create Docker images

examples/aggregator/scripts/docker_push
ipetr0v: It will only work if you run this search after you build an example.
Because consecutive builds invalidate build targets in the cache IIUC

conradgrobler: It did not work on my machine even after running `build_example` with `-i`. Perhaps I am running a different version of Bazel or my Bazel is configured differently.

Issue #819, opened 2020-04-07
Implement HTTPS client node


conradgrobler:
This should be similar to the gRPC client node.

Some requirements (although not all of these have to be there as part of the initial implementation):

- Trusted root CA certificates should be hard-coded in the runtime to validate TLS connections
- Each non-public instance of the node should only be able to connect to a single domain, and should validate the URL before connecting
- The secrecy label for the node should be based on the target domain

PR #833, opened 2020-04-09
Use bytes to represent gRPC message body

tiziano88:
This will also be useful for #819

### DIFF
--- a/oak/server/module_invocation.cc
+++ b/oak/server/module_invocation.cc
@@ -96,9 +96,7 @@ void ModuleInvocation::ProcessRequest(bool ok) {
   // Build an encapsulation of the gRPC request invocation and put it in a Message.
   oak::encap::GrpcRequest grpc_request;
   grpc_request.set_method_name(context_.method());
-  google::protobuf::Any* any = new google::protobuf::Any();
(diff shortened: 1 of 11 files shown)


oak/server/module_invocation.cc
```diff
@@ -223,7 +221,7 @@ void ModuleInvocation::BlockingSendResponse() {
     auto callback = new std::function<void(bool)>(
         std::bind(&ModuleInvocation::SendResponse, this, std::placeholders::_1));
     stream_.Write(bb, options, callback);
-  } else if (!grpc_response.has_rsp_msg()) {
+  } else if (grpc_response.rsp_msg().empty()) {
```
tiziano88, line 224: This is causing tests to fail now, I am guessing because an empty bytes field is indistinguishable from a default message. @daviddrysdale do we need to keep this distinction between last message with and without payload?

tiziano88: I fixed it by assuming that only exactly one between `rsp_msg` and `status` is set. @ipetr0v is this compatible with the Rust implementation too? If so, should we make that a oneof, and / or explain in a comment in the proto message?

PR #838, opened 2020-04-13
Add TLS handling to the gRPC pseudo-node

### LINKED ISSUE #806: TLS for gRPC pseudo-node
After finishing https://github.com/project-oak/oak/issues/751 we need to add a way to configure TLS for the gRPC pseudo-node

Based on the following comment https://github.com/project-oak/oak/issues/751#issuecomment-603838068

~~Blocked by https://github.com/project-oak/oak/issues/890~~

### LINKED ISSUE #813: Reduce the number of `clone` calls in the gRPC pseudo-node
If I understand correctly - Hyper creates a new instance of a server for each user request.
More than that, it can pass a server generator to other threads that may work with individual user connections.
So we need a `make_service_fn` that creates an asynchronous server generator used by different threads, and a `service_fn` function that is used by a thread to generate a server instance to `serve` each user request (`serve` is also `async`).

In order to deal with the Rust borrow checker we need to `clone` the `server` 3 times:
https://github.com/project-oak/oak/blob/1d0e43667cdf1b2e47390913c0d3d11dcc8e4b32/oak/server/rust/oak_runtime/src/node/grpc_server.rs#L314-L323

Based on the [Slack thread](https://project-oak.slack.com/archives/CHE9E13C3/p1586016767103600).

ipetr0v:
This change:
- Adds TLS handling logic to the gRPC pseudo-Node
- Reimplements gRPC processing using Tonic library

Fixes #806
Fixes #813

~~Blocked by https://github.com/project-oak/oak/pull/891~~

# Checklist
      construction.

### DIFF
--- a/oak/server/rust/oak_loader/src/main.rs
+++ b/oak/server/rust/oak_loader/src/main.rs
@@ -18,12 +18,10 @@
 //!
 //! To invoke, run the following command from the root of the repository:
 //!
-//! ```
 //! cargo run --package=oak_loader -- \
 //!     --application=<APP_CONFIG_PATH> \
 //!     --grpc-tls-private-key=<PRIVATE_KEY_PATH> \
 //!     --grpc-tls-certificate=<CERTIFICATE_PATH>
-//! ```
 
 use log::info;
 use oak_runtime::{configure_and_run, proto::oak::application::ApplicationConfiguration};
--- a/oak/server/rust/oak_loader/src/main.rs
+++ b/oak/server/rust/oak_loader/src/main.rs
@@ -19,7 +19,10 @@
 //! To invoke, run the following command from the root of the repository:
 //!
 //! ```
-//! cargo run --package=oak_loader -- --application=<APP_CONFIG_PATH>
+//! cargo run --package=oak_loader -- \
+//!     --application=<APP_CONFIG_PATH> \
+//!     --grpc-tls-private-key=<PRIVATE_KEY_PATH> \
+//!     --grpc-tls-certificate=<CERTIFICATE_PATH>
 //! ```
 
 use log::info;
@@ -35,25 +38,27 @@ use std::{
 };
 use structopt::StructOpt;
 
+use oak_runtime::proto::oak::application::node_configuration::ConfigType::GrpcServerConfig;
+
 #[derive(StructOpt, Clone)]
 #[structopt(about = "Oak Loader")]
 pub struct Opt {
     #[structopt(long, help = "Application configuration file")]
     application: String,
-    // TODO(#806): Make arguments non-optional once TLS support is enabled.
-    #[structopt(long, help = "Path to the PEM-encoded CA root certificate")]
-    ca_cert: Option<String>,
-    #[structopt(long, help = "Path to the private key")]
-    private_key: Option<String>,
-    #[structopt(long, help = "Path to the PEM-encoded certificate chain")]
-    cert_chain: Option<String>,
+    #[structopt(long, help = "Private RSA key file used by gRPC server pseudo-Nodes")]
+    grpc_tls_private_key: String,
+    #[structopt(
+        long,
+        help = "PEM encoded X.509 TLS certificate file used by gRPC server pseudo-Nodes"
+    )]
+    grpc_tls_certificate: String,
     #[structopt(
         long,
         default_value = "3030",
-        help = "Metrics server port number. Defaults to 3030."
+        help = "Metrics server port number (the default value is 3030)"
     )]
     metrics_port: u16,
-    #[structopt(long, help = "Starts the Runtime without a metrics server.")]
+    #[structopt(long, help = "Starts the Runtime without a metrics server")]
     no_metrics: bool,
     #[structopt(
         long,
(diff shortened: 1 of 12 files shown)


oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -146,66 +129,230 @@ impl GrpcServerNode {
...
+    type Item = Vec<u8>;
+    type Error = tonic::Status;
+
+    fn decode(&mut self, src: &mut DecodeBuf<'_>) -> Result<Option<Self::Item>, Self::Error> {
+        use bytes::Buf;
+
+        // Decoder should empty the buffer, or the "Unexpected EOF decoding stream" will occure.
+        // https://github.com/hyperium/tonic/blob/0fa2bf1cea9d1166d49e40f2211268611b6993de/tonic/src/codec/decode.rs#L248
+        let item = Vec::from(src.bytes());
+        src.advance(src.bytes().len());
```
tiziano88, line 292: Does https://docs.rs/bytes/0.5.4/bytes/buf/ext/struct.Reader.html help here?

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -146,66 +129,230 @@ impl GrpcServerNode {
...
+
+        let future = async move {
+            debug!("Processing an HTTP/2 request: {:?}", request);
+            let mut grpc_service = Grpc::new(VecCodec::default());
+            let response = grpc_service.unary(grpc_handler, request).await;
+            debug!("Sending an HTTP/2 response: {:?}", response);
+            Ok(response)
+        };
+
+        Box::pin(future)
```
tiziano88, line 238: I am not entirely sure why all this needs to happen, vs just getting the bytes out of the `http::Request`. Are you sure we need all these extra abstractions that ultimately don't do anything (apart from converting between different formats for bytes)?

conradgrobler: I agree. I think it should be possible to just use `hyper::body::aggregate` to get the bytes from the request body, rather than using a codec and a grpc service.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -28,18 +31,17 @@ use prost::Message;
...
     application: String,
-    // TODO(#806): Make arguments non-optional once TLS support is enabled.
-    #[structopt(long, help = "Path to the PEM-encoded CA root certificate")]
-    ca_cert: Option<String>,
-    #[structopt(long, help = "Path to the private key")]
-    private_key: Option<String>,
-    #[structopt(long, help = "Path to the PEM-encoded certificate chain")]
-    cert_chain: Option<String>,
+    #[structopt(long, help = "Private key file")]
+    private_key: String,
```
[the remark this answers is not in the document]
ipetr0v: Done
Added `Private RSA key file used by gRPC pseudo-Nodes` and `PEM encoded X.509 TLS certificate file used by gRPC pseudo-Nodes`

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -146,66 +129,230 @@ impl GrpcServerNode {
...
-                warn!("Couldn't aggregate request body: {}", error);
-                error
-            })?;
+    /// Main node worker thread.
+    fn worker_thread(mut self) {
+        let pretty_name = pretty_name_for_thread(&thread::current());
+
+        // Receive a `channel_writer` handle used to pass handles for temporary channels.
+        info!("{}: Waiting for a channel writer", pretty_name);
+        self.init_channel_writer()
```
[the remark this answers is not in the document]
ipetr0v: Yes, it's only used in the `HttpRequestHandler` and `GrpcRequestHandler` (that is created by the former).
I just was storing it there for convenience and to use it as a reference point for comments.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
tiziano88: Fine for now, but note that we will probably end up with a HTTP server pseudo-node, so this will end up being basically that, and that logic will probably be pushed into the SDK.

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -148,66 +131,227 @@ impl GrpcServerNode {
...
+            debug!("Sending an HTTP/2 response: {:?}", response);
+            Ok(response)
+        };
+
+        Box::pin(future)
+    }
+}
+
+/// Encapsulates [`VecEncoder`] and [`VecDecoder`] types and is used by [`tonic::server::Grpc`].
+#[derive(Clone, Debug, Default)]
```
tiziano88, line 246: Does this really need to be `Clone`, `Debug` and `Default`? Are these needed by `tonic` itself? (and below)

oak/server/rust/oak_runtime/src/node/grpc_server.rs
```diff
@@ -146,66 +129,230 @@ impl GrpcServerNode {
...
+                })?;
+
+            let response = handler
+                // Handle a gRPC request and send it into the Runtime.
+                .handle_grpc_request(grpc_request)
+                // Read a gRPC response from the Runtime.
+                .and_then(|response_reader| handler.handle_grpc_response(response_reader))
+                // Convert an error to a gRPC error status without sending clients descriptions for
+                // internal errors.
+                .map_err(|_| tonic::Status::new(tonic::Code::Internal, ""))?;
```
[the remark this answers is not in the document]
tiziano88: Could you add a comment here then?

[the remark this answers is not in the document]
ipetr0v: Done

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -28,18 +31,17 @@ use prost::Message;
...
     #[structopt(long, help = "Application configuration file")]
     application: String,
-    // TODO(#806): Make arguments non-optional once TLS support is enabled.
-    #[structopt(long, help = "Path to the PEM-encoded CA root certificate")]
-    ca_cert: Option<String>,
-    #[structopt(long, help = "Path to the private key")]
-    private_key: Option<String>,
-    #[structopt(long, help = "Path to the PEM-encoded certificate chain")]
-    cert_chain: Option<String>,
+    #[structopt(long, help = "Private RSA key file used by a gRPC pseudo-Node")]
```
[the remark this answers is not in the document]
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/grpc_server.rs
ipetr0v: Deleted, only `Default` is left

PR #847, opened 2020-04-15
Simplify node thread creation and management

tiziano88:
Each node now just returns a JoinHandle to a main thread, and the main
thread is encapsulated in a separate method, in which the node itself is
moved, instead of having to clone all the fields from the parent node.

# Checklist

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/external.rs
+++ b/oak/server/rust/oak_runtime/src/node/external.rs
@@ -19,7 +19,7 @@ use crate::{
     NodeId,
 };
 use lazy_static::lazy_static;
-use log::{error, info};
+use log::info;
 use oak_abi::OakStatus;
 use std::{sync::RwLock, thread, thread::JoinHandle};
 
@@ -40,7 +40,6 @@ pub struct PseudoNode {
     config_name: String,
     runtime: RuntimeProxy,
     reader: Handle,
-    thread_handle: Option<JoinHandle<()>>,
 }
 
 impl PseudoNode {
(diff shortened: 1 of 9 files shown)


tiziano88:
@rbehjati friendly ping

oak/server/rust/oak_runtime/src/node/external.rs
```diff
@@ -49,13 +48,12 @@ impl PseudoNode {
...
             runtime,
             reader,
-            thread_handle: None,
         }
     }
 }
 
 impl super::Node for PseudoNode {
-    fn start(&mut self) -> Result<(), OakStatus> {
+    fn start(self: Box<Self>) -> Result<JoinHandle<()>, OakStatus> {
```
rbehjati, line 56: Is it possible to refactor the closure inside the `spawn` here to a `worker_thread` like the rest of the PR?

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -948,53 +979,10 @@ impl super::Node for WasmNode {
             self.config_name, self.entrypoint
         );
 
-        // Clone or copy all the captured values and move them into the closure, for simplicity.
```
rbehjati, line 951: This is code from before... but is it possible to refactor the block of code above this line into a method, possibly called `enumerate_exports`, to make this `start` method read more similarly to the other `start` methods?

oak/server/rust/oak_runtime/src/node/external.rs
tiziano88: Done.

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
tiziano88: Good call, I clarified the logic and added more logging since I was at it.

PR #860, opened 2020-04-17
Add script to generate reproducibility index file

tiziano88:
Also add a cloud build step to build the Rust oak_loader binary.

Unfortunately, the artifacts are not actually reproducibly buildable
right now, so we cannot enforce this yet, but I'm checking in the
generated file that I get on my machine, which seems to differ between
my machine and GCP.

When we figure out the source of the discrepancy, then we can enforce
that the hashes are actually checked in when changes are made to the
source.

Ref #861

### DIFF
--- a/reproducibility_index
+++ b/reproducibility_index
@@ -1,0 +1,10 @@
+61fdf25b4e2b4171a579dd9e4716b24348fb3f9f90ad936f9c91046618c93ed5  ./target/wasm32-unknown-unknown/release/abitest_0_frontend.wasm
+32a5f6aa9cbac2f71f40be10c52da32653a5b579810e5927af42df171a0f9dfc  ./target/wasm32-unknown-unknown/release/abitest_1_backend.wasm
+c697982360ea5d5a95b795889b4633ed27d3bc892f9bc9ac1df480c3759cb9b5  ./target/wasm32-unknown-unknown/release/aggregator.wasm
+2f5b176dbe8c9269d6bc2253c82d627d556db8a7214de8133ebfa185075a08dc  ./target/wasm32-unknown-unknown/release/chat.wasm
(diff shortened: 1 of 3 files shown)


reproducibility_index
```diff
@@ -0,0 +1,10 @@
+61fdf25b4e2b4171a579dd9e4716b24348fb3f9f90ad936f9c91046618c93ed5  ./target/wasm32-unknown-unknown/release/abitest_0_frontend.wasm
```
ipetr0v, line 1: Should it be in a separate `artifact` directory?

scripts/build_reproducibility_index
```diff
@@ -0,0 +1,25 @@
...
+
+readonly SCRIPTS_DIR="$(dirname "$(readlink -f "$0")")"
+# shellcheck source=scripts/common
+source "$SCRIPTS_DIR/common"
+
+# List of artifacts that are expected to be reproducibly built.
+readonly REPRODUCIBLE_ARTIFACTS=(
+  ./target/wasm32-unknown-unknown/release/*.wasm
+  ./bazel-clang-bin/oak/server/loader/oak_runner
+  ./target/debug/oak_loader
```
ipetr0v, line 17: Maybe it's better to use `release` than `debug` ?

reproducibility_index
tiziano88: - we don't have an `artifact` folder (yet)
- if we had one, its contents would not be checked in, but this file should still be checked in

so I would leave this here for now, but happy to restructure things more generally at some point

scripts/build_reproducibility_index
tiziano88: AFAICT that's not what is built / used by the other scripts?

https://github.com/project-oak/oak/blob/55a93d7b7f3a23687db42d152d51eaf69a1fc234/scripts/build_server#L50

We should probably switch that to `--release`, at which point we can change this list too.

PR #881, opened 2020-04-20
Lint Dockerfile using Hadolint

tiziano88:
See https://github.com/hadolint/hadolint

Fix various issues brought up by hadolint, plus extra ones.

- update base image to a more recent one
- use consistent naming for `ARG` (lowercase)
- use `curl` instead of `wget`
- move `&&` continuations to new line
- clear apt cache after install commands
- set PIPEFAIL option

### DIFF
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -1,0 +1,8 @@
+ignored:
+  # Pin versions in apt get install.
+  # See https://github.com/hadolint/hadolint/wiki/DL3008.
+  - DL3008
+  # Set the SHELL option -o pipefail before RUN with a pipe in.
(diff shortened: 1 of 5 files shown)


tiziano88:
cc @ipetr0v for the Android Dockerfile changes (I will make sure everything keeps working, but let me know if you have any comments)

tiziano88:
@ipetr0v actually this now fails when building the Android Docker image, do you have any idea why? https://pantheon.corp.google.com/cloud-build/builds/7e8307bd-7270-40e2-938f-b0189cf3cc99;step=15?project=oak-ci

tiziano88:
It looks like it may be the `pipefail` setting, though I have no idea why

Issue #855, opened 2020-04-16
OpenID Connect: Add support to the Oak Runtime


conradgrobler:
The Oak Runtime needs to provide some services to support authentication via OpenID Connect:

- Provide nonces to clients to prevent replay attacks
  - The nonce must be stored for future reference to be able to validate the resulting ID token
  - The nonce must have enough entropy to avoid collisions (eg 256 bits)
- Exchange ID tokens for bearer tokens
  - [Validate ID token](https://developers.google.com/identity/protocols/oauth2/openid-connect#validatinganidtoken)
    - This involves downloading current certificates from a Google server
    - The certificates change about once per day, so ideally it would be good to cache these for the `max-age` specified in the `cache-control` response header when downloading
  - Generate bearer token that expires (eg in 24 hours)
  - Store mapping between bearer token and unique user id (`sub` field in ID token)
  - Return bearer token to the client

The Oak Runtime must store some state to support the authentication process
- The list of nonces that were issued so that it could be used to later validate the resulting ID tokens
- The list of nonces that were already used to validate ID tokens to avoid replay attacks
- The mappings between unique user accounts and bearer tokens (including expiry date of the bearer tokens)
  - The bearer token will be for short term use, but it should be possible to resolve the unique identity of the user they apply to
  - It would be possible to skip storing this if the bearer token included the user unique ID and a signature, but this would make the size of the token much bigger. In that case it might be better to always use ID tokens directly instead of exchanging them for bearer tokens
- Mapping between unique users accounts and refresh tokens (if we decide to support refresh tokens)

conradgrobler:
As part of the validation of the ID token, the client ID used for generating the ID token must be known. This could be done by either:

- Assume there is a single client ID that will be used by the Oak instance, so include it in the server config; or
- Require it as a parameter for generating a nonce, and store a mapping between the nonce and the client ID

conradgrobler:
One issue that needs to be clarified is how the list of nonces will be persisted to prevent replay attacks.
- The nonce must be stored for at least as long as the expiry time of the identity token
- If multiple Oak instances accept tokens from the same client ID, there must be some mechanism to propage this between them, or they should use shared external state storage (eg for load balancing etc)

tiziano88:
How about a slightly different design:

[...]

PR #888, opened 2020-04-21
Add initial OpenID Connect auth client example

conradgrobler:
This is the first part of the authentication client example. It opens the browser to allow the user to authenticate, and then receives the resulting authentication code. The example will be extended to exhange the auth code for an identity token once the authentication service is availble to support it and then pass the token to the runtime as proof of identity (#855).

# Checklist

      construction.

### DIFF
--- a/examples/authentication/README.md
+++ b/examples/authentication/README.md
@@ -1,0 +1,24 @@
+# OpenID Connect Authentication Example
+
+This example authenticates using the
+[Google Identity Platform](https://developers.google.com/identity/), which
+implements OpenID Connect.
+
+The client application opens a browser which allows the user to authenticate. On
+successful authentication, the browser is redirected to connect to a web server
+running on localhost. The authentication code is extracted from the query string
+of the redirection URL and returned to the main thread.
+
+TODO(#855): The authenctication code can then be exchanged for an identity token
+by the authentication server.
+
+The client can be executed using:
+
+```bash
+
+cargo run -b auth_client -- --client-id $YOUR_CLIENT_ID
+
+```
+
+This requires a valid OAuth Client ID, which can be created as described in
+https://developers.google.com/identity/protocols/oauth2/openid-connect#getcredentials
--- a/examples/authentication/README.md
+++ b/examples/authentication/README.md
@@ -9,14 +9,14 @@ successful authentication, the browser is redirected to connect to a web server
 running on localhost. The authentication code is extracted from the query string
 of the redirection URL and returned to the main thread.
 
-TODO(#855): The authenctication code can then be exchanged for an identity token
+TODO(#855): The authentication code can then be exchanged for an identity token
 by the authentication server.
 
 The client can be executed using:
 
 ```bash
 
-cargo run -b auth_client -- --client-id $YOUR_CLIENT_ID
+cargo run --package auth_client -- --client-id $YOUR_CLIENT_ID
 
 ```
 
--- a/examples/authentication/client/src/main.rs
+++ b/examples/authentication/client/src/main.rs
@@ -19,16 +19,18 @@
 //! This example uses the Google Identity Platform.
 //! https://developers.google.com/identity/
 
-use futures::channel::oneshot::{self, Sender};
 use futures_util::future;
 use hyper::{service::Service, Body, Request, Response, Server, StatusCode};
 use log::{info, warn};
 use std::{
     collections::HashMap,
-    sync::{Arc, Mutex},
     task::{Context, Poll},
 };
 use structopt::StructOpt;
+use tokio::sync::{
+    mpsc::{self, Sender},
+    oneshot,
+};
 use url::{form_urlencoded, Url};
 
 #[derive(StructOpt, Clone)]
@@ -59,14 +61,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let redirect_address = opt.redirect_address.parse()?;
     info!("Listening for redirect on {}", &redirect_address);
 
-    let (sender, receiver) = oneshot::channel();
-    let result_sender = Arc::new(Mutex::new(ResultSender {
-        sender: Some(sender),
-        result: None,
-    }));
-    let maker = MakeHandler {
-        result_sender: result_sender.clone(),
-    };
+    let (shutdown_sender, shutdown_receiver) = oneshot::channel();
+    let (result_sender, mut result_receiver) = mpsc::channel(1);
+    let maker = MakeHandler { result_sender };
 
     let mut runner = tokio::runtime::Runtime::new().unwrap();
     let task = runner.spawn(async move {
@@ -74,9 +71,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             .serve(maker)
             // Use oneshot channel as signal for shutdown.
             .with_graceful_shutdown(async move {
-                receiver.await.unwrap();
+                shutdown_receiver.await.ok();
             })
             .await
+            .ok();
     });
 
     let request_uri = get_authentication_request_url(opt);
@@ -85,11 +83,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Open the URL in the system-configured default browser.
     open::that(request_uri)?;
 
-    // Wait until notified that the redirect has been processed.
-    runner.block_on(task)??;
+    // Wait for result sent by redirect handler.
+    let result = runner.block_on(result_receiver.recv());
+    result_receiver.close();
 
-    let result = result_sender.lock().unwrap();
-    let code = result.result.as_ref().unwrap().as_ref().ok().unwrap();
+    // Notify server to shutdown.
+    shutdown_sender.send(()).ok();
+
+    // Wait until graceful shutdown is completed..
+    runner.block_on(task)?;
+
+    let code = result.unwrap().unwrap();
     info!("Received code: {}", code);
     Ok(())
 }
@@ -114,23 +118,6 @@ fn get_authentication_request_url(opt: Opt) -> String {
     auth_endpoint.to_string()
 }
 
-/// Provides storage for the result of proccessing the redirect and providing a notification to the
-/// main thread that it has been processed.
-///
-/// The oneshot Sender can only be used once.
-struct ResultSender {
-    sender: Option<Sender<()>>,
-    result: Option<Result<String, String>>,
-}
-
-impl ResultSender {
-    fn notify(&mut self) {
-        if self.sender.is_some() {
-            self.sender.take().unwrap().send(()).unwrap();
-        }
-    }
-}
-
 /// Handles the redirects to extract code from the query string.
 ///
 /// A new instance is created for every incoming request. The redirect URL contains the result of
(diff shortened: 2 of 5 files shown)


examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+    open::that(request_uri)?;
+    runner.block_on(task)??;
+
+    let result = result_sender.lock().unwrap();
+    let code = result.result.as_ref().unwrap().as_ref().ok().unwrap();
+    info!("Received code: {}", code);
+    Ok(())
+}
+
+fn get_request(opt: Opt) -> String {
```
tiziano88, line 92: Rename to something more specific and / or add some documentation.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Ok(()).into()
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        if let Some(query) = request.uri().query() {
+            let lookup: HashMap<_, _> = form_urlencoded::parse(query.as_bytes()).collect();
+            let result_sender = &mut self.result_sender.lock().unwrap();
+            if lookup.contains_key("code") {
```
tiziano88, line 139: Does this work?

```suggestion
            if let Some(code) = lookup.get("code") {
```

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+        .query_pairs_mut()
+        .append_pair("scope", scope)
+        .append_pair("response_type", "code")
+        .append_pair("redirect_uri", &redirect_uri)
+        .append_pair("client_id", &opt.client_id);
+    auth_endpoint.to_string()
+}
+
+struct ResultSender {
+    sender: Option<Sender<()>>,
```
tiziano88, line 109: Does `Sender<Result<String, String>>` not work here, instead of the two separate fields? In fact if that works, we don't really need the struct at all.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+                Response::builder()
+                    .status(StatusCode::BAD_REQUEST)
+                    .body(Body::from("No query string!"))
+                    .unwrap(),
+            )
+        }
+    }
+}
+
+// Produces instances of the redirect handler service.
```
tiziano88, line 175: Please add some more documentation, either here or on `RedirectHandler`, the code seems fine but it takes a while to figure out what's talking to what, so perhaps a high level description may help.

examples/authentication/README.md
```diff
@@ -0,0 +1,24 @@
...
+This example authenticates using the
+[Google Identity Platform](https://developers.google.com/identity/), which
+implements OpenID Connect.
+
+The client application opens a browser which allows the user to authenticate. On
+successful authentication, the browser is redirected to connect to a web server
+running on localhost. The authentication code is extracted from the query string
+of the redirection URL and returned to the main thread.
+
+TODO(#855): The authenctication code can then be exchanged for an identity token
```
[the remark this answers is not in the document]
conradgrobler: Done.

examples/authentication/README.md
```diff
@@ -0,0 +1,24 @@
...
+by the authentication server.
+
+The client can be executed using:
+
+```bash
+
+cargo run -b auth_client -- --client-id $YOUR_CLIENT_ID
```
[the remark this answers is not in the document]
conradgrobler: It is the same as `--bin`. I changed it to `--package` for consistency with other examples.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+    let task = runner.spawn(async move {
+        Server::bind(&redirect_address)
+            .serve(maker)
+            .with_graceful_shutdown(async move {
+                receiver.await.unwrap();
+            })
+            .await
+    });
+    open::that(request_uri)?;
+    runner.block_on(task)??;
```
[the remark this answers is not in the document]
conradgrobler: Yes, `block_on` returns a `Result` that contains another `Result`.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+    let mut runner = tokio::runtime::Runtime::new().unwrap();
+    let task = runner.spawn(async move {
+        Server::bind(&redirect_address)
+            .serve(maker)
+            .with_graceful_shutdown(async move {
+                receiver.await.unwrap();
+            })
+            .await
+    });
+    open::that(request_uri)?;
```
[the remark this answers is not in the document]
conradgrobler: Yes. Added comment.

examples/authentication/client/src/main.rs
conradgrobler: I started that way, but it had its own complications. The `Receiver` gets the result within the `async` section inside `with_graceful_shutdown`. This code runs in a different thread, so another cross-thread communication mechanism is needed to get the results from the `Receiver` back to the main thread after the server is shut down. It could in principle be done with two oneshot channels as well, but I am not sure that is any clearer.

Also, oneshot `Sender`s and `Receiver`s cannot be copied, so it would need something like Arc<Mutex<Option<Result<String, String>>>>.

examples/authentication/client/src/main.rs
conradgrobler: Done.

examples/authentication/client/src/main.rs
conradgrobler: Done.

examples/authentication/client/src/main.rs
conradgrobler: Done.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    let result = result_sender.lock().unwrap();
+    let code = result.result.as_ref().unwrap().as_ref().ok().unwrap();
+    info!("Received code: {}", code);
+    Ok(())
+}
+
+fn get_request(opt: Opt) -> String {
+    // TODO(#886): Retrieve endpoint from server.
+    let mut auth_endpoint = Url::parse("https://accounts.google.com/o/oauth2/v2/auth").unwrap();
```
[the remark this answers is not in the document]
conradgrobler: Done.

examples/authentication/client/src/main.rs
tiziano88: Ah I see, `Sender` and `Receiver` work with futures, not with threads, so probably not a great fit here. How about https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html instead? That should be nicer to use across threads.

conradgrobler: Using `mpsc` solves one problem, in that it removes the need for `Arc` and `Mutex`, but there is still a need for a second channel. There are two signals needed. One to give the results to the main thread (which uses `result` on the struct) and one to notify Hyper to shut down the web server (which uses `sender` on the struct).

conradgrobler: I have now refactored the code to use `mspc` to notify `main` of the result of the redirect handling and `oneshot` from `main` to notify Hyper to shut down the server. This simplifies the flow a bit.

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+            .serve(maker)
+            // Use oneshot channel as signal for shutdown.
+            .with_graceful_shutdown(async move {
+                shutdown_receiver.await.ok();
+            })
+            .await
+            .ok();
+    });
+
+    let request_uri = get_authentication_request_url(opt);
```
tiziano88, line 80: (nit) consistent `uri` vs `url`

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+
+/// Produces instances of the redirect handler service.
+///
+/// Hyper uses it to create a new handler for every incoming request. The handler needs a copy of
+/// the reference to the ResultSender to communicate the results back to the main thread.
+struct MakeHandler {
+    result_sender: Sender<Result<String, String>>,
+}
+
+impl<T> Service<T> for MakeHandler {
```
tiziano88, line 189: I find the naming still a bit confusing; is it correct to read `MakeHandler` as handling connections, and `RedirectHandler` as handling requests? If so, maybe it would be more readable to actually replace `T` with `Connection` (or whatever concrete type) (if it works).

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+        .append_pair("redirect_uri", &redirect_uri)
+        .append_pair("client_id", &opt.client_id);
+    auth_endpoint.to_string()
+}
+
+/// Handles the redirects to extract code from the query string.
+///
+/// A new instance is created for every incoming request. The redirect URL contains the result of
+/// the authentication as query string parameters. If the the authentication is successful the
+/// authentication code is be passed in the `code` paramter. If it failed the reason is passed in
```
tiziano88, line 125:
```suggestion
/// authentication code is passed in the `code` parameter. If it failed the reason is passed in
```

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+/// the `error` paramter.
```
tiziano88, line 126:
```suggestion
/// the `error` parameter.
```

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+                warn!("Error: {:?}", error);
+                self.result_sender.try_send(Err(error)).unwrap();
+                future::ok(
+                    Response::builder()
+                        .status(StatusCode::UNAUTHORIZED)
+                        .body(Body::from("Authentication failed!"))
+                        .unwrap(),
+                )
+            } else {
+                future::ok(
```
tiziano88, line 163: Shouldn't this also send some error back? And also the else below?

examples/authentication/client/src/main.rs
```diff
@@ -0,0 +1,203 @@
...
+    info!("URI: {}", &request_uri);
+    // Open the URL in the system-configured default browser.
+    open::that(request_uri)?;
+
+    // Wait for result sent by redirect handler.
+    let result = runner.block_on(result_receiver.recv());
+    result_receiver.close();
+
+    // Notify server to shutdown.
+    shutdown_sender.send(()).ok();
```
tiziano88, line 91: What does the `ok` actually do?

examples/authentication/client/src/main.rs
conradgrobler: MakeHandler does not really handle connections, even though a I think connection is passed in to `call` but ignored. It is similar a producer pattern where it is a service responsible for creating one instance of the request handler for each incoming request. I used the naming pattern of beginning the name with `Make` as this seems to be what all the hyper examples do. I have now changed the name to RedirectHandlerProducer as it is clearer about the intention of the service. I don't think it is possible to make it a concrete type, as it could be basically anything that supports `AsyncRead`, `AsyncWrite`. In the case of TLS, it is the TLS stream rather than the original TCP connection that is used. I also think it differs depending on whether the client is using HTTP v1 or v2.

examples/authentication/client/src/main.rs
conradgrobler: Sorry, missed this one. Now done.

tiziano88:
@conradgrobler I assume you have already created an OAuth client for testing, could you add the details to the README? In particular, I think the client_id can be added directly in the snippet for now, and we can keep the client secret somewhere safe internally (perhaps as part of the oak-ci project on GCP).

Issue #861, opened 2020-04-17
Ensure that root-of-trust binary artifacts can be reproducibly built


tiziano88:
By "root-of-trust binary artifacts" I mean things like the executable oak_loader binary (C++ / Rust), and any Oak Module that may be used for declassification (e.g. the aggregator module), identified by a label with the corresponding hash.

PR #897, opened 2020-04-23
Fix posting reproducibility index to PR

tiziano88:
The current version does not work because when running on merge events
the PR number is not present, so we need to look it up based on commit
SHA.

Ref #861


tiziano88:
:crossed_fingers:

PR #906, opened 2020-04-27
Add initial authentication server implementation

conradgrobler:
This initial version of the auth server only provides the client ID and OAuth endpoint to the client. Subsequent versions will expand the parameters provided and implement the logic the exchange the authentication code for an identity token.

# Checklist

      construction.

### DIFF
--- a/examples/authentication/client/src/main.rs
+++ b/examples/authentication/client/src/main.rs
@@ -20,7 +20,7 @@
 //! https://developers.google.com/identity/
 
 use hyper::Server;
-use log::info;
+use log::{info, warn};
 use structopt::StructOpt;
 use tokio::sync::{mpsc, oneshot};
 
@@ -105,8 +105,10 @@ async fn get_authentication_code(
     // Notify local HTTP server to shutdown.
     shutdown_sender.send(()).unwrap();
 
-    // Wait until graceful shutdown is completed, ignoring shutdown errors.
-    task.await.ok();
+    // Wait until graceful shutdown is completed, logging shutdown errors.
+    if let Err(error) = task.await {
+        warn!("Error while shutting down server: {}", error);
+    }
     let code = result.unwrap().unwrap();
     Ok(code)
 }
--- a/examples/authentication/client/src/main.rs
+++ b/examples/authentication/client/src/main.rs
@@ -19,19 +19,13 @@
 //! This example uses the Google Identity Platform.
 //! https://developers.google.com/identity/
 
-use futures_util::future;
-use hyper::{service::Service, Body, Request, Response, Server, StatusCode};
-use log::{info, warn};
-use std::{
-    collections::HashMap,
-    task::{Context, Poll},
-};
+use hyper::Server;
+use log::info;
 use structopt::StructOpt;
-use tokio::sync::{
-    mpsc::{self, Sender},
-    oneshot,
-};
-use url::{form_urlencoded, Url};
+use tokio::sync::{mpsc, oneshot};
+
+mod auth_client;
+mod redirect_handler;
 
 #[derive(StructOpt, Clone)]
 #[structopt(about = "OpenID Connect Client Example.")]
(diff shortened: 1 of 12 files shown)


examples/authentication/client/src/main.rs
```diff
@@ -77,127 +93,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    // Notify server to shutdown.
+    // Notify local HTTP server to shutdown.
     shutdown_sender.send(()).unwrap();
 
-    // Wait until graceful shutdown is completed..
-    runner.block_on(task)?;
-
+    // Wait until graceful shutdown is completed, ignoring shutdown errors.
+    task.await.ok();
```
tiziano88, line 109: If the intent of `ok()` is to ignore the return value, I think `let _ = task.await;` would be more explicit. Or you may as well match on the `Err` case and log it :)

examples/authentication/proto/authentication.proto
```diff
@@ -0,0 +1,32 @@
...
+// Parameters needed by clients to initiate OpenID Connect authentication requests.
+message AuthParameters {
+  string client_id = 1;
+  string auth_endpoint = 2;
+  // TODO(#886): Add support for scope, state and code challenge.
+}
+
+service Authentication {
+  rpc GetAuthParameters(google.protobuf.Empty) returns (AuthParameters);
+}
```
tiziano88, line 32: I don't immediately see why we need a gRPC server for this, as opposed to just an endpoint that returns a single string which is the URL to open in the browser? Is there value in preserving all this structure?

conradgrobler: The `redirect_uri` querystring parameter in the resulting URL is dependent on the client, which is why I decided to let the client construct the final URL. It could be handled by passing the `redirect_uri` as a paramter to the auth server, or by including a token that the client replaces. I don't think either of these would make the code much clearer though.

Also, the next step is to exchange the code for a token, and validating the token. This can once again be done with URL parameters and a get request that returns the token, but I don't think manually handling URL parameters and base64 (or JSON) encoding and decoding of the token bytes will make the code clearer or simpler.

examples/authentication/client/src/main.rs
```diff
@@ -19,54 +19,70 @@
...
+        &opt.ca_cert,
+        &opt.auth_server,
+        &opt.redirect_address,
+    ))?;
+    let code = runner.block_on(get_authentication_code(&opt.redirect_address, &request_url))?;
+    info!("Received code: {}", code);
+    Ok(())
+}
 
+/// Gets the OIDC authentication code by opening the default browser and extracting it from the
```
ipetr0v, line 73: Does the browser open in the headless mode?
Also, why do we need a browser instead a single HTTP request (does it require user interaction)?

examples/authentication/client/src/main.rs
```diff
@@ -19,54 +19,70 @@
...
     let opt = Opt::from_args();
-    let redirect_address = opt.redirect_address.parse()?;
-    info!("Listening for redirect on {}", &redirect_address);
+    let mut runner = tokio::runtime::Runtime::new().unwrap();
+    let request_url = runner.block_on(auth_client::get_authentication_request_url(
+        &opt.ca_cert,
+        &opt.auth_server,
+        &opt.redirect_address,
+    ))?;
+    let code = runner.block_on(get_authentication_code(&opt.redirect_address, &request_url))?;
```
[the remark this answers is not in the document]
conradgrobler: It evolved from a simpler version that did not use it. Changed it to use it now.

examples/authentication/client/src/main.rs
conradgrobler: It opens a normal browser window. It requires user interaction: the user has to log in to their Google account or select which one to use if they are already logged in.

examples/authentication/client/src/redirect_handler.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+}
+
+/// Produces instances of the redirect handler service.
+///
+/// Hyper uses it to create a new handler for every incoming request. The handler needs a copy of
+/// the reference to the ResultSender to communicate the results back to the main thread.
+pub struct RedirectHandlerProducer {
+    result_sender: Sender<Result<String, String>>,
```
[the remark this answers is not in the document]
conradgrobler: Added a new `AuthError` struct to make it clear what the error part of ther result represents.

examples/authentication/server/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[tonic::async_trait]
+impl Authentication for AuthenticationHandler {
+    async fn get_auth_parameters(
+        &self,
+        _req: Request<()>,
+    ) -> Result<Response<AuthParameters>, Status> {
+        Ok(Response::new(AuthParameters {
+            client_id: self.client_id.clone(),
+            auth_endpoint: "https://accounts.google.com/o/oauth2/v2/auth".to_string(),
```
tiziano88, line 53: Add a comment to show where this comes from?

conradgrobler: Done.

conradgrobler:
@googlebot ping

PR #920, opened 2020-04-30
Do not log termination as error

tiziano88:
It is confusing to see error log messages for normal working conditions.

### DIFF
--- a/sdk/rust/oak/src/lib.rs
+++ b/sdk/rust/oak/src/lib.rs
@@ -15,7 +15,7 @@
 //
 
 use byteorder::{ReadBytesExt, WriteBytesExt};
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use serde::{Deserialize, Serialize};
 
 // Re-export ABI constants that are also visible as part of the SDK API.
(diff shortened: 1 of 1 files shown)


sdk/rust/oak/src/lib.rs
```diff
@@ -465,15 +465,24 @@ pub fn run_event_loop<T: crate::io::Decodable, N: Node<T>>(mut node: N, in_handl
...
+                    ErrBadHandle | ErrChannelClosed => {
+                        warn!(
+                            "non-transient error waiting for command: {:?}; terminating event loop",
+                            status
+                        );
                         return;
                     }
                     _ => {
-                        info!("(possibly) transient error: continuing event loop");
+                        warn!("(possibly) transient error waiting for command: {:?}; continuing event loop", status);
```
juliettepretot, line 485: I like that now all the logs related this Error are one line (as opposed to error followed by info), makes it very clear that it's about he same thing

juliettepretot: Also being able to pattern match is nice, it's good to see for this humble JavaScript developer.

PR #929, opened 2020-05-01
Configure shell to use rustup command before using it

juliettepretot:
Super duper minor, but we one needs to to configure the shell before running `rustup`. Else you'd get "rustup: command not found".

Should make things ever so slightly easier to setup for beginners.


tiziano88:
Thanks! Feel free to merge when the checks have completed. We usually use the "squash and merge" strategy.

PR #930, opened 2020-05-01
Compile Docker server with clang

ipetr0v:
This change adds parameters for choosing server type in Docker images.
Change is necessary, since an Oak server for Docker images was not compiled using `clang`.

# Checklist

### DIFF
--- a/scripts/build_example
+++ b/scripts/build_example
@@ -6,17 +6,19 @@ source "${SCRIPTS_DIR}/common"
 
 language="rust"
 compilation_mode='fastbuild'
-docker_image=false
-while getopts "l:ide:h" opt; do
+docker_config=''
+while getopts "l:i:de:h" opt; do
   case "${opt}" in
     h)
-      echo -e "Usage: ${0} [-h] [-l rust|cpp] [-i] [-d] -e EXAMPLE
+      echo -e "Usage: ${0} [-h] [-l rust|cpp] [-i base|logless] [-d] -e EXAMPLE
 
 Build the given example Oak Application and client.
 
 Options:
   -e    Example application name (required)
-  -i    Package application in a Docker image
+  -i    Package application in a Docker image with one of the following server types:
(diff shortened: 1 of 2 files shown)


scripts/build_example
```diff
@@ -6,17 +6,19 @@ source "${SCRIPTS_DIR}/common"
...
     h)
-      echo -e "Usage: ${0} [-h] [-l rust|cpp] [-i] [-d] -e EXAMPLE
+      echo -e "Usage: ${0} [-h] [-l rust|cpp] [-i base|logless] [-d] -e EXAMPLE
 
 Build the given example Oak Application and client.
 
 Options:
   -e    Example application name (required)
-  -i    Package application in a Docker image
+  -i    Package application in a Docker image with one of the following server types:
```
tiziano88, line 19: I guess if this is not specified, it does not package it as a Docker image at all? Perhaps clarify somewhere (basically this flag enables docker packaging, *and* specifies the version)

scripts/build_example
```diff
@@ -79,16 +89,18 @@ case "${language}" in
...
 
-if [[ "${docker_image}" == true ]]; then
+if [[ -n "${docker_config}" ]]; then
   if [[ "${language}" != "rust" ]]; then
     echo "Docker images are only supported for Rust modules"
     exit 1
   fi
 
-  bazel build "${bazel_build_flags[@]}" "//examples/${EXAMPLE}/server:${EXAMPLE}_image.tar"
+  readonly DOCKER_BAZEL_BUILD_FLAGS=("${bazel_build_flags[@]}" "--config=${docker_config}")
```
tiziano88, line 98: I think is more confusing than helpful, perhaps just do this below

```
bazel build "${bazel_build_flags[@]}" "--config=${docker_config}" "//examples/${EXAMPLE}/server:${EXAMPLE}_image.tar"
```

up to you really, anyways perhaps add a comment

PR #935, opened 2020-05-01
Add build examples and build server to runner

tiziano88:
It's not quite on par with the other scripts there, but getting there

Add `./scripts/runner` to make it easier to invoke.

# Checklist

      construction.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -23,7 +23,10 @@
 //! cargo run --package=runner
 //! ```
 
-use std::{io::Read, path::PathBuf};
+use std::{
+    io::Read,
+    path::{Path, PathBuf},
+};
 use structopt::StructOpt;
 
 mod internal;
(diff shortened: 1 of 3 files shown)


runner/src/main.rs
```diff
@@ -45,15 +181,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    if statuses.contains(&StatusResultValue::Error) {
-        std::process::exit(1);
     }
+}
 
-    Ok(())
+fn examples() -> impl Iterator<Item = PathBuf> {
+    let walker = walkdir::WalkDir::new("./examples").into_iter();
+    walker
+        .filter_entry(|e| !is_ignored_entry(e))
```
ipetr0v, line 190: Does it also ignore `./examples/certs` ?
https://github.com/tiziano88/oak/blob/a8554dbebeeac6bb9e660e0bc20d548c854ccdb9/runner/src/main.rs#L197-L206

runner/src/main.rs
```diff
@@ -23,16 +23,152 @@
...
+                &[
+                    "build",
+                    "--release",
+                    "--target=x86_64-unknown-linux-musl",
+                    "--package=oak_loader",
+                ],
+            ),
+        },
+        v => {
+            let config = match v {
```
ipetr0v, line 142: It doesn't yet support the `rust` server, right?

runner/src/main.rs
```diff
@@ -23,16 +23,152 @@
...
+fn bazel_build(extra_build_flags: &[&str], targets: &[&str]) -> Box<dyn Runnable> {
+    let mut flags = Vec::new();
+    flags.push("build");
+
+    let b = bazel_build_flags();
+    flags.extend(as_ref(&b));
+
+    flags.extend(extra_build_flags);
+    flags.push("--");
+    flags.extend(targets);
```
ipetr0v, line 123: Do we need to also support excluded targets?

runner/src/main.rs
```diff
@@ -23,16 +23,152 @@
...
+}
+
+fn build_examples(opt: &BuildExamples) -> Step {
+    match opt.variant.as_str() {
+        "rust" => Step::Multiple {
+            name: "root".to_string(),
+            steps: examples()
+                .filter(is_cargo_toml_file)
+                .filter(|entry| {
+                    (*entry != PathBuf::from("./examples/aggregator/backend/Cargo.toml"))
```
ipetr0v, line 61: Could you please add a comment, why these targets are excluded?

runner/src/main.rs
tiziano88: Not yet (soon!)

runner/src/main.rs
tiziano88: No need for now (also Bazel will be even less relevant going forward)

runner/src/main.rs
tiziano88: Done.

runner/src/main.rs
tiziano88: No, but this is used with extra filtering (only including Cargo.toml files), so in the end that does not affect it.

PR #941, opened 2020-05-04
Manage prometheus registry in Oak Runtime

### LINKED ISSUE #899: Improve metrics

rbehjati:
- Instantiates and manages prometheus registry in the Runtime,
  instead of using a static instance
- Uses labels for finer grained collection of metrics
- Better instrumentation for gRPC server pseudo-node
- Exposes metrics on `/metrics` endpoint

Fixes #899 

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/metrics/mod.rs
+++ b/oak/server/rust/oak_runtime/src/metrics/mod.rs
@@ -15,53 +15,126 @@
 //
 
 use prometheus::{
-    opts, register_histogram, register_int_counter, register_int_gauge, Histogram, IntCounter,
-    IntGauge,
+    proto::MetricFamily, HistogramOpts, HistogramVec, IntCounterVec, IntGauge, Opts, Registry,
 };
 
 pub mod server;
 
-/// Struct that collects all the metrics in one place
-pub struct Metrics {
-    pub grpc_request_duration: Histogram,
-    pub grpc_requests_total: IntCounter,
-    pub grpc_response_size: Histogram,
-    pub runtime_nodes_count: IntGauge,
+/// Helper struct with functions for creating and registering metrics.
+struct MetricsBuilder {
+    pub registry: Registry,
 }
 
-// TODO(#899): For testability implement a trait with methods for updating the metrics.
-// TODO(#899): Instead of using a global Registry, the Runtime should instantiate and manage the
-// Registry
-impl Metrics {
-    fn new() -> Self {
+/// Mostly copied from https://github.com/grpc-ecosystem/java-grpc-prometheus
+#[derive(Clone)]
+pub struct GrpcServerMetrics {
+    /// Total number of RPCs started on the server.
+    pub grpc_server_started_total: IntCounterVec,
+    /// Total number of RPCs completed on the server, regardless of success or failure.
+    pub grpc_server_handled_total: IntCounterVec,
+    /// Histogram of response latency of RPCs handled by the server, in seconds.
+    pub grpc_server_handled_latency_seconds: HistogramVec,
+    /// Histogram of response sizes of RPCs handled by the server.
+    pub grpc_response_size_bytes: HistogramVec,
+}
+
+/// Struct that collects all metrics for monitoring the Oak Runtime.
+#[derive(Clone)]
+pub struct RuntimeMetrics {
+    pub runtime_nodes_total: IntGauge,
+}
+
+/// Struct that collects all the metrics in one place
+#[derive(Clone)]
+pub struct Metrics {
+    registry: Registry,
+    pub runtime_metrics: RuntimeMetrics,
+    pub grpc_server_metrics: GrpcServerMetrics,
+}
+
+impl MetricsBuilder {
+    pub fn new() -> Self {
         Self {
-            grpc_request_duration: register_histogram!(
-                "grpc_request_duration_seconds",
-                "The gRPC request latencies in seconds."
-            )
-            .expect("Creating grpc_request_duration_seconds metric failed."),
+            registry: Registry::new(),
+        }
+    }
 
-            grpc_requests_total: register_int_counter!(opts!(
-                "grpc_requests_total",
-                "Total number of gRPC requests received."
-            ))
-            .expect("Creating grpc_requests_total metric failed."),
+    fn register<T: 'static + prometheus::core::Collector + Clone>(&self, metric: T) -> T {
+        self.registry.register(Box::new(metric.clone())).unwrap();
+        metric
+    }
+}
 
-            grpc_response_size: register_histogram!(
+fn counter_vec(metric_name: &str, labels: &[&str], help: &str) -> IntCounterVec {
+    let opts = Opts::new(metric_name, help);
+    IntCounterVec::new(opts, labels).unwrap()
+}
+
+fn histogram_vec(metric_name: &str, label: &[&str], help: &str) -> HistogramVec {
+    let opts = HistogramOpts::new(metric_name, help);
+    HistogramVec::new(opts, label).unwrap()
+}
+
+fn int_gauge(metric_name: &str, help: &str) -> IntGauge {
+    let opts = Opts::new(metric_name, help);
+    IntGauge::with_opts(opts).unwrap()
+}
+
+impl GrpcServerMetrics {
+    fn new(builder: &MetricsBuilder) -> Self {
+        GrpcServerMetrics {
+            grpc_server_started_total: builder.register(counter_vec(
+                "grpc_server_started_total",
+                &["method_name"],
+                "Total number of RPCs started on the server.",
+            )),
+            grpc_server_handled_total: builder.register(counter_vec(
+                "grpc_server_handled_total",
+                &["method_name", "status_code"],
+                "Total number of RPCs completed on the server, regardless of success or failure.",
+            )),
+            grpc_server_handled_latency_seconds: builder.register(histogram_vec(
+                "grpc_server_handled_latency_seconds",
+                &["method_name"],
+                "Histogram of response latency of RPCs handled by the server.",
+            )),
+            grpc_response_size_bytes: builder.register(histogram_vec(
                 "grpc_response_size_bytes",
-                "The gRPC response sizes in bytes."
-            )
-            .expect("Creating grpc_response_size_bytes metric failed."),
-
-            runtime_nodes_count: register_int_gauge!(opts!(
-                "runtime_nodes_count",
-                "Number of nodes in the runtime."
-            ))
-            .expect("Creating runtime_nodes_count metric failed."),
+                &["method_name"],
+                "Histogram of response sizes of RPCs handled by the server.",
+            )),
         }
     }
 }
 
-lazy_static::lazy_static! {
-    pub static ref METRICS: Metrics = Metrics::new();
+impl RuntimeMetrics {
+    fn new(builder: &MetricsBuilder) -> Self {
+        RuntimeMetrics {
+            runtime_nodes_total: builder.register(int_gauge(
+                "runtime_nodes_total",
+                "Number of nodes in the runtime.",
+            )),
+        }
+    }
+}
+
+impl Metrics {
+    pub fn new() -> Self {
+        let builder = MetricsBuilder::new();
+        Self {
+            runtime_metrics: RuntimeMetrics::new(&builder),
+            grpc_server_metrics: GrpcServerMetrics::new(&builder),
+            registry: builder.registry,
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


rbehjati:
@tiziano88 I am using a HashMap to allow dynamically adding metrics, instead of having a huge object with all the metrics pre-defined. So now each node can register the metrics it needs. 
If there are multiple instances of each node, each instance can have its own metrics; as I have done for `grpc_server`. Although I am not sure if different `grpc_server` nodes would have different `config_name`s.  Is it possible to let the nodes know their `NodeId`s?

What do you think about adding metrics dynamically like this?

tiziano88:
Thanks! I had a quick look, but this looks too dynamic to me, it is essentially re-implementing a façade on top of the underlying registry, forwarding all the types and methods. I would suggest to stick to a more static approach in which all the metrics are defined in one place, and can be accessed via struct fields, following the usual Rust ownership and borrowing rules. Similarly to how we prefer defining flags all in place, rather than scattering them around in the various modules. WDYTY? Did you have a specific issue in mind that made you prefer the approach in this PR?

rbehjati:
The main reason is to allow different instances of the same type of node to have separate metrics. With a static approach, we won't be able to distinguish between metrics from two different grpc nodes, for example. Additionally, with a dynamic approach you won't get stats for metrics that you don't use (with a static approach you'll get that for all the struct fields, event though they'll be all zeros). 
Also, later on, to allow application-level metrics, we'll need a way to allow registering metrics dynamically. I don't think static metrics would work for that use case. 

Scattering does not bother me that much, since the access to the metrics is scattered anyways.

tiziano88:
I think distinguishing between node instances could be done by using labels. If each instance uses a different metric name, how do we even know what metrics to look for when monitoring the binary?

rbehjati:
I am not sure I understand your question, but the names are not random, they'd have names similar to what you get with labels. 

I'll see if I can make it work with labels, but I am still wondering how we'll support application-level metrics with pre-defined metrics.

tiziano88:
The names would not be random, but they would also not be known upfront, right? If they depend on the name of the nodes that are part of the application, it would not be easy to create, say, a dashboard that shows incoming requests for gRPC server nodes. Or am I misunderstanding how metric names work?

tiziano88:
And yes you are right that application-level metrics would probably need some special treatment, my comment applies to runtime-level metrics.

oak/server/rust/oak_runtime/src/metrics/mod.rs
```diff
@@ -15,53 +15,142 @@
...
-                "The gRPC response sizes in bytes."
-            )
-            .expect("Creating grpc_response_size_bytes metric failed."),
-
-            runtime_nodes_count: register_int_gauge!(opts!(
+            registry: Registry::new(),
+        }
+    }
+
+    fn register<T: 'static + prometheus::core::Collector + Clone>(&self, metric: T) -> T {
```
ipetr0v, line 67: Why it should be `static`?

oak/server/rust/oak_runtime/src/metrics/mod.rs
```diff
@@ -15,53 +15,142 @@
...
+#[derive(Clone)]
+pub struct GrpcServerMetrics {
+    /// Total number of RPCs started on the server.
+    pub grpc_server_started_total: IntCounterVec,
+    /// Total number of RPCs completed on the server, regardless of success or failure.
+    pub grpc_server_handled_total: IntCounterVec,
+    /// Histogram of response latency of RPCs handled by the server, in seconds.
+    pub grpc_server_handled_latency_seconds: HistogramVec,
+    /// Histogram of response sizes of RPCs handled by the server.
+    pub grpc_response_size: HistogramVec,
```
tiziano88, line 38: `grpc_response_size_bytes` (see https://prometheus.io/docs/practices/naming/#metric-names)

oak/server/rust/oak_runtime/src/metrics/mod.rs
```diff
@@ -15,53 +15,142 @@
...
+    /// Total number of stream messages received from the client.
+    pub grpc_server_msg_received_total: IntCounterVec,
+    /// Total number of stream messages sent by the server.
+    pub grpc_server_msg_sent_total: IntCounterVec,
+}
+
+/// Struct that collects all metrics for monitoring the Oak Runtime.
+#[derive(Clone)]
+pub struct RuntimeMetrics {
+    pub runtime_nodes_count: IntGaugeVec,
```
tiziano88, line 49: `runtime_nodes_total` for consistency with the others? or `..._count_total`?

oak/server/rust/oak_runtime/src/metrics/mod.rs
```diff
@@ -15,53 +15,142 @@
...
+
+    fn register<T: 'static + prometheus::core::Collector + Clone>(&self, metric: T) -> T {
+        self.registry.register(Box::new(metric.clone())).unwrap();
+        metric
+    }
+
+    fn counter(&self, metric_name: &str, labels: &[&str], help: &str) -> IntCounterVec {
+        let opts = Opts::new(metric_name, help);
+        IntCounterVec::new(opts, labels).unwrap()
+    }
```
tiziano88, line 75: AFAICT these methods don't do anything with `self`, in fact I am not sure whether they need to be defined in this `impl` block in the first place, is this an intentional decision?

oak/server/rust/oak_runtime/src/metrics/server.rs
```diff
@@ -42,9 +43,11 @@ impl std::fmt::Display for MetricsServerError {
 
 impl std::error::Error for MetricsServerError {}
 
-async fn serve_metrics(_req: Request<Body>) -> Result<Response<Body>, MetricsServerError> {
+async fn handle_metrics_request(
+    runtime: Arc<Runtime>,
```
tiziano88, line 47: Why does this need to be an `Arc`? I thought it should be enough to have a regular reference. Or does that complicate the lifetime of this method?

oak/server/rust/oak_runtime/src/metrics/server.rs
```diff
@@ -53,25 +56,43 @@ async fn serve_metrics(_req: Request<Body>) -> Result<Response<Body>, MetricsSer
...
         })
 }
 
-async fn make_server(port: u16) -> Result<(), hyper::error::Error> {
+async fn serve_metrics(
+    runtime: Arc<Runtime>,
+    req: Request<Body>,
+) -> Result<Response<Body>, MetricsServerError> {
+    match (req.method(), req.uri().path()) {
+        (&Method::GET, "/metrics") => handle_metrics_request(runtime).await,
```
tiziano88, line 72: Nice pattern for matching on request!

oak/server/rust/oak_runtime/src/metrics/server.rs
```diff
@@ -53,25 +56,43 @@ async fn serve_metrics(_req: Request<Body>) -> Result<Response<Body>, MetricsSer
...
+
+async fn make_server(runtime: Arc<Runtime>, port: u16) -> Result<(), hyper::error::Error> {
     let addr = SocketAddr::from(([0, 0, 0, 0], port));
 
     // A `Service` is needed for every connection, so this
     // creates one from the `process_metrics` function.
-    let make_svc = make_service_fn(|_conn| async {
-        // service_fn converts our function into a `Service`
-        Ok::<_, hyper::Error>(service_fn(serve_metrics))
+    let make_service = make_service_fn(move |_| {
```
tiziano88, line 85: I think it's useful to keep the parameter name as `_conn`, for documentation purposes.

oak/server/rust/oak_runtime/src/node/grpc/server.rs
```diff
@@ -182,11 +181,19 @@ impl Service<http::Request<hyper::Body>> for HttpRequestHandler {
...
+        let metrics_data = self.runtime.metrics_data();
         let future = async move {
             debug!("Processing an HTTP/2 request: {:?}", request);
             let mut grpc_service = Grpc::new(VecCodec::default());
             let response = grpc_service.unary(grpc_handler, request).await;
             debug!("Sending an HTTP/2 response: {:?}", response);
+            let stc = format!("{}", response.status());
+            metrics_data
+                .grpc_server_metrics
+                .grpc_server_handled_total
```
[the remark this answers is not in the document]
tiziano88: Nice! Though I think the status code was not defined at metric creation, so does it actually work?

oak/server/rust/oak_runtime/src/metrics/server.rs
```diff
@@ -53,25 +54,47 @@ async fn serve_metrics(_req: Request<Body>) -> Result<Response<Body>, MetricsSer
...
     // creates one from the `process_metrics` function.
-    let make_svc = make_service_fn(|_conn| async {
-        // service_fn converts our function into a `Service`
-        Ok::<_, hyper::Error>(service_fn(serve_metrics))
+    let make_service = make_service_fn(move |_conn| {
+        let runtime = runtime.clone();
+        async move {
+            Ok::<_, hyper::Error>(service_fn(move |req| {
+                let runtime = runtime.clone();
+                async move { serve_metrics(runtime, req).await }
```
tiziano88, line 92: I would have thought that `async { foo.await }` is equivalent to just `foo`, or is this closure needed because of the explicit move?

oak/server/rust/oak_runtime/src/metrics/mod.rs
```diff
@@ -15,53 +15,127 @@
...
+        }
+    }
+}
+
+impl RuntimeMetrics {
+    fn new(builder: &MetricsBuilder) -> Self {
+        RuntimeMetrics {
+            runtime_nodes_total: builder.register(int_gauge(
+                "runtime_nodes_total",
+                &["instance"],
```
tiziano88, line 115: What is this label for? AFAICT it's always set to `runtime`?

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -1101,12 +1111,16 @@ impl Runtime {
...
-        );
+        self.metrics_data
+            .runtime_metrics
+            .runtime_nodes_total
+            .with_label_values(&["runtime"])
+            .set(
+                self.node_infos
+                    .read()
+                    .expect("could not acquire lock on node_infos")
+                    .len() as i64,
```
tiziano88, line 1122: Maybe refactor this in a separate method, to make it clear that it does not hold any locks.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -34,6 +33,9 @@ use std::{
     thread::JoinHandle,
 };
```
tiziano88, line 35: Remove blank line.

oak/server/rust/oak_runtime/src/metrics/mod.rs
rbehjati: I wanted to keep everything as vector. With vector metrics, label is needed. I have now changed the type to `IntGauge` instead of `IntGaugeVec`.

oak/server/rust/oak_runtime/src/metrics/server.rs
rbehjati: Yes. Thanks. Fixed. Must have been leftover from my chain of changes to make Rust happy.

oak/server/rust/oak_runtime/src/runtime/mod.rs
rbehjati: Done.

oak/server/rust/oak_runtime/src/runtime/mod.rs
rbehjati: Thanks. Done!

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -16,14 +16,15 @@
 
 use crate::{
     message::{Message, NodeMessage},
-    metrics::METRICS,
+    metrics::*,
```
tiziano88, line 19: I think we try to avoid wildcard imports unless necessary.

PR #949, opened 2020-05-05
Use `RUST_LOG` in `simple_logger`

ipetr0v:
This change makes `simple_logger` use `RUST_LOG` environment variable to set up log level.

# Checklist
      construction.


tiziano88:
Deferring to @daviddrysdale who set that up, not sure if there is any reason behind the original choice.

PR #950, opened 2020-05-05
Move codec to a separate directory

ipetr0v:
This change moves a gRPC encoder/decoder to a separate directory.

# Checklist
      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/grpc/server.rs
+++ b/oak/server/rust/oak_runtime/src/node/grpc/server.rs
@@ -24,7 +24,6 @@ use std::{
     thread::{self, JoinHandle},
 };
 use tonic::{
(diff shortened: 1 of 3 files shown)


oak/server/rust/oak_runtime/src/node/grpc/server.rs
```diff
@@ -34,7 +33,13 @@ use oak_abi::{
     grpc::encap_request, label::Label, proto::oak::encap::GrpcRequest, ChannelReadStatus, OakStatus,
 };
 
-use crate::{metrics::METRICS, node::Node, pretty_name_for_thread, runtime::RuntimeProxy, Handle};
+use crate::{
```
tiziano88, line 36: Move this up with the other `use` lines

ipetr0v: Done

PR #958, opened 2020-05-06
Run both C++ and Rust Loader

### LINKED ISSUE #874: Provide gRPC read handle to the Oak entrypoint

ipetr0v:
This change updates macro and scripts in order to be able to run C++ and Rust Oak Loaders on the same examples.

Fixes #874

# Checklist

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -88,10 +88,16 @@ steps:
     timeout: 60m
     entrypoint: 'bash'
     args: ['./scripts/run_examples', '-s', 'base']
+  - name: 'gcr.io/oak-ci/oak:latest'
+    id: run_examples_rust
+    waitFor: ['run_examples']
+    timeout: 60m
+    entrypoint: 'bash'
+    args: ['./scripts/run_examples', '-s', 'rust']
   # TODO(942): Reenable `run_examples` with `asan` and `tsan`.
   # - name: 'gcr.io/oak-ci/oak:latest'
   #   id: run_examples_asan
-  #   waitFor: ['run_examples']
+  #   waitFor: ['run_examples_rust']
   #   timeout: 60m
   #   entrypoint: 'bash'
   #   args: ['./scripts/run_examples', '-s', 'asan']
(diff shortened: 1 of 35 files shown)


tiziano88:
FYI, as discussed in the on-demand, I think we should go ahead and switch directly from running the examples from C++ to Rust, without trying to artificially support both. I think when gRPC client and server nodes are supported, we can have a PR that switches running the examples with the Rust runtime, together with any SDK changes that are necessary. The storage and roughtime examples can be disabled as part of that, we'll figure out what to do about them in a separate PR (either delete them, or re-enable them when the corresponding functionality is available in the Rust loader).

ipetr0v:
I thought to add this commit in order to make later changes (such as gRPC client and streaming support) be easier to implement - since they all will be tested in the CI and they will no become a single giant PR.

PR #984, opened 2020-05-13
Support downgrade of tags

tiziano88:
Ref #630

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/runtime/mod.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/mod.rs
@@ -23,11 +23,14 @@ use crate::{
 use core::sync::atomic::{AtomicBool, AtomicU64, Ordering::SeqCst};
 use itertools::Itertools;
 use log::{debug, error, info, trace, warn};
-use oak_abi::{label::Label, ChannelReadStatus, OakStatus};
+use oak_abi::{
+    label::{Label, Tag},
+    ChannelReadStatus, OakStatus,
+};
 use prometheus::proto::MetricFamily;
 use rand::RngCore;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt::Write,
     string::String,
     sync::{Arc, Mutex, RwLock},
@@ -103,6 +106,9 @@ struct NodeInfo {
     /// See https://github.com/project-oak/oak/blob/master/docs/concepts.md#labels
     label: Label,
 
+    /// The downgrading privilege of this Node.
+    privilege: NodePrivilege,
+
     /// Map of ABI handles to channels.
     abi_handles: HashMap<oak_abi::Handle, ChannelHalf>,
 
@@ -110,6 +116,18 @@ struct NodeInfo {
     /// small exception, when the Runtime is in the process of closing down and
     /// the [`NodeStopper`] is held by the shutdown processing code).
     node_stopper: Option<NodeStopper>,
+}
+
+/// The downgrading (declassification + endorsement) privilege associated with a Node instance.
+///
+/// See https://github.com/project-oak/oak/blob/master/docs/concepts.md#downgrades
+#[derive(Debug, Default, Clone)]
+pub struct NodePrivilege {
+    /// Tags that may be declassified (removed from the secrecy component of a label) by the Node.
+    can_declassify_secrecy_tags: HashSet<Tag>,
+
+    /// Tags that may be endorsed (added to the integrity component of a label) by the Node.
+    can_endorse_integrity_tags: HashSet<Tag>,
 }
 
 impl std::fmt::Debug for NodeInfo {
@@ -341,6 +359,7 @@ impl RuntimeProxy {
             proxy.node_id,
             "implicit.initial",
             &Label::public_trusted(),
+            &NodePrivilege::default(),
         );
         proxy
     }
@@ -725,6 +744,42 @@ impl Runtime {
         node_info.label.clone()
     }
 
+    /// Returns the least restrictive (i.e. least secret, most trusted) label that this Node may
+    /// downgrade to. This takes into account all the [downgrade privilege](NodeInfo::privilege)
+    /// that the node possesses.
+    fn get_node_downgraded_label(&self, node_id: NodeId) -> Label {
+        // Original (static) Node label.
+        let node_label = self.get_node_label(node_id);
+        // Retrieve the set of tags that the node may downgrade.
+        let node_privilege = self.get_node_privilege(node_id);
+        Label {
+            // Remove all the secrecy tags that the Node may declassify.
+            secrecy_tags: node_label
+                .secrecy_tags
+                .iter()
+                .filter(|t| !node_privilege.can_declassify_secrecy_tags.contains(t))
+                .cloned()
+                .collect(),
+            // Add all the integrity tags that the Node may endorse.
+            integrity_tags: node_label
+                .integrity_tags
+                .iter()
+                .chain(node_privilege.can_endorse_integrity_tags.iter())
+                .cloned()
+                .collect(),
+        }
+    }
+
+    /// Returns a clone of the [`NodePrivilege`] of the provided Node.
+    fn get_node_privilege(&self, node_id: NodeId) -> NodePrivilege {
+        let node_infos = self
+            .node_infos
+            .read()
+            .expect("could not acquire lock on node_infos");
+        let node_info = node_infos.get(&node_id).expect("invalid node_id");
+        node_info.privilege.clone()
+    }
+
     /// Returns a clone of the [`Label`] associated with the provided reader `channel_half`.
     ///
     /// Returns an error if `channel_half` is not a valid read half.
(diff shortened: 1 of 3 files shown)


oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -679,6 +691,41 @@ impl Runtime {
...
+                .collect(),
+            // Add (endorse) integrity tags that the Node can downgrade.
+            integrity_tags: node_label
+                .integrity_tags
+                .iter()
+                .chain(node_can_downgrade_tags.iter())
+                .cloned()
+                .collect(),
+        }
+    }
```
tiziano88, line 718: @aferr do you think this logic (and associated comments) makes sense? The idea is that each node has a static (immutable) label, and a set of capabilities to downgrade specific tags (for now assume they are "dual" capabilities, as Flume calls them, i.e. they can be both endorsed and declassified). Here I am computing the "most public" label that the node can declassify to, and use that to perform the actual "flows to" comparison when reading from and writing to channels. Is this correct?

aferr: Overall this does seem correct and like a fine way to do the downgrade check. The comments are also consistent with what the code does. Running through roughly what you said, if you wanted to have a top-level API like
downgrading_channel_send(self, node_id, channel_handle, ..., down_label)

roughly this would work like:
get_node_downgraded_label(self, nod_id) flowsTo down_label. /* to check that this downgrade is allowed */
down_label flows_to get_channel_label(channel_handle) /* to check that the downgrade is sufficient to do the send */

Another way to use this that I think is a bit less safe:
You could also just relax the check that gets done on the normal channel_send by checking that get_node_downgraded_label(self) flows_to get_ch_label(handle)
but I think this is less safe because then programmers will do downgrades when they don't realize they are doing it. I think it's better to make downgrades explicit because they are exceptions to isolation.

You could make this more general (And I recommend that you do!):
The idea is that if t is in node_can_downgrade_tags, that means it can either declassify away t from confidentiality or add t to integrity? This seems fine, but you might get better generality if you have separate capabilities for declassifying/endorsing (so there is node_can_declassify_tags and node_can_endorse_tags). It seems useful to be able to control declassifying/endorsing separately and basically free to allow this. The Flume paper actually does have separate capabilities for "adding labels" and "removing labels", which is a bit confusing because for confidentiality adding is upgrading and removing is downgrading. So in Flume, t- is the capability for declassifying and for reducing integrity (which is not endorsing!) and then t+ is endorsing/conf_upgrading (whacky!). The flume dual label is the joint capability t+ and t- which doesn't allow you to control declassifying/endorsing separately.

I think a slightly better way to do this than what Flume does is to have a tag t_dec for declassifying t and t_end for endorsing t. Then upgrades are always allowed for either type of label.
You would then also need:
declassifying_channel_send(...
endorsing_channel_recv(...

As long as "upgrades" of data being moved are always allowed, I actually see no need to have a a send that endorses or a recv that declassifies. If we did need to declassify on recvs and endorse on sends, the API gets more difficult because we really would need a "downgrading" channel call which does both. BUT AFAICT it looks like we can cleanly separate declassification/endorsement privilege if you like this idea.

tiziano88: Thanks for the detailed comments, everything you said made sense to me! Let me reply to some of your points individually.

## explicit downgrading send / receive

I thought about this, but since the Node can already inspect the label of the channel it is about to send things over, I think this can be emulated by an inspection plus a comparison to the current node label, and based on that deciding whether to do the send or not. Perhaps we can package it up as a single function in the Rust SDK, but I think it does not require to add a different version of send / receive to the ABI. But happy to revisit at some later point, let's keep discussing this offline.

## separate declassify / endorse capabilities

I actually started with this approach of them being separate (following the Flume model), but I realized that for most (all?) practical cases I would always grant both of them anyways, so it seems to just add a new field which would always be the same as another one. Do you have any use case in which it is necessary to keep them separate?

## separate declassifying and endorsing tags

Not sure I completely follow your point here, are you saying to split the principals so that we always talk about more specific ones in the context of either declassification or endorsement? If so, I guess this is similar to always using explicit integrity and confidentiality projections in Flam, rather then the underlying joint principal. Is this correct?

aferr: ### explicit downgrading send / receive
Your response makes sense. Loosely the choice is defaulting towards throwing an error if there is a chance of information release vs. more conveniently assuming the node wants to use all privileges it has.

I like the idea of implementing it one way in the ABI and emulating other variants in the SDK. I think separating the downgrading case from the non-downgrading case in the ABI might also make a formalism about the ABI easier, but we are maybe too far away from that direction to make this a priority. .

I won't keep pushing for one answer or the other, though.

### separate declassify / endorse capabilities
Given that confidentiality / integrity are separate labels in the first place, it seems awkward to not control downgrades for them separately as well. If you want to collapse the ability to downgrade to one object, why not also collapse labels to one object? (You can also map a product lattice onto a single lattice). 

I think the ability to separate conf/integ downgrading privilege is easy enough to add that you might as well just include it even without a concrete use-case, yet. I think folks will be imaginative enough to find a use for this. 

I think the second-best option would be to collapse labels to one object and have downgrading capability as the only other object. The midpoint where there are separate conf/integ labels, but only one downgrading object seems awkward to me. Maybe it's possible to translate from the (conf, integ, decl, endo) model to the (label, downgrade) model.

### separate declassifying and endorsing tags
I was trying to suggest we improve over the Flume model in the following way.

Flume has a capability called t+ and another called t-.
t+ is two things: the ability to declassify secrecy, and the ability to "upgrade" integrity. (Where upgrading integrity means going from trusted to untrusted. This is the direction that is always allowed).
t- is dual to this.

I think this design unnecessarily constrains upgrades which should just be allowed by default. To be more concrete, I think flows from [conf: {}, ... ] -> [conf: {alice}, ... ] should just always be allowed.

I think it is better to have:
t_decl: which just controls declassification (but not integrity upgrades which are now always allowed)
t_endo: controls endorsement (but not secrecy upgrades which are now always allowed)

tiziano88: (focusing on **separate declassifying and endorsing tags** only)

> Flume has a capability called t+ and another called t-.
t+ is two things: the ability to declassify secrecy, and the ability to "upgrade" integrity. (Where upgrading integrity means going from trusted to untrusted. This is the direction that is always allowed).
t- is dual to this.

I thought t- is the ability to declassify secrecy, and t+ is the ability to endorse integrity? Or I clearly misunderstood something :)

BTW I think the current design in this PR already allows for movements up the lattice, because it uses the "flows to" relation between labels, so, say, [conf: {}, ... ] -> [conf: {alice}, ... ] is indeed allowed. Here is the implementation of the "flows to" relation between labels: 

https://github.com/project-oak/oak/blob/6312906a7f711b4ba4508a9dcd85bbadd8c78b6f/oak/server/rust/oak_abi/src/label/mod.rs#L57-L71

aferr:
> I thought t- is the ability to declassify secrecy, and t+ is the ability to endorse integrity? Or I clearly misunderstood something :)

You are right, I said it backwards by accident. Sorry.

> BTW I think the current design in this PR already allows for movements up the lattice

Flume does use capabilities to control dynamic label raises, which is really what I was getting at. I realize Oak does not have dynamic label raises, but maybe you were going to use the upgrading part of the capability for channel_create. I guess not, so never mind.

tiziano88:
> Flume does use capabilities to control dynamic label raises, which is really what I was getting at. I realize Oak does not have dynamic label raises, but maybe you were going to use the upgrading part of the capability for channel_create. I guess not, so never mind.

So actually `channel_create` and `node_create` also allow upgrading label:

https://github.com/project-oak/oak/blob/6312906a7f711b4ba4508a9dcd85bbadd8c78b6f/oak/server/rust/oak_runtime/src/runtime/mod.rs#L811-L817

but this also comes from the "flows to" relationship, and as part of this PR I am actually changing it to take into account the downgraded label rather than the "static" label. Since the downgraded label can trivially be upgraded again, that keeps working, plus nodes now get the ability to create other nodes and channels with less secret or more endorsed labels (according to their capability). But please do double check over the code if you have time, it would be good to have a second pair of eyes on it. Thanks again!

aferr:
>  ... allow upgrading label... but this also comes from the "flows to" relationship

OK then you are already doing the thing I was hoping for. Sounds good.

> and as part of this PR I am actually changing it to take into account the downgraded label rather than the "static" label

Yes, I think you need that. Makes sense.

Please do keep in mind the option to emulate explicitly separated downgrading and non-grading versions of channel_create and node_create in the API as well.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -75,6 +81,18 @@ struct NodeInfo {
...
+/// See https://github.com/project-oak/oak/blob/master/docs/concepts.md#downgrades
+#[derive(Debug, Default, Clone)]
+pub struct NodePrivilege {
+    /// Tags that may be declassified (removed from the secrecy component of a label) by the Node.
+    can_declassify_secrecy_tags: HashSet<Tag>,
+
+    /// Tags that may be endorsed (added to the integrity component of a label) by the Node.
+    can_endorse_integrity_tags: HashSet<Tag>,
+}
+
```
tiziano88, line 95: @aferr I'm encapsulating the capabilities (of both kinds) into this struct for now, is "privilege" a correct / good enough term? I am trying to avoid "capability" because it may have other implications in the context of operating systems and handles to channels etc.

FWIW I am still hoping eventually to unify this with the hierarchy of principals a la Flam, but that's future work :)

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -75,6 +81,18 @@ struct NodeInfo {
     join_handle: Option<JoinHandle<()>>,
 }
 
+/// The downgrading (declassification + endorsement) privilege associated with a Node instance.
+///
+/// See https://github.com/project-oak/oak/blob/master/docs/concepts.md#downgrades
+#[derive(Debug, Default, Clone)]
+pub struct NodePrivilege {
+    /// Tags that may be declassified (removed from the secrecy component of a label) by the Node.
+    can_declassify_secrecy_tags: HashSet<Tag>,
```
aferr, line 90: :)

oak/server/rust/oak_runtime/src/runtime/mod.rs
aferr: Yeah looks good I like the way you did it.

Not using the word capability does make sense and privilege is a good alternative!

We might want something easier than FLAM to explain an understand but has the benefits of FLAM.One tricky part about making a "simpler FLAM" might be that (I think) having two separate objects for "integrity" and "declassifying privilege" is easier to explain than when they are the same thing, but suspect (but am not certain) separating them will have implications w.r.t. what theorems we can prove. Also something to consider as part of that future work.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -688,6 +707,41 @@ impl Runtime {
         node_info.label.clone()
     }
 
+    /// Returns a version of the [`Label`] of the provided Node, taking into account all the
```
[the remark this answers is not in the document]
tiziano88: Thanks, reads better now!

But I'm confused by the second part; you said:

> If there is some label L such that L flowsTo get_node_downgraded_label(...), then this node could downgrade to L.

but **I think** it should be really:

"If there is some label L such that get_node_downgraded_label(...) flowsTo L, then this node could downgrade to L"

right?

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -702,62 +756,66 @@ impl Runtime {
...
+    ) -> Result<(), OakStatus> {
+        let downgraded_node_label = self.get_node_downgraded_label(node_id);
         trace!(
-            "{:?}: node_label={:?}, channel_label={:?}",
+            "{:?}: can {:?} read from {:?}?",
             node_id,
-            node_label,
-            channel_label
+            downgraded_node_label,
+            label
```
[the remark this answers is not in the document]
tiziano88: I had it there, but removed it because it increases the risk that in some future refactoring we would mix up the labels :) but I'll think about how to log more stuff in the future.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -702,62 +756,66 @@ impl Runtime {
...
+    fn validate_can_write_to_label(&self, node_id: NodeId, label: &Label) -> Result<(), OakStatus> {
+        let downgraded_node_label = self.get_node_downgraded_label(node_id);
         trace!(
-            "{:?}: node_label={:?}, channel_label={:?}",
+            "{:?}: can {:?} write to {:?}?",
             node_id,
-            node_label,
-            channel_label
+            downgraded_node_label,
+            label
```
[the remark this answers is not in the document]
tiziano88: "

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -1169,7 +1212,12 @@ impl Runtime {
             config.node_subname(entrypoint),
             new_node_id.0
         );
-        self.node_configure_instance(new_node_id, &new_node_name, label);
+        self.node_configure_instance(
+            new_node_id,
+            &new_node_name,
+            label,
+            &NodePrivilege::default(),
```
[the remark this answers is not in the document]
tiziano88: Yes, but the privilege will be granted by the runtime itself (e.g. based on the Wasm code hash), not by the creator node. Does this make sense? Let me know if you think we are not on the same page about this, because I suspect this is not really "standard" in other IFC system.

oak/server/rust/oak_runtime/src/runtime/mod.rs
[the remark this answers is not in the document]
aferr: Oops, you are right, I said it backwards again. There are unlimited opportunities to do things backwards in IFC.

oak/server/rust/oak_runtime/src/runtime/mod.rs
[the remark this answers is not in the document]
aferr: I don't think I fully follow the way you are doing it, but I think I don't need to fully follow how privilege movement works right now before you merge this.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -739,62 +794,67 @@ impl Runtime {
...
-    /// Returns whether the given Node is allowed to read from the provided channel, according to
-    /// their respective [`Label`]s.
+    /// Returns whether the given Node is allowed to read from the provided channel read half,
+    /// according to their respective [`Label`]s.
     fn validate_can_read_from_channel(
         &self,
         node_id: NodeId,
         channel_half: &ChannelHalf,
     ) -> Result<(), OakStatus> {
-        trace!(
```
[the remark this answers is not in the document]
tiziano88: it has just moved inside the `validate_can_read_from_label` method. I think that's enough, if it turns out we need more details I'll add more tracing in the future.

Issue #982, opened 2020-05-13
Refactor Node thread creation (use `tokio::sync::oneshot`)


ipetr0v:
Currently two pseudo-Nodes (gRPC server and client) are using [Tokio Runtime](https://docs.rs/tokio/0.2.20/tokio/runtime/struct.Runtime.html) in order to run `async` routines:
https://github.com/project-oak/oak/blob/11f96d62555364286881e3cc081c98d5754c5f19/oak/server/rust/oak_runtime/src/node/grpc/server.rs#L130-L138
https://github.com/project-oak/oak/blob/11f96d62555364286881e3cc081c98d5754c5f19/oak/server/rust/oak_runtime/src/node/grpc/server.rs#L145
And because these threads are blocked, they cannot `join` correctly when the Runtime is trying to stop the Node.

We might need to use `tokio::sync::oneshot`, that is already used by the `AuxServer`:
https://github.com/project-oak/oak/blob/38522c9462591b25ff1f64865f2390ab5ef6eced/oak/server/rust/oak_runtime/src/runtime/introspect.rs#L170-L176
https://github.com/project-oak/oak/blob/38522c9462591b25ff1f64865f2390ab5ef6eced/oak/server/rust/oak_runtime/src/runtime/mod.rs#L160-L165
https://github.com/project-oak/oak/blob/38522c9462591b25ff1f64865f2390ab5ef6eced/oak/server/rust/oak_runtime/src/runtime/mod.rs#L177-L188

PR #985, opened 2020-05-13
Use futures select for stopping gRPC nodes

ipetr0v:
This change:
- Adds futures select for Tokio Runtime
- Temporary disables gRPC client tests in `abitest` since they now require TLS certificates (will be re-enabled after examples start using Rust Loader)

Fixes https://github.com/project-oak/oak/issues/982

# Checklist
      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/logger.rs
+++ b/oak/server/rust/oak_runtime/src/node/logger.rs
@@ -19,6 +19,7 @@ use log::{error, info, log};
 use oak_abi::proto::oak::log::{Level, LogMessage};
 use prost::Message;
 use std::string::String;
+use tokio::sync::oneshot;
 
 /// Logging pseudo-Node.
 pub struct LogNode {
@@ -37,7 +38,12 @@ impl LogNode {
 impl super::Node for LogNode {
     /// Main execution loop for the logging pseudo-Node just waits for incoming
     /// `LogMessage`s and outputs them.
-    fn run(self: Box<Self>, runtime: RuntimeProxy, handle: oak_abi::Handle) {
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        _notify_receiver: oneshot::Receiver<()>,
+    ) {
         loop {
             // An error indicates the Runtime is terminating. We ignore it here and keep trying to
             // read in case a Wasm Node wants to emit remaining messages. We will return
--- a/oak/server/rust/oak_runtime/src/node/mod.rs
+++ b/oak/server/rust/oak_runtime/src/node/mod.rs
@@ -21,6 +21,7 @@ use std::{
     string::String,
     sync::Arc,
 };
+use tokio::sync::oneshot;
 use tonic::transport::{Certificate, Identity, Uri};
 
 pub mod external;
@@ -30,9 +31,17 @@ mod wasm;
 
 /// Trait encapsulating execution of a Node or pseudo-Node.
 pub trait Node {
-    /// Execute the Node, using the provided `Runtime` reference and initial handle.  The method
+    /// Execute the Node, using the provided `Runtime` reference and initial handle. The method
     /// should continue execution until the Node terminates.
-    fn run(self: Box<Self>, runtime: RuntimeProxy, handle: oak_abi::Handle);
+    ///
+    /// `notify_receiver` receives a notification from the Runtime upon termination. This
+    /// notification can be used by the Node to gracefully shut down.
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        notify_receiver: oneshot::Receiver<()>,
+    );
 }
 
 /// A [`Configuration`] corresponds to a [`NodeConfiguration`] protobuf message, with parsed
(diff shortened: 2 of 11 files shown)


oak/server/rust/oak_runtime/src/node/logger.rs
```diff
@@ -37,7 +38,12 @@ impl LogNode {
 impl super::Node for LogNode {
     /// Main execution loop for the logging pseudo-Node just waits for incoming
     /// `LogMessage`s and outputs them.
-    fn run(self: Box<Self>, runtime: RuntimeProxy, handle: oak_abi::Handle) {
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        _: oneshot::Receiver<()>,
```
tiziano88, line 45: Keep the name of the parameter here for documentation, i.e. `_notify_receiver` (and elsewhere)

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -32,7 +33,12 @@ mod wasm;
 pub trait Node {
     /// Execute the Node, using the provided `Runtime` reference and initial handle.  The method
     /// should continue execution until the Node terminates.
```
tiziano88, line 35: Update docs, please explain the `notify_receiver` param function.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -72,6 +73,10 @@ struct NodeInfo {
     /// (with one small exception, when the Runtime is in the process of closing
     /// down and the [`JoinHandle`] is held by the shutdown processing code).
     join_handle: Option<JoinHandle<()>>,
+
+    /// If the Node is currently running, holds the [`oneshot::Sender`] for
+    /// shutting down the async routine executed by the Nodes thread.
```
tiziano88, line 78: This comment could be clearer I think; basically this is a notification object whose receiver is sent to the node, and this is the sender; the agreement is that the runtime will notify it upon termination and then start waiting on the join handle; it's up to the node to figure out how to actually terminate when receiving a notification.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -72,6 +73,10 @@ struct NodeInfo {
     /// (with one small exception, when the Runtime is in the process of closing
     /// down and the [`JoinHandle`] is held by the shutdown processing code).
     join_handle: Option<JoinHandle<()>>,
+
+    /// If the Node is currently running, holds the [`oneshot::Sender`] for
```
tiziano88, line 77: Is it worth coalescing `join_handle` and `notify_sender` into a single struct (not sure what name), so that it can be held inside a single Option, and also it may expose a single method to notify and wait at the same time. WDYT?

ipetr0v: Maybe `struct NodeStopper` or `NodeNotifier`, which one is better?

ipetr0v: `NodeTerminator` probably will be too much =)

tiziano88: `NodeNotifier` seems fine, we can rename later if we come up with something better.

ipetr0v: The new struct will still need to have `Option` arguments, because [`join`](https://doc.rust-lang.org/std/thread/struct.JoinHandle.html) takes ownership of `self`.

tiziano88: I think you can `take` the entire struct from the option, and then I think you can just use its components? But if not, don't worry about it for now

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -30,9 +31,16 @@ mod wasm;
...
-    /// Execute the Node, using the provided `Runtime` reference and initial handle.  The method
+    /// Execute the Node, using the provided `Runtime` reference and initial handle. The method
     /// should continue execution until the Node terminates.
-    fn run(self: Box<Self>, runtime: RuntimeProxy, handle: oak_abi::Handle);
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        // Object that receives a notification from the Runtime before it stopps the Node. This
+        // notification can be used to correctly shut down async routines executed by the Node.
```
tiziano88, line 41:
```suggestion
        // Object that receives a notification from the Runtime upon termination. This
        // notification can be used by the Node to gracefully shut down.
```

but also I don't think this comment shows up anywhere (e.g. in the IDE or in generated docs), it would be better to have it as part of the method documentation I think?

ipetr0v: Done

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -68,18 +98,20 @@ struct NodeInfo {
...
 
 impl std::fmt::Debug for NodeInfo {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
             "NodeInfo {{'{}', label={:?}, join_handle={:?}, handles=[",
-            self.name, self.label, self.join_handle,
+            self.name,
+            self.label,
+            self.node_stopper.as_ref().map(|s| s.get_join_handle()),
```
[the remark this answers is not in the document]
ipetr0v: Done

PR #996, opened 2020-05-15
Add prometheus-to-sd to scrape metrics

### LINKED ISSUE #976: Visualise prometheus metrics on GCP
Subtask of #736:

Cloud Monitoring recommends using [OpenCensus for collecting custom metrics](https://cloud.google.com/monitoring/custom-metrics/open-census). However, there don't seem to be a client library for Rust. Furthermore, since we already use prometheus to collect metrics, using a client library for OpenCensus sounds redundant. 

Alternatively, we could deploy an OpenCensus server (or configure it accordingly if one is by default available for Kubernetes). OpenCensus can be configured to [receive prometheus metrics](https://opencensus.io/service/receivers/prometheus/) (also see [full config](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus.yml) for a prometheus server), and to [export them to stackdriver](https://opencensus.io/service/exporters/stackdriver/metrics/) (which should allow viewing the metrics in GCP).

The other alternative, which might be the simplest way to export prometheus metrics to GCP, is to use the [Stackdriver Prometheus sidecar](https://github.com/Stackdriver/stackdriver-prometheus-sidecar). 

Another alternative: [prometheus-to-sd](https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd) is a simple component that can scrape metrics stored in prometheus text format.

rbehjati:
Added a new container from [prometheus-to-sd](https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd)  to the aggregator pod. This container scrapes metrics from the aggregator oak-application.

Fixes #976 

# Checklist

      construction.

### DIFF
--- a/examples/aggregator/gcp/pod.yaml
+++ b/examples/aggregator/gcp/pod.yaml
@@ -10,6 +10,9 @@ spec:
       image: gcr.io/oak-ci/oak-aggregator:latest
       ports:
         - containerPort: 8080
+          name: aggregator
+        - containerPort: 3030
+          name: metrics
       volumeMounts:
         - name: tls-secret
           mountPath: '/etc/oak-secrets'
@@ -24,6 +27,31 @@ spec:
       env:
         - name: RUST_LOG
           value: info
+    # Additional container to scrape prometheus metrics:
+    # https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd
+    - name: prometheus-to-stackdriver
+      image: gcr.io/google-containers/prometheus-to-sd:v0.9.2
+      ports:
+        - containerPort: 6060
+      command:
+        - /monitor
+        # To be able to push custom metrics to the Stackdriver, the `stackdriver-prefix`
+        # flag must be set to `custom.googleapis.com`.
+        # Ref: https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd#custom-metrics
+        # On custom metrics: https://cloud.google.com/monitoring/custom-metrics#about_custom_metrics
+        - --stackdriver-prefix=custom.googleapis.com
+        - --source=oak-application:http://localhost:3030/metrics
+        - --pod-id=$(POD_NAME)
+        - --namespace-id=$(POD_NAMESPACE)
+      env:
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
   volumes:
     - name: tls-secret
       secret:
--- a/examples/aggregator/README.md
+++ b/examples/aggregator/README.md
@@ -101,6 +101,6 @@ Deployment requires Docker images to be uploaded to the
 following command:
 
 ```bash
-./scripts/build_example -e aggregator -i
+./scripts/build_example -e aggregator -i base
 ./examples/aggregator/scripts/docker_push
 ```


examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,26 @@ spec:
       env:
         - name: RUST_LOG
           value: info
+    - name: prometheus-to-sd
+      image: gcr.io/google-containers/prometheus-to-sd:v0.9.0
+      ports:
+        - name: profiler
```
tiziano88, line 33: What does `profiler` refer to?

examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,26 @@ spec:
...
+          containerPort: 6060
+      command:
+        - /monitor
+        - --stackdriver-prefix=custom.googleapis.com
```
tiziano88, line 37: Does this need to be this exact value? If so, could you add a link to some source?

examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,26 @@ spec:
       env:
         - name: RUST_LOG
           value: info
+    - name: prometheus-to-sd
```
tiziano88, line 30: Could you add a comment here? in particular, the `sd` part of the name is quite obscure. The link you have in the commit message should probably be here instead.

examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,26 @@ spec:
...
+      image: gcr.io/google-containers/prometheus-to-sd:v0.9.0
+      ports:
+        - name: profiler
+          containerPort: 6060
+      command:
+        - /monitor
+        - --stackdriver-prefix=custom.googleapis.com
+        - --source=oak-application:http://localhost:3030/metrics
```
tiziano88, line 38: Does this support a single replica of a single source? How will it work with a more realistic deployment?

examples/aggregator/server/BUILD
```diff
@@ -26,6 +26,7 @@ oak_docker(
     application = "//examples/aggregator/config:config",
     ports = [
         "8080/tcp",  # Oak Application gRPC port.
+        "3030/tcp",  # Metrics port.
```
ipetr0v, line 29: Does it also need the `6060` port?

examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,30 @@ spec:
...
+    # https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd
+    - name: prometheus-to-stackdriver
+      image: gcr.io/google-containers/prometheus-to-sd:v0.9.0
+      ports:
+        - containerPort: 6060
+      command:
+        - /monitor
+        # To be able to push custom metrics to the Stackdriver, the `stackdriver-prefix` 
+        # flag must be set to `custom.googleapis.com`.
+        # Ref: https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd#custom-metrics
```
tiziano88, line 40: Perhaps also add this https://cloud.google.com/monitoring/custom-metrics#about_custom_metrics which is the actual GCP reference.

examples/aggregator/gcp/pod.yaml
```diff
@@ -24,6 +27,30 @@ spec:
       env:
         - name: RUST_LOG
           value: info
+    # Additional container to scrape prometheus metrics:
+    # https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd
+    - name: prometheus-to-stackdriver
+      image: gcr.io/google-containers/prometheus-to-sd:v0.9.0
```
tiziano88, line 33: FYI it looks like the latest version is 0.9.2, is there any reason why it was not chosen?

See https://pantheon.corp.google.com/gcr/images/google-containers/GLOBAL/prometheus-to-sd?gcrImageListsize=30

tiziano88:
Note from https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd#alternatives :

> Google develops prometheus-to-sd primarily for Google Kubernetes Engine to collect metrics from system services in order to support Kubernetes users. We designed the tool to be lean when deployed as a sidecar in your pod. It's intended to support only the metrics the Kubernetes team at Google needs; you can use it to monitor your applications, but the feature set is limited.

@rbehjati it seems we should be using https://github.com/Stackdriver/stackdriver-prometheus-sidecar instead?

rbehjati:
> Note from https://github.com/GoogleCloudPlatform/k8s-stackdriver/tree/master/prometheus-to-sd#alternatives :
> 
> > Google develops prometheus-to-sd primarily for Google Kubernetes Engine to collect metrics from system services in order to support Kubernetes users. We designed the tool to be lean when deployed as a sidecar in your pod. It's intended to support only the metrics the Kubernetes team at Google needs; you can use it to monitor your applications, but the feature set is limited.
> 
> @rbehjati it seems we should be using https://github.com/Stackdriver/stackdriver-prometheus-sidecar instead?
> 
> also see https://cloud.google.com/monitoring/kubernetes-engine/prometheus

Yes. I can try `stackdriver-prometheus-sidecar` next. For that I think we have to have a deployment (instead of just a pod). I can fix that.

PR #997, opened 2020-05-15
Make scripts fail if any part of a pipeline fails

tiziano88:
Currently only the return value of the last command in a pipeline is
taken into account, any other failures are silently ignored.

In the reproducibility script this was actually causing the `sha256sum`
command failure to be ignored.

See
https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

# Checklist

      construction.

### DIFF
--- a/scripts/build_reproducibility_index
+++ b/scripts/build_reproducibility_index
@@ -13,7 +13,8 @@ source "$SCRIPTS_DIR/common"
 # List of artifacts that are expected to be reproducibly built.
 readonly REPRODUCIBLE_ARTIFACTS=(
   ./target/wasm32-unknown-unknown/release/*.wasm
-  ./target/x86_64-unknown-linux-musl/release/oak_loader
+  # TODO(#725): Include this line when `./scripts/build_server -s rust` actually builds the binary.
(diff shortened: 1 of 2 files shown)


scripts/build_reproducibility_index
```diff
@@ -13,7 +13,8 @@ source "$SCRIPTS_DIR/common"
 # List of artifacts that are expected to be reproducibly built.
 readonly REPRODUCIBLE_ARTIFACTS=(
   ./target/wasm32-unknown-unknown/release/*.wasm
-  ./target/x86_64-unknown-linux-musl/release/oak_loader
+  # TODO(#725): Include this line when `./scripts/build_server -s rust` actually builds the binary.
+  # ./target/x86_64-unknown-linux-musl/release/oak_loader
```
tiziano88, line 17: @ipetr0v BTW I am not sure why this is not the case currently? Is it intended?

tiziano88: Ah I think https://github.com/project-oak/oak/pull/932 was the cause.

ipetr0v: Yes, currently even the Rust Loader is built by Bazel

PR #1000, opened 2020-05-19
Specify zone in gcloud command

tiziano88:
This did not seem to work otherwise


tiziano88:
@rbehjati @ipetr0v how did it work for you? Do you have it set in some env variable or config file?

rbehjati:
I remember earlier I had to change the script to make it work (I did not commit that change). Then I found a way to configure the default zone for my gcloud installation. 
This is how you set it:
```
gcloud config set compute/zone <zone>
```
And you get the value via: 
```
gcloud config get-value compute/zone
or 
gcloud config list
```

I think it is good to have the change you have made in this PR. Otherwise, we should add somewhere in the docs something about setting the default zone.

PR #1001, opened 2020-05-19
Collect metrics using stackdriver-prometheus-sidecar

### LINKED ISSUE #976: Visualise prometheus metrics on GCP

rbehjati:
Adds a new deployment with a `prometheus server`, and a `stackdriver-prometheus-sidecar` to scrape metrics and send them to Cloud Monitoring. 

Fixes #976 

# Checklist

      construction.


rbehjati:
@tiziano88 Could you please run the script `examples/aggregator/scripts/deploy`, or give my user the `["container.clusterRoleBindings.create"]` permission?

Without a dedicated service account with the permissions specified in the script, prometheus is unable to use k8s service discovery, which seems to be the correct way to scrape metrics from components on k8s. Here is the error message from prometheus: 
`"level=error ts=2020-05-20T13:28:01.574Z caller=klog.go:94 component=k8s_client_runtime func=ErrorDepth msg="/app/discovery/kubernetes/kubernetes.go:283: Failed to list *v1.Pod: pods is forbidden: User \"system:serviceaccount:default:default\" cannot list resource \"pods\" in API group \"\" at the cluster scope"
"`

tiziano88:
I just granted that to project-oak-team@google.com (I think), could you check and let me know if it worked for you?

rbehjati:
Yes. That worked. Thanks.

examples/aggregator/gcp/deployment.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
```
ipetr0v, line 1: Why are we using Deployment instead of Pod?

oak/server/rust/oak_glue/src/lib.rs
```diff
@@ -101,7 +101,7 @@ pub unsafe extern "C" fn glue_start(
             }
         };
         let runtime_config = oak_runtime::RuntimeConfiguration {
-            metrics_port: Some(3030),
+            metrics_port: Some(9090),
```
ipetr0v, line 104: Why the port has changed?

examples/aggregator/gcp/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+              name: aggregator
+            - containerPort: 9090
+              name: metrics
+          volumeMounts:
+            - name: tls-secret
+              mountPath: '/etc/oak-secrets'
+              readOnly: true
+          args:
+            - --application=config.bin
+            - --ca_cert=/etc/oak-secrets/ca.pem
```
ipetr0v, line 31: Parameters have been changed in https://github.com/project-oak/oak/pull/1012

examples/aggregator/gcp/deployment.yaml
rbehjati: Deployment allows scaling up and down the application. It basically controls the number of pods in an application.

oak/server/rust/oak_glue/src/lib.rs
rbehjati: It is the default port that k8s service discovery (indicated by `kubernetes_sd_configs`) looks for. I guess I could change the configuration for the prometheus server, but I thought it would be better to keep the standard default. The prometheus server itself runs on port 9090 too. That is why earlier I thought we have to use a different port for our metrics server, but that is not necessary.

examples/aggregator/gcp/metrics-sidecar.yaml
```diff
@@ -0,0 +1,167 @@
...
+      labels:
+        app: metrics-sidecar
+    spec:
+      serviceAccount: prometheus
+      securityContext:
+        runAsUser: 0
+      containers:
+        - name: prometheus
+          # 2.11 seems to be the highest version that is tested.
+          # TO-DO: experiment with newer versions
```
tiziano88, line 100: `TODO` is the format we use throughout.

examples/aggregator/gcp/metrics-sidecar.yaml
```diff
@@ -0,0 +1,167 @@
...
+          image: prom/prometheus:v2.11.2
+          args:
+            - '--config.file=/etc/prometheus/config/prometheus.yml'
+            - '--storage.tsdb.path=/data'
+            - '--storage.tsdb.min-block-duration=15m'
+            - '--storage.tsdb.max-block-duration=4h'
+            - '--storage.tsdb.retention=48h'
```
tiziano88, line 107: I guess the quotes are not really necessary, since they are not used elsewhere.

rbehjati: Thanks. Fixed.

examples/aggregator/gcp/metrics-sidecar.yaml
rbehjati: Thanks. I change the version to v2.18.1 (newest version), and removed the comment.

Issue #689, opened 2020-03-09
Refactor parameter specification


ipetr0v:
After talking to @tiziano88 we decided that it's worth thinking about refactoring a way parameters for Oak applications are specified.

Currently we only have parameters inside an application configuration `config.textproto` file that is compiled in a `config.bin` file using Protobuf:
https://github.com/project-oak/oak/blob/c4865b206ade84fdc3d730a855870ca675ad0fbc/examples/hello_world/config/config.textproto#L1-L13
And any change requires a recompilation, so it's not very convenient.

Basically, there are two types of possible parameters for an Oak Application:
- **Secure parameters**:
    - Parameters that should change the hash value of a Wasm module.
    - An example of a secure parameter could be the `SAMPLE_THRESHOLD` from the [Aggregator](https://github.com/project-oak/oak/issues/619) example. This `SAMPLE_THRESHOLD` specifies the amount of samples that should be collected from users before revealing the aggregated value.
- **Insecure parameters**:
    - Parameters that could be freely changed without worrying about affecting the user trust.
    - An example of an insecure parameter could be an external IP address to connect to (as long as Oak Application checks its TLS certificates).

In the case of insecure parameters it may feel strange to always recompile a `config.bin` whenever developers need to change something, for example when a backend service changed an IP address.

So I think it makes sense to have secure parameters hardcoded (for now), since all changes will definitely effect the Wasm module hash value. And for insecure parameters the most obvious solution would be to have an additional argument{s} for `oak_runner`:
```bash
./asylo_oak_runner --application=APPLICATION [-- MODULE_ARGUMENTS]
```

Possible ways of implementing include:
- Pass arguments in the `oak_main` function, so they could be parsed inside the module, and developers could use whatever configuration they want (`JSON`, `YAML` etc.).
- Have an additional `config` pseudo-node with a corresponding `fn get(name: &String) -> String` function.

cc @project-oak/core

P.S.
Also I think that `grpc_port` could be a command-line argument too, so it would be more convenient to deploy Oak applications (as long as we will have only one gRPC server pseudo-node in Oak applications).

tiziano88:
@ipetr0v let's start with insecure parameters only for now, since we will need them to configure things like gRPC endpoints, logging, authentication, etc.

To take a concrete example, let's imagine we need to provide a username and password to use to connect to a non-Oak server (e.g. for storage). 

[...]

PR #1004, opened 2020-05-19
Introduce config map format and parsing logic

tiziano88:
This will be used to pass configuration files and secrets to Oak
Applications, by specifying them via command line flags to the
`oak_loader` binary.

Also switch the top-level `oak_loader` error handling to the `anyhow`
crate.

Ref #689

# Checklist

### DIFF
--- a/oak/server/rust/oak_loader/Cargo.toml
+++ b/oak/server/rust/oak_loader/Cargo.toml
@@ -13,9 +13,13 @@ oak_debug = []
 default = ["oak_debug"]
 
 [dependencies]
+anyhow = "*"
 log = "*"
 oak_runtime = "=0.1.0"
 prost = "*"
 signal-hook = "*"
 simple_logger = "*"
 structopt = "*"
+
+[dev-dependencies]
+maplit = "*"
--- a/oak/server/rust/oak_loader/src/main.rs
+++ b/oak/server/rust/oak_loader/src/main.rs
@@ -24,10 +24,13 @@
 //!     --grpc-tls-certificate=<CERTIFICATE_PATH> \
 //!     --root-tls-certificate=<CERTIFICATE_PATH>
 
-use log::info;
+use anyhow::anyhow;
+use core::str::FromStr;
+use log::{debug, info};
 use oak_runtime::{configure_and_run, proto::oak::application::ApplicationConfiguration};
 use prost::Message;
 use std::{
+    collections::HashMap,
     fs::{read_to_string, File},
     io::Read,
     sync::{
@@ -37,11 +40,15 @@ use std::{
 };
 use structopt::StructOpt;
 
-use oak_runtime::proto::oak::application::node_configuration::ConfigType::{
-    GrpcClientConfig, GrpcServerConfig,
+#[cfg(test)]
+mod tests;
+
+use oak_runtime::proto::oak::application::{
+    node_configuration::ConfigType::{GrpcClientConfig, GrpcServerConfig},
+    ConfigMap,
 };
 
-#[derive(StructOpt, Clone)]
+#[derive(StructOpt, Clone, Debug)]
 #[structopt(about = "Oak Loader")]
 pub struct Opt {
     #[structopt(long, help = "Application configuration file.")]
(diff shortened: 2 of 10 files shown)


oak/server/rust/oak_loader/Cargo.toml
```diff
@@ -13,9 +13,13 @@ oak_debug = []
 default = ["oak_debug"]
 
 [dependencies]
+anyhow = "*"
```
ipetr0v, line 16: I think this should also be added to `oak/server/rust/oak_loader/BUILD`

tiziano88: Done.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -70,24 +77,73 @@ pub struct Opt {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ConfigEntry {
+    key: String,
+    filename: String,
+}
+
+impl FromStr for ConfigEntry {
+    type Err = anyhow::Error;
+    fn from_str(v: &str) -> Result<Self, Self::Err> {
+        let parts = v.split("=").collect::<Vec<_>>();
```
conradgrobler, line 98: This seems quite brittle. It might be nice in future to have some better defined parsing.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -70,24 +77,73 @@ pub struct Opt {
...
         eprintln!("No debugging output configured at build time");
     }
     let opt = Opt::from_args();
+    debug!("parsed opts: {:?}", opt);
+
+    let config_map = parse_config_map(&opt.config_files)?;
+    // We only log the keys here, since the values may be secret.
+    debug!("parsed config map entries: {:?}", config_map.items.keys());
+    // TODO(#689): Pass the `config_map` object to the Runtime instance, and make it available to
+    // the running Oak Application.
```
conradgrobler, line 146: I am not clear on how this will be used. Is the intention to make this available to wasm nodes? Or to dynamically replace tokens in the application configuration?

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -70,24 +77,73 @@ pub struct Opt {
...
+        .map_err(|error| anyhow!("Failed to read file <{}>: {:?}", filename, error))?;
     Ok(buffer)
 }
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn parse_config_files(config_entries: &[ConfigEntry]) -> anyhow::Result<HashMap<String, Vec<u8>>> {
+    let mut file_map = HashMap::new();
+    for config_entry in config_entries {
+        let file_content = read_file(&config_entry.filename)?;
+        file_map.insert(config_entry.key.to_string(), file_content);
```
conradgrobler, line 122: Perhaps log a warning if the key already exists and the value is being overwritten?

oak/server/rust/oak_loader/src/main.rs
tiziano88: I think this is fine for the foreseeable future, we fully control the file names anyways, so there is no risk that we end up with `=` in the file name itself. But sure, if it does happen we can extend it.

oak/server/rust/oak_loader/src/main.rs
tiziano88: Good point, done and added a test.

oak/server/rust/oak_loader/src/main.rs
tiziano88: The intention is to make it available to the first node of an Oak application, which can then take it apart and distribute it to further nodes if it needs to. It will be easier after #917 is merged, so I'll wait for that before doing the rest of the work.

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Seeing that this is also intended for sensitive data, is there a plan to add labels or some other mechanism to protect the data from untrusted nodes?

PR #1010, opened 2020-05-21
Rename secrecy to confidentiality

tiziano88:
Seems a more accurate representation of the fact that data is kept
confidential to a set of subjects, as opposed to just secret, which
feels more like all-or-nothing.

# Checklist

      construction.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -140,7 +140,7 @@ its ephemeral Channel handles.
 - Nodes and Channels have **Labels**, fixed at creation time
 - The Runtime keeps track of Labels, and enforces flow of data between Nodes and
   Channels based on them
-- Each Label has two **components**: **secrecy** and **integrity**
+- Each Label has two **components**: **confidentiality** and **integrity**
 - Each component is an unordered set of **tags**
 - Each tag is a structured representation of a **security principal**
 - A security principal is either a **user** or a **computation**
@@ -163,7 +163,7 @@ creation of side channels based on Labels.
 
 #### Components
 
-A Label has two components: **secrecy** and **integrity**.
+A Label has two components: **confidentiality** and **integrity**.
 
 Each component is represented as an (unordered) set of tags, and each tag refers
 to an individual security principal in the system.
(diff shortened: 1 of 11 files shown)


aferr:
I think the terminology change is an improvement.

Initially, I did not have a strong opinion, and thought the two terms "secrecy" and "confidentiality" were interchangeable. Though thinking about it I think I tend to use the word "confidentiality" more often myself.

I thought a more objective way to look at this was to do a quick scan of some IFC papers. The information is a bit gray, but it looks like confidentiality might be the better term.

**Some highly-cited surveys:**
https://www.cs.cornell.edu/andru/papers/jsac/sm-jsac03.pdf

This one actually compares the terms "confidentiality", "secrecy" and "confinement in Section II.

"Terminology for security properties relating to confidentiality is somewhat inconsistent ... However, both “confinement” and “secrecy” have been used to describe related
but weaker security properties. ..."

https://www.cs.yale.edu/homes/dsc5/papers/sabelfeld-sands-jcs07.pdf

Uses both terms with similar frequency and interchangably.

https://link.springer.com/content/pdf/10.1007%2F3-540-45608-2_6.pdf

Primarily uses the term secrecy (though this survey is now a bit outdated).


**Somewhat arbitrarily chosen papers by different sets of authors than the above about IFC**
https://users.soe.ucsc.edu/~cormac/papers/popl12b.pdf
http://askarov.net/depsec.pdf
https://people.mpi-sws.org/~dg/papers/csf2018.pdf'

All use confidentiality more often than secrecy, or do not use secrecy in the main text at all (just in titles of papers in references).

These are all "PL papers"

https://www.scs.stanford.edu/~dm/home/papers/zeldovich:histar.pdf
Uses secrecy

https://pdos.csail.mit.edu/papers/asbestos-sosp05.pdf
Uses secrecy 2x, confidentiality 1x, seemingly interchangably.

docs/concepts.md
```diff
@@ -215,17 +216,17 @@ Intuitively, data can only flow from `a` to `b` if:
 The least privileged label is usually referred to as "public trusted" (and
```
aferr, line 216: Not related to the change you just made, but I would consider changing this to say:

"The information flow label  "public trusted" corresponds to a Node or Channel which has only observed public data, and is universally trusted. This is the bottom of the information flow label (and it may also be represented as `⊥`, pronounced "bottom"), which means information from sources with this label are allowed to influence anything."

Reasons for this suggestion:
- Avoids a non-standard use of "endorsed" -- endorsed usually means the integrity label has changed (down the lattice)
- Updates some prose around the explanation of the "trusted" part of this label (It sounds like it used to be "untrusted")
- [I am somewhat less convinced about this part of the change myself] avoids the word privilege, because maybe privilege implies "authority lattice" wherein public_untrusted has the least privilege (can't declassify anything, can't endorse anything). Though, I suppose the information flow lattice is also about privilege in a way because it is privilege to receive information in the first place.

tiziano88:
Thanks for the literature review @aferr , it's interesting to have here as reference! If you are happy with the change, could you mark the PR as approved?

aferr:
> Thanks for the literature review @aferr , it's interesting to have here as reference! If you are happy with the change, could you mark the PR as approved?

I did just have one comment.

docs/concepts.md
tiziano88: I have another PR that clarifies some of this, I'll incorporate your feedback in that!

aferr: OK, SG approved.

PR #1031, opened 2020-05-26
Print reproducibility index diff

tiziano88:
This should make it easier to actually understand what targets were
affected by a PR.

Ref. #861


tiziano88:
Now let's see if it *actually* works

PR #1033, opened 2020-05-27
Fix name of `public_untrusted` label

### LINKED ISSUE #980: Rename `public_trusted` label to `public_untrusted`
https://github.com/project-oak/oak/blob/11f96d62555364286881e3cc081c98d5754c5f19/oak/server/rust/oak_abi/src/label/mod.rs#L47-L55

I think the (incorrect) `public_trusted` naming was due to my misunderstanding of the IFC lattice structure in some early discussions with @aferr . 

@aferr to clarify: this label is the least privileged (in [Flume](https://pdos.csail.mit.edu/papers/flume-sosp07.pdf) terms, it has empty confidentiality and integrity components) so I believe it should be called `public_untrusted`, right? Or should we call it `bottom`, or something else?

Aside: if anything, `public_trusted` is perhaps the label that the oak runtime itself would have, if it had one (though it does not have an explicit one currently), since it is indeed trusted by clients, e.g. over gRPC, thanks to remote attestation (but that's another discussion anyways).

tiziano88:
This was incorrectly referred to as `public_trusted`, while in fact it
is not trusted by anyone, as it has an empty integrity label.

Fix #980

### DIFF
--- a/oak_abi/src/label/tests.rs
+++ b/oak_abi/src/label/tests.rs
@@ -60,7 +60,7 @@ fn label_flow() {
     let tag_0 = authorization_bearer_token_hmac_tag(&[0, 0, 0]);
     let tag_1 = authorization_bearer_token_hmac_tag(&[1, 1, 1]);
 
-    let public_trusted = Label::public_trusted();
+    let public_untrusted = Label::public_untrusted();
 
     // A label that corresponds to the confidentiality of tag_0.
     //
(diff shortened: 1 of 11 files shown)


oak_abi/src/label/tests.rs
```diff
@@ -101,10 +101,10 @@ fn label_flow() {
     //       label_0    label_1
     //             \     /
     //              \   /
-    //          public_trusted
+    //          public_untrusted
```
aferr, line 104: but public_untrusted isn't the bottom label.

tiziano88: Not even in the context of the 5 labels of this example? This is not meant to be an exhaustive diagram :)

aferr: Ooopsies you are right.

PR #1039, opened 2020-05-28
Exclude `oak_abi` and `oak_utils` from the workspace

rbehjati:
Fixes the problem described in [this comment](https://github.com/project-oak/oak/issues/971#issuecomment-635201450).

# Checklist

      construction.


tiziano88:
I guess this will have to be undone as part of #971 ? The whole point of that is to end up with 2/3 non-overlapping workspaces

rbehjati:
I think this fixes the problem with the build, and the docs. Since `oak_abi` is in the same folder as `Cargo.toml` it has to be a member of the workspace. We should be able to make `oak_abi` a standalone crate after the rest of #971 is completed.

tiziano88:
> Since `oak_abi` is in the same folder as `Cargo.toml` it has to be a member of the workspace. 

I thought the error message I posted earlier (https://github.com/project-oak/oak/issues/971#issuecomment-635201450) actually mentions a couple of ways of excluding it from the workspace, doesn't it?

rbehjati:
> I thought the error message I posted earlier ([#971 (comment)](https://github.com/project-oak/oak/issues/971#issuecomment-635201450)) actually mentions a couple of ways of excluding it from the workspace, doesn't it?

I am using `workspace.exclude` now. This required excluding `oak_utils` from the workspace too, to avoid circular dependencies. I tried to make it a module in the `oak_abi` package, but I am not sure how to make it work, since `oak_utils` is used in `build.rs`. Any suggestions?

tiziano88:
Do you mean that if moved under `oak_abi::utils`, then it cannot be used from the `build.rs` file?

rbehjati:
> Do you mean that if moved under `oak_abi::utils`, then it cannot be used from the `build.rs` file?

Yes.

tiziano88:
If that's the case, it does sound like `oak_utils` needs to be a top-level crate too, unfortunately.

rbehjati:
> If that's the case, it does sound like `oak_utils` needs to be a top-level crate too, unfortunately.

Yes. It is unfortunate. 
This change does not fix the problem with the docs (#1037). I'll fix that in a different PR.

PR #1045, opened 2020-05-29
Make examples a separate workspace

rbehjati:
Ref #971 and #765

# Checklist

      construction.


examples/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+aggregator_backend = { path = "aggregator/backend" }
+aggregator_common = { path = "aggregator/common" }
+chat = { path = "chat/module/rust" }
+hello_world = { path = "hello_world/module/rust" }
+machine_learning = { path = "machine_learning/module/rust" }
+private_set_intersection = { path = "private_set_intersection/module/rust" }
+running_average = { path = "running_average/module/rust" }
+rustfmt = { path = "rustfmt/module/rust" }
+translator = { path = "translator/module/rust" }
+translator_common = { path = "translator/common" }
```
tiziano88, line 48: Why are all these patches needed?

examples/aggregator/scripts/run_backend
```diff
@@ -5,6 +5,7 @@ readonly SCRIPTS_DIR="$(dirname "$(readlink -f "$0")")"
 # shellcheck source=scripts/common
 source "${GLOBAL_SCRIPTS_DIR}/common"
 
+cd "examples"
```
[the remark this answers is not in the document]
tiziano88: Please don't `cd` in the middle of scripts, unless necessary, and even so, please use a subshell instead.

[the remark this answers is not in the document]
rbehjati: Apparently it is not needed, but the script does not quite work even on master.

examples/Cargo.toml
rbehjati: I guess Cargo needs these to find the dependencies. Previously, they were in the top-level `Cargo.toml`.

tiziano88: I'm aware they were there, but I don't think they are actually relevant any more. I think they were just a convenience so that we could use `--package=translator`, but now that they are in a separate workspace we probably need to specify the entire manifest path anyways, so I don't think they buy us anything. Or it would be good to know if they are actually used for anything else :)

rbehjati: I see! I removed them except from `translator_common = { path = "translator/common" }`. That one is still needed, since it is a dependency for `translator` and `hello-world`. Specifying the manifest-path does not seem to bring `translator_common` into scope.

Issue #1024, opened 2020-05-26
Add Trusted Information Retrieval example


ipetr0v:
After adding the Aggregator example (https://github.com/project-oak/oak/pull/626 https://github.com/project-oak/oak/pull/699), the next possible good example of Oak usage could Trusted Information Retrieval (similar to [Private Information Retrieval](https://en.wikipedia.org/wiki/Private_information_retrieval)).

I think the simplest PoC implementation could be:
- Server that contains coordinates of different example points of interest (shops, etc.)
- Client that sends its coordinates to the server and retrieves the nearest point of interest
- The whole database could be in-memory (as the initial PoC)
- [optional] Client authentication using TLS
- [optional] An Android client application

cc @project-oak/core

tiziano88:
Nice plan!

FYI I think #1004 may be very useful here, to pass in the database to use for the lookups to the Oak Application, without having to recompile its source code every time.

As for interesting data sources, I have been looking around for something that could be used for this; it would be nice to have a list of Points Of Interest (POI) indexed by location, and being able to privately look them up based on the contents of an incoming request from a client, but I could not find such database yet. In the meanwhile though, I came across the [Transport For London API](https://api-portal.tfl.gov.uk/docs), which has, among others, an XML file with the availability of santander bikes for hire. Perhaps a nice initial demo would be to ingest this data in the application, and serve it to client in a private way. Then we can change it so that the Oak Application periodically downloads a fresh version of this file, which will require making sure that there is no interference between nodes that see secret data, and the node that downloads the data feed.

PR #1049, opened 2020-05-29
Add Trusted Information Retrieval example

ipetr0v:
This change adds a Trusted Information Retrieval example.

Fixes https://github.com/project-oak/oak/issues/1024

# Checklist


examples/private_information_retrieval/client/private_information_retrieval.cc
```diff
@@ -0,0 +1,84 @@
...
+      oak::ApplicationClient::CreateTlsChannel(address, ca_cert));
+
+  // Parse arguments.
+  auto location = absl::GetFlag(FLAGS_location);
+  if (location.size() != 2) {
+    LOG(FATAL) << "Incorrect number of coordinates: " << location.size() << " (expected 2)";
+  }
+  float latitude;
+  if (!absl::SimpleAtof(location.front(), &latitude)) {
+    LOG(FATAL) << "Incorrect latitude: " << location.front();
```
conradgrobler, line 73: Perhaps "Latitude must be a valid floating point number." or something like that. Simlarly for Longitude.

conradgrobler: Also might be worthwhile doing additional range validation (eg longitude >= -180 and <= 180, latitude >=-90 and <= 90)

examples/private_information_retrieval/module/rust/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+};
+
+/// Oak Node that contains an in-memory database of Points Of Interest.
+pub struct PrivateInformationRetrievalNode {
+    points_of_interest: Vec<PointOfInterest>,
+}
+
+impl PrivateInformationRetrievalNode {
+    /// Returns a Euclidean distance between two locations.
+    fn distance(first: Location, second: Location) -> f32 {
```
conradgrobler, line 53: Probably not important, as this is just an example, but Euclidean distance assumes a flat plane - not a sphere (which is implied by longitude and latitude).

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+    long: f32,
+    installed: bool,
+    locked: bool,
+    #[serde(rename = "installDate", default)]
+    install_date: String,
+    #[serde(rename = "removalDate", default)]
+    removal_date: String,
+    temporary: bool,
+    #[serde(rename = "nbBikes", default)]
+    nb_bikes: u32,
```
conradgrobler, line 57: What does the `nb_` prefix mean? Pehaps rename to `number_of_bikes` or something similar?

scripts/run_server
```diff
@@ -66,16 +76,12 @@ if [[ -n ${EXAMPLE+x} ]]; then
...
 
-declare -a ADDITIONAL_ARGS
-ADDITIONAL_ARGS+=(
-  "--grpc-tls-private-key=${SCRIPTS_DIR}/../examples/certs/local/local.key"
-  "--grpc-tls-certificate=${SCRIPTS_DIR}/../examples/certs/local/local.pem"
-  "--root-tls-certificate=${SCRIPTS_DIR}/../examples/certs/local/ca.pem"
-)
-if [[ -n "${OIDC_CLIENT+x}" ]]; then
-  readonly ADDITIONAL_ARGS+=("--oidc-client=${OIDC_CLIENT}")
+  if [[ "${EXAMPLE}" == "private_information_retrieval" ]]; then
```
conradgrobler, line 80: Although this works for a single example, this is not maintainable if multiple examples start using different config files. We should thinkg about a different way of passing this into the script.

Cargo.toml
```diff
@@ -53,6 +54,7 @@ aggregator_common = { path = "examples/aggregator/common" }
 chat = { path = "examples/chat/module/rust" }
 hello_world = { path = "examples/hello_world/module/rust" }
 machine_learning = { path = "examples/machine_learning/module/rust" }
+private_information_retrieval = { path = "examples/private_information_retrieval/module/rust" }
```
tiziano88, line 57: I don't think this is necessary?

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
   "examples/chat/module/rust",
   "examples/hello_world/module/rust",
   "examples/machine_learning/module/rust",
+  "examples/private_information_retrieval/module/rust",
```
tiziano88, line 14: BTW, should we rename this example to `trusted_information_retrieval` or something like that? perhaps start a thread on slack to discuss

examples/private_information_retrieval/database.xml
```diff
@@ -0,0 +1 @@
...</lat><long>-0.129698963</long><installed>true</installed><locked>false</locked><installDate>1392039000000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>14</nbBikes><nbEmptyDocks>5</nbEmptyDocks><nbDocks>21</nbDocks></station><station><id>763</id><name>Mile End Park Leisure Centre, Mile End</name><terminalName>300027</terminalName><lat>51.52059714</lat><long>-0.032566533</long><installed>true</installed><locked>false</locked><installDate>1392043020000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>34</nbBikes><nbEmptyDocks>12</nbEmptyDocks><nbDocks>47</nbDocks></station><station><id>764</id><name>St. John's Road, Clapham Junction</name><terminalName>200185</terminalName><lat>51.46239255</lat><long>-0.16829214</long><installed>true</installed><locked>false</locked><installDate>1392633480000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>29</nbBikes><nbEmptyDocks>4</nbEmptyDocks><nbDocks>33</nbDocks></station><station><id>765</id><name>Ranelagh Gardens, Fulham</name><terminalName>300064</terminalName><lat>51.46760141</lat><long>-0.20682737</long><installed>true</installed><locked>false</locked>
```
tiziano88, line 1: Can you add a README file with a description of this file, its provenance, a license if needed, and some instructions on how to re-generate it when needed?

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+    lat: f32,
+    long: f32,
+    installed: bool,
+    locked: bool,
+    #[serde(rename = "installDate", default)]
+    install_date: String,
+    #[serde(rename = "removalDate", default)]
+    removal_date: String,
+    temporary: bool,
+    #[serde(rename = "nbBikes", default)]
```
tiziano88, line 56: Where does all this structure come from? If it's documented somewhere on the internet, please link to it. If not, please explain how you arrived to it.

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+// limitations under the License.
+//
+
+//! Private Information Retrieval example for Project Oak.
+//!
+//! This shows how an Oak Node can aggregate data samples and report aggregated values if there are
+//! enough samples to hide individual contributors (enforces k-anonymity).
+//!
+//! Clients invoke the module by providing data samples that contain a bucket ID
+//! and a Sparse Vector - a dictionary with integer keys.
```
tiziano88, line 23: Wrong comment.

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+            Err(OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs))
+        }
+    }
+}
+
+/// Parse an XML database into a vector of [`PointOfInterest`].
+pub fn parse_database(xml_database: &[u8]) -> Result<Vec<PointOfInterest>, OakError> {
+    let database: Database = from_str(
+        String::from_utf8(xml_database.to_vec())
+            .expect("Couldn't convert vector to string")
```
tiziano88, line 89: It seems excessive to panic for a malformed file.

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+    .map_err(|error| {
+        error!("Couldn't parse XML data: {:?}", error);
+        OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
+    })?;
+
+    let points_of_interest = database
+        .stations
+        .iter()
+        // Filter out closed stations, and stations with no bikes.
+        .filter(|station| station.installed && !station.locked && station.removal_date == "")
```
tiziano88, line 101:
```suggestion
        .filter(|station| station.installed && !station.locked && station.removal_date.is_empty())
```

examples/private_information_retrieval/module/rust/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    Location, PointOfInterest, PrivateInformationRetrieval, PrivateInformationRetrievalDispatcher,
+};
+
+/// Oak Node that contains an in-memory database of Points Of Interest.
+pub struct PrivateInformationRetrievalNode {
+    points_of_interest: Vec<PointOfInterest>,
+}
+
+impl PrivateInformationRetrievalNode {
+    /// Returns a Euclidean distance between two locations.
```
tiziano88, line 52: What's the unit of this distance? I guess degrees?

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+struct Station {
+    id: u32,
+    name: String,
+    #[serde(rename = "terminalName", default)]
+    terminal_name: String,
+    lat: f32,
+    long: f32,
```
tiziano88, line 48: Are these expressed in degrees? If so, perhaps `latitude_degrees` and `longitude_degrees`

examples/private_information_retrieval/module/rust/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+                if distance < min_distance {
+                    (Some(point.clone()), distance)
+                } else {
+                    (closest_point, min_distance)
+                }
+            },
+        );
+        match nearest_point.0 {
+            Some(point) => {
+                debug!("Sending the nearest Point Of Interest: {:?}", point);
```
tiziano88, line 82: I don't think this is sending anything, perhaps "Found nearest Point Of Interest"?

examples/private_information_retrieval/module/rust/src/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+use assert_matches::assert_matches;
+use maplit::hashmap;
+use oak::grpc;
+use oak_abi::{proto::oak::application::ConfigMap, OakStatus};
+use oak_runtime::io::Encodable;
+
+const MODULE_CONFIG_NAME: &str = "private_information_retrieval";
+
+const XML_DATABASE: &str =
+"<?xml version=\"1.0\" encoding=\"utf-8\"?><stations lastUpdate=\"1590775020879\" version=\"2.0\">
```
tiziano88, line 30: Probably easier to use a raw string literal.

examples/private_information_retrieval/proto/private_information_retrieval.proto
```diff
@@ -0,0 +1,35 @@
...
+  float longitude = 2;
+}
+
+message PointOfInterest {
+  string name = 1;
+  Location location = 2;
+}
+
+service PrivateInformationRetrieval {
+  rpc GetNearestPointOfInterest(Location) returns (PointOfInterest);
```
tiziano88, line 34: It's good practice to define explicit request and response messages, see https://cloud.google.com/apis/design/naming_convention#request_and_response_messages . 

This is the perfect example: imagine we want to add a parameter to specify how many entries to return to the request, and return multiple entries in the response: in the current definition it would not be possible to do either of these things without making a breaking change.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -136,6 +141,18 @@ pub fn parse_config_map(config_files: &[ConfigEntry]) -> anyhow::Result<ConfigMa
...
+/// Send configuration map to the initial Oak Node.
+fn send_config_map(
+    config_map: ConfigMap,
+    runtime: &RuntimeProxy,
+    handle: Handle,
+) -> anyhow::Result<()> {
+    let sender = Sender::new(handle);
+    sender
+        .send(config_map, runtime)
+        .map_err(|status| anyhow!("could configuration map to the initial Node: {:?}", status))
```
tiziano88, line 153: Missing some words from the error message?

scripts/run_server
tiziano88: If only we could use a real programming language to run the examples... https://github.com/project-oak/oak/tree/master/runner :)

examples/private_information_retrieval/proto/private_information_retrieval.proto
ipetr0v: Done

examples/private_information_retrieval/module/rust/src/database.rs
```diff
@@ -0,0 +1,110 @@
...
+    )
+    .map_err(|error| {
+        error!("Couldn't parse XML data: {:?}", error);
+        OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
+    })?;
+
+    let points_of_interest = database
+        .stations
+        .iter()
+        // Filter out closed stations, and stations with no bikes.
```
tiziano88, line 99: Which part of the filter removes the stations with no bikes? Is that what `installed` means?

examples/private_information_retrieval/proto/private_information_retrieval.proto
```diff
@@ -0,0 +1,43 @@
...
+message Request {
+  Location location = 1;
+}
+
+message Response {
+  PointOfInterest point_of_interest = 2;
+}
+
+service PrivateInformationRetrieval {
+  rpc GetNearestPointOfInterest(Request) returns (Response);
```
tiziano88, line 42: Please do follow standard naming conventions: 
- https://cloud.google.com/apis/design/naming_convention#method_names
- https://cloud.google.com/apis/design/naming_convention#request_and_response_messages

s/GetNearestPointOfInterest/ListNearestPointsOfInterest/
s/Request/ListNearestPointsOfInterestRequest/
s/Response/ListNearestPointsOfInterestResponse/

(or even omit "nearest" entirely)

I know it seems verbose, but otherwise it seems almost impossible to evolve this API in a reasonable way.

tiziano88, line 42: Could you add a short doc comment to this method?

examples/private_information_retrieval/module/rust/src/lib.rs
tiziano88: Personally I think converting to km is complete overkill, especially since AFAICT you are not actually using the absolute value for anything, you just want to find the minimum one. Or am I missing something?

examples/private_information_retrieval/proto/private_information_retrieval.proto
ipetr0v: Done

ipetr0v: Done
But the service currently returns a single POI (but probably will be updated to return a list).

PR #1077, opened 2020-06-03
Pass additional args to aggregator example

tiziano88:


tiziano88:
I'm not sure why it was not done before though, let me know if there is something wrong with it.

tiziano88:
@daviddrysdale could you double check my bash homework? :)

PR #1080, opened 2020-06-03
More accurate logging of gRPC requests lifecycle

tiziano88:
If there is an error when reading a gRPC request from an invocation,
then attempt to send an error response to the caller rather than leave
it hanging forever (this happens in case of misplaced labels).

# Checklist

### DIFF
--- a/sdk/rust/oak/src/grpc/mod.rs
+++ b/sdk/rust/oak/src/grpc/mod.rs
@@ -17,7 +17,7 @@
 //! Functionality to help Oak Nodes interact with gRPC.
 
 use crate::OakError;
-use log::error;
+use log::{error, warn};
 use oak_abi::proto::google::rpc;
 pub use oak_abi::proto::{
     google::rpc::*,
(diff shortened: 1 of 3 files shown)


sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -155,24 +155,37 @@ impl<T: ServerNode> crate::Node<Invocation> for T {
...
+                // TODO(#1078): If it's clear that the issue is caused by a too restrictive label
+                // (permission denied), then return a more specific error to the caller.
+                warn!(
+                    "could not receive gRPC request from gRPC invocation: {}",
+                    err
+                );
+                // If anything fails while reading the request, then at least try to send a response
+                // back to the caller, rather than just returning from this method with `?`, since
+                // otherwise that would leave the caller hanging forever.
+                return response_writer.close(Result::Err(build_status(
```
conradgrobler, line 174: This seems a bit weird. Even though an error happened, the caller will receive an `Ok` status if the channel closes successfully.

tiziano88: You are right, that was not great, I made it so that the result of the closing is just logged, but the original error is returned to the caller. Thanks for the suggestion!

PR #1084, opened 2020-06-04
Improve logging of gRPC request handling 

tiziano88:
The main change is that we log one info entry for each gRPC request,
including details of its label, so that it's possible to debug label
mismatch issues.

Also do not log the contents of the request, only its size, and the
client IP address (though this does not seem to work reliably yet).


tiziano88:
This is now blocked by #865, since the actual has of the wasm module is different on my machine and on cloud build.

PR #1085, opened 2020-06-04
Manage Node.js dependencies with bazel

juliettepretot:
This will make it easier to integrate this example with our scripts

# Checklist

      construction.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -14,7 +14,10 @@
 # limitations under the License.
 #
 
-workspace(name = "oak")
+workspace(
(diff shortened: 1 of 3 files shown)


juliettepretot:
@conradgrobler Good point! I'm a bit out of the loop on this, if we decided to move away from bazel & happy to close this and rely on the node installation inside of our docker file instead.

tiziano88:
Yeah Docker is preferable, if it works for you @juliettepretot

juliettepretot:
Can do :)

PR #1105, opened 2020-06-05
Define example manifest files

tiziano88:
# Checklist

      construction.


examples/aggregator/example.toml
```diff
@@ -0,0 +1,9 @@
+name = "aggregator"
+
+[modules.module]
+lang = "Rust"
```
anghelcovici, line 4: You probably don't need both the lang and cargo_manifest? If it has a cargo_manifest it should be clear it's cargo :)

examples/aggregator/example.toml
```diff
@@ -0,0 +1,9 @@
...
+cargo_manifest = "examples/aggregator/module/rust/Cargo.toml"
+
+[clients.cpp]
+lang = "Cpp"
+additional_args = ["--bucket=test", "--data=1:10,2:20,3:30"]
```
anghelcovici, line 9: Same here - if you have a clients.cpp maybe you don't need to specify the lang?

examples/aggregator/example.toml
tiziano88: In pure Rust this would be an enum with different fields, and everything would be type safe. In principle, that could be done for TOML too, but the parser does not seem to like those kind of enums :/ https://github.com/alexcrichton/toml-rs/issues/390

examples/abitest/example.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[modules]
+module_0 = { Cargo = { cargo_manifest = "examples/abitest/module_0/rust/Cargo.toml" } }
+module_1 = { Cargo = { cargo_manifest = "examples/abitest/module_1/rust/Cargo.toml" } }
+
+[clients]
+cpp = { Bazel = { bazel_target = "//examples/abitest/client:client" }, additional_args = [
+  "--test_exclude=(Storage|GrpcServerServerStreamingMethod)",
+  "--cert_chain=../../../../../../../../examples/certs/local/local.pem",
+  "--private_key=../../../../../../../../examples/certs/local/local.key",
```
juliettepretot, line 11: I know it's a minor thing, but the `../../../../../../../../` paths honestly confuse me so much. But I guess with us moving away from bazel that's a seperate thing. :)

tiziano88: Yeah, that's orthogonal to this change, please ignore for now :)

runner/src/main.rs
```diff
@@ -319,26 +351,42 @@ fn run_example(example: &Example) -> Step {
...
+                        .collect(),
                 }),
             },
         ],
     }
 }
 
+fn run_client(name: &str, client: &Client) -> Step {
+    match &client.target {
+        Target::Cargo { .. } => todo!(),
```
juliettepretot, line 369: TIL that Rust had a literal `todo` macro. 😅

runner/src/main.rs
```diff
@@ -319,26 +351,42 @@ fn run_example(example: &Example) -> Step {
...
-                            "--ca_cert=../../../../../../../../examples/certs/local/ca.pem"
-                                .to_string(),
-                        ]
+                    steps: example
+                        .clients
                         .iter()
-                        .chain(example.additional_client_flags.iter()),
-                    ),
+                        .map(|(name, client)| run_client(name, &client))
+                        .collect(),
```
juliettepretot, line 360: That's nice, it seems like this PR is unofficially adding support for running multiple clients

runner/src/main.rs
```diff
@@ -289,10 +303,28 @@ fn run_example_server(application_file: &str) -> Cmd {
...
+    modules: HashMap<String, Target>,
+    clients: HashMap<String, Client>,
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+enum Target {
+    Bazel { bazel_target: String },
+    Cargo { cargo_manifest: String },
+}
```
juliettepretot, line 319: What I like about the manifest is that it specifies everything needed to run an example in one place, as opposed if-conditions in our various build scripts. :) 

With that approach, it has occurred to me that maybe it would make sense to keep the target logic outside of the runner and instead in each manifest. 

Eg instead of specifying a target, and the runner having instructions on how to build with that target, could the manifest simply list a command that should be run, leaving the details of the build system up to it?

Lemme know if that makes sense. :) 

Also curious to hear @rbehjati thoughts, since this somewhat relates to our conversation in the Oak Channel.

examples/abitest/example.toml
juliettepretot: Feel free to ignore this since it's not actionable, but purely to satisfy my curiosity,: 

This basically navigates us out of the bazel runner directory and back into the working directory to load the certs, right? Had we considered instead copying the certs in into the bazel runner dir with a build rule?

runner/src/main.rs
rbehjati: Sorry for the delay. I like what you are proposing @juliettepretot. Whenever there is a discussion of reuse, it reminds me of this [hilarious talk](https://vimeo.com/233817459). 
I am also a bit worried about lines like this (although I think this line can easily be removed, and be added to each manifest):
https://github.com/project-oak/oak/blob/2709157404874ad8a7f34f9b6c6f102f9d78f096/runner/src/main.rs#L408

juliettepretot: Thanks for sharing, that was a fun talk. :) 

Re the line you mentioned, that's unrelated to the manifest. The same line exists in our bash script, and thankfully it will be phased out as we switch away from bazel.

rbehjati: I think `--ca_cert=...` could be added to the `additional_args` list for each example, to make the manifests self-contained. 

It is not an issue now, especially since we are going to remove bazel. But I think in general it is better to keep the manifests self-contained. I am also thinking of these manifest files as a form of documentation and a reference for how to run the examples. Similar to what you wrote :) 

> Then in order to contribute an example, one needs to merely understand the manifest, not have a mental model of all of our build scripts.

Issue #970, opened 2020-05-11
Fix the documentation for `wait_on_channels`


rbehjati:
A more precise description of the behaviour of `wait_on_channels` is needed. 
Ref: https://github.com/project-oak/oak/pull/960/files#r422917549

Issue #918, opened 2020-04-30
Separate input and output buffers for `wait_on_channels` ABI call


tiziano88:
Currently `wait_on_channels` uses a unified buffer provided by the caller to read channels to wait on, and to write back individual channel status, interleaving 1 byte every 8 for the output.

It may be simpler to understand and to implement if instead it had an input-only buffer for the channels to wait on, and an output-only buffer in which it writes back status of the corresponding channels.

May also help with #803 .

cc @daviddrysdale

PR #1114, opened 2020-06-08
Return ChannelReadStatus from Receiver::wait

### LINKED ISSUE #1078: Return channel read status from `Receiver::wait`
Currently `Receiver::wait` only returns `ErrBadHandle` as overall status if the (only) channel that is being waited on resulted in any error. There seems to be no way for a `Receiver` user to figure out why the wait actually failed, in that case.

https://github.com/project-oak/oak/blob/474e890121fe00211cdfb4e3445100697a64a47b/sdk/rust/oak/src/io/receiver.rs#L68-L78

@rbehjati since you are most familiar with this logic, could you look into it?

rbehjati:
Fixes #1078 

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/node/wasm/mod.rs
+++ b/oak/server/rust/oak_runtime/src/node/wasm/mod.rs
@@ -421,6 +421,13 @@ impl WasmInterface {
     }
 
     /// Corresponds to the host ABI function [`wait_on_channels`](https://github.com/project-oak/oak/blob/master/docs/abi.md#wait_on_channels).
+    ///
+    /// May return the following error statuses:
+    /// - `Err(OakStatus::ErrTerminated)`: if [`Runtime`] is terminating. In this case, read
+    ///   statuses in `status_buff` do not reflect the actual read statuses.
+    /// - `Err(OakStatus::ErrInvalidArgs)`: if cannot parse `status_buff` into channel handles.
+    ///
+    /// [`Runtime`]: crate::runtime::Runtime
     fn wait_on_channels(
         &mut self,
         status_buff: AbiPointer,
@@ -435,7 +442,10 @@ impl WasmInterface {
 
         let handles_raw: Vec<u8> = self
             .get_memory()
-            .get(status_buff, handles_count as usize * 9)
+            .get(
+                status_buff,
+                handles_count as usize * oak_abi::SPACE_BYTES_PER_HANDLE,
+            )
             .map_err(|err| {
                 error!(
                     "{}: wait_on_channels(): Unable to read handles from guest memory: {:?}",
--- a/oak/server/rust/oak_runtime/src/runtime/mod.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/mod.rs
@@ -595,23 +595,23 @@ impl Runtime {
             .collect()
     }
 
-    // TODO(#970): Improve the documentation.
-    /// Waits on a slice of `ChannelHalf`s, blocking until one of the following conditions:
+    /// Given a slice of `ChannelHalf`s representing channel read handles:
     /// - If the [`Runtime`] is terminating this will return immediately with an `ErrTerminated`
-    ///   status for each channel.
-    /// - If any of the readers is in an erroneous status, e.g. when a channel is orphaned, this
+    ///   status.
+    /// - If any of the channels is in an erroneous status, e.g. when a channel is orphaned, this
     ///   will immediately return with all the channels statuses set in the returned vector.
-    /// - If any of the channels is able to read a message, the corresponding element in the
-    ///   returned vector will be set to `Ok(ReadReady)`, with `Ok(NotReady)` signaling the channel
-    ///   has no message available.
-    ///
-    /// In particular, if all channels are in a good status but no messages are available on any of
-    /// the channels (i.e., all channels have status [`ChannelReadStatus::NotReady`]),
-    /// [`Runtime::wait_on_channels`] will continue to block until a message is available on one of
-    /// the channels, or one of the channels is orphaned.
+    /// - If all channels are in a good status but no messages are available on any of the channels
+    ///   (i.e., all channels have status [`ChannelReadStatus::NotReady`]),
+    ///   [`Runtime::wait_on_channels`] blocks until a message is available on one of the channels,
+    ///   or one of the channels is orphaned. In both cases a vector of all the channels statuses
+    ///   will be returned, unless the [`Runtime`] is terminating, in which case
+    ///   `Err(ErrTerminated)` will be returned.
     ///
     /// Invariant: The returned vector of [`ChannelReadStatus`] values will be in 1-1
     /// correspondence with the passed-in vector of [`oak_abi::Handle`]s.
+    ///
+    /// See also the host ABI function
+    /// [`wait_on_channels`](https://github.com/project-oak/oak/blob/master/docs/abi.md#wait_on_channels).
     ///
     /// [`Runtime`]: crate::runtime::Runtime
     fn wait_on_channels(
--- a/sdk/rust/oak/src/io/receiver.rs
+++ b/sdk/rust/oak/src/io/receiver.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 //
 
-use crate::{io::Decodable, OakError, OakStatus, ReadHandle};
+use crate::{io::Decodable, ChannelReadStatus, OakError, OakStatus, ReadHandle};
 use serde::{Deserialize, Serialize};
 
 /// Wrapper for a handle to the read half of a channel, allowing to receive data that can be decoded
@@ -66,14 +66,34 @@ impl<T: Decodable> Receiver<T> {
     }
 
     /// Wait for a value to be available from this receiver.
+    ///
+    /// Returns [`ChannelReadStatus`] of the wrapped handle, or `Err(OakStatus::ErrTerminated)` if
+    /// the Oak Runtime is terminating,
     #[allow(clippy::trivially_copy_pass_by_ref)]
-    pub fn wait(&self) -> Result<(), OakStatus> {
+    pub fn wait(&self) -> Result<ChannelReadStatus, OakStatus> {
         // TODO(#500): Consider creating the handle notification space once and for all in `new`.
         let read_handles = vec![self.handle];
         let mut space = crate::new_handle_space(&read_handles);
         crate::prep_handle_space(&mut space);
         let status =
             unsafe { oak_abi::wait_on_channels(space.as_mut_ptr(), read_handles.len() as u32) };
-        crate::result_from_status(status as i32, ())
+
+        match crate::result_from_status(status as i32, ()) {
+            Ok(()) => space
+                .get(oak_abi::SPACE_BYTES_PER_HANDLE - 1)
+                .cloned()
+                .map(i32::from)
+                .and_then(ChannelReadStatus::from_i32)
+                .ok_or_else(|| {
+                    panic!(
+                        "wait_on_channels always yields a valid channel read status if the returned status is not Err(OakStatus::ErrTerminated)."
+                    )
+                }),
+            Err(OakStatus::ErrTerminated) => Err(OakStatus::ErrTerminated),
+            Err(OakStatus::ErrInvalidArgs) => {
+                panic!("Indicates that `space` is corrupted. This should never happen.")
+            }
+            _ => panic!("Other `OakStatus` error codes never happen."),
+        }
     }
 }
--- a/sdk/rust/oak/src/lib.rs
+++ b/sdk/rust/oak/src/lib.rs
@@ -505,23 +505,29 @@ pub fn run_event_loop<T: crate::io::Decodable, N: Node<T>>(
                         );
                         return;
                     }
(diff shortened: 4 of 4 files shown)


sdk/rust/oak/src/lib.rs
```diff
@@ -505,7 +505,7 @@ pub fn run_event_loop<T: crate::io::Decodable, N: Node<T>>(
                         );
                         return;
                     }
-                    ErrBadHandle | ErrChannelClosed | ErrPermissionDenied => {
+                    ErrBadHandle | ErrChannelClosed | ErrPermissionDenied | ErrInternal => {
```
rbehjati, line 508: @tiziano88 I don't think any of these error statuses (apart from ErrInternal) will be returned any more. Should I still keep them here for completeness?

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -596,19 +596,17 @@ impl Runtime {
     }
 
     // TODO(#970): Improve the documentation.
```
rbehjati, line 598: @tiziano88 If the modified documentation is good enough, I can remove this line and close #970. WDYT?

sdk/rust/oak/src/lib.rs
```diff
@@ -519,9 +519,17 @@ pub fn run_event_loop<T: crate::io::Decodable, N: Node<T>>(
                     }
                 }
             }
-            Ok(()) => {
-                trace!("{:?}: wait over", in_channel);
-            }
+            Ok(status) => match status {
+                ChannelReadStatus::ReadReady => trace!("{:?}: wait over", in_channel),
+                _ => {
```
rbehjati, line 524: If the status is `InvalidChannel`, `Orphaned`, or `PermissionDenied` we reach this arm. I suppose in all these cases we want to return immediately (the status will never be `NotReady` IIUC).

tiziano88: Could you enumerate them explicitly in this arm? So if we ever add a new status, we will be forced to reconsider this match statement. Also, based on your comment, it sounds like we want to log an `error` in the `NotReady` case, since I guess it would mean there is a logic error somewhere, right?

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,30 @@ impl<T: Decodable> Receiver<T> {
...
         // TODO(#500): Consider creating the handle notification space once and for all in `new`.
         let read_handles = vec![self.handle];
         let mut space = crate::new_handle_space(&read_handles);
         crate::prep_handle_space(&mut space);
         let status =
             unsafe { oak_abi::wait_on_channels(space.as_mut_ptr(), read_handles.len() as u32) };
-        crate::result_from_status(status as i32, ())
+
+        match space
+            .get(oak_abi::SPACE_BYTES_PER_HANDLE - 1)
```
tiziano88, line 84: Trying to get the byte by addressing it from the end seems harder to follow and more fragile in case the layout changes (assuming it's more likely we will add things to the end of this representation). In fact I think perhaps we should consider a proper struct (perhaps tagged with `repr(c)`) to make sure that `new_handle_space` and this logic are in sync and binary-compatible (also ideally with the Runtime side of things, but I'm not sure that's feasible. Or we should just bite the bullet and separate inputs and outputs as per https://github.com/project-oak/oak/issues/918. WDYT? cc @daviddrysdale

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,30 @@ impl<T: Decodable> Receiver<T> {
...
+            .cloned()
+            .map(i32::from)
+            .and_then(ChannelReadStatus::from_i32)
+        {
+            Some(channel_status) => Ok(channel_status),
+            None => {
+                crate::result_from_status(status as i32, ()).map(|_| Err(OakStatus::ErrInternal))?
```
tiziano88, line 91: Log the error before discarding?

sdk/rust/oak/src/lib.rs
tiziano88: If they should not happen, please capture them explicitly and log an error. If you had such insight to prompt you to write it down in this comment, why disregard this valuable information in the code? It would also make the code self-explanatory.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -596,19 +596,17 @@ impl Runtime {
...
-    /// Waits on a slice of `ChannelHalf`s, blocking until one of the following conditions:
+    /// Waits on a slice of `ChannelHalf`s, until at least one of the channels is able to read a
+    /// message, provided that all of the channels are valid; otherwise returns immediately.
```
tiziano88, line 600: IIUC it will stop waiting if a channel becomes orphan, which means that "waits until at least one of the channels is able to read a message" is not always true.

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,30 @@ impl<T: Decodable> Receiver<T> {
     }
 
     /// Wait for a value to be available from this receiver.
+    ///
+    /// May return the following error statuses:
+    /// - `Err(OakStatus::ErrTerminated)`: if the Oak Runtime is terminating.
+    /// - `Err(OakStatus::ErrInternal)`: if cannot extract the `ChannelReadStatus` form the buffer
```
conradgrobler, line 72: s/form/from/?

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,29 @@ impl<T: Decodable> Receiver<T> {
...
+        match crate::result_from_status(status as i32, ()) {
+            Err(OakStatus::ErrTerminated) => Err(OakStatus::ErrTerminated),
+            _ => match space
+                .get(oak_abi::SPACE_BYTES_PER_HANDLE - 1)
+                .cloned()
+                .map(i32::from)
+                .and_then(ChannelReadStatus::from_i32)
+            {
+                Some(channel_status) => Ok(channel_status),
+                None => unreachable!("wait_on_channels always yields a valid channel read status if the returned status is not Err(OakStatus::ErrTerminated)."),
```
tiziano88, line 90: Reading the documentation of `unreachable`, it feels like it is meant for much more self-contained impossible conditions (e.g. truly unreachable match arms, code after a loop that never terminates, etc): https://doc.rust-lang.org/std/macro.unreachable.html

I don't think it should be used for this kind of precondition checks, for which panic seems the default choice.

Alternatively we may decide to log at `error` level, since that's reserved for logic errors. I guess it is worth having a quick discussion on panic vs error on a GitHub issue, could you start one? For now, feel free to use what you prefer, it should be easy to later on grep for `error!` or `panic!` and make sure they are consistent, hopefully.

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,29 @@ impl<T: Decodable> Receiver<T> {
...
         let read_handles = vec![self.handle];
         let mut space = crate::new_handle_space(&read_handles);
         crate::prep_handle_space(&mut space);
         let status =
             unsafe { oak_abi::wait_on_channels(space.as_mut_ptr(), read_handles.len() as u32) };
-        crate::result_from_status(status as i32, ())
+
+        match crate::result_from_status(status as i32, ()) {
+            Err(OakStatus::ErrTerminated) => Err(OakStatus::ErrTerminated),
+            _ => match space
```
tiziano88, line 83: Shouldn't this pass through also `ErrInvalidArgs` and others? Either way, could you explain why some values are matched and some ignored in a comment, for future reference?

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,29 @@ impl<T: Decodable> Receiver<T> {
...
+                .get(oak_abi::SPACE_BYTES_PER_HANDLE - 1)
+                .cloned()
+                .map(i32::from)
+                .and_then(ChannelReadStatus::from_i32)
+            {
```
tiziano88, line 88: How about

```rust
..
.and_then(ChannelReadStatus::from_i32)
.ok_or_else(|| unreachable!(...))
```

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -421,6 +421,14 @@ impl WasmInterface {
     }
 
     /// Corresponds to the host ABI function [`wait_on_channels`](https://github.com/project-oak/oak/blob/master/docs/abi.md#wait_on_channels).
+    ///
+    /// May return the following error statuses:
+    /// - `Err(OakStatus::ErrTerminated)`: if [`Runtime`] is terminating. In this cases, read
```
tiziano88, line 426:
```suggestion
    /// - `Err(OakStatus::ErrTerminated)`: if [`Runtime`] is terminating. In this case, read
```

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -421,6 +421,14 @@ impl WasmInterface {
...
+    ///   statuses in `status_buff` do not reflect the actual read statuses.
+    /// - `Err(OakStatus::ErrInvalidArgs)`: if cannot parse `status_buff` into channel handles.
+    /// - `Err(OakStatus::ErrBadHandle)`: if all channels have erroneous statuses.
```
tiziano88, line 429: Just to double check (I know, I should know by now :P ): if there are 10 channels, and 9 of them are invalid and 1 ready, it returns Ok, but if all of them are invalid then it returns ErrBadHandle?

oak/server/rust/oak_runtime/src/runtime/mod.rs
tiziano88: SG, but would it make sense to actually update https://github.com/project-oak/oak/blob/master/docs/abi.md#wait_on_channels and redirect both this documentation and that on the wasm node implementation to that? Otherwise we still have 3 pieces of text saying essentially the same thing in slightly different ways.

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
rbehjati: That is true, and it sounds counterintuitive! IIRC the reason for adding the check for all invalid channels and then returning `Err(OakStatus::ErrBadHandle)` was to have an extra flag to let the caller functions more quickly detect the situation. But the caller functions may as well check the situation themselves.
That check can be removed as I have done in this [PR](https://github.com/project-oak/oak/pull/1123).

tiziano88:
BTW, would it help to add `must_use` somewhere? https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

rbehjati:
> BTW, would it help to add `must_use` somewhere? https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

Like where? Does not `clippy` already check these things? Or maybe I don't completely understand your comment.

sdk/rust/oak/src/io/receiver.rs
```diff
@@ -66,14 +66,33 @@ impl<T: Decodable> Receiver<T> {
...
         crate::prep_handle_space(&mut space);
         let status =
             unsafe { oak_abi::wait_on_channels(space.as_mut_ptr(), read_handles.len() as u32) };
-        crate::result_from_status(status as i32, ())
+
+        match crate::result_from_status(status as i32, ()) {
+            Err(OakStatus::ErrTerminated) => Err(OakStatus::ErrTerminated),
+            Err(OakStatus::ErrInvalidArgs) => panic!("Indicates that `space` is corrupted. This should never happen."),
+            // The only other output from `oak_abi::wait_on_channels` is `OakStatus::Ok`, 
+            // which is handled below. Other `OakStatus` error codes never happen.
```
tiziano88, line 85: Is it possible to add the logic that shows this is actually the case?

```rust
Err(_) => panic!("should never happen")
Ok(...) => space...
```

tiziano88:
> > BTW, would it help to add `must_use` somewhere? https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute
> 
> Like where? Does not `clippy` already check these things? Or maybe I don't completely understand your comment.

I meant like tagging ChannelReadStatus with `must_use`, so that if we forget to use it, clippy will complain.

sdk/rust/oak/src/io/receiver.rs
rbehjati: I tried that, but `clippy` does not like it:
```
error: `Err(_)` matches all errors
  --> sdk/rust/oak/src/io/receiver.rs:84:13
   |
84 |             Err(_) => panic!("Other errors never happen."),
   |             ^^^^^^
   |
   = note: `-D clippy::match-wild-err-arm` implied by `-D warnings`
   = note: match each error separately or use the error output
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_wild_err_arm
```
I can still add a separate arm for the `Ok` case.

tiziano88: Yeah that's even better!

rbehjati:
> I meant like tagging ChannelReadStatus with `must_use`, so that if we forget to use it, clippy will complain.

Interesting. I can look into that.

PR #1116, opened 2020-06-08
Install NPM deps when building an example

juliettepretot:
This reimplements #1096, which I reverted (#1103) due to permission issues in docker. 

## Background on the permissions issues

I was stuck on this for a little while, but I figured it out now.

We set up our docker image with the root user (it executes everything in the Dockerfile), but then we run it with a non root user to avoid creating permission issues in the cache folders that are shared with the docker image. (I imagine we moved these outside of the docker image to enable better caching for our GCP builds.)

So far so good. In the [`docker_run` script](https://github.com/project-oak/oak/blob/master/scripts/docker_run#L32) we pass the UID / GID of the current user to the docker image, which uses them to identify that user. That works fine, since [docker does not require a user to actually exist within its VM](https://docs.docker.com/engine/reference/run/#user). Since the user does not actually exist inside of the VM, it also does not have a `$HOME` dir. Instead that env variable resolves to the root of the docker file system. Eg `~/foo` will resolve to `/foo`. 

That's when the issue comes in when using rookie tools like npm or yarn. Basically they always create/edit caches & config files in the `$HOME` directory of the current user, and offer only limited overrides. This happens every-time one installs dependencies, such as during our `build_example` script/

In our case, because `~/` === `/` they then try to create some root level of the docker file system, which our non root user does not have the privilege of doing. Ergo things crash. 

## How this fixes things

This PR prevents those permission issues from happening by switching to the yarn node package manager, which unlike npm handles permission issues in its default cache folder by using fallbacks in the `/tmp` dir. (It's also slightly faster & [more robust](https://npm.community/t/packages-with-peerdependencies-are-incorrectly-hoisted/4794), which is nice, but not strictly needed). The remaining file that is hardcoded as being created in the `$HOME` dir (`yarnrc`) is instead provided via the mount option, so it's not actually created in the VM's root dir. 

## Alternative fixes

- We could create an actual `$HOME` directory for the non root docker user, which should avoid these issues for not just npm/yarn, but similar software that creates files in those dirs. This seemed like a larger change though, so I didn't want to do it right away. Also maybe a bit tricky, since the UID/GID of the non root user is currently different for each person that runs the [`docker_run` script](https://github.com/project-oak/oak/blob/master/scripts/docker_run#L32)

- We could `chmod 777` the relevant root dir pathes in the initial docker file, so any user will be able to use them, but I really didn't want to do that 

Some helpful background on docker permission: https://jtreminio.com/blog/running-docker-containers-as-current-host-user/

# Checklist

      construction.


juliettepretot:
Closed in favor of https://github.com/project-oak/oak/pull/1121

PR #1121, opened 2020-06-09
Create a home dir for the docker_run user, to avoid permission problems

juliettepretot:
Previously the the non-root user used in the `docker_run` script would not have its own home directory. This meant that `$HOME` or `~/` would evaluate to the root dir of the docker image, `/`. 

Since the non root user does not have write permission for its `$HOME` dir (aka the root dir), we [alias](https://github.com/project-oak/oak/blob/5d8891f380573a2f5183136288d328f67b48b2a1/scripts/docker_run#L36) all relevant paths it might write to. Eg the default bazel cache lives at `~/.cache/bazel`, which evaluates to `/.cache/bazel`, which is aliased [here](https://github.com/project-oak/oak/blob/5d8891f380573a2f5183136288d328f67b48b2a1/scripts/docker_run#L36).

Doing this worked ok so far, but doesn't work for all tools out there. For example, Node's NPM always creates some caches/config files under `~/.npm` when installing packages. And unlike previous cases, we wouldn't be able to alias the root path `/.npm` path with an outside folder, since we install NPM packages both in our build_example script and also in the root-ran Dockerfile. In the absence of a home-dir, both use the same folder. Aliasing the root `/.npm` folder would hence write root-owned files into our workspace during the docker build process, [which we don't want](https://github.com/project-oak/oak/blob/master/scripts/docker_run#L16).

This PR addresses the permissions issue, by creating an actual home-dir that is owned by the non-root user used to run our scripts in docker. It can safely write files there, meaning that new tools can be added easily without encountering permissions issues. Caches can still be aliased to outside folders, as an enhancement to enable fasters builds on GCP. 

By avoiding a class of permission issues, it should hopefully make contributing to our docker build system overall more friendly. 

## More in depth

This is a more elegant solution to the permissions issues I encountered in: #1116. If you're curious you can see my first attempt at navigating the permission issues there.

# Checklist

      construction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -196,3 +196,11 @@ RUN rustup component add \
 
 RUN cargo install cargo-deadlinks
 RUN cargo install cargo-deny
+
+ARG DOCKER_UID
+ARG DOCKER_GID
+RUN if [ ${DOCKER_GID:-0} -ne 0 ] ; then groupadd -g ${DOCKER_GID} docker_runner; fi \
+  && if [ ${DOCKER_UID:-0} -ne 0 ] ; then \
+  useradd --uid ${DOCKER_UID} --gid ${DOCKER_GID} docker_runner --home-dir /home/docker_runner --create-home \
+  && chown -R ${DOCKER_UID}:${DOCKER_GID} /home/docker_runner \
+  ;fi \
--- a/scripts/docker_run
+++ b/scripts/docker_run
@@ -24,6 +24,8 @@ mkdir -p './cargo-cache'
 docker build \
   --cache-from="$DOCKER_IMAGE_NAME:latest" \
   --tag="$DOCKER_IMAGE_NAME:latest" \
+  --build-arg DOCKER_UID="$DOCKER_UID" \
+  --build-arg DOCKER_GID="$DOCKER_GID" \
   . 1>&2
 
 docker_run_flags=(
@@ -33,7 +35,7 @@ docker_run_flags=(
   "--env=USER=$DOCKER_USER"
   '--env=BAZEL_REMOTE_CACHE_ENABLED'
   '--env=BAZEL_GOOGLE_CREDENTIALS'
-  "--volume=$PWD/bazel-cache:/.cache/bazel"
+  "--volume=$PWD/bazel-cache:/home/docker_runner/.cache/bazel"
   "--volume=$PWD/cargo-cache:/usr/local/cargo/registry"
   "--volume=$PWD:/opt/my-project"
   '--workdir=/opt/my-project'


tiziano88:
I haven't looked at the changes closely, but I suspect this will mean that it is not possible to share the Docker image across users (since in general UID will be different, and that becomes part of the image).

tiziano88:
> Afaik we always run the docker image through the `docker_run` script, which freshly builds the image each time. The changes in this PR pass the UID/GID to the build process at runtime, so it will always correspond to the current user.
> 
> After a run has finished, the image is scrapped, so no artifacts of the user that previously ran it are left over.

This is not quite accurate; we do rely on docker image caching, both locally and in CI. We also have a couple of scripts to push and pull docker images to / from Google Container Registry:

- https://github.com/project-oak/oak/blob/master/scripts/docker_pull
- https://github.com/project-oak/oak/blob/master/scripts/docker_push

In fact the first step of the CI build is to pull the `latest` image from the registry:

https://github.com/project-oak/oak/blob/19c442b3785ff2df55ceea8fd5e56bacf64bbc22/cloudbuild.yaml#L5-L9

Without this cache, I think building the Docker image from scratch every time would take substantially longer, but we can wait until cloud build runs on this PR to see how much that actually impacts the overall time.

tiziano88:
Ah nevermind, I just realised that your changes are the last step in the Docker image, so what I mentioned is technically accurate, but only for the last layer of the image, which is very thin. It does mean that that layer is different for each user, which I guess is fine as long as it always remains the last one? I wish there were a cleaner way of doing this though :/

juliettepretot:
@tiziano88 Interesting, thanks for sharing those details. :) I will look into it.

I do admit that this change is more substantial, so it's good to have a close look. If preferred I'm happy to fallback to #1116 to solve my immediate issue, but it feels a bit messier.

juliettepretot:
@tiziano88 RE your second point, that makes sense. That makes sense, relying on it being merely the last step seems a bit hacky/problematic. Permissions are hard, esp with docker. `

Technically another approach could be to always use the same non-root user for docker runs, instead of the UID/GID of the host user. This would solve caching issues, but I'm unfamiliar with other implications of that strategy.

tiziano88:
Can you just make the relevant folders world readable / writable? That seems fine, it's not like the Docker image is shared by different users anyways

juliettepretot:
@tiziano88 As in just `chmod -r 777` them? I had some hesitation to do so, but it might be smallest evil and get the job done™.

tiziano88:
We already have a precedent anyways:

https://github.com/project-oak/oak/blob/19c442b3785ff2df55ceea8fd5e56bacf64bbc22/Dockerfile#L171

juliettepretot:
@tiziano88 Oh I hadn't seen that. I shall do that then. :)

PR #1128, opened 2020-06-10
Install NPM deps when building an example

juliettepretot:
# Checklist

      construction.


Dockerfile
```diff
@@ -60,6 +60,8 @@ RUN curl --location "${bazel_url}" > bazel.deb \
 # Install Node.js and npm.
 RUN curl --location https://deb.nodesource.com/setup_12.x | bash - \
   && apt-get install --no-install-recommends --yes nodejs \
+  && mkdir "/.npm" \
+  && chmod a+rwx "/.npm" \
```
juliettepretot, line 64: Problematic fave

juliettepretot: For background, NPM expects to be able to write to `~/.npm` during installs for caching. Since for our non root user this is the root dir, we have to create & enable rwx for this dir. 

For background see #1121

PR #1135, opened 2020-06-10
Add support for format and CI steps in runner

tiziano88:
Merge BuildExamples and RunExamples under the same sub-command.

Start running without waiting for an event when using `--watch`.

Refactor `is_ignored_path` to work both with absolute and relative
paths.

Support "check" and "fix" mode for formatting.


runner/src/internal.rs
```diff
@@ -34,24 +34,11 @@ pub struct Opt {
...
     RunExamples(RunExamples),
     BuildServer(BuildServer),
+    Format,
     RunTests,
-}
-
-#[derive(StructOpt, Clone)]
-pub struct BuildExamples {
-    #[structopt(long, help = "package application in a Docker image")]
-    docker: bool,
```
juliettepretot, line 46: Do we no longer need this option? It seems like it wasn't implemented either way, but am curious if we want it

runner/src/internal.rs
```diff
@@ -62,6 +49,10 @@ pub struct RunExamples {
         default_value = "rust"
     )]
     pub application_variant: String,
+    #[structopt(long, help = "name of a single example to run")]
+    pub example_name: Option<String>,
```
juliettepretot, line 53: Interesting, so the way to run one example is to run `RunExamples`, but with an option to run actually just one. 
Similarly to build an example ones one runs `RunExamples` but uses a bool to indicate to actually just build. 

It feels a bit counterintuitive, but I assume it's done for less code duplication. 

One thing that would seem more intuitive to me would be use a `example_names` Vec, as opposed to a single string. To run a single example, one would then pass a Vec with just that entry.

runner/src/main.rs
```diff
@@ -252,28 +245,46 @@ fn build_server(opt: &BuildServer) -> Step {
...
+}
+
+fn run_ci() -> Step {
+    Step::Multiple {
+        name: "ci".to_string(),
+        steps: vec![
+            run_tests(),
+            run_examples(&RunExamples {
+                application_variant: "rust".to_string(),
+                example_name: None,
```
juliettepretot, line 283: If I understand Line 140 correctly, this means that all examples will be run?

Asking because as somebody not familiar with the code (and maybe just Rust), `example_name: None` reads a bit like "don't run any examples", but maybe that's just me. Perhaps could be improved by changing the naming to something like "only_run_example_name", but 🤷‍♀️

runner/src/main.rs
```diff
@@ -331,40 +342,54 @@ struct Client {
...
                         .collect(),
+                },
+                build_example_config(&example.name),
+                // Build the server first so that when running it in the next step it will start up
+                // faster.
+                build_server(&BuildServer {
+                    variant: "rust".to_string(),
                 }),
+            ],
+            if opt.build_only {
```
juliettepretot, line 370: Passing the whole `RunExamples` struct as arg, only to use a single key seems a bit heavy. That's since the `application_variant ` and `example_name` keys are still present but not relevant. 

Could be a bit misleading to, if `opt.example_name` were ever set to a different name than `example.name` :).

Doesn't seem ideal from an API perspective, but I'm not super picky about it, since this is an internal function and not exposed as part of the runner API.

runner/src/main.rs
```diff
@@ -331,40 +342,54 @@ struct Client {
...
+                            .map(|(name, client)| run_client(name, &client))
+                            .collect(),
+                    }),
+                }]
             },
-        ],
+        ]
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>(),
```
juliettepretot, line 392: It interesting to see assembling a Vec based on conditions is done in Rust.

In JS land, the spread operator makes that very easy.

runner/src/main.rs
```diff
@@ -146,7 +135,11 @@ fn run_examples(opt: &RunExamples) -> Step {
             /// top-level method first.
             steps: examples
                 .iter()
-                .map(|example| run_example(example))
+                .filter(|example| match &opt.example_name {
+                    Some(example_name) => &example.name == example_name,
+                    None => true,
+                })
+                .map(|example| run_example(opt, example))
```
juliettepretot, line 142: As explained in my comment on the `run_example` func, I think it might be nicer to pass just the `build_only` key, instead of the whole opt struct. :)

runner/src/main.rs
juliettepretot: (I'm aware due to the current implementation of L140, the misleading case I outlined never happens, but imho "the func API doesn't allow for misleading args" > "The func is currently never called in a misleading way")

runner/src/main.rs
tiziano88: We will also need access to the application variant from within `run_example` at some point, plus I think most other arguments to that struct (e.g. whether to compile in debug mode) so I think it is fine to just pass the entire thing. It acts kind of as a context of how to run the examples.

runner/src/main.rs
tiziano88: Yeah I agree, added a TODO on the struct field itself for now.

runner/src/main.rs
tiziano88: Let me think some more about it, perhaps I can split the struct in two and solve both this and your other point. Will follow up with a separate PR.

runner/src/main.rs
tiziano88: Yeah this is quite verbose, but this is effectively allocating a new vec composed of those parts, and Rust always makes allocations very explicit.

juliettepretot: Yeah, I figured that was the advantage of that. :)

runner/src/internal.rs
juliettepretot: Makes sense :). My intuition is always to make API as friendly to future iteration with things like that, but admittedly a lot of that comes from working in a dynamic lang, which makes changing them down the road much harder.

runner/src/main.rs
juliettepretot: That makes sense. Perhaps it's more akin to a closure, than the API of a reusable function then. :)

PR #1141, opened 2020-06-11
Add trusted CA certs to Oak runtime

### LINKED ISSUE #999: Embed root CAs in oak_loader binary
the `oak_loader` binary should not depend on external roots of trust that may affect the meaning of any Information Flow Control labels; labels are expected to have the same meaning across instances, and therefore it must not be possible to override root CAs either by specifying different flags or configuring the system differently.

We should end up with a self-contained `oak_loader` binary that includes the list of all allowed root CAs in the binary itself (and therefore contribute to its remotely attestable state by affecting its hash). In particular such root CAs are the ones that are used to establish the validity of the TLS endpoint tag:

https://github.com/project-oak/oak/blob/54e2ab272f74ef1897b377d0cc5a0dfcf8a9a35d/oak/proto/label.proto#L65-L81

@conradgrobler I think we discussed something along those lines in the past, could you look into it, and figure out what's necessary? In particular I think we need to answer the following:

- where to get an authoritative list of CAs in the first place
- how to embed it in the `oak_loader` rust binary
- how to ensure that gRPC server pseudo-node makes use of them (and only them)
- how this affects declassification of TLS endpoint tags

cc @ipetr0v

conradgrobler:
# Checklist

      construction.

Fixes #999


oak/server/rust/oak_loader/src/main.rs
```diff
@@ -199,8 +200,14 @@ fn main() -> anyhow::Result<()> {
         read_to_string(&opt.grpc_tls_private_key).context("could not read gRPC TLS private key")?;
     let grpc_tls_certificate =
         read_to_string(&opt.grpc_tls_certificate).context("could not read gRPC TLS certificate")?;
-    let root_tls_certificate =
-        read_to_string(&opt.root_tls_certificate).context("could not read root TLS certificate")?;
+    let root_tls_certificate = match &opt.root_tls_certificate {
+        Some(certificate_path) if cfg!(feature = "oak_debug") => {
+            read_to_string(certificate_path).context("could not read root TLS certificate")?
+        }
+        _ => std::str::from_utf8(include_bytes!("certs/roots.pem"))
```
ipetr0v, line 207: Can it also show an error if we specify a gRPC cert parameter without enabling `debug`?

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -199,8 +200,14 @@ fn main() -> anyhow::Result<()> {
         read_to_string(&opt.grpc_tls_private_key).context("could not read gRPC TLS private key")?;
     let grpc_tls_certificate =
         read_to_string(&opt.grpc_tls_certificate).context("could not read gRPC TLS certificate")?;
-    let root_tls_certificate =
-        read_to_string(&opt.root_tls_certificate).context("could not read root TLS certificate")?;
+    let root_tls_certificate = match &opt.root_tls_certificate {
+        Some(certificate_path) if cfg!(feature = "oak_debug") => {
```
ipetr0v, line 204: Does it mean that we can only run CI in debug?

scripts/generate_root_ca_certs
```diff
@@ -0,0 +1,37 @@
...
+
+mkdir "${EXTRACT_DIR}"
+
+tar --extract --gzip --file="${DOWNLOAD_FILE}" --directory="${EXTRACT_DIR}"
+
+files="$(find ./"${EXTRACT_DIR}"/*)"
+
+for file in ${files}; do
+    { echo "# $(openssl x509 -in "${file}" --noout --issuer)"; \
+        openssl x509 -in "${file}"; echo ""; } >> "${TARGET_FILE}"
```
ipetr0v, line 36: Could you please add comments about these commands?

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Yes.

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Done.

scripts/generate_root_ca_certs
conradgrobler: Done.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -72,7 +73,7 @@ pub struct Opt {
         long,
         help = "PEM encoded X.509 TLS root certificate file used to authenticate an external gRPC service."
     )]
-    root_tls_certificate: String,
+    root_tls_certificate: Option<String>,
```
tiziano88, line 76: Can we make it so that this field (and flag) only actually exist in the debug build? No need to get out of your way, if there is no simple way.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -199,8 +200,15 @@ fn main() -> anyhow::Result<()> {
         read_to_string(&opt.grpc_tls_private_key).context("could not read gRPC TLS private key")?;
     let grpc_tls_certificate =
         read_to_string(&opt.grpc_tls_certificate).context("could not read gRPC TLS certificate")?;
-    let root_tls_certificate =
-        read_to_string(&opt.root_tls_certificate).context("could not read root TLS certificate")?;
+    let root_tls_certificate = match &opt.root_tls_certificate {
+        Some(certificate_path) if cfg!(feature = "oak_debug") => {
+            read_to_string(certificate_path).context("could not read root TLS certificate")?
+        }
+        Some(_) => bail!("Specifying `root-tls-certificate` requires the `oak_debug` feature."),
```
tiziano88, line 207: Personally I think this would read better as an if inside a single branch, rather than as a match guard, but up to you. This catches my eye because it looks like it is ignoring something.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -199,8 +200,15 @@ fn main() -> anyhow::Result<()> {
...
+        _ => std::str::from_utf8(include_bytes!("certs/roots.pem"))
```
tiziano88, line 208: Could you replace this with `None`

scripts/generate_root_ca_certs
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+readonly SCRIPTS_DIR="$(dirname "$0")"
+# shellcheck source=scripts/common
+source "$SCRIPTS_DIR/common"
+
+# Commit that updated the certificates for Android 10.
+readonly COMMIT="be85a54767b3559e7bb262e4f1a78e1c5829e478"
+
+readonly DOWNLOAD_DIR="${SCRIPTS_DIR}/../target/tmp"
```
tiziano88, line 10: Why is this under `target`? That is the directory that cargo uses for its own artifacts. I would create a dedicated top level folder. In fact, it seems most of this logic is emulating something like git submodules, have you considered just using that? Or is the ca-certificates not a stand-alone git repo?

scripts/generate_root_ca_certs
```diff
@@ -0,0 +1,38 @@
...
+    --output "${DOWNLOAD_FILE}" \
+    --create-dirs
+
+mkdir "${EXTRACT_DIR}"
+
+tar --extract --gzip --file="${DOWNLOAD_FILE}" --directory="${EXTRACT_DIR}"
+
+files="$(find ./"${EXTRACT_DIR}"/*)"
+
+for file in ${files}; do
```
tiziano88, line 34: Could you add some comments to explain what's going on here? e.g. that this is concatenating all the certs in a single pem file one by one, and perhaps point to some documentation that says that this is fine to do (in my ignorance, I would not have expected that I can just concatenate them together like this)

oak/server/rust/oak_loader/src/main.rs
conradgrobler: I could not find an easy way to completely remove it, but I changed it so that the structopt attribute is only added when `oak_debug` is enabled. This means that the flag is not available to be set without the feature.

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Restructured it.

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Done.

scripts/generate_root_ca_certs
conradgrobler: Changed it to use a `downloads` directory and added it to `.gitignore`. I did consider using a git sub-module, but was not sure how to keep it pinned to a specific commit unless we explicitly decide to change it. As far as I understand, a 'git pull' will recursively pull new changes from sub-modules as well.

oak/server/rust/oak_loader/src/main.rs
```diff
@@ -199,8 +212,16 @@ fn main() -> anyhow::Result<()> {
...
-        read_to_string(&opt.root_tls_certificate).context("could not read root TLS certificate")?;
+    let root_tls_certificate = match &opt.root_tls_certificate {
+        Some(certificate_path) => {
+            if !cfg!(feature = "oak_debug") {
+                // It should not be possible to specify the flag without `oak_debug`.
+                bail!("Specifying `root-tls-certificate` requires the `oak_debug` feature.");
+            };
+            read_to_string(certificate_path).context("could not read root TLS certificate")?
+        }
+        _ => get_default_root_tls_certs()?,
```
tiziano88, line 223: Please use a more precise pattern rather than a wildcard whenever possible.

scripts/generate_root_ca_certs
conradgrobler: Added comments.

oak/server/rust/oak_loader/src/main.rs
conradgrobler: Done.

PR #1149, opened 2020-06-12
Integrate the Node.js client with the rust runner

juliettepretot:
# Checklist

      construction.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -331,7 +331,7 @@ struct Example {
 enum Target {
     Bazel { bazel_target: String },
     Cargo { cargo_manifest: String },
-    NPM { package_directory: String },
+    Npm { package_directory: String },
 }
 
 #[derive(serde::Deserialize, Debug)]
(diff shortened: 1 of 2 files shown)


runner/src/main.rs
```diff
@@ -331,6 +331,7 @@ struct Example {
 enum Target {
     Bazel { bazel_target: String },
     Cargo { cargo_manifest: String },
+    NPM { package_directory: String },
```
tiziano88, line 334: I don't think Rust has a clear style preference for initialisms, but I think `Npm` would be more in line with enum naming conventions here (all uppercase is usually reserved for constants). Do use the correct capitalisation in strings and log entries though.

runner/src/main.rs
```diff
@@ -413,6 +415,33 @@ fn run_client(name: &str, client: &Client) -> Step {
                 .flatten(),
             ),
         },
+        Target::NPM { package_directory } => Step::Multiple {
+            name: format!("Running the {} NPM client at {}", name, package_directory),
```
tiziano88, line 419: I would make these shorter if possible, or the indentation makes them go quickly off screen :)

juliettepretot:
@tiziano88 thanks for you comments, updated :)

Issue #1133, opened 2020-06-10
Add real database support for Trusted Information Retrieval


ipetr0v:
After adding a Trusted Information Retrieval example application https://github.com/project-oak/oak/pull/1049, we can update it, so it will be able to work with a real database, such as [Redis](https://redis.io/) or [Spanner](https://cloud.google.com/spanner).
As a first step we may use Redis since it's a simple key-value storage.

For this particular example we will consider the following assumptions:
- Database cannot fit into Oak's memory, and thus is hosted outside Oak (in the untrusted zone)
- Database can be split into chunks (i.e. first 100 entries, second 100 entries, etc.)
- Client can only request a single database entry through a gRPC request (we do not consider complex queries for now)
- Attacker can read all communications between Oak and the database (requests, responses)
- Attacker can perform network analysis
    - Can correlate database chunks requested by Oak and responses sent to clients

In order to provide privacy while being able to request entries from the database, we may implement the following scheme in the Oak application:
- Oak receives a request for a specific database entry
- It starts continuously requesting chunks of data from the database
- Processes them and finds the requested entry
- Sends a response only when all chunks have been requested from the database (even though it might have found the requested value earlier)
    - This is necessary so we will not leak information about a particular chunk that contains the requested entry, since an attacker could analyse network traffic
- [optional] Some optimizations may be applied, for example, add all requests in a queue that will share requests sent to the database (do not request all database chunks for each individual client's request)

Also, in order to connect Oak to the database we need to implement a non-Oak proxy service that will:
- Receive gRPC requests from Oak
- Redirect them into the database
- Return data entries as a gRPC response

It can also be implemented as a new pseudo-Node, but it will unnecessarily increase the TCB.

cc @project-oak/core

PR #1152, opened 2020-06-15
Add an external database to TIR example

ipetr0v:
This change:
- Adds an example external database (a Rust binary) to a TIR example.
- Adds an interface for downloading chunks of data from the database

Ref https://github.com/project-oak/oak/issues/1133

# Checklist


examples/trusted_information_retrieval/backend/src/database.rs
```diff
@@ -0,0 +1,97 @@
...
+            longitude_degrees: station.longitude_degrees,
+            installed: station.installed,
+            locked: station.locked,
+            install_date: station.install_date.to_string(),
+            removal_date: station.removal_date.to_string(),
+            temporary: station.temporary,
+            number_of_bikes: station.number_of_bikes,
+            number_of_empty_docks: station.number_of_empty_docks,
+            number_of_docks: station.number_of_docks,
+        })
```
tiziano88, line 95: May be more readable to separately implement `From` or `To` for these types.

examples/trusted_information_retrieval/backend/src/database.rs
```diff
@@ -0,0 +1,97 @@
...
+    number_of_bikes: u32,
+    #[serde(rename = "nbEmptyDocks", default)]
+    number_of_empty_docks: u32,
+    #[serde(rename = "nbDocks", default)]
+    number_of_docks: u32,
+}
+
+/// Load an XML database from `url`.
+pub async fn load_database(url: Url) -> Result<Vec<DatabaseEntry>, String> {
+    debug!("Loading database");
```
tiziano88, line 61: Print the URL too?

examples/trusted_information_retrieval/backend/src/database.rs
```diff
@@ -0,0 +1,97 @@
...
+    let response = reqwest::get(url)
+        .await
+        .map_err(|error| format!("Couldn't load database: {:?}", error))?;
```
tiziano88, line 64: I would s/load/download/ in the messages

examples/trusted_information_retrieval/backend/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+    info!("Running backend database");
+
+    let private_key = tokio::fs::read(&opt.grpc_tls_private_key).await?;
+    let certificate = tokio::fs::read(&opt.grpc_tls_certificate).await?;
+    let database_url = Url::parse(&opt.database_url).expect("Couldn't parse database URL");
+
+    // Create a mutex-protected database.
+    let database = Arc::new(Mutex::new(vec![]));
+
+    // Run a separate thread for updating the database.
```
tiziano88, line 126: I don't think `tokio` actually runs a separate thread? But I may be wrong.

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -32,22 +32,18 @@ mod proto {
...
     TrustedInformationRetrieval, TrustedInformationRetrievalDispatcher,
 };
 
-/// Oak Node that contains an in-memory database of Points Of Interest.
-pub struct TrustedInformationRetrievalNode {
-    points_of_interest: Vec<PointOfInterest>,
-}
+/// Oak Node that connects to an external database of Points Of Interest.
+#[derive(Default)]
+pub struct TrustedInformationRetrievalNode {}
```
tiziano88, line 46:
```suggestion
pub struct TrustedInformationRetrievalNode;
```

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
         // Compute distance.
         earth_radius * central_angle
     }
+
+    /// Create a gRPC client pseudo-Node.
+    fn get_grpc_client() -> Result<DatabaseClient, String> {
+        match oak::grpc::client::Client::new_with_label(
+            &oak::node_config::grpc_client("https://localhost:8888"),
```
tiziano88, line 77: This should be a config from the config map?

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
...
+            &Label::public_untrusted(),
+        )
+        .map(DatabaseClient)
+        {
+            Some(grpc_client) => Ok(grpc_client),
+            None => Err("Couldn't create a gRPC client".to_string()),
+        }
```
tiziano88, line 84: Does `.or("Couldn't create a gRPC client".to_string())` work here instead of the `match`?

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
         // Compute distance.
         earth_radius * central_angle
     }
+
+    /// Create a gRPC client pseudo-Node.
+    fn get_grpc_client() -> Result<DatabaseClient, String> {
```
tiziano88, line 75: May as well use `anyhow` here, seems more idiomatic than a String error.

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
...
+            None => Err("Couldn't create a gRPC client".to_string()),
+        }
+    }
+
+    /// Get database size through a gRPC request.
+    fn get_database_size() -> Result<u32, String> {
+        match Self::get_grpc_client()?.get_database_size(GetDatabaseSizeRequest {}) {
+            Ok(response) => Ok(response.size as u32),
+            Err(error) => Err(format!("gRPC send error: {:?}", error)),
+        }
```
tiziano88, line 92:
```rust
.map(|response| response.size as u32).or_else(|| format!("gRPC send error: {:?}", error))
```

(or better, use anyhow)

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
...
+    }
+
+    /// Load a database subset defined by the `first` and the `last` entry indices (the `last`
+    /// entry is included in the response).
+    fn list_database_entries(first: u32, last: u32) -> Result<Vec<PointOfInterest>, String> {
```
tiziano88, line 97: Could you make this `first` and `page_size`, so that it is more in line with https://cloud.google.com/apis/design/design_patterns#list_pagination ?

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +70,57 @@ impl TrustedInformationRetrievalNode {
...
+        let request = ListDatabaseEntriesRequest {
+            first_entry: first as i32,
+            last_entry: last as i32,
+        };
+        match Self::get_grpc_client()?.list_database_entries(request) {
```
tiziano88, line 102: `.map(...).or(...)`?

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -85,11 +132,24 @@ impl TrustedInformationRetrieval for TrustedInformationRetrievalNode {
     ) -> grpc::Result<ListPointsOfInterestResponse> {
         debug!("Received request: {:?}", request);
         let location = request.location.ok_or_else(|| {
-            let err = "Location is not specified".to_string();
-            error!("Bad request: {:?}", err);
-            grpc::build_status(grpc::Code::InvalidArgument, &err)
+            let error = "Location is not specified".to_string();
+            error!("Bad request: {:?}", error);
+            grpc::build_status(grpc::Code::InvalidArgument, &error)
```
tiziano88, line 137: May be clearer to build the status first, and then log it as it is (so that even the code is logged)

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
```diff
@@ -42,3 +42,41 @@ service TrustedInformationRetrieval {
...
+
+message GetDatabaseSizeRequest {}
+
+message GetDatabaseSizeResponse {
+  int32 size = 1;
+}
+
+message ListDatabaseEntriesRequest {
+  int32 first_entry = 1;
+  int32 last_entry = 2;
```
tiziano88, line 70: Same comment as above re: `first` and `page_size` https://cloud.google.com/apis/design/design_patterns#list_pagination (no need to mess around with pagination tokens for now, it would just complicate things)

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
```diff
@@ -42,3 +42,41 @@ service TrustedInformationRetrieval {
...
+  uint32 number_of_bikes = 11;
+  uint32 number_of_empty_docks = 12;
+  uint32 number_of_docks = 13;
+}
+
+message GetDatabaseSizeRequest {}
+
+message GetDatabaseSizeResponse {
+  int32 size = 1;
+}
```
tiziano88, line 66: I don't think this is needed at all if we use proper pagination

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
```diff
@@ -42,3 +42,41 @@ service TrustedInformationRetrieval {
   // Find the nearest Point Of Interest based on linear search in the database.
   rpc ListPointsOfInterest(ListPointsOfInterestRequest) returns (ListPointsOfInterestResponse);
 }
+
+message DatabaseEntry {
```
rbehjati, line 46: Can this be renamed into `BikeStation`? `DatabaseEntry` sounds too generic for the field names in this message.

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
ipetr0v: Pagination tokens will be necessary for stopping requests due to the end of the database.

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
ipetr0v: In the next PR I'll remove specific fields from this message, and will just return a serialized string (in order to create a generic reusable module).

examples/trusted_information_retrieval/backend/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+    entries: Arc<Mutex<Vec<DatabaseEntry>>>,
+}
+
+#[tonic::async_trait]
+impl Database for DatabaseService {
+    async fn get_database_size(
+        &self,
+        _request: Request<GetDatabaseSizeRequest>,
+    ) -> Result<Response<GetDatabaseSizeResponse>, Status> {
+        let entries_lock = self.entries.lock().await;
```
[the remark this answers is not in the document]
ipetr0v: Done

examples/trusted_information_retrieval/backend/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+    let certificate = tokio::fs::read(&opt.grpc_tls_certificate).await?;
+    let database_url = Url::parse(&opt.database_url).expect("Couldn't parse database URL");
+
+    // Create a mutex-protected database.
+    let database = Arc::new(Mutex::new(vec![]));
+
+    // Run a separate thread for updating the database.
+    tokio::spawn(update_database_loop(database_url, database.clone()));
+
+    let handler = DatabaseService { entries: database };
```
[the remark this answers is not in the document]
ipetr0v: It would require copying `DatabaseService` into a separate asynchronous task or using it as an `Arc<Mutex<...`.
And `entries` inside `DatabaseService` are already an `Arc<Mutex<...`. So it would add lock inside a lock.

examples/trusted_information_retrieval/module/rust/src/lib.rs
ipetr0v: Done

examples/trusted_information_retrieval/proto/trusted_information_retrieval.proto
ipetr0v: Done

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -74,6 +75,54 @@ impl TrustedInformationRetrievalNode {
...
+        .ok_or(anyhow!("Couldn't create a gRPC client"))
+    }
+
+    /// Load a database subset defined by `offset` and the number of requested elements
+    /// (`page_size`).
+    fn list_database_entries(
+        &self,
+        offset: u32,
+        page_size: u32,
+    ) -> anyhow::Result<(Vec<PointOfInterest>, bool)> {
```
tiziano88, line 95: Not sure the bool is relevant, just return things until empty or the number of entries is less than the page size.

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -85,53 +134,86 @@ impl TrustedInformationRetrieval for TrustedInformationRetrievalNode {
...
+                .map_err(|error| {
+                    let status = grpc::build_status(grpc::Code::Unavailable, "");
+                    error!("Couldn't get database entries: {:?}, {:?}", error, status);
+                    status
+                })?;
+
+            // Find the nearest point of interest.
+            nearest_point = points_of_interest
+                .iter()
+                .fold(
```
tiziano88, line 158: Would https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by_key work here?

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -85,53 +134,86 @@ impl TrustedInformationRetrieval for TrustedInformationRetrievalNode {
...
+                    (nearest_point, f32::MAX),
```
tiziano88, line 159: nearest_point vs closest_point seems a very confusing terminology, if the intention is just not to use the same name twice

examples/trusted_information_retrieval/backend/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+structopt = "*"
+tokio = { version = "*", features = [
+  "fs",
+  "macros",
+  "rt-threaded",
+  "sync",
+  "stream"
+] }
+tonic = { version = "*", features = ["tls"] }
+quick-xml = { version = "*", features = ["serialize"] }
```
tiziano88, line 27: Why do we need the serialize feature? It seems we only need to deserialize.

examples/trusted_information_retrieval/module/rust/src/lib.rs
```diff
@@ -85,53 +134,86 @@ impl TrustedInformationRetrieval for TrustedInformationRetrievalNode {
     ) -> grpc::Result<ListPointsOfInterestResponse> {
         debug!("Received request: {:?}", request);
         let location = request.location.ok_or_else(|| {
-            let err = "Location is not specified".to_string();
-            error!("Bad request: {:?}", err);
-            grpc::build_status(grpc::Code::InvalidArgument, &err)
+            let error = "Location is not specified".to_string();
+            error!("Bad request: {:?}", error);
```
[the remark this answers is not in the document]
ipetr0v: Done

examples/trusted_information_retrieval/module/rust/src/lib.rs
ipetr0v:
```
the trait `std::cmp::Ord` is not implemented for `f32`
```
`min_by_key` doesn't work for `PartialOrd`

examples/trusted_information_retrieval/backend/Cargo.toml
ipetr0v: It's for both serializing and deserializing:
https://docs.rs/quick-xml/0.18.1/quick_xml/#features

Issue #1162, opened 2020-06-16
Investigate use of sccache for CI


tiziano88:
See https://github.com/mozilla/sccache

PR #1165, opened 2020-06-17
Enable sccache for Rust compilation 

tiziano88:
Modify the volume mount points so that they match the default values in
VScode, in order to gain additional shared caching.

Ref #1162

# Checklist

      construction.


tiziano88:
This does not seem to work in all cases yet, but it's a starting point. In particular, it seems that buiding the server binaries is faster, but I have not noticed any speedup in tests, though I am not sure I am doing it right. Hopefully we can iterate on it more easily once it is checked in though.

rbehjati (approved):

PR #1166, opened 2020-06-17
Separate Database Proxy module in TIR

ipetr0v:
This change:
- Separates Database Proxy module from the main module in TIR (reusability)
- Refactors the TIR logic to key/value search instead of searching for the closest point (generalization)

Ref https://github.com/project-oak/oak/issues/1133

# Checklist


examples/trusted_information_retrieval/backend/build.rs
```diff
@@ -18,15 +18,17 @@ use std::path::Path;
...
 
-    tonic_build::configure()
-        .build_client(false)
-        .build_server(true)
-        .compile(&[file_path.as_path()], &[proto_path])?;
+        tonic_build::configure()
+            .build_client(false)
+            .build_server(true)
+            .compile(&[file_path.as_path()], &[proto_path])?;
+    }
```
tiziano88, line 32: Why was this changed? Is it not possible to compile them at once any more? Please add enough comments for future maintainers to figure out what's going on here.

examples/trusted_information_retrieval/backend/src/main.rs
```diff
@@ -76,19 +78,37 @@ impl Database for DatabaseService {
...
+                .map(|entry| {
+                    // Serialize database entry in a String.
+                    let mut serialized_entry = Vec::new();
+                    entry.encode(&mut serialized_entry).unwrap();
+                    DatabaseEntry {
+                        key: entry.id.to_string(),
+                        value: String::from_utf8(serialized_entry).unwrap_or_else(|error| {
+                            let error = format!("Data entry cannot be serialized: {:?}", error);
+                            warn!("{}", error);
+                            error
```
tiziano88, line 100: Why is this logging **and** returning the same error?

examples/trusted_information_retrieval/backend/src/tests.rs
```diff
@@ -56,7 +56,7 @@ fn test_parse_database() {
     assert_matches!(database, Ok(_));
     let database = database.unwrap();
     assert_eq!(database.len(), 2);
-    assert_eq!(database[0].name, "Opened station 1");
+    assert_eq!(database[0].street_name, "Opened station 1");
```
tiziano88, line 59: This does not read as a street name, does anything need to change?

examples/trusted_information_retrieval/client/trusted_information_retrieval.cc
```diff
@@ -27,33 +27,28 @@
 
 ABSL_FLAG(std::string, address, "localhost:8080", "Address of the Oak application to connect to");
 ABSL_FLAG(std::string, ca_cert, "", "Path to the PEM-encoded CA root certificate");
-ABSL_FLAG(float, latitude, float{}, "Requested location's latitude in degrees (WGS84)");
-ABSL_FLAG(float, longitude, float{}, "Requested location's longitude in degrees (WGS84)");
+ABSL_FLAG(std::string, id, "", "Database ID of the point of interest");
```
tiziano88, line 30: What's a database id?

examples/trusted_information_retrieval/client/trusted_information_retrieval.cc
```diff
@@ -27,33 +27,28 @@
...
-  ListPointsOfInterestResponse response;
-  grpc::Status status = stub->ListPointsOfInterest(&context, request, &response);
+  LOG(INFO) << "Getting point of interest:";
+  GetPointOfInterestRequest request;
+  request.set_id(id);
+  GetPointOfInterestResponse response;
+  grpc::Status status = stub->GetPointOfInterest(&context, request, &response);
   if (!status.ok()) {
-    LOG(ERROR) << "Could not get nearest point of interest: " << status.error_code() << ": "
+    LOG(ERROR) << "Could not get point of interest: " << status.error_code() << ": "
```
tiziano88, line 45: Should this be FATAL? (I think I made this same change several times already :P )

examples/trusted_information_retrieval/module_0/rust/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+        &mut self,
+        request: GetPointOfInterestRequest,
+    ) -> grpc::Result<GetPointOfInterestResponse> {
+        debug!("Received request: {:?}", request);
+
+        // Get requested database entry from the database proxy Node.
+        let entry = get_database_entry(&request.id, &self.database_url).map_err(|error| {
+            let status = grpc::build_status(grpc::Code::Unavailable, "");
+            error!("Couldn't get point of interest: {:?}, {:?}", error, status);
+            status
```
tiziano88, line 66: Why Unavailable? Is the intention that the client should just retry? Maybe add some comments.

examples/trusted_information_retrieval/module_0/rust/src/tests.rs
```diff
@@ -0,0 +1,30 @@
...
+use assert_matches::assert_matches;
+
+const CONFIG_FILE: &str = r#"
+database_url = "https://localhost:8888"
+"#;
+
+#[test]
+fn test_parse_config() {
+    let config = parse_config_file(CONFIG_FILE.as_bytes().to_vec());
+    assert_matches!(config, Ok(_));
```
tiziano88, line 27: It seems you know exactly what the result should be, may as well inline everything.

```
let parsed = ...
let expected = Database {
  ...
};
assert_eq!(Ok(expected)), parsed)
```

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+pub fn get_database_entry(id: &str, database_url: &str) -> grpc::Result<DatabaseEntry> {
+    let client = oak::grpc::client::Client::new_with_label(
+        &oak::node_config::wasm("database_proxy", "database_proxy_main"),
+        &Label::public_untrusted(),
+    )
+    .map(DatabaseProxyClient)
+    .ok_or_else(|| {
+        grpc::build_status(
+            grpc::Code::Unavailable,
+            "Couldn't create a database proxy Node",
```
tiziano88, line 57:
```suggestion
            "Couldn't create database proxy Node",
```

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    })?;
+
+    let request = GetDatabaseEntryRequest {
+        key: id.to_string(),
+        database_url: database_url.to_string(),
+    };
+    client.get_database_entry(request).and_then(|response| {
+        response
+            .entry
+            .ok_or_else(|| grpc::build_status(grpc::Code::NotFound, "Empty response"))
```
tiziano88, line 68: "Empty response" seems quite mysterious if I got it back as an error over gRPC

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+        database_url: &str,
+    ) -> anyhow::Result<Vec<DatabaseEntry>> {
+        let request = ListDatabaseEntriesRequest {
+            offset: offset as i32,
+            page_size: page_size as i32,
+        };
+        Self::get_grpc_client(database_url)?
+            .list_database_entries(request)
+            .map(|response| response.entries)
+            .map_err(|error| anyhow!("gRPC send error: {:?}", error))
```
tiziano88, line 101: does `.context` not work here?

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+/// A gRPC service implementation for the database proxy Node.
+impl DatabaseProxy for DatabaseProxyNode {
+    fn get_database_entry(
+        &mut self,
+        request: GetDatabaseEntryRequest,
+    ) -> grpc::Result<GetDatabaseEntryResponse> {
+        // Request entries from the database until the last entry page is reached.
```
tiziano88, line 111: This definitely needs a much more detailed explanation.

Why is the logic implemented like this?

Why does the loop not break immediately once the correct entry is found?

What are the invariants that need to be maintained?

Is it fine to log stuff in this loop?

etc.

ipetr0v: General explanation of the idea is at the top of the file.
Added a comment about a loop.

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+            if entries.len() < DATABASE_PAGE_SIZE as usize {
+                break;
+            }
+            current_offset += DATABASE_PAGE_SIZE;
+        }
+
+        requested_entry
+            .map(|entry| GetDatabaseEntryResponse { entry: Some(entry) })
+            .ok_or_else(|| grpc::build_status(grpc::Code::NotFound, ""))
+    }
```
tiziano88, line 147: IIUC, this will return an error to the caller, but not log anything in the server, which may make debugging hard / impossible. Or does it get logged somewhere else?

examples/trusted_information_retrieval/proto/database.proto
```diff
@@ -0,0 +1,59 @@
...
+  int32 page_size = 2;
+}
+
+message ListDatabaseEntriesResponse {
+  repeated DatabaseEntry entries = 1;
+}
+
+// Service definition used by the backend database which Oak Node connects to.
+service Database {
+  // Get a subset of entries from the database based on the offset and the number of entries.
```
tiziano88, line 57: I think this could do with a bit more explanation: what subset? Are entries assumed to be sorted? If so, by what? What if more entries are requested than available?

examples/trusted_information_retrieval/module_1/rust/src/lib.rs
ipetr0v: I think so. it would require reimplementing this for `rpc::Status`:
https://github.com/dtolnay/anyhow/blob/55343b9d6cb8fe12486b916fcc1e1588dc4cef11/src/context.rs#L42-L60

PR #1175, opened 2020-06-19
Move NPM build step inside of client variant logic inside of build_example

juliettepretot:
# Checklist

      construction.

### DIFF
--- a/scripts/build_example
+++ b/scripts/build_example
@@ -5,13 +5,13 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 source "${SCRIPTS_DIR}/common"
 
 application_language="rust"
-client_variant="bazel"
+client_variants="bazel"
 compilation_mode='fastbuild'
 docker_config=''
 while getopts "e:a:c:di:h" opt; do
   case "${opt}" in
     h)
-      echo -e "Usage: ${0} [-h] [-a rust|cpp] [-c cargo|bazel] [-i base|logless] [-d] -e EXAMPLE
+      echo -e "Usage: ${0} [-h] [-a rust|cpp] [-c [cargo|bazel|npm]...] [-i base|logless] [-d] -e EXAMPLE
 
 Build the given example Oak Application and client.
 
@@ -20,9 +20,11 @@ Options:
   -a    Example application variant:
           - rust (used by default)
           - cpp
-  -c    Example client variant:
-          - cargo
-          - bazel (used by default)
+  -c    Example client variants. Multiple clients can be specified by space
+        separating them. Eg -c \"cargo bazel npm\".
+        - cargo
+        - bazel (used by default)
+        - npm
   -d    Build C++ code for example using debug mode
   -i    This flag enables packaging the application into a Docker image,
         and specifies the version of the Oak server, used by the application:
@@ -35,7 +37,7 @@ Options:
     a)
       application_language="${OPTARG}";;
     c)
-      client_variant="${OPTARG}";;
+      client_variants="${OPTARG}";;
     d)
       compilation_mode='dbg';;
     i)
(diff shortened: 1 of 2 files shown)


scripts/build_example
```diff
@@ -5,13 +5,13 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
...
 application_language="rust"
-client_language="cpp"
+client_variant="bazel"
 compilation_mode='fastbuild'
 docker_config=''
 while getopts "e:a:c:di:h" opt; do
   case "${opt}" in
     h)
-      echo -e "Usage: ${0} [-h] [-a rust|cpp] [-c rust|cpp] [-i base|logless] [-d] -e EXAMPLE
+      echo -e "Usage: ${0} [-h] [-a rust|cpp] [-c cargo|bazel] [-i base|logless] [-d] -e EXAMPLE
```
anghelcovici, line 14: Looks like there is a npm build step at line 122, why not include npm in the help text?

scripts/run_examples
```diff
@@ -46,12 +46,12 @@ done
...
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e aggregator -- --bucket=test --data=1:10,2:20,3:30
   elif [[ "${example}" == 'chat' ]]; then
-      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_languages}" -e chat -- --test
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e chat -- --test
   elif [[ "${example}" == 'trusted_information_retrieval' ]]; then
-      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_languages}" -e trusted_information_retrieval -- --latitude 0.0 --longitude 0.0
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e trusted_information_retrieval -- --latitude 0.0 --longitude 0.0
   else
-      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_languages}" -e "${example}"
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e "${example}"
```
anghelcovici, line 55: Can we maybe have one variable for `"${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e "${example}"` and only add a suffix for each of the examples that need it? There is a lot of copy pasting here :P

scripts/run_example
```diff
@@ -110,37 +110,37 @@ fi
...
       fi
       ;;
-    cpp)
-      cpp_client="./bazel-client-bin/examples/${EXAMPLE}/client/client"
-      if [[ -f "${cpp_client}" ]]; then
-          "${cpp_client}" "${TLS_ARGS[@]}" "${ADDITIONAL_ARGS[@]-}" "${CLIENT_ARGS[@]-}"
+    bazel)
+      build_file="${SCRIPTS_DIR}/../examples/${EXAMPLE}/client/BUILD"
+      if [[ -f "${build_file}" ]]; then
+          "${SCRIPTS_DIR}/../bazel-client-bin/examples/${EXAMPLE}/client/client" "${TLS_ARGS[@]}" "${ADDITIONAL_ARGS[@]-}" "${CLIENT_ARGS[@]-}"
```
anghelcovici, line 129: Probably outside of scope of this review, but we should probably use `bazel run ....` and let it figure out paths instead of manually pointing to the file.

juliettepretot: Agreed. Though given I heard we also want to migrate away from bazel, I'm not sure if we want to invest more effort into our bazel code rn. :)

juliettepretot:
@anghelcovici 

Thanks for your comment! Most of the code in this draft PR (the first commit) actually was PR'ed in https://github.com/project-oak/oak/pull/1170 & merged earlier. 

I left this PR in draft as further work, but it's not fully ready yet. The aspect this PR will address specifically is into the client variant in the build_example step. :)

scripts/build_example
juliettepretot: Added that. :) For now I've been using this draft PR as a way to run/test CI in the background. 

We only conditionally call the `build_example` step inside of run_example, and I wanted to check which things would break if I moved `npm` into the variant, but it's not ready yet.

scripts/run_examples
juliettepretot: Could be an improvement yeah. But on a higher level, I don't think that using a lot conditionals inside of the run script is such a maintainable thing to begin with. 

Ideally the example should specify in one place which args it needs, instead of our bash script conditionally checking the example name to insert them. That's what I like about the manifest approach taken in the rust runner. 

Spoke to @daviddrysdale earlier, and it might make sense to use a similar if not same manifest in bash. It comes in handy to store other relevant data about an example (eg which clients it offers) which we currently have to deduct based on folder structure. See: https://github.com/project-oak/oak/pull/1169

PR #1176, opened 2020-06-19
Always run all clients in run_examples

juliettepretot:
Replaces the current way of deducting available clients from the folder structure (#1148) with instead explicitly specifying them in the conditionals of the `run_examples` script. 

This is nice, because it allows us to explicitly fail if a client is not avaible.

# Checklist

      construction.

### DIFF
--- a/scripts/run_example
+++ b/scripts/run_example
@@ -115,29 +115,16 @@ do
   case "${client_variant}" in
     cargo)
       rust_client_manifest="${SCRIPTS_DIR}/../examples/${EXAMPLE}/client/rust/Cargo.toml"
-      if [[ -f "${rust_client_manifest}" ]]; then
           cargo run --release --manifest-path="${rust_client_manifest}" -- \
             --root-tls-certificate="${SCRIPTS_DIR}/../examples/certs/local/ca.pem" \
             "${CLIENT_ARGS[@]-}"
-      else
-        echo "The ${EXAMPLE} example does not contain a ${client_variant} client. Skipping this client."
-      fi
       ;;
     bazel)
-      build_file="${SCRIPTS_DIR}/../examples/${EXAMPLE}/client/BUILD"
-      if [[ -f "${build_file}" ]]; then
           "${SCRIPTS_DIR}/../bazel-client-bin/examples/${EXAMPLE}/client/client" "${TLS_ARGS[@]}" "${ADDITIONAL_ARGS[@]-}" "${CLIENT_ARGS[@]-}"
-      else
-        echo "The ${EXAMPLE} example does not contain a ${client_variant} client. Skipping this client."
-      fi
       ;;
     npm)
       npm_client="${SCRIPTS_DIR}/../examples/${EXAMPLE}/client/nodejs"
-      if [[ -d "${npm_client}" ]]; then
           npm start --prefix "${npm_client}"
-      else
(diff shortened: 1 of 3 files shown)


scripts/run_examples
```diff
@@ -46,12 +41,14 @@ done
...
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -e chat -- --test
   elif [[ "${example}" == 'trusted_information_retrieval' ]]; then
-      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e trusted_information_retrieval -- --latitude 0.0 --longitude 0.0
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "bazel cargo" -e trusted_information_retrieval -- --latitude 0.0 --longitude 0.0
+  elif [[ "${example}" == 'hello_world' ]]; then
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "bazel npm" -e "${example}"
   else
-      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -c "${client_variants}" -e "${example}"
+      "${SCRIPTS_DIR}/run_example" -s "${server}" -a "${application_language}" -e "${example}"
   fi
```
juliettepretot, line 53: Personally I think keeping a kind of manifest is preferable over all these conditionals. 

However given that we still haven't decided on the bash vs rust question of runner scripts, I don't want to put too much effort into refactoring the bash scripts to remove redundancy.

scripts/run_example
```diff
@@ -115,29 +115,16 @@ do
...
     npm)
       npm_client="${SCRIPTS_DIR}/../examples/${EXAMPLE}/client/nodejs"
-      if [[ -d "${npm_client}" ]]; then
           npm start --prefix "${npm_client}"
-      else
-        echo "The ${EXAMPLE} example does not contain a ${client_variant} client. Skipping this client."
-      fi
       ;;
     *)
       echo "Invalid example client variant: ${client_variant}"
```
ipetr0v, line 130: Which example variant will be run, if nothing is specified in the `-c`?
Will it just be Bazel?
https://github.com/project-oak/oak/blob/e9db4c6f4b342bc6066fccb71104210b4984dbb3/scripts/run_example#L9

scripts/run_examples
tiziano88: I note that (to me) this seems a good argument for the Rust + manifest, so feel free to weigh in on the discussion :) I would like it to be driven by use cases, and this seems a good one to me

scripts/run_example
juliettepretot: Yup, since it's the default. :) https://github.com/project-oak/oak/blob/018389a2f8c1080be13eea0ec248747dd366505f/scripts/run_example#L32

We can change that if we'd like to run all available example by default, but that seems like a separate task :)

scripts/run_examples
juliettepretot: Agreed. And as [I mentioned earlier](https://github.com/project-oak/oak/pull/1105#discussion_r438030248), replacing these scattered conditionals is what appeals to me most about the manifest approach. 

Though I suppose while Rust makes it easier to maintain complex logic such manifests, one could always argue that theoretically the same pattern could be implemented in bash. :)

Issue #1029, opened 2020-05-26
defense-in-depth configuration of `oak_loader`


tiziano88:
After #917 is submitted, it will not be possible to rely on manual inspection of an application configuration file in order to determine the upper bound of the kinds of nodes that an Oak Application may create.

Several levels of defense-in-depth mechanisms are desirable in production systems, in order to prevent malicious actors from being able to manipulate the system so that it act outside of the boundaries that the developer / deployer initially intended.

For `oak_loader`, we should support (at least) the following **permissions** (as a whitelist, passed in as command line flags or as part of some config file):

- logging
- whitelisted TLS domains / IP addresses for egress traffic (gRPC and HTTPS)
- "exposed" files from host (partly implemented in #1004)

cc @daviddrysdale

PR #1198, opened 2020-06-25
Move helper functions in oak_loader to a separate submodule

rbehjati:
A bit of refactoring before starting with #1029

### DIFF
--- a/oak/server/rust/oak_loader/src/main.rs
+++ b/oak/server/rust/oak_loader/src/main.rs
@@ -26,123 +26,18 @@
 //!     --root-tls-certificate=<CERTIFICATE_PATH>
 //! ```
 
-use anyhow::{anyhow, bail, Context};
-use core::str::FromStr;
-use log::{debug, info};
-use oak_abi::proto::oak::application::{ApplicationConfiguration, ConfigMap};
-use oak_runtime::{
-    auth::oidc_utils::parse_client_info_json,
-    config::{configure_and_run, load_certificate},
+use anyhow::Context;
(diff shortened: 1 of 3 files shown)


oak/server/rust/oak_loader/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+    collections::HashMap,
+    fs::{read, read_to_string},
+};
+use structopt::StructOpt;
+use tonic::transport::Identity;
+
+/// Command line options for the Oak loader.
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(about = "Oak Loader")]
+pub struct Opt {
```
conradgrobler, line 39: I am not sure what the reason is for moving this to separate library crate rather than a submodule.

If the plan is to reuse it from other binaries, it might be better to not include the command-line arguments. Command-line flags used in libraries often lead to conflicts or confusion, especially if multiple libraries try to declare their own flags.

If there are no plans for reuse, it might be better to just move it to a submodule rather that a separate library crate.

tiziano88: I am also unsure why this was all moved to a lib.rs file, what's the intention behind it @rbehjati ?

conradgrobler: Yes, I think a submodule is better. Perhaps also change the title of the PR to remain consistent?

rbehjati: Good point. Thanks.

PR #1214, opened 2020-06-30
Remove Bazel dependency for serialiser and docker images

### LINKED ISSUE #1058: Run Config Serializer without Bazel
After removing C++ from our codebase (https://github.com/project-oak/oak/pull/1016) we are still using Bazel to create a serialized application configuration:
https://github.com/project-oak/oak/blob/9f5d57ffe2887bdfd0b5bdede5fb4404fcfdd735/examples/hello_world/config/BUILD#L26-L33

It could be rewritten as a `bash` script.

Originated from https://project-oak.slack.com/archives/CHE9E13C3/p1591028897061100

anghelcovici:
Fixes #1058. I think this is the first step towards replacing the whole serialiser with a Rust version. I also want to make it a bit more self contained, but that is for the next version.

# Checklist

      construction.


examples/abitest/config/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+readonly MODULES_DIR="${DIR}/../../target/wasm32-unknown-unknown/release"
+
+bazel run //oak/common:app_config_serializer -- \
+    --textproto="${DIR}/config.textproto"\
+    --modules="frontend_module:${MODULES_DIR}/abitest_0_frontend.wasm","backend_module:${MODULES_DIR}/abitest_1_backend.wasm" \
+    --output_file="${DIR}/config.bin"
```
tiziano88, line 9: I think we should probably start considering a dedicated output directory for build artifacts (e.g. `/out`?) Otherwise it's pretty hard to figure out what's source and what's generated (sure, the gitignore file gives you some ideas, but it's not great). So this may be `/out/abitest.bin`, which seems also a bit more self-explanatory.

examples/abitest/config/build.sh
```diff
@@ -0,0 +1,10 @@
...
+
```
tiziano88, line 10: Please add proper newlines at EOF.

examples/abitest/config/build.sh
anghelcovici: Good point, I was thinking about it. We already have a billion output directories and out sounds generic, but I'll try to find something that makes sense

anghelcovici: @tiziano88 - the other annoying thing is that afaict, Cargo doesn't allow you to specify a different path for the output (or a link to that). There are a couple of in progress issues, and an experimental feature, but otherwise it sucks. It would be nice to have a /bin/ folder for each example only with the binary artefacts needed to run it

Dockerfile
```diff
@@ -31,13 +46,23 @@ RUN apt-get --yes update \
...
+  # Cleanup
   && apt-get clean \
   && rm --recursive --force /var/lib/apt/lists/* \
   # Print version of various installed tools.
   && git --version \
-  && clang-format -version \
   && shellcheck --version
 
+
+RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
```
tiziano88, line 57: What are all these flags to `curl`? Could you expand them, so we don't have to remind ourselves what's going on?

examples/abitest/config/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+readonly MODULES_DIR="${DIR}/../../target/wasm32-unknown-unknown/release"
+readonly OUT_DIR="${DIR}/../bin"
```
tiziano88, line 5: Are we settling on `bin` as the output directory? Why not... `out`? No need to change it anyways, we'll see later on.

cc @ipetr0v

examples/aggregator/config/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+readonly MODULES_DIR="${DIR}/../../target/wasm32-unknown-unknown/release"
+readonly OUT_DIR="${DIR}/../bin"
+mkdir -p "${OUT_DIR}"
```
tiziano88, line 6:
```suggestion
mkdir --parents "${OUT_DIR}"
```

examples/aggregator/config/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+readonly DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
tiziano88, line 3: No idea what's going on here, is it trying to resolve its own location?

Could it be a function defined in `scripts/common`?

examples/aggregator/config/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
tiziano88, line 2: Could you also source `scripts/common` here, or is it too complicated?

examples/aggregator/server/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM gcr.io/distroless/cc@sha256:f81e5db8287d66b012d874a6f7fea8da5b96d9cc509aa5a9b5d095a604d4bca1
+
+COPY //:examples/target/x86_64-unknown-linux-gnu/release/aggregator_backend .
```
tiziano88, line 3: What does this `//:` syntax mean?

examples/aggregator/server/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+EXPOSE 8888/tcp
```
tiziano88, line 4: The oak server Dockerfile has `EXPOSE` after `ENTRYPOINT` (which also reads more naturally I think), perhaps make this consistent?

scripts/docker_run
```diff
@@ -17,6 +17,7 @@ source "$SCRIPTS_DIR/common"
 readonly DOCKER_UID="${UID:-0}"
 readonly DOCKER_GID="${GID:-0}"
 readonly DOCKER_USER="${USER:-root}"
+readonly DOCKER_GROUP="$(getent group docker | cut -d: -f3)"
```
tiziano88, line 20: AFAICT this is literally the group of the `docker` user, but the other `DOCKER_*` vars actually refer to the user and group of the user outside of docker, so that it can be remapped inside. Not sure it's worth renaming things, but at least please make it clear in a comment, I can definitely see myself trying to figure this out again in the future :)

examples/abitest/config/build.sh
anghelcovici: I like the name bin better, as it is a place where we dump binary artefacts. Out seems... strange.

examples/aggregator/config/build.sh
anghelcovici: Maybe... it's apparently the safest bet to get the current folder, but it is not consistent with other places, so I'll change it.

examples/aggregator/server/Dockerfile
anghelcovici: That is wrong, not sure why it doesn't complain. good catch

scripts/docker_run
anghelcovici: It is the outside docker group id. Yeah, naming is confusing, I'll change it.

examples/aggregator/config/build.sh
anghelcovici: It's not, but afaict, these bash script will be there for a very short time (as @ipetr0v  is replacing them with toml files). @ipetr0v, is that correct?

examples/aggregator/config/build.sh
anghelcovici: It can... but this is a bit different. As we source this, I wanted to get the folder of the script, not of the location where this is called. the method we use in most of the scripts works, as we mostly call them directly, not via `source`.  I think it's a somehow unique usecase. As the comment below, I wouldn't spend too much time on this, as I do think we should remove these files soon and replace them with `.toml`s

.gitignore
```diff
@@ -35,6 +35,9 @@
 # Ignore downloaded temporary files, such as the trusted root CA certificates.
 /downloads/
 
+# Ignore generated binary example files
+**/examples/**/bin
```
ipetr0v, line 39: Is this directory used to store Wasm files and serialized application configs?

Dockerfile
```diff
@@ -7,14 +7,29 @@ FROM debian/snapshot:${debian_snapshot}
...
 
+# Uncomment the RUN below if the default snapshot package manager is slow.
+# Please not that this might cause issues and affects reproducible builds,
+# so only use for development.
+# RUN echo \
+#  deb [arch=amd64] http://ukdebian.mirror.anlx.net/debian buster main non-free contrib\
+# > /etc/apt/sources.list
+
+# Getting curl and certificates dependecies.
+RUN apt-get --yes update \
```
ipetr0v, line 18: Why is it a separate `apt-get` call?

examples/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM oak_docker:latest
+
+# This runs in the context of each example, so we just copy the bin
+COPY bin/config.bin .
```
ipetr0v, line 4: Which example does it copy?
Or it is being run sequentially for every example?
Is `bin/config.bin` rewritten every time a new example is built?

examples/abitest/config/build.sh
ipetr0v: Is there an out directory for `oak_loader`?

examples/abitest/config/build.sh
```diff
@@ -0,0 +1,11 @@
...
+mkdir --parents "${OUT_DIR}"
+
+bazel run //oak/common:app_config_serializer -- \
```
ipetr0v, line 8: So this PR doesn't remove Bazel for building examples.
Why was it moved to `build.sh`?

examples/aggregator/config/build.sh
ipetr0v: Yes, but `app_config_serializer` is the only thing that changes in https://github.com/project-oak/oak/pull/1219.
Examples are still compiled from `scripts/build_example`

scripts/run_clang_tidy
```diff
@@ -8,6 +8,7 @@ bazel_build_flags+=(
   '--experimental_action_listener=@io_kythe//kythe/cxx/tools/generate_compile_commands:extract_json'
   '--noshow_progress'
   '--noshow_loading_progress'
+  '--experimental_extra_action_top_level_only'
```
ipetr0v, line 11: What does it do?

.gitignore
anghelcovici: Yes

tiziano88: I think we should just pick a directory name (`bin` is fine, `out` was also suggested) and use it to generate everything everywhere (not just examples) (including binaries, wasm modules, additional stuff), rather than trying to get things out from `target/wasm32-unknown-unknown` and similar.

For `cargo build`, there is an `--out-dir` flag that should do that already.

Dockerfile
anghelcovici: No good reason. I thought is going to optimise docker runs, but if the execution doesn't complete it doesn't save an intermediate step. I will merge it.

examples/Dockerfile
anghelcovici: Docker has a "context" in which is ran. You can think of it as the "parent folder". For each example image, I use the folder examples/EXAMPLE_NAME as the context, so bin/config.bin is the file for that e.g. examples/hello_world/bin/config.bin

scripts/build_example
```diff
@@ -95,15 +94,16 @@ esac
 
 if [[ -n "${docker_config}" ]]; then
   "${SCRIPTS_DIR}/build_server" -s "${docker_config}"
-  if [[ "${EXAMPLE}" == "tensorflow" ]]; then
-    bazel --output_base="${CACHE_DIR}/emscripten" build "${bazel_build_flags[@]}" --config=emscripten //examples/tensorflow/server:tensorflow_image.tar
-  else
-    bazel build "${bazel_build_flags[@]}" "--config=clang" "//examples/${EXAMPLE}/server:${EXAMPLE}_image.tar"
-  fi
+  # Build the base server.
+  docker build -t oak_docker -f ./oak/server/Dockerfile ./oak/server
```
tiziano88, line 98:
```suggestion
  docker build --tag=oak_docker --file=./oak/server/Dockerfile ./oak/server
```

scripts/build_example
```diff
@@ -95,15 +94,16 @@ esac
...
+  # Build and save the example
+  docker build -t "${EXAMPLE}_image" -f ./examples/Dockerfile "./examples/${EXAMPLE}"
```
tiziano88, line 100: Why the `_image` suffix? All docker images are... images.

scripts/build_example
```diff
@@ -95,15 +94,16 @@ esac
...
+  docker save "${EXAMPLE}_image" > "./examples/${EXAMPLE}/bin/${EXAMPLE}_image.tar"
```
tiziano88, line 101: What is this step for?

examples/abitest/config/build.sh
anghelcovici: It's about the linked issue, which is removing Bazel from the invocation of the app_serializer. It is described in the first comment, but not reflected in the title, so I will change that. I think we will always build examples with Bazel (c++ ones), we don't have a goal to remove that.

examples/Dockerfile
tiziano88: It does read a bit weird though. I wonder whether having an explicit `ARG config_file` which is passed as `docker build --build-arg=config_file=...` would be more readable.

scripts/run_clang_tidy
anghelcovici: clang_tidy was not working, not really related to the PR. I am still not 100% sure why, but the "hack-ish" experimental code was trying to compile all bazel targets and getting into some GO code that it didn't configure well. This keeps it to only the first level of targets, which is exactly what we need.

examples/abitest/config/build.sh
anghelcovici: No, this is for the examples. I don't think the `oak_loader` should be there

examples/Dockerfile
anghelcovici: What does read a bit weird even mean? This is how the Dockerfiles work... maybe the comment should be different?

examples/abitest/config/build.sh
tiziano88: I think we should have it as `oak/server/bin/oak_loader` at some point.

examples/Dockerfile
tiziano88: The weirdness I was referring to is that usually a Docker image is built from a known location, but here you are expecting to move it around and be able to rebuild it in other folders.

scripts/build_example
anghelcovici: We first build the server image, on top of that we put the "application" and we save it so we can later "push it". I am not really sure we need the .tar step, but @ipetr0v almost convinced me in a slack chat :)

scripts/build_example
anghelcovici: I am just reusing the existing names. Happy to change it, but I think we use it in a lot of places. cc: @ipetr0v - any reason for using _image?

examples/abitest/config/build.sh
anghelcovici: +1 to @tiziano88  suggestion

.gitignore
anghelcovici: Just pointing out that I believe --out-dir is still experimental

examples/Dockerfile
anghelcovici: We are building an image of an example, so it kind make sense.. but I do see where you are coming from.

scripts/build_example
anghelcovici: It's actually quite interesting. We could, potentially, have 1 single image, but the issue is that the Docker context would need to be the first common parent directory of the example and server, which is oak top-most directory. And because we have a .dockerignore there for the "overall" image, that conflicts with this. So this is why I made it a two step image. It's also nicer, since we can ideally build the server image once and add to it for each example

scripts/build_example
ipetr0v: Since it's an example that is being loaded by Oak server inside Docker, it also could be `"${EXAMPLE}_server"`.
But just `"${EXAMPLE}"` is also fine.

PR #1217, opened 2020-06-30
Use brackets in bash variables

ipetr0v:
This change adds brackets in bash variables.


tiziano88:
Why didn't this use quotes in the end?

ipetr0v:
Because `client_variants` is a string, not an array, and quotes made it `for` over a single string.

PR #1221, opened 2020-07-02
Use existing binaries for Rust tools, where possible

anghelcovici:
# Checklist

      construction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -195,7 +195,26 @@ RUN rustup component add \
   rust-src \
   rustfmt
 
+# No binary available on Github, have to use cargo install
 RUN cargo install cargo-deadlinks
-RUN cargo install cargo-deny
-RUN cargo install cargo-crev
-RUN cargo install grcov
+
+# Where to install rust tooling
+ARG install_dir=${rustup_dir}/bin
+
+# Install grcov
+ARG grcov_version=v0.5.15
+ARG grcov_location=https://github.com/mozilla/grcov/releases/download/${grcov_version}/grcov-linux-x86_64.tar.bz2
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -195,7 +195,26 @@ RUN rustup component add \
...
   rustfmt
 
+# No binary available on Github, have to use cargo install
 RUN cargo install cargo-deadlinks
-RUN cargo install cargo-deny
-RUN cargo install cargo-crev
-RUN cargo install grcov
+
+# Where to install rust tooling
+ARG install_dir=${rustup_dir}/bin
```
tiziano88, line 202: Maybe we should have an overall `tools_dir` and use it for most single-binary tools?

anghelcovici: I did initially that. But then I noticed that this is where the tools went initially, and as these were rust based, so it seemed natural.
Usually the tools install in `/usr/local/bin` or `/usr/bin`, but we could have another folder.

anghelcovici:
This is actually quite horrible.. again problems with users in Docker. If I install cargo-crev it works to run it, but any operation fails, and I think is because we install and run rustup as root. I'll try to see if there is any workaround

tiziano88:
> This is actually quite horrible.. again problems with users in Docker. If I install cargo-crev it works to run it, but any operation fails, and I think is because we install and run rustup as root. I'll try to see if there is any workaround

I think you just need to change the invocation from `cargo crev repo fetch url https://github.com/dpc/crev-proofs` to `cargo-crev crev repo fetch url https://github.com/dpc/crev-proofs` (there is a missing `crev` currently)

PR #1227, opened 2020-07-03
Small clean-up of Dockerfile

anghelcovici:
I've tried to remove unused stuff and reuse existing binaries where possible.
This shaves about 1 GB from the image size, according to my local docker image ls and hopefully makes building the whole image faster.

I've also changed the order for GCB, to run the spellchecking first.

# Checklist

      construction.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -164,21 +164,21 @@ steps:
 
   - name: 'gcr.io/oak-ci/oak:latest'
     id: check_formatting
-    waitFor: ['git_check_diff', 'build_android_image']
+    waitFor: ['git_init', 'build_android_image']
     timeout: 30m
     entrypoint: 'bash'
     args: ['./scripts/check_formatting']
 
   - name: 'gcr.io/oak-ci/oak:latest'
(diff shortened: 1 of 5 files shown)


cloudbuild.yaml
```diff
@@ -164,21 +164,21 @@ steps:
...
-    waitFor: ['check_formatting']
+    waitFor: ['check_formatting', 'git_check_diff']
     timeout: 30m
     entrypoint: 'bash'
     args: ['./scripts/check_cargo_deny']
 
   - name: 'gcr.io/oak-ci/oak:latest'
     id: check_docs
-    waitFor: ['check_formatting']
+    waitFor: ['check_formatting', 'git_check_diff']
```
rbehjati, line 181: I think this step and `check_cargo_deny` are small, inexpensive steps that can be done earlier without having to wait for `git_check_diff`. And IIUC they do not produce any files that need to be checked in the `git_check_diff` step. On the other hand, these steps do not fail as often as `check_formatting `, so it might not be as annoying to run them at the end...

Issue #1183, opened 2020-06-22
Distribution of pre-compiled Wasm modules


tiziano88:
One of the main features of Oak is to allow declassification by nodes that have a WebAssembly logic that is trusted by the owner of the data. For instance, a client may want to specify, via the label / IFC system, that a piece of data can be declassified by a particular Wasm node that aggregates or anonymizes the data before releasing it or further processing it.

In practice this is done by labelling the data with the principal corresponding to that specific node (e.g. the actual hash of the Wasm module, or a key belonging to an entity that is trusted to sign a family of nodes with the desired functionality, or some other schemes involving verifiable logs).

Ideally, an (untrusted) developer of an Oak Application would reuse pieces of logic that implement declassification functionality that have been written / verified by other parties that are more trusted by actual clients / users. For this, we should clarify how a developer would go about including pre-built (and potentially also signed) Wasm modules in a larger Oak Application.

cc @project-oak/core

tiziano88:
Thinking some more about this:

- we need a way for "trusted" module authors to publish their compiled Wasm modules (e.g. aggregation, redaction, etc.), and ideally show provenance from some Rust source code via a reproducible build
- we need a way for third party application developers to use "trusted" modules as part of their (generally untrusted) application

My current thinking is that there should be a mechanism in the [app config serializer](https://github.com/project-oak/oak/blob/main/oak/common/app_config_serializer.cc) to [specify](https://github.com/project-oak/oak/blob/51a4fffc58b54881c1268f9d1da78bb83dac4918/oak/common/app_config_serializer.cc#L34-L36) that a particular module is to be downloaded from a URL, and checked against a provided hash.

For instance, in the aggregator example, we should have a step that builds the Wasm node from sources and publishes that on Google Cloud Store, or as a GitHub release artifact (I think we need to split the aggregator logic from the part of the application that actually creates the gRPC node etc, cc @ipetr0v ). [We already have hashes generated for all the Wasm modules.](https://github.com/project-oak/oak/blob/hashes/reproducibility_index) Then as part of the overall application, we would build the untrusted part of the application locally (does not even need to be reproducibly buildable), and reuse the pre-compiled aggregation logic (which is the one that actually uses its declassification privilege in IFC) by downloading it from from where it is published.

seefeldb:
FYI, https://github.com/WICG/webpackage might be useful prior art for distributing signed content on the web as files (see signed http exchange). It would also be a way to bundle resources with a module.

[...]

PR #1229, opened 2020-07-03
Add push_example script

ipetr0v:
This change adds a script for saving example modules in Google Storage.

Ref https://github.com/project-oak/oak/issues/1183

# Checklist
      construction.

### DIFF
--- a/scripts/push_example
+++ b/scripts/push_example
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+readonly SCRIPTS_DIR="$(dirname "$0")"
+# shellcheck source=scripts/common
+source "${SCRIPTS_DIR}/common"
+
+while getopts "e:h" opt; do
+  case "${opt}" in
+    h)
+      echo -e "Usage: ${0} [-h] -e EXAMPLE
+
+Save example modules in Google Cloud Storage.
+
+Options:
+  -e    Example application name (required)
+  -h    Print Help (this message) and exit"
+      exit 0;;
+    e)
+      readonly EXAMPLE="${OPTARG}";;
+    *)
+      echo "Invalid argument: ${OPTARG}"
+      exit 1;;
+  esac
+done
+
+if [[ -z "${EXAMPLE+z}" ]]; then
+  echo "Missing required option: -e EXAMPLE"
+  exit 1
+fi
+
+readonly EXAMPLE_BIN_DIR="${PWD}/examples/${EXAMPLE}/bin"
+for module in "${EXAMPLE_BIN_DIR:?}"/*.wasm; do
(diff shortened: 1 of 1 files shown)


scripts/push_example
```diff
@@ -0,0 +1,37 @@
...
+  -e    Example application name (required)
+  -h    Print Help (this message) and exit"
+      exit 0;;
+    e)
+      readonly EXAMPLE="${OPTARG}";;
+    *)
+      echo "Invalid argument: ${OPTARG}"
+      exit 1;;
+  esac
+done
```
anghelcovici, line 24: This seems excessive :) The error message below should be enough I think?

scripts/push_example
```diff
@@ -0,0 +1,37 @@
...
+
+if [[ -z "${EXAMPLE+z}" ]]; then
+  echo "Missing required option: -e EXAMPLE"
+  exit 1
+fi
+
+readonly EXAMPLE_BIN_DIR="${PWD}/examples/${EXAMPLE}/bin"
+for module in "${EXAMPLE_BIN_DIR:?}"/*.wasm; do
+  module_hash="$(sha256sum ${module} | awk '{print $1}')"
+  module_name="$(basename ${module} | cut -f 1 -d '.')"
```
anghelcovici, line 34: Can you use either awk or cut for both? This is slightly confusing

anghelcovici: (the first one can be written as cut -f1 -d" ")

scripts/push_example
ipetr0v: Just making it similar to other scripts that use `-e` for referring to an example.

scripts/push_example
ipetr0v: Done

ipetr0v:
The only problem with this script, is that in order to build reproducibly, we need to build examples in Docker.
But our Docker image doesn't have GCP api installed (and probably shouldn't), so pushing examples should be done outside of Docker.

scripts/push_example
anghelcovici: I still suggest we use -e for the example.. but seeing that this has only one way to use it, I don't think we need all the `getops` code for it. If you remove this and it's not used correctly, it will still print the correct message

tiziano88: I have only one word to add, and I am not even going to actually mention it :) (it starts with R)

ipetr0v: I can just make it so that `${1}` will be `-e` and `${2}` will be a parameter.
But I think it will be less readable.

tiziano88: let's keep as it is for now, and think what to do about it in `runner` perhaps in the future.

PR #1230, opened 2020-07-03
Add bin directory for oak_loader

ipetr0v:
This change adds a binary directory for `oak_loader`.

Fixes https://github.com/project-oak/oak/issues/719
Originated from https://github.com/project-oak/oak/pull/1214#discussion_r448934995

# Checklist
      construction.


ipetr0v:
If we will compile `oak_loader` in `bin`, do we need to run it from there?
Or we should still use `cargo run ...`?
cc @anghelcovici @tiziano88

tiziano88:
> If we will compile `oak_loader` in `bin`, do we need to run it from there?
> Or we should still use `cargo run ...`?
> cc @anghelcovici @tiziano88

I feel we should stop using `cargo run` and `bazel run` at some point, it's not very realistic of how the binaries are normally run. So I vote for building first, and then running as binary later. Happy to hear other opinions though.

.gitignore
```diff
@@ -35,8 +35,9 @@
 # Ignore downloaded temporary files, such as the trusted root CA certificates.
 /downloads/
 
-# Ignore generated binary example files
+# Ignore generated binary files.
 **/examples/**/bin
```
tiziano88, line 39: Can we just ignore `**/bin/`, like we ignore `**/target/` below?

ipetr0v: Done

PR #1232, opened 2020-07-03
Add external Wasm modules

ipetr0v:
This change:
- Adds the ability to load external Wasm modules with `oak_config_serializer`
- Makes TIR example use an external precompiled Wasm module

Ref https://github.com/project-oak/oak/issues/1183

# Checklist


sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,44 @@ impl Default for InitialNodeConfig {
     }
 }
 
+fn check_sha256_sum(data: &Vec<u8>, sum: &str) -> bool {
```
tiziano88, line 90: `&Vec<u8>` seems an odd type, would `&[u8]` work?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,44 @@ impl Default for InitialNodeConfig {
...
+    let mut hasher = Sha256::new();
+    hasher.update(data);
+    let result_sum = hasher.finalize();
+    let extected_sum = hex::decode(sum).expect("Couldn't decode SHA256 sum");
```
tiziano88, line 94: s/extected/??/

sdk/rust/oak_config_serializer/src/main.rs
tiziano88, line 90: It feels this function does two things, would it be simpler to just return the sha256 and let the caller do the comparison? Also a panic on failure seems excessive!

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,44 @@ impl Default for InitialNodeConfig {
...
         Module::Path(path) => {
             fs::read(&path).with_context(|| format!("Couldn't read file {}", path))
         }
+        Module::External(external) => {
+            let url: Url = format!("{}/{}", &external.url, &external.sha256)
+                .parse()
+                .context("Couldn't parse URL")?;
+
+            debug!("Downloading module from: {}", url);
+            let response = reqwest::get(url)
```
tiziano88, line 113: I guess at some point we need to have a cache of modules, since this will re-download them every time, perhaps add a TODO + issue?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,44 @@ impl Default for InitialNodeConfig {
...
+                .await
+                .context("Couldn't download module")?;
```
tiziano88, line 115: Print the failing URL too?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,44 @@ impl Default for InitialNodeConfig {
...
+            let data = response
+                .bytes()
+                .await
+                .context("Couldn't retreive module from HTTP response")?
```
tiziano88, line 119:
```suggestion
                .context("Couldn't retrieve module from HTTP response")?
```

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Added an issue https://github.com/project-oak/oak/issues/1240

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,46 @@ impl Default for InitialNodeConfig {
...
+async fn load_module(module: &Module) -> anyhow::Result<Vec<u8>> {
     match module {
-        Module::Url(_url) => unimplemented!(),
         Module::Path(path) => {
             fs::read(&path).with_context(|| format!("Couldn't read file {}", path))
         }
+        Module::External(external) => {
+            let expected_sha256 =
+                hex::decode(&external.sha256).context("Couldn't decode SHA256 sum")?;
+            let url: Url = format!("{}/{}", &external.url, &external.sha256)
```
tiziano88, line 105: Wait, so is it not using the actual provided URL? I think it should use whatever it's in the manifest, and then use the SHA as a checksum. Otherwise it becomes impossible to reuse modules that have been published with different names, right?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,46 @@ impl Default for InitialNodeConfig {
...
+                .parse()
+                .context("Couldn't parse URL")?;
+
+            debug!("Downloading module from: {}", url);
+            // TODO(1240): Add a Wasm module cache.
```
tiziano88, line 110:
```suggestion
            // TODO(#1240): Add a Wasm module cache.
```

tiziano88: (I think that's what we do elsewhere?)

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,46 @@ impl Default for InitialNodeConfig {
...
+                .to_vec();
+
+            if get_sha256(&data)
+                .iter()
+                .zip(expected_sha256.iter())
+                .all(|(first, second)| first == second)
+            {
+                Ok(data)
+            } else {
+                Err(anyhow!("Incorrect SHA256 sum"))
```
tiziano88, line 127: Print the actual and expected ones?

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Currently the name of the file is it's hash.
And the directory is called `database_proxy` for example.

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

sdk/rust/oak_config_serializer/src/main.rs
tiziano88: I think the URL should be used as it is. If the uploader wants to use the hash of the module as file name, they can, but they should not be forced to do it.

ipetr0v: Done

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,45 @@ impl Default for InitialNodeConfig {
     }
 }
 
+fn get_sha256(data: &[u8]) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(data);
+    hex::encode(hasher.finalize().as_slice().to_vec())
```
tiziano88, line 93: Not sure if this works

```suggestion
    hex::encode(hasher.finalize().collect())
```

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,45 @@ impl Default for InitialNodeConfig {
     }
 }
 
+fn get_sha256(data: &[u8]) -> String {
```
tiziano88, line 90: Add a comment (mostly to clarify that it's already hex encoded)

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -78,13 +87,45 @@ impl Default for InitialNodeConfig {
...
             fs::read(&path).with_context(|| format!("Couldn't read file {}", path))
         }
+        Module::External(external) => {
+            let url: Url = external.url
+                .parse()
+                .context("Couldn't parse URL")?;
+
+            debug!("Downloading module from: {}", url);
+            // TODO(#1240): Add a Wasm module cache.
+            let response = reqwest::get(url.clone())
```
tiziano88, line 109: Why can `url` not be passed by reference?

ipetr0v: It's used in `.with_context(|| format!("Couldn't download module from {}", url))?;` that moves it in the context of lambda.

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Finalize returns a [`GenericArray`](https://docs.rs/generic-array/0.14.2/generic_array/struct.GenericArray.html) and it doesn't have `collect`.

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

PR #1236, opened 2020-07-03
Fix issues found by shellcheck

tiziano88:
These scripts were not under `./scripts` and therefore missed by
`shellcheck`.

As part of #1190, `runner` actually runs `shellcheck` on them, and
pointed out a few existing errors.

# Checklist

      construction.

### DIFF
--- a/examples/aggregator/scripts/docker_push
+++ b/examples/aggregator/scripts/docker_push
@@ -1,31 +1,30 @@
 #!/usr/bin/env bash
 
 readonly GLOBAL_SCRIPTS_DIR="$(dirname "$(readlink -f "$0")")/../../../scripts/"
-readonly SCRIPTS_DIR="$(dirname "$(readlink -f "$0")")"
 # shellcheck source=scripts/common
 source "${GLOBAL_SCRIPTS_DIR}/common"
 
 readonly AGGREGATOR_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-aggregator'
 readonly BACKEND_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-aggregator-backend'
 
-declare -a image backend_image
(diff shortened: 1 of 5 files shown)


tiziano88:
> The build.sh files have been removed. Can you rebase first?

Done.

> Why didn't these cause a failure in our build?

See commit message / PR description (`shellcheck` was only linting files under `scripts`. `runner` is a bit more flexible and found these files even outside of `scripts`)

anghelcovici:
How is runner more flexible? (as in, doesn't it require a search path also?)

examples/aggregator/scripts/docker_push
```diff
@@ -1,31 +1,30 @@
...
+if [[ ${#images[@]} -eq 0 || ${#backend_images[@]} -eq 0 ]]; then
+  echo "Docker images have not been built, run './scripts/build_example -e aggregator -i base'"
   exit 1
 fi
-if [[ ${#image[@]} > 1 || ${#backend_image[@]} > 1 ]]; then
-  echo "Too many Docker images: ${image}, ${backend_image}"
+if [[ ${#images[@]} -gt 1 || ${#backend_images[@]} -gt 1 ]]; then
+  echo "Too many Docker images: " "${images[@]}" ", " "${backend_images[@]}"
   exit 0
 fi
```
tiziano88, line 20: FYI @ipetr0v this was also incorrect (was mixing up strings and array types, and shellcheck found a lot of issues). I have no idea what I'm doing, but it seems to work now, does it look reasonable to you?

ipetr0v: Yes, seems reasonable.

tiziano88:
> How is runner more flexible? (as in, doesn't it require a search path also?)

I made the version of runner in my own branch run `shellcheck` on all files starting with `#!`. I guess it could also be done via bash scripts. :)

Issue #1252, opened 2020-07-10
[runner] run binaries directly instead of via `cargo run`


tiziano88:
https://github.com/project-oak/oak/blob/c099895766237fcebf24aae2886d6be2dbe2052a/runner/src/main.rs#L441-L461

we use `cargo run` and `bazel run` in a few places in scripts and in the runner, as a convenience for getting things up and running quickly. This is not necessarily representative of how those binaries will be used in practice (e.g. we still need to invoke the binary itself from the docker image). Also by using `cargo run`, a single step is really doing two things: building and then running the binary, which means that if we just `cargo run` the server, it may take anywhere between a couple of seconds to several minutes for the server to actually start, based on whether the build step was cached.

cc @ipetr0v

ipetr0v:
We used `readonly SERVER_BIN_DIR="${PWD}/oak/server/bin"` in shell scripts:
https://github.com/project-oak/oak/blob/d1e6497e73223afe46ddd9f3618f971bac1b5205/scripts/build_server#L37-L39
https://github.com/project-oak/oak/blob/d1e6497e73223afe46ddd9f3618f971bac1b5205/scripts/run_server#L90-L91

Will add the same this to `runner`.

ipetr0v:
We also are currently copying artifacts to `bin` directory:
https://github.com/project-oak/oak/blob/c099895766237fcebf24aae2886d6be2dbe2052a/runner/src/main.rs#L266-L284

Should we use `out-dir` (it's unstable)?

tiziano88:
`out-dir` looks reasonable, and I think we already use it somewhere? Or maybe it was just in the previous scripts?

ipetr0v:
Yes, it was in `build_server`.

PR #1258, opened 2020-07-13
Remove cargo run from runner

ipetr0v:
This change:
- Removes `cargo run` from `runner`
- Copies `oak_loader` into `bin` and use it to run Oak server

Fixes https://github.com/project-oak/oak/issues/1252

# Checklist

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -291,48 +291,63 @@ fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
 
 fn build_server(opt: &BuildServer) -> Step {
     match opt.server_variant.as_str() {
-        "base" | "logless" => Step::Single {
-            name: format!("build server ({})", opt.server_variant),
-            command: Cmd::new_with_env(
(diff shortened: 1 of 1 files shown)


runner/src/main.rs
```diff
@@ -291,48 +291,63 @@ fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
...
-                } else {
-                    hashmap! {}
+        "base" | "logless" => Step::Multiple {
+            name: "server".to_string(),
+            steps: vec![
+                Step::Single {
+                    name: "create bin folder".to_string(),
+                    command: Cmd::new(
+                        "mkdir",
+                        vec!["--parent".to_string(), "oak/server/bin".to_string()],
```
tiziano88, line 301: I think @daviddrysdale mentioned that `--parents` does not work on macos, only `-p` does, so maybe we should stick to that. Also BTW it is `--parents`, not `--parent`, so are you sure this actually worked? :)

PR #1260, opened 2020-07-15
Add protobuf schema for introspection events

juliettepretot:
See [David's comment](https://github.com/project-oak/oak/issues/913#issuecomment-655652315) for background on this change in #913. These events will allow dynamic introspect, they're based on his [stringly typed](https://wiki.c2.com/?StringlyTyped) version of'em. 

# Checklist

      construction.

### DIFF
--- a/oak/proto/introspection_events.proto
+++ b/oak/proto/introspection_events.proto
@@ -1,0 +1,89 @@
+//
+// Copyright 2019 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package oak.event;
+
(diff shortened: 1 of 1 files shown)


oak/proto/introspection_events.proto
```diff
@@ -0,0 +1,105 @@
...
+
+message MessageDequeueing {
+  // The id of the destroyed node
+  uint64 node_id = 1;
+
+  // The id of the created node
+  uint64 channel_id = 2;
+
+  repeated uint64 acquired_handles = 3;
+}
```
juliettepretot, line 105: It would be possible combine these to be DRYer, and use `optional` or `oneof` types for the different fields.

However I like spelling them out verbosely, as it's quite clear and makes sure that no invalid combinations of fields are possible.

oak/proto/introspection_events.proto
```diff
@@ -0,0 +1,105 @@
...
+import "google/protobuf/timestamp.proto";
+
+package oak.event;
+
+// This messages defines an event sent by the Oak Runtime to allow for dynamic
+// introspection in non-production enviroments. It provides a "schema" to
+// describe the creation/destruction of nodes, channels, handles, as well as the
+// queuing of messages.
+message Event {
+  google.protobuf.Timestamp timestamp = 1;
```
juliettepretot, line 28: See https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto on details for the type. If preferred we could also use a simple `int64` and insert a UNIX timestamp ourselves.

Timestamps would be useful, since they allow for "replaying" of events. I know precise timestamps can pose security risks, but these events are only sent in non-prod enviroments.

tiziano88:
I'm OOO so please don't wait for my review, thanks!

juliettepretot:
@tiziano88 didn't mean to wait for your review, go enjoy your vacation <3

PR #1263, opened 2020-07-16
Use crate prefix when generating prost code

ipetr0v:
This change:
- Checks if `proto` files are generated for `oak` crate, and uses `crate` instead of `::oak`
- Updates `prost` patch file

# Checklist


ipetr0v:
Apparently, when tests are generated, they are not a part of `oak` crate, but the environment variable `CARGO_PKG_NAME` is set to `oak`:
```
error[E0433]: failed to resolve: unresolved import
/opt/my-project/sdk/target/debug/build/oak-a64860617a590aac/out/handle_tests/tests.rs:23:29

EitherSender(crate::io::Sender<super::TestMessageType>),
                                ^^
                                 |
                                 unresolved import
                                 help: a similar path exists: `oak::io`
```

ipetr0v:
Since `cargo` doesn't give us an environment variable to check if we a building an integration test, I can just add a new parameter to:
https://github.com/project-oak/oak/blob/d96093e0b61fb2812747905dc71d32d2e475cfe8/oak_utils/src/lib.rs#L166-L167

oak_utils/src/lib.rs
```diff
@@ -182,6 +182,13 @@ pub struct ProtoOptions {
     pub derive_handle_visit: bool,
 
     pub out_dir_override: Option<std::path::PathBuf>,
+
+    /// Replace `::oak` with `crate` in the generated proto code.
+    ///
+    /// Code generator expects an `oak` crate to be available for resolving `::oak::io::Sender`.
+    /// The generated path is valid for all crates that list `oak` as a dependency, but is not
+    /// currently available for `oak` itself.
+    pub replace_oak_with_crate: bool,
```
conradgrobler, line 191: Is this needed? It seems like building the test was fixed by using self.package elsewhere. This might cause problems that this is used to determine the package name in L223, but the it still uses oak_package() in L48 for the same purpose, so it might get out of sync.

ipetr0v: it doesn't fix tests everywhere. SDK `compile_protos_with_options` will use `CARGO_PKG_NAME == "oak"` for tests and thus set `crate` in `#[derive(crate::handle::HandleVisit)]`.

ipetr0v: Maybe there is another way of learning the package name in `sdk`? Similar to `self.package` from `prost`.

conradgrobler: I think the underlying problem is that the `build.rs` is only called when the sdk is built, and the code for the protos is generated at that point (see https://github.com/project-oak/oak/pull/1108#discussion_r450987029).

I guess it would work as it is at the moment, but this will break if these tests ever include gRPC service definitions, as the gRPC code generation in L48 will still use the old logic. For now it is probably fine, as I don't think we plan on adding gRPC service definitions to the SDK test proto files.

tiziano88:
@ipetr0v is this PR still useful / necessary?

conradgrobler:
I don't think this is needed anymore since #1400

ipetr0v:
Closing as outdated

PR #1272, opened 2020-07-17
Move cloudbuild step of packaging an example into the runner

juliettepretot:
See the discussion in https://project-oak.slack.com/archives/CHE9E13C3/p1594984760219500 for background. 

# Checklist

      construction.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -449,6 +449,22 @@ fn run_ci() -> Step {
                     coverage: false,
                 },
             }),
+            // Package the Hello World application in a Docker image.
+            run_examples(&RunExamples {
+                application_variant: "rust".to_string(),
+                example_name: Some("hello_world".to_string()),
+                run_server: Some(false),
+                run_clients: Some(false),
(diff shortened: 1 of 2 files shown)


runner/src/main.rs
```diff
@@ -449,6 +449,22 @@ fn run_ci() -> Step {
                     coverage: false,
                 },
             }),
+            // Package the Hello World application in a Docker image.
+            run_examples(&RunExamples {
```
rbehjati, line 453: Could this be a separate command that can be run individually? I think that would be helpful when working locally.

juliettepretot: If you wanted to run just this step, you can do so via: `./scripts/docker_run ./scripts/runner run-examples --example-name=hello_world --run-server=false --run-clients=false --build-docker`

juliettepretot: The code here basically provides the same arguments, just in the shape of a step inside of the `run-ci` task.

rbehjati: You are right. Thanks :)

PR #1275, opened 2020-07-21
Exclude `https://codecov.io/gh/project-oak/oak` from link check

rbehjati:
Link check often fails when checking `https://codecov.io/gh/project-oak/oak`. To make the CI-builds more stable it is excluded from `liche`.


runner/src/main.rs
```diff
@@ -1070,7 +1070,7 @@ fn run_liche() -> Step {
                     // We exclude the following URLs from the checks:
                     // - https://groups.google.com/g/project-oak-discuss : not publicly accessible
                     // - https://crates.io/crates : returns 404 (see https://github.com/raviqqe/liche/issues/39)
```
tiziano88, line 1072: Could you add a line to the comment too?

rbehjati: Done! Thanks :)

PR #1276, opened 2020-07-21
Implement introspection events 

juliettepretot:
Implements `ChannelCreated`, `HandleCreated`, `HandleDestroyed`, `NodeCreated`, `NodeDestroyed`

Omits the `ChannelDestroyed` event for now, for reasons explained here: https://github.com/project-oak/oak/issues/913#issuecomment-661794936

# Checklist

      construction.

### DIFF
--- a/oak/server/rust/oak_runtime/src/runtime/channel.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/channel.rs
@@ -47,7 +47,7 @@ type WaitingThreads = Mutex<HashMap<ThreadId, Weak<Thread>>>;
 /// is included in the `reader_count` or `writer_count`.
 pub struct Channel {
     // An internal identifier for this channel. Purely for disambiguation in debugging output.
-    id: ChannelId,
+    pub id: ChannelId,
 
     pub messages: RwLock<Messages>,
 
@@ -85,6 +85,11 @@ impl std::fmt::Debug for Channel {
 
 /// A reference to one half of a [`Channel`].
 pub struct ChannelHalf {
+    // Make the associated channel public for introspection
+    #[cfg(feature = "oak_debug")]
+    pub channel: Arc<Channel>,
+    // Ensure the associated channel is private
+    #[cfg(not(feature = "oak_debug"))]
     channel: Arc<Channel>,
     pub direction: ChannelHalfDirection,
 }
--- a/oak/server/rust/oak_runtime/src/runtime/channel.rs
+++ b/oak/server/rust/oak_runtime/src/runtime/channel.rs
@@ -47,7 +47,7 @@ type WaitingThreads = Mutex<HashMap<ThreadId, Weak<Thread>>>;
 /// is included in the `reader_count` or `writer_count`.
 pub struct Channel {
     // An internal identifier for this channel. Purely for disambiguation in debugging output.
-    pub id: ChannelId,
+    id: ChannelId,
 
     pub messages: RwLock<Messages>,
 
@@ -85,11 +85,6 @@ impl std::fmt::Debug for Channel {
 
 /// A reference to one half of a [`Channel`].
 pub struct ChannelHalf {
-    // Make the associated channel public for introspection
-    #[cfg(feature = "oak_debug")]
-    pub channel: Arc<Channel>,
-    // Ensure the associated channel is private
-    #[cfg(not(feature = "oak_debug"))]
     channel: Arc<Channel>,
     pub direction: ChannelHalfDirection,
 }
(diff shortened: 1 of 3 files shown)


oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -269,7 +274,17 @@ impl Runtime {
...
                     node_id, candidate, half
                 );
+
+                let event_details = HandleCreated {
+                    node_id: node_id_to_primitive(node_id),
+                    handle: candidate,
+                    channel_id: half.get_id(),
+                };
+
                 node_info.abi_handles.insert(candidate, half);
```
rbehjati, line 284: nit: I'd move this line of code before the line that creates `event_details`.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -594,6 +627,12 @@ impl Runtime {
             write_handle,
             read_handle,
         );
+
+        self.introspection_event(EventDetails::ChannelCreated(ChannelCreated {
```
rbehjati, line 631: Is this event really needed? I noticed the `ChannelDestroyed` event is never used, and it does not seem very straightforward when that one should be created. 

In general, I think it is nice to keep these events in sync with changes to the `node_infos`. Looks like there is no specific change to the `node_infos` that corresponds to the creation of a channel. So perhaps this could be removed. Especially since it is covered by the `HandleCreated` event, which has a matching `HandleDestroyed` event.

oak/server/rust/oak_runtime/src/runtime/mod.rs
juliettepretot: I'm still new to Rust, but afaik that's not possible. This line moves the `half` variable out of scope, and I need to access it to get the `channel_id` for the event details. :)

oak/server/rust/oak_runtime/src/runtime/introspection_events.rs
```diff
@@ -16,12 +16,17 @@
...
     proto::oak::introspection_events::{event::EventDetails, Event},
-    runtime::Runtime,
+    runtime::{NodeId, Runtime},
 };
 use log::info;
 use prost_types::Timestamp;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+pub fn node_id_to_primitive(node_id: NodeId) -> u64 {
+    let NodeId(node_id_as_primitive) = node_id;
```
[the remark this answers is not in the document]
juliettepretot: Oh nice, I didn't know of this syntax, will use it instead.

oak/server/rust/oak_runtime/src/runtime/mod.rs
juliettepretot: Fair point. :) I does seem like the `Handle` events should give us sufficient information where this event could technically be omitted. 

The goal currently is to still implement a `ChannelDestroyed` event down the road, but the implementation is currently still unclear, since it requires refactoring the `introspection_event ` method. See details here: https://github.com/project-oak/oak/issues/913#issuecomment-661794936

In principle I do think that events for channels are still useful. They allow for easier updating, since channels are represented in the introspection tools. You are correct though that their creation & destruction could be deducted from `Handle` (and the yet to be added `Message`) events.

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -278,11 +292,29 @@ impl Runtime {
...
+        };
+
+        let result = node_info
             .abi_handles
             .remove(&handle)
             .ok_or(OakStatus::ErrBadHandle)
-            .map(|_half| ())
+            .map(|_half| ());
+
+        match result {
```
conradgrobler, line 312: nit: am I not sure what is best in terms of idiomatic Rust, but I would have thought a standard `if result.is_ok()` is easier to read.

oak/server/rust/oak_runtime/src/runtime/channel.rs
```diff
@@ -100,6 +100,12 @@ impl ChannelHalf {
         ChannelHalf { channel, direction }
     }
 
+    /// Get the ID of the underlying channel.  For debugging/introspection
+    /// purposes.
+    pub fn get_id(&self) -> u64 {
```
conradgrobler, line 105: nit: perhaps `get_channel_id` seeing that the id does not refer to the channel half itself?

oak/server/rust/oak_runtime/src/runtime/mod.rs
rbehjati: Yes. I thought maybe there is a way around it. It should be possible to send the  `introspection_event` first and then update the `node_info`. But it is not a big deal... :)

juliettepretot: Yes! I could do that. I guess my intention is to send the event _after_ the action has been performed, but it's my obsessive focus on details more than anything else. ^^

oak/server/rust/oak_runtime/src/runtime/mod.rs
juliettepretot: Makes sense, happy to use that instead. To be entirely honest I largely used a match since I didn't know the `is_ok()` method existed yet. ^^

Issue #1250, opened 2020-07-10
allow running the C++ hello world module with `runner`


tiziano88:
After https://github.com/project-oak/oak/pull/1190 is submitted, the hello world example is only run in its Rust variant (though the C++ variant is built). In order to support running the C++ variant too, we should either create a separate manifest file, and treat it as a separate example, or extend the manifest format to support multiple application variants for each example.

ipetr0v:
Why do we need a separate manifest file? And what kind of file?
If it's about a `.toml` file with an application configuration, then we already have separate files for Rust and CPP:
https://github.com/project-oak/oak/blob/120484391822adf383695f1312bada02d8d0be12/examples/hello_world/config/config.toml#L1-L5
https://github.com/project-oak/oak/blob/120484391822adf383695f1312bada02d8d0be12/examples/hello_world/config/config_cpp.toml#L1-L4

ipetr0v:
Got it, this is the manifest file:
https://github.com/project-oak/oak/blob/120484391822adf383695f1312bada02d8d0be12/examples/hello_world/example.toml#L3-L5

PR #1281, opened 2020-07-23
Enable C++ examples in runner

ipetr0v:
This change enables C++ example modules in `runner`.

Fixes https://github.com/project-oak/oak/issues/1250

# Checklist


examples/hello_world/example.toml
```diff
@@ -1,14 +1,13 @@
 name = "hello_world"
 
 [application]
+variant = "rust"
```
tiziano88, line 4: (optional) Since you are adding an application variant, we may as well have multiple application variants in the same example manifest now. This would probably require making `modules` a field of `application`, which I think makes sense anyways.

anghelcovici: I actually think this is really nice. I don't like the fact that we hardcode "example_cpp.toml" in the runner, especially since I would like to have a "rust only" version of an example, and that would require yet another hardcoded version.

ipetr0v: Done

Issue #1234, opened 2020-07-03
Make `runner` show a summary of failed steps at the end of the execution


tiziano88:
I think it is great that this command runs all the steps even if something fails in between. 
I suggest logging a summary of the failures or the list of failed steps in the end, so that one does not have to go through the entire log to locate the problem.

_Originally posted by @rbehjati in https://github.com/project-oak/oak/pull/1190_

ipetr0v:
Should it only apply to `run_ci`?
If so, we need to also add a `name` for each `Step`.

ipetr0v:
Adding a PR https://github.com/project-oak/oak/pull/1286.
Error summary in the end will look like an exact list of failed steps:
```
Finished with errors:
 ❯ examples ❯ hello_world ❯ build application ⊢ [ERROR]
```

PR #1286, opened 2020-07-24
Show error summary in runner

ipetr0v:
This change adds a short failure summary in `runner`.

Fixes https://github.com/project-oak/oak/issues/1234

# Checklist

### DIFF
--- a/runner/src/internal.rs
+++ b/runner/src/internal.rs
@@ -312,7 +312,6 @@ pub async fn run_step(context: &Context, step: Step) -> StepResult {
                     eprintln!("{} {} {}", context.prefix, "║".blue(), line);
                 }
                 eprintln!("{} {}", context.prefix, "╚════════════════════════".blue());
-                step_result.failed_steps_prefixes.push(context.prefix);
             }
 
             step_result.values = foreground_result.values;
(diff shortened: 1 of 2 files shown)


runner/src/internal.rs
```diff
@@ -288,8 +301,9 @@ pub async fn run_step(context: &Context, step: Step) -> HashSet<StatusResultValu
                 "{} ⊢ (finished) {}",
                 context.prefix, background_status.value
             );
+
             if (background_status.value == StatusResultValue::Error
-                || values.contains(&StatusResultValue::Error)
+                || foreground_result.values.contains(&StatusResultValue::Error)
                 || context.opt.logs)
                 && !logs.is_empty()
```
rbehjati, line 308: Is this check needed for pushing the step to `step_result.failed_steps_prefixes`? Or should `step_result.failed_steps_prefixes` be updated regardless of the content of `logs`?

runner/src/main.rs
```diff
@@ -171,8 +171,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             std::process::exit(-1);
         });
         let statuses = run().await;
+
+        if !statuses.failed_steps_prefixes.is_empty() {
+            eprintln!("Finished with errors:");
```
rbehjati, line 176: nit: "List of failed steps:" may be a better title.

runner/src/internal.rs
ipetr0v: Done

Issue #1240, opened 2020-07-06
Implement a Wasm module cache


ipetr0v:
After https://github.com/project-oak/oak/pull/1232 is merged, we need to add a cache of Wasm modules, that are currently being re-downloaded every time an application configuration is serialized.

Originated from https://github.com/project-oak/oak/pull/1232/files#r450288627
Ref https://github.com/project-oak/oak/issues/1183

PR #1289, opened 2020-07-27
Add Wasm module cache

ipetr0v:
This change adds the ability to:
- Cache downloaded Wasm modules
- Set SHA256 sum for all Wasm modules used by `oak_config_serializer`

Fixes https://github.com/project-oak/oak/issues/1240

# Checklist


sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -94,37 +86,55 @@ fn get_sha256(data: &[u8]) -> String {
     hex::encode(hasher.finalize().as_slice().to_vec())
 }
 
-/// Load Wasm module from URL or a file, if URL is not specified.
+/// Download Wasm module from `url` and caches it in `filename`.
```
tiziano88, line 89: This makes filename both an input and an output, and makes the caching mechanism poorly encapsulated. I would suggest to keep the previous version, and just cache things under a `.oak` did, or directly under `/tmp`, don't know what are the current best practices

tiziano88: In fact maybe just under the bin/out dir that we use for final artifacts, add a `oak_module_cache` subfolder or similar

ipetr0v: Which previous version? From this PR?
Because prior to this PR we didn't cache downloaded files.

tiziano88: I meant keep the interface of the previous version, by which either a URL or a filename need to be provided

ipetr0v: > In fact maybe just under the bin/out dir that we use for final artifacts, add a oak_module_cache subfolder or similar
Currently it saves downloaded files in the `${EXAMPLE}/bin` directory.

tiziano88: I thought it downloads them every time though? Caching should be a transparent optimization so that if a file with the correct name and hash is already available, then it won't get downloaded again. I'm not sure that's already the case, if so perhaps there's nothing else to do? :)

ipetr0v: I mean this PR makes it download a module and save it into `bin`.
And every time it checks whether this file is available in `bin` or not.

tiziano88: Ah I see, then perhaps all I'm saying is that it should just hide the filename from the user, and generate one based on just the hash? Think from the point of view of a user, how would they pick a filename? The tool should do as much as possible for them and hide unnecessary details

ipetr0v: If we want to use `bin` for caching, it should be provided as a parameter to `config_serializer`.
So in order to hide it from the user, I'll just add a `/tmp/oak_config_serializer/${NAME}` directory.

tiziano88: We already pass the output file as flag though, right? Maybe we can just create a directory within the target containing directory. Up to you anyways. I am afraid if we just use /tmp it will be hard / impossible to clean up over time

ipetr0v: Agree, cleaning up `/tmp` is annoying

tiziano88: Yet another option: just create a subdirectory under the PWD (which we would just git ignore). This is perhaps the most similar to how git and Bazel and cargo work? Although not exactly

tiziano88: If we wanted to make this most similar to cargo, it would have just a flag for the manifest file, and then figure out where to put cached and final artifacts based on that. I guess it would need an extra field in the manifest file for the application name, to be used as file name, if it's not already there. Pretty much the same as rust cargo.toml anyway :)

ipetr0v: Done
Saving into `.oak/`

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -57,18 +57,18 @@ struct Config {
...
 #[derive(Debug, serde::Deserialize)]
 #[serde(deny_unknown_fields)]
-enum Module {
-    #[serde(rename = "path")]
-    Path(String),
-    #[serde(rename = "external")]
-    External(External),
+struct Module {
+    file: WasmFile,
+    sha256: Option<String>,
```
tiziano88, line 62: I'm not sure about this change; IIUC, this allows omitting the hash for remote files, which I think we should actively discourage, and ideally forbid entirely. For local files instead, there is not much value in having an explicit hash anyways, since it usually means that the module was already built locally, or comes from a trusted source.

Personally I would leave the interface as it was, and then perhaps make more structural change (e.g. adding the hash to local files too) in a separate, dedicated PR.

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,44 +87,67 @@ impl Default for InitialNodeConfig {
     }
 }
 
+const CACHE_DIRECTORY: &str = ".oak";
```
tiziano88, line 90: Add a comment. Also, what is this relative to?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,44 +87,67 @@ impl Default for InitialNodeConfig {
...
+
 /// Computes SHA256 sum from `data` and returns it as a HEX encoded string.
 fn get_sha256(data: &[u8]) -> String {
     let mut hasher = Sha256::new();
     hasher.update(data);
     hex::encode(hasher.finalize().as_slice().to_vec())
 }
 
-/// Load Wasm module from URL or a file, if URL is not specified.
+/// Download Wasm module from `url`.
```
tiziano88, line 99: I see nothing about wasm modules in this fn, it seems a generic download file from URL?

ipetr0v: Changed

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Added comments

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,35 +87,55 @@ impl Default for InitialNodeConfig {
...
 fn get_sha256(data: &[u8]) -> String {
     let mut hasher = Sha256::new();
     hasher.update(data);
     hex::encode(hasher.finalize().as_slice().to_vec())
 }
 
-/// Load Wasm module from URL or a file, if URL is not specified.
+/// Download file from `url`.
+async fn download_module_from_url(url: &str) -> anyhow::Result<Vec<u8>> {
+    let url: Url = url.parse().context("Couldn't parse URL")?;
```
tiziano88, line 103: Maybe also print the URL in the context?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,35 +87,55 @@ impl Default for InitialNodeConfig {
...
 /// Computes SHA256 sum from `data` and returns it as a HEX encoded string.
 fn get_sha256(data: &[u8]) -> String {
     let mut hasher = Sha256::new();
     hasher.update(data);
     hex::encode(hasher.finalize().as_slice().to_vec())
 }
 
-/// Load Wasm module from URL or a file, if URL is not specified.
+/// Download file from `url`.
+async fn download_module_from_url(url: &str) -> anyhow::Result<Vec<u8>> {
```
tiziano88, line 102: I think also the fn name and the error messages should talk about files, not modules.

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,35 +87,55 @@ impl Default for InitialNodeConfig {
...
-                .bytes()
-                .await
-                .context("Couldn't retrieve module from HTTP response")?
-                .to_vec();
+            let mut cache_path = std::env::current_dir().unwrap();
+            cache_path.push(CACHE_DIRECTORY);
+            std::fs::create_dir_all(cache_path.as_path())
+                .context("Couldn't create cache directory")?;
 
+            let data = download_module_from_url(&external.url).await?;
```
tiziano88, line 130: I may be missing something, but it looks to me that this is downloading the file every time, so how does the caching actually work?

ipetr0v: Thanks!
I somehow lost the most important functionality
Looks like WFH finally started affecting me =)

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,33 +87,71 @@ impl Default for InitialNodeConfig {
...
-                .with_context(|| format!("Couldn't download module from {}", url))?;
-            let data = response
-                .bytes()
-                .await
-                .context("Couldn't retrieve module from HTTP response")?
-                .to_vec();
-
+            let mut cache_path = std::env::current_dir().unwrap();
+            cache_path.push(CACHE_DIRECTORY);
+            cache_path.push(&external.sha256);
```
tiziano88, line 128: I think this could be a separate method, with a bit more documentation (e.g. explain that the cached files are named after the hash)l

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,33 +87,71 @@ impl Default for InitialNodeConfig {
...
+
+            // Try to load module from cache, if failed, download it from URL.
+            let data = match fs::read(&cache_path) {
+                Ok(data) => {
+                    debug!("Loaded module from cache {:?}", cache_path.as_path());
```
tiziano88, line 133: I would make these logs `info`, I think normal users would like to see them, even if they are not debugging the binary, right?

sdk/rust/oak_config_serializer/src/main.rs
```diff
@@ -87,33 +87,71 @@ impl Default for InitialNodeConfig {
...
+                        .context("Couldn't create cache directory")?;
+                    fs::write(&cache_path, &data).with_context(|| {
+                        format!("Couldn't write file {:?}", cache_path.as_path())
+                    })?;
+                    data
+                }
+            };
+
+            // Check SHA256 sum of the Wasm module.
             let received_sha256 = get_sha256(&data);
```
tiziano88, line 155: Maybe rename this, since it is not always "received"

sdk/rust/oak_config_serializer/src/main.rs
ipetr0v: Done

PR #1290, opened 2020-07-28
Docs

benlaurie:
Document quick start a bit better, plus make it correct.


docs/Docker.md
```diff
@@ -0,0 +1,23 @@
...
+
+```bash
+brew install docker
+```
+
+And then what? Oak's install script fails with
+
+```bash
+Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
+```
```
juliettepretot, line 23: It appears that the docker daemon cannot natively run on Mac (uses Linux kernel features), hence a VM is needed t run it. Here's a guide: https://medium.com/@yutafujii_59175/a-complete-one-by-one-guide-to-install-docker-on-your-mac-os-using-homebrew-e818eb4cfc3

I assume the docker.com setup includes something akin to this.

juliettepretot: Alternatively one can install the docker.com GUI application via brew through `brew cask install docker`.

tiziano88:
Would it make sense to point to https://docs.docker.com/get-started/#set-up-your-docker-environment or some other existing documentation, instead of maintaining our own fork of it?

benlaurie:
> Would it make sense to point to https://docs.docker.com/get-started/#set-up-your-docker-environment or some other existing documentation, instead of maintaining our own fork of it?

It would, except it doesn't tell me how to deal with the situation I am now (back) in...

```
$ docker run hello-world
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

docs/Docker.md
benlaurie: So simple!

juliettepretot:
@benlaurie It seems to you need to run `./scripts/runner format` to format the update docs & thereby patch the failing check. :) 

Opened a separate PR (#1297) to make this clearer in our contribution docs.

Issue #1291, opened 2020-07-28
Stop runner immediately after a failure


ipetr0v:
Currently `runner` continues execution after some steps have failed. For example, it could continue working for 30 mins after `formatter` have shown an error: https://github.com/project-oak/oak/issues/1283#issuecomment-664954151

cc @anghelcovici @tiziano88 @rbehjati

rbehjati:
I think that is intentional. So that you get a list of **all** the failing steps with a single CI-build, instead of having to trigger the ci-build again and again, to find out about all the failing steps. 

Stopping the `runner` immediately after the first failing step makes it almost pointless. In that case, I think it would be much better to have separate `cloudbuild` steps, like we had before, instead of having a single `runner_ci` step.

ipetr0v:
Closing, since the behavior is intended.

Issue #1283, opened 2020-07-24
Runner suggestions


anghelcovici:
I've been playing a bit with the Runner and I want to brain dump some possible improvements. Not sure if these are already covered by other issues:

- [ ] There is no progress indicator. Getting stuck on a step for 300+ seconds is not great. Can we have a spinning thing? :)
- [x] I didn't have Bazel installed. The error was not clear (`panic occurred: could not spawn command: Os { code: 2, kind: NotFound, message: "No such file or directory" }`) and it has hard to investigate. Maybe improve error messages to show/dump detailed logs when crashing?
    - Error logs seem more helpful now: `panic occurred: could not spawn command: Command { std: "prettier" "--write" "./oak_sign/Cargo.toml", kill_on_drop: false }: No such file or directory (os error 2)`
- [x] On [this run](https://pantheon.corp.google.com/cloud-build/builds/571d674e-ff87-4deb-b7fd-dc6333965e7a;step=4?project=oak-ci), the formatter had an error, but the build continued for 30 minutes. Can we stop things as soon as the error happens? (Issue: https://github.com/project-oak/oak/issues/1291) 
    - [DONE: Behaviour changed in 3ead617fc405d4574beb7c15d31cc0afc7054271]
- [ ] Building Runner takes 3 minutes on my mac and it also require a nightly compiler. Can we prebuild it? Also, do we really need experimental features in a tool?
- [ ] The builds are not deterministic (as in the order is not the same). I've ran this in docker on may machine and in docker on a mac and they proceed in different order. Should we make them deterministic?
- [ ] There is no output by default for the examples. Can we enable the running of examples to show output by default?
- [ ] When building the hello_world example there is no way to exclude the cpp module from being built. This adds quite a lot of time for the first build I did.
- [x] A C++ module for the `hello_world` example is not currently being run (Issue: https://github.com/project-oak/oak/issues/1250, PR: https://github.com/project-oak/oak/pull/1281).
- [x] I had an error in the "build application" step and the runner continued to running the server. Is this intentional?
    - [DONE: It was, but behaviour changed in 3ead617fc405d4574beb7c15d31cc0afc7054271]
- [x] We are not building the Rust client for TIR. Do we support Rust clients in the runner? (Issue: https://github.com/project-oak/oak/issues/1285, PR: https://github.com/project-oak/oak/pull/1287)
[...]

PR #1295, opened 2020-07-29
Add time-of-day, and fail/pass status to each log line of `runner`

rbehjati:
* Adds time-of-day (UTC) in the `%H:%M:%S` format to each log line
* Adds run status in the `E:<error-count>,O:<ok-count>,R:<remaining-count>`
  format to each log line

Example log line:
```
[11:43:59; E:0,O:6,R:3]:  ❯ ci ❯ examples ❯ aggregator ❯ run ❯ background server ❯ run clients ❯ cpp } ⊢ [OK] [3s]
```

This change can hopefully address some of the issues raised in #1283 and #1291.

### DIFF
--- a/runner/src/internal.rs
+++ b/runner/src/internal.rs
@@ -101,25 +101,80 @@ pub struct BuildServer {
     pub coverage: bool,
 }
 
+/// A construct to keep track of the status of the execution. It only cares about the top-level
+/// steps.
+#[derive(Clone)]
+pub struct Status {
+    error: usize,
+    ok: usize,
+    remaining: usize,
+}
+
+impl Status {
+    pub fn new(remaining: usize) -> Self {
+        Status {
+            error: 0,
+            ok: 0,
+            remaining,
+        }
+    }
+
+    /// Guarantees that the `error`, `ok`, and `remaining` counts are updated only after the
+    /// completions of each top-level step.
+    fn update(&mut self, context: &Context, step_has_error: bool) {
+        // Update the status with results from the step, only if it is a top-level step.
+        if context.depth() == 1 {
+            self.remaining -= 1;
+            // We only care about pass (`ok`) and fail (`error`). If an entire step is skipped, we
+            // count it as a passed step.
+            if step_has_error {
+                self.error += 1;
+            } else {
+                self.ok += 1;
+            }
+        }
+    }
+}
+
+/// Formats the status as `E:<error-count>,O:<ok-count>,R:<remaining-count>`, suitable for
+/// annotating the log lines.
+impl std::fmt::Display for Status {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "E:{},O:{},R:{}", self.error, self.ok, self.remaining)
+    }
+}
+
 /// Encapsulates all the local state relative to a step, and is propagated to child steps.
 pub struct Context {
     opt: Opt,
-    prefix: String,
+    prefix: Vec<String>,
 }
 
 impl Context {
     pub fn root(opt: &Opt) -> Self {
         Context {
             opt: opt.clone(),
-            prefix: "".to_string(),
+            prefix: vec![],
         }
     }
 
-    fn child(&self, prefix: &str) -> Self {
+    fn child(&self, name: &str) -> Self {
+        let mut prefix = self.prefix.clone();
+        prefix.push(name.to_string());
         Context {
             opt: self.opt.clone(),
-            prefix: format!("{} ❯ {}", self.prefix, prefix),
+            prefix,
         }
+    }
+
+    fn depth(&self) -> usize {
+        self.prefix.len()
+    }
+}
+
+impl std::fmt::Display for Context {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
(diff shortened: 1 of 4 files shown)


runner/src/internal.rs
```diff
@@ -97,24 +97,70 @@ pub struct BuildServer {
...
+                .count();
+        }
+    }
+}
+
+impl std::fmt::Display for Status {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "F:{},P:{},R:{}",
```
ipetr0v, line 137: Are remaining `R` global or only with the same depth?

runner/src/internal.rs
```diff
@@ -280,7 +353,7 @@ pub async fn run_step(context: &Context, step: Step) -> StepResult {
             let background_stdout_future = tokio::spawn(read_to_end(running_background.stdout()));
             let background_stderr_future = tokio::spawn(read_to_end(running_background.stderr()));
 
-            let mut foreground_result = run_step(&context, *foreground).await;
+            let mut foreground_result = run_step(&context, *foreground, run_status.clone()).await;
```
ipetr0v, line 356: If the `run_status` is changed in the `foreground`, should it influence the `run_status` value (why is it cloned)?
For example if there were failures inside `foreground`.

runner/src/internal.rs
```diff
@@ -97,24 +97,76 @@ pub struct BuildServer {
     pub coverage: bool,
 }
 
+#[derive(Clone)]
```
tiziano88, line 100: I think we usually keep these annotations after the comment, near the actual struct?

runner/src/internal.rs
```diff
@@ -97,24 +97,76 @@ pub struct BuildServer {
...
+/// A construct to keep track of the status of the execution. It only cares about the top-level
+/// steps.
+pub struct Status {
+    failed: usize,
+    passed: usize,
+    remaining: usize,
```
tiziano88, line 106: This terminology seems unnecessarily different from the OK / Error terminology that is already used throughout the code and the logs. FWIW I avoided the pass / fail terminology because not all steps are "tests".

runner/src/internal.rs
```diff
@@ -97,24 +97,76 @@ pub struct BuildServer {
...
+        Status {
+            failed: 0,
+            passed: 0,
+            remaining,
+        }
+    }
+
+    /// Guarantees that the `failed`, `passed`, and `remaining` counts are updated only after the
+    /// completions of each top-level step.
+    fn update(&mut self, context: &Context, step_failed: bool) {
```
tiziano88, line 120: This seems essentially another way of keeping track of the result from each compound step, which is already done in part by 

https://github.com/project-oak/oak/blob/9eb2b43e9b415c203943ada57da1a455a793ea9a/runner/src/internal.rs#L168

I think a more elegant way of doing this would be to compute the top-level list of futures, and check how many have already been resolved, and to what value? I think this makes sense especially seeing that this struct does not do anything apart from the top-level context, but it still needs to be passed around, which seems unnecessary

runner/src/internal.rs
```diff
@@ -97,24 +97,76 @@ pub struct BuildServer {
...
+    }
+}
+
 /// Encapsulates all the local state relative to a step, and is propagated to child steps.
 pub struct Context {
     opt: Opt,
     prefix: String,
+    /// For convenience; to make updating of the `Status` easier.
+    /// It matches the number of atomic step names in the `prefix`.
+    depth: u8,
```
tiziano88, line 153: May be simpler / more elegant to change prefix to a `Vec<String>` with the various segments, and just count them.

runner/src/internal.rs
```diff
@@ -192,8 +263,10 @@ where
 /// Run the provided step, printing out information about the execution, and returning a set of
 /// status results from the single or multiple steps that were executed.
 #[async_recursion]
-pub async fn run_step(context: &Context, step: Step) -> StepResult {
+pub async fn run_step(context: &Context, step: Step, mut run_status: Status) -> StepResult {
```
tiziano88, line 266: It seems that this `run_status` is used as an output parameter, which is usually not idiomatic. As I mentioned elsewhere, there is overlap between this new Status and the StepResult value, I think it makes sense to unify them, and make it look like an actual return value.

runner/src/internal.rs
juliettepretot: I like that idea, it seems nice :)

runner/src/internal.rs
juliettepretot: I like that idea. :) If the futures already give us the information of which steps ran & and with which result it makes a lot of sense to use that, instead of separately keeping track of that status. :)

runner/src/internal.rs
rbehjati: It is not just an output. It has to be an input as well, because it is used for annotating all the log lines with the global state of the execution. If it is changed to a return value, then it can only annotate each log line with the local execution state, which is not very interesting. Actually, in that case, it would not add any information to the log line beyond what is already there. 

So, as I stated in another comment, this variable is keeping a global state, and could be replaced with a global mutable variable. Or it could be part of the context, but I did not want to make `context` mutable. 

Let me know what you think :)

Issue #1294, opened 2020-07-29
Choose client variants in runner


ipetr0v:
Currently `runner` runs all client variants (Rust, C++, etc.) for examples.
It has a `--run-clients`, but currently it's a `bool` used to disable clients.

PR #1298, opened 2020-07-29
Choose client variants in runner

ipetr0v:
This change adds a `client_variant` parameter to `runner` (replaces `run_clients`).

Fixes https://github.com/project-oak/oak/issues/1294

# Checklist

### DIFF
--- a/runner/src/internal.rs
+++ b/runner/src/internal.rs
@@ -69,8 +69,12 @@ pub struct RunExamples {
     pub build_server: BuildServer,
     #[structopt(long, help = "run server [default: true]")]
     pub run_server: Option<bool>,
-    #[structopt(long, help = "run clients [default: true]")]
-    pub run_clients: Option<bool>,
+    #[structopt(
+        long,
+        help = "client variant: [all, rust, cpp, go, nodejs, none] [default: all]",
+        default_value = "all"
+    )]
+    pub client_variant: String,
     #[structopt(long, help = "additional arguments to pass to clients")]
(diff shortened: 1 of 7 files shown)


runner/src/main.rs
```diff
@@ -614,7 +618,7 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
     // clients in the foreground.
     #[allow(clippy::collapsible_if)]
     let run_backend_server_clients: Step = if opt.run_server.unwrap_or(true) {
-        let run_server_clients = if opt.run_clients.unwrap_or(true) {
+        let run_server_clients = if opt.client_variant != "none" {
```
tiziano88, line 621: Could you make `none` a const somewhere?

runner/src/internal.rs
```diff
@@ -69,8 +69,12 @@ pub struct RunExamples {
     pub build_server: BuildServer,
     #[structopt(long, help = "run server [default: true]")]
     pub run_server: Option<bool>,
-    #[structopt(long, help = "run clients [default: true]")]
-    pub run_clients: Option<bool>,
+    #[structopt(
+        long,
+        help = "client variant: [all, rust, cpp, go, nodejs, none] [default: all]",
```
rbehjati, line 74: Just curious... since `all` is an option here, does it also make sense to support any subset of the clients? I think that could be done if the `--client-variant` flag takes a list of the clients.

ipetr0v: Will add in a separate PR

PR #1302, opened 2020-07-30
Add a new module and struct for configuring permitted features

rbehjati:
Ref #1029

### DIFF
--- a/oak/server/rust/oak_runtime/src/permissions.rs
+++ b/oak/server/rust/oak_runtime/src/permissions.rs
@@ -1,0 +1,48 @@
+//
+// Copyright 2020 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/// Provides a declarative description of the features that are permitted
+/// for an Oak application.
+///
+/// To run an Oak application, one must provide, as input to the `oak_loader`, a TOML
+/// file of permissions, which is parsed into an instance of `PermissionsConfiguration`.
+/// The permissions are applied dynamically at runtime, when creating the Nodes.
+#[allow(dead_code)]
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct PermissionsConfiguration {
+    /// Whether to enable the use of GrpcServerNode.
+    #[serde(default)]
+    allow_grpc_server_nodes: bool,
+
+    /// Whether to enable the use of HttpServerNode.
+    #[serde(default)]
+    allow_http_server_nodes: bool,
+
+    /// Whether to enable the use of LogNode.
+    #[serde(default)]
+    allow_log_nodes: bool,
+
+    /// Whether to allow introspection. If true, the introspection auxiliary service will be
+    /// started.
+    #[serde(default)]
+    allow_introspection: bool,
+
+    /// List of the allowed URIs that the application can connect to. Only GrpcClientNodes
+    /// that are configured to connect to URIs in this list can be created.
+    #[serde(default)]
+    allow_grpc_egress_addresses: Vec<String>,
+}
--- a/oak/server/Cargo.lock
+++ b/oak/server/Cargo.lock
@@ -676,6 +676,7 @@ dependencies = [
  "oak_utils 0.1.0",
  "prost 0.6.1",
  "rustls 0.17.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.112 (registry+https://github.com/rust-lang/crates.io-index)",
  "signal-hook 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "structopt 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "tonic 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/oak/server/rust/oak_loader/Cargo.toml
+++ b/oak/server/rust/oak_loader/Cargo.toml
@@ -21,6 +21,7 @@ oak_runtime = "=0.1.0"
 oak_utils = "=0.1.0"
 prost = "*"
 rustls = "*"
+serde = { version = "*", features = ["derive"] }
 signal-hook = "*"
 structopt = "*"
 tonic = { version = "*", features = ["tls"] }
--- a/oak/server/rust/oak_runtime/src/lib.rs
+++ b/oak/server/rust/oak_runtime/src/lib.rs
@@ -69,6 +69,7 @@ mod io;
 mod message;
 mod metrics;
 mod node;
+mod permissions;
 mod proto;
 mod proxy;
 #[cfg(test)]


oak/proto/permissions.proto
```diff
@@ -0,0 +1,59 @@
...
+syntax = "proto3";
+
+package oak.permissions;
+
+// A PermissionsConfiguration provides a declarative description of the features that are permitted
+// for an Oak application.
+//
+// The permissions are applied dynamically at runtime, when creating the Nodes.
+message PermissionsConfiguration {
+  // Whether to enable the use of a GrpcServerNode as the front-door to the application.
```
tiziano88, line 26: Is there a difference between using a gRPC server node as front-door vs non front-door? Maybe it would be simpler to just call the various options like the corresponding pseudo-nodes that they enable?

oak/proto/permissions.proto
```diff
@@ -0,0 +1,59 @@
...
+  bool http_front_door = 2;
+
+  // Whether to allow application logging. If true, allows creating a LogNode for the application.
+  bool logging = 3;
+
+  // Whether to allow monitoring (for the Runtime and the application). If true, the monitoring
+  // auxiliary service will be started.
+  bool monitoring = 4;
+
+  // Whether to allow introspection. If true, the introspection auxiliary service will be started.
```
tiziano88, line 39: I guess this is only for the "debug" builds, not sure whether we need to specify it here though

oak/proto/permissions.proto
```diff
@@ -0,0 +1,59 @@
...
+
+  // List of the allowed domain names or IP addresses that the application can connect to. Only
+  // GrpcClientNodes that are configured to connect to servers in this list can be created.
+  repeated string grpc_client_domains = 6;
+
+  // List of allowed roughtime servers that the application can connect to.
+  repeated string roughtime_servers = 7;
+
+  // Map from Wasm module names to the list of files they are allowed to access.
+  map<string, FileConfig> permitted_files = 8;
```
tiziano88, line 50: I am not sure how this will be used; this configuration should be agnostic of the oak application configuration, so how would someone configure the appropriate wasm module names?

oak_abi/build.rs
```diff
@@ -26,6 +26,7 @@ fn main() {
             "../oak/proto/log.proto",
             "../oak/proto/oak_abi.proto",
             "../oak/proto/roughtime_service.proto",
+            "../oak/proto/permissions.proto",
```
tiziano88, line 29: Do we really expect this file to be used as part of the ABI? I was thinking it would only be used as part of the oak_loader binary?

oak/proto/permissions.proto
conradgrobler: I thought it shoud be OK to allow introstpection for non-debug builds now that the node graph structure can no longer depend on non-public information.

oak/proto/permissions.proto
```diff
@@ -0,0 +1,59 @@
...
+
+  // Whether to allow introspection. If true, the introspection auxiliary service will be started.
+  bool introspection = 5;
+
+  // List of the allowed domain names or IP addresses that the application can connect to. Only
+  // GrpcClientNodes that are configured to connect to servers in this list can be created.
+  repeated string grpc_client_domains = 6;
+
+  // List of allowed roughtime servers that the application can connect to.
+  repeated string roughtime_servers = 7;
```
conradgrobler, line 47: This is potentially problematic. This could allow the owner of the site to specify only a single roughtime server (perhaps one they control), which would mean that the roughtime loses a lot of the trust related to the multi-server approach.

oak/proto/permissions.proto
conradgrobler: I am struggling to think of a use case for limiting access to the contents of a file to only a specific WASM node. Do you have some use cases in mind?

oak/proto/permissions.proto
```diff
@@ -0,0 +1,59 @@
...
+
+syntax = "proto3";
+
+package oak.permissions;
+
+// A PermissionsConfiguration provides a declarative description of the features that are permitted
+// for an Oak application.
+//
+// The permissions are applied dynamically at runtime, when creating the Nodes.
+message PermissionsConfiguration {
```
conradgrobler, line 25: Do you have more information on how this will be used in a document or issue somewhere? I am not clear on why having this as a separate proto (rather than extending the application config) adds security benefits. Is the plan to share this with the client in future as part of remote attestation once it is available?

oak/proto/permissions.proto
rbehjati: I think there will be only one GrpcServerNode in an Oak application. But there could potentially be several HttpServer nodes, for example for application monitoring. Is that right? 

My intention with `grpc_front_door` and `http_front_door` is to allow specifying how the application is exposed to the outside world. In case of the HttpServerNodes, I think we would want to separate the config for the front-door, and the monitoring server node (currently no flag for it here BTW).

Or maybe we just need to allowlist the egress traffic? Again in the case of HTTP servers, we probably need separate allowlist for the front-door and monitoring servers.

oak/proto/permissions.proto
rbehjati: I am not sure either! :)

How about monitoring? Is that also something that can be enabled for non-debug builds? My understanding was that the same concerns apply to monitoring too. 

Does it make sense to allowlist the egress traffic here too (domains/IP to which we send the monitoring or introspection data)?

oak/proto/permissions.proto
rbehjati: I think the idea is that this config is static, and publicly available. So it can be reviewed, and people will know that an app with such a config cannot be trusted. Does that make it better?

oak/proto/permissions.proto
rbehjati: @conradgrobler I don't have a specific use case. But the idea was to have something similar to volumes in k8s deployment configs, where volumes are defined, and each container specifies which of the volumes it wants to mount to. 
@tiziano88 If not referencing the wasm modules, I don't know how else we can make a restriction on the files that a wasm module can use. Why does this config have to be agnostic of the oak application config? Or maybe I have misunderstood the requirement here.

oak_abi/build.rs
rbehjati: It is needed in the Runtime. I have moved it to oak_runtime.

oak/proto/permissions.proto
tiziano88: We should not assume that any remote client would review this config (hence why it is a defense in depth approach). And as @conradgrobler mentioned, we should not let whoever is running the runtime select the list runtime servers that are allowed; the list of servers should be hardcoded in the implementation of the runtime itself (which makes it part of the TCB and the remotely attestable data).

oak/proto/permissions.proto
tiziano88: This is a config that someone who wants to run applications would write once and for all to determine what they are comfortable with, before having an actual application to deploy. It is meant to enable or disable features of the runtime that are not expected to be used in a particular environment. For instance, this would be something that we would show during a launch review, to convince the security team that we are not going to connect to arbitrary websites etc. But this audience would not know much / anything about the oak application that will run on the runtime.

Additionally, since files need to be explicitly passed to the runtime anyways, I don't think this level of defense in depth is particularly useful, it is just going to be a copy of the same list of files that we would pass there.

oak/proto/permissions.proto
rbehjati: As discussed earlier today, it is now a flag for allowing the creation of any grpc server pseudo-node. 
Also as discussed, there is no need to have any limitations on ingress traffic, as it is too difficult to maintain, and the functionality is generally covered by firewalls.

oak/proto/permissions.proto
rbehjati: I removed the flag for monitoring, but the flag for introspection is kept.

oak/proto/permissions.proto
rbehjati: Removed, as discussed.

oak/proto/permissions.proto
rbehjati: Removed, as discussed.

oak/proto/BUILD
```diff
@@ -143,3 +143,8 @@ proto_library(
     srcs = ["handle.proto"],
     deps = ["@com_google_protobuf//:descriptor_proto"],
 )
+
+proto_library(
```
conradgrobler, line 147: nit: this is probably not needed, seeing that it will only be used by the loader and runtime which uses Prost. There is probably no need to make it available via Bazel to other languages.

oak/proto/permissions.proto
```diff
@@ -0,0 +1,41 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak.permissions;
+
+// A PermissionsConfiguration provides a declarative description of the features that are permitted
+// for an Oak application.
```
tiziano88, line 22: Didn't we arrive to the conclusion that this should be a TOML file instead of proto in the end? Otherwise, how is it going to be parsed? Or did you figure out if / how it can be done in Rust using textproto format?

oak/proto/permissions.proto
```diff
@@ -0,0 +1,41 @@
...
+// The permissions are applied dynamically at runtime, when creating the Nodes.
+message PermissionsConfiguration {
+  // Whether to enable the use of a GrpcServerNode.
+  bool allow_grpc_server_nodes = 1;
+
+  // Whether to enable the use of an HttpServerNode.
+  bool allow_http_server_nodes = 2;
+
+  // Whether to allow application logging. If true, allows creating a LogNode for the application.
+  bool allow_logging = 3;
```
tiziano88, line 33: Since there is runtime logging and application logging, and this is really about the log pseudo node, why not just keep a consistent terminology and make the bools refer to actual node types? it seems simpler?

oak/proto/permissions.proto
rbehjati: I might be missing something here, but I think it can be a protobuf, the same way we have `application.proto`, but the actual permission files that we use as input to the `oak_loader` will be toml files, the same way we now have a `config.toml` for each example. Alternatively, this could be a simple `struct`, but I thought having it as a protobuf is more consistent with the rest of the codebase.

oak/proto/permissions.proto
rbehjati: Good point. I changed the name. 

`allow_introspection` and `allowed_grpc_egress_addresses` do not directly refer to a node type. So, I did not change them. But let me know if you can think of better names :)

oak/proto/permissions.proto
tiziano88: since we need to define the actual struct for TOML deserialization (something like [this](https://github.com/project-oak/oak/blob/bfc2a023dba2611e115f647d79369f0e31ceaf80/runner/src/main.rs#L582-L589)), I don't see how the protobuf helps with anything?

rbehjati: Replaced the protobuf with a struct.

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -100,6 +100,37 @@ impl FromStr for ConfigEntry {
...
 
+/// A PermissionsConfiguration provides a declarative description of the features that are permitted
+/// for an Oak application.
+///
+/// The permissions are applied dynamically at runtime, when creating the Nodes.
+#[allow(dead_code)]
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Permissions {
+    /// Whether to enable the use of a GrpcServerNode.
```
tiziano88, line 111: (nit) comment talks about single node, field name refers to multiple nodes.

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -100,6 +100,37 @@ impl FromStr for ConfigEntry {
     }
 }
 
+/// A PermissionsConfiguration provides a declarative description of the features that are permitted
+/// for an Oak application.
```
tiziano88, line 104: Maybe mention somewhere that this is a parsed file, and what format, etc.

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -100,6 +100,37 @@ impl FromStr for ConfigEntry {
...
+///
+/// The permissions are applied dynamically at runtime, when creating the Nodes.
+#[allow(dead_code)]
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Permissions {
```
tiziano88, line 110: Are you sure this should be in oak_loader? I was expecting it to be part of oak_runtime? Otherwise we may end up with a circular dep. I may be wrong though.

oak/server/rust/oak_runtime/src/permissions.rs
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/// Provides a declarative description of the features that are permitted
+/// for an Oak application.
+///
+/// To run an Oak application, one must provide, as input to the `oak_loader`, a toml
```
tiziano88, line 20:
```suggestion
/// To run an Oak application, one must provide, as input to the `oak_loader`, a TOML
```

oak/server/rust/oak_runtime/src/permissions.rs
```diff
@@ -0,0 +1,48 @@
...
+    allow_http_server_nodes: bool,
+
+    /// Whether to enable the use of LogNode.
+    #[serde(default)]
+    allow_log_nodes: bool,
+
+    /// Whether to allow introspection. If true, the introspection auxiliary service will be
+    /// started.
+    #[serde(default)]
+    allow_introspection: bool,
```
tiziano88, line 42: I'm still not entirely sure who / how would set this, but I guess we can keep it around and see what use cases we find for it.

rbehjati: My understanding from the discussions yesterday is that `introspection` is going to be allowed in non-debug mode. But in the non-debug mode, we need a mechanism to remove (or anonymise?) messages in the channels. I don't know if that part is already there or not. 

This flag only says if the introspection server will actually be started. Does this make sense?

PR #1309, opened 2020-07-31
Record introspection events in a Runtime queue to be consumed introspection server

juliettepretot:
~It took me some trial & and error to wrap my head around channels, threads etc, but I'm quite happy that I got it to work. :)~

See #913 for context.

# Checklist

      construction.


oak/server/rust/oak_runtime/src/runtime/introspect.rs
```diff
@@ -150,10 +151,11 @@ async fn make_server(
     info!("starting introspection server on {:?}", addr);
 
     // And a MakeService to handle each connection...
+    let runtime_clone = runtime.clone();
     let make_service = make_service_fn(move |_| {
-        // The `Arc<Runtime>` is moved into this closure, but needs to be cloned
+        // The runtime_clone is moved into this closure, but needs to be cloned
         // because this closure is called for every connection.
-        let runtime = runtime.clone();
+        let runtime = runtime_clone.clone();
```
juliettepretot, line 158: I'll admit this isn't the most beautiful naming, but essentially I need to clone the runtime initially, so that `runtime` isn't moved into the closure and can be used later for the introspection events.

juliettepretot:
(this currently still prints warnings when trying to send events before the introspection server has booted up, looking into it)

oak/server/rust/oak_runtime/src/runtime/proxy.rs
```diff
@@ -61,6 +62,7 @@ impl RuntimeProxy {
         application_configuration: &ApplicationConfiguration,
         grpc_configuration: &GrpcConfiguration,
     ) -> RuntimeProxy {
+        let (introspection_event_sender, _) = broadcast::channel(100);
```
tiziano88, line 65: Please make this a const with a meaningful name, and ideally also some comment :)

oak/server/rust/oak_runtime/src/runtime/mod.rs
```diff
@@ -248,6 +248,8 @@ pub struct Runtime {
 
     aux_servers: Mutex<Vec<AuxServer>>,
 
+    introspection_event_sender: broadcast::Sender<Event>,
```
tiziano88, line 251: comment pls

oak/server/rust/oak_runtime/src/runtime/introspect.rs
```diff
@@ -170,6 +172,24 @@ async fn make_server(
         let _ = termination_notificiation_receiver.await;
     });
 
+    // Listen for introspection events
+    let introspection_event_receiver = runtime.introspection_event_sender.clone().subscribe();
```
tiziano88, line 176: Could you elaborate on why the clone call is necessary, in a comment? Also, could this be moved inside the async closure below?

oak/server/rust/oak_runtime/src/runtime/introspection_events.rs
```diff
@@ -44,9 +44,10 @@ impl Runtime {
             event_details: Some(event_details),
         };
 
-        // TODO(#913): Push the event over to a channel, to be consumed by the
-        // introspection aux server.
-        debug!("Introspection event recorded: {:?}", event);
+        let sender = self.introspection_event_sender.clone();
+        if let Err(error) = sender.send(event) {
+            warn!("Failed to send introspection event to channel: {:?}", error);
+        }
```
tiziano88, line 50: Why is this cloning the sender for each new event that needs to be broacast?

oak/server/rust/oak_runtime/src/runtime/introspect.rs
juliettepretot: Actually the clone is not needed, moved it inside of the closure. :)

oak/server/rust/oak_runtime/src/runtime/introspection_events.rs
juliettepretot: Also not needed, I think I carried over the cloning from another channel lib, which required it & didn't double check whether it was still needed.

oak/server/rust/oak_runtime/src/runtime/proxy.rs
juliettepretot: Done! Just to make sure understand you correctly, the suggestion is to make the argument (`100`) of the `broadcast::channel` function as const? I also added a comment to explain that value.

oak/server/rust/oak_runtime/src/runtime/mod.rs
juliettepretot: Added! Please let me know if this is the type of comment you were looking for.

juliettepretot:
@tiziano88 Thanks for your review, addressed the comments. :]

The remaining issue in this PR is the first few messages failing to be sent (no receivers) prior to the aux server having started up. 

I have some more less qualified ideas to address it (secondary queue if there are no receivers yet), but maybe it's best to discuss that in our meeting tomorrow. :)

juliettepretot:
One of option to the aforementioned problem of missed early events would be to maintain an unused initial receiver, thereby making the broadcast channel maintain a queue of past events. A bit hacky, but could work. 

See how it looks in code here: https://github.com/juliettepretot/oak/commit/6a70d6129286487796cd0379c8888416fbd8c055

juliettepretot:
Chatted about the first lost messages issues irl with @tiziano88 & @daviddrysdale, we decided to refactor to using a `VecDeque` to store events on the runtime for now. There's some potential scalability concerns to that approach, but esp since this is debug only it's more pragmatic for now. :)

tiziano88:
@juliettepretot just to make sure we are not in a deadlock: are you going to modify the PR so that it uses a `VecDeque`, or should I review it as it is? Thanks!

juliettepretot:
@tiziano88 Will modify it. :) Was busy last week, but my plan is still to update this PR. Will remove the WIP label & rerequest your review once it's ready.

oak/server/rust/oak_runtime/src/introspection_events.rs
```diff
@@ -44,9 +43,10 @@ impl Runtime {
             event_details: Some(event_details),
         };
 
-        // TODO(#913): Push the event over to a channel, to be consumed by the
-        // introspection aux server.
-        debug!("Introspection event recorded: {:?}", event);
+        self.introspection_event_queue
+            .lock()
+            .unwrap()
+            .push_back(event);
```
juliettepretot, line 49: Is it bad idea to call `lock` from within the Runtime? If the introspection aux server were to currently hold the lock for that `Mutex` it seems we'd block the Runtime, which seems troublesome. 

I could spawn a thread to mitigate this.

juliettepretot:
@tiziano88 ready :)

oak/server/rust/oak_runtime/src/proxy.rs
```diff
@@ -67,6 +68,7 @@ impl RuntimeProxy {
             node_infos: RwLock::new(HashMap::new()),
             next_node_id: AtomicU64::new(0),
             aux_servers: Mutex::new(Vec::new()),
+            introspection_event_queue,
```
daviddrysdale, line 71: nit: might as well be consistent with the other fields and have `introspection_event_queue: Mutex::new(VecDeque::new()),` inline.

oak/server/rust/oak_runtime/src/introspection_events.rs
daviddrysdale: Are you expecting the introspection aux server to hold the lock for a long period (e.g. while a user is scrubbing backwards and forwards through the events)?  If so, then you're right we'd need something to prevent it blocking the Runtime.

OTOH if the introspection code is only going to grab the lock momentarily> then it should be OK.  (However, that might in turn assume that the ordering of events in the `VecDeque` is stable, e.g. that it's OK to lock/read-entry-17/unlock  then lock/read-entry-16/unlock and have the numbering not change underneath.)

juliettepretot: For now it should _probably_ be fine. There's a point to be made that responding to scrubbing should probably be done client side, in which case past events would be serialized to JSON upon requesting `/graph`. Hence the lock should be temporary. 

For now we should be ok with ordering, since the order of the queue is of infinite length & chronological. However admittedly that might change down the road.

juliettepretot:
Found this on a secret internal memepage :D 
image

(It's funny but agree that pragmatic solutions are better)

Issue #1304, opened 2020-07-30
unable to evaluate symlinks in Dockerfile path: lstat /workspace/android.Dockerfile: no such file or directory


tiziano88:
This keeps happening more and more often in CI jobs, perhaps some interference between the android and non-android build steps:

https://pantheon.corp.google.com/cloud-build/builds/4d64b51b-5c64-4ff5-b136-3b633b35c39d?project=oak-ci

```log
Step #8 - "build_android_image": unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /workspace/android.Dockerfile: no such file or directory
```

tiziano88:
From what @rbehjati and @juliettepretot investigated, this appears to be an issue in GCB when retriggering workflows from GitHub. I don't think there is anything we should do, apart from notifying the GCB team (which @juliettepretot has already done) and waiting for a fix there. Assigning to @juliettepretot just so we don't forget to update this once we get a reply from the GCB team.

juliettepretot:
Thanks for opening this issue to track. Current status is that the GCB folks confirmed its a known bug. Hopefully it'll be patched soon.

PR #1310, opened 2020-07-31
Disable liche

rbehjati:
Tried to disable liche with minimum change to the code... Is there a better way to do this?

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -419,7 +419,8 @@ fn check_format() -> Step {
             run_prettier(FormatMode::Check),
             run_markdownlint(FormatMode::Check),
             run_embedmd(FormatMode::Check),
-            run_liche(),
+            // TODO(#1304): Uncomment, when re-run from GitHub is fixed.
+            // run_liche(),
             run_cargo_fmt(FormatMode::Check),
             run_hadolint(),
             run_shellcheck(),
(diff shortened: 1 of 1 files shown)


benlaurie:
Why are you deleting FORCE_CI?

rbehjati:
> Why are you deleting FORCE_CI?

I can put it back! I thought now that I have added a TODO for #1304, `FORCE_CI` is not longer needed to remind us about the issue :) Also, since `liche` is removed, it is less likely that the build would fail out of flakiness.

runner/src/main.rs
```diff
@@ -1098,7 +1103,9 @@ fn run_liche() -> Step {
     Step::Multiple {
```
tiziano88, line 1103: Isn't it more readable to just comment out the lines that call to `run_liche` and add the TODO there to uncomment them?

rbehjati: Oh! Interesting. I can comment out the whole `run_liche` function, without getting any `clippy` complaints (I am used to style checks complaining about commented out **code**).

runner/src/main.rs
```diff
@@ -1094,27 +1095,28 @@ fn run_embedmd(mode: FormatMode) -> Step {
...
-                    // - https://crates.io/crates : returns 404 (see https://github.com/raviqqe/liche/issues/39)
-                    // - https://codecov.io/gh/project-oak/oak : unstable, and often unaccessible on CI-build
-                    "--exclude=(https://groups.google.com/g/project-oak-discuss|https://crates.io/crates|https://codecov.io/gh/project-oak/oak)",
-                    &entry,
-                ]),
-            })
-            .collect(),
-    }
-}
+// TODO(#1304): uncomment, when re-run from GitHub is fixed.
```
tiziano88, line 1098: I would just add an ignore attribute for clippy instead of commenting out the entire function, so at least it keeps being compiled correctly, but up to you.

rbehjati: Thanks. Done :)

Issue #1274, opened 2020-07-20
Wasm module signature label


tiziano88:
One of the ways that Oak allows for delegation is in the choice of declassification modules. 

Currently it is possible to label data with a label referring to the hash of an individual Wasm module with the following:

https://github.com/project-oak/oak/blob/1b3b4b22a8c8bd29f9b567970142065de56f35a0/oak/proto/label.proto#L58-L65

But in many cases it is not possible or desirable to enumerate the list of allowed Wasm modules upfront at label creation time. Oak should provide a way of specifying a family of Wasm modules.

This proposal is to allow cryptographic signatures of Wasm module to be used as labels.

A possible implementation would add a tag such as:

```proto
message WebAssemblyModuleSignature {
  bytes public_key = 1;
}
```

in which the public key may be a `minisign` key: https://jedisct1.github.io/minisign/ or some other public key suitable for signing arbitrary data.

Any Wasm module signed with the provided key would be granted the privilege to declassify tags referring to that key.

Key distribution and discovery is a separate problem, i.e. it needs to happen out-of-band. How keys are published and obtained (and also revoked) is out of scope for this issue, but should probably be investigated more closely in a separate issue at some point once we have a better idea.

ipetr0v:
What if we need the data to be processed by two nodes **A** and **B**, signed by the same party?

What would the label look like if we will not use exact module hashes?

If it will look like `google_public_key ∧ google_public_key`, then it will just merge into `google_public_key`, and it will be possible for Oak to release the data without processing it by the node **B**.

ipetr0v:
I think the label should all include the name of the module:
```protobuf
message WebAssemblyModuleSignature {
  string name = 1;
  bytes public_key = 2;
}
```

tiziano88:
> What if we need the data to be processed by two nodes **A** and **B**, signed by the same party?
> 
> What would the label look like if we will not use exact module hashes?
> 
> If it will look like `google_public_key ∧ google_public_key`, then it will just merge into `google_public_key`, and it will be possible for Oak to release the data without processing it by the node **B**.

That should be the same as having only one module signed by that party? I don't think there is an intuitive meaning of "needs to be declassified by two arbitrary modules signed by google.com, but really not just one, there must be exactly two of them"?

ipetr0v:
It could be just two different modules that serve different purpose.
For example, a third-party developer wants to use already existing Google modules, and he will not be able to ask Google to merge these two modules into a single one.

[...]

PR #1312, opened 2020-08-02
Add Wasm module signature label

ipetr0v:
This change adds:
- Wasm module ed25519 signature labels
- Adds the ability to add ed25519 signatures and public keys to the application ocnfiguration

Fixes https://github.com/project-oak/oak/issues/1274

# Checklist


examples/hello_world/config/config.toml
```diff
@@ -1,5 +1,5 @@
 name = "hello_world"
 
 [modules]
-app = { path = "examples/hello_world/bin/hello_world.wasm" }
+app = { path = "examples/hello_world/bin/hello_world.wasm", signature = "" }
```
tiziano88, line 4: I don't expect the signatures to be part of the application configuration; roughly, I think the application stays as it is, but separately, we provide a list of signatures to the oak_loader, which comes from somewhere else. the idea is that eventually the signatures would come from a verifiable log or anywhere else on the internet, they are not part of the modules nor the application. this would also allow for instance multiple parties to sign the same module, and publish their own signatures in whatever way they prefer.

ipetr0v: But we need to somehow specify a public key and the signature value for a particular Wasm module

ipetr0v: Currently I was thinking about this configuration:
```
name = "trusted_information_retrieval"

[modules.app]
location = { path = "examples/trusted_information_retrieval/bin/trusted_information_retrieval.wasm" }
signature = { public_key = "examples/certs/label/label.pub", value = "examples/hello_world/bin/trusted_information_retrieval.wasm.sign" }

[modules.database_proxy]
location = { external = { url = "https://storage.googleapis.com/oak-modules/database_proxy/3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f", sha256 = "3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f" } }
```

tiziano88: - I think the signature should just be over the hash of the module (maybe in the future we can add additional "tags" or names to a module, but let's ignore the module name for now perhaps).
- a signature would be a (protobuf?) object containing the hash being signed, the public key, and the signature itself.
- the oak_loader is provided with a file (or files, or URLs) of such signatures, and loads them at startup
- when instantiating new modules, the oak runtime looks to see if the module is signed, and by what public key, and verifies the signature against that key (using the minisign format)
- one desired outcome is that if A wants to sign B's module, A can do so without having to re-package B's module; A only needs to publish A's signature over B's module hash

ipetr0v: Personally I'm leaning towards either having a `.toml` file or packing everything into a config file.

tiziano88:
Are you sure we should be using the minisign crate? It seems bad that it does not depend on any standard crypto crate, so if we use it, I think we should only use it for parsing, not for verifying signatures. WDYT?

ipetr0v:
I'm currently using `minisign` because [`ed25519_dalek`](https://docs.rs/ed25519-dalek/0.9.1/ed25519_dalek/) couldn't parse keys and signatures, created by `minisign`.
I think we can keep `minisign` for now, since we are using it to create signatures anyway.

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -421,6 +432,26 @@ impl Runtime {
     pub fn gather_metrics(&self) -> Vec<MetricFamily> {
         self.metrics_data.gather()
     }
+
+    /// Check Wasm module signatures.
+    pub(crate) fn verify_module_signatures(&self) -> Result<(), OakStatus> {
+        for (name, module) in &self.application_configuration.wasm_modules {
+            // Get signature by module name.
+            if let Some(signature_item) = self.signature_configuration.get(name) {
```
rbehjati, line 440: Looks like having a signature is optional. Maybe add a comment somewhere explaining what that means?

docs/development.md
```diff
@@ -245,3 +245,15 @@ I0511 10:15:29.585389 244858 hello_world.cc:50] Request: WORLDS
 I0511 10:15:29.591434 244858 hello_world.cc:57] Response: HELLO WORLDS!
 I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
 ```
+
+### Signing Oak Wasm modules
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. The resulting signatures can be used by Oak clients as IFC labels.
```
tiziano88, line 252: I thought the public keys can be used as labels?

examples/certs/label/label.key
```diff
@@ -0,0 +1,2 @@
+untrusted comment: minisign encrypted secret key
```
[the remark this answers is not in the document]
tiziano88: +1

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -80,6 +87,20 @@ pub struct Opt {
     config_files: Vec<ConfigEntry>,
 }
 
+#[derive(Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct SignatureManifest {
+    name: String,
+    signatures: HashMap<String, Signature>,
```
tiziano88, line 94: Please clarify what the keys of this map are.

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -93,6 +93,7 @@ pub fn create_node(
     application_configuration: &ApplicationConfiguration,
     node_configuration: &NodeConfiguration,
     grpc_configuration: &GrpcConfiguration,
+    signature_configuration: &HashMap<String, SignatureConfiguration>,
```
tiziano88, line 96: I think the term "configuration" is already overloaded, how about signature database / repository / list / map / table or something like that?

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -798,21 +801,29 @@ impl WasmNode {
...
             .map_err(ConfigurationError::WasmiModuleInializationError)?;
         let entrypoint_name = node_configuration.wasm_entrypoint_name;
         validate_entrypoint(&module, &entrypoint_name).map_err(|err| {
             warn!("could not validate entrypoint: {:?}", err);
             ConfigurationError::IncorrectWebAssemblyModuleName
         })?;
+
+        let public_key = signature_configuration
+            .get(&node_configuration.wasm_module_name)
+            .map(|signature_item| signature_item.public_key.to_bytes());
```
tiziano88, line 821: Each module may have more than one signature, right? So I think this should return a Vec, which would also make sense to replace the Option with a Vec in the `wasm_node_privilege` method.

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -824,15 +835,26 @@ impl WasmNode {
 
 /// Computes the [`NodePrivilege`] granted to a WebAssembly Node running the specified WebAssembly
 /// module.
```
tiziano88, line 837: Update doc

oak/server/Cargo.lock
```diff
@@ -1126,15 +1340,27 @@ dependencies = [
 
 [[package]]
 name = "sha2"
-version = "0.9.0"
+version = "0.8.2"
```
tiziano88, line 1343: Why was this downgraded? I think the API is also different, I remember actually upgrading it to the new API at some point.

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -55,6 +57,11 @@ pub struct Opt {
         )
     )]
     root_tls_certificate: Option<String>,
+    #[structopt(
+        long,
+        help = "Path to a Toml file containing information about Oak modules' signatures."
```
tiziano88, line 62:
```suggestion
        help = "Path to a TOML file containing information about Oak modules' signatures."
```

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -421,6 +432,26 @@ impl Runtime {
...
+                let public_key = &signature_item.public_key;
+                let signature = &signature_item.signature;
+                let data_reader = std::io::Cursor::new(module);
+                verify(public_key, signature, data_reader, true, false).map_err(|error| {
```
tiziano88, line 444: could you extract the magic true and false into named consts in this scope?

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -824,15 +835,26 @@ impl WasmNode {
...
+
+    match wasm_module_public_key {
+        Some(public_key) => {
+            let signature_tag = oak_abi::label::web_assembly_module_signature_tag(&public_key);
+            NodePrivilege::new(
+                hashset! { hash_tag.clone(), signature_tag.clone() },
+                hashset! { hash_tag, signature_tag },
+            )
+        }
+        None => NodePrivilege::new(hashset! { hash_tag.clone() }, hashset! { hash_tag }),
```
tiziano88, line 856: This does not seem very elegant, one branch is a copy of the other plus more stuff. Perhaps create two separate sets, and then join then together at the end? It would also make the multiple signatures case easier to handle.

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -80,6 +87,20 @@ pub struct Opt {
...
+#[derive(Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct SignatureManifest {
+    name: String,
+    signatures: HashMap<String, Signature>,
+}
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct Signature {
```
tiziano88, line 99: How is this different from SignatureConfiguration ? could you add comments to clarify?

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
ipetr0v: Done

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
ipetr0v: Done

oak/server/Cargo.lock
ipetr0v: Upgraded back
Downgraded version was needed to rust [ed25519](https://docs.rs/ed25519-dalek/0.9.1/ed25519_dalek/), but it's not used anymore

docs/development.md
```diff
@@ -278,3 +278,15 @@ I0511 10:15:29.585389 244858 hello_world.cc:50] Request: WORLDS
 I0511 10:15:29.591434 244858 hello_world.cc:57] Response: HELLO WORLDS!
 I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
 ```
+
+### Signing Oak Wasm modules
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
```
rbehjati, line 284: Is it explained anywhere how these signatures are used (for example for determining the privileges, if I understand correctly)? 

There was also a comment from Tiziano about modules with multiple signatures. Is it possible to make multiple signatures with this command or would that require a different command? 

There is some interesting discussion and description of what these signatures are for in #1274. I think it would be nice to have a summary documented somewhere in our docs, if it is not already there.

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
```diff
@@ -824,15 +835,26 @@ impl WasmNode {
 
 /// Computes the [`NodePrivilege`] granted to a WebAssembly Node running the specified WebAssembly
 /// module.
-fn wasm_node_privilege(wasm_module_bytes: &[u8]) -> NodePrivilege {
+/// Created [`NodePrivilege`] consists of Wasm module hash and signature.
+fn wasm_node_privilege(wasm_module_bytes: &[u8], signatures: Vec<Signature>) -> NodePrivilege {
```
tiziano88, line 839: It seems this should be able to just get a `&[Signature]` argument, which would be more consistent and also let you avoid the cloning on line 820.

oak/server/testdata/key.pub
```diff
@@ -0,0 +1,2 @@
+untrusted comment: minisign public key 199FF2AA1C6A7CDD
```
tiziano88, line 1: Where is the corresponding private key? How do these need to be re-generated if things change? Please document things so that if anyone else had to make any change and you are not around, they would be able to find their way around.

oak_abi/src/label/mod.rs
```diff
@@ -93,6 +93,20 @@ pub fn web_assembly_module_tag(web_assembly_module_hash_sha_256: &[u8]) -> Tag {
     }
 }
 
+/// Creates a [`Tag`] having as principal the provided WebAssembly module Ed25519 signature.
```
tiziano88, line 96: I think the pricipal is the key, not the signature.

examples/trusted_information_retrieval/signature.toml
```diff
@@ -0,0 +1,5 @@
+name = "trusted_information_retrieval"
+
+[[signatures.database_proxy]]
```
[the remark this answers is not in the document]
ipetr0v: I have added a module name in order to know, which keys correspond to which modules.
Otherwise we would need to check all the modules with all the keys.

I can change module names to module hashes.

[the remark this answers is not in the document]
tiziano88: My question is why is the module name part of the struct? If you just need it as a reminder, it can just be a comment. Also the signature path on line 5 already has the module name in the path itself it seems (though perhaps that will change to a hash too?)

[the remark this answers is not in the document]
ipetr0v: It's not just a reminder, it's used by the `oak_runtime` for verifying Wasm modules.
It the `oak_runtime` doesn't have information about which key corresponds to a specific module, then it would need to check all the keys for every single module.

> Also the signature path on line 5 already has the module name ...
Yes, but it's only happened, because we have chosen the examples to be in this directory. Not all applications will contain such an information.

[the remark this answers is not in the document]
tiziano88: Another problem with having the name as map key here is that, IIUC, you can still only have at most one signature per module, which we discussed is too limiting.

If we had the hash here, we could have a comment that says "database_proxy" in the TOML file, and I think that will serve the purpose. The runtime will not use the name for anything, and signing modules does not involve knowing their name, or their name being consistent.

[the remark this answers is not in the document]
ipetr0v: I've already updated this, it accepts multiple signatures per module.
In this file double square brackets `[[` mean that this is an array.
So it may looks like this:
```toml
[[signatures.database_proxy]]
...

[[signatures.database_proxy]]
...
```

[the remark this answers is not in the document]
ipetr0v: I'll change it into a hash, and use comments for module names.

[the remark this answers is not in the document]
tiziano88: ah got it, thanks, I wasn't familiar with the syntax! still the other points stand :) maybe we can use the hash as part of the array, but have a comment with the name. basically I don't think the runtime nor the signing process should even know about the name. WDYT?

examples/trusted_information_retrieval/example.toml
```diff
@@ -17,6 +17,7 @@ module_1 = { Cargo = { cargo_manifest = "examples/trusted_information_retrieval/
 [server]
 additional_args = [
   "--config-files=config=examples/trusted_information_retrieval/database.toml",
+  "--signature-manifest=examples/trusted_information_retrieval/signature.toml"
```
tiziano88, line 20: In general, this file name contains signatures (plural), right? So not sure why it's called signature (singular).

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -80,6 +87,22 @@ pub struct Opt {
     config_files: Vec<ConfigEntry>,
 }
 
+#[derive(Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct SignatureManifest {
+    application_name: String,
```
tiziano88, line 93: What is this used for? I cannot think of anything useful for it. Also, the signatures should be agnostic of the actual application, right? Or are you imagining they work only for a specific application? If only for a specific application, why is the application identified by its name only? It does not seem to offer any security guarantee

oak/server/rust/oak_loader/src/options.rs
```diff
@@ -181,6 +208,42 @@ fn create_grpc_config(opt: &Opt) -> anyhow::Result<oak_runtime::GrpcConfiguratio
     Ok(grpc_config)
 }
 
+/// Create a signature table for Oak runtime.
+/// Returns a `HashMap` that maps each module hash to a vector of [`oak_runtime::Signature`].
```
tiziano88, line 212: Mention that the signatures are not verified (AFAICT)?

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -108,6 +112,14 @@ pub struct GrpcConfiguration {
     pub grpc_client_root_tls_certificate: Option<Certificate>,
 }
 
+/// Configuration options related to Wasm module signatures.
+/// Consists of parsed `minisign::PublicKey` and `minisign::SignatureBox`.
```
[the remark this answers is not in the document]
tiziano88: Probably worth a comment saying that the signatures in this struct are not guaranteed to have been verified already? Otherwise if anyone assumes that they are verified, it will break the entire trust assumptions.

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -424,6 +437,44 @@ impl Runtime {
     pub fn gather_metrics(&self) -> Vec<MetricFamily> {
         self.metrics_data.gather()
     }
+
+    /// Check Wasm module signatures.
```
tiziano88, line 441:
```suggestion
    /// Checks Wasm module signatures.
```

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -424,6 +437,44 @@ impl Runtime {
...
+            // Get signature by module name.
+            if let Some(signatures) = self.signature_table.get(&module_hash) {
+                for signature_item in signatures.iter() {
+                    let public_key = &signature_item.public_key;
+                    let signature = &signature_item.signature;
+                    let data_reader = std::io::Cursor::new(module);
+
+                    // Output status information to `stderr`.
+                    // https://docs.rs/minisign/0.5.20/minisign/fn.verify.html#arguments
+                    const STATUS_TO_STDERR: bool = false;
```
tiziano88, line 456: Not sure where these names come from, AFAICT the params are `quiet` and `output`, so actually they should probably be `true` and `false`, but the names need to correspond to the actual parameters. See https://docs.rs/minisign/0.5.20/minisign/fn.verify.html

```rust
const QUIET: bool = true;
const OUTPUT: bool = false;
```

Or is there any reason why that would not work here?

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -424,6 +437,44 @@ impl Runtime {
...
+                        data_reader,
+                        STATUS_TO_STDERR,
+                        DATA_TO_STDOUT,
+                    )
+                    .map_err(|error| {
+                        error!(
+                            "Wasm module signature verification failed for {}: {}",
+                            name, error
+                        );
+                        OakStatus::ErrInvalidArgs
```
tiziano88, line 471: Not sure we want to fail this way long term; it's fine for now, but worth thinking about it in a future PR.

oak/server/rust/oak_runtime/src/node/wasm/mod.rs
tiziano88: I think it's fine to unwrap_or_default in the calling code, but I don't see why this arguments needs to be a Vec, it should be fine with `&[Signature]` and use `.as_ref` on the vec, even for empty vec? https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-AsRef%3C%5BT%5D%3E

oak/server/rust/oak_runtime/src/node/wasm/tests.rs
```diff
@@ -25,17 +25,32 @@ use oak_abi::{
 };
 use wat::{parse_file, parse_str};
 
-fn start_node(wasm_module: Vec<u8>, entrypoint_name: &str) -> Result<(), OakStatus> {
+fn start_node(
+    wasm_module: Vec<u8>,
+    entrypoint_name: &str,
+    signature: Option<Signature>,
```
tiziano88, line 31: Seems simpler and also more realistic to just pass `Vec<Signature>` (or `&[Signature]`).

oak_abi/proto/label.proto
```diff
@@ -65,6 +66,12 @@ message WebAssemblyModuleTag {
   bytes web_assembly_module_hash_sha_256 = 1;
 }
 
+// Policies related to modules, referring to the signature of the native WebAssembly node.
+message WebAssemblyModuleSignatureTag {
+  // Public key counterpart of the private key used to sign WebAssembly module.
+  bytes public_key = 1;
```
tiziano88, line 72: Please add some more details about the format of such key.

oak/server/testdata/minimal.sign
```diff
@@ -0,0 +1,4 @@
+untrusted comment: signature from minisign secret key
+RWTdfGocqvKfGWhWjqcrQDSBUwIn3yP3P9g0kb1xknK3ir/iifo1uFWzFRp+m/AkZr9Nk+ygR/A0vb9e4kswc2iQ2ry2uBKEKwE=
+trusted comment: timestamp:1597152961	file:minimal.wasm
```
tiziano88, line 3: Do we require rebuilding those wasm files now, if the wat files change? Could you confirm this is already documented / automated?

oak_abi/proto/label.proto
ipetr0v: Minisign doesn't specify the serialization format. Looks like they just concatenate the hash algorithm, keynum and the key itself:
https://github.com/jedisct1/rust-minisign/blob/1d909fa0272b7177cb86717f280f104de2658fbc/src/public_key.rs#L100-L116

Will add a comment that it's a serialized `minisign` key

oak/server/rust/oak_runtime/src/node/wasm/tests.rs
ipetr0v: Done

oak/server/rust/oak_runtime/src/lib.rs
ipetr0v: Do you mean `ErrInvalidArgs` or just erroring with an `OakStatus`?

tiziano88: Either way, don't have a strong opinion yet.

oak/server/rust/oak_runtime/src/node/mod.rs
```diff
@@ -93,6 +93,7 @@ pub fn create_node(
     application_configuration: &ApplicationConfiguration,
     node_configuration: &NodeConfiguration,
     grpc_configuration: &GrpcConfiguration,
+    signature_table: &SignatureTable,
```
tiziano88, line 96: FYI a way of making this more flexible (e.g. allowing lookup of signatures from internet) may be to replace this struct with a lookup function (or trait) that given the module hash, returns potential signatures (which the runtime then independently verifies). Then the current implementation may correspond to a lookup function that uses the signature table, but a different one may do HTTP requests or other things to actual return that. I think it would be a nice way of abstracting this behaviour. WDYT? (for another PR of course, but perhaps worth creating an issue to track, if you think it's a good idea? then we can also have a flag to oak_loader to choose which lookup mechanism to use, so we can switch between the two)

oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -1184,3 +1238,12 @@ impl Runtime {
             .len() as i64
     }
 }
+
+/// Computes a Sha256 digest of `bytes` and returns it in a hex encoded string.
```
tiziano88, line 1242:
```suggestion
/// Computes a SHA-256 digest of `bytes` and returns it in a hex encoded string.
```

for consistency with some of your own other comments

oak/server/testdata/minimal.sign
tiziano88: Either way seems fine, but it needs to be clear and automated. I would like to keep the `wat` files around to show how the `wasm` files are generated in any case.

scripts/sign_module
```diff
@@ -0,0 +1,46 @@
...
+
+Options:
+  -i    Input Wasm file
+  -o    Output signature file
+  -h    Print Help (this message) and exit"
+      exit 0;;
+    i)
+      readonly INPUT_FILE="${OPTARG}";;
+    o)
+      readonly OUTPUT_FILE="${OPTARG}";;
```
tiziano88, line 24: FWIW this script seems already complex enough that it may be easier to replace it with an actual small Rust binary, and then we also get the benefit that it can just generate self-contained signature files (including the hash, the public key, and the signature, in a single file). Initially it can literally spawn out to `minisign` if needed, or use the `minisign` crate if it works and can rely on the existing private key. Could you create a PR to track?

tiziano88 (approved):

oak/server/testdata/minimal.sign
ipetr0v: Done, changed to loading `.wasm` files (keeping `.wat`)

PR #1318, opened 2020-08-03
Move most runtime files up one level

tiziano88:
This avoids the `oak_runtime/runtime` stutter, and I think makes more
sense anyways.

# Checklist

      construction.


oak/server/rust/oak_runtime/src/lib.rs
```diff
@@ -74,3 +106,1084 @@ pub struct GrpcConfiguration {
     // TODO(#999): Remove user-configurable root CA.
     pub grpc_client_root_tls_certificate: Option<Certificate>,
 }
+
+struct NodeStopper {
```
tiziano88, line 110: This is all copy-pasted verbatim from `oak_runtime/src/runtime.mod`

Issue #1303, opened 2020-07-30
Self-sign a TLS certificate for oakexamples.dev


rbehjati:
While waiting for provisioning of a TLS certificate from Google, we can get a self-signed certificate to continue with rest of the configuration, and testing of the APIs.

rbehjati:
Fixed in #1319

PR #1319, opened 2020-08-03
Self-signed TLS certificates for the aggregator example

rbehjati:
Ref #1303

- New certificate files are created for use on GCP, following the guidelines in `examples/certs/README.md`.
- The new TLS certificate is signed for `aggregator.oakexamples.dev` in addition to the IP address.
- Updated aggregator's GCP config files and scripts
- Updated aggregator's README file

Tested with the following command that starts the aggregator client that connects to the aggregator app running on Kubernetes. 
```
./bazel-client-bin/examples/aggregator/client/client \
 --address=aggregator.oakexamples.dev \
 --ca_cert=./examples/certs/gcp/ca.pem  \
 --bucket=test \
 --data=1:10,2:20,3:30
```


examples/certs/gcp/gcp.ext
```diff
@@ -4,4 +4,5 @@ keyUsage = digitalSignature, nonRepudiation, keyEncipherment, dataEncipherment
 subjectAltName = @alt_names
 
 [alt_names]
+DNS.1 = aggregator.oakexamples.dev
 IP.1 = 35.246.87.178
```
tiziano88, line 8: do we still need the IP address here?

examples/aggregator/scripts/deploy
```diff
@@ -29,6 +29,12 @@ then
   "${SCRIPTS_DIR}/undeploy" || true
 else
   "${SCRIPTS_DIR}/undeploy" -m || true
+
+  # Deleting the service does not immediately delete the node ports, so `deploy` fails 
+  # if it is run immediately after `undeploy`.
+  # Ref: https://github.com/kubernetes/kubernetes/issues/32987#issuecomment-344514151
+  echo "Waiting for 5 mins for k8s to remove all the old resources..."
+  sleep 5m
```
tiziano88, line 37: BTW What's the story for more realistic deployments? Obviously we cannot undeploy every single time, since it's time consuming and also comes with downtime every time.

examples/certs/gcp/gcp.key
```diff
@@ -1,27 +1,27 @@
 -----BEGIN RSA PRIVATE KEY-----
```
tiziano88, line 1: I assume this was regenerated using the instructions in https://github.com/project-oak/oak/blob/main/examples/certs/README.md

@conradgrobler is there any reason why that is not an actual executable shell script with comments? also from what I can tell the process is still pretty interactive, i.e. it asks for country code etc., hopefully that can be scripted too?

examples/certs/gcp/gcp.ext
rbehjati: I think we don't really need it. But it is not harmful to keep it. 
Or is it actually needed for certain scenarios @conradgrobler?

examples/certs/gcp/gcp.key
conradgrobler: There is not a good reason. At the time I thought it would not be something we had to rerun often so went with just documenting the steps and leaving it as an interactive process.

The other fields can be specified in the config file that is passed to `openssl req` for generating the CSR (e.g. https://gist.github.com/yidas/af42d2952d85c0951c1722fcd68716c6)

examples/aggregator/scripts/deploy
rbehjati: Normally, we don't need the `-m` flag, which redeploys the metrics-sidecar. Without `-m`, the `deploy` command is very fast. I had to redeploy the metrics-sidecar because I had made some changes to the config (actually mistakenly! I had to make those changes to the aggregator container!). 

So, we'll hardly ever run `deploy` with the `-m` command. Though, it would be nice to run `sleep 5m` only if the `undeploy` command actually deletes some resources. Is it possible to return a value from a bash script?

I think more realistic deployments would retry the `deploy` if it fails.

examples/certs/gcp/gcp.ext
conradgrobler: It is not really needed, but in future if the WebPKI crate does start supporting IP addresses in certs, it could be used for connections as well.

Issue #1315, opened 2020-08-03
move oak_abi protos to oak_abi folder


tiziano88:
currently, most of the proto files used by oak_abi are actually in the oak folder: 

https://github.com/project-oak/oak/blob/3a8f9c1e4f43269ad2afaedcb64c6414b5d1ad49/oak_abi/build.rs#L21-L31

In general, we should start moving towards a structure in which someone developing an oak application never has to use / depend on anything in the oak folder, just oak_abi and the Rust SDK.

ipetr0v:
Do we need to move files used by `oak` to `oak_abi` as well?
https://github.com/project-oak/oak/blob/e12f38d36e1f8aaff1736581adf7695fed06cd9d/sdk/rust/oak/build.rs#L19-L23

tiziano88:
I think so, either under oak_abi or in some SDK folder

ipetr0v:
I have moved them to `oak_abi`, but started receiving the following error:
```
error[E0277]: the trait bound `proto::oak_abi::label::Tag: std::cmp::Eq` is not satisfied
  --> src/label/mod.rs:61:86
   |
61 |         let self_confidentiality_tags: HashSet<_> = self.confidentiality_tags.iter().collect();
   |                                                                                      ^^^^^^^ the trait `std::cmp::Eq` is not implemented for `proto::oak_abi::label::Tag`
   |
   = note: required because of the requirements on the impl of `std::cmp::Eq` for `&proto::oak_abi::label::Tag`
   = note: required because of the requirements on the impl of `std::iter::FromIterator<&proto::oak_abi::label::Tag>` for `std::collections::HashSet<&proto::oak_abi::label::Tag>`
```
Is it because we are generating some custom code for `proto` files?

tiziano88:
I think this needs to change:

https://github.com/project-oak/oak/blob/e11c5c26f775ee45ec17507a306c189fdbabbefe/oak_utils/src/lib.rs#L241

ipetr0v:
If we want to move `handle.proto` into `oak_abi` (not just folder, but also to refer to it as `oak_abi::proto::handle`), we need to move [`oak_runtime::io`](https://github.com/project-oak/oak/blob/main/oak/server/rust/oak_runtime/src/io.rs) to `oak_abi` as well.

PR #1323, opened 2020-08-05
Move oak::proto to oak_abi

ipetr0v:
This change moves `oak::proto` to `oak_abi`.

Fixes https://github.com/project-oak/oak/issues/1315

# Checklist


ipetr0v:
Couple of notes regarding this commit:
- Files that were not moved from `oak/proto` to `oak_abi/proto`: `authentication.proto`, `handle.proto`, introspection_events.proto`
- Even though files used by `oak_abi` have been moved, objects are still referred as `oak_abi::proto::oak::...`

cc @tiziano88

PR #1326, opened 2020-08-05
Rename app serializer to app builder

tiziano88:
Make it behave more closely to how other tools work:

- take as input a manifest file
- emit output artifacts relative to the manifest file
- cache modules relative to the manifest file
- make output file name be suffixed with `.oak` extension
- more consistent terminology throughout (in particular avoid the
overloaded word `config`)

# Checklist

      construction.


sdk/rust/oak_app_build/src/main.rs
```diff
@@ -14,15 +14,17 @@
 // limitations under the License.
```
ipetr0v, line 14: Should it be called `oak_app_builder` or `oak_app_build`?

tiziano88: I like oak_app_build better, but I don't have a strong preference. I also thought that this is more a linker, so maybe it should be called oak_app_link(er)? Anyways, we can easily change the name in the future if we like something else.

PR #1327, opened 2020-08-06
Improve dev container support in VS code 

tiziano88:
Split Docker image build and run steps

Create script for docker build, and use it in cloud build too

Create user as part of the docker build step, so that it is a real user
with a proper $HOME and most tools work as expected

Add instructions on how to get things up and running on VS Code using
the remote container extension

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -240,3 +240,24 @@ ARG deny_version=0.7.0
 ARG deny_location=https://github.com/EmbarkStudios/cargo-deny/releases/download/${deny_version}/cargo-deny-${deny_version}-x86_64-unknown-linux-musl.tar.gz
 RUN curl --location ${deny_location} | tar --extract --gzip --directory=${install_dir} --strip-components=1
 RUN chmod +x ${install_dir}/cargo-deny
+
+# Unset $CARGO_HOME so that the new user will use the default value for it, which will point it to
+# its own home folder.
+ENV CARGO_HOME ""
+
+# Placeholder args that are expected to be passed in at image build time.
+# See https://code.visualstudio.com/docs/remote/containers-advanced#_creating-a-nonroot-user
+ARG USERNAME=user-name-goes-here
+ARG USER_UID=1000
+ARG USER_GID=${USER_UID}
+
+# Create the specified user and group.
+#
+# Ignore errors if the user or group already exist (it should only happen if the image is being
+# built as root, which happens on GCB).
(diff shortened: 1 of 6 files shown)


docs/development.md
```diff
@@ -40,6 +40,37 @@ The remainder of this document explores what's going on under the covers here,
...
+
+and then select "Remote-Containers: Reopen in Container".
+
+This should attach VS Code to an instance of the Docker container with all the
+dev tools installed, and configured with most linters and Rust tools.
+
+The Rust-Analyzer extension may prompt you to download the rust-analyzer server,
+in which case allow that by clicking on "Download now" in the notification.
+
+To test that things work correctly, open a Rust file, start typing `std::`
```
rbehjati, line 71: (nit): Maybe mention that you should give vs code some time to build the workspace (and rust-analyzer to build its index, I guess).

Dockerfile
```diff
@@ -240,3 +240,24 @@ ARG deny_version=0.7.0
...
 RUN curl --location ${deny_location} | tar --extract --gzip --directory=${install_dir} --strip-components=1
 RUN chmod +x ${install_dir}/cargo-deny
+
+# Unset $CARGO_HOME so that the new user will use the default value for it, which will point it to
+# its own home folder.
+ENV CARGO_HOME ""
+
+# Placeholder args that are expected to be passed in at image build time.
+# See https://code.visualstudio.com/docs/remote/containers-advanced#_creating-a-nonroot-user
+ARG USERNAME=user-name-goes-here
```
rbehjati, line 250: What happens if no value is assigned here? If that produces a meaningful error message when building the image, it might be better not to assign a value here.

docs/development.md
tiziano88: Done.

scripts/docker_build
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Usage: docker_run [OPTIONS] COMMAND
```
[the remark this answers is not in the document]
tiziano88: Done.

scripts/docker_build
```diff
@@ -0,0 +1,27 @@
...
+#
+# Options:
+#    --detach    Run container in background and print container ID
```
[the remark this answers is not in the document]
tiziano88: Done.

Dockerfile
rbehjati: In that case I agree that it is better to assign a value like this :)

PR #1328, opened 2020-08-06
Formalization of Oak for proving NI. Candidate security def.

aferr:
# Checklist

      construction.


aferr:
At this point there is a first-draft candidate of a top-level security condition when there are no downgrades. The top-level security condition is called "conjecture_possibilistic_ni" in PossibilisticNI.v

There are a handful of ways in which both this model and theorem are (necessarily) a bit different. So extra eyes on whether or not this is a good specification of security would be really helpful.

I am still pretty new to coq and I think there are likely a lot of ways to write things that have different implications as to how difficult it is to prove something. Any feedback on how to make things more provable would also be greatly appreciated!

jieungkim:
1. GenericMap.v: There should be several standard and/or extended libraries that provide generic total/partial maps. You do not need to change this library immediately, but you can think about using those pre-built libraries (e.g. coq-mathcomp-finmap / coq-int-map - opam packages). If you want to use those built-in map libraries, you may need to change the key of those map definitions (currently you are using any decidable values, built-in map libraries may use Z or Int types). But making an injection function should not be hard. 

2. Lattice.v: You defined lattice as a type class. But, think about defining it such as this file (by using modules: https://github.com/jieungkim/SF-prep/blob/jieung/Challenges/ZSet.v). We have used this one in CertiKOS since we want to reduce hypothesis when extracting codes. But, if it is not related to the code extraction, using type classes with out instantiations should be fine. 

3. GenericModel.v What are potential problems of handle_fresh and nid_fresh? 
-> And I agree that "option types" might be better when checking the value is in the partial map or not, because the current proposition definition cannot distinguish the empty map and the map that is not empty but does not contain the value associated with the key.

4. for "step_system": will "(state_upd_call id c' s')" change the current id? otherwise, it seems "step_system" still using the same id during its evaluation. 

5. for "step_system": when modeling the full system semantics, you may need the environmental context that defines a valid node set in the system. And, it seems you need to define init_state as well. 

6. for decidability, you can take a look at "https://github.com/jieungkim/SF-prep/blob/jieung/Challenges/Decision.v". This does not make a dramatic simplicity, but it is quite useful for readability in my sense. 

7. For defining step relations and proving invariants that are satisfied by those step relations, seeing the CompCert's Smallstep library (http://compcert.inria.fr/doc-1.6/html/Smallstep.html) or other related things.

aferr:
Thanks Jieung! That's super helpful!

1, 2, 6, 7: I will take a look at those and probably make any changes in later CLs. Thanks for the pointers.

3. I was referring to exactly the problem you mentioned: Inability to distinguish between a real channel/node that looks like empty and one that just does not exist. I was just putting off changing to option types. Is it better if I change to option types after looking at other map libraries?

4. In the current definition there is no current id in the global state. In the ValidStep constructor of step_system, it just uses any (id: node_id) s.t. id maps to a node called n. With this definition, I think for example nid = 1 can be used to step from s to s', and then nid = 2 can be used to step from s' to s''. So I think this definition does allow the node that executes to change from one transition to another.

5. valid nodes: Once I switch to option types, I was thinking of modeling the valid nodes in the system as any node_id that maps to Some (n) (and this change would go in H0 of valid_step). I think if this is the case the node_state map is enough. But it could also be defined by not having node_id at all and instead just using a set of nodes rather than a partial map from node_ids to nodes. Is one better than the other for writing proofs, or are they about the same?

init_state: I think one invariant I need on states is that there are never read handles that are held by more than one node, and I might also need this relation to hold on initial states, but I haven't hit a point where I need this yet, so I don't have it.

Some other questions while I have your attention:
- Does "conjecture_possibilistic_ni" make sense to you? 
- I am building traces of states/events in EvAugSemantics.v. Is this a good way to build traces?

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,222 @@
...
+    ms: list message
+}.
+
+(* ABI Calls *)
+Inductive call: Type :=
+    | WriteChannel (h: handle)(m: message): call
+    | ReadChannel (h: handle): call
+    | CreateChannel (lbl: level): call
+    | CreateNode (lbl: level)(h: handle): call
+    | Internal: call.
```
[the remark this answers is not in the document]
aferr: It's meant to be any action done by the node other than make an ABI call. It looks like a NOP from the perspective of the rest of the system since the only way to influence things outside the node is the ABI. I added a comment but kept the old name. I don't have a strong opinion on the name if you think NoOp makes more sense.

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,222 @@
...
+    chans: chan_state
+}.
+
+(*============================================================================
+* Empty
+============================================================================*)
+Definition empty_chan := {| clbl := top; ms := []; |}.
+Definition empty_node := {|
+        nlbl := top;
+        rhans := Empty_set handle;
```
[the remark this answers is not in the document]
aferr: I fixed the names in a later commit.

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,222 @@
...
+        whans := Empty_set handle;
+        ncall := Internal;
+    |}.
+Definition empty_state := {|
```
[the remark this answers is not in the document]
aferr: It's a value of type state that I am using to represent an empty one.  It actually wasn't used, so I just deleted it.

jieungkim:
- I was referring to exactly the problem you mentioned: Inability to distinguish between a real channel/node that looks like empty and one that just does not exist. I was just putting off changing to option types. Is it better if I change to option types after looking at other map libraries?
=> I guess at least Map library in CompCert and IntMap (https://www.cs.yale.edu/flint/cs428/coq/library/Coq.IntMap.Map.html) are partial maps. They are good for differentiating empty values and non-existing values, but we always have to unwrap the values when we can get the value from the map. It is sometime bothersome, but they are at least guarantee the total correctness. 

To use simple representation with several map operations, we may be able to use monadic style representations, but I think that will not be the immediate issue.

- Does "conjecture_possibilistic_ni" make sense to you?
=> In my understanding, yes. I guess... you want to say that if the initial traces are same, and if there is a non-deterministic evaluation that ends with a certain trace, there always have a possible transitions that ends with the equal trace at the end (it seems you want to show the observable equivalence of any evaluation sequences - thus prove security of the system). 

One thing that you can improve this proof is... you can parameterize "step_system_ev_multi". 
=> make a polymorphic multi step rules 
Variable State : Type.
Variable single_step : State -> State -> Prop. 
Inductive multi_step: Type -> Type -> Prop :=
| multi_step_refl : forall st st', single_step st st' -> multi_step st st'
| multi_step_trans: forall st1 st2 st3, 
  single_step st1 st2 ->
  multi_step st2 st3 ->
  multi_step st1 st3.
=> Define several lemmas and proofs for this parameterized multi_step rules 
=> then use the polymorphic proofs for several places. 

CompCert's Smallstep library is very good at this kind of thing. 

- I am building traces of states/events in EvAugSemantics.v. Is this a good way to build traces?
=> I think that's fine and correct. It seems the transition rule is from the pair of state and trace to the pair of state and trace, thus it is quite similar to the trace recording that I did in CertiKOS (we put our trace in the state itself though).  I am just curious why you also record all result state in the trace. Is there a specific reason that you do that? 

=> I think your multi_step rule only allows at most two steps. Please look at the above polymorphic multi_step example, and CompCert's Smallstep library.

=> minor thing: I think you have a typo at the last line of the "step_node_ev" relation. 
[step_node_ev id Internal s t s ((s', ev) :: t).] => [step_node_ev id Internal s t s' ((s', ev) :: t).] 
Am I correct?

experimental/ni_coq/vfiles/Events.v
```diff
@@ -0,0 +1,115 @@
...
+        exists t2, <c, s2> => t2 /\ t1 =L t2.
+
+    In other words there is some way to reach an execution trace
+    that looks the same beginning from the other state.
+    
+    Trapeze uses a possibilistic definition of security:
+    https://pdfs.semanticscholar.org/809b/f2702a765b9e7dba4624a1dbc53af11579db.pdf
+    See also:
+    https://www.cs.cornell.edu/andru/papers/csfw03.pdf
+*)
```
alastairreid, line 54: This comment and the next big comment about the alternative definition is really useful.
In fact, as I was reading this code, I had already added a request for this kind of explanation of why you chose one style of definition instead of another in PossibilisticNI.v.

I am not sure that this is the best place to put the explanations.
I think you have put them here because I think that the Inductive types are used to represent judgements in proofs of each of the two NI definitions.
I think it would be better to put them somewhere else so that a reader who is more interested in the model and the theorems (but is skipping the proof) would see this explanation.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -0,0 +1,38 @@
+Require Import OakIFC.Parameters.
+Require Import OakIFC.RuntimeModel.
+Require Import OakIFC.EvAugSemantics.
+Require Import OakIFC.Events.
+Require Import OakIFC.LowEquivalences.
+Require Import Coq.Lists.List.
+Import ListNotations.
+
```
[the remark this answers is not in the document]
alastairreid: Sorry, I meant to amend this comment to refer to your description in Events.v that explains this (and which asks whether Events.v is the best place for the description).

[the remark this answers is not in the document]
aferr: Changing the "exists" to "forall" would make the definition more like observational determinism whereas now the definition is closer to possibilistic noninterference. (Though as mentioned in the Events.v file, the definition of trace low-equivalence would need to change)

One reason I went with the possibilistic definition (at least for now) is that the only thread-granularity IFC runtime that I know of (Trapeze) (https://pdfs.semanticscholar.org/809b/f2702a765b9e7dba4624a1dbc53af11579db.pdf) uses this kind of possibilistic security definition, so starting with this definition would be an example of beginning the project by trying to replicate the state of the art.

On advantage of observational determinism is that it is preserved by refinement, so for something like the earlier systems verification work, it would have been more important/useful to try to target a definition that is based on O.D. But for this project, we are not doing a refinement proof connecting the model and implementation, so this difference does not matter. To prove O.D., I think we'd also need to prove some lemmas about data race freedom. I do not remember the details about what needs to be proven and have not dug into the details about the synchronization primitives of the Oak Runtime to know whether or not the data race freedom propositions (or more accurately, a sensible translation of these propositions from a language-based domain to ones about this model) would be true. It might be something I'll at least think about looking into more deeply again. Though at the moment it does not seem to me to be the best place to spend effort.

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -0,0 +1,92 @@
...
+Definition msg_is_head (ch: channel)(m: message): Prop :=
+    match ch.(ms) with
+        | [] => False
+        | m' :: ms' => m' = m
+    end. 
+
+Inductive step_node_ev: node_id -> call -> state -> trace -> state -> trace -> Prop :=
+    | SWriteChan s id han msg s' t ev
+        (H0: step_node id (WriteChannel han msg) s s')
+        (H1: ev = (EvL (OutEv msg) (s.(nodes) id).(nlbl))):
```
blaxill, line 54: Optional: It might be worth adding some notation to make these props more readable, up to you though. e.g.

```Coq
Notation "'event_is' ev msg" = (ev = (EvL (msg (s.(nodes) id).(nlbl))).
```

I think also since the `ev` is only used in the index of `step_node_ev`, absorbing `H1` into the RHS is equivalent? (although would change proof structure)

```Coq
Inductive step_node_ev: node_id -> call -> state -> trace -> state -> trace -> Prop :=
    | SWriteChan s id han msg s' t ev
        (H0: step_node id (WriteChannel han msg) s s'):
         step_node_ev id (WriteChannel han msg) s t s' ((s', (EvL (OutEv msg) (s.(nodes) id).(nlbl))) :: t)
...
```

Again notation for `(EvL (OutEv msg) (s.(nodes) id).(nlbl))` could help out with readability.

This is all just stylistic choice though so completely up to you.

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -0,0 +1,92 @@
...
+        step_node_ev id (CreateChannel lbl) s t s' ((s', ev) :: t)
+    | SCreateNode s id lbl h s' t ev
+            (* model observation that a node is created ?? *)
+        (H0: step_node id (CreateNode lbl h) s s')
+        (H1: ev = (EvL NilEv (s.(nodes) id).(nlbl))):
+        step_node_ev id (CreateNode lbl h) s t s' ((s', ev) :: t)
+    | SInternal s id s' t ev
+        (H0: step_node id Internal s s')
+        (H1: ev = (EvL NilEv (s.(nodes) id).(nlbl))):
+        step_node_ev id Internal s t s ((s', ev) :: t).
```
jieungkim, line 75: It seems the last line should be 
        step_node_ev id Internal s t s' ((s', ev) :: t).
Is it correct?

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -0,0 +1,92 @@
...
+        step_system_ev s t (state_upd_call id c' s') t'.
+
+
+Inductive step_system_ev_multi: state -> trace -> state -> trace -> Prop :=
+    | multi_system_ev_refl s t s' t'
+        (H: step_system_ev s t s' t'):
+        step_system_ev_multi s t s' t'
+    | multi_system_ev_tran s1 t1 s2 t2 s3 t3 
+        (H1: step_system_ev s1 t1 s2 t2)
+        (H2: step_system_ev s2 t2 s3 t3):
```
jieungkim, line 91: I guess your intention in here is  (H2: step_system_ev_multi s2 t2 s3 t3)

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -0,0 +1,92 @@
+Require Import OakIFC.Lattice.
+Require Import OakIFC.Parameters.
+Require Import OakIFC.GenericMap.
```
jieungkim, line 3: You can use the following notations for better readability and maintenance. 

From OakIFC Require Import
  Lattice 
  Parameters
  GenericMap.

experimental/ni_coq/vfiles/Events.v
```diff
@@ -0,0 +1,115 @@
...
+    trace-low equivalence we should target. The O.D. style is stronger, but it
+    might not be possible for us to prove it.
+*)
+Inductive stut_trace_low_eq: level -> trace -> trace -> Prop :=
+    | SNilEQ ell: stut_trace_low_eq ell [] []
+    | SAddBoth ell xs xe ys ye t1 t2 
+        (H1: stut_trace_low_eq ell t1 t2)
+        (H2: ev_low_eq ell xe ye)
+        (H3: state_low_eq ell xs ys):
+        stut_trace_low_eq ell ((xs, xe)::t1) ((ys, ye)::t2)
```
blaxill, line 105: Another completely optional style suggestion, since you are not giving the propositions detailed names or using them on the RHS, you _could_ leave them unnamed:

```Coq
Inductive stut_trace_low_eq: level -> trace -> trace -> Prop :=
...
    | SAddBoth ell xs xe ys ye t1 t2: 
        stut_trace_low_eq ell t1 t2
        -> ev_low_eq ell xe ye
        -> state_low_eq ell xs ys
        -> stut_trace_low_eq ell ((xs, xe)::t1) ((ys, ye)::t2)
```

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -0,0 +1,66 @@
+Require Import OakIFC.RuntimeModel.
+Require Import OakIFC.Parameters.
+Require Import OakIFC.LowEquivalences.
+Require Import Coq.Lists.List.
+Require Import OakIFC.GenericMap.
+Import ListNotations.
+
+Theorem chan_upd_preserves_nodes: forall h ch s,
+    (state_upd_chan h ch s).(nodes) = s.(nodes).
+    intros; simpl; reflexivity.
```
blaxill, line 10: very minor: for consistency's sake I'd still start with `Proof.`

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -0,0 +1,38 @@
...
+        (step_system_ev_multi s21 [] s2n t2n) /\
+        (trace_low_eq ell t1n t2n)).
+
+Theorem possibilistic_ni_1step: forall ell s11 s21 s12 t12,
+    (state_low_eq ell s11 s21) /\
+    (step_system_ev s11 [] s12 t12) ->
+    (exists s22 t22,
+        (step_system_ev s21 [] s22 t22) /\
+        (trace_low_eq ell t12 t22)).
+Admitted.
```
blaxill, line 22: Perhaps a comment as to why this is currently admitted?

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,219 @@
...
+        (H2: handle_fresh s h):
+            let s0 := (state_upd_chan h {| ms := []; clbl := lbl; |} s) in
+            let s1 := (state_node_add_rhan h cid s0) in
+            let s' := (state_node_add_whan h cid s1) in
+            step_node cid (CreateChannel lbl) s s'
+    | SCreateNode s cid nid lbl h
+        (H0: (s.(nodes) cid).(nlbl) << lbl)
+        (H1: (nid_fresh s nid)):
+            (* create new node with read handle *)
+        let s0 := (state_upd_node nid {|
```
blaxill, line 192: Optional: perhaps worth putting these bindings into their own definitions (or notation)?

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,219 @@
...
+            read_handles := Ensembles.Subtract old_n.(read_handles) h;
+            write_handles := old_n.(write_handles);
+            ncall := old_n.(ncall);
+        |} s.
+
+(* 
+NOTE it might actually be better to use option types
+for the range of both the node state and channel states 
+*)
+(* There may be potential problems with these definitions *)
```
blaxill, line 151: I dont see any immediate problems, is there some issue in particular you were thinking of?

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -0,0 +1,224 @@
...
+Require Import List.
+Import ListNotations.
+Require Import Coq.Sets.Ensembles.
+
+(* Ensembles don't have implicit type params and these lines fix that *)
+Arguments Ensembles.In {U}.
+Arguments Ensembles.Add {U}.
+Arguments Ensembles.Subtract {U}.
+Arguments Ensembles.Singleton {U}.
+
```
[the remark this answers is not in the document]
aferr: Addressed in latest commit. Also added a note about channel reads specifically.

experimental/ni_coq/vfiles/vestigial/StateEqb.v
```diff
@@ -0,0 +1,52 @@
+Require Import OakIFC.RuntimeModel.
+Require Import OakIFC.Lattice.
+Require Import OakIFC.Parameters.
+Require Import OakIFC.GenericMap.
+Require Import Coq.Lists.List.
+Require Import Coq.Bool.Bool.
+
```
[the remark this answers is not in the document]
aferr: Addressed in latest commit, which also mentions that this file is no longer in use, like the other files in this subdirectory.

experimental/ni_coq/README.md
```diff
@@ -0,0 +1,5 @@
+Experimental formalization of Oak + NI result.
+
+Instructions to re-generate makefile:
+(see also: https://coq.inria.fr/refman/practical-tools/utilities.html)
+coq_makefile -f _CoqProject -o Makefile
```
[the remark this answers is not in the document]
aferr: I added a paragraph about noninterference and pointed to the two main files, but I'll iterate on it again to talk more about the progression as you requested.

I might like to hold off on talking about the file structure too much other than maybe the two main files for the runtime and the security definition at the moment because I bet I'll restructure it a lot and forget to update the readme.

experimental/ni_coq/vfiles/Events.v
aferr: Added some (partly overlapping) discussion in PossibilisticNI.v in a later commit.

experimental/ni_coq/vfiles/Events.v
```diff
@@ -0,0 +1,115 @@
...
+        (H1: ~((lvl el2) << ell))
+        (H2: ~((lvl el1) << ell)):
+        ev_low_eq ell el1 el2.
+
+Definition trace := list (state * event_l).
+
+(* We might need two different definitions of trace low-equivalence
+* depending on the top-level security condition *)
+
+(* This is a strateforward definition of trace low-equivalence
```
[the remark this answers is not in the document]
aferr: At least it was strateforward to fix.

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -0,0 +1,92 @@
...
+informtion because the choice of call is always essentially non-deterministic,
+so the call is a piece of state that probably does not matter at the moment)
+
+ When downgrades events are added, a trace might need to be
+ list (state * (Ensemble event))                  (Ensembles are sets)
+ rather than
+ list (state * event)
+ since individual calls might produce more than one event. For example,
+ when a read call does a declassification it would produce an input event
+ and a downgrade event.
```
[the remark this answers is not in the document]
aferr: That's not a naive question. Actually the fact that this definition involves both states and events at the same time is slightly unusual. Usually you have one or the other, but I think we need both here.

I was also thinking about the option you described and I don't have a definitive answer for which is the better way to specify it. The reason I was leaning towards sets of events at least at the time of the comment is that if the type is
list (state * Ensemble event) then I think I can more cleanly explain what a single trace element is: it's the state at the time of a transition along with the set of all the events that happened at that point in the trace.

If instead, the future downgrading send call pushes two elements onto the trace, then the same state (at the end of the call) gets repeated in the trace twice. So explaining what a trace element is would seem slightly less clear.

I don't really know which one is easier to work with or overall better. I *think* the two ways of writing it would turn out to be equivalent in terms of which "runtime models" they would permit. I also *think* this might still be true for either of the two trace low-equivalence definitions on traces.

experimental/ni_coq/vfiles/RuntimeModel.v
aferr: If some value for node/channel is used rather than an option type, then handle_fresh and nid_fresh don't distinguish between the case where the runtime really hasn't made such a handle or node yet or the case where there is a channel or node, but it happens to be exactly the empty channel or node. These weird edge cases, but maybe it matters.

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -81,12 +82,11 @@ Inductive step_system_ev: state -> trace -> state -> trace -> Prop :=
...
 
-
 Inductive step_system_ev_multi: state -> trace -> state -> trace -> Prop :=
     | multi_system_ev_refl s t s' t'
         (H: step_system_ev s t s' t'):
         step_system_ev_multi s t s' t'
     | multi_system_ev_tran s1 t1 s2 t2 s3 t3 
-        (H1: step_system_ev s1 t1 s2 t2)
-        (H2: step_system_ev s2 t2 s3 t3):
+        (H1: step_system_ev_multi s1 t1 s2 t2)
```
jieungkim, line 90: I guess this one is also correct, but for proof, revert this one to the previous one 
        (H1: step_system_ev s1 t1 s2 t2)
is also good. 

So, it could be    
     (H1: step_system_ev s1 t1 s2 t2)
     (H2: step_system_ev_multi s2 t2 s3 t3)

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: This was a typo. Fixed now.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: Thanks. This is done now.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: Yes, it was a typo. Fixed now.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: Ah okay. It looks like in PL Foundations, they also use (H1: single step ... ) (H2: multi step ... ). I guess this is better for the proof because then there is only one way to construct H1. I did it this way in dbac1ff. Thanks.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: That is much better! I was putting off using notation until I was more settled on how to write various definitions (though maybe it's actually easier if you have notation earlier in the process).
I have added notation for events now in d421739

experimental/ni_coq/vfiles/Events.v
aferr: I did like the style I was using better, but you've now sold me on the style you suggested. :) The change is here: b541584

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: Fixed: 9b8e627

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Fixed: 9b8e627

Also, I don't plan to make a habit of checking in WIP proofs. I iterated on the definitions a few times and wanted to just start the proofs a bit to see if they are reasonable before trying to collect feedback.

experimental/ni_coq/vfiles/RuntimeModel.v
aferr: That sounds like a good idea. I do want more notation in general, but for state updates specifically, I want to hold off because I might want to first try out the libraries Jieung suggested.

aferr:
I think I have resolved everything. Please feel free to unresolve anything if you feel I have not. Approve if you think it is ready.

Issue #1324, opened 2020-08-05
Oak ABI / SDK layering


tiziano88:
We have a few open issues about figuring out where various types should live.

I believe @daviddrysdale and @anghelcovici originally suggested a layered approach, which I think is a great idea.

How about the following crate structure:

- **oak_abi**
  + contains low-level Wasm functions and primitive types, only handles raw bytes + handles
  + no dependencies on any other crate
- **oak_io**
  + contains helper types for higher level, strongly typed, passing of bytes + handles (`Sender`, `Receiver`, `Encodable`, `Decodable`)
  + depends on oak_abi
- **oak_services**
  + contains protobuf definitions of the interface for the various node types (e.g. gRPC invocation, logging protos, etc.)
  + depends on oak_io and oak_abi
- **oak_sdk**
  + additional types and helpers to make the interaction with services more ergonomic from Rust
  + depends on oak_services, oak_io and oak_abi
- **oak_runtime**
  + internal implementation of Runtime
  + developers of Oak applications must not depend on this -- not even transitively
  + depends on oak_services, oak_io and oak_abi

cc @ipetr0v @conradgrobler

Issue #1224, opened 2020-07-02
de-duplicate `Message`, `Encodable`, `Decodable` types between Runtime and SDK


tiziano88:
Runtime `NodeMessage`:

https://github.com/project-oak/oak/blob/2cf69a76494d21189c171263679a9a4199866e29/oak/server/rust/oak_runtime/src/message.rs#L30-L39

SDK `Message`:

https://github.com/project-oak/oak/blob/2cf69a76494d21189c171263679a9a4199866e29/sdk/rust/oak/src/io/mod.rs#L46-L51

Runtime `Encodable`:

https://github.com/project-oak/oak/blob/2cf69a76494d21189c171263679a9a4199866e29/oak/server/rust/oak_runtime/src/io.rs#L42-L57

SDK `Encodable`:

https://github.com/project-oak/oak/blob/2cf69a76494d21189c171263679a9a4199866e29/sdk/rust/oak/src/io/encodable.rs#L19-L31

etc.

Note that this is currently blocked by #473, since otherwise this would introduce a dependency cycle.

Probably most of this will be solved at least partially by #1108.

conradgrobler:
I dont think the runtime should depend on the SDK (or vice versa), so these should either be moved to the ABI (although it is not strictly part of the ABI) or a new shared crate that can be referenced by the SDK and the runtime.

tiziano88:
I agree in principle, but I am not sure adding even more crates is actually a viable strategy. In practice even the split between ABI and SDK seems like a premature optimization (I don't expect anyone to actually depend on ABI but not SDK for the foreseeable future, until someone comes up with a completely different SDK on top of the same ABI).

So yes, for now I think moving them to the ABI makes sense.

tiziano88:
@conradgrobler it sounds like we need to have a clear layering between crates, and decide what goes where (as a side note I believe @daviddrysdale has put some thought into this in the past already). We may consider adding a new intermediate crate between oak_abi and the SDK crate if it helps with anything, but could you detail how you suggest to go about it @conradgrobler ?

conradgrobler:
#1324 will help with this. As discussed, we also need to remove the inconsistencies that are currenlty blocking the de-duplication:

- We should have a single `Handle` type: the suggestion is to standardise on the ABI handle
- We should use `OakError` consistently for the `Encodable` and `Decodable` traits
[...]

PR #1330, opened 2020-08-06
Move implementations from SDK Sender and Receiver

conradgrobler:
Move the context-dependent implementations from the SDK's `Sender` and `Receiver` structs into extension traits.

This is a first step towards implementing #1224, #1249, #1324, and generally making it possible to serialise handles in protos in the runtime pseudo nodes as well. This requires harmonising the `Sender` and `Receiver` structs between the SDK and runtime. Some of the functions are incompatible between the runtime and SDK, so have to be moved out of the implementation of the struct itself. The next step is to do something similar for the runtime versions of these structs.

- `SenderExt`
  - `send`
  - `close`
- `ReceiverExt`
  - `receive`
  - `wait`
  - `try_receive`
  - `close`

# Checklist

      construction.

### DIFF
--- a/examples/abitest/module_0/rust/src/lib.rs
+++ b/examples/abitest/module_0/rust/src/lib.rs
@@ -1625,8 +1625,7 @@ impl FrontendNode {
                 new_write.handle, internal_req,
             );
             let new_channel = oak::io::Sender::new(*new_write);
-            new_channel
-                .send(&internal_req)
+            oak::io::send(&new_channel, &internal_req)
                 .expect("could not send request over channel");
             oak::channel_close(new_write.handle).expect("could not close channel");
 
@@ -1751,7 +1750,7 @@ impl FrontendNode {
         // Close the only read-handle for the invocation handle, which should
         // trigger the gRPC server pseudo-Node to terminate (but we can't
         // check that here).
-        expect_eq!(Ok(()), invocation_receiver.close());
+        expect_eq!(Ok(()), oak::io::close_receiver(&invocation_receiver));
         Ok(())
     }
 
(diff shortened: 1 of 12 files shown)


examples/abitest/module_0/rust/src/lib.rs
```diff
@@ -1869,7 +1868,7 @@ impl FrontendNode {
             .map_err(from_proto)?;
         let mut count = 0;
         loop {
-            match receiver.receive() {
+            match oak::io::receive(&receiver) {
```
tiziano88, line 1871: I think for the SDK crate we can actually recover most of the existing semantics by declaring an extension that exposes the `.receive` and other methods directly on the sender / receiver types (the usual pattern is to have a `ReceiverExt` and `SenderExt` trait, but the naming does not really matter).

conradgrobler: I changed it to use extension traits as you suggested.

examples/injection/module/rust/src/lib.rs
```diff
@@ -66,7 +66,7 @@ mod proto {
 
 use oak::{
     grpc,
-    io::{Receiver, Sender},
+    io::{Receiver, ReceiverExt, Sender, SenderExt},
```
tiziano88, line 69: I imagine eventually `oak::io::Receiver` and `oak::io::Sender` would be replaced by simpler objects (possibly with same name) defined in `oak_abi` , and also generated by our custom prost logic, is that correct?

conradgrobler: Yes, although it would probably move to the new `oak_io` crate as part of #1324.

PR #1338, opened 2020-08-10
Add SDK support for HTTP server pseudo-node

rbehjati:
* Adds an init function to the SDK for creating HTTP server pseudo-nodes
* Replaces grpc Invocation in the SDK with a generic Invocation to use for
  both HTTP and Grpc server pseudo-nodes
* Adds a simple example that uses an HTTP server pseudo-node as front-door
* Updates docs accordingly

Ref #1186 

# Checklist

      construction.


examples/http_server/client/client
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+curl localhost:8080
```
conradgrobler, line 3: I don't think any of our examples should use non-secure connections. The HTTP server should use TLS, and the local CA cert can be passed to curl using the `--cacert ` command line argument.

sdk/rust/oak/src/invocations.rs
```diff
@@ -0,0 +1,154 @@
...
+        if let Err(error) = self.request_receiver.close() {
+            warn!("Error when trying to close request_receiver: {}", error);
+        }
+        if let Err(error) = self.response_sender.close() {
+            warn!("Error when trying to close response_sender: {}", error);
+        }
+    }
+}
+
+pub struct InitialServerInvocation<
```
conradgrobler, line 99: Perhaps add a doc comment to clarify the reason for creating this?

examples/http_server/module/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+impl Node<HttpInvocation> for StaticHttpServer {
+    fn handle_command(&mut self, invocation: HttpInvocation) -> Result<(), OakError> {
+        let request = invocation.receive()?;
+
+        info!("Handling a request to {}.", request.path);
+
+        let http_response = HttpResponse {
+            body: "Hello from HTTP server!\n".to_string().into_bytes(),
+            status: 200,
+            headers: std::collections::HashMap::new(),
```
conradgrobler, line 44: Is this needed?

examples/http_server/client/client
rbehjati: I have not added TLS to the HTTP server node yet. I will fix this, when TLS support is added.

examples/http_server/module/src/lib.rs
rbehjati: I could use `..Default::default()`, but I think it is better to be explicit, since this is an example showcasing the use of HTTP server pseudo-node.

sdk/rust/oak/src/invocations.rs
rbehjati: Done. But maybe it is not the best description. WDYT?

sdk/rust/oak/src/invocations.rs
```diff
@@ -0,0 +1,156 @@
...
+        if let Err(error) = self.request_receiver.close() {
+            warn!("Error when trying to close request_receiver: {}", error);
+        }
+        if let Err(error) = self.response_sender.close() {
+            warn!("Error when trying to close response_sender: {}", error);
+        }
+    }
+}
+
+/// The write half of a channel sending external requests as `Invocations` to an Oak Node that
```
conradgrobler, line 99: I think the write half of the channel is quite deeply buried in the object structure Perhaps something like `Wrapper for a sender that can be used to send invocations to an Oak node that will handle the requests.`

rbehjati: Thanks. Done.

examples/http_server/module/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+impl Node<HttpInvocation> for StaticHttpServer {
+    fn handle_command(&mut self, invocation: HttpInvocation) -> Result<(), OakError> {
+        let request = invocation.receive()?;
+
+        info!("Handling a request to {}.", request.path);
+
+        let http_response = HttpResponse {
+            body: "Hello from HTTP server!\n".to_string().into_bytes(),
+            status: 200,
+            // headers: std::collections::HashMap::new(),
```
juliettepretot, line 44: Should we remove this comment? :)

examples/http_server/module/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+    let http_channel =
+        oak::http::init("[::]:8080").expect("Could not create HTTP server pseudo-Node!");
+
+    oak::run_event_loop(node, http_channel);
+});
+
+struct StaticHttpServer {}
+
+impl Node<HttpInvocation> for StaticHttpServer {
+    fn handle_command(&mut self, invocation: HttpInvocation) -> Result<(), OakError> {
```
juliettepretot, line 36: Why does this need `&mut self`?

rbehjati: I think generally `handle_command` (declared in `Node`) may need `self` to be mutable. `StaticHttpServer` is implementing `Node` and cannot change the signature of `handle_command`, even though it is not really using `self`.

examples/http_server/module/src/lib.rs
rbehjati: Yes. Fixed :)

oak/server/rust/oak_runtime/src/node/http/mod.rs
```diff
@@ -428,8 +428,8 @@ fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
             OakStatus::ErrInvalidArgs
         }),
         None => {
-            warn!("No HTTP label found:");
-            Err(OakStatus::ErrInvalidArgs)
+            warn!("No HTTP label found: using public_untrusted.");
+            Ok(Label::public_untrusted())
```
tiziano88, line 432: This seems a very dangerous change: it is going to assume that any data sent to this server is public unless an explicit header is provided, but surely in that case Oak is not doing anything useful at all, so it's almost definitely a mistake from the client. We should fail closed in this case.

sdk/rust/oak/src/http/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+//! Functionality to help Oak Nodes create HTTP server pseudo-Nodes.
+use crate::{
+    invocations::{Invocation, ServerInvocationChannel},
+    io::{Receiver, ReceiverExt, SenderExt},
+    OakStatus,
+};
+
+use oak_abi::proto::oak::encap::{HttpRequest, HttpResponse};
+
+pub type HttpInvocation = Invocation<HttpRequest, HttpResponse>;
```
tiziano88, line 26: Maybe just call this `Invocation`, so that it can be referred to as `oak::http::Invocation`? And similarly for the gRPC one.

sdk/rust/oak/src/invocations.rs
```diff
@@ -0,0 +1,156 @@
...
+//
+
+use crate::io::{ReceiverExt, SenderExt};
+use log::warn;
+
+/// An invocation, consisting of exactly two channels: one to read incoming requests from the
+/// client, and one to write outgoing responses to the client.
+pub struct Invocation<
+    R: prost::Message + crate::handle::HandleVisit + crate::io::Decodable + std::default::Default,
+    S: prost::Message + crate::handle::HandleVisit + crate::io::Encodable + std::clone::Clone,
```
tiziano88, line 24: Do we really need all these type bounds? I would have expected just `R: Decodable, S: Encodable`.

sdk/rust/oak/src/invocations.rs
```diff
@@ -0,0 +1,156 @@
...
+            warn!("Error when trying to close request_receiver: {}", error);
+        }
+        if let Err(error) = self.response_sender.close() {
+            warn!("Error when trying to close response_sender: {}", error);
+        }
+    }
+}
+
+/// Wrapper for a sender that can be used to send invocations to an Oak node that will handle the
+/// requests.
```
tiziano88, line 100: I am not convinced this type is actually necessary / useful, but I guess fine to keep if it simplifies any of the logic.

sdk/rust/oak/src/http/mod.rs
tiziano88: edit: in fact it seems that's already the case for the gRPC one, so this should probably just follow the same pattern.

sdk/rust/oak/src/invocations.rs
rbehjati: This replaces `GrpcServerInit`, previously in the `grpc::Server` module. `oak::io::channel_create` requires something that is `Encodable + Decodable`. As far as I can see the only purpose of this type is to  implements those traits. Is there a better way to do it?

oak/server/rust/oak_runtime/src/node/http/mod.rs
rbehjati: I'll fix it with #1345.

sdk/rust/oak/src/invocations.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    pub fn close_channels(&self) {
+        if let Err(error) = self.request_receiver.close() {
+            warn!("Error when trying to close request_receiver: {}", error);
+        }
+        if let Err(error) = self.response_sender.close() {
+            warn!("Error when trying to close response_sender: {}", error);
+        }
+    }
+}
```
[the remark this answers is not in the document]
rbehjati: I think this is more readable :)

sdk/rust/oak/src/invocations.rs
rbehjati: I believe it will work if we implement `Encodable` and `Decodable` for sender.

tiziano88:
> Thanks for the comments @tiziano88, including the styling ones. Is there a version of clippy or rustfmt that warns about these things or fixes them automatically?

not that I know of, but if you figure it out please add it to https://github.com/project-oak/oak/blob/main/.rustfmt.toml

PR #1342, opened 2020-08-11
Drop node_id from channel introspection events

juliettepretot:
As suggested by @daviddrysdale: https://project-oak.slack.com/archives/CHE9E13C3/p1595049790239800?thread_ts=1594995744.232500&cid=CHE9E13C3

Imho this makes sense conceptually, since channels aren't tied to Nodes, but connected to them via handles. 

See the related issue #913

# Checklist

      construction.

### DIFF
--- a/oak/proto/introspection_events.proto
+++ b/oak/proto/introspection_events.proto
@@ -47,15 +47,11 @@ message NodeDestroyed {
 }
 
 message ChannelCreated {
-  uint64 node_id = 1;
-
-  uint64 channel_id = 2;
+  uint64 channel_id = 1;
 }
 
 message ChannelDestroyed {
-  uint64 node_id = 1;
-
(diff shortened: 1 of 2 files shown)


oak/proto/introspection_events.proto
```diff
@@ -47,15 +47,11 @@ message NodeDestroyed {
...
-  uint64 channel_id = 2;
+  uint64 channel_id = 1;
 }
 
 message ChannelDestroyed {
-  uint64 node_id = 1;
-
-  uint64 channel_id = 2;
+  uint64 channel_id = 1;
 }
```
juliettepretot, line 55: It did appear to me that the idiomatic proto thing to do would be to keep the original field number of two. However since this proto is only consumed by ourselves it seemed nice to keep the one field of the message as field number 1.

PR #1366, opened 2020-08-18
EvAug cuts out states; progress on PossibilisticNI

aferr:
The EvAug semantics is refactored so that it no longer is a state ->
trace -> state -> trace -> Prop relation but instead just a trace ->
trace -> Prop relation. This removes the need to have various
propositions about the states and traces involved in the step relation
agreeing. This also updates the statement of possibilistic NI and the
single step theorem. In the process, some problems with both of those
theorems are fixed. There is also some progress on the proof
for possibilistic NI. It is mostly done aside from some smaller lemmas
that I think are true but are admitted (the proof of the single-step
theorem is also still admitted).

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -7,6 +7,7 @@ From OakIFC Require Import
     Events
     LowEquivalences.
 
+
 (*
 This is the top-level candidate security condition. This is a
 "possibilistic security condition". A possibilistic security condition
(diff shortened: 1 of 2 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -53,34 +54,107 @@ https://www.cs.cornell.edu/andru/papers/csfw03.pdf
     ***
 *)
 
-Definition conjecture_possibilistic_ni := forall ell s11 s21 s1n t1n,
-    (state_low_eq ell s11 s21) /\
-    (step_system_ev_multi s11 [] s1n t1n) ->
-    (exists s2n t2n,
-        (step_system_ev_multi s21 [] s2n t2n) /\
+Definition is_init(t: trace) := length t = 1.
```
jieungkim, line 57: For clarification, 0 or 1?

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -53,34 +54,107 @@ https://www.cs.cornell.edu/andru/papers/csfw03.pdf
...
+    step_system_ev_multi t1 t2 ->
+    step_system_ev_multi t2 t3 ->
+    step_system_ev_multi t1 t3.
+Proof.
+Admitted.
+
 Theorem possibilistic_ni: conjecture_possibilistic_ni. Proof.
 unfold conjecture_possibilistic_ni.
-intros ell s11 s21 s1n t1n [H1 H2].
+intros ell t1_init t2_init t1n [H1 [H2 [H3 H4]]].
```
jieungkim, line 125: One of my suggestion in the proof is... using meaningful names for hypotheses instead of using Hn if it is not too much bothersome.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -53,34 +54,107 @@ https://www.cs.cornell.edu/andru/papers/csfw03.pdf
...
-        (trace_low_eq ell t12 t22)).
+Theorem possibilistic_ni_1step: forall ell t1 t2 t1',
+    (trace_low_eq ell t1 t2) ->
+    (step_system_ev t1 t1') ->
+    (exists t2',
+        (step_system_ev t2 t2') /\
+        (trace_low_eq ell t1' t2')).
 Admitted. (* NOTE: work in progress *)
 
+Theorem app_emp: forall {T} (x: T) l,
```
jieungkim, line 76: cons_not_nil?  (By referring the convention in https://coq.inria.fr/library/Coq.Lists.List.html  - app_cons_not_nil).

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: It's actually 1. It was effectively 0 before because the initial state was always "[]", but really the transition relation can't take a step from the empty state! The initial trace has to be a trace including just one state.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: This is done now in the latest commit.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: That doesn't exactly work, but nil_cons almost does.

I now have nil_cons_rev as a separate theorem because this is needed twice. This theorem seems inelegant to have though. Is there a way to get rid of it?

Actually, if you have time can you take a look at the proof of t_upd_mono_nil? There is one last proof obligation that I can't see how to prove (there is a line with "admit"), but it seems almost like something that should be in a standard library. I also can't find it.

jieungkim: I am sorry. but, I couldn't find "t_upd_mono_nil" in my current local repo. Could you let me know where it is?

jieungkim: Please add

" destruct p; reflexivity."

and let me know if it is working in your side as well.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Resolving, but unresolve if you want to.

jieungkim: Thanks for the comment. I understood that :D

PR #1367, opened 2020-08-18
Deduplicate Message, Encodable and Decodable

### LINKED ISSUE #1224: de-duplicate `Message`, `Encodable`, `Decodable` types between Runtime and SDK

### LINKED ISSUE #1381: `oak_derive` does not work for numeric enums
When trying to automatically derive HandleVisit for numeric enums (e.g. https://github.com/project-oak/oak/blob/main/oak_services/proto/log.proto#L40) the resulting code does not compile.

conradgrobler:
Deduplicate `Message` struct and `Encodable` and `Decodable` traits between Runtime and SDK. Moving the `Encodable` and `Decodable` traits also required moving the `HandleVisit` trait.

The main changes were:

- Move `Message` struct and `Encodable` and `Decodable` traits from the SDK to `oak_io`
- Move `HandleVisit` trait and associated functionality from the SDK to `oak_io`
- Add `oak_io` as a dependency to examples to support derivation of `HandleVisit` trait on protos
- Changre runtime to use `Message` struct and `Encodable` and `Decodable` traits in `oak_io`

Fixes #1224
Fixes #1381

# Checklist

      construction.


oak_io/src/handle.rs
```diff
@@ -43,3 +44,201 @@ impl std::fmt::Debug for WriteHandle {
...
+handle_visit_blanket_impl!((), f64, f32, i32, i64, u32, u64, bool, String, Vec<u8>);
+
+// Provide an implementation for the Any type in Prost.
+handle_visit_blanket_impl!(prost_types::Any);
+
+// Provide an implementation for oak_abi type that an implementation cannot be derived for.
+// It does not contain handles, so a blanket impl is sufficient.
+handle_visit_blanket_impl!(oak_abi::proto::oak::application::ConfigMap);
+
+pub use oak_derive::HandleVisit;
```
rbehjati, line 196: What is this? Does not it cause a circular dependency? Perhaps add a comment to explain what it does.

rbehjati (approved):
LGTM.

ipetr0v (approved):

oak_io/src/handle.rs
conradgrobler: This allows other crates to use  `#[derive(HandleVisit)]` without having to explicitly reference the `oak_derive` crate. Added a comment.

rbehjati: Thanks. But there is still a circular dependency: `oak_io` is dependent on `oak_derive` and `oak_derive` is dependent on `oak_io`. Does Rust/Cargo generally allow circular dependency? 
Can `oak_derive` be merged into `oak_io`?

conradgrobler: Cargo does not allow circular references. But it is not really a circular reference. `oak_derive` is not dependent on `oak_io`, it is only the integration tests that depends on both `oak_derive` and `oak_io` to test derivation of the `HandleVisit` trait. Seeing that the integration tests end up being built as a separate binary package, the dependency graph is not circular.

The `oak_derive` crate could be merged into `oak_io`, but it seems to be considered standard practice in rust to keep the macro derivation in a separate crate so that other crates can still use the trait even if they don't want to use derivation. It would be possible to move the `oak_derive` integration tests into the `oak_io` package to remove the need for the dev dependency, but I am not sure this would make it less confusing as the tests would be removed from the code it is testing.

rbehjati: Thanks for the clarification Conrad. 
I did not know about the practice of keeping the macro derivation in a separate package. I agree that it is better to keep the integration test close to the code it is testing.

PR #1375, opened 2020-08-20
Introspection server: Expose endpoint to load introspection events

juliettepretot:
Exposes endpoint on the introspection server for a future front-end client to load the list introspection events

# Checklist

      construction.


oak_runtime/src/introspect.rs
```diff
@@ -120,6 +122,23 @@ fn handle_request(
             "Object Counts",
             &runtime.html_counts(),
         ))));
+    // Endpoint to load the list introspection events, serialized via protobuf
+    } else if path == "/introspection-events" {
+        let events_message = Events {
+            events: Vec::from(runtime.introspection_event_queue.lock().unwrap().clone()),
```
juliettepretot, line 128: I need to clone here since `Vec::from` won't accept the `MutexGuard` wrapping the `VecDeque`. Is there a better way to do this? Should I leave comment?

oak_runtime/src/introspect.rs
```diff
@@ -120,6 +122,23 @@ fn handle_request(
...
+
+        let mut buffer: Vec<u8> = Vec::new();
+        events_message.encode(&mut buffer).unwrap();
+
+        let mut response = Response::new(Body::from(buffer));
+        // Set the Access-Control-Allow-Origin header, so this endpoint can be
+        // called from within a browser.
+        response
+            .headers_mut()
+            .insert(ACCESS_CONTROL_ALLOW_ORIGIN, HeaderValue::from_static("*"));
```
juliettepretot, line 139: Allows all origins to call this, the least secure policy, but seems fair for our case.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

Issue #1378, opened 2020-08-20
Create an `oak_sign` crate


ipetr0v:
Currently Wasm module signatures are created by a `sign_module` script (that uses `minisign` and adds a new dependency to the Docker image).
Instead we can create a separate `oak_sign` crate, that will sign Wasm Oak modules and generate self-contained signature files (including the hash, the public key, and the signature, in a single file).

Originated from https://github.com/project-oak/oak/pull/1312/files#r472943185
cc @tiziano88

PR #1384, opened 2020-08-22
Create `oak_sign` crate

ipetr0v:
This change adds `oak_sign` crate, that can:
- Generate Ed25519 keys
- Sign Wasm modules
- Verify Wasm module signatures
- Save and load keys/signatures using PEM format

Fixes https://github.com/project-oak/oak/issues/1378

# Checklist
- [x] Pull request includes prototype/experimental work that is under construction.


oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    rand,
+    signature::{self, KeyPair},
+};
+use std::{
+    collections::HashMap,
+    fs::{read, write},
+};
+use structopt::StructOpt;
+
+/// Command line options for the Oak sign.
```
tiziano88, line 52:
```suggestion
/// Command line options for `oak_sign`.
```

oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub public_key: String,
+}
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "Sign Wasm module")]
+struct Sign {
+    #[structopt(long, help = "Input private key file path")]
+    pub private_key: String,
+    #[structopt(long, help = "Input Wasm module path")]
+    pub module: String,
```
tiziano88, line 83: Is this specific about Wasm? AFAICT it just signs an arbitrary file, so I think the naming should be more generic. For instance, it may be used to sign other things in the future if we wanted to. So maybe just `--input` or `--input_file`?

oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
+const SIGNATURE_TAG: &str = "SIGNATURE";
+
+/// Creates a PEM structure for encoding.
+fn create_pem(tag: &str, contents: &[u8]) -> Pem {
+    Pem {
+        tag: tag.to_string(),
+        contents: contents.to_vec(),
+    }
+}
```
tiziano88, line 108: I'm curious why you decided to use PEM for storing this output, could you explain how you got there? I imagine other options may be protobuf or TOML or JSON, curious how they compare.

(I'm not suggesting changing to any of those format, I would just like to learn more about the differences myself)

oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+                .with_context(|| format!("Couldn't read private key file {}", &opt.private_key))?;
+            let module_bytes = read(&opt.module)
+                .with_context(|| format!("Couldn't Wasm module file {}", &opt.module))?;
+
+            // Sign Wasm module
+            let private_key_pem =
+                parse(private_key_file).context("Couldn't parse PEM encoded private key file")?;
+            let key_pair =
+                signature::Ed25519KeyPair::from_pkcs8(&private_key_pem.contents.as_ref())
+                    .expect("Couldn't parse PKCS encoded private key");
```
tiziano88, line 151:
```suggestion
                    .expect("Couldn't parse PKCS8 encoded private key");
```

tiziano88 (approved):
LG but please wait for @daviddrysdale 's review too

oak_sign/src/main.rs
ipetr0v: I was thinking of PEM as a key format that everyone is familiar with (and got used to).
Also PEM implies using base64 for encoding keys, which would not be obvious in TOML or JSON without an additional explanation.

ipetr0v: Thus, using non-standard formats would create similar problem as described in https://github.com/project-oak/oak/issues/1385:
- Someone, who would want to check the signature would need to go deep in the documentation in order to understand, in what encoding keys and signatures are stored.

oak_sign/src/main.rs
ipetr0v: Done

oak_sign/src/main.rs
ipetr0v: Done

oak_sign/src/main.rs
ipetr0v: Done

oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[structopt(long, help = "Output signature path")]
+    pub signature: String,
+}
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "Verify Wasm module signature")]
+struct Verify {
+    #[structopt(long, help = "Input signature path")]
+    pub signature: String,
+    #[structopt(long, help = "Input Wasm module path")]
```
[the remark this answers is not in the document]
ipetr0v: Done

oak_sign/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+enum Command {
+    Generate(Generate),
+    Sign(Sign),
+    Verify(Verify),
+}
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "Generate Ed25519 key pair")]
+struct Generate {
+    #[structopt(long, help = "Output private key file path")]
```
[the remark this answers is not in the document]
ipetr0v: Done
I think so, I'm just using standard PEM. Although I'm packing public key and signature to the same file.

Issue #1385, opened 2020-08-24
Move from `minisign` to `ring`


ipetr0v:
After finishing https://github.com/project-oak/oak/pull/1384 we will be able to switch from `minisign` library used for signing Wasm modules - to `ring` library.

The problems with `minisign` are:
- It creates individual files for public keys and signatures (can't use PEM format)
- Uses a non-standard file format
- Is self sufficient, i.e. doesn't use any established Rust crypto library

Issue #1066, opened 2020-06-02
Implement labels for examples


tiziano88:
Similar to #972, but for the rest of the examples (which presumably will need simpler labels).

ipetr0v:
Are hash-based labels already supported in Oak?

conradgrobler:
Regarding the use of signature labels in the TIR example, I don't think that hash-based labels or signature-based labels are appropriate for the TIR example, as it should use per-user labels to be secure.

If hash- or signature-based labels are used, the output from the node is public once the data leaves the node. If an untrusted node receives the output (rather than the gRPC server pseudo node) then the response can be sent anywhere, not only to the user who requested it.

ipetr0v:
But if we will use per-user labels, will Oak be able to send requests to an external database?

conradgrobler:
Sending a query to an external database is not appropriate for private information retrieval either. The typical use case for private/trusted information retrieval is to not leak the sensitive information that is required to do the query. If the sensitive information that should be protected is removed by the node during declassification, the query can no longer be based on this sensitive data.

My understanding was that the external database would be sent into the node (perhaps in batches if it is too big to be send all at once) that has the user data, rather than the node making a query.

tiziano88:
I think I agree with @conradgrobler .

@ipetr0v how do you think using a signature-based label helps with this example in particular? Perhaps could you describe the workflow you have in mind including who signs what, and why they should be trusted?

ipetr0v:
Currently we have two modules in TIR:
- the **TIR** itself
    - It processes specific user requests
- and a **database_proxy**
    - Is a generic implementation of private information retrieval, that requests a specific database entry id without revealing it

Client sends requests for a specific database entry **E** to TIR, but client doesn't trust TIR module, because it's just a specific implementation created by a third-party. Client trusts the **database_proxy** (ex. that was signed by Google) to request **E** from a database. Thus, client assigns a Google's public key to its data, because only **database_proxy** is trusted to know **E** and request it from an external database without revealing it.

conradgrobler:
What labels are associated with TIR and database_proxy? Where does the output from database_proxy go?

ipetr0v:
There are no labels for TIR and a `public_key` label for `database_proxy`.

`database_proxy` works as follows:
- Some other module requests **E** from it
- `database_proxy` sends multiple requests to an external database and iteratively searches for an entry corresponding to **E**
- Then it returns the found entry to the module that asked for it

[...]

PR #1387, opened 2020-08-24
Move from minisign to ring

ipetr0v:
This change switches `minisign` library for `ring`.

Fixes https://github.com/project-oak/oak/issues/1385

# Checklist


docs/development.md
```diff
@@ -288,5 +288,8 @@ In order create a signature `hello_world.sign` for a Wasm module
 `hello_world.wasm` use the following script:
```
daviddrysdale, line 288: (for a future PR)

This text talks about how to generate a signature of a module, but I don't think there's any explanation of *why* you might want such a thing.   Could we add an explanation somewhere (probably in concepts.md, possibly adapted from your helpful [explanation on Slack](https://project-oak.slack.com/archives/CHE9E13C3/p1598369605051800)) of what signature labels mean, then reference it from here?

oak_loader/src/options.rs
```diff
@@ -223,25 +223,23 @@ fn create_sign_table(opt: &Opt) -> anyhow::Result<SignatureTable> {
...
-                    signature,
-                });
+                let loaded_signature = match &signature_location {
+                    SignatureLocation::Path(path) => {
+                        let signature_file = read(&path)
+                            .with_context(|| format!("Couldn't read signature file {}", &path))?;
+                        parse_pem_signature(&signature_file)
+                            .with_context(|| format!("Couldn't parse signature file {}", &path))?
+                    }
+                    SignatureLocation::Url(_url) => unimplemented!(),
```
daviddrysdale, line 236: Could we get a `TODO(#999)` here please?

oak_runtime/src/node/wasm/mod.rs
```diff
@@ -847,9 +847,8 @@ fn wasm_node_privilege(wasm_module_bytes: &[u8], signatures: &[Signature]) -> No
     let mut integrity_tags = hashset! { hash_tag };
 
     for signature_item in signatures.iter() {
-        let signature_tag = oak_abi::label::web_assembly_module_signature_tag(
-            &signature_item.public_key.to_bytes(),
-        );
+        let signature_tag =
+            oak_abi::label::web_assembly_module_signature_tag(&signature_item.public_key);
         confidentiality_tags.insert(signature_tag.clone());
```
daviddrysdale, line 852: (pre-existing) Quick question to check my understanding: would we expect this secrecy label to normally use disjunctions rather than (as here) conjunctions, when disjunctions are available?  

AIUI this currently builds a label like `code:hash ∧ pubkey:A ∧ pubkey:B ∧ pubkey:C`, meaning that the client has to specify *all* of those tags.

tiziano88: Note that this function computes the **privilege** of the node, i.e. the **maximum authority** that the node can use in declassifying (or endorsing) data.

This means that, for instance, a node with privilege `code:hash ∧ pubkey:A ∧ pubkey:B ∧ pubkey:C` may declassify **any of** those tags; so data labelled with a confidentiality label of `pubkey:A` (alone) may be declassified by this node, and so can data labelled with `pubkey:B`, etc.

Please do let me know if it is still unclear

examples/trusted_information_retrieval/signatures.toml
```diff
@@ -1,4 +1,5 @@
 # database_proxy
-[[signatures.3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f]]
-public_key = "examples/certs/minisign/minisign.pub"
-signature = "examples/trusted_information_retrieval/database_proxy.sign"
+[signatures]
+3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f = [
+  { path = "examples/trusted_information_retrieval/database_proxy.sign" },
+]
```
tiziano88, line 5: Note that this key-value format now splits the information about the hash of the code to this file, while public key and signature are in the file pointed to by path. 

I think it is worth trying to think ahead how this will be used in the future, which I imagine will be a verifiable log or a simple database that ties together hash, public key, and signature of each module. e.g. if you imagine that as a mySQL table, it would be a table with three columns, so I think it makes sense to encode the three pieces of information in the same file to begin with.

WDYT?

Can be a follow-up PR, but I'm mostly curious if you think it will evolve in that direction, or do you have something else in mind @ipetr0v

oak_runtime/src/node/wasm/tests.rs
```diff
@@ -73,15 +73,9 @@ fn start_node(
...
-fn create_signature(public_key_path: &str, signature_path: &str) -> Signature {
-    let public_key =
-        minisign::PublicKey::from_file(public_key_path).expect("Could not read public key file");
-    let signature =
-        minisign::SignatureBox::from_file(signature_path).expect("Could not read signature file");
-    Signature {
-        public_key,
-        signature,
-    }
+fn create_signature(signature_path: &str) -> Signature {
```
tiziano88, line 76: I may be missing something, but it seems that this is loading a signature, not creating one?

oak_runtime/src/node/wasm/mod.rs
daviddrysdale: Doh, it's a privilege not a label &ndash; apologies for not reading closely enough.

docs/development.md
ipetr0v: I have a partial explanation here:
https://github.com/project-oak/oak/pull/1396

Not sure if it's enough

examples/trusted_information_retrieval/signatures.toml
ipetr0v: I was thinking that it would look like this:
```toml
[signatures]

# database_proxy
3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f = [
  { path = "examples/trusted_information_retrieval/database_proxy.sign" },
  { url = "https://oaksignatures.io/signature1.sign" },
  { url = "https://oaksignatures.io/signature2.sign" },
]

# some_other_module
3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f = [
  { url = "https://oaksignatures.io/signature3.sign" },
]
```
So that it will just contain paths/URLs of signature files. And thus signatures will accessible just by a URL.

I'm also not sure if we will need a verifiable log for signatures (we may only need it for the Oak binary hash itself).

tiziano88:
adding @rbehjati as reviewer to try and disseminate knowledge and spread review a bit more across the team :)

examples/trusted_information_retrieval/database_proxy.sign
```diff
@@ -1,4 +1,8 @@
-untrusted comment: signature from minisign secret key
-RWTdfGocqvKfGVf1MpX2AnaNRZh+ov+9ZPoSdMS2C9tCqd9fmvqksLutCxI4zYy9hYqUnZf10RPAcw5GfggDCZBZfLZCegLoRwc=
-trusted comment: timestamp:1597147477	file:3606a86d9fa01d223dc3cb092c1aceb9e3afe0ed9799ac5d5b214b1610209f8f
-wMrsDIuJOM4556brM1vAVqUsbkXl/tPjsnb1SNzPmBaRlPdHteTomSXcfIgK8LUk/HeZluWBS7Y+P6YcpiF4Ag==
+-----BEGIN PUBLIC KEY-----
+f41SClNtR4i46v2Tuh1fQLbt/ZqRr1lENajCW92jyP4=
+-----END PUBLIC KEY-----
+
```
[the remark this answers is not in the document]
daviddrysdale: No, I mean a separate source for the public key.

In a non-Oak context, if I were to download some app from a website, and also got both a signature and a public key from the same website, then it's tempting to verify all three together.  However, doing so doesn't really tell me anything &ndash; anyone who takes over the website can modify all three:
 - create a hacked version of the app
 - create a new keypair and serve the public key
 - use the new private key to create a new signature over the hacked app, and serve that.

For the signature to mean anything, I need to trust the public key, which means I need to acquire it by some separate path, i.e. a PKI.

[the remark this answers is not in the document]
tiziano88: @daviddrysdale I see your point, but **I think** it does not apply here, since this is handling signatures for pieces of code, but not making any assumptions about how someone else (dat owner) would label their data with something that refers to the public key corresponding to those signatures.

This is similar to signing an Android application, and the play store checking that the signature indeed matches a public key which is bundled together with it. How someone (an end user, most likely) would get to find out about and trust such public key in the first place, is a separate problem (and one that is yet unsolved for Oak, I would say).

For now, from the data owner point of view, we expect that they magically know what public keys to trust, and would just label data accordingly, then Oak would just respect those labels.

How to make public keys discoverable, tie them to higher level identities, make sure there are no split view attacks, etc. are problems that I hope will be somewhat solved / solvable by something like key transparency / keybase, or some other PKI. Speaking of which, do you think Tink would help with any of these? (in my mind it serves a somewhat similar purpose but on the server side, i.e. handling private key rotation and matching, but maybe there is something more that it may be used for).

I also think we should look a bit more closely at how similar mechanisms work on Android, iOS and macOS, since I think there are some similarities.

examples/trusted_information_retrieval/signatures.toml
tiziano88: But why not make the hash part of the thing that is downloaded via the URL? To avoid fetching them upfront?

examples/certs/ed25519/README.md
```diff
@@ -0,0 +1,12 @@
+# Example keys
```
tiziano88, line 1: I don't remember if I commented on this before (cannot find it now, at least), but why are these under the existing `certs` folder? I would expect a `keys` folder to be more appropriate?

examples/trusted_information_retrieval/signatures.toml
tiziano88: this is what I am suggesting:

```toml
signatures = [
  # database_proxy
  { path = "examples/trusted_information_retrieval/database_proxy.sign" },
  { url = "https://oaksignatures.io/signature1.sign" },
  { url = "https://oaksignatures.io/signature2.sign" },
  # some_other_module
  { url = "https://oaksignatures.io/signature3.sign" },
]
```

docs/development.md
```diff
@@ -288,5 +288,8 @@ In order create a signature `hello_world.sign` for a Wasm module
...
 
 ```bash
-./scripts/sign_module -i examples/hello_world/bin/hello_world.wasm -o examples/hello_world/bin/hello_world.sign
+./scripts/oak_sign sign \
+  --private-key=examples/certs/ed25519/test.key \
+  --input-file=examples/hello_world/bin/hello_world.wasm \
```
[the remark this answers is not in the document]
ipetr0v: After a discussion in https://github.com/project-oak/oak/issues/1066, we decided not to use signatures in the `truster_information_retrieval` example: https://github.com/project-oak/oak/issues/1407

rbehjati:
I think the following line must be updated too:
https://github.com/project-oak/oak/blob/b8e35fae6a43dbb34cd0abc829ea979db43d6632/oak_abi/proto/label.proto#L72

docs/development.md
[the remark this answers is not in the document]
rbehjati: Then perhaps add signing and a `signature.toml` file to the `hello_world` example (if there is not already a PR for that).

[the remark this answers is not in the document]
ipetr0v: I'll repurpose the following PR (https://github.com/project-oak/oak/pull/1392) for this

oak_runtime/src/lib.rs
```diff
@@ -1255,3 +1253,31 @@ pub fn sha_256_hex(bytes: &[u8]) -> String {
...
     hex::encode(hash_value)
 }
+
+// PEM file tags.
+pub(crate) const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
+pub(crate) const SIGNATURE_TAG: &str = "SIGNATURE";
+
+/// Parses public key and signature encoded using PEM format.
+/// https://tools.ietf.org/html/rfc1421
+pub fn parse_pem_signature(signature_pem: &[u8]) -> Result<Signature, OakStatus> {
```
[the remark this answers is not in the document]
ipetr0v: Similar function is also defined in `oak_runtime`.
But currently we don't have an `oak_common` crate. Maybe we'll define it, but there is nothing much to be put there.

oak_loader/src/options.rs
```diff
@@ -223,25 +223,26 @@ fn create_sign_table(opt: &Opt) -> anyhow::Result<SignatureTable> {
...
-                let signature = SignatureBox::from_file(&signature_item.signature)
-                    .context("Couldn't parse signature file")?;
-                debug!("Parsed signature: {}", signature.to_string());
-                parsed_signatures.push(Signature {
-                    public_key,
-                    signature,
-                });
+                let loaded_signature = match &signature_location {
+                    SignatureLocation::Path(path) => {
+                        let signature_file = read(&path)
```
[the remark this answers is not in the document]
ipetr0v: I think in a case of a corrupted file it's better for the maintainer (a person who runs Oak) to just exclude it from the `signatures.toml`, because it's a critical error and it might be easy to miss it as an `error!` log.
And if a signature couldn't be loaded, Oak will work fine, but clients wouldn't be able to connect to Oak - and it's harder to check.

oak_runtime/src/lib.rs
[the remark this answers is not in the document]
rbehjati: Do you mean there are other functions that are also used only in tests and in oak_loader?

rbehjati (approved):
LGTM!

oak_runtime/src/lib.rs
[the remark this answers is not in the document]
ipetr0v: No, I think there are not, that's why I didn't think about creating an `oak_common` crate, since it will be too small.

oak_sign/src/main.rs
```diff
@@ -152,11 +161,13 @@ fn main() -> anyhow::Result<()> {
                     .expect("Couldn't parse PKCS8 encoded private key");
             let public_key_bytes = key_pair.public_key();
             let signature_bytes = key_pair.sign(&input_file_bytes);
```
tiziano88, line 163: Is it wise to generate a hash, store that in the signature file, but then calculate the actual signature on the original input bytes instead of the computed hash (which will presumably still sign the hash, though it may or may not be the same hash)? It seems it may make more sense to just sign the computed hash (even if it ends up being re-hashed as part of the signature)?

@benlaurie do you have an opinion on this?

tiziano88 (approved):

oak_sign/src/main.rs
ipetr0v: [Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519) already uses SHA-512 for signing.
I think if we will start signing SHA-256 hashes with Ed25519 it may raise a lot of questions from the Oak application developers in the future.

ipetr0v: > Is it wise to generate a hash, store that in the signature file, but then calculate the actual signature on the original input bytes instead of the computed hash

Hash that is stored in the file is used just as a reference, so when we want to check a Wasm file's signature, we compute its hash, and then find a corresponding file, that contains the same hash.

tiziano88: Let's follow up separately, so we don't block this PR. Could you create a separate issue to track, so we don't forget?

ipetr0v: Created an issue https://github.com/project-oak/oak/issues/1437

PR #1388, opened 2020-08-25
Fix oak_runtime tests and formatting

conradgrobler:
# Checklist

      construction.


tiziano88 (approved):
Thanks for cleaning up after my mistake. I wonder whether we should have a more resilient logic for running tests, perhaps run non-workspace cargo files instead?

conradgrobler:
No problem :)

Another option is to store the paths to the workspace roots in a config file and use that to validate whether the workspace detection is correct, so that the paths for testing, formatting, etc. are based on two independent sources of information.

Issue #1344, opened 2020-08-11
Use Wasm signature labels in clients


ipetr0v:
After finishing https://github.com/project-oak/oak/issues/1274 we need to add the ability to add Wasm signature labels in example clients.
For example we can add this to the Trusted Information Retrieval client.

PR #1392, opened 2020-08-25
Use signature label in Private Set Intersection

ipetr0v:
This change adds a signature label to Private Set Intersection example.

Fixes https://github.com/project-oak/oak/issues/1344
Ref https://github.com/project-oak/oak/issues/1066

# Checklist


ipetr0v:
@tiziano88 since we will not add Wasm signature labels to TIR, but still need an example with it for CI, what could be a better choice for it?
I think `private_set_intersection` is a good fit.

tiziano88:
Yes I agree, I think that's a good fit

examples/private_set_intersection/client/private_set_intersection.cc
```diff
@@ -68,7 +68,12 @@ int main(int argc, char** argv) {
 
   // TODO(#1066): Use a more restrictive Label, based on a bearer token shared between the two
   // clients.
-  oak::label::Label label = oak::PublicUntrustedLabel();
+  std::string public_key_base64 = "f41SClNtR4i46v2Tuh1fQLbt/ZqRr1lENajCW92jyP4=";
```
tiziano88, line 71: Link to where this is coming from in a comment.

examples/private_set_intersection/client/private_set_intersection.cc
```diff
@@ -68,7 +68,12 @@ int main(int argc, char** argv) {
...
+  std::string public_key;
+  if (!absl::Base64Unescape(public_key_base64, &public_key)) {
+    LOG(FATAL) << "Could not decode public key: " << public_key_base64;
+  }
+  oak::label::Label label = oak::WebAssemblyModuleSignatureLabel(public_key);
```
tiziano88, line 76: Could we have part of the example that uses an invalid key too? Please make sure it runs as part of CI too, either as example or as test.

examples/private_set_intersection/private_set_intersection.sign
```diff
@@ -3,10 +3,10 @@ f41SClNtR4i46v2Tuh1fQLbt/ZqRr1lENajCW92jyP4=
 -----END PUBLIC KEY-----
 
 -----BEGIN SIGNATURE-----
-XTMXZqNPeVO5mfT+GURXMBXU3GrjVtIVlpdrnACbkgePpszy+RuL7wzrok6bU+P5
-1JY5gTU4vSU8kljTOzJvDw==
+z/qheXvmNUY/1YY06sw+1KQ0KjD9JF3/8aXHn9M8QtjTl0VY8rHWhnrWsWFiPmMG
+Ee7ZndTs3yVQRUIa7BMoBg==
```
tiziano88, line 7: Out of curiosity, if you re-sign the same file with the same key, does it produce the same signature? Or does it use some nondeterministic value as part of the signature?

oak_runtime/src/node/wasm/mod.rs
```diff
@@ -837,20 +831,28 @@ impl WasmNode {
 /// Computes the [`NodePrivilege`] granted to a WebAssembly Node running the specified WebAssembly
 /// module.
 /// Created [`NodePrivilege`] consists of Wasm module hash and signature.
```
tiziano88, line 833:
```suggestion
/// Created [`NodePrivilege`] consists of Wasm module hash and any matching signatures.
```

oak_runtime/src/node/wasm/mod.rs
```diff
@@ -815,15 +815,9 @@ impl WasmNode {
...
-        let signatures = signature_table
-            .values
-            .get(&node_configuration.wasm_module_name)
-            .cloned()
-            .unwrap_or_default();
-
         // We compute the node privilege once and for all at start and just store it, since it does
         // not change throughout the node execution.
-        let node_privilege = wasm_node_privilege(&wasm_module_bytes, signatures.as_ref());
+        let node_privilege = wasm_node_privilege(&wasm_module_bytes, signature_table);
```
tiziano88, line 820: I'm curious why this was changed, did the previous one not work any more?

tiziano88 (approved):

oak_runtime/src/node/wasm/mod.rs
ipetr0v: Yes, it was a bug

examples/private_set_intersection/client/private_set_intersection.cc
```diff
@@ -78,14 +81,36 @@ int main(int argc, char** argv) {
...
+  std::string invalid_public_key;
+  if (!absl::Base64Unescape(invalid_public_key_base64, &invalid_public_key)) {
+    LOG(FATAL) << "Could not decode public key: " << invalid_public_key_base64;
+  }
+  oak::label::Label invalid_label = oak::WebAssemblyModuleSignatureLabel(invalid_public_key);
+  auto invalid_stub = PrivateSetIntersection::NewStub(oak::ApplicationClient::CreateChannel(
+      address, oak::ApplicationClient::GetTlsChannelCredentials(ca_cert), invalid_label));
+  std::vector<std::string> set_2{"c", "d", "e"};
+  auto submit_status_2 = SubmitSet(invalid_stub.get(), set_2);
+  if (submit_status_2.ok()) {
```
tiziano88, line 108: Could you make this match a bit more specific? Currently it would probably succeed if there is no connectivity, for instance.

tiziano88 (approved):

examples/private_set_intersection/client/private_set_intersection.cc
```diff
@@ -68,7 +65,13 @@ int main(int argc, char** argv) {
 
   // TODO(#1066): Use a more restrictive Label, based on a bearer token shared between the two
   // clients.
-  oak::label::Label label = oak::PublicUntrustedLabel();
+  // Public key is stored in the `examples/keys/ed25519/test.pub` file.
+  std::string public_key_base64 = "f41SClNtR4i46v2Tuh1fQLbt/ZqRr1lENajCW92jyP4=";
```
rbehjati, line 69: Could this be read from the file instead of being hard coded here?

rbehjati (approved):

examples/private_set_intersection/client/private_set_intersection.cc
ipetr0v: Done

examples/private_set_intersection/client/private_set_intersection.cc
ipetr0v: Done

PR #1396, opened 2020-08-26
Update docs for signature labels

ipetr0v:
This change updates docs regarding Wasm signature labels.

# Checklist

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -322,6 +322,71 @@ represented by a message containing two Channels handles:
     connection itself; this represents the actual authority of the caller.
   - an empty integrity component.
 
+### Signature labels
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. Such signatures may be created by an Oak module reviewer who then
+publishes signatures and corresponding public keys on a public resource. By
+signing a module reviewers endorse it's functionality and show that they trust
+this particular module not to have any undocumented functionality. The same
+reviewer may also use distinct keys for different purposes, for example to use a
+separate public key to sign a family of related modules.
+
+In case when a client trusts a certain reviewer, they can download a
+corresponding reviewer's public key and use it as a data label. This will allow
+an Oak server to download and use new versions of modules signed by the same
+reviewer without client's notice, i.e. client wouldn't need to constantly update
+module hashes in data labels. It will also allow new versions of Oak modules to
+access old client's data saved by Oak.
+
+Reviewer's public key can also be assigned to outgoing requests by the client
+library, which was created by the Oak application developer. In this case client
+doesn't need to manually download a public key, but they will need to trust that
+the client library assigns these public keys to outgoing requests correctly.
+
+In order to more clearly show how Oak module signatures are used, here is a
+possible life cycle of an Oak application: from development to usage.
+
+#### Oak signature life cycle
+
+The parties involved in the signature life cycle are:
+
+- Application developer
+  - Person/organization who writes an Oak application
+- Oak module developer
+  - Person/organization who develops and publishes Oak Wasm modules to be used
+    by the **Application developer**
+- Reviewer
+  - Person/organization who reviews and signs Oak modules
+- Operator
+  - Person/organization who deploys Oak applications on a production server
+- Client
+  - Person/organization who sends data to a deployed Oak application
+
+The cycle itself looks as follows:
+
+- **Reviewer**
+  - Reviews existing Oak modules and checks their behavior
+  - Signs Oak modules and publishes corresponding signatures (as well as public
+    keys used for signing)
+- **Application developer**:
+  - Chooses which third-party Oak modules to use
+  - Creates an Oak application using these third-party Oak modules
+  - Publishes the application
+- **Operator**
+  - Downloads an Oak application
+  - Downloads files containing Oak modules' signatures/keys
+  - Deploys the Oak application and passes it a list of Oak module
+    signatures/keys
+- **Client**
+  - Downloads and runs a designated client library
+  - Client library downloads corresponding reviewers' public keys
+  - Client library assigns these keys as labels to the data and sends it to a
+    deployed Oak application
+
+For a step-by-step description on how to sign Oak modules see
+[Signing Oak Wasm modules](/docs/development.md#signing-oak-wasm-modules).
+
 ## Pseudo-Nodes
 
 An Oak Node is limited to exchanging messages with other Nodes via Channels; to
--- a/docs/development.md
+++ b/docs/development.md
@@ -10,6 +10,7 @@
   - [Build Runtime Server](#build-runtime-server)
   - [Run Runtime Server](#run-runtime-server)
   - [Run Example Client](#run-example-client)
+  - [Signing Oak Wasm modules](#signing-oak-wasm-modules)
 
 ## Quick Start
 
@@ -281,15 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
 
 ### Signing Oak Wasm modules
 
-Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
-scheme.
+As described in [Oak Concepts](/docs/concepts.md#labels) Oak Wasm modules can be
+signed using [Ed25519](https://ed25519.cr.yp.to/) scheme.
 
-In order create a signature `hello_world.sign` for a Wasm module
-`hello_world.wasm` use the following script:
+If you are reviewing an Oak module (for example `private_set_intersection.wasm`)
+and want to create a signature for it, first you need a private/public key pair.
+It can be generated using `oak_sign` utility:
+
+```bash
+./scripts/oak_sign generate \
+  --private-key=examples/keys/ed25519/test.key \
+  --public-key=examples/keys/ed25519/test.pub
+```
+
+It generates two key files (`test.key` and `test.pub`) that are encoded using
+[PEM](https://tools.ietf.org/html/rfc1421) format.
+
+Then a Wasm module can be signed using the following command:
 
 ```bash
 ./scripts/oak_sign sign \
-  --private-key=examples/certs/ed25519/test.key \
-  --input-file=examples/hello_world/bin/hello_world.wasm \
-  --signature=examples/hello_world/bin/hello_world.sign
+  --private-key=examples/keys/ed25519/test.key \
+  --input-file=examples/private_set_intersection/bin/private_set_intersection.wasm \
+  --signature=examples/private_set_intersection/bin/signature.pem
 ```
+
+This command generates a signature file `signature.pem` containing a public key,
+a Wasm module signature (also encoded with PEM) and corresponding Wasm module
+SHA-256 hash.
+
+Wasm module signatures should be passed to `oak_loader` by an Oak operator using
+the `--signatures-manifest=signatures.toml`, where `signatures.toml` file
+contains paths to signature files for each signed Oak Wasm module. Also, since
+each module can be signed by multiple entities, each module may have multiple
+signatures.
+
+The `signatures.toml` file can look like this:
+
+```toml
+signatures = [
+  { path = "examples/private_set_intersection/signature.pem" },
+  { path = "examples/private_set_intersection/additional_signature.pem" },
+]
+
+```
+
+Where `signature.pem` and `additional_signature.pem` are signatures that could
+be provided by different reviewers.
+
+Oak clients can download the `test.pub` public key (from the reviewer's public
+repository) and use it as a
+[label](/docs/programming-oak.md#using-an-oak-application-from-a-client) for
+requests sent to the corresponding Oak application.
--- a/docs/development.md
+++ b/docs/development.md
@@ -287,16 +287,14 @@ signed using [Ed25519](https://ed25519.cr.yp.to/) scheme.
 
 If you are reviewing an Oak module (for example `private_set_intersection.wasm`)
 and want to create a signature for it, first you need a private/public key pair.
-It can be generated using `oak_sign` utility:
+It can be generated with `oak_sign` utility using the following script (that
+will generate two key files: `test.key` and `test.pub`)
 
 ```bash
 ./scripts/oak_sign generate \
   --private-key=examples/keys/ed25519/test.key \
   --public-key=examples/keys/ed25519/test.pub
 ```
-
-It generates two key files (`test.key` and `test.pub`) that are encoded using
-[PEM](https://tools.ietf.org/html/rfc1421) format.
 
 Then a Wasm module can be signed using the following command:
 
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -356,26 +356,31 @@ Because the Oak Application is available as a gRPC service, clients written in
 any language that supports gRPC can use the service. For example in Rust:
 
 <!-- prettier-ignore-start -->
-[embedmd]:# (../examples/private_set_intersection/client/rust/src/main.rs Rust /.*web_assembly_module_signature_tag/ /Ok(request)*/)
-```rust
-  let public_key_tag = web_assembly_module_signature_tag(&public_key);
-  let mut label = Vec::new();
-  Label {
-      confidentiality_tags: vec![public_key_tag],
-      integrity_tags: vec![],
-  }
-  .encode(&mut label)
-  .context("Error encoding label")?;
-  Ok(PrivateSetIntersectionClient::with_interceptor(
-      channel,
-      move |mut request: Request<()>| {
-          request.metadata_mut().insert_bin(
-              oak_abi::OAK_LABEL_GRPC_METADATA_KEY,
-              MetadataValue::from_bytes(label.as_ref()),
-          );
-          Ok(request)
-      },
-  ))
+[embedmd]:# (../examples/private_set_intersection/client/rust/src/main.rs Rust /^.*create_client.*/ /^}/)
+```Rust
+fn create_client(
+    channel: Channel,
+    public_key: &[u8],
+) -> anyhow::Result<PrivateSetIntersectionClient<Channel>> {
+    let public_key_tag = web_assembly_module_signature_tag(&public_key);
+    let mut label = Vec::new();
+    Label {
+        confidentiality_tags: vec![public_key_tag],
+        integrity_tags: vec![],
+    }
+    .encode(&mut label)
+    .context("Error encoding label")?;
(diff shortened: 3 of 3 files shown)


docs/development.md
```diff
@@ -284,9 +284,60 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+./scripts/oak_sign sign \
+  --private-key=examples/certs/ed25519/test.key \
+  --input-file=examples/hello_world/bin/hello_world.wasm \
+  --signature=examples/hello_world/bin/hello_world.sign
+```
+
+The generated signature file `hello_world.sign` contains both a public key and a
+Wasm modules signature (also encoded with PEM).
+
+Wasm module signatures should be passed to `oak_loader` using the
```
tiziano88, line 320: by whom?

docs/development.md
```diff
@@ -284,9 +284,60 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+`--signatures-manifest=signature.toml`, where `signature.toml` file contains
+paths to signature files for each signed Oak Wasm module (each module is
+referred by its SHA-256 hash). Also, since each module can be signed by multiple
+entities, modules can have multiple signatures.
```
tiziano88, line 324:
```suggestion
entities, each module may have multiple signatures.
```

docs/development.md
```diff
@@ -284,9 +284,60 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+a client trusts a certain Oak Wasm modules developer, it can use a corresponding
+author's public key as a data label. This will allow an Oak server to download and
+use new versions of modules produced by this developer without client's notice,
+i.e. client wouldn't need to constantly update module hashes in data labels. It
+will also allow new versions of Oak modules to access old client's data saved by
+Oak.
+
 In order create a signature `hello_world.sign` for a Wasm module
-`hello_world.wasm` use the following script:
+`hello_world.wasm`, first we need a private/public key pair. It can be generated
```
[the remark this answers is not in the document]
ipetr0v: Added an updated version of the following comment https://github.com/project-oak/oak/issues/1274#issuecomment-670005123 in the `concepts.md`

docs/development.md
```diff
@@ -284,9 +284,60 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+```bash
+./scripts/oak_sign generate \
+  --private-key=examples/certs/ed25519/test.key \
+  --public-key=examples/certs/ed25519/test.pub
+```
+
+It generates key files that are encoded using
+[PEM](https://tools.ietf.org/html/rfc1421) format.
+
+Then a Wasm module can be signed using the following command:
```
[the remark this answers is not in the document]
tiziano88: I mean, what is the meaning behind signing a module? I guess some sort of endorsement of the functionality of the module itself, for instance. Also mention that the same reviewer may use distinct keys for different purpose, so each key is really endorsing a family of related modules, in a way, right?

docs/development.md
```diff
@@ -281,12 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+  --private-key=examples/certs/ed25519/test.key \
+  --input-file=examples/hello_world/bin/hello_world.wasm \
+  --signature=examples/hello_world/bin/hello_world.sign
+```
+
+This command generates a signature file `hello_world.sign` contains both a
+public key and a Wasm modules signature (also encoded with PEM).
+
+Wasm module signatures should be passed to `oak_loader` by an Oak server
+maintainer using the `--signatures-manifest=signature.toml`, where
```
tiziano88, line 314: Previously it sounded like the term is Oak application maintainer, but here it says Oak server maintainer, though I guess you meant the same thing in both, right?

docs/development.md
```diff
@@ -281,12 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+module (each module is referred by its SHA-256 hash). Also, since each module
+can be signed by multiple entities, each module may have multiple signatures.
+
+The `signature.toml` file can look like this:
+
+```toml
+[signatures]
+d584ac26782b91e5d47f35a1e0eca6836b7b6aaa09e624a3baea09b6f01ffe5b = [
+  { path = "examples/hello_world/bin/hello_world.sign" },
+]
```
tiziano88, line 325: I commented on the other PR, but I really think that the hash should be part of the signature. If a reviewer signs a Wasm module, the resulting file should have all the information necessary to link the signature, public key and module identifier. In the current form, imagine I sign something, I generate `hello_world.sign`, now I send it to you, and the file itself is not sufficient, I also need to tell you (in some other way) what the hash is. May as well make it part of the file, no? Again, if this were a mySQL table, it would have 3 columns I imagine, so it makes sense that these three pieces of information are part of a single entry.

ipetr0v: Updated

docs/concepts.md
```diff
@@ -322,6 +322,62 @@ represented by a message containing two Channels handles:
     connection itself; this represents the actual authority of the caller.
   - an empty integrity component.
 
+### Signature labels
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. Such signatures may be created by an Oak module reviewer and published
+on a public resource for Oak clients to download.
```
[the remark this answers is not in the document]
ipetr0v: Yes, clients download public keys, updated

docs/concepts.md
```diff
@@ -322,6 +322,62 @@ represented by a message containing two Channels handles:
...
+
+In case when a client trusts a certain reviewer, it can use a corresponding
+author's public key and use it as a data label. This will allow an Oak server to
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/concepts.md
```diff
@@ -322,6 +322,62 @@ represented by a message containing two Channels handles:
...
+The parties involved in the signature life cycle are:
+
+- Application developer
+  - Person/organization who writes an Oak application
+- Oak module developer
+  - Person/organization who develops and publishes Oak Wasm modules to be used
+    by the **Application developer**
+- Reviewer
+  - Person/organization who reviews and signs Oak modules
+- Maintainer
```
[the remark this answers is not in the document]
ipetr0v: Changed to operator

docs/concepts.md
```diff
@@ -322,6 +322,62 @@ represented by a message containing two Channels handles:
...
+client's data saved by Oak.
+
+In order to clearer understand how Oak module signatures are used, here is a
+possible life cycle of an Oak application: from development to usage.
+
+#### Oak signature life cycle
+
+The parties involved in the signature life cycle are:
+
+- Application developer
```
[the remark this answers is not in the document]
ipetr0v: I tried to merge these two lists, but if feels a little bit odd, since the second list should be a life cycle (too much text will be in one place)

docs/development.md
ipetr0v: Changed to `Oak operator` (similar to the life cycle)

docs/development.md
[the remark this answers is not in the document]
ipetr0v: Added this to `concepts.md`

docs/concepts.md
```diff
@@ -322,6 +322,66 @@ represented by a message containing two Channels handles:
     connection itself; this represents the actual authority of the caller.
   - an empty integrity component.
 
+### Signature labels
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. Such signatures may be created by an Oak module reviewer who then
+publish corresponding public keys on a public resource. By signing a module
```
tiziano88, line 329:
```suggestion
publishes corresponding public keys on a public resource. By signing a module
```

tiziano88, line 329: The main thing is that they publish the signatures. Then their public key is also assumed to be published already.

docs/concepts.md
```diff
@@ -322,6 +322,66 @@ represented by a message containing two Channels handles:
...
+key to sign a family of related modules.
+
+In case when a client trusts a certain reviewer, it can download a corresponding
+reviewer's public key and use it as a data label. This will allow an Oak server
+to download and use new versions of modules signed by the same reviewer without
+client's notice, i.e. client wouldn't need to constantly update module hashes in
+data labels. It will also allow new versions of Oak modules to access old
+client's data saved by Oak.
+
+In order to clearer understand how Oak module signatures are used, here is a
```
tiziano88, line 342:
```suggestion
In order to more clearly show how Oak module signatures are used, here is a
```

docs/development.md
```diff
@@ -281,15 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+It can be generated using `oak_sign` utility:
+
+```bash
+./scripts/oak_sign generate \
+  --private-key=examples/keys/ed25519/test.key \
+  --public-key=examples/keys/ed25519/test.pub
+```
+
+It generates two key files (`test.key` and `test.pub`) that are encoded using
+[PEM](https://tools.ietf.org/html/rfc1421) format.
```
tiziano88, line 299: I think the fact that they are PEM encoded is irrelevant and an implementation detail. Or am I missing something for which this is useful at all?

docs/development.md
```diff
@@ -281,15 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
 ./scripts/oak_sign sign \
-  --private-key=examples/certs/ed25519/test.key \
-  --input-file=examples/hello_world/bin/hello_world.wasm \
-  --signature=examples/hello_world/bin/hello_world.sign
+  --private-key=examples/keys/ed25519/test.key \
+  --input-file=examples/private_set_intersection/bin/private_set_intersection.wasm \
+  --signature=examples/private_set_intersection/bin/private_set_intersection.sign
+```
+
+This command generates a signature file `private_set_intersection.sign` contains
```
tiziano88, line 310:
```suggestion
This command generates a signature file `private_set_intersection.sign` containing
```

docs/development.md
```diff
@@ -281,15 +282,55 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+The `signatures.toml` file can look like this:
+
+```toml
+signatures = [
+  { path = "examples/private_set_intersection/private_set_intersection.sign" },
+]
+
+```
+
+Where SHA-256 value was obtained by:
```
tiziano88, line 328: There are no hashes to be manually passed around now.

docs/concepts.md
ipetr0v: Done

ipetr0v: Updated `publishes signatures and corresponding public keys`

docs/concepts.md
ipetr0v: Done

docs/development.md
ipetr0v: It's just a clarification, so the reader will understand what are key files. So we would not copy the `minisign` mistake, where documentation doesn't say what format is used for keys.

docs/development.md
ipetr0v: Done

docs/development.md
ipetr0v: Deleted

docs/concepts.md
```diff
@@ -322,6 +322,66 @@ represented by a message containing two Channels handles:
...
+  - Creates an Oak application using these third-party Oak modules
+  - Publishes the application
+- **Operator**
+  - Downloads an Oak application
+  - Downloads files containing Oak modules' signatures/keys
+  - Deploys the Oak application and passes it a list of Oak module
+    signatures/keys
+- **Client**
+  - Chooses which Oak modules' reviewer to trust
+  - Downloads corresponding reviewers' public keys
```
conradgrobler, line 378: This does not seem to be a realistic scenario. Only a very small number of users would feel comfortable with donwloading a public key and ensuring that the appropriate public key abel is attached to all data.

docs/concepts.md
```diff
@@ -322,6 +322,66 @@ represented by a message containing two Channels handles:
...
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. Such signatures may be created by an Oak module reviewer who then
+publishes signatures and corresponding public keys on a public resource. By
+signing a module reviewers endorse it's functionality and show that they trust
+this particular module not to have any undocumented functionality. The same
+reviewer may also use distinct keys for different purpose, for example to use a
+separate public key to sign a family of related modules.
+
+In case when a client trusts a certain reviewer, it can download a corresponding
```
conradgrobler, line 335: I don't think this will be a typical use case. The main use case would be where the end-user trusts the application developer, and the application developer ensures that the client application attaches the correct public key label to the data to ensure that it could be processed in future (e.g. after being stored) even if code must be updated (e.g. to fix a bug).

docs/development.md
```diff
@@ -281,15 +282,50 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
-  --private-key=examples/certs/ed25519/test.key \
-  --input-file=examples/hello_world/bin/hello_world.wasm \
-  --signature=examples/hello_world/bin/hello_world.sign
+  --private-key=examples/keys/ed25519/test.key \
+  --input-file=examples/private_set_intersection/bin/private_set_intersection.wasm \
+  --signature=examples/private_set_intersection/bin/private_set_intersection.sign
+```
+
+This command generates a signature file `private_set_intersection.sign`
+containing a public key, a Wasm modules signature (also encoded with PEM) and
```
rbehjati, line 311:
```suggestion
containing a public key, a Wasm module signature (also encoded with PEM) and
```

docs/concepts.md
```diff
@@ -322,6 +322,66 @@ represented by a message containing two Channels handles:
...
   - an empty integrity component.
 
+### Signature labels
+
+Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
+scheme. Such signatures may be created by an Oak module reviewer who then
+publishes signatures and corresponding public keys on a public resource. By
+signing a module reviewers endorse it's functionality and show that they trust
+this particular module not to have any undocumented functionality. The same
+reviewer may also use distinct keys for different purpose, for example to use a
```
rbehjati, line 332:
```suggestion
reviewer may also use distinct keys for different purposes, for example to use a
```

docs/concepts.md
rbehjati: Is it uncomfortable because users do not trust this approach or because it is too much work?

docs/development.md
```diff
@@ -281,15 +282,50 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+the `--signatures-manifest=signatures.toml`, where `signatures.toml` file
+contains paths to signature files for each signed Oak Wasm module. Also, since
+each module can be signed by multiple entities, each module may have multiple
+signatures.
+
+The `signatures.toml` file can look like this:
+
+```toml
+signatures = [
+  { path = "examples/private_set_intersection/private_set_intersection.sign" },
```
rbehjati, line 324:
```suggestion
  { path = "examples/private_set_intersection/reviewer_1.sign" },
```
I guess each signature in this file could be signed by a different reviewer. Is that correct?

docs/development.md
```diff
@@ -281,15 +282,50 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+]
+
 ```
+
+Oak clients can download the `test.pub` public key and use it as a
```
rbehjati, line 329:
```suggestion
Oak clients can download the `test.pub` public key (from the reviewer's public repository) and use it as a
```

docs/development.md
```diff
@@ -281,15 +282,50 @@ I0511 10:15:29.593106 244858 hello_world.cc:57] Response: HELLO AGAIN WORLDS!
...
+Then a Wasm module can be signed using the following command:
 
 ```bash
 ./scripts/oak_sign sign \
-  --private-key=examples/certs/ed25519/test.key \
-  --input-file=examples/hello_world/bin/hello_world.wasm \
-  --signature=examples/hello_world/bin/hello_world.sign
+  --private-key=examples/keys/ed25519/test.key \
+  --input-file=examples/private_set_intersection/bin/private_set_intersection.wasm \
+  --signature=examples/private_set_intersection/bin/private_set_intersection.sign
```
rbehjati, line 307:
```suggestion
  --signature=examples/private_set_intersection/bin/reviewer_1.sign
```
Just a suggestion. See also my suggestion for `signatures.toml` below :)

rbehjati (approved):
Thanks.

docs/concepts.md
conradgrobler: I think it would be confusing for most users.

docs/concepts.md
ipetr0v: Done
Added:
```
Reviewer's public key can also be assigned to outgoing requests by the client library, which was created by the
Oak application developer. In this case client doesn't need to manually download
a public key, but they will need to trust that the client library assigns these public keys to outgoing requests correctly.
```

docs/concepts.md
ipetr0v: Updated

docs/development.md
ipetr0v: Changed to `signature.pem` simply because we already have this file (in https://github.com/project-oak/oak/pull/1512)

docs/development.md
ipetr0v: Correct, Oak could use any number of public keys. Also the same reviewer can create multiple different signatures for different modules.
Added a second key and a comment:
```
Where `signature.pem` and `additional_signature.pem` are signatures that could
be provided by different reviewers.
```

conradgrobler (approved):

docs/development.md
tiziano88: I think in our case, we actually don't want others to use these keys for anything else. I am not sure why we would want to document them at all, if we do then obviously this becomes a public API and we cannot change it any more, which does not seem worthwhile to me. What use cases are you thinking in which someone would parse such keys and use them without relying on oak utilities?

tiziano88 (approved):
I am not sure the current version of the documentation would be particularly useful to anyone outside the Oak team, but it is a good start. Please let us iterate on it though, and make sure that it is eventually more convincing.

docs/development.md
ipetr0v: Deleted

PR #1397, opened 2020-08-26
Negative test for HTTP server pseudo-node

rbehjati:
Add a test to check that HTTP server node rejects unsecure requests

### DIFF
--- a/oak_runtime/src/node/http/tests.rs
+++ b/oak_runtime/src/node/http/tests.rs
@@ -17,24 +17,110 @@
 use super::*;
 use maplit::hashmap;
 use oak_abi::{label::Label, proto::oak::application::ApplicationConfiguration};
-use std::{fs, thread};
+use std::{fs, option::Option, thread::JoinHandle};
+
+struct HttpServerTester {
+    runtime: RuntimeProxy,
+    server_node_thread_handle: Option<JoinHandle<()>>,
+    oak_node_simulator_thread_handle: Option<JoinHandle<()>>,
+    notify_sender: Option<tokio::sync::oneshot::Sender<()>>,
+}
+
+impl HttpServerTester {
+    fn new(port: u32, with_simulator_thread: bool) -> HttpServerTester {
+        let runtime = create_runtime();
+        let server_node = create_server_node(port);
+        let (init_receiver, invocation_receiver) = create_communication_channel(&runtime);
+
+        // Start the server node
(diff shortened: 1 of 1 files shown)


conradgrobler (approved):

oak_runtime/src/node/http/tests.rs
```diff
@@ -17,72 +17,127 @@
...
+        }
+    }
+
+    fn cleanup(&mut self) {
+        if let Some(sender) = self.notify_sender.take() {
+            sender.send(()).unwrap_or_else(|()| {
+                debug!("Test node already dropped `notify_receiver`.");
+            })
+        }
+        let _ = self.server_node_thread_handle.take().map(JoinHandle::join);
```
tiziano88, line 67: Does this actually do anything? I would expect it just gets optimized away by the compiler. Either way, in general it is not good practice to use functional idioms (map, filter) just for their side effects.

oak_runtime/src/node/http/tests.rs
```diff
@@ -17,72 +17,127 @@
...
+        let (notify_sender, notify_receiver) = tokio::sync::oneshot::channel::<()>();
+        let runtime_proxy = runtime.clone();
+        // TODO(#1186): Use tokio instead of spawning a thread.
+        let server_node_thread = std::thread::spawn(move || {
+            server_node.run(runtime_proxy, init_receiver, notify_receiver)
+        });
+
+        // Simulate an Oak node that responds with 200 (OK) to every request it receives
+        // TODO(#1186): Use tokio instead of spawning a thread.
+        let mut oak_node_simulator_thread = None;
```
tiziano88, line 45: I think the mutable binding can be avoided, since `if / else` blocks are expressions in Rust (and not just statements).

rbehjati: Good point. Thanks.

oak_runtime/src/node/http/tests.rs
rbehjati: Looks like it does something. If the simulator thread is created in the case of `test_https_server_cannot_serve_http_requests`, then the test gets blocked because of the `wait_on_channel` in `oak_node_simulator` and the join here. If I remove this line, then the test ends without a problem. 

But I can see why using functional idioms only for their side effect is discouraged. I have fixed it in a separate PR.

rbehjati (commented):
Thanks for the review @tiziano88. I have addressed your comments in #1402.

PR #1401, opened 2020-08-27
Fix channel_write channel_read calls

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/README.md
+++ b/experimental/ni_coq/README.md
@@ -24,7 +24,12 @@ The following are dependencies of this project:
 - coq-mathcomp-finmap (1.5.0)
 - coq-record-update (0.2.0)
 
-All of the above may be installed with [opam](https://opam.ocaml.org/).
+All of the above may be installed with [opam](https://opam.ocaml.org/), and the
+names given above are opam package names. The coq package repo should first be
+added with by running:
+
+opam repo add coq-released
+[https://coq.inria.fr/opam/released](https://coq.inria.fr/opam/released)
 
 The included makefile has been generated with CoqProject: (see also:
 [the instructions for CoqProject](https://coq.inria.fr/refman/practical-tools/utilities.html))
--- a/experimental/ni_coq/README.md
+++ b/experimental/ni_coq/README.md
@@ -18,10 +18,13 @@ is in vfiles/RuntimeModel.v. Both are evolving.
 
 ## Instructions
 
-The following are dependencies of this project: coq (8.11.0) coq-mathcomp-finmap
-(1.5.0)
+The following are dependencies of this project:
 
-Both of the above may be installed with [opam](https://opam.ocaml.org/).
+- coq (8.11.0)
+- coq-mathcomp-finmap (1.5.0)
+- coq-record-update (0.2.0)
+
+All of the above may be installed with [opam](https://opam.ocaml.org/).
 
 The included makefile has been generated with CoqProject: (see also:
 [the instructions for CoqProject](https://coq.inria.fr/refman/practical-tools/utilities.html))
--- a/experimental/ni_coq/vfiles/EvAugSemantics.v
+++ b/experimental/ni_coq/vfiles/EvAugSemantics.v
@@ -49,11 +49,6 @@ Notation "n '--->' msg":= (EvL (OutEv msg) n.(nlbl)) (at level 10).
 Notation "n '<---' msg":= (EvL (InEv msg) n.(nlbl)) (at level 10).
 Notation "n '---'":= (EvL NilEv n.(nlbl)) (at level 10).
 
-Definition msg_is_head (ch: channel)(m: message): Prop :=
-    match ch.(ms) with
-        | [] => False
-        | m' :: ms' => m' = m
-    end. 
 
 Definition head_st (t: trace) :=
     match t with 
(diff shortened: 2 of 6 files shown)


aferr:
Tiziano (or possibly Ben could know): I am not sure if the way channel reads/writes work here is right. (Aside from the fact that read handles need to be made linear, which I think is not yet implemented)

Here channels have essentially a stack of messages, a set of read handles, and a set of write handles. When a channel gets read, the reader gets the top message of the stack, all the read handles, and all the write handles. Is this right?

Or should I think of it more like a stack of triples: (messages * read_handles * write_handles) ?

Or is there no stack, and it's just one triple.

Or something else?

experimental/ni_coq/README.md
```diff
@@ -18,10 +18,13 @@ is in vfiles/RuntimeModel.v. Both are evolving.
 
 ## Instructions
 
-The following are dependencies of this project: coq (8.11.0) coq-mathcomp-finmap
-(1.5.0)
+The following are dependencies of this project:
```
jieungkim, line 22: It will be better if we mention adding a coq package repo is desired to install Coq packages with opam: "opam repo add coq-released https://coq.inria.fr/opam/released"

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -12,6 +12,8 @@ Arguments Ensembles.In {U}.
 Arguments Ensembles.Add {U}.
 Arguments Ensembles.Subtract {U}.
 Arguments Ensembles.Singleton {U}.
+From RecordUpdate Require Import RecordSet.
```
jieungkim, line 15: Cool! :D I didn't know this one. Thanks for letting me know.

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -122,32 +147,45 @@ but that there is no premise checking that this call is really the one used
 in the relation. This is checked in the global transition relation just below.
 *)
 Inductive step_node: node_id -> call -> state -> state -> Prop :=
```
jieungkim, line 149: One (really weak) recommendation: 

It seems all parts in this rule do not change the first argument (node_id). One way to represent this one is 

Inductive step_node (id : node_id) : call -> state -> state -> Prop :=

To show that it is actually using the same node id as an explicit argument of this rule. But, it's really for the sake of readability.

jieungkim:
To Andrew: I hope someone else approve this change after reviewing channel reads/writes parts (that I am not familiar with), but let me know if you want me to approve this change :D

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -22,31 +35,46 @@ Open Scope fset_scope.
...
+(* etachannel just enumerates the fields of the record, as provided
+* by https://github.com/tchajed/coq-record-update *)
+(* When a new field is added to channel, be sure to add it here as well *)
+Instance etachannel : Settable _ := settable! Chan<clbl; ms; rhs; whs>.
+
 (* ABI Calls *)
 Inductive call: Type :=
-    | WriteChannel (h: handle)(m: message): call
+    | WriteChannel (h: handle)(m: message)
+            (rhs: Ensemble handle)(whs: Ensemble handle): call
```
tiziano88, line 52: This is not quite correct. Each message is itself made of bytes + handles. There are no handles associated with a channel directly. When a reader reads from a channel, it gets the first message in the queue, which means it sees the bytes that make the data of the message, and acquires access to the handles that are part of that message.

If it helps, this is very similar to capability-based systems with message passing, e.g. Fuchsia: https://fuchsia.dev/fuchsia-src/reference/syscalls/channel_write (this is a good description that is very close to what Oak does).

Let me know if you want to know more details about this.

aferr: This sounds a lot like the second option I listed above: it's essentially a stack of triples: (data * read_handles * write_handles).

A reader uses a read handle that it already has to access a channel. When it does, it gets the top message. Messages include data (the bytes) as well as some handles. When the read happens, the reader gets the data and all the handles in the message.

aferr: Or phrasing it another way:

A channel has a queue of messages. Messages are triples with: data, read handles, write handles.

(That's not what is implemented here now, but if you also think that makes sense, I can update it to that.)

experimental/ni_coq/README.md
aferr: Updated in latest commit.

experimental/ni_coq/vfiles/RuntimeModel.v
tiziano88: Yes it's essentially the same as your suggestion, but just to clarify, each message itself is made up of (bytes + handles), so I would avoid talking about (message + handles), or the terminology would not match.

For reference: 

https://github.com/project-oak/oak/blob/e174def0a34f1e975d1c0d9cd905cf05c856b4f4/oak_io/src/lib.rs#L34-L39

(feel free to include it as a comment, if it helps)

tiziano88: The other difference is that in this implementation, there is only one field for handles, both read and write, but I think it makes sense to keep them separate in your model, at least for now.

aferr: OK got it. That makes sense.

Re handles as one combined thing: 
I wasn't sure if splitting them was fine either. I was briefly trying to write it so that "handle" is a sumtype of read_handles and write_handles, but for some reason it didn't seem great. I might look at it again later as long as it's not too bad to keep them separate for now.

aferr: See latest commit. Let me know what you think.

experimental/ni_coq/vfiles/RuntimeModel.v
aferr: That is much better! Fixed in latest commit.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: It seems quite nice actually. I was not sure if I should use it at first since it seemed it may have made the proofs harder, but it doesn't. You can just do an "unfold" and get right back to the usual definitions for record update.

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -147,38 +161,51 @@ Definition nid_fresh (s: state)(nid: node_id): Prop :=
...
-        ch.(clbl) <<L n.(nlbl) ->
-        step_node id (ReadChannel han) s 
-            (state_upd_chan han (chan_pop ch) s)
-    | SCreateChan s id n h lbl:
+        (msg_is_head ch msg) ->
+        ch.(clbl) <<L n.(nlbl) -> (* label of caller flowsTo label of ch *)
+        let n' := node_get_hans n msg in
+            (* node gets handles from channel *)
+        let ch' := chan_pop ch in
+            (* pop the message, clear the read/write handles in ch *)
```
tiziano88, line 196: I find this a bit unnatural to read, in my mind it would read better if it was something like

```coq
let (msg, ch') := chan_pop ch in
  n' := node_get_hans n msg
...
```

or something along those lines. Would that make sense?

tiziano88 (approved):

experimental/ni_coq/vfiles/RuntimeModel.v
aferr: That looks elegant but I think that's tricky to do here because of the case where the channel is empty. Here the message is "retrieved from the channel" by the premise (msg_is_head ch msg). (And so this rule does not apply when the channel is empty).

It would be better to model the case where the channel is empty and an error is thrown, but that isn't done yet. (I might do this by adding another inference rule)

Issue #1246, opened 2020-07-07
Aggregator example should not create new nodes from existing nodes


tiziano88:
https://github.com/project-oak/oak/blob/dc23f27ad67f1707a8d92fb126fed932a80f4599/examples/aggregator/module/rust/src/lib.rs#L108-L111

Once we start labelling created nodes with more sensible labels, the main programming pattern will be for all nodes and channels to be created directly by the initial node of an application, while "worker" nodes just receive handles to pre-existing objects.

For the aggregator example, this means creating the gRPC client pseudo-node as part of the `oak_main` entrypoint:

https://github.com/project-oak/oak/blob/dc23f27ad67f1707a8d92fb126fed932a80f4599/examples/aggregator/module/rust/src/lib.rs#L166-L182

and then pass its handle to the gRPC worker.

The complication is that the gRPC worker currently only accepts gRPC invocation messages as input, so it needs to be modified so that it can accept a new type of message which is essentially an enum of either an "init" message (containing the initial handles and any other configuration), or the actual gRPC message to process.

ipetr0v:
I remember that I have reviewed a commit that was adding an "init" message, but I don't remember where it was, and have it been added yet.

ipetr0v:
Found it:
https://github.com/project-oak/oak/blob/7f33be4ff8f1a0949514643e36ad9f1a2253141c/sdk/rust/oak/src/grpc/server.rs#L21-L23

But it's a different "init" message.

ipetr0v:
Also, why should it be an enum of {Init, gRPC_invocation}?
Could it just read the first init message and then pass handle to the gRPC dispatcher?

tiziano88:
I don't think it's a good idea to rely on the order of the messages to figure out what they should do. We should use strongly typed messages instead, and each node should handle only one type of messages ever, rather than switch between different types.

So for this case, I think the implementation of the aggregator node should accept a `oneof` between an init message, and a gRPC invocation. If the gRPC invocation variant is received, then it should be dispatched using the code generated by prost.

cc @conradgrobler as we chatted about this earlier.

ipetr0v:
We are also loading `ConfigMap` the same way (as a first message in the initial handle):
https://github.com/project-oak/oak/blob/7b89fbdada697230672eae48b10fbe7904a33b2f/examples/trusted_information_retrieval/module_0/rust/src/lib.rs#L76-L79

tiziano88:
That handle is not used for anything else afterwards though

[...]

PR #1404, opened 2020-08-28
Use protos for Invocations

### LINKED ISSUE #1249: Replace `Invocation` struct with protobuf

conradgrobler:
This PR replaces Invocation structs in the SDK and Runtime with protos. It also enables using protobuf handle serialisation in the runtime.

Sumary of changes:

- Replace Invocation structs with protos
- Add GrpcInvocationSender and HttpInvocationSender protos for sending the Invocation channel during startup
- Update c++ modules to support GrpcInvocationSender
- Use updated versions of externally stored binaries for aggregator and trusted information retrieval examples

Fixes #1249

Unblocks #1246 

# Checklist

      construction.


oak_runtime/src/io.rs
```diff
@@ -22,13 +22,23 @@ use oak_abi::{ChannelReadStatus, OakStatus};
 use oak_io::{Decodable, Encodable, OakError};
 pub use oak_io::{Receiver, Sender};
 
+/// Wrapper for storing a decodable message and the size of the original encoded version.
+pub struct MessageWithSize<T> {
+    pub message: T,
+    pub size: usize,
+}
+
```
tiziano88, line 30: Why is this necessary? And what happens if the value of the size field does not match the actual size? AFAICT this is only used for metrics, and I feel the extra complexity is not warranted. We can look into the message and see its size there, if needed?

oak_runtime/src/node/grpc/invocation.rs
```diff
@@ -16,70 +16,54 @@
...
-            }),
-            sender: Sender::new(WriteHandle {
-                handle: message.handles[1],
-            }),
-        })
-    }
-}
-
 impl Invocation {
     pub fn close(self, runtime: &RuntimeProxy) {
```
tiziano88, line 30: Could you add a comment saying that this ignores any errors while closing the underlying channels?

oak_runtime/src/node/grpc/invocation.rs
```diff
@@ -16,70 +16,54 @@
...
+            }
+            None => error!("No sender on invocation."),
+        };
     }
+
     pub fn receive_request(&self, runtime: &RuntimeProxy) -> Result<GrpcRequest, OakError> {
-        self.receiver.receive(runtime)
+        match &self.receiver {
+            Some(receiver) => receiver.receive(runtime),
+            None => Err(OakError::OakStatus(OakStatus::ErrBadHandle)),
```
tiziano88, line 52: I thought this should be a similar error to above when no receiver or sender is present. Not exactly the same, but they should be related?

oak_runtime/src/node/grpc/server/mod.rs
```diff
@@ -574,7 +537,7 @@ impl Drop for MetricsRecorder {
 
 struct GrpcResponseIterator {
     runtime: RuntimeProxy,
-    response_reader: oak_abi::Handle,
+    response_reader: ReadHandle,
```
tiziano88, line 540: Should this field be of type `Receiver`? Or should it be renamed to `_handle`?

oak_runtime/src/io.rs
conradgrobler: This was to be able to use a Receiver rather than manually wait on channels, receive and decode the message to simplify the logic in the GrpcResponseIterator. I assumed there might be a need in future other cases might als use the size of the messages for monitoring, so thought it would be useful to make it possible to use a Receiver in these cases.

I am not sure what you mean by looking into the message and see its size, but I assume you mean go back to manually waiting on the channels and reading the raw message. Is that correct?

oak_runtime/src/node/grpc/invocation.rs
conradgrobler: The above function logs the error as it does not return a result, where this returns a result, so I thought it would be sufficient to propagate the error to be logged at a higher level. Do you suggest logging here and also returning the error or something else?

oak_runtime/src/node/grpc/server/mod.rs
conradgrobler: If I move back to waiting on channels and reading raw messages I will change this back to being a raw Handle and rename the field.

sdk/rust/oak/build.rs
```diff
@@ -17,12 +17,13 @@
...
     oak_utils::compile_protos_with_options(
         &[
+            "../../../oak_services/proto/grpc_invocation.proto",
+            "../../../oak_services/proto/http_invocation.proto",
             "../../../oak_services/proto/storage_service.proto",
             "../../../oak_services/proto/roughtime_service.proto",
         ],
         &["../../.."],
         oak_utils::ProtoOptions {
-            derive_handle_visit: false,
```
ipetr0v, line 25: Is `handle_visit` not necessary here? What does it do?

sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -243,8 +251,8 @@ where
     // Build an Invocation holding the two channels and send it down the
     // specified channel.
     let invocation = Invocation {
-        request_receiver: req_receiver.clone(),
-        response_sender: rsp_sender.clone(),
+        receiver: Some(req_receiver.clone()),
```
ipetr0v, line 254: Does `GrpcInvocationSender` check that both fields are not `None`? Since it will be an internal error

ipetr0v (approved):

oak_runtime/src/node/grpc/invocation.rs
```diff
@@ -16,70 +16,54 @@
...
-                handle: message.handles[0],
-            }),
-            sender: Sender::new(WriteHandle {
-                handle: message.handles[1],
-            }),
-        })
-    }
-}
-
 impl Invocation {
```
rbehjati, line 29: Could this and a similar implementation of HttpInvocation in `sdk/rust/oak/src/http/mod.rs` be unified into a generic implementation?

oak_io/src/handle.rs
```diff
@@ -187,12 +187,15 @@ macro_rules! handle_visit_blanket_impl {
 // See: https://github.com/danburkert/prost#scalar-values
 handle_visit_blanket_impl!((), f64, f32, i32, i64, u32, u64, bool, String, Vec<u8>);
 
-// Provide an implementation for the Any type in Prost.
-handle_visit_blanket_impl!(prost_types::Any);
+// Provide an implementation for the Any and Timestamp types in Prost.
+handle_visit_blanket_impl!(prost_types::Any, prost_types::Timestamp);
```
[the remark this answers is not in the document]
conradgrobler: Any types that want to implement Encodable or Decodable must implement HandleVisit. Any objects that will be sent using SenderExt must implement Encodable and any objects received using ReceiverExt must implement Decodable. To be able to send protos around they must all implement HandleVisit.

HandleVisit provides the ability to recursively visit all properties an a struct. This means that all properties that are on a struct that implements HandleVisit must in turn also implement HandleVisit. prost_types::Timestamp is used by the IntrospecitionEvents proto.

In cases where there are no handles there is nothing to do, but the trait must still be implemented. So the handle_visit_blanket_impl macro just implements a no-op to satisfy the type constraint.

sdk/rust/oak/build.rs
conradgrobler: By removing this line the HandleVisit trait will be automatically derived for the protos during compilation (the default is true). This is required for the new protos to be sent and received using Sender and Receiver.

oak_io/src/handle.rs
```diff
@@ -187,12 +187,15 @@ macro_rules! handle_visit_blanket_impl {
...
 
 // Provide an implementation for oak_abi type that an implementation cannot be derived for.
 // It does not contain handles, so a blanket impl is sufficient.
-handle_visit_blanket_impl!(oak_abi::proto::oak::application::ConfigMap);
+handle_visit_blanket_impl!(
```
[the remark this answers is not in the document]
conradgrobler: As noted in another comment in this reply this is a no-op to satisfy the type constraints, as the Label proto is used by the storage service protos.

oak_runtime/src/io.rs
conradgrobler: As discused separately, removed this and used the size of the rsp_msg field rather thatn the size of the entire GrpcResponse proto for monitoring.

oak_runtime/src/node/grpc/invocation.rs
conradgrobler: Added a comment.

@ipetr0v the ignoring of errors was a pre-existing pattern in the function. I assume the reasoning is that the primary reason for the errors would be that the channel was already closed previously, so logging the error should be sufficient.

oak_io/src/handle.rs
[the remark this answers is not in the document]
rbehjati: Thanks for the explanation.

sdk/rust/oak/src/grpc/mod.rs
conradgrobler: There is not a `send` function on GrpcInvocationSender. Its WriteHandle is extracted and used in `get_invocation_channel` and an Error is returned there if the `sender` is `None`.

oak_runtime/src/node/grpc/server/mod.rs
conradgrobler: Changed it to an `Option<Receiver>` (it must be an `Option` to support consuming `receiver` when calling `close` in `drop`)

oak_runtime/src/node/grpc/invocation.rs
conradgrobler: I don't think this can be done easily. This `GrpcInvocation` implementation is in the Runtime and the implementation of the `HttpInvocation` is in the SDK. The Runtime and SDK use different versions of SenderExt and ReceiverExt with different signatures for `send`, `receive` and `close` to support different mechanisms for interacting with handles depending on whether the code runs in a WASM node or not.

rbehjati: hmm! I did not realize that we have two different versions of `SenderExt` and `ReceiverExt `.

rbehjati (approved):

PR #1409, opened 2020-09-01
Ensure NodeCreated is fired for the initial node

juliettepretot:
Moves the `NodeCreated` introspection event from being fired in the `Runtime.node_create` method to the `Runtime.node_configure_instance` method. This is done since the first node (`implicit.initial`) is created directly in `RuntimeProxy.create_runtime` and does not go through `Runtime.node_create`.

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/lib.rs
+++ b/oak_runtime/src/lib.rs
@@ -1150,13 +1150,6 @@ impl Runtime {
             OakStatus::ErrInvalidArgs
         })?;
 
-        // TODO(#913): Add automated tests that verify that NodeCreated is
-        // always fired prior to any other introspection events related to the
-        // node.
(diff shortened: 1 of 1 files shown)


oak_runtime/src/lib.rs
```diff
@@ -1244,10 +1237,19 @@ impl Runtime {
...
-        RuntimeProxy {
+        let proxy = RuntimeProxy {
             runtime: self.clone(),
             node_id: self.new_node_id(),
-        }
+        };
+
+        // TODO(#913): Add automated tests that verify that NodeCreated is
+        // always fired prior to any other introspection events related to the
+        // node.
```
juliettepretot, line 1247: Another reason to add tests, so this doesn't break down the road. :)

daviddrysdale (approved):

Issue #1407, opened 2020-08-28
Re-implement TIR example


ipetr0v:
After a discussion in https://github.com/project-oak/oak/issues/1066 we need to re-implement Trusted Information Retrieval example in the following way:
- A separate TIR node should be created for each new client request (similar to https://github.com/project-oak/oak/issues/1246)
- `database_proxy` node should download the whole external database
    - We are removing a memory restriction and start assuming that the whole database could fit into Oak
- TIR node shouldn't send any requests to the `database_proxy` node, instead we need a `router` node that:
    - Will not be tainted with user data (it just forwards invocations)
    - Will request the database and pass it to the TIR node

The idea behind these changes is to remove the necessity of trusting any particular node in the application and just to assign a client token label to data.

cc @tiziano88 @conradgrobler

Issue #1406, opened 2020-08-28
Implement support in the SDK for finding the Label associated with a channel


conradgrobler:
To support a router pattern, where a node routes channels and invocations to specific nodes based on the associated labels, we need to provide a mechanism in the ABI and SDK for looking up the label associated with a channel.

ipetr0v:
> node routes channels and invocations to specific nodes based on the associated labels
Should it choose a destination node based on the incoming data label?
What if multiple nodes have corresponding labels?

conradgrobler:
The typical scenario where the router is needed is for per-user nodes. The router should know which user the data is intended for, and either create a new node with that label if it doesn't exist, or send the data to the right node if it does exist. At the moment the router node cannot tell which user the data is from. This is a prerequisite for #1407.

In cases where multiple nodes have the same label, the router would need to have some additional logic to make the decision (e.g. perhaps round-robin load balancing), but would still need to know which label is associated with the data.

ipetr0v:
I think such hidden logic may not be obvious for clients and application developers, since Oak will choose which node to set as a destination based on some additional logic.

I think, if we want to create per user nodes:
- We either might do something similar to [AWS Lambda](https://aws.amazon.com/lambda/), where nodes are created per each client request, and router doesn't need to route requests to a specific existing (per-client) node, it will just create a new one every time
- Or we can add a `node_id` as a part of the request, if we want to create a per-client node once and reuse it afterwards

conradgrobler:
If we move to a different model of node creation we can add logic to the gRPC server node to create new nodes, but with our current implementation all gRPC invocations are sent down a single channel. The node receiving this will have to create the per-user nodes and foward the invocations.

It most cases the node can just create a new node for every request and not worry about reusing nodes, but it still needs to know which label to set on the newly created node to maintain the right security.

ipetr0v:
> but it still needs to know which label to set on the newly created node to maintain the right security.

Why do we need to add a label to a newly created node?
If this node won't have a label, it will not be able to declassify the client data, and only be able to send it back to the original client. IIUC this is the intended behavior.

conradgrobler:
If the node has no label, there is nothing stopping it from creating another node (such as a gRPC client or log node) and sending the data outside of Oak.

conradgrobler:
[...]

PR #1423, opened 2020-09-02
Add Trusted Database example

ipetr0v:
This change adds a Trusted Database example.

Fixes https://github.com/project-oak/oak/issues/1407

# Checklist


examples/trusted_database/client/rust/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    )]
+    root_tls_certificate: String,
+    #[structopt(long, help = "Requested location's latitude in degrees (WGS84)")]
+    latitude: f32,
+    #[structopt(long, help = "Requested location's longitude in degrees (WGS84)")]
+    longitude: f32,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
tiziano88, line 57: Can we use `anyhow` for the return type, and for some of the error handling here?

examples/trusted_database/client/rust/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        .expect("Couldn't create TLS configuration")
+        .connect()
+        .await
+        .expect("Couldn't connect to Oak Application");
+
+    // TODO(#1097): Turn the following logic into a proper reusable client library.
+    let mut label = Vec::new();
+    Label::public_untrusted()
+        .encode(&mut label)
+        .expect("Error encoding label");
```
tiziano88, line 94: This will need to be changed to a per-user label and auth, right? Which at some point will rely on the challenge-response auth mechanism that @rbehjati is working on.

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Load an XML database from [`oak::ReadHandle`] and parse it.
+pub fn load_database(config_map: ConfigMap) -> Result<Vec<PointOfInterest>, OakError> {
+    debug!("Loading database");
+    match config_map.items.get("database") {
+        Some(xml_database) => {
+            let points_of_interest = parse_database(xml_database).map_err(|error| {
+                error!("Couldn't parse database: {:?}", error);
+                OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
+            })?;
+            debug!("Database loaded - size: {:?}", points_of_interest.len());
```
tiziano88, line 69:
```suggestion
            debug!("Database loaded - size: {} entries", points_of_interest.len());
```

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,113 @@
...
+        None => {
+            error!("`database` configuration argument is not specified");
+            Err(OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs))
+        }
+    }
+}
+
+/// Parse an XML database into a vector of [`PointOfInterest`].
+pub fn parse_database(xml_database: &[u8]) -> Result<Vec<PointOfInterest>, OakError> {
+    let database: Database = from_str(
```
tiziano88, line 81: I think the code would read better if `from_str` is fully qualified here instead of being imported.

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,113 @@
...
+                && station.removal_date.is_empty()
+                && station.number_of_bikes > 0
+        })
+        .map(|station| PointOfInterest {
+            name: station.name.to_string(),
+            location: Some(Location {
+                latitude: station.latitude_degrees,
+                longitude: station.longitude_degrees,
+            }),
+        })
```
tiziano88, line 110: Could this be defined as `From` or `To` between these types?

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,113 @@
...
+    )
+    .map_err(|error| {
+        error!("Couldn't parse XML data: {:?}", error);
+        OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
+    })?;
+
+    let points_of_interest = database
+        .stations
+        .iter()
+        // Filter out uninstalled, closed and removed stations, and stations with no bikes.
```
tiziano88, line 97: Why?

examples/trusted_database/module_0/rust/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+impl Node<grpc::Invocation> for TrustedDatabaseNode {
+    fn handle_command(&mut self, invocation: grpc::Invocation) -> Result<(), oak::OakError> {
+        // Create a client request handler Node.
+        debug!("Creating handler Node");
+        let (sender, receiver) = oak::io::channel_create::<TrustedDatabaseCommand>()
+            .expect("Couldn't create command channel");
+        oak::node_create(
+            &oak::node_config::wasm("trusted_database_handler", "handler_oak_main"),
+            receiver.handle,
+        )
```
tiziano88, line 53: Am I missing something, or is this not actually doing anything with respect to labels?

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+                    env!("OUT_DIR"),
+                    "/oak.examples.trusted_database_command.rs"
+                ));
+            }
+        }
+    }
+}
+
+// #[cfg(test)]
+// mod tests;
```
tiziano88, line 47: remove?

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+
+use log::{debug, error};
+use oak::{grpc, io::ReceiverExt};
+use proto::oak::examples::trusted_database::{
+    ListPointsOfInterestRequest, ListPointsOfInterestResponse, Location, PointOfInterest,
+    TrustedDatabase, TrustedDatabaseCommand, TrustedDatabaseDispatcher,
+};
+
+/// Oak Node that contains an in-memory database of Points Of Interest.
```
tiziano88, line 56: Is it intentional that the documentation is the same as the other node? That seems confusing, maybe add more details

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+pub struct TrustedDatabaseHandlerNode {
+    points_of_interest: Vec<PointOfInterest>,
+}
+
+impl TrustedDatabaseHandlerNode {
+    /// Returns a distance (in kilometers) between two locations using the Haversine formula
+    /// (ignoring height variations):
+    /// https://en.wikipedia.org/wiki/Haversine_formula
+    fn distance(first: Location, second: Location) -> f32 {
```
tiziano88, line 65: Not sure why this method is defined here and not as a top level fn.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+/// A gRPC service implementation for the Private Information Retrieval example.
+impl TrustedDatabase for TrustedDatabaseHandlerNode {
+    /// Find the nearest Point Of Interest based on linear search in the database.
+    fn list_points_of_interest(
+        &mut self,
+        request: ListPointsOfInterestRequest,
+    ) -> grpc::Result<ListPointsOfInterestResponse> {
+        debug!("Received request: {:?}", request);
+        let location = request.location.ok_or_else(|| {
+            let err = "Location is not specified".to_string();
```
tiziano88, line 97: const?

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        match nearest_point.0 {
+            Some(point) => {
+                debug!("Found the nearest Point Of Interest: {:?}", point);
+                Ok(ListPointsOfInterestResponse {
+                    point_of_interest: Some(point),
+                })
+            }
+            None => {
+                let err = "Empty database".to_string();
+                error!("Couldn't find the nearest Point Of Interest: {:?}", err);
```
tiziano88, line 122: Maybe it's clearer to actually build the error first, and then log it, so that you don't have to repeat the string, and also the code gets logged too.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+                Err(grpc::build_status(grpc::Code::InvalidArgument, &err))
```
conradgrobler, line 123: nit: if the database is empty returning an InvalidArgument error to the client is confusing. Something like an internal error would be more accurate, as there is nothing that the client can do (by changing the input argument) to fix this.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+
+oak::entrypoint!(handler_oak_main => |in_channel| {
+    oak::logger::init_default();
+
+    // Receive command.
+    let command_receiver = oak::io::Receiver::<TrustedDatabaseCommand>::new(in_channel);
+    command_receiver.wait().expect("Couldn't wait for command of channel");
```
conradgrobler, line 134: Is there a reason for using `wait` and `try_receive` separately rather than just using `receive` that does both?

rbehjati (commented):
I am trying to match the implementation here with the description in #1407. 
Is `TrustedDatabaseNode` the `database_proxy` node in #1407?
Is `TrustedDatabaseHandlerNode` the TIR node that is exposing gRPC endpoints to the clients?
The `router` node is created in line 50 of `examples/trusted_database/module_0/rust/src/lib.rs` per request. Is that correct? I expected `router` to be created by the TIR node (instead of the `database_proxy`) and have a different type than `TrustedDatabaseHandlerNode`.
It still feels like that the TIR node is sending the requests to `database_proxy`, as opposed to what is required by #1407. Or maybe I am missing something.

examples/trusted_database/module_0/rust/src/database.rs
ipetr0v: Since we are trying to find a nearest bike station that has bikes in it.

examples/trusted_database/client/rust/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Client for the Trusted Database example.
+
+pub mod proto {
+    tonic::include_proto!("oak.examples.trusted_database");
```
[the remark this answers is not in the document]
ipetr0v: IIUC `main.rs` and `lib.rs` are different compilation targets that don't intersect.
`lib.rs` is mainly needed for tests, so we will be able to use a `TrustedDatabaseClient`.

examples/trusted_database/client/rust/src/main.rs
ipetr0v: I think it should be changed in a different PR, since gRPC server declassification is not yet implemented

examples/trusted_database/module_0/rust/src/lib.rs
ipetr0v: I think since `module_0` only proxies invocation - it does not interact with labels

ipetr0v:
> I am trying to match the implementation here with the description in #1407.
Is TrustedDatabaseNode the database_proxy node in #1407?

It's a merged `database_proxy` and `router` Nodes.

> Is TrustedDatabaseHandlerNode the TIR node that is exposing gRPC endpoints to the clients?

Correct.

> The router node is created in line 50 of examples/trusted_database/module_0/rust/src/lib.rs per request. Is that correct? I expected router to be created by the TIR node (instead of the database_proxy) and have a different type than TrustedDatabaseHandlerNode.

The TIR (that was renamed to `Handler` here) is created per request. And the `Main` node (`TrustedDatabaseNode`) proxies client requests as well as the database to each newly created `Handler` Node.

> It still feels like that the TIR node is sending the requests to database_proxy, as opposed to what is required by #1407. Or maybe I am missing something.

It only reads proxied client requests and the database, and returns the response.
And since `TrustedDatabaseNode` only works with invocations - it doesn't see client messages (requests/responses).

examples/trusted_database/database.xml
```diff
@@ -0,0 +1 @@
...</lat><long>-0.129698963</long><installed>true</installed><locked>false</locked><installDate>1392039000000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>14</nbBikes><nbEmptyDocks>5</nbEmptyDocks><nbDocks>21</nbDocks></station><station><id>763</id><name>Mile End Park Leisure Centre, Mile End</name><terminalName>300027</terminalName><lat>51.52059714</lat><long>-0.032566533</long><installed>true</installed><locked>false</locked><installDate>1392043020000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>34</nbBikes><nbEmptyDocks>12</nbEmptyDocks><nbDocks>47</nbDocks></station><station><id>764</id><name>St. John's Road, Clapham Junction</name><terminalName>200185</terminalName><lat>51.46239255</lat><long>-0.16829214</long><installed>true</installed><locked>false</locked><installDate>1392633480000</installDate><removalDate></removalDate><temporary>false</temporary><nbBikes>29</nbBikes><nbEmptyDocks>4</nbEmptyDocks><nbDocks>33</nbDocks></station><station><id>765</id><name>Ranelagh Gardens, Fulham</name><terminalName>300064</terminalName><lat>51.46760141</lat><long>-0.20682737</long><installed>true</installed><locked>false</locked>
```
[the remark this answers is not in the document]
ipetr0v: It's just a downloaded database file. Probably need to format it, since it's very inconvenient to check if responses are correct.

examples/trusted_database/module_0/rust/src/database.rs
ipetr0v: It's not necessarily `From`/`To` since it uses only 3 fields from `Station`

tiziano88:
High level comment: why do we need two separate Rust crates (for the two modules) for the implementation of the Oak application?

examples/trusted_database/module_0/rust/src/lib.rs
tiziano88: Then it's not doing much in terms of Oak specific functionality, right? Or in other words, when we add labels from the client point of view, this logic will have to change to propagate them correctly, which could have been done now, but I guess we can postpone. But do you agree that if it doesn't do anything label related, then it's probably wrong? Or is the magic happening elsewhere?

ipetr0v:
> High level comment: why do we need two separate Rust crates (for the two modules) for the implementation of the Oak application?

So the example will show that in principal the implementation module could be any 3rd-party module.

examples/trusted_database/module_0/rust/src/lib.rs
conradgrobler: I don't think is possible to create a node with the right label until #1406 is resolved. Perhaps add a TODO to add the label logic here once #1406 is complete, and as part of adding more restrictive labels to the client. When the client label is more restrictive, the trusted_database_handler will not be able to read the invocation unless it has the same label.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+    let command_receiver = oak::io::Receiver::<TrustedDatabaseCommand>::new(in_channel);
+    let command: TrustedDatabaseCommand =
+        command_receiver.receive().expect("Couldn't receive command");
+    let receiver = command.invocation_receiver.expect("Couldn't receive gRPC invocation receiver");
+
+    // Run event loop and handle incoming invocations.
+    let node = TrustedDatabaseHandlerNode { points_of_interest: command.points_of_interest };
+    let dispatcher = TrustedDatabaseDispatcher::new(node);
+    let invocation_receiver = receiver.receiver.expect("Empty gRPC invocation receiver");
+    oak::run_event_loop(dispatcher, invocation_receiver);
```
conradgrobler, line 141: Should this run the event loop? If a new node is generated for every invocation, it should probably just read the one invocation, write the response and then terminate. Or ar least make the comments very clear that the event loop will only run once because the invocation channel is closed by the node that created it. Otherwise it might be confusing to people who read through the example.

examples/trusted_database/module_1/rust/src/tests.rs
```diff
@@ -0,0 +1,45 @@
...
+fn assert_approx_eq(left: f32, right: f32) {
+    if (left - right).abs() > 0.01 {
+        panic!("{} is not equal to {}", left, right);
+    }
+}
+
+#[test]
+fn test_distance() {
+    assert_approx_eq(get_distance((0.0, 0.0), (0.0, 0.0)), 0.0);
+    assert_approx_eq(get_distance((0.0, 0.0), (10.0, 10.0)), 1568.5204);
```
conradgrobler, line 41: nit: it seems a bit weird that the expected value goes to 4 decimal digits if the assertion only checks up to 2.

conradgrobler (approved):

examples/trusted_database/oak_app_manifest.toml
```diff
@@ -0,0 +1,5 @@
+name = "trusted_database"
+
+[modules]
+app = { path = "examples/trusted_database/bin/trusted_database.wasm" }
+trusted_database_handler = { path = "examples/trusted_database/bin/trusted_database_handler.wasm" }
```
rbehjati, line 5: I guess I don't quite understand the plumbing here. Do you still need this here? Is not it enough that `TrustedDatabaseNode` is creating `trusted_database_handler` nodes per request? 
The same applies to the integration_test for `module_0`. Do you need to build both wasm modules for the test?

examples/trusted_database/module_0/rust/tests/integration_test.rs
```diff
@@ -0,0 +1,172 @@
...
+            name: "Opened station 1".to_string(),
+            location: Some(Location {
+                latitude: 0.0,
+                longitude: 0.0,
+            }),
+        }),
+    };
+    assert_eq!(got, want);
+
+    let result = list_points_of_interest(&mut client, 9.0, -10.0).await;
```
rbehjati, line 157: Is this second request testing something in addition to the first request? If so, please add a comment to explain what it is testing. Otherwise, I guess one request is enough to show that the correct point of interest is retrieved.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+use proto::oak::examples::trusted_database::{
+    ListPointsOfInterestRequest, ListPointsOfInterestResponse, Location, PointOfInterest,
+    TrustedDatabase, TrustedDatabaseCommand, TrustedDatabaseDispatcher,
+};
+
+// Error messages.
+const NO_LOCATION_ERROR: &str = "Location is not specified";
+const EMPTY_DATABASE_ERROR: &str = "Empty database";
+
+/// Oak Node that contains a copy of the database .
```
rbehjati, line 57:
```suggestion
/// Oak Node that contains a copy of the database.
```
nit :P

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+    let central_angle = 2.0
+        * ((latitude_difference_radians / 2.0).sin().powi(2)
+            + (first_latitude_radians.cos()
+                * second_latitude_radians.cos()
+                * (longitude_difference_radians / 2.0).sin().powi(2)))
+        .sqrt()
+        .asin();
+
+    // Earth radius in kilometers.
+    let earth_radius = 6371.0_f32;
```
rbehjati, line 122: Could this be a const?

examples/trusted_database/module_0/rust/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        // Send a command massage to the handler Node.
+        debug!("Sending command message to handler Node");
+        sender
+            .send(&command)
+            .expect("Couldn't send command to handler Node");
+        oak::channel_close(sender.handle.handle).expect("Couldn't close sender channel");
+
+        // Send a gRPC invocation.
+        debug!("Sending gRPC invocation to handler Node");
```
rbehjati, line 85: IIUC the "handler node" in this log line is different from the "handler node" in the previous debug line. If so, it is better to use different names to be clear.

tiziano88:
> > High level comment: why do we need two separate Rust crates (for the two modules) for the implementation of the Oak application?
> 
> So the example will show that in principal the implementation module could be any 3rd-party module.

But the whole point here is that it does not require trust in any particular code, so the split into different crates is just an annoyance, no one will ever reuse either of the crates anyways, they will just write their own code, possibly reusing some of the functions, but not the Wasm module as a whole.

examples/trusted_database/client/rust/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+async fn main() -> anyhow::Result<()> {
+    env_logger::init();
+    let opt = Opt::from_args();
+
+    let uri = opt.uri.parse().context("Error parsing URI")?;
+    let root_tls_certificate = tokio::fs::read(&opt.root_tls_certificate)
+        .await
+        .context("Couldn't load certificate file")?;
+    let latitude = opt.latitude;
+    if latitude < -90.0 || latitude > 90.0 {
```
tiziano88, line 67: May use https://docs.rs/anyhow/1.0.32/anyhow/macro.ensure.html here and below

examples/trusted_database/client/rust/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+        .context("Couldn't connect to Oak Application")?;
+
+    // TODO(#1097): Turn the following logic into a proper reusable client library.
+    let mut label = Vec::new();
+    Label::public_untrusted()
+        .encode(&mut label)
+        .context("Error encoding label")?;
+    let mut client =
+        TrustedDatabaseClient::with_interceptor(channel, move |mut request: Request<()>| {
+            request.metadata_mut().insert_bin(
```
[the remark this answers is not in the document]
tiziano88: FWIW, I think eventually the labelling logic would benefit from being encapsulated in an interceptor anyways eventually, so this may actually be closer to what we expect in the final form. (similarly for authentication, which is not here)

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,115 @@
...
+//
+
+use log::{debug, error};
+use oak::OakError;
+use oak_abi::proto::oak::application::ConfigMap;
+use serde::Deserialize;
+use trusted_database_handler::proto::oak::examples::trusted_database::{Location, PointOfInterest};
+
+/// Database structure represents internal XML fields from the following database:
+/// https://tfl.gov.uk/tfl/syndication/feeds/cycle-hire/livecyclehireupdates.xml
```
tiziano88, line 24: Did you read the licensing terms? https://data.london.gov.uk/dataset/cycle-hire-availability

I think it means we need to link back to their license, right?

examples/trusted_database/module_0/rust/src/database.rs
```diff
@@ -0,0 +1,115 @@
...
+    temporary: bool,
+    #[serde(rename = "nbBikes", default)]
+    number_of_bikes: u32,
+    #[serde(rename = "nbEmptyDocks", default)]
+    number_of_empty_docks: u32,
+    #[serde(rename = "nbDocks", default)]
+    number_of_docks: u32,
+}
+
+/// Load an XML database from [`oak::ReadHandle`] and parse it.
```
tiziano88, line 59: Is this correct? It looks like it is reading it from the config map.

examples/trusted_database/module_0/rust/src/database.rs
tiziano88: Perhaps add it to the comment? I thought it was just returning the closest station.

examples/trusted_database/module_0/rust/src/lib.rs
tiziano88: You are right @conradgrobler , thanks!

examples/trusted_database/module_0/rust/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        invocation_sender
+            .send(&invocation)
+            .expect("Couldn't send gRPC invocation to handler Node");
```
tiziano88, line 88: Just checking, is it fine to have all these `expect`s in production code? I guess here it is fine since there is nothing that can be done to recover, but it may still be a transient issue that may be fine to handle by propagating the error up to the caller, or not?

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+impl TrustedDatabase for TrustedDatabaseHandlerNode {
+    /// Find the nearest Point Of Interest based on linear search in the database.
+    fn list_points_of_interest(
+        &mut self,
+        request: ListPointsOfInterestRequest,
+    ) -> grpc::Result<ListPointsOfInterestResponse> {
+        debug!("Received request: {:?}", request);
+        let location = request.location.ok_or_else(|| {
+            let err = grpc::build_status(grpc::Code::InvalidArgument, &NO_LOCATION_ERROR);
+            error!("{:?}", err);
```
tiziano88, line 72: `error` seems excessive for something that can be trivially triggered by a user request.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+            err
+        })?;
+        let nearest_point = self.points_of_interest.iter().fold(
+            (None, f32::MAX),
+            |(closest_point, min_distance), point| {
```
tiziano88, line 77: What's the difference between `nearest_point` and `closest_point`?

tiziano88, line 77: Maybe rename `min_distance` to `closest_point_distance` or similar? or rename `closest_point` to `min_point` (though this sounds wrong)

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+                let point_location = point.location.clone().expect("Non-existing location");
```
tiziano88, line 78: I guess it makes more sense to check for existence before cloning?

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+                let distance = distance(location.clone(), point_location);
```
tiziano88, line 79: I think renaming `location` to `request_location` or similar may make things more readable.

examples/trusted_database/module_1/rust/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+        &mut self,
+        request: ListPointsOfInterestRequest,
+    ) -> grpc::Result<ListPointsOfInterestResponse> {
+        debug!("Received request: {:?}", request);
+        let location = request.location.ok_or_else(|| {
+            let err = grpc::build_status(grpc::Code::InvalidArgument, &NO_LOCATION_ERROR);
+            error!("{:?}", err);
+            err
+        })?;
+        let nearest_point = self.points_of_interest.iter().fold(
```
tiziano88, line 75: Would it make sense to use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by or https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by_key here instead of `fold`?

examples/trusted_database/module_1/rust/src/tests.rs
tiziano88: Yeah 0.01 seems quite coarse for a f32 comparison, I would expect 0.000001 or similar would still be fine, for such small numbers?

examples/trusted_database/proto/trusted_database.proto
```diff
@@ -0,0 +1,60 @@
...
+
+syntax = "proto3";
+
+package oak.examples.trusted_database;
+
+import "google/protobuf/empty.proto";
+
+message Location {
+  float latitude = 1;
+  float longitude = 2;
```
tiziano88, line 25: radians? degrees? something else?

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: Done

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: Done

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: Done

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: Done

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: I used `run_event_loop` in order to use automatically generated `TrustedDatabaseDispatcher` and not implement `handler` by hand.
Added a comment.

examples/trusted_database/oak_app_manifest.toml
ipetr0v: `oak_app_manifest.toml` is used by `oak_loader` to load corresponding Wasm modules. So we still need to specify all possible modules here.

examples/trusted_database/module_0/rust/src/database.rs
ipetr0v: Thanks! I overlooked this license first. http://www.nationalarchives.gov.uk/doc/open-government-licence/version/2/
Since it says:
```
You must, where you do any of the above:
- acknowledge the source of the Information by including any attribution statement specified by the Information Provider(s) and, where possible, provide a link to this licence;
```
Does it mean that we shouldn't use this database, or we just need to add a LICENSE file?

examples/trusted_database/module_0/rust/tests/integration_test.rs
tiziano88: I agree with @ipetr0v , I think 2 examples is better here, to corroborate (although not prove, of course) the idea that at least it is not always returning the first entry or something like that.

examples/trusted_database/module_1/rust/src/lib.rs
tiziano88: I think you are already comparing pairs of elements anyways, so `min_by` seems fine? Anyway not a big deal.

examples/trusted_database/module_0/rust/tests/integration_test.rs
rbehjati: Makes sense. Would be nice to have a comment line to mention this :)

rbehjati (approved):

examples/trusted_database/module_0/rust/src/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+        <nbEmptyDocks>1</nbEmptyDocks>
+        <nbDocks>1</nbDocks>
+    </station>
+</stations>"#;
+
+#[test]
+fn test_parse_database() {
+    let database = parse_database(&XML_DATABASE.as_bytes().to_vec());
+    assert_matches!(database, Ok(_));
+    // Check that locked and uninstalled stations are filtered out.
```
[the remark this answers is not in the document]
ipetr0v: Deleted

examples/trusted_database/module_0/rust/tests/integration_test.rs
ipetr0v: Done

examples/trusted_database/module_1/rust/src/lib.rs
ipetr0v: Added to `third_party/data/cycle-hire-availability/LICENSE`

examples/trusted_database/database.xml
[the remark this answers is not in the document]
ipetr0v: Done
Formatted the database and moved it into `third_party`

PR #1425, opened 2020-09-02
Update docs about the use of labels with HTTP requests

rbehjati:
* Adds json schema files for IFC labels, generated using [protoc-gen-jsonschema](https://github.com/chrusty/protoc-gen-jsonschema)
* Updates documentation about the use of IFC labels with HTTP requests

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -322,6 +322,28 @@ represented by a message containing two Channels handles:
     connection itself; this represents the actual authority of the caller.
   - an empty integrity component.
 
+### HTTP and user labels
+
+Similar to gRPC requests, each incoming HTTP request must have a label
+specifying the confidentiality guarantees that the caller wants to impose on the
+request message. The label is either provided as a JSON formatted header named
+`oak-label` or a [protobuf message](/oak_abi/proto/label.proto) in a header
+named `oak-label-bin`. JSON formatted labels must conform to the
+[canonical encoding of Proto3 to JSON](https://developers.google.com/protocol-buffers/docs/proto3#json).
+
+Internally, each incoming HTTP invocation is represented by a message containing
+two Channel handles:
+
+- a "request receiver" read handle to a Channel whose label has a
+  confidentiality component explicitly set to the confidentiality component of
+  the label provided as a header (i.e., `oak-label` or `oak-label-bin`) in the
+  HTTP request. This represents the confidentiality guarantees that the caller
+  wants to impose on the request message.
+
+- a "response sender" write handle to a Channel. To avoid any accidental data
+  leaks, the label of the Channel is set to have a confidentiality component set
+  based on the identity of the user, and an empty integrity component.
+
 ### Signature labels
 
 Oak Wasm modules can be signed using [Ed25519](https://ed25519.cr.yp.to/)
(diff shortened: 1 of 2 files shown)


docs/concepts.md
```diff
@@ -322,6 +322,39 @@ represented by a message containing two Channels handles:
...
+    HTTP connection itself; this represents the actual authority of the caller.
+
+- a "response sender" write handle to a Channel, whose Label is currently
+  `public_untrusted`. However, to avoid any accidental data leaks, the label of
+  the Channel should be set such that it has:
+
+  - a confidentiality component set based on the identity of the user. This is
+    essentially the same as the integrity component described above for the
+    "request receiver".
+  - an empty integrity component.
```
juliettepretot, line 356: How are messages to this channel handled? Does it result in an outgoing HTTP invocation to the client that sent the original invocation?

rbehjati: Yes. The messages in the channel are only read by the HTTP server pseudo-node. The messages are in fact HTTP responses that will be sent to the client. But I will add a the confidentiality label to avoid any accidental data leaks. (This is a WIP description of how it should work, and it is not yet implemented, I have created #1428 to track this.)

rbehjati:
@tiziano88 @juliettepretot  I am not sure if it makes sense to check-in the `.jsonscheme` files in `oak_abi/json`. Proto3 provides a canonical JSON encoding, and Protobuf 3.3.0, comes with a built-in JSON serializer and parser. So, just having the proto file should be enough; no need for explicitly sharing a json schema. What do you think?

tiziano88:
> @tiziano88 @juliettepretot I am not sure if it makes sense to check-in the `.jsonscheme` files in `oak_abi/json`. Proto3 provides a canonical JSON encoding, and Protobuf 3.3.0, comes with a built-in JSON serializer and parser. So, just having the proto file should be enough; no need for explicitly sharing a json schema. What do you think?

How were these files genereted? How can other people re-generated? Is there any mechanism to make sure they are in sync? Personally I don't think they add any value on top of the protobuf files, but I don't know if they are useful for anything else?

rbehjati:
@tiziano88 I totally agree that they don't provide any value! I have generated them using [protoc-gen-jsonschema](https://github.com/chrusty/protoc-gen-jsonschema). Keeping them in sync will be a serious problem. I could not find a Bazel rule for generating them, and I am not sure we have a good place to document how these could be generated (currently only mentioned in this PR!).

juliettepretot:
Agreed we don't need these files if they can be deducted from protobuf as a source of truth. What are the JSON schema files used for? Just documentation?

rbehjati:
> Agreed we don't need these files if they can be deducted from protobuf as a source of truth. What are the JSON schema files used for? Just documentation?

Just documentation :)

juliettepretot:
Would it make sense to leave instructions in the docs to explain how people can deduct the JSON shape from the protobuf file?

rbehjati:
> Would it make sense to leave instructions in the docs to explain how people can deduct the JSON shape from the protobuf file?

Added a link to the description of the mapping.

juliettepretot (approved):
LGTM

PR #1426, opened 2020-09-02
Allow both JSON and protobuf labels in HTTP requests

rbehjati:
Ref [this comment](https://github.com/project-oak/oak/pull/1398#discussion_r478527779).

### DIFF
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -50,6 +50,7 @@ use futures_util::{
     future::TryFutureExt,
     stream::{StreamExt, TryStreamExt},
 };
+use prost::Message;
 use tokio_rustls::TlsAcceptor;
 
 #[cfg(test)]
@@ -480,11 +481,24 @@ impl Pipe {
     }
 }
 
+// HTTP requests can either provide JSON formatted labels or protobuf encoded labels. But exactly
+// one of these should be provided. This method checks that exactly one label is provided in a
+// header in the request and extracts it for use for further handling of the request.
 fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
-    match req.headers.get(oak_abi::OAK_LABEL_HTTP_KEY) {
-        Some(label_str) => parse_json_label(label_str.to_vec()),
-        None => {
-            warn!("No HTTP label found.");
+    let headers = (
+        req.headers.get(oak_abi::OAK_LABEL_HTTP_JSON_KEY),
+        req.headers.get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY),
+    );
+
+    match headers {
+        (Some(json_label), None) => parse_json_label(json_label.to_vec()),
+        (None, Some(protobuf_label)) => parse_protobuf_label(&protobuf_label[..]),
+        _ => {
+            warn!(
+                "Exactly one header must be provided as an {} or {} header.",
+                oak_abi::OAK_LABEL_HTTP_JSON_KEY,
+                oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY
+            );
             Err(OakStatus::ErrInvalidArgs)
         }
     }
@@ -499,6 +513,13 @@ fn parse_json_label(label_str: Vec<u8>) -> Result<Label, OakStatus> {
         OakStatus::ErrInvalidArgs
     })?;
     serde_json::from_str(&label_str).map_err(|err| {
+        warn!("Could not parse HTTP label: {}", err);
+        OakStatus::ErrInvalidArgs
+    })
+}
+
+fn parse_protobuf_label(protobuf_label: &[u8]) -> Result<Label, OakStatus> {
+    Label::decode(protobuf_label).map_err(|err| {
         warn!("Could not parse HTTP label: {}", err);
         OakStatus::ErrInvalidArgs
     })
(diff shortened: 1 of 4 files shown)


oak_runtime/src/node/http/tests.rs
```diff
@@ -203,9 +203,8 @@ fn oak_node_simulator(runtime: &RuntimeProxy, invocation_receiver: oak_abi::Hand
 async fn send_request(uri: &str) -> Result<http::response::Response<hyper::Body>, hyper::Error> {
     // Send a request, and wait for the response
     let label = oak_abi::label::Label::public_untrusted();
-    let label_bytes = serde_json::to_string(&label)
-        .expect("Could not serialize the label to JSON")
-        .into_bytes();
+    let mut label_bytes = vec![];
+    let _ = label.encode(&mut label_bytes);
```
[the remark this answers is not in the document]
juliettepretot: Eg in this case simply `label.encode(&mut label_bytes);`

oak_runtime/src/node/http/mod.rs
```diff
@@ -481,12 +482,31 @@ impl Pipe {
...
+    } else if let Some(label_bytes) = protobuf_label {
+        Label::decode(&label_bytes[..]).map_err(|err| {
+            warn!("Could not parse HTTP label: {}", err);
+            OakStatus::ErrInvalidArgs
+        })
+    } else {
+        // Should never arrive here
+        warn!("No HTTP label found.");
+        Err(OakStatus::ErrInvalidArgs)
     }
```
juliettepretot, line 510: I'm still new to Rust, but it seems to me this could be restructured into a match expression? 

```rust
fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
    let headers = (
        req.headers.get(oak_abi::OAK_LABEL_HTTP_JSON_KEY),
        req.headers.get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY),
    );

    match headers {
        (Some(json_label), None) => parse_json_label(json_label.to_vec()),
        (None, Some(protobuf_label)) => Label::decode(&protobuf_label[..]).map_err(|err| {
            warn!("Could not parse HTTP label: {}", err);
            OakStatus::ErrInvalidArgs
        }),
        _ => {
            error!(
                "Exactly one header must be provided as an {} or {} header.",
                oak_abi::OAK_LABEL_HTTP_JSON_KEY,
                oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY
            );
            Err(OakStatus::ErrInvalidArgs)
        }
    }
}
```

oak_runtime/src/node/http/tests.rs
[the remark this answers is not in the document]
juliettepretot: Had a look by cloning this branch and it seems the reason is the `rustc(unused_must_use)` warning. 

Using `let _` seems like bypassing it to me, but if it's common thing who am I to judge. :) Seems it doesn't actually handle a potential error though.

image

oak_runtime/src/node/http/mod.rs
rbehjati: I really like this. Thanks :)

oak_runtime/src/node/http/tests.rs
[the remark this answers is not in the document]
rbehjati: I added `let _ =` only to avoid the warning. I have added error handling code now.

oak_runtime/src/node/http/mod.rs
```diff
@@ -480,11 +481,27 @@ impl Pipe {
...
-            warn!("No HTTP label found.");
+    let headers = (
+        req.headers.get(oak_abi::OAK_LABEL_HTTP_JSON_KEY),
+        req.headers.get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY),
+    );
+
+    match headers {
+        (Some(json_label), None) => parse_json_label(json_label.to_vec()),
+        (None, Some(protobuf_label)) => Label::decode(&protobuf_label[..]).map_err(|err| {
+            warn!("Could not parse HTTP label: {}", err);
```
tiziano88, line 496: Is there any specific reason why an error decoding from protobuf produces a `warn`, but not from JSON?

oak_runtime/src/node/http/mod.rs
```diff
@@ -480,11 +481,27 @@ impl Pipe {
...
+            OakStatus::ErrInvalidArgs
+        }),
+        _ => {
+            error!(
```
tiziano88, line 500: This should probably be `warn` too?

oak_runtime/src/node/http/mod.rs
```diff
@@ -480,11 +481,27 @@ impl Pipe {
     }
 }
 
+// HTTP requests can either provide JSON formatted labels or protobuf encoded labels. But exactly
+// one of these should be provided. This method checks that exactly one label is provided in a
+// header in the request and extracts it for use for further handling of the request.
 fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
```
tiziano88, line 487: Is this in sync with the logic from the gRPC server pseudo-node? if not, should it be?

tiziano88 (approved):

oak_runtime/src/node/http/mod.rs
juliettepretot: Rust's pattern matching is quite powerful, I really miss it when writing code in other languages now.

juliettepretot (approved):
LGTM

oak_runtime/src/node/http/mod.rs
tiziano88: Yes that's what I mean, also I didn't realise that the gRPC metadata ends up with a different header name, my main point about this logic was that I thought HTTP and gRPC would use the same header name, and therefore some disambiguation is necessary (especially if we plan to eventually merge the two node types together)

rbehjati: I think we need to discuss this further.

PR #1427, opened 2020-09-03
Add basic TypeScript introspection client

juliettepretot:
Broke the introspection client code out of #1382, so I can continue to iterate on it with small PRs (instead of one massive one down the road) while #1382 (the Rust code to bundle & serve these files) is still pending review. 

# Checklist

      construction.


oak_runtime/src/introspection_browser_client/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
conradgrobler, line 1: Is the usage of this script documented anywhere? Also, in spite of reading the discussion on #1382 I am not clear on how this script will interoperate with the build system or whether this should be here or in the scripts directory.

conradgrobler (commented):
It feels a bit weird to have all these files under `src/` which is usually mostly `.rs` files. Is it possible to move it elsewhere and copy it during the build as part of the build.rs to have cleaner separation?

oak_runtime/src/introspection_browser_client/build.sh
juliettepretot: This script is used by NPM for the `npm run build` command, used to build the client. Commonly these are inlined in `package.json` (see the `dev` script), but since this one is a bit more complex it felt sensible to move it to its own file. 

Personally I'd argue against moving it into the scripts directory, since it isn't meant to be called directly by developers. 

I'd argue the integration of it into the build system is part of #1382, where it is called in the runner.

juliettepretot:
@conradgrobler RE the location of this code, yup, that's debate-able. If it's part of the runtime crate, as the consensus in #1382 appears to be, it seems sensible to place in the source of that crate. Happy to move this at a later point as well.

oak_runtime/src/introspection_browser_client/build.sh
rbehjati: I think `oak_runtime/src` is where we have all the rust code. I'd suggest `introspection_browser_client` to be moved to its own top-level folder. Is that possible?

oak_runtime/src/introspection_browser_client/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import introspectionEventsProto from './proto/introspection_events_pb';
+
+function loadSerializedEvents(): Promise<Uint8Array> {
```
rbehjati, line 21: Perhaps add a comment explaining what it does. In particular, it might be nice to give a bit of context about what the `Events` are.

oak_runtime/src/introspection_browser_client/build.sh
tiziano88: I agree that it feels odd to have non-rust code inside the `src` directory of a Rust crate. If possible, another directory at the same level as `src` seems more appropriate?

tiziano88 (approved):

oak_runtime/src/introspection_browser_client/build.sh
juliettepretot: Makes sense. :) I moved the folder up to live at the same level as `src`. I do think it's sensible to keep these files inside of the crate, since they are bundled by it.

https://github.com/project-oak/oak/blob/410c648a17de46838ff3e8a990f7cc0262677b74/oak_runtime/src/introspect.rs#L115

oak_runtime/introspection_browser_client/build
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#
+# Build browser introspection client.
+
+# Install npm modules
+npm ci
+
+# Generate JavaScript code from the introspection_events proto
+mkdir -p proto
+PROTOC_GEN_TS_PATH="./node_modules/.bin/protoc-gen-ts"
```
tiziano88, line 10:
```suggestion
readonly PROTOC_GEN_TS_PATH='./node_modules/.bin/protoc-gen-ts'
```

tiziano88 (approved):

oak_runtime/introspection_browser_client/build
juliettepretot: Good catch )

PR #1429, opened 2020-09-03
Possibilistic single-step for system transition

aferr:
This commit includes a proof of possibilistic_ni_1step in
PossibilisticNI.v, which is the unwinding condition. However, it still
depends on several theorems are still admitted, but should be true. The
definition for havocking calls for nodes that have just executed has
also been fixed in the system-wide state transition relation.
Previously, the call havocking part of the definition assumed that the
contents of the node have not changed as a result of the node's
execution, but clearly that is wrong. Now it properly updates the
(possibly changed) node that is pointed to by the node id that was
executed.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -7,6 +7,9 @@ From OakIFC Require Import
     Events
     LowEquivalences
     TraceTheorems.
+From mathcomp Require Import all_ssreflect finmap.
+From RecordUpdate Require Import RecordSet.
+Import RecordSetNotations.
 
 
 (*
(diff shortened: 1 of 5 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,84 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+        destruct t2_sn; subst.
+        - exfalso. apply (node_no_steps_to_empty t2 id (ncall n2)). assumption.
+        - exists (fst p). simpl. destruct p. reflexivity.
+    }
+    destruct Hs2' as [s2' Hs2']. 
+    remember (head_set_call t2_sn id c') as t2'.
+    exists t2'. split.
+    + subst. (* can step from t2_sn to t2' *)
+        apply (ValidStep id n2 (ncall n2) c' s2 t2 s2' t2_sn);
+            (try assumption; try reflexivity).
```
jadephilipoom, line 140: If this is expected to solve the goal, you could do `(assumption || reflexivity)` to try `assumption` first and fall back to `reflexivity` if it fails.

experimental/ni_coq/vfiles/TraceTheorems.v
```diff
@@ -6,25 +6,51 @@ From OakIFC Require Import
...
-        ((let (_, e) := p in (s, e)) :: t).
+    intros; destruct t; destruct t'.
+    - reflexivity.
+    - inversion H.
+    - inversion H. destruct p. inversion H1.
+    - inversion H. destruct p. inversion H1. subst. reflexivity.
+Qed. 
+
+Theorem head_set_call_not_nil: forall t id c,
+    ~(t = nil) -> ~((head_set_call t id c) = nil).
```
jadephilipoom, line 26: May be more readable using the `<>` sign, as in `t <> nil -> head_set_call t id c <> nil`?

experimental/ni_coq/vfiles/TraceTheorems.v
```diff
@@ -6,25 +6,51 @@ From OakIFC Require Import
...
+    intros. unfold head_set_call. unfold s_set_call.
+    destruct t. assumption.
+    destruct p. destruct (finmap.fnd (finmap.ffun_of_fmap (nodes s)) id);
     apply nil_cons_rev.
-    destruct p. reflexivity.
+Qed.
+
+Lemma head_set_call_preserves_tail: forall t t' a id c,
+    head_set_call t id c = a :: t' ->
+    exists a', t = a' :: t'.
```
jadephilipoom, line 36: I'm not sure which form is more convenient for your use-case, but you could also phrase this as `tl (head_set_call t id c) = tl t`, to avoid introducing extra variables for the first list elements.

jadephilipoom (approved):
LGTM. I added a few comments with possible rephrasings.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,84 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
     (step_system_ev t1 t1') ->
     (exists t2',
         (step_system_ev t2 t2') /\
         (trace_low_eq ell t1' t2')).
-Admitted. (* NOTE: work in progress *)
+Proof.
+    intros.
+    inversion H0; subst.
+    rename t' into t1_sn. rename s into s1. rename s' into s1'. rename n into n1.
+    assert (Ht2call: exists s2 n2,
```
blaxill, line 103: The proof might be easier to follow with the `assert`s as separate lemmas.

blaxill (approved):

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Ah nice to know! I am using this now.

experimental/ni_coq/vfiles/TraceTheorems.v
aferr: You're right. That's better. I'm doing that in the latest commit.

experimental/ni_coq/vfiles/TraceTheorems.v
aferr: I like your way better. I'm using it in the latest commit.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: I buy that. Working on it.

aferr: See latest commit. I split the longest one into a separate lemma. I changed a lot of the asserts that just apply a lemma into "specializes" which I only just learned about. I think specialize has also improved readability. I couldn't think of a good way to convert one of the asserts into a separate lemma, so I left it but I also added a comment near it.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,100 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+    (exists s2 n2,
+        head_st t2 = Some s2 /\ 
+        (nodes s2).[? id] = Some n2 ).
+Proof.
+    intros.
+    destruct (head_st t2) eqn:Ht2head.
+    - (* some s *)
+        destruct (nodes s).[? id] eqn:Hsid.
+            + (* some n *)
+                exists s. exists n.
```
jieungkim, line 114: exists s, n.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,100 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+                split; (reflexivity || assumption).
+            + (* none *)
+                inversion H; subst.
```
jieungkim, line 117: You will need this pattern in many places. So, making your own inversion tactic is quite a good strategy. The simplest one will be:

Ltac inv H := inversion H; subst.

and, we can extend this user-defined inversion when it is desired.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,100 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+                * (* t1 = [] and t2 = [] *) discriminate H0.
+                * exfalso. inversion H4; subst. simpl in Ht2head.
+                 unfold node_state_low_eq in H5.
+                 inversion Ht2head. rewrite H8 in H5. specialize (H5 id).
+                 rewrite Hsid in H5. simpl in H0. inversion H0.
+                 rewrite H9 in H5. rewrite H1 in H5. assumption.
+    - (* none *)
+        inversion H; subst.
```
jieungkim, line 125: inversion H; subst; simpl in *; discriminate.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -66,16 +69,100 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+                 inversion Ht2head. rewrite H8 in H5. specialize (H5 id).
+                 rewrite Hsid in H5. simpl in H0. inversion H0.
+                 rewrite H9 in H5. rewrite H1 in H5. assumption.
+    - (* none *)
+        inversion H; subst. 
+            + discriminate H0.
+            + inversion Ht2head.
+Qed.
+ 
 Theorem possibilistic_ni_1step: forall ell t1 t2 t1',
```
jieungkim, line 130: One good way to make more meaningful hypothesis names is naming premises in the theorem definition.

e.g.

Theorem possibilistic_ni_1step: forall ell t1 t2 t1'
    (Ht_eq: trace_low_eq ell t1 t2)
    (Hstep_ev: step_system_ev t1 t1'),
    (exists t2',
        (step_system_ev t2 t2') /\
        (trace_low_eq ell t1' t2')).

Then, when you [intros] all hypotheses, user-defined names will be used as names for premises. It could be a good way when extending and managing the proof.

PR #1432, opened 2020-09-03
Introspection client: Get state from events

juliettepretot:
Implements the logic to deduct the state of an Oak application from a list of events describing state mutations. "Time traveling" to view a past state then becomes as easy as calculating the state from a subset of the present event list.

The UI currently isn't very pretty and shows only the current state. I'll be working with a designer to create a more beautiful and useful version.

image

# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/index.tsx
+++ b/oak_runtime/introspection_browser_client/index.tsx
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
   });
 }
 
+interface OakApplicationState {
+  nodeInfos: NodeInfos;
+  channels: Channels;
+}
+
+type NodeId = number;
+type AbiHandle = number;
+type NodeInfos = Map<NodeId, NodeInfo>;
+interface NodeInfo {
+  abiHandles: Map<AbiHandle, ChannelHalf>;
+}
+
+type ChannelID = number;
+enum ChannelHalfDirection {
+  Read,
+  Write,
+}
+interface ChannelHalf {
+  channelId: ChannelID;
+  direction: ChannelHalfDirection;
+}
+interface Message {
+  data: Uint8Array;
+  channels: ChannelHalf[];
+}
+interface Channel {
+  id: ChannelID;
+  messages: Message[];
+}
+type Channels = Map<ChannelID, Channel>;
+
+function eventReducer(
+  applicationState: OakApplicationState,
+  event: introspectionEventsProto.Event
+) {
+  const eventType = event.getEventDetailsCase();
+  const { EventDetailsCase } = introspectionEventsProto.Event;
+
+  switch (eventType) {
+    case EventDetailsCase.NODE_CREATED:
+      applicationState.nodeInfos.set(event.getNodeCreated().getNodeId(), {
+        abiHandles: new Map(),
+      });
+
+      break;
+    case EventDetailsCase.NODE_DESTROYED:
+      applicationState.nodeInfos.delete(event.getNodeDestroyed().getNodeId());
+
+      break;
+    case EventDetailsCase.CHANNEL_CREATED:
+      {
+        const channelId = event.getChannelCreated().getChannelId();
+        applicationState.channels.set(channelId, {
+          id: channelId,
+          messages: [],
+        });
+      }
+
+      break;
+    case EventDetailsCase.CHANNEL_DESTROYED:
+      applicationState.channels.delete(
+        event.getChannelDestroyed().getChannelId()
+      );
+
+      break;
+    case EventDetailsCase.HANDLE_CREATED:
+      {
+        const details = event.getHandleCreated();
+        applicationState.nodeInfos
+          .get(details.getNodeId())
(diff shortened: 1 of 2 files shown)


oak_runtime/introspection_browser_client/index.tsx
```diff
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
...
+}
+interface Message {
+  data: Uint8Array;
+  channels: ChannelHalf[];
+}
+interface Channel {
+  id: ChannelID;
+  messages: Message[];
+}
+type Channels = Map<ChannelID, Channel>;
```
juliettepretot, line 77: These data structures describing the `OakApplicationState` largely mirror the ones used in the actual runtime. The only difference is channels being centrally stored as a top level property, instead of being referenced by handles. 

This due to differences between TypeScript & Rust, the former of which has no true `Arc` equivalent.

oak_runtime/introspection_browser_client/index.tsx
```diff
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
...
+    case EventDetailsCase.MESSAGE_ENQUEUED:
+      // TODO(#913): Add support for displaying messages
+      break;
+    case EventDetailsCase.MESSAGE_DEQUEUED:
+      // TODO(#913): Add support for displaying messages
+      break;
+    default:
+      // This should never happen
+      throw new Error(`Encountered unhandled event of type ${eventType}`);
+  }
```
juliettepretot, line 141: This `switch` statement can basically be thought of as a poor man's version of pattern matching, implementing different logic for each event type. 

This would be a lot nicer if written Rust, but for now I felt that the overhead introduced by Rust/JS interop does not justify writing this logic in Rust, even if it would be nicer.

oak_runtime/introspection_browser_client/index.tsx
```diff
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
...
+        applicationState.nodeInfos
+          .get(details.getNodeId())
+          .abiHandles.delete(details.getHandle());
+      }
+      break;
+    case EventDetailsCase.MESSAGE_ENQUEUED:
+      // TODO(#913): Add support for displaying messages
+      break;
+    case EventDetailsCase.MESSAGE_DEQUEUED:
+      // TODO(#913): Add support for displaying messages
```
conradgrobler, line 136: At the moment it is not clear to me how the state, including where channel-falves are stored will be fully reconconstructed. Will this event include a list of channel halves that was passed in the message?

conradgrobler (approved):

oak_runtime/introspection_browser_client/index.tsx
juliettepretot: Yup! See the associated proto of this event: 

https://github.com/project-oak/oak/blob/c3d9e43aeab28095643788d1e2e91f984d31ddfe/proto/introspection_events.proto#L82

oak_runtime/introspection_browser_client/index.tsx
```diff
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
...
+      );
+
+      break;
+    case EventDetailsCase.HANDLE_CREATED:
+      {
+        const details = event.getHandleCreated();
+        applicationState.nodeInfos
+          .get(details.getNodeId())
+          .abiHandles.set(details.getHandle(), {
+            channelId: details.getChannelId(),
```
rbehjati, line 119: Does it require a check to make sure the `channelId` exists in `applicationState.channels`?

oak_runtime/introspection_browser_client/index.tsx
```diff
@@ -45,6 +45,104 @@ function loadSerializedEvents(): Promise<Uint8Array> {
...
+            direction: 0
```
rbehjati, line 120: Is it possible to use `read` or `write` here? Also is it always `read` (0)?

oak_runtime/introspection_browser_client/index.tsx
juliettepretot: If introspection events are fired correctly the `channelId ` should always exist. In case it doesn't, a runtime error will occur, along the lines of `"Cannot read property 'abiHandles' of undefined"`. 

I could add a check here to include a more descriptive error message, though I'm not sure how much value it'd add. :)

juliettepretot: Turns out that I can enable a setting TypeScript so this does require a check. (`strictNullChecks`). Added a commit to enable that & add null checks with proper error messages, similar to how it is done in Rust.

rbehjati: Nice. I was kind of thinking about how `nodeInfos` and `channels` in `OakApplicationState` remain in sync. If that is at all necessary. I guess, as you mentioned, if the events are fired correctly, `nodeInfos` and `channels` will be synced.

rbehjati (approved):

PR #1438, opened 2020-09-07
Exclude protoc generated files from formatting

juliettepretot:
Patches an issue when running `check-format` locally, in which JavaScript & TypeScript files generated by protobuf compiler are flagged. 

Thanks to @ebeasant-arm for catching. :) 

# Checklist

      construction.


tiziano88 (approved):
How come this was not caught in CI though?

juliettepretot:
@tiziano88 These files are created when building the runtime via the runner. In CI, the runner runs `check-format` first, hence these files do not yet exist. 

If we wanted to catch cases like this we could consider moving the `check-format` check down.

PR #1442, opened 2020-09-08
Update prettier to version 2

juliettepretot:
See: https://prettier.io/blog/2020/03/21/2.0.0.html

Matches the major version of the Prettier VSCode plugin, thereby avoiding formatting inconsistencies between that & our format script.

# Checklist

      construction.

### DIFF
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -262,6 +262,7 @@ name = "hello_world"
 [modules]
 app = { path = "examples/hello_world/bin/hello_world.wasm" }
 translator = { path = "examples/hello_world/bin/translator.wasm" }
+
 ```
 
 All these steps are implemented as a part of the
@@ -278,6 +279,7 @@ name = "hello_world"
 [modules]
 app = { external = { url = "https://storage.googleapis.com/oak-modules/hello_world/57ba0bcebf2c01389d0413736b0a3bb261312bcf0d6e87181359e402df751d50", sha256 = "57ba0bcebf2c01389d0413736b0a3bb261312bcf0d6e87181359e402df751d50" } }
 translator = { path = "examples/hello_world/bin/translator.wasm" }
+
 ```
 
 It is also possible to save compiled Wasm modules to Google Cloud Storage using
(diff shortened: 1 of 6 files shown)


docs/programming-oak.md
```diff
@@ -262,6 +262,7 @@ name = "hello_world"
 [modules]
 app = { path = "examples/hello_world/bin/hello_world.wasm" }
 translator = { path = "examples/hello_world/bin/translator.wasm" }
+
```
tiziano88, line 265: These blank lines seem odd, are they added by prettier? Why only here?

tiziano88 (approved):

docs/programming-oak.md
juliettepretot: It's the TOML plugin we use, it always adds an empty line at the end of toml files. Previously it would skip it for toml excerpts in markdown files, but it seems that's no longer the case.

tiziano88: Which TOML plugin do we use? :eyes:

juliettepretot: It's a bit strange, but unfortunately it seems that plugin is no longer maintained. Imho admittedly strange formatting, but probably not worth staying on the previous prettier version to avoid this

tiziano88: do you mean that prettier is also formatting code inside code blocks if it has a plugin for that particular file format?

ipetr0v (approved):

docs/programming-oak.md
tiziano88: but also, the version of the prettier-toml plugin has not changed, so I'm even more confused

juliettepretot: [As of 2.1 prettier does apply formatting to code blocks it recognizes](https://prettier.io/blog/2020/08/24/2.1.0.html#add---embedded-language-formattingautooff-option-7875httpsgithubcomprettierprettierpull7875-by-bakkothttpsgithubcombakkot-8825httpsgithubcomprettierprettierpull8825-by-fiskerhttpsgithubcomfiskers). Hence previously the plugin was not applied to the code blocks, but since prettier was updated it now is.

juliettepretot: We can turn off this option (formatting code-blocks) if desired.

tiziano88: Ah cool, thanks for the link, that's interesting! Happy to keep as it is

tiziano88:
Please make sure to push the latest version of the Docker image once this is merged

Issue #1430, opened 2020-09-03
HTTP server pseudo-node crashes when accessed from Chrome


tiziano88:
To reproduce:

- run the following command from within `./scripts/docker_sh`

  ```bash
  ./scripts/runner --logs run-examples --example-name=http_server --client-variant=none
  ```

- open a Chrome window
- connect to `https://localhost:8080` via the URL bar

The logs of the application show the following error:

```
[2020-09-03T19:02:42Z INFO  oak_runtime::node::logger] Starting HTTP server pseudo-node on port 8080.
[2020-09-03T19:02:42Z INFO  oak_runtime] NodeId(1): start node instance NodeId(3) with privilege NodePrivilege { can_declassify_confidentiality_tags: {}, can_endorse_integrity_tags: {} }
[2020-09-03T19:02:42Z INFO  oak_runtime::node::http] http_server: Waiting for invocation channel
[2020-09-03T19:02:42Z INFO  oak_runtime::node::http] Invocation channel write handle received: 9390602833236434704
[2020-09-03T19:02:42Z INFO  oak_runtime::node::http] http_server: Starting HTTP server pseudo-Node on: [::]:8080
[2020-09-03T19:02:42Z INFO  oak_runtime::node::logger] Receiver(ReadHandle(6869080766937487096)): starting event loop
[2020-09-03T19:02:42Z INFO  oak_runtime::node::http] ThreadId(6): Started HTTP server pseudo-node on port 8080
[2020-09-03T19:02:46Z ERROR rustls::session] TLS alert received: Message {
        typ: Alert,
        version: TLSv1_3,
        payload: Alert(
            AlertMessagePayload {
                level: Fatal,
                description: CertificateUnknown,
            },
        ),
    }
[2020-09-03T19:02:46Z ERROR oak_runtime::node::http] Client-connection error: Custom { kind: InvalidData, error: AlertReceived(CertificateUnknown) }
[2020-09-03T19:02:46Z ERROR oak_runtime::node::http] server error: error accepting connection: TLS Error: Custom { kind: InvalidData, error: AlertReceived(CertificateUnknown) }
[2020-09-03T19:02:46Z INFO  oak_runtime::node::http] HTTP server pseudo-node terminated with Ok(())
```

There are two issues here:

- not sure why it is failing, perhaps some TLS configuration
- even if it does fail, it should not terminate the entire server node, it should keep going and try to handle future requests

rbehjati:
[...]

PR #1443, opened 2020-09-08
Fix early termination of HTTP server node

rbehjati:
Ref #1430 

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -78,7 +78,22 @@ impl hyper::server::accept::Accept for TlsServer<'_> {
         mut self: Pin<&mut Self>,
         cx: &mut Context,
     ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
-        Pin::new(&mut self.acceptor).poll_next(cx)
+        let connection = Pin::new(&mut self.acceptor).poll_next(cx);
+        match connection {
+            // When a connections fails, for example due to a `CertificateUnknown` error, we want to
+            // keep the server alive and waiting for new connections. The `poll_accept` method has
+            // to return `Poll::Pending` in these cases, to in turn cause the `poll` method of the
+            // hyper server to return `Poll::Pending`, making the server wait for the next
+            // connection. The `poll_watch` method in `hyper::server::conn::SpawnAll`,
+            // which calls `poll_accept` will make sure that the server is scheduled to
+            // be awoken when there is a new incoming connection, by calling `wake` on
+            // the `waker` in the context `cx`.
+            Poll::Ready(Some(Err(e))) => {
+                log::error!("Error when processing TLS stream: {:?}", e);
+                Poll::Pending
+            }
+            _ => connection,
+        }
     }
 }
 
@@ -130,7 +145,7 @@ impl HttpServerNode {
         let mut tcp = TcpListener::bind(&self.address)
             .await
             .expect("Could not create TCP listener.");
-        let tls_server = self.build_tls_server(&mut tcp).await;
+        let tls_server = self.build_tls_server(&mut tcp);
         let server = Server::builder(tls_server).serve(service);
 
         let graceful_server = server.with_graceful_shutdown(async {
@@ -144,19 +159,12 @@ impl HttpServerNode {
         );
 
         // Run until asked to terminate...
-        let result = graceful_server
-            // Terminate the server only if a shutdown signal is received.
-            // In case of errors, only log them instead of terminating the server.
-            .or_else(|e| async move {
-                log::error!("server error: {}", e);
-                Ok::<_, hyper::Error>(())
-            })
-            .await;
+        let result = graceful_server.await;
         info!("HTTP server pseudo-node terminated with {:?}", result);
     }
 
     /// Build a server that checks incoming TCP connections for TLS handshake.
-    async fn build_tls_server<'a>(&'a self, tcp: &'a mut TcpListener) -> TlsServer<'a> {
+    fn build_tls_server<'a>(&'a self, tcp: &'a mut TcpListener) -> TlsServer<'a> {
         let tls_cfg = crate::tls::to_server_config(self.tls_config.clone());
         let tls_acceptor = TlsAcceptor::from(tls_cfg);
 
@@ -166,6 +174,7 @@ impl HttpServerNode {
                 io::Error::new(io::ErrorKind::Other, format!("Incoming failed: {:?}", err))
             })
             .and_then(move |stream| {
+                log::debug!("Received incoming TLS stream: {:?}", stream);
                 tls_acceptor.accept(stream).map_err(|err| {
                     log::error!("Client-connection error: {:?}", err);
                     io::Error::new(io::ErrorKind::Other, format!("TLS Error: {:?}", err))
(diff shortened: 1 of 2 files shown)


oak_runtime/src/node/http/mod.rs
```diff
@@ -78,7 +78,16 @@ impl hyper::server::accept::Accept for TlsServer<'_> {
...
         cx: &mut Context,
     ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
-        Pin::new(&mut self.acceptor).poll_next(cx)
+        let connection = Pin::new(&mut self.acceptor).poll_next(cx);
+        match connection {
+            // In case of error (e.g., CertificateUnknown error), return Poll::Pending, to avoid
+            // early termination of the server.
+            Poll::Ready(Some(Err(e))) => {
+                log::error!("Error when processing TLS stream: {:?}", e);
+                Poll::Pending
```
tiziano88, line 87: What is the difference between this and `Poll::Ready(None)`?

In fact, I don't really understand why `Pending` works here, given what it says here https://doc.rust-lang.org/nightly/core/task/enum.Poll.html#variant.Pending :

> When a function returns Pending, the function must also ensure that the current task is scheduled to be awoken when progress can be made.

rbehjati: I also expected `Poll::Ready(None)` to work, but it does not. Which is strange, because as far as I can see, this is where `poll_accept` is called: 

https://github.com/hyperium/hyper/blob/66fc127c8d4f81aed9300c9d0f13246b8206067a/src/server/conn.rs#L773

If I return `Poll::Ready(None)`, the server terminates as before. 
I have added the call to `cx.waker().clone().wake()` before returning `Poll::Pending`, but I am not sure what impact it has.

tiziano88: I don't think that's what is expected to happen (at least, waking up the task before calling Pending does not match my mental model). Could you try and see what is happening exactly, and if this is the way to go, add more comments to explain how you arrived to your conclusion?

conradgrobler: Having looked at the Hyper code, I think returning `Poll::Pending` is appropriate. It treats `poll_accept` similarly to an async stream. Returning `Poll::Ready(None)` (or an error) signals the end of the stream, so it returns ouf of the loop.

I don't think `cx.waker().clone().wake()` does anything useful. `Poll::Pending` works because the context is passed to the underlying acceptor, which would call `wake` on the `waker` when there is a new incoming connection. From the server's point of view, it was woken up when there was a new connection, but the connection failed so there is nothing for it to do but wait for the next connection.

rbehjati: Thanks for the nice explanation @conradgrobler. This is my understanding too (based on what I observed), but I don't see where in the code `Poll::Ready(None)` signals the end of the stream. Do you know where in the code it happens?

conradgrobler: I believe it happens at https://github.com/hyperium/hyper/blob/66fc127c8d4f81aed9300c9d0f13246b8206067a/src/server/conn.rs#L852

rbehjati: But I expect to reach that line with both `Poll::Pending` and `Poll::Ready(None)`. Does not the code reach that line with `Poll::Pending` too?

conradgrobler: The `ready!` macro in https://github.com/hyperium/hyper/blob/66fc127c8d4f81aed9300c9d0f13246b8206067a/src/server/conn.rs#L843 means that the function would return earlly with `Poll::Pending` when it receives `Poll::Pending`. This in turn will cause the server to return `Poll::Pending` from its implementation of `poll`, which would in effect cause a wait until there is a new connection.

rbehjati: Oh! Of course! `ready!` is a macro. Thanks.

tiziano88: Also see https://docs.rs/hyper/0.13.7/src/hyper/server/tcp.rs.html#151

BTW can we just use `AddrIncoming` from that crate?

rbehjati: Is `AddrIncoming` handling TLS connections?

tiziano88: uhmm no, you are right, my bad :)

oak_runtime/src/node/http/mod.rs
```diff
@@ -166,6 +174,7 @@ impl HttpServerNode {
                 io::Error::new(io::ErrorKind::Other, format!("Incoming failed: {:?}", err))
             })
             .and_then(move |stream| {
```
tiziano88, line 176: just checking: is it possible that it would be simpler to handle the issue at this level? have you already tried that?

rbehjati: I did think of it but I cannot figure out how. 
It would be nice to call a `filter` method after `and_then` here. But there does not seem to be any such method. 
I also tried to implement something like https://github.com/hyperium/tonic/blob/04a8c0c82a4007f48c3bf3539a3f2312746fedd1/tonic/src/transport/server/incoming.rs#L18

but the types are too complicated, and there is no useful documentation. I would like to get something like `Result<TlsStream<TcpStream>, Error>` from `tokio_rustls::Accept`, and filter out the error cases, and only yield the ones that have a valid TlsStream. Let me know if you have any ideas for that conversion.

oak_runtime/src/node/http/mod.rs
tiziano88: Why would `ready!` not work here too then, given this method seems quite similar in structure to `AddrIncoming` anyways?

tiziano88 (approved):

oak_runtime/src/node/http/mod.rs
rbehjati: The logic here is different from `ready!`. `ready!` would return an error, we want to return `Poll::Pending`. Also, I don't think `ready!` is public (its definition does not have a `#[macro_export]` attribute).

tiziano88: I thought this is it, I'm not sure why it wouldn't work

https://docs.rs/futures-util/0.3.5/futures_util/macro.ready.html

rbehjati: I think the `ready!` macro used in hyper is this one: 
https://github.com/hyperium/hyper/blob/master/src/common/mod.rs

tiziano88: I meant that, in conjunction with some additional error handling logic, to be clear. Anyways, if it doesn't work, feel free to ignore.

tiziano88: > I think the `ready!` macro used in hyper is this one:
> https://github.com/hyperium/hyper/blob/master/src/common/mod.rs

I understand that, but why can we not use https://docs.rs/futures-core/0.3.5/src/futures_core/task/poll.rs.html#5-11 (which AFAICT has an equivalent implementation) and make the code look more similar to https://docs.rs/hyper/0.13.7/src/hyper/server/tcp.rs.html#98 for instance? To me it seems the pattern is quite similar. Anyways, we can try that out separately perhaps.

rbehjati: Yes. The pattern is similar. I can try that.

rbehjati: As discussed, I'll leave it as is (without using `ready!`) for now, since the `poll_accept` does not need to process the returned value apart from replacing it with `Poll::Pending` in case of an error. If in the future, we need a more complicated logic, we could use `ready!`.

PR #1446, opened 2020-09-08
Add labels to introspection events

juliettepretot:
# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/components/ApplicationStateOverview/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/ApplicationStateOverview/index.tsx
@@ -36,6 +36,8 @@ export default function ApplicationStateOverview({
               <dd>{id}</dd>
               <dt>Name:</dt>
               <dd>{nodeInfo.name}</dd>
+              <dt>Labels:</dt>
+              <dd>{nodeInfo.name}</dd>
               <dt>Handles:</dt>
               <dd>
                 <ul>
--- a/oak_runtime/introspection_browser_client/components/ApplicationStateOverview/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/ApplicationStateOverview/index.tsx
@@ -37,7 +37,7 @@ export default function ApplicationStateOverview({
               <dt>Name:</dt>
               <dd>{nodeInfo.name}</dd>
               <dt>Labels:</dt>
-              <dd>{nodeInfo.name}</dd>
+              <dd>{JSON.stringify(nodeInfo.label.toObject())}</dd>
               <dt>Handles:</dt>
               <dd>
                 <ul>
--- a/oak_utils/src/lib.rs
+++ b/oak_utils/src/lib.rs
@@ -229,7 +229,8 @@ where
             .type_attribute(".", "#[derive(::oak_io::handle::HandleVisit)]")
             // Link relevant Oak protos to the appropirate oak_io and oak_services types.
             .extern_path(".oak.handle", "::oak_io::handle")
-            .extern_path(".oak.encap", "::oak_services::proto::oak::encap");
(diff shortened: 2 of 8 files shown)


oak_utils/src/lib.rs
```diff
@@ -229,7 +229,8 @@ where
             .type_attribute(".", "#[derive(::oak_io::handle::HandleVisit)]")
             // Link relevant Oak protos to the appropirate oak_io and oak_services types.
             .extern_path(".oak.handle", "::oak_io::handle")
-            .extern_path(".oak.encap", "::oak_services::proto::oak::encap");
+            .extern_path(".oak.encap", "::oak_services::proto::oak::encap")
+            .extern_path(".oak.label", "::oak_abi::proto::oak::label");
```
conradgrobler, line 233: I don't think that this is required. Just importing `label.proto` in `introspection.proto` (which is already done) should be sufficient to use the labels there. These existing `extern_path` statements should only be needed to support handles in protos, e.g.:

```
    oak.handle.Sender join_room = 1 [(oak.handle.message_type) = ".oak.encap.GrpcResponse"];
```

juliettepretot: I'm inclined to think the same ([the prost docs seems to suggest this](https://docs.rs/prost-build/0.6.1/prost_build/struct.Config.html#method.extern_path)). 

I had previously thought that we already do this for the `oak.encap` proto, but I just saw that we do implement our own trait in the Rust module referenced here.

However when merely importing the `label.proto` in `introspection.proto` I seem to run into an issue with Prost. It does include & compile `label.proto`, yet I get the following error when compiling:

image


Is this an issue with Prost? Am I overlooking something?

juliettepretot: While the compiled label proto exists in the same folder, it's never imported, I think that's why the super reference does not resolve?

conradgrobler: Yes, I think so. I am not sure of the best approach. The current solution you have works, but seems excessive to add it to the shared `oak_utils` crate for something that will only be used in one place. Having said that, the Label struct is part of the oak_abi and might be used by more protos in future.

As you mentioned, another option is to include the generated rs file from the proto again in the runtime while compiling, but this seems like unnecessary duplication as the runtime already imports the type from oak_abi.

The last option I can think of (and probably my personal preference) is to add a reference to the existing Label struct in `oak_runtime/src/proto/mod.rs` to make sure `super::label::Label` can be resolved:

Something like

``` rust
    pub mod label {
        pub use oak_abi::proto::oak::label::Label;
    }
```

conradgrobler (approved):

oak_runtime/introspection_browser_client/components/ApplicationStateOverview/index.tsx
```diff
@@ -36,6 +36,8 @@ export default function ApplicationStateOverview({
               <dd>{id}</dd>
               <dt>Name:</dt>
               <dd>{nodeInfo.name}</dd>
+              <dt>Labels:</dt>
+              <dd>{nodeInfo.name}</dd>
```
conradgrobler, line 40: Sorry, missed this in my previous review. This should probably use something else rather than `nodeInfo.name`.

juliettepretot: Oh yes! Classic me, overlooked this. :)

oak_utils/src/lib.rs
juliettepretot: I quite like the last option you mentioned, it seems nicer than hacking into `extern_path` which doesn't seem to be intended to be used like this. 

I favor it over the second option. I tried that one earlier, but then we duplicate the type and need to convert from the `oak_abi` proto derived `Label` type to the `oak_runtime` proto derived `Label`.

PR #1447, opened 2020-09-08
Remove need for step_system_multi_backwards admit

jadephilipoom:
This PR, instead of proving the admitted lemma `step_system_multi_backwards`, removes the need for it by:
- flipping the order of transitivity in multi-step semantics so that forward steps are more natural
- replacing the use of `step_system_multi_backwards` with simpler reasoning based on the new semantics

Essentially, the change to the multi-step semantics made it so that, when constructing an inductive that proves there exists a sequence of valid steps starting with trace `t1` and ending with trace `t2`, we increase the sequence length by *adding events to `t2`*, rather than *removing them from `t1`*.

Most of the line-by-line changes are slight adjustments to the `TraceTheorems.v` proofs to match them with the new transitivity order.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -175,12 +175,13 @@ induction t1n.
             (a::t1n) Hinit_tleq H) as [t2n [E1 E2]].
             exists t2n. split. constructor. assumption. assumption.
         + rename Ht1_mstep_t1n into Ht1_mstep_at1n.
-        specialize (step_system_multi_backwards t2 t1n a H0) as E.
-        assert (H': step_system_ev_multi t1_init t2) by
-            (constructor; assumption).
-        specialize (IHt1n (step_system_transitive t1_init t2 t1n H' E))
+          match goal with
+            H : _ |- _ =>
+            apply step_system_ev_uncons in H end.
+          cbn [tl] in *; subst.
+          specialize (IHt1n ltac:(assumption))
             as [t2n [Hm_t2_init_t2n Hleq_t1n_t2n]].
-        specialize (step_system_multi_extends t2 t1n a H0) as E2.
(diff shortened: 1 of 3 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -175,12 +175,13 @@ induction t1n.
             (a::t1n) Hinit_tleq H) as [t2n [E1 E2]].
             exists t2n. split. constructor. assumption. assumption.
         + rename Ht1_mstep_t1n into Ht1_mstep_at1n.
-        specialize (step_system_multi_backwards t2 t1n a H0) as E.
-        assert (H': step_system_ev_multi t1_init t2) by
-            (constructor; assumption).
-        specialize (IHt1n (step_system_transitive t1_init t2 t1n H' E))
+          match goal with
+            H : _ |- _ =>
```
aferr, line 179: Just for my education, can you give me an example of a goal that this matches on if it's not too much trouble? (I'm just new to automation).

aferr (approved):

experimental/ni_coq/vfiles/PossibilisticNI.v
jadephilipoom: This one matches on any goal which contains a hypothesis `H` such that `step_system_ev_uncons` can by `apply`ed to `H`. Because of the shape of `step_system_ev_uncons`, that means any hypothesis of the form `step_system_ev _ _`, so as long as there's a goal like that in the context the match will succeed.

jadephilipoom: Concrete example (the simplest one) would be:
```coq
t1, t2 : trace
X : step_system_ev t1 t2
======
False
```
After the match, it would look like:
```coq
t1, t2: trace
X : t1 = tl t2
======
False
```

aferr: Thanks!

PR #1449, opened 2020-09-08
Compress introspection client files with gzip

juliettepretot:
Cuts the size of the generated JS bundle in half, so this seems sensible. 

# Checklist

      construction.


tiziano88 (approved):
This seems a very premature optimization to me, but fine to go ahead with it

juliettepretot:
@tiziano88 Very fair point. ^^ The reason I added it is that webpack kept nagging me by printing a warning about the performance implications of large assets. 🙈

PR #1451, opened 2020-09-09
Render oakApplicationState with graphviz

juliettepretot:
Replicates the behavior of Rust introspection by rendering the application state as a graph

# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
@@ -15,6 +15,7 @@
 //
 
 import React from 'react';
+import { graphviz } from 'd3-graphviz';
 import {
   OakApplicationState,
   ChannelHalfDirection,
@@ -22,10 +23,6 @@ import {
   ChannelID,
   AbiHandle,
 } from '~/components/Root';
-
-type StateGraphProps = {
-  applicationState: OakApplicationState;
-};
 
 // Generate a Graphviz dot graph that shows the shape of the Nodes and Channels
 function getGraphFromState(applicationState: OakApplicationState) {
@@ -91,14 +88,16 @@ function getGraphFromState(applicationState: OakApplicationState) {
   `;
 }
 
+type StateGraphProps = {
+  applicationState: OakApplicationState;
+};
+
 export default function StateGraph({ applicationState }: StateGraphProps) {
-  return (
-    <section
-      dangerouslySetInnerHTML={{
-        __html: getGraphFromState(applicationState)
-          .replace(/ /g, '&nbsp;')
-          .replace(/\n/g, '<br/>'),
-      }}
-    ></section>
-  );
+  const ref = React.useRef<HTMLDivElement>(null);
+  React.useEffect(() => {
+    const dotGraph = getGraphFromState(applicationState);
+    graphviz(ref.current).renderDot(dotGraph);
+  }, [applicationState]);
+
+  return <div ref={ref} />;
 }
--- a/oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
@@ -1,0 +1,104 @@
+//
+// Copyright 2020 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import React from 'react';
+import {
+  OakApplicationState,
+  ChannelHalfDirection,
+  NodeId,
+  ChannelID,
+  AbiHandle,
+} from '~/components/Root';
+
+type StateGraphProps = {
+  applicationState: OakApplicationState;
+};
+
+// Generate a Graphviz dot graph that shows the shape of the Nodes and Channels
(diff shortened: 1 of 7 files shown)


oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      ${oakHandles.join('\n      ')}
+    }
+    {
+      node [shape=ellipse style=filled fillcolor=green]
+      ${oakChannels.join('\n      ')}
+    }
+    ${connections.join('\n    ')}
+  }
+  `;
+}
```
juliettepretot, line 89: Replicates the `Runtime.graph` method in TypeScript.

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+} from '~/components/Root';
+
+// Generate a Graphviz dot graph that shows the shape of the Nodes and Channels
+function getGraphFromState(applicationState: OakApplicationState) {
+  const getNodeDotId = (nodeId: NodeId) => `node${nodeId}`;
+  const getChannelDotId = (channelId: ChannelID) => `channel${channelId}`;
+  const getHandleDotId = (nodeId: NodeId, handle: AbiHandle) =>
+    `node${nodeId}_${handle}`;
+
+  const oakNodes = [...applicationState.nodeInfos.entries()].map(
```
tiziano88, line 34: not sure whether the `oak` prefix actually adds much, it does not seem consistent throughout anyways.

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const oakHandles = [...applicationState.nodeInfos.entries()]
+    .map(([nodeId, nodeInfo]) =>
+      [...nodeInfo.abiHandles.entries()].map(
+        ([handle]) =>
+          `${getHandleDotId(nodeId, handle)} [label="${nodeId}:${handle}"]`
+      )
+    )
+    .flat();
+  const oakChannels = [
+    ...applicationState.channels.entries(),
```
tiziano88, line 46: What is this pattern with `[...something]`? Do they not cancel each other out?

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  ].map(([channelId]) => getChannelDotId(channelId));
+  const connections = [...applicationState.nodeInfos.entries()]
+    .map(([nodeId, nodeInfo]) =>
+      [...nodeInfo.abiHandles.entries()].map(([handle, channelHalf]) => {
+        switch (channelHalf.direction) {
+          case ChannelHalfDirection.Write:
+            return `${getNodeDotId(nodeId)} -> ${getHandleDotId(
+              nodeId,
+              handle
+            )} -> ${getChannelDotId(channelHalf.channelId)}`;
```
tiziano88, line 56: AFAICT the result of `getHandleDotId` gets interpolated in the string; if so, would it be possible to keep in one line? Or does autoformat also act inside strings?

tiziano88 (approved):

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
juliettepretot: You're correct that it gets interpolated into it. Agreed that the formatting is a bit strange, but enforced by prettier. It won't add line breaks inside of the actual string, but will add breaks within `${}` expressions.

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
juliettepretot: fair, I'll remove it. :)

oak_runtime/introspection_browser_client/components/StateGraph/index.tsx
juliettepretot: Lemme explain. :) The expression `[]` creates an array, and the spread expression (`...`) is used to spread the [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries) returned by `applicationState.channels.entries()` _inside_ of it. 

`const array = [...iterator]` is equivalent to 

```js
const array = []
for (const value of iterator) {
    array.push(value);
}
```


Another way to write this would be `Array.from(iterator)`, which might be clearer?

tiziano88: Thanks, I figured this was similar to `foo.iterator().collect()` in Rust :) but I assumed that since you are calling `map`, then you don't need the array to be materialized there?

Basically, why not just do the following?

```typescript
applicationState.channels.entries().map(...)
```

Is it just that you want the final type to be Array instead of iterator? If so though, it seems better to just do that at the end of the chain (which in Rust would be a final call to `collect`).

If what I'm suggesting does not make sense or it's not best practice, feel free to ignore, I'm mostly curious when it comes to type systems and functional idioms in languages I am not familiar with, so thanks for all the explanations so far already!

juliettepretot: Yup, I don't technically need an array here. At the end of the function I call `join` on this array, the final output type is a string.

The reason I convert the iterator to an array is because the `map` method (alongside `filter`, `flat`, `join` etc) is defined on the prototype of the Array object. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

The same method(s) does not exist for the iterator returned by `applicationState.channels.entries()`. 

Technically I could define my own functions (or methods) to do these things, which would be more memory efficient. 

```js
function* mapIterator(iterator, map) {
    for (const value of iterator) {
	    yield map(value)
	}
}
```

But in practice it's probably wise to save that optimization for later once/if it's needed & simply convert the iterators to arrays for now.

PR #1455, opened 2020-09-10
Introspection: Add basic styles & navigation

juliettepretot:
2020-09-10 2 03 53 pm

# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/components/MainTabs/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/MainTabs/index.tsx
@@ -1,0 +1,84 @@
+import React from 'react';
+import { makeStyles, Theme } from '@material-ui/core/styles';
+import AppBar from '@material-ui/core/AppBar';
+import Tabs from '@material-ui/core/Tabs';
+import Tab from '@material-ui/core/Tab';
+
+interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  value: number;
+}
+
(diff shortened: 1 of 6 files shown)


oak_runtime/introspection_browser_client/components/MainTabs/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const useTabPanelStyles = makeStyles((theme: Theme) => ({
+  root: {
+    display: 'flex',
+    height: '100%',
+  },
+}));
+
+function TabPanel({ children, value, index, ...other }: TabPanelProps) {
```
conradgrobler, line 20: Is there a guideline somewhere explaining why some functions are named using Pascal case and other camel case?

conradgrobler (approved):

oak_runtime/introspection_browser_client/components/MainTabs/index.tsx
juliettepretot: Yup! Normal JavaScript function are generally named in camelCase. 

In this case many functions are capitzaled because they are React components: https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized. They're not called directly, but rather from within a React tree.

Similarly by convention constructor functions are usually capitalized. Eg

```js
function Car(make, model, year) {
  this.make = make;
  this.model = model;
  this.year = year;
}

const car1 = new Car('Eagle', 'Talon TSi', 1993);
```

PR #1457, opened 2020-09-10
Proof automation in TraceTheorems.v

jadephilipoom:
Following up on a conversation with @aferr, this PR creates some tactic scripts from the proofs in `TraceTheorems.v` and uses them instead of manual manipulations. This should make the proofs less sensitive to code changes, and take some pain out of proving future theorems.

Some of the tactics (e.g. `destruct_match`) are pretty general-purpose and could be factored out so other files can import them.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/TraceTheorems.v
+++ b/experimental/ni_coq/vfiles/TraceTheorems.v
@@ -6,36 +6,55 @@ From OakIFC Require Import
     Events
     EvAugSemantics.
 
-Theorem nil_cons_rev: forall {A : Type} (x : A) (l : list A), x :: l <> [].
-Proof.
-    unfold not. intros. inversion H.
-Qed.
+(* inner loop of destruct_match *)
+Local Ltac destruct_match' e :=
+  lazymatch e with
+  | context [match ?x with _ => _ end] => destruct_match' x
+  | _ => destruct e
+  end.
+(* Finds the first match it encounters in the goal and destructs the
+   most deeply-nested match within it. *)
+Local Ltac destruct_match :=
+  match goal with
+  | |- context [match ?x with _ => _ end] => destruct_match' x
+  | H : context [match ?x with _ => _ end] |- _ => destruct_match' x
+  end.
+
+(* Runs [inversion] and then clears the original hypothesis, and runs
+   [subst], in order to prevent cluttering the context.*)
+Ltac invert_clean H :=
+  progress (inversion H; clear H); subst.
+
+(* Single step of [crush] *)
+Local Ltac crush_step :=
+  first [ progress intros
+        | progress subst
+        | progress cbn [length tl] in *
+        | match goal with
+          | H : head_st [] = Some _ |- _ =>
+            solve [inversion H]
+          | H : head_st (_ :: _) = Some _ |- _ =>
+            invert_clean H
+          | H : _ :: _ = _ :: _ |- _ =>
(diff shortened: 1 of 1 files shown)


experimental/ni_coq/vfiles/TraceTheorems.v
```diff
@@ -6,36 +6,55 @@ From OakIFC Require Import
...
     head_set_call t id c = t' ->
     length t = length t'.
-Proof.
-    intros; destruct t; destruct t'.
-    - reflexivity.
-    - inversion H.
-    - inversion H. destruct p. inversion H1.
-    - inversion H. destruct p. inversion H1. subst. reflexivity.
-Qed. 
+Proof. cbv [head_set_call]. crush. Qed.
```
aferr, line 49: Can I check my understanding of cbv/cbn with you? The coq documentation I could find mostly talks about using cbv with their built-in reduction rules (like beta, gamma, iota, etc). I think I understand what that does: it just applies the given reduction rule. When you pass in a specific definition like you're doing here, it looks like this applies the given definition to reduce the goal as much as it can. From this reference (https://github.com/coq/coq/wiki/Quick-Reference-for-Beginners), it sounds like the difference between cbv/cbn is that cbn will not continue to apply the definition recursively. Is that about right?

aferr (approved):

experimental/ni_coq/vfiles/TraceTheorems.v
```diff
@@ -6,36 +6,55 @@ From OakIFC Require Import
...
+Ltac invert_clean H :=
+  progress (inversion H; clear H); subst.
+
+(* Single step of [crush] *)
+Local Ltac crush_step :=
+  first [ progress intros
+        | progress subst
+        | progress cbn [length tl] in *
+        | match goal with
+          | H : head_st [] = Some _ |- _ =>
```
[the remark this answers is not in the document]
jadephilipoom: It depends on the structure of the codebase. For a file like this, where there are lots of proofs that need to do similar manipulations on a small set of definitions, I'd make a file-local `crush`-style tactic that includes those definitions. But if other collections of proofs end up needing something similar, I might make a generic one and combine it with the few specific clauses needed in each group.

experimental/ni_coq/vfiles/TraceTheorems.v
jadephilipoom: That's the gist of it; `cbv` will unfold the definition *no matter what*, as much as it can. And, to a first approximation, `cbn` will unfold the definition only if doing so makes the goal simpler somehow -- it can remove a match inside the function body, for instance. The way I use `cbn` most often is when I have some recursive definition and one level of the recursive argument is exposed, so that `cbn` will only unfold one call to the recursive function. However, in the case highlighted here, none of the arguments have any structure exposed, so as far as `cbn` is concerned it doesn't make the goal simpler to inline `head_set_call` and (untested, but I think so) `cbn` won't do anything.

It should be mentioned that these arguments are technically arguments to the `delta` argument of `cbv`, which deals with inlining functions. Just writing `cbv` or `cbn` is interpreted by default as using all reduction rules, e.g. `cbv iota beta zeta delta`. For this purpose, `cbv delta [head_set_call]` would probably do the same thing.

aferr: Thanks. I really appreciate your expanded explanation!

PR #1460, opened 2020-09-10
Respond with bad request if oak-label is not set in HTTP request

rbehjati:
Ref #1430

### DIFF
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -301,19 +301,45 @@ struct HttpRequestHandler {
 impl HttpRequestHandler {
     async fn handle(&self, req: Request<Body>) -> Result<Response<Body>, OakStatus> {
         let request = HttpServerNode::map_to_http_request(req).await;
-        let oak_label = get_oak_label(&request)?;
-        info!(
-            "Handling HTTP request; request size: {} bytes, label: {:?}",
-            request.body.len(),
-            oak_label
-        );
+        match get_oak_label(&request) {
+            Ok(oak_label) => {
+                info!(
+                    "Handling HTTP request; request body size: {} bytes, label: {:?}",
+                    request.body.len(),
+                    oak_label
+                );
 
-        debug!("Inject the request into the Oak Node");
-        let response = self
-            .inject_http_request(request, &oak_label)
-            .map_err(|_| OakStatus::ErrInternal)?;
+                debug!("Injecting the request into the Oak Node");
+                let response = self
+                    .inject_http_request(request, &oak_label)
+                    .map_err(|err| {
+                        warn!(
+                            "Error when injecting the request into the Oak Node: {:?}",
+                            err
+                        );
+                        OakStatus::ErrInternal
+                    })?;
 
-        Ok(response.to_response())
+                response.to_response().map_err(|e| {
+                    warn!("Could not create response: {}", e);
+                    OakStatus::ErrInternal
+                })
+            }
+            Err(OakStatus::ErrInvalidArgs) => http::response::Builder::new()
+                .status(StatusCode::BAD_REQUEST)
+                .body(Body::from("Invalid or missing Oak label."))
+                .map_err(|e| {
+                    warn!("Could not create response: {}", e);
+                    OakStatus::ErrInternal
+                }),
+            Err(_oak_status) => http::response::Builder::new()
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                .body(Body::from("Internal server error."))
+                .map_err(|e| {
+                    warn!("Could not create response: {}", e);
+                    OakStatus::ErrInternal
+                }),
+        }
     }
 
     fn inject_http_request(
@@ -488,24 +514,21 @@ impl HttpResponseIterator {
         response_receiver.receive(&self.runtime)
     }
 
-    fn to_response(&self) -> Response<Body> {
+    fn to_response(&self) -> Result<Response<Body>, http::Error> {
         info!(
             "Generating response for runtime {} and reader {:?}.",
             self.runtime.node_id.0, self.response_reader
         );
-        let mut response = Response::new(Body::empty());
         match self.read_response() {
-            Ok(http_response) => {
-                let status_code = http_response.status as u16;
-                *response.body_mut() = Body::from(http_response.body);
-                *response.status_mut() = StatusCode::from_u16(status_code)
-                    .unwrap_or_else(|_| panic!("Error when creating status code {}", status_code));
-            }
+            Ok(http_response) => http::response::Builder::new()
+                .status(http_response.status as u16)
+                .body(Body::from(http_response.body)),
             Err(status) => {
                 error!("Could not read response: {}", status);
-                *response.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
+                http::response::Builder::new()
+                    .status(StatusCode::INTERNAL_SERVER_ERROR)
+                    .body(Body::empty())
             }
         }
-        response
     }
 }


oak_runtime/src/node/http/mod.rs
```diff
@@ -444,35 +460,28 @@ fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
...
+    serde_json::from_str(&label_str).map_err(|err| format!("Could not parse HTTP label: {}", err))
 }
 
-fn parse_protobuf_label(protobuf_label: &[u8]) -> Result<Label, OakStatus> {
-    Label::decode(protobuf_label).map_err(|err| {
-        warn!("Could not parse HTTP label: {}", err);
-        OakStatus::ErrInvalidArgs
-    })
+fn parse_protobuf_label(protobuf_label: &[u8]) -> Result<Label, String> {
+    Label::decode(protobuf_label).map_err(|err| format!("Could not parse HTTP label: {}", err))
```
tiziano88, line 484: I think we need to be careful about anything that results in a user-facing message, for instance AFAICT now all these strings may end up being displayed to a user in a browser because we use them as errors in the various Results. A better practice is instead to log the error, and return a more fixed (and explicitly whitelisted) message to the user, which actually maps quite nicely to the usual pattern in Rust of defining an enum for the possible errors, and then having a way of converting that enum to a message in the function in which all the errors are handled.

In our case, I think we should probably be even more cautious, so I think it's fine to log the precise error on the server, and return a more generic error to the client, without risking revealing too much information to the user.

oak_runtime/src/node/http/mod.rs
```diff
@@ -488,24 +497,26 @@ impl HttpResponseIterator {
...
             "Generating response for runtime {} and reader {:?}.",
             self.runtime.node_id.0, self.response_reader
         );
-        let mut response = Response::new(Body::empty());
         match self.read_response() {
             Ok(http_response) => {
                 let status_code = http_response.status as u16;
-                *response.body_mut() = Body::from(http_response.body);
-                *response.status_mut() = StatusCode::from_u16(status_code)
+                let status = StatusCode::from_u16(status_code)
```
tiziano88, line 508: (pre-existing) Does this mean that if someone passes an invalid status code (e.g. `1`) this will panic?

tiziano88, line 508: (nit) the naming seems reversed here, it seems that `status_code` and `status` should be swapped?

oak_runtime/src/node/http/mod.rs
```diff
@@ -301,19 +301,35 @@ struct HttpRequestHandler {
...
+                info!(
+                    "Handling HTTP request; request size: {} bytes, label: {:?}",
+                    request.body.len(),
+                    oak_label
+                );
+
+                debug!("Inject the request into the Oak Node");
+                let response = self
+                    .inject_http_request(request, &oak_label)
+                    .map_err(|_| OakStatus::ErrInternal)?;
```
tiziano88, line 315: Log the error before discarding, or comment why it is not needed.

oak_runtime/src/node/http/mod.rs
```diff
@@ -301,19 +301,35 @@ struct HttpRequestHandler {
...
-        debug!("Inject the request into the Oak Node");
-        let response = self
-            .inject_http_request(request, &oak_label)
-            .map_err(|_| OakStatus::ErrInternal)?;
-
-        Ok(response.to_response())
+        match get_oak_label(&request) {
+            Ok(oak_label) => {
+                info!(
+                    "Handling HTTP request; request size: {} bytes, label: {:?}",
```
tiziano88, line 307:
```suggestion
                    "Handling HTTP request; request body size: {} bytes, label: {:?}",
```

I believe the size is that of the body, not of the whole request, right?

oak_runtime/src/node/http/mod.rs
```diff
@@ -301,19 +301,35 @@ struct HttpRequestHandler {
...
+                    request.body.len(),
+                    oak_label
+                );
+
+                debug!("Inject the request into the Oak Node");
```
tiziano88, line 312:
```suggestion
                debug!("Injecting the request into the Oak Node");
```

for consistency

oak_runtime/src/node/http/mod.rs
rbehjati: I agree. How about just responding with a generic response like "Invalid or missing oak-label" in line 332?

oak_runtime/src/node/http/mod.rs
rbehjati: Yes. The status comes from the Oak Node.

rbehjati: I used status because the method name is `status`. I guess both could be `status_code`.

oak_runtime/src/node/http/mod.rs
tiziano88: Yeah I think that's the safest for now, and as long as we log the more precise error, we should be able to debug things if anything goes wrong, at least while we have logging enabled.

But I would still change all the errors, it is not very idiomatic to use strings as errors. If you want unstructured errors, then perhaps anyhow could be used, but we don't normally use it at the library level, only at top level for binaries.

If there is no enum that does the job here, you could define a new one, or just use () as the error type, or use Option instead of Result.

oak_runtime/src/node/http/mod.rs
```diff
@@ -488,24 +503,26 @@ impl HttpResponseIterator {
...
-        let mut response = Response::new(Body::empty());
         match self.read_response() {
             Ok(http_response) => {
                 let status_code = http_response.status as u16;
-                *response.body_mut() = Body::from(http_response.body);
-                *response.status_mut() = StatusCode::from_u16(status_code)
+                let status_code = StatusCode::from_u16(status_code)
                     .unwrap_or_else(|_| panic!("Error when creating status code {}", status_code));
+                http::response::Builder::new()
+                    .status(status_code)
```
tiziano88, line 517: I think since now `body` is actually doing the check on everything from the builder, you can just pass `http_response.status` here, no need to parse it first. If it fails to parse, it should result in an http::Error anyways, right? If this does not work, then it may be worth adding manual logic to make the method return Err anyways.

oak_runtime/src/node/http/mod.rs
rbehjati: Done.

oak_runtime/src/node/http/mod.rs
rbehjati: I don't think I can just pass `http_response.status` the types are different.

oak_runtime/src/node/http/mod.rs
```diff
@@ -301,19 +301,45 @@ struct HttpRequestHandler {
...
+                })
+            }
+            Err(OakStatus::ErrInvalidArgs) => http::response::Builder::new()
+                .status(StatusCode::BAD_REQUEST)
+                .body(Body::from("Invalid or missing Oak label."))
+                .map_err(|e| {
+                    warn!("Could not create response: {}", e);
+                    OakStatus::ErrInternal
+                }),
+            Err(_oak_status) => http::response::Builder::new()
```
rbehjati, line 335: @tiziano88 Do you think we need a separate match arm and status code for each `OakStatus`? I am not even sure what the mapping would be, since we are not even using all OakStatuses. Perhaps it is best to define a separate Error type here. WDYT?

tiziano88: No, I think the current logic is fine. Basically we only distinguish between user error and internal error. An alternative may be to define a mapping from `OakStatus` to `StatusCode`, similar to what I think we already have for gRPC status codes.

oak_runtime/src/node/http/mod.rs
tiziano88: weird, looking at the type, to me it looks like it should work: https://docs.rs/http/0.2.1/http/response/struct.Builder.html#method.status

anyways, I'm probably missing something, maybe the error types don't line up

tiziano88 (approved):

oak_runtime/src/node/http/mod.rs
rbehjati: `http_response.status as u16` works.

oak_runtime/src/node/http/mod.rs
rbehjati: And return a `StatusCode` from `get_oak_label` instead of `OakStatus`?

tiziano88: No, keep returning OakStatus everywhere as it currently is, but in this function convert from OakStatus to StatusCode.

PR #1461, opened 2020-09-10
Implement more comprehensive HTTP example

tiziano88:
Rename path to uri in proto message.

Add instructions on how to run and test the example via a browser.

# Checklist

      construction.

### DIFF
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -16,7 +16,7 @@
 
 //! Simple example starting an Oak Application serving a static page over HTTP.
 
-use log::info;
+use log::{info, warn};
 use maplit::hashmap;
 use oak::{http::Invocation, Node, OakError};
 use oak_services::proto::oak::encap::HttpResponse;
@@ -37,15 +37,37 @@ impl Node<Invocation> for StaticHttpServer {
     fn handle_command(&mut self, invocation: Invocation) -> Result<(), OakError> {
         let request = invocation.receive()?;
 
-        info!("Handling a request to {}.", request.path);
+        info!("Handling a request to {}.", request.uri);
 
-        let http_response = HttpResponse {
-            body: "Hello from HTTP server!\n".to_string().into_bytes(),
-            status: 200,
-            headers: hashmap! {},
+        // Workaround for https://rust-lang.github.io/rust-clippy/master/index.html#borrow_interior_mutable_const.
+        static CONTENT_TYPE: http::header::HeaderName = http::header::CONTENT_TYPE;
+
+        let response = match request.uri.parse::<http::Uri>() {
+            Ok(uri) => match uri.path().as_ref() {
+                "/" => HttpResponse {
+                    body: include_bytes!("../static/index.html").to_vec(),
+                    status: http::StatusCode::OK.as_u16() as i32,
+                    headers: hashmap! {
+                        CONTENT_TYPE.as_str().to_string() => "text/html; charset=UTF-8".to_string().into_bytes(),
+                    },
+                },
+                _ => HttpResponse {
+                    body: "not found".to_string().into_bytes(),
+                    status: http::StatusCode::NOT_FOUND.as_u16() as i32,
+                    headers: hashmap! {},
+                },
+            },
+            Err(err) => {
+                warn!("could not parse URI: {}", err);
+                HttpResponse {
+                    body: "invalid URI".to_string().into_bytes(),
+                    status: http::StatusCode::BAD_REQUEST.as_u16() as i32,
+                    headers: hashmap! {},
+                }
+            }
         };
 
-        let res = invocation.send(&http_response);
+        let res = invocation.send(&response);
         invocation.close_channels();
         res
     }
--- a/oak_services/proto/http_encap.proto
+++ b/oak_services/proto/http_encap.proto
@@ -20,8 +20,8 @@ package oak.encap;
 
 // Protocol buffer encoding representing an HTTP request.
 message HttpRequest {
-  // The path to which the request was sent.
-  string path = 1;
+  // The URI (including scheme and authority) to which the request was sent.
+  string uri = 1;
   // The HTTP request method used.
   string method = 2;
   // The body of the request.
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -741,6 +741,7 @@ dependencies = [
 name = "http_server"
 version = "0.1.0"
 dependencies = [
+ "http",
  "log",
  "maplit",
  "oak",
--- a/examples/http_server/README.md
+++ b/examples/http_server/README.md
@@ -8,3 +8,27 @@ This Oak Application has an
 `StaticHttpServer` that implements
 [Node](https://project-oak.github.io/oak/sdk/doc/oak/trait.Node.html).
 `StaticHttpServer` is the node serving HTTP requests.
+
+To run the example interactively:
+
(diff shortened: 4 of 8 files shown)


examples/http_server/module/src/lib.rs
```diff
@@ -37,15 +37,37 @@ impl Node<Invocation> for StaticHttpServer {
...
 
-        info!("Handling a request to {}.", request.path);
+        info!("Handling a request to {}.", request.uri);
 
-        let http_response = HttpResponse {
-            body: "Hello from HTTP server!\n".to_string().into_bytes(),
-            status: 200,
-            headers: hashmap! {},
+        // Workaround for https://rust-lang.github.io/rust-clippy/master/index.html#borrow_interior_mutable_const.
+        static CONTENT_TYPE: http::header::HeaderName = http::header::CONTENT_TYPE;
```
rbehjati, line 43: Can this be moved outside the function?

tiziano88:
Uhmm this seems to make the HTTP node tests fail sometimes, but not consistently. Not sure why, will look into it. @rbehjati let me know if you have any suggestions on why that may be happening

oak_services/proto/http_encap.proto
```diff
@@ -20,8 +20,8 @@ package oak.encap;
 
 // Protocol buffer encoding representing an HTTP request.
 message HttpRequest {
-  // The path to which the request was sent.
-  string path = 1;
+  // The URI (including scheme and authority) to which the request was sent.
```
rbehjati, line 23: Not sure if it is true. The line that prints the `request.uri` in `examples/http_server/module/src/lib.rs` only prints "/", no scheme or authority.

rbehjati (approved):
This is nice. I have two minor comments.

oak_services/proto/http_encap.proto
tiziano88: Really? It was printing the entire URI when I tried on my machine, which is why I decided to change it. But also just looking at the code, it takes the URI from the incoming request, so I would expect that the entire URI gets propagated.

examples/http_server/module/src/lib.rs
tiziano88: Possibly yes, but since it's used only here, I think this is fine? Also TBH I don't really understand why clippy complains about it, the type seems pretty simple to me, so I'm not sure why it considers it internally mutable, maybe the macros used to generate the various header names are doing something dodgy internally.

tiziano88:
> Uhmm this seems to make the HTTP node tests fail sometimes, but not consistently. Not sure why, will look into it. @rbehjati let me know if you have any suggestions on why that may be happening

so actually I tried again from `main`, and it seems that tests are flaky there too. To reproduce:

```bash
cd oak_runtime
cargo test --package oak_runtime --lib -- node --nocapture
```

and then rerun the `cargo test` a few times. I got a failure roughtly 1 out of 5 times.

oak_services/proto/http_encap.proto
tiziano88: I'll wait until you confirm this, if what you say is correct then my change will probably break your use case, so I need to look into it more closely to understand what's going on.

rbehjati: That makes sense. I get only the path ("/") for requests that are send from the rust client. For requests from the browser and the bash client the entire URI is printed.

tiziano88: Thanks for confirming. Just to clarify: which Rust client are you referring to? AFAICT if it performs an actual HTTP request, it should result in the full URI anyways

rbehjati:
> Uhmm this seems to make the HTTP node tests fail sometimes, but not consistently. Not sure why, will look into it. @rbehjati let me know if you have any suggestions on why that may be happening

Do they actually fail or get stuck? Do you have an error log?

oak_services/proto/http_encap.proto
rbehjati: This is the line that sends the request:
https://github.com/project-oak/oak/blob/79de881763b06e83a00e80aeba740978ebd0949e/examples/http_server/client/src/main.rs#L61

tiziano88:
They fail, I'll post some logs later tonight.

examples/http_server/module/src/lib.rs
rbehjati: I also cannot see where the internal mutability comes from by having a quick look at the code for `HeaderName`.

tiziano88:
> > Uhmm this seems to make the HTTP node tests fail sometimes, but not consistently. Not sure why, will look into it. @rbehjati let me know if you have any suggestions on why that may be happening
> 
> Do they actually fail or get stuck? Do you have an error log?

Posted them here: https://github.com/project-oak/oak/issues/1462

oak_services/proto/http_encap.proto
tiziano88: I tried various combinations, but I still always see it logged as "https://localhost:8080" (which is what I would expect). Do you have steps that make it log "/" as the URL / path?

tiziano88: (since the discrepancy is already in `main`, I will go ahead with this PR anyways, but please do let me know how to reproduce)

rbehjati: I run the server using `./scripts/runner --logs run-examples --example-name=http_server --client-variant=none` (inside docker), then I run the rust client:
```
cd examples/http_server/client
cargo run
```

tiziano88: Thanks, I managed to reproduce that way, that's odd! Maybe the hyper client strips the uri before sending the request? not sure what's going on.

conradgrobler: I this the issue is that the hyper server reconstructs the URI from client headers. The raw HTTP request does not have the full URI of the request as a single field. In HTTP2 there are multiple headers (`:scheme`, `:host`, ":path") that can be used to reconstrutct the URI on the server. In HTTP1 the method and path are sent as the first line of the request body. A client can optionally also add a host header, but this is not required. Perhaps the rust client is defaulting to HTTP1 and does not use HTTP2.

tiziano88: Good point @conradgrobler , I didn't realise that was the case. @rbehjati could we make the HTTP server only use HTTP2 anyways? Not sure whether that is sufficient to solve this issue, but it is something we discussed we should do anyways.

rbehjati: What I remember from earlier discussions (and comments on #1261) is that HTTP server pseudo-node is supposed to only support HTTP/1 and JSON. Do we now have a consensus that we should completely switch to HTTP/2?

tiziano88: I don't see why not, anyways it's just a matter of disabling HTTP1 with https://docs.rs/hyper/0.13.7/hyper/server/struct.Builder.html#method.http2_only , everything else should stay the same, right? The Rust client, `curl` and Chrome should all still keep working (but let me know if that's not the case).

rbehjati: It would be interesting if everything stays the same. The pain of using hyper would pay off.

rbehjati: I could not find a way to make the rust client work. I have created this PR: https://github.com/project-oak/oak/pull/1481
The error is "failed to lookup address information: Name or service not known". It could be the same problem we were experiencing with TLS for gRPC that did not work without a domain name. Let me know if you can think of any solutions.

Issue #1458, opened 2020-09-10
HTTP headers should not be represented as a map


tiziano88:
https://github.com/project-oak/oak/blob/79de881763b06e83a00e80aeba740978ebd0949e/oak_services/proto/http_encap.proto#L29-L30

HTTP headers may be repeated, so a map representation is not appropriate

tiziano88:
cc @rbehjati

PR #1463, opened 2020-09-10
Make http_server example client actually log

tiziano88:
It should help debugging the issue described in https://github.com/project-oak/oak/pull/1461#discussion_r486658233

# Checklist

      construction.

### DIFF
--- a/examples/http_server/client/src/main.rs
+++ b/examples/http_server/client/src/main.rs
@@ -31,6 +31,7 @@ pub struct Opt {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    env_logger::init();
     // Send a request, and wait for the response
     let label = oak_abi::label::Label::public_untrusted();
     let label_bytes = serde_json::to_string(&label)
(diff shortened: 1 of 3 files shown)


examples/http_server/client/src/main.rs
```diff
@@ -68,6 +69,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .await
         .expect("Error while awaiting response");
 
-    log::info!("Got response: {:?}", resp);
+    log::info!("response: {:?}", resp);
```
tiziano88, line 72: BTW it seems that the response headers (e.g. content-type) are not there when this is logged, @rbehjati , not sure whether they are lost along the way, or I am doing something wrong when logging

rbehjati: Nope! My bad. The headers are not set in the response! I can fix it with #1458.

tiziano88: SG, thanks! I think this PR can go ahead without that anyways (and hopefully will help to check that issue is fixed, once you have a PR for it)

examples/http_server/client/src/main.rs
```diff
@@ -68,6 +69,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    log::info!(
```
rbehjati, line 73: Does not the previous line already print the body of the response too?

rbehjati (approved):

examples/http_server/client/src/main.rs
tiziano88: No, it just prints `Body(stream)` or something like that.

PR #1464, opened 2020-09-11
Add PSI Rust client

ipetr0v:
This change adds a Rust Private Set Intersection client.

# Checklist

### DIFF
--- a/examples/private_set_intersection/client/rust/Cargo.toml
+++ b/examples/private_set_intersection/client/rust/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "private_set_intersection_client"
+version = "0.1.0"
+authors = ["Ivan Petrov <ivanpetrov@google.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+name = "private_set_intersection_client"
+path = "src/lib.rs"
+
+[[bin]]
+name = "psi_client_bin"
+path = "src/main.rs"
+
+[dependencies]
+anyhow = "*"
+base64 = "*"
+env_logger = "*"
+http = "*"
+log = "*"
+maplit = "*"
+oak_abi = "=0.1.0"
+pem = "*"
+prost = "*"
+structopt = "*"
+tokio = { version = "*", features = ["fs", "macros", "sync", "stream"] }
+tonic = { version = "*", features = ["tls"] }
+
+[build-dependencies]
+oak_utils = "*"
--- a/examples/private_set_intersection/client/rust/src/main.rs
+++ b/examples/private_set_intersection/client/rust/src/main.rs
@@ -1,0 +1,185 @@
+//
+// Copyright 2020 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Client for the Private Set Intersection example.
+
+use anyhow::{anyhow, Context};
+use http::uri::Uri;
+use log::info;
+use maplit::hashset;
+use oak_abi::label::{web_assembly_module_signature_tag, Label};
+use private_set_intersection_client::proto::{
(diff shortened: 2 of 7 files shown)


examples/private_set_intersection/client/rust/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+name = "private_set_intersection_client"
+path = "src/lib.rs"
+
+[[bin]]
+name = "psi_client_bin"
+path = "src/main.rs"
```
tiziano88, line 14: Why are these both needed?

examples/private_set_intersection/client/rust/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+    Channel::builder(uri.clone())
+        .tls_config(tls_config)
+        .context("Couldn't create TLS configuration")?
+        .connect()
+        .await
+        .context("Couldn't connect to Oak Application")
+}
+
+/// Create Oak gRPC client.
+async fn create_client(
```
tiziano88, line 68: What is async about this function?

examples/private_set_intersection/client/rust/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+        .context("Couldn't load certificate file")?;
+    let public_key_file = tokio::fs::read(&opt.public_key)
+        .await
+        .context("Couldn't load public key file")?;
+    let public_key_bytes = pem::parse(public_key_file)
+        .context("Empty public key file")?
+        .contents;
+
+    // Submit sets from different clients.
+    let channel_1 = create_channel(&uri, &root_tls_certificate)
```
tiziano88, line 109: Who does 1-based indexing? :)

examples/private_set_intersection/client/rust/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+        .context("Couldn't create gRPC client")?;
+
+    let request = Request::new(SubmitSetRequest {
+        values: vec!["c".to_string(), "d".to_string(), "e".to_string()],
+    });
+    invalid_client
+        .submit_set(request)
+        .await
+        .err()
+        .context("Submitted set with an invalid public key label")?;
```
tiziano88, line 154: I am not sure I am reading this correctly, I thought you would expect an error in this case? Could you make it clearer anyways?

tiziano88 (approved):

examples/private_set_intersection/client/rust/Cargo.toml
ipetr0v: `lib` is needed for integration tests, in order to include `private_set_intersection_client::PrivateSetIntersectionClient`

examples/private_set_intersection/client/rust/src/main.rs
ipetr0v: Removed

examples/private_set_intersection/client/rust/src/main.rs
ipetr0v: We do =)
https://github.com/project-oak/oak/blob/754c3c9d49939f8fa177bef80b7a0e8a99be1f5c/examples/private_set_intersection/client/private_set_intersection.cc#L72-L78

examples/private_set_intersection/client/rust/src/main.rs
ipetr0v: Yes, it fails if there is no error. `.err()` returns `Some` if there is error and `None` otherwise.
Added a comment

examples/private_set_intersection/client/rust/src/main.rs
ipetr0v: Not sure if it's better to rename it to `first_channel` etc.
Since we need to connect to Oak from two distinct clients.

rbehjati (approved):

examples/private_set_intersection/client/rust/src/main.rs
tiziano88: The c++ code you linked does 0-based indexing, right? (Not that it matters much, but not sure I missed your point)

first and second seem worse options to me, we already have numbers that we can use to enumerate things, why use words? :)

ipetr0v:
So, if we have multiple clients in the same example (that blocks new submissions after client sent a request), then we are not able to run both clients on the same server instance (without rerunning it).
Should I remove PSI C++ client, or make `runner` rerun the server on each example?

cc @tiziano88

tiziano88:
Maybe we should have a bucket or set id, to keep different sets of clients separate?

tiziano88:
I note that this is now essentially the same as the aggregator example, in which the aggregation function is "set intersection".

PR #1467, opened 2020-09-11
Broaden the supported architectures for the Oak CI and Runner.

ebeasant-arm:
This is a reasonably sized PR covering steps needed to support more generic architechtures in Oak.
This comprises a Rust change to the runner to select apropriate target specifications for varying architectures,
and some CI changes with a little aarch64 specific content. Please see the Comit Message for more details.

# Checklist

      construction.


tiziano88 (approved):
Thanks, LG but please wait for @ipetr0v to take a look too before submitting!

scripts/common
```diff
@@ -11,8 +11,36 @@ set -o pipefail
 export RUST_LOG="${RUST_LOG:-info}"
 
 # See https://pantheon.corp.google.com/gcr/images/oak-ci/GLOBAL/oak
-readonly DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak'
-readonly SERVER_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-server'
+# Select the correct Dockerfile for the underlying arch.
+# Note that these can all be overridden from the environment as desired.
+# Docker's target types align wih Go, and not with bash/linux, hence the
```
ipetr0v, line 16:
```suggestion
# Docker's target types align with Go, and not with bash/linux, hence the
```

ipetr0v (approved):

arm64v8.Dockerfile
```diff
@@ -214,36 +236,37 @@ RUN npm install --global \
...
-#ENV PATH "${hadolint_dir}:${PATH}"
-#RUN mkdir --parents ${hadolint_dir} \
-#  && curl --location https://github.com/hadolint/hadolint/releases/download/v${hadolint_version}/hadolint-Linux-x86_64 > ${hadolint_bin} \
-#  && sha256sum --binary ${hadolint_bin} && echo "${hadolint_sha256} *${hadolint_bin}" | sha256sum --check \
-#  && chmod +x ${hadolint_bin} \
-#  && hadolint --version
+# TODO: Add hadolint version
+# Debian's Stack is not recent enough in the Buster distribution to build or run Hadolint.
+# We use Cabal to build it fron scratch.
+# This requires a rebuild of cabal in the process to update to 3.2.0
```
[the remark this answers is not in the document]
ebeasant-arm: We're invetigating why GHC, Cabal, Stack dropped all arm64 binary releases: its notionally due to lack of CI to run them, but that really causes problems for any distributions that are cross-platform. Much of this is due to people using their own CI to drop containerised binaries (!) direct from GitHub/Lab, circumventing any of the distro maker's tooling. Its deeply frustrsating, but hopefully we can organise to help provide some support to GitHub/Lab's free tier for aarch64/arm64v8 builders.

runner/src/main.rs
```diff
@@ -50,15 +51,23 @@ use check_license::CheckLicense;
...
 
-#[cfg(target_os = "macos")]
-const DEFAULT_SERVER_RUST_TARGET: &str = "x86_64-apple-darwin";
-#[cfg(not(target_os = "macos"))]
-const DEFAULT_SERVER_RUST_TARGET: &str = "x86_64-unknown-linux-musl";
+use lazy_static::lazy_static;
 
 #[cfg(target_os = "macos")]
-const DEFAULT_EXAMPLE_BACKEND_RUST_TARGET: &str = "x86_64-apple-darwin";
+lazy_static! {
```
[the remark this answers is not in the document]
ebeasant-arm: I'll have a look at once_cell: Thanks for the pointer. I used the runtime composition here to avoid having to embed the multiple combinations in constants. Its not elegant, but it is hopefully more of a 'write once' approach.

arm64v8.Dockerfile
```diff
@@ -133,45 +147,54 @@ RUN echo 'deb http://deb.debian.org/debian buster-backports main' > /etc/apt/sou
...
 # Emscripten brings Node with it, we need to allow non-root access to temp folders
-#RUN mkdir "/.npm" && chmod a+rwx "/.npm"
+RUN mkdir "/.npm" && chmod a+rwx "/.npm"
 
 # Install Go.
 ARG golang_version=1.14.6
 # [amd64] ARG golang_sha256=aed845e4185a0b2a3c3d5e1d0a35491702c55889192bb9c30e67a3de6849c067
-ARG golang_sha256=291bccfd7d7f1915599bbcc90e49d9fccfcb0004b7c62a2f5cdf0f96a09d6a3e
+# ARG golang_sha256=291bccfd7d7f1915599bbcc90e49d9fccfcb0004b7c62a2f5cdf0f96a09d6a3e
+#  && sha256sum --binary ${golang_temp} && echo "${golang_sha256} *${golang_temp}" | sha256sum --check \
```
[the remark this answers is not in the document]
ebeasant-arm: Good catch - I'll get the most recent sha256sum for the image and restore it.

runner/src/main.rs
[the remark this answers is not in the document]
ebeasant-arm: I knew there was a reasonI jumped for lazy_static - it supports a no_std spinlock implementation as well: If Oak never plans to be no_std (?), then I'll convert to once_cell.

arm64v8.Dockerfile
```diff
@@ -283,21 +306,17 @@ ARG install_dir=${rustup_dir}/bin
 
 # Install grcov.
 # https://github.com/mozilla/grcov
-ARG grcov_version=v0.5.15
-ARG grcov_location=https://github.com/mozilla/grcov/releases/download/${grcov_version}/grcov-linux-x86_64.tar.bz2
-RUN curl --location ${grcov_location} | tar --extract --bzip2 --directory=${install_dir}
-RUN chmod +x ${install_dir}/grcov
+ARG grcov_version=0.5.15
+RUN cargo install --version ${grcov_version} grcov
```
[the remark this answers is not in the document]
ebeasant-arm: No problem, on it ...

runner/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[cfg(target_os = "macos")]
+    {
+        dsrt = consts::ARCH.to_owned() + "-apple-darwin";
+        debrt = consts::ARCH.to_owned() + "-apple-darwin";
+    }
+    #[cfg(not(target_os = "macos"))]
+    {
+        dsrt = consts::ARCH.to_owned() + "-unknown-" + consts::OS + "-musl";
+        debrt = consts::ARCH.to_owned() + "-unknown-" + consts::OS + "-gnu";
+    }
```
tiziano88, line 31: would something like this work? Up to you though, the current one is just fine.

```rust
let dsrt = if cfg!(target_os = "macos") {
  ...
} else {
  ...
};
```

runner/build.rs
```diff
@@ -0,0 +1,37 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use std::env::consts;
+
+fn main() {
+    let dsrt;
+    let debrt;
```
tiziano88, line 21: What do these names mean? Could you add a comment / reference or rename them to something more explicit?

edit: I realised only now that they are the initals of the respective env vars, that was quite obscure and not very idiomatic in Rust :P Could you give them appropriate names, even if they are long? Also please explain in a comment why this approach is necessary.

runner/src/main.rs
```diff
@@ -285,15 +275,15 @@ fn build_server(opt: &BuildServer) -> Step {
                         "-Zunstable-options".to_string(),
                         ...match opt.server_variant {
                             ServerVariant::Logless => vec!["--no-default-features".to_string(),
-                                format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET)),
+                                format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(&std::env::var("DEFAULT_SERVER_RUST_TARGET").unwrap())),
```
tiziano88, line 278: I am not convinced this approach is worth the effort of passing these as env variables that are set in the build script. What is the problem that it solves, vs for instance having a method in this file that returns the appropriate value for each of these strings? Is it because `const::ARCH` is only available at build time? Either way, it would be great to have it written down, for future developers touching this code.

runner/build.rs
ebeasant-arm: *facepalm* once again, guilty of no comments because I moved on to testing and forgot to note to make sure they were explained...

runner/src/main.rs
ebeasant-arm: Yep, comments comments and comments needed: The long and short of it is that the structopt default_values are insufficent when the defaults change with the platform on which the system is built. By taking the specifics of the cludge out of the main code and into the build script, it keeps the rest of it generic. That means any new platform can tweak in the build.rs without having to fiddle with the main runner codebase. That particularly egrigious line is due to the type-coercion needed here for String to &str, and the intent for a deliberate failure if the DEFAULT_SERVER_TARGET is unset by the build script, which is the case only on an unrecognised platform (it will blow up on the unwrap).

ebeasant-arm: If I wrote a specific structopt parser for the strings here, I think it could improve it further, and would remove the nasty hackage here.

runner/build.rs
```diff
@@ -0,0 +1,78 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use platforms::{platform::Platform, target::OS};
+
+fn main() {
```
tiziano88, line 19: I am still not a big fan of build.rs files in general, why would the same logic not work as part of main.rs?

tiziano88: IIUC your other comment, it is just to have a separation so that adding a new platform only requires modifying build.rs and not main.rs? If that's the only reason, IMO it does not justify the complexity of the build script and passing arguments as strings over env variables. The same logic may (and should) still be encapsulated in main.rs so that someone just wanting to add a new platform would be able to do so without modifying (or understanding) the rest of the file.

ebeasant-arm: Its peculiar that we have quite opposing views from the similar reasoning! :
The use of build.rs for me is a reduction in complexity of main.rs, and represents a reduction in cludginess of encoding platform strings in the functional code. It separates the platform adaptions cleanly from the functional code, and by using a separate file it helps de-conflict unrelated merges of code/platform hacking (it would keep me out of main.rs ;-). The code uses the environmental variables to help avoid needing to use once-cell type patterns or globals in a similar way, which to my mind further increases complexity (and decreases readability). Conversely, If the code in was imported to main.rs, the volume of code remains would probably actually increase a little, along with the dependencies lists for that module.
I was trying to save everyone a few headaches in future, deconflict development activities and aid the reabability :-)

All this said:

I'll happily put the code in main.rs, your codebase, your rules :-)

Updated patch incoming, I had to also remove the system NPM implementation as is was conflicting... more changes ahoy....

tiziano88: Ah I see, but I don't think the once-cell optimizations are necessary, a method that re-computes the string every time is more than fine, I don't think we should worry about a few extra string allocations in a dev tool. Does that seem a reasonable compromise to you?

ebeasant-arm: All works for me: I almost had it completely cleanly solved once, then ran into the 'can't implement a trait from one crate on a struct from another' rule, which scuppered that design too... I'll beat it into submission one day ;-) thanks for the tolerance, Cheers !

tiziano88: I think what I'm suggesting boils down to just moving the logic from this file to two methods in main.rs: `get_server_rust_target()` and `get_client_rust_target()`, and just invoke them as many times as necessary throughout the code, no need to worry about trying to do it exactly once. Also no need to make the structopt default line up, I think it's fine to make that an Option, and either use the provided value (if any), or if not, rely on the one comeing from the appropriate aforementioned method.

tiziano88: Also if you prefer, feel free to put these in a separate platform.rs file, and include that from main.rs, to keep the logic separate, up to you.

tiziano88:
Hey @ebeasant-arm , are you still looking into this at all? We would still be keen to merge this if you are up for it, I think it needs rebasing and a few comments need to be replied to, let me know how you want to proceed, happy to help to get this through!

ebeasant-arm:
Hi there - Haskell was a blocker for a bit, I'll pick this up and rebase it and see how we go from there, many thanks!

Issue #1456, opened 2020-09-10
running examples twice in a row with no modifications results in the introspection client code and the oak loader binary being recompiled


tiziano88:
To reproduce, run the following command from docker:

```bash
./scripts/runner --logs run-examples --example-name=http_server
```

(or any other example)

then run it again immediately after.

Observed: `build introspection browser client` and `build server` steps are re-run, taking up to a few minutes every time

Expected: these steps should be instantaneous since everything should still be cached

@juliettepretot could you look into it? perhaps there is some flag to make sure that the compilation of those assets only happens when they changed?

juliettepretot:
thx for reporting, will investigate

juliettepretot:
There isn't a build in flag to do this, but will try to write small script that handles this.

tiziano88:
Are the output files deterministic at least? Or does their contents change? I guess either way, `include_bytes` will think they have been updated, even if only the timestamp is different, right?

juliettepretot:
Yes they should be deterministic. What happens right now is that we build client files _every time_, since Webpack & NPM offer no build in cache. Since they are rebuilt every-time. Pretty sure that `include_bytes` indeed only checks the timestamp.

Either way we probs want to skip building the client if its source has not changed, since that itself takes around 15s on my machine.

juliettepretot:
Fixed the issue of cargo rebuilding the `oak_loader` in #1487. 
Reduced the time spent on building introspection assets from ~40s to ~4s with webpack caching. 

I could probably cache building introspection assets even more aggressively (bringing it down to <1s by skipping webpack altogether if source files did not change), but the maintenance cost might not be worth it. 

@tiziano88 Can I close this issue?

tiziano88:
Yeah I think the current state of things looks good, thanks!

juliettepretot:
Cool beans, closing this then :)

PR #1476, opened 2020-09-15
Fix oak_loader being compiled on each runner run

juliettepretot:
Addresses (one half of) https://github.com/project-oak/oak/issues/1456:

Background: The introspection client is built on each run, changing the timestamp on output assets, even if the resulting file is identical. Cargo checks this timestamp to determine if these files have changed. Since timestamps are different on each run, cargo recompiles each binary that includes these files, on every run. This patches this via a webpack plugin that ensures that the timestamp of the introspection assets only changes if the contents changed.  

Reduces the time spent on the `build server` step from ~55s to <1s on subsequent runs. 

Next up I'll look into some optimizations to cache/skip build steps of the introspection client to speed up that step as well. (Currently ~30s on every run)

# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/webpack.config.js
+++ b/oak_runtime/introspection_browser_client/webpack.config.js
@@ -17,11 +17,88 @@
 const path = require('path');
 const CopyPlugin = require('copy-webpack-plugin');
 const CompressionPlugin = require('compression-webpack-plugin');
+const fs = require('fs');
+const { promisify } = require('util');
+const rimraf = require('rimraf');
+
+const readFileAsync = promisify(fs.readFile);
+const writeFileAsync = promisify(fs.writeFile);
+const rimrafAsync = promisify(rimraf);
+
+const OUTPUT_DIR = './dist';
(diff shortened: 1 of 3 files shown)


tiziano88:
Adding @ipetr0v to review, who implemented something similar in rust at some point, though it turned out not to be necessary in the end

oak_runtime/introspection_browser_client/webpack.config.js
```diff
@@ -17,11 +17,88 @@
...
+// Ref: https://github.com/project-oak/oak/issues/1456
+class CopyIfNonExistent {
+  constructor(
+    options = {
+      sourceDir: undefined,
+      destinationDir: undefined,
+      deleteSourceDir: false,
+    }
+  ) {
+    if ([undefined].includes(options.sourceDir, options.destinationDir)) {
```
juliettepretot, line 44:
```suggestion
    if ([options.sourceDir, options.destinationDir].includes(undefined)) {
```

juliettepretot:
Closing in favor of #1487

PR #1484, opened 2020-09-16
Skip NPM installs if dependencies didn't change

juliettepretot:
Addresses #1456 through skipping time expensive NPM installs on subsequent runs, if dependencies did not change.

When dependencies are initially installed, copy of the `package-lock.json` file representing the installed dependencies is stored in a cache. On subsequent builds, if the cached `package-lock.json` matches the present `package-lock.json` dependency installations can be safely skipped. 

# Checklist

      construction.


juliettepretot:
Since this adds quite a few bash scripts to the root of the introspection client dir I'll follow this up with a PR to restructure that directory for clarity. 

Eg separate directories for `scripts`, `src`, etc

oak_runtime/introspection_browser_client/build
```diff
@@ -2,8 +2,12 @@
 #
 # Build browser introspection client.
 
-# Install npm modules
-npm ci
+# Install npm modules if needed
+readonly CURRENT_PACKAGE_LOCK="./package-lock.json"
+readonly LAST_INSTALLED_PACKAGE_LOCK="./node_modules/.cache/last-install-package-lock.json"
+if ! cmp -s "$CURRENT_PACKAGE_LOCK" "$LAST_INSTALLED_PACKAGE_LOCK"; then
```
tiziano88, line 8: Would this read better with `diff` instead of `cmp`? (I must admit I had never heard of `cmp` before)

tiziano88, line 8:
```suggestion
if ! cmp --silent "$CURRENT_PACKAGE_LOCK" "$LAST_INSTALLED_PACKAGE_LOCK"; then
```

oak_runtime/introspection_browser_client/npm_hook_package_added_or_removed
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+#
+# Removes the stale ./node_modules/.cache/last-install-package-lock.json after
+# an NPM package has been added or removed. This additional hook is needed since
+# that file can only be genereated after dependencies are installed as whole,
+# but not when individual dependencies are added or removed.
+
+# Check if this hooks was called from a dependency, not from the source package.
+if [[ $INIT_CWD != $PATH ]]; then
```
tiziano88, line 9:
```suggestion
if [[ "$INIT_CWD" != "$PATH" ]]; then
```
(btw I would expect shellcheck to complain about this, not sure why it's not)

oak_runtime/introspection_browser_client/npm_hook_package_added_or_removed
```diff
@@ -0,0 +1,12 @@
...
+  cd $INIT_CWD
```
tiziano88, line 10:
```suggestion
  cd "$INIT_CWD"
```

oak_runtime/introspection_browser_client/npm_hook_package_added_or_removed
tiziano88, line 9: where is this `INIT_CWD` variable coming from? Could you link to some documentation / reference?

oak_runtime/introspection_browser_client/postinstall
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#
+# Maintains a copy representing the package-lock.json at the time of the
+# last NPM install in ./node_modules/.cache/last-install-package-lock.json.
+# Doing this permits the build script to skip npm installs on subsequent runs
+# if dependencies did not change.
+
+# Setup hook to run on subsequent additions and removals of dependencies.
+# Ref: https://docs.npmjs.com/misc/scripts#hook-scripts
+if [[ ! -d "./node_modules/.hooks/postinstall" || ! -d "./node_modules/.hooks/postuninstall" ]]; then
```
tiziano88, line 10: If possible use consistent quoting for these paths, here and below

tiziano88 (approved):

oak_runtime/introspection_browser_client/build
juliettepretot: Could be! `cmp` is faster since it stops at the first byte difference, but admittedly with files this small it probably wouldn't make a practical difference. 

Do you think I should switch to `diff` for legibility?

PR #1488, opened 2020-09-16
Use "standard" Monoid definition from frunk crate

tiziano88:
# Checklist

      construction.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -72,6 +72,7 @@ dependencies = [
  "aggregator_grpc",
  "assert_matches",
  "env_logger",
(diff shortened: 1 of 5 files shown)


tiziano88:
This is not particularly useful, but there are other interesting traits in the https://docs.rs/frunk/0.3.1/frunk/index.html crate that it would be nice to explore at some point, so I thought I'd pull it in to make sure it compiles etc, even if we are only using two traits for now.

ipetr0v (approved):

tiziano88:
There is some conflict in the transitive dependencies, I'll switch this back to draft for now.

PR #1490, opened 2020-09-16
Write call single-step, single-thread unwind

aferr:
This commit finishes a proof of the single-step, single-threaded
unwinding condition for just the write call and just in the case where
the label of the node making the call does flow to the label of the
observer. Several new smaller theorems are introduced but admitted for
now. They should be relatively simple. The files are reorganized a bit.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -1,68 +1,146 @@
 From OakIFC Require Import
-        RuntimeModel
         Parameters
+        RuntimeModel
+        EvAugSemantics
+        Events
         LowEquivalences
-        GenericMap.
+        Lattice.
 Require Import Coq.Lists.List.
 Import ListNotations.
+From mathcomp Require Import all_ssreflect finmap.
+From RecordUpdate Require Import RecordSet.
+Import RecordSetNotations.
 
-Theorem chan_upd_preserves_nodes: forall h ch s,
-    (state_upd_chan h ch s).(nodes) = s.(nodes).
+
+Theorem trace_leq_imples_head_st_leq: forall ell t1 t2 s1 s2,
+    (head_st t1 = Some s1) ->
+    (head_st t2 = Some s2) ->
+    (trace_low_eq ell t1 t2) ->
+    (state_low_eq ell s1 s2).
 Proof.
-    intros; simpl; reflexivity.
+    inversion 3. 
+    - 
+        exfalso. rewrite <- H3 in H. inversion H.
+    - 
+        assert (xs = s1). {
+            assert (head_st ((xs, xe) :: t0 ) = Some xs).
+            reflexivity. congruence.
+        }
+        assert (ys = s2). {
+            assert (head_st ((ys, ye) :: t3 ) = Some ys).
+            reflexivity. congruence.
+        }
+    congruence.
 Qed.
 
-Theorem chan_append_unwind: forall ell m ch1 ch2,
-    (chan_low_eq ell ch1 ch2) ->
-    (chan_low_eq ell (chan_append ch1 m) (chan_append ch2 m)). 
+Theorem state_leq_and_flowsto_to_node_eq: forall ell s1 s2 id n1 n2,
+    (nodes s1).[? id] = Some n1 ->
+    (nodes s2).[? id] = Some n2 ->
+    (state_low_eq ell s1 s2) -> 
+    (nlbl n1 <<L ell) ->
+    n1 = n2.
 Proof.
-    intros. inversion H. 
-    constructor. rewrite H0. reflexivity.
-    constructor 2.
-    apply H1. apply H2.
-Qed. 
+    inversion 3. specialize (H2 id). rewrite H H0 in H2.
+    inversion H2. subst. reflexivity. contradiction.
+Qed.
 
-Theorem leq_ch_upd_preserves_sleq: forall (ell: level)
-    (h:handle)(ch1 ch2: channel)(s1 s2:state),
-    (chan_low_eq ell ch1 ch2) ->
-    (state_low_eq ell s1 s2) ->
-    (state_low_eq ell 
-        (state_upd_chan h ch1 s1)
-        (state_upd_chan h ch2 s2)).
+Theorem state_upd_chan_preserves_node_state_leq:
+    forall ell s1 s2 han1 ch1 han2 ch2,
+    node_state_low_eq ell (nodes s1) (nodes s2) ->
+    node_state_low_eq ell
+        (nodes (state_upd_chan han1 ch1 s1))
+        (nodes (state_upd_chan han2 ch2 s2)).
+Proof.
+Admitted. (* WIP // TODO *)
+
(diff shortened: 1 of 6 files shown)


experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -1,68 +1,148 @@
...
+        (chans (state_upd_chan han ch1 s1)) (chans (state_upd_chan han ch2 s2)).
+Proof.
+Admitted. (* WIP // TODO *)
+
+Theorem node_low_eq_reflexive:
+    forall ell n, node_low_eq ell n n.
+Proof.
+Admitted. (* WIP // TODO *)
+
+
```
jadephilipoom, line 88: nit: extra line break

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -1,68 +1,148 @@
...
+Theorem other_chan_exists_from_chans_leq: forall ell s1 s2 han ch1,
+    chan_state_low_eq ell (chans s1) (chans s2) ->
+    (chans s1).[? han] = Some ch1 ->
+    exists ch2, ((chans s2).[? han ] = Some ch2) /\
+        (chan_low_eq ell ch1 ch2).
+Proof.
+Admitted. (* WIP // TODO *)
+
+
```
jadephilipoom, line 97: nit: extra line break

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -80,124 +83,65 @@ Theorem possibilistic_ni_1step_node: forall ell id t1 t2 s1 s2 n1 n2 t1',
         (step_node_ev id (ncall n2) t2 t2') /\
         (trace_low_eq ell t1' t2')).
 Proof.
+    intros.
+    assert (Hsleq_s1s2: state_low_eq ell s1 s2) by
+        (eapply trace_leq_imples_head_st_leq; eassumption).
+    inversion H4; subst.
+    - (* WriteChannel *)
+        inversion H8. assert (s = s1) by congruence.
+        assert (n0 = n) by congruence. subst.
```
jadephilipoom, line 92:
```suggestion
        replace n0 with n by congruence.
```

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -80,124 +83,65 @@ Theorem possibilistic_ni_1step_node: forall ell id t1 t2 s1 s2 n1 n2 t1',
...
+        assert (n = n1) by congruence. subst.
```
jadephilipoom, line 93:
```suggestion
        replace n1 with n by congruence.
```

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -80,124 +83,65 @@ Theorem possibilistic_ni_1step_node: forall ell id t1 t2 s1 s2 n1 n2 t1',
         (step_node_ev id (ncall n2) t2 t2') /\
         (trace_low_eq ell t1' t2')).
 Proof.
+    intros.
+    assert (Hsleq_s1s2: state_low_eq ell s1 s2) by
+        (eapply trace_leq_imples_head_st_leq; eassumption).
+    inversion H4; subst.
+    - (* WriteChannel *)
+        inversion H8. assert (s = s1) by congruence.
```
jadephilipoom, line 91:
```suggestion
        inversion H8. replace s1 with s by congruence.
```

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -80,124 +83,65 @@ Theorem possibilistic_ni_1step_node: forall ell id t1 t2 s1 s2 n1 n2 t1',
...
+        --  (* the step happens *)
+            assert ((ncall n) = WriteChannel han msg) by congruence.
+            rewrite H9.
+            eapply SWriteChan; auto; try eassumption; try congruence.
+            rewrite Heqs2'. (* replace n' with (node_del_rhans (rhs msg) n ) in *. *)
+            rewrite Heqch2'.
+            eapply RuntimeModel.SWriteChan; auto; try eassumption; try congruence.
+            inversion Hchanleq; subst; congruence.
+        -- (* the traces are low-equivalent*)
+            constructor; try congruence.
```
jadephilipoom, line 123: If you're expecting `congruence` to leave a specific number of subgoals unsolved here, it can be nice to use the `[ | ... | ]` syntax to indicate how many. For instance, if you expect three remaining subgoals, you could write `constructor; try congruence; [ | | ]` and then you'd get a failure if there was a different number of goals remaining. It's easier to get an error early than getting one inside a bullet-point and having to figure out that this was a subgoal you expected `congruence` to dispatch. If you expect `congruence` to solve all but the last subgoal, you could use `[ congruence .. | ]` instead, which could have speed benefits because Coq won't even try to run `congruence` on the last goal.

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -0,0 +1,105 @@
...
+    intros. inversion H0. 
+    split.
+    - (* nodes *) 
+        intros id'. destruct (id' =P id). 
+        + (* eq *) 
+            rewrite e;
+            rewrite !state_upd_node_eq; assumption.
+        + (* neq *) 
+            rewrite !state_upd_node_neq.
+            apply (H1 id').  assumption. assumption.
```
jadephilipoom, line 57: nit:
```suggestion
            apply (H1 id'); assumption.
```

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: I like the way you wrote it better, but the proofs didn't work for me locally anymore, so I went back to the old way just for that reason. (I also tried replace ... in * by congruence). Similar situation for the others.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Neither of the bracket syntax suggestions seem to work for me. Are we using the same version of coq, I wonder? I'm on 8.12.0.

This looks useful anyway.

jadephilipoom: Maybe the subgoal you were expecting to leave unsolved isn't the last subgoal in the sequence produced by `constructor`? The bracket syntax goes `[ <tactic for subgoal 1> | <tactic for subgoal 2> | <tactic for subgoal 3> ]` for however many subgoals you have. Writing `..` within the `[ ]` works like `*` in a regular expression, so the `[ congruence .. | ]` example says "run congruence on every goal except the last one, leaving the last one untouched". (The `..` can be used anywhere in the bracket expression also; `[ congruence | .. | tauto ]` would run congruence on the first goal, `tauto` on the last one, and leave the rest untouched.)

Anyway, not an essential change for this piece of code, just a useful bit of syntax I wanted to mention :)

Issue #1492, opened 2020-09-16
Oak Chrome extension


tiziano88:
Tracking issue for a Chrome extension that would enforce that a web page is talking to an Oak Application, attaches labels to outgoing requests, and prevents non-Oak requests once a web page switches to Oak mode.

https://developer.chrome.com/extensions/webRequest

juliettepretot:
I'm curious, would this really need to be an extension proprietary to chrome? It seems to me that a [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) would be able to achieve most of these tasks.

tiziano88:
AFAICT workers are not a great fit here because:

- there is no UI associated with them, to signal to the user whether a page is *actually* in Oak mode (the extension in the future should make this clearer, e.g. by changing the icon color or something).
- workers need to be associated with a URL, but here the context is that of a tab; i.e. this is similar to "incognito mode", which is a per-tab setting, rather than a per-URL setting.
- it seems that workers are mostly meant to "help" by offloading some of the request handling to local code, but they don't seem to be aimed at capturing all of the outgoing traffic; the extension API seems more specifically geared towards making guarantees that all requests are intercepted.

juliettepretot:
@tiziano88 Thanks for the summary, appreciate it. :) 

- Absence of UI is a fair point. :) Though this would also be possible as a progressive enhancement in a service worker + extension combo. 
- Out of curiosity, why is per-tab behavior desired? This seems potentially confusing from a user perspective. (Also, how do the chrome extensions handle websites that sync state across tabs? Seems like potential way of leaking data)
- True that they are largely advertised as a performance improvement. Though they do serve as a network proxies, it seems quite plausible they could achieve the same purpose.

tiziano88:
pasting my own comment from https://project-oak.slack.com/archives/CHE9E13C3/p1600718595019500?thread_ts=1600691759.016400&cid=CHE9E13C3 :

[...]

PR #1493, opened 2020-09-17
Create Oak Chrome extension skeleton

tiziano88:
This is a very experimental placeholder, which exposes an API for tabs
to switch to "Oak private mode", allowing the extension to then
intercept and block requests.

Ref #1492

# Checklist

      construction.

### DIFF
--- a/chrome_extension/README.md
+++ b/chrome_extension/README.md
@@ -1,0 +1,44 @@
+# Oak Chrome Extension
+
+The Oak Chrome Extension is a proof of concept to show how Oak may be used as
+part of web applications.
+
+A regular web page may at any point enter Oak private mode, which means that
+from that point onwards it will be able to communicate exclusively with an Oak
+application that can enforce that any data it receives is kept private for the
+current session. This is similar to "incognito mode" but for the server-side
+parts of an application.
+
+A web page enters Oak private mode by invoking the following Javascript code:
+
+```javascript
+window.postMessage('oak_enter');
+```
+
+This is an irreversible process: at this point, the Oak extension takes full
+control of all the communication initiated by that particular tab, blocks any
+non-Oak requests (including any static assets, which the application should take
+care to load before entering Oak mode), attaches an appropriate label to any Oak
+requests, and authenticates against a remote Oak application.
+
+Note that there is no way of getting out of Oak mode for a tab; if there were,
+it may be used to leak potentially confidential data from the current tab. The
+only way out of Oak mode is to manually close the tab, and then create a new
+one. Again, this is similar to how it is not possible to take a Chrome incognito
+tab and turn it back into a regular tab.
+
+Note: for now, it seems that closing a tab and then re-opening the last closed
+tab actually re-creates a tab with the same URL but outside of Oak mode.
+
+## Installation
+
+In order to install the Chrome extension locally for development:
+
+- navigate to `chrome://extensions`
+- turn on "Developer mode"
+- click on the "Load unpacked" button
+- select the `chrome_extension` folder (the one containing the `manifest.json`
+  file) and click "Open"
+
+After the extension is installed, in order to reload it from disk, it is
+sufficient to click on the refresh arrow button next to the extension.
--- a/chrome_extension/background.js
+++ b/chrome_extension/background.js
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 The Project Oak Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+// TODO(#1492): Generate an ECDSA key pair, and use the private key for authentication and the
+// public key as Oak label.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign
+
+// Ids of tabs that have entered Oak mode.
+const enabledTabs = new Set();
+
+// Intercept outgoing requests, and block them if they are coming from a tab that has entered Oak
+// mode.
+//
+// TODO(#1492): This only intercepts new WebSocket connections, but not individual messages on
+// previously established connections. e.g. https://www.websocket.org/echo.html.
+//
+// TODO(#1492): This does not catch cases in which a tab opens another tab via JavaScript, e.g.
+// `window.open('https://google.com/?q=123');`.
+chrome.webRequest.onBeforeRequest.addListener(
+  (details) => {
+    const tabId = details.tabId;
+    const oakMode = enabledTabs.has(tabId);
+    console.log(
+      'request from tab ' + tabId + ': ' + (oakMode ? 'blocked' : 'allowed')
+    );
+    // For now, we just cancel all outgoing requests once a tab has entered Oak mode, which is not
+    // particularly meaningful (and in fact quite annoying, since it is irreversible). But this serves
+    // as a proof of concepts that we can intercept requests and block them with arbitrary logic.
+    //
+    // TODO(#1492): Instead of just cancelling all requests, we should allow Oak requests through,
+    // and attach authentication credentials and labels to them. For this, we need to determine
+    // which ones are legitimate Oak requests though.
+    return { cancel: oakMode };
+  },
+  // filters
+  { urls: ['<all_urls>'] },
+  // extraInfoSpec
+  ['blocking']
+);
+
+// Listen for messages from `content.js` that signal whether to enable Oak mode for a specific tab.
+chrome.runtime.onMessage.addListener((message, sender) => {
+  console.log('message received', message, sender);
+  if (message == 'oak_enter') {
+    const tabId = sender.tab.id;
+    console.log('entering Oak mode for tab ' + tabId);
+    enabledTabs.add(tabId);
+  }
+});
--- a/chrome_extension/content.js
+++ b/chrome_extension/content.js
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 The Project Oak Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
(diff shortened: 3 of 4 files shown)


tiziano88:
@juliettepretot dog

chrome_extension/README.md
```diff
@@ -0,0 +1,44 @@
+# Oak Chrome Extension
+
+The Oak Chrome Extension is a proof of concept to show how Oak may be used as
+part of web applications.
+
+A regular web page may at any point enter Oak private mode, which means that
+from that point onwards it will be able to communicate exclusively with an Oak
+application that can enforce that any data it receives is kept private for the
+current session. This is similar to "incognito mode" but for the server-side
+parts of an application.
```
ipetr0v, line 10: I think this extension will only be used on Oak application pages (that don't have any 3rd-party resources downloaded by the client, since we interrupt everything). So it may make sense not to make a `Private` button, and just invoke the extension automatically, if the Oak app page was opened.

chrome_extension/README.md
```diff
@@ -0,0 +1,44 @@
+# Oak Chrome Extension
```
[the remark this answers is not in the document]
juliettepretot: Maybe yes. Though I'm not sure we that many things to put in it yet. For now `chrome_extension` seems nice

chrome_extension/README.md
juliettepretot, line 10: Should this be a service worker instead? That's a standard web API which creates a proxy that intercepts all network request. This could similarly enforce exclusive communication. 

https://developers.google.com/web/ilt/pwa/introduction-to-service-worker

juliettepretot: Upsides are it works in all browsers & requires not separate installation.

juliettepretot: Afaik service workers intercept all network communication, hence could offer similar enforcement. Of course this would break if there's a way to bypass the service worker, which I haven't investigated. 

Curious if we considered going down this route? Of course it's possible to start with a chrome extension now (seems like a potentially safer bet, since Arcs etc use these) and consider refactoring later.

chrome_extension/README.md
```diff
@@ -0,0 +1,44 @@
...
+A regular web page may at any point enter Oak private mode, which means that
+from that point onwards it will be able to communicate exclusively with an Oak
+application that can enforce that any data it receives is kept private for the
+current session. This is similar to "incognito mode" but for the server-side
+parts of an application.
+
+A web page enters Oak private mode by invoking the following Javascript code:
+
+```js
+window.oak.enter();
```
juliettepretot, line 15: Note that this conceivably overlaps with websites that use this key already. Are there other planned properties for the global `oak` object? Would it make sense to use `window.postMessage("oak_enter")` instead?

chrome_extension/background.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+// TODO: Generate an ECDSA key pair, and use the private key for authentication and the public key
```
juliettepretot, line 3: I believe this TODO is missing an issue number

chrome_extension/background.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
```
juliettepretot, line 1: Linting rules are flagging this due to a lacking license header, seems we should add it?

chrome_extension/background.js
```diff
@@ -0,0 +1,44 @@
...
+//
+// Only contains entries for which the value is `true`.
+const tab_status = {};
+
+// Intercept outgoing requests, and block them if they are coming from a tab that has entered Oak
+// mode.
+chrome.webRequest.onBeforeRequest.addListener(
+  (details) => {
+    let tab_id = details.tabId;
+    let oak_mode = tab_status[tab_id];
```
juliettepretot, line 18: Nit but personally I'd make these constants

chrome_extension/background.js
```diff
@@ -0,0 +1,44 @@
...
+
+// TODO: Generate an ECDSA key pair, and use the private key for authentication and the public key
+// as Oak label.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign
+
+// Maps a tab id to a boolean signalling whether that tab has entered Oak mode.
+//
+// Only contains entries for which the value is `true`.
+const tab_status = {};
```
juliettepretot, line 11: Personally I'd make this a map. :) 

The reason being that when looking up properties on an object, JavaScript goes up all the way up the prototype chain. This is less performant, but more importantly can lead to unexpected results. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

Map also exposes a `has` [method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has), which is more predictable than reading an object property and evaluating it as boolean.

chrome_extension/content.js
```diff
@@ -0,0 +1,46 @@
...
+  false
+);
+
+// Injects a script in the context of the active tab, so that it can operate on the `window` object
+// of that tab.
+function inject(fn) {
+  const script = document.createElement('script');
+  script.text = `(${fn.toString()})();`;
+  document.documentElement.appendChild(script);
+}
```
juliettepretot, line 31: Can we use `chrome.tabs.executeScript` instead?
https://developer.chrome.com/extensions/tabs#method-executeScript

Injecting scripts like this could lead to unexpected consequences, by polluting the global JavaScript namespace of that page.

juliettepretot:
I assume people might be keen to use TypeScript instead of direct JS? Happy to help set it up

chrome_extension/README.md
tiziano88: thanks for this and for the comment on the issue too, I replied there so that the discussion is more easily discoverable and does not implode in a black hole whenever I rebase this PR:

https://github.com/project-oak/oak/issues/1492#issuecomment-694296909

chrome_extension/README.md
tiziano88: Fair enough, I was trying to make it simpler for web pages, but I think sticking to `postMessage` actually also avoids the need of extra weird injected scripts, so let's go with that.

chrome_extension/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+This is an irreversible process: at this point, the Oak extension takes full
+control of all the communication initiated by that particular tab, blocks any
+non-Oak requests (including any static assets, which the application should take
+care to load before entering Oak mode), attaches an appropriate label to any Oak
+requests, and authenticates against a remote Oak application.
```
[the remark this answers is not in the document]
tiziano88: Yes, though only new WS connections, not individual messages. I guess this may be a problem for streaming HTTP requests too.

chrome_extension/background.js
tiziano88: Done.

chrome_extension/background.js
tiziano88: Yes, others too, will add TODO numbers (and create issues) before submitting.

chrome_extension/background.js
tiziano88: Great, thanks!

chrome_extension/background.js
tiziano88: Done.

chrome_extension/content.js
tiziano88: Not needed any more.

tiziano88:
> I assume people might be keen to use TypeScript instead of direct JS? Happy to help set it up

Thanks, I thought about that too, though my hope is that this will stay quite minimal and easily inspectable, so perhaps let's see how far we can get with plain js for the time being

chrome_extension/background.js
```diff
@@ -0,0 +1,61 @@
...
+
+'use strict';
+
+// TODO: Generate an ECDSA key pair, and use the private key for authentication and the public key
+// as Oak label.
+//
+// See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign
+
+// Ids of tabs that have entered Oak mode.
+const enabled_tabs = new Set();
```
juliettepretot, line 25: nit: I will caution that camlCase is the most common naming convention in JavaScript land, but I'm not precious about it :)

juliettepretot (approved):

chrome_extension/background.js
```diff
@@ -0,0 +1,61 @@
...
+
+// Intercept outgoing requests, and block them if they are coming from a tab that has entered Oak
+// mode.
+//
+// TODO: This only intercepts new WebSocket connections, but not individual messages on previously
+// established connections. e.g. https://www.websocket.org/echo.html.
+chrome.webRequest.onBeforeRequest.addListener(
```
conradgrobler, line 32: Does this also catch content loading in child IFrames? Does this catch opening of new windows/tabs?

chrome_extension/background.js
tiziano88: ahh, I personally dislike camelCase, but since the google JS style guide also states it (https://google.github.io/styleguide/jsguide.html#naming-local-variable-names), I switched over to that.

chrome_extension/background.js
tiziano88: iframes: yes
new windows: no (I was able to exfiltrate data via 

```js
window.open('https://google.com/?q=123');
```

@juliettepretot do you think there is a any way of catching this?

ipetr0v (approved):

chrome_extension/background.js
juliettepretot: There's a CSP working draft to prevent this, but so far it's not implemented in any browser. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/navigate-to

`chrome.webNavigation.onBeforeNavigate` might be able to stop it, but I haven't looked into it: https://developer.chrome.com/extensions/webNavigation#event-onBeforeNavigate

juliettepretot: On a higher level I'm a bit unsure how practical the approach of treating all client code as untrusted is. In practice that means pretty heavy restrictions. Any oak website couldn't include external links, as it'd risk the same kind of data leak.

juliettepretot: And even if those are successfully imposed, data can still be leaked by other evil chrome extensions reading the Oak website's content.

rbehjati (approved):
This was interesting to look at. I guess there is going to be a UI (like a popup) in the future as well?

chrome_extension/background.js
tiziano88: yeah good point re: external links.

I think eventually we would need something more drastic, like incognito mode, in which even links that you open from incognito are themselves opened in more incognito tabs. But this is a good starting point I think, despite all the assumptions.

tiziano88:
> This was interesting to look at. I guess there is going to be a UI (like a popup) in the future as well?

yeah possibly, though for now I want this to be focused on "oak private mode" (label + auth token), so really there is only 1 bit of information that needs to be conveyed, so I am hoping doing something on the extension icon itself should be enough. But yes, if necessary, the chrome extension API does allow creating pop up dialogs from the extension icon too.

chrome_extension/background.js
tiziano88: FYI I made all the TODOs point to the same issue for now, I don't want to spam the issue tracker with this stuff while it's still so experimental, but happy to start splitting it out if more people start looking into it.

tiziano88:
@juliettepretot I noticed that prettier does not format comments, is there any way to do that automatically?

juliettepretot:
> @juliettepretot I noticed that prettier does not format comments, is there any way to do that automatically?

Been thinking about writing a plugin for that, prettier indeed does not offer it. There's an ESLint rule that does, but haven't looked into combining it with prettier. 

For now this VSCode extension might be easiest. :/ https://github.com/stkb/Rewrap

PR #1469, opened 2020-09-11
Retry sending requests in HTTP tests to make the tests stable

rbehjati:
Fixes #1462

PR #1494, opened 2020-09-17
Use the runtime to start the test HTTP server node

rbehjati:
Ref [comment on 1469](https://github.com/project-oak/oak/pull/1469#discussion_r489710316)

### DIFF
--- a/oak_runtime/src/node/http/tests.rs
+++ b/oak_runtime/src/node/http/tests.rs
@@ -16,7 +16,13 @@
 
 use super::*;
 use maplit::hashmap;
-use oak_abi::{label::Label, proto::oak::application::ApplicationConfiguration};
+use oak_abi::{
+    label::Label,
+    proto::oak::application::{
+        node_configuration::ConfigType, ApplicationConfiguration, HttpServerConfiguration,
+        NodeConfiguration,
+    },
+};
 use std::{fs, option::Option, thread::JoinHandle};
 
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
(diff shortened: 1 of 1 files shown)


oak_runtime/src/node/http/tests.rs
```diff
@@ -159,34 +145,35 @@ fn create_runtime() -> RuntimeProxy {
...
-            },
-            tls_config,
-        )
-        .expect("Could not create server node"),
-    )
+    crate::RuntimeProxy::create_runtime(&configuration, &secure_server_config, &signature_table)
+}
+
+fn create_server_node(runtime: &RuntimeProxy, port: u32) -> oak_abi::Handle {
+    let (init_receiver, invocation_receiver) = create_communication_channel(runtime);
```
tiziano88, line 164: Could this (or the underlying method) use the typed version of `channel_create` with actual types? Since below you are wrapping that anyways into a typed Receiver. BTW we should probably clarify that that is the preferred one, instead of the one that returns "raw" handles.

tiziano88 (approved):
Thanks!

oak_runtime/src/node/http/tests.rs
rbehjati: The one in the `sdk`? Can I use that one or is there another one in the Runtime?

oak_runtime/src/node/http/tests.rs
```diff
@@ -221,33 +208,32 @@ fn create_communication_channel(runtime: &RuntimeProxy) -> (oak_abi::Handle, oak
...
-            body: vec![],
-            status: http::status::StatusCode::OK.as_u16() as i32,
-            headers: hashmap! {},
-        };
-        invocation
-            .sender
-            .expect("Empty sender on invocation.")
-            .send(resp, runtime)
-            .unwrap();
+    loop {
```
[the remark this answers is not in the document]
rbehjati: Thanks, this is much nicer.

oak_runtime/src/node/http/tests.rs
tiziano88: Good point, I think it's missing from the Runtime, could you create an issue for it so we don't forget? Perhaps it could go in `oak_runtime::io`.

rbehjati: Created #1496. I won't add a TODO here, because there are many other places in the code that should be updated as well.

PR #1510, opened 2020-09-22
Move tactics into their own file and test them

jadephilipoom:
In preparation for doing more proof automation, I moved the more general-purpose tactics from `TraceTheorems.v` into their own file and, since they'll now be exported for use in other files, I added some tests for them. The tests aren't strictly necessary, but they can be a nice warning system if you change the expected behavior of a tactic that's used in multiple files, so I think they might come in handy.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/Tactics.v
+++ b/experimental/ni_coq/vfiles/Tactics.v
@@ -1,0 +1,94 @@
+(** Generically useful tactics **)
+
+(* inner loop of destruct_match *)
+Ltac destruct_match' e :=
+  lazymatch e with
+  | context [match ?x with _ => _ end] => destruct_match' x
+  | _ => destruct e
+  end.
+(* Finds the first match it encounters in the goal and destructs the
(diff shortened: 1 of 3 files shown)


experimental/ni_coq/vfiles/Tactics.v
```diff
@@ -0,0 +1,94 @@
...
+                          | O => n
+                          | S _ => n
+                          end) with
+                   | O => m
+                   | S _ => m
+                   end = m).
+  Proof.
+    intros n m.
+    destruct_match; [ | ]. (* expect 2 subgoals *)
+    (* we expect that the match on m, not n, was the one destructed,
```
aferr, line 46: I appreciate your thorough comments!

Also, it is nice to see how to write tests for tactics. I would not have known how to do that.

aferr:
The tests do seem like a good thing to have, and it does seem like a better way to organize the files.

aferr (approved):

Issue #1491, opened 2020-09-16
Add functionality to `oak_sign` to sign challenge for user authentication


tiziano88:
Related to #1357 .

`oak_sign` may be extended so that it may be used offline to sign a challenge (from command line).

In the future, ideally this logic would instead be made part of a client library for connecting to Oak applications, which would generate key pairs, and use them to sign a challenge obtained from a specific Oak Runtime instance. For this, I think it may make sense to turn `oak_sign` into a library (i.e. separating, exposing and documenting the relevant functions) so that it may be used both by the `oak_sign` binary itself, as well as by a Rust client library.

cc @rbehjati @ipetr0v

rbehjati:
I agree that this should be a library.

ipetr0v:
> oak_sign may be extended so that it may be used offline to sign a challenge (from command line).

It's already used from a command line:
https://github.com/project-oak/oak/blob/605c64a90d279af254fef41be9b24609962edb30/scripts/oak_sign#L7

Also, agree about making it a library. It will also PEM related code:
https://github.com/project-oak/oak/blob/caa91c86f4f2b9d451d7a8bae9b56d82375f4b3e/oak_runtime/src/lib.rs#L1301-L1305

tiziano88:
> > oak_sign may be extended so that it may be used offline to sign a challenge (from command line).
> 
> It's already used from a command line:
> https://github.com/project-oak/oak/blob/605c64a90d279af254fef41be9b24609962edb30/scripts/oak_sign#L7

Not for signing a challenge to be used for web auth though, right?

> Also, agree about making it a library. It will also PEM related code:
> https://github.com/project-oak/oak/blob/caa91c86f4f2b9d451d7a8bae9b56d82375f4b3e/oak_runtime/src/lib.rs#L1301-L1305

What about the PEM code (maybe you missed a word)?

ipetr0v:
`It will also deduplicate PEM related code`, updated the original comment

ipetr0v:
> Not for signing a challenge to be used for web auth though, right?

tiziano88:
> > Not for signing a challenge to be used for web auth though, right?
> 
> Yes, but we will not be able to run a console command from a web browser.

Definitely, but I was thinking for initial experiments we would just get the challenge from the server, run `oak_sign` manually via command line, passing the challenge, and paste the signature back in the browser.

[...]

Issue #1437, opened 2020-09-07
Sign Wasm module hashes


ipetr0v:
Follow up of a discussion started in https://github.com/project-oak/oak/pull/1387#discussion_r483963855

Currently Wasm modules are signed using [Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519).
And after https://github.com/project-oak/oak/pull/1387 we will also hash them with SHA-256 and store these hashes in corresponding `signature.pem` files (along with actual signatures and public keys). These hashes are necessary in order to understand to which Wasm module a signature corresponds to.

**Proposal**: We can as well just sign SHA-256 hashes instead of actual Wasm data.

Two counter-arguments against this:
- Ed25519 already uses SHA-512 for signing. So if we will start signing SHA-256 hashes with Ed25519 it may raise a lot of questions from the Oak application developers in the future.
- Hash that is stored in the file is used just as a reference, so when we want to check a Wasm file's signature, we compute its hash, and then find a corresponding file, that contains the same hash.

cc @tiziano88

PR #1512, opened 2020-09-22
Refactor oak_sign

ipetr0v:
This change:
- Refactors and removes code duplication
- Signs SHA-256 hashes instead of file contents
- Adds the ability to sign input strings

Fixes https://github.com/project-oak/oak/issues/1437
Fixes https://github.com/project-oak/oak/issues/1491

# Checklist


oak_sign/src/main.rs
```diff
@@ -124,96 +93,31 @@ fn main() -> anyhow::Result<()> {
...
-            write(&opt.signature, &encoded_signature)
-                .with_context(|| format!("Couldn't write signature file {}", &opt.signature))?;
+            let private_key = read_pem_file(&opt.private_key)?;
+            let input = match (&opt.input_file, &opt.input_string) {
+                (Some(_), Some(_)) => Err(anyhow!(
+                    "Both `input_file` and `input_string` are specified"
+                )),
+                (Some(input_file), None) => std::fs::read(&input_file)
+                    .with_context(|| format!("Couldn't read input file {}", &input_file)),
+                (None, Some(input_string)) => Ok(input_string.as_bytes().to_vec()),
```
conradgrobler, line 109: Should there be some validation on minimum length? Does the hasher accept very short or empty strings?

oak_sign/src/main.rs
```diff
@@ -80,8 +71,10 @@ struct Generate {
 struct Sign {
     #[structopt(long, help = "Input PEM encoded Ed25519 private key file path")]
     pub private_key: String,
-    #[structopt(long, help = "Input file path")]
-    pub input_file: String,
+    #[structopt(long, conflicts_with = "input_string", help = "Input file path")]
+    pub input_file: Option<String>,
+    #[structopt(long, conflicts_with = "input_file", help = "Input string")]
```
conradgrobler, line 76: Perhaps add a bit more detail in the help to make it clear how it would be used?

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub signature: Vec<u8>,
+    pub hash: Vec<u8>,
+}
+
+impl Signature {
+    pub fn create(input: &[u8], private_key: &[u8]) -> anyhow::Result<Signature> {
+        let key_pair = signature::Ed25519KeyPair::from_pkcs8(private_key)
+            .ok()
+            .context("Couldn't parse PKCS8 encoded private key")?;
+        let hash = get_sha256(input);
```
conradgrobler, line 70: Is there documentation somewhere that needs to be updated to highlight that this is effectively signing the hash of the hash of the input file?

conradgrobler (approved):

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+};
+
+// PEM file tags.
+pub const PRIVATE_KEY_TAG: &str = "PRIVATE KEY";
+pub const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
+pub const SIGNATURE_TAG: &str = "SIGNATURE";
+pub const HASH_TAG: &str = "HASH";
+
+/// Convenience struct for private and public key.
+pub struct KeyPairStruct {
```
tiziano88, line 35: Why the `Struct` suffix? Is it because `ring` already has a type with the same name? If so, I guess we could refer to that as the fully qualified name, I don't think adding `Struct` suffix on the local one is particularly meaningful (nor scalable).

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+        let public_key = key_pair.public_key();
+        Ok(KeyPairStruct {
+            private_key: private_key.as_ref().to_vec(),
+            public_key: public_key.as_ref().to_vec(),
+        })
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct Signature {
```
tiziano88, line 59: This instead may benefit from having a more precise name, especially since it has a field called `signature` itself. How about `SignatureBundle` or `SignaturePack` or something?

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub public_key: Vec<u8>,
+    pub signature: Vec<u8>,
+    pub hash: Vec<u8>,
+}
+
+impl Signature {
+    pub fn create(input: &[u8], private_key: &[u8]) -> anyhow::Result<Signature> {
```
tiziano88, line 66: Why is this accepting a byte slice for the private key? It seems more idiomatic to pass the acual parsed key pair in, which will be parsed separately by the caller, and errors handled there. The current code seems to unnecessarily couple the two.

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn from_pem_file(filename: &str) -> anyhow::Result<Signature> {
+        let file =
+            read(filename).with_context(|| format!("Couldn't read signature file {}", filename))?;
+        let file_content =
+            pem::parse_many(file)
+                .iter()
+                .fold(HashMap::new(), |mut content, entry| {
+                    content.insert(entry.tag.to_string(), entry.contents.to_vec());
+                    content
+                });
```
tiziano88, line 104: Couldn't this just be `.map(...).collect()`, as long as the `map` returns `(key, value)` pairs?

oak_sign/src/main.rs
```diff
@@ -124,96 +93,31 @@ fn main() -> anyhow::Result<()> {
...
-            let hash_pem = create_pem(HASH_TAG, hash_bytes.as_ref());
-            let encoded_signature = encode_many(&[public_key_pem, signature_pem, hash_pem]);
-
-            // Write signature file.
-            write(&opt.signature, &encoded_signature)
-                .with_context(|| format!("Couldn't write signature file {}", &opt.signature))?;
+            let private_key = read_pem_file(&opt.private_key)?;
+            let input = match (&opt.input_file, &opt.input_string) {
+                (Some(_), Some(_)) => Err(anyhow!(
+                    "Both `input_file` and `input_string` are specified"
```
tiziano88, line 105: I think it would be more useful if the message said what should be done, e.g. "Exactly one between `input_file` and `input_string` must be specified" (and below).

tiziano88 (approved):

rbehjati (approved):
This is very nice. I have only some minor comments.

oak_sign/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    let encoded_pem = pem::encode(&Pem {
+        tag: tag.to_string(),
+        contents: contents.to_vec(),
+    });
+    write(filename, encoded_pem).with_context(|| format!("Couldn't write file {}", filename))?;
+    Ok(())
+}
+
+/// Computes a SHA-256 digest of `input` and returns it in a form of raw bytes.
+pub fn get_sha256(input: &[u8]) -> Vec<u8> {
```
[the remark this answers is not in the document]
rbehjati: The Wasm module uses `get_sha256_hex`. Right?

oak_sign/src/main.rs
```diff
@@ -124,96 +99,38 @@ fn main() -> anyhow::Result<()> {
...
-            let hash_pem = create_pem(HASH_TAG, hash_bytes.as_ref());
-            let encoded_signature = encode_many(&[public_key_pem, signature_pem, hash_pem]);
-
-            // Write signature file.
-            write(&opt.signature, &encoded_signature)
-                .with_context(|| format!("Couldn't write signature file {}", &opt.signature))?;
+            let private_key = read_pem_file(&opt.private_key)?;
+            let key_pair = KeyBundle::parse(&private_key)?;
+            let input = match (&opt.input_file, &opt.input_string) {
+                (Some(_), Some(_)) => Err(anyhow!(
```
rbehjati, line 111: Not sure what is more idiomatic, but this arm and the last one can be merged into one arm at the end:
```
                _ => Err(anyhow!(
                    "Exactly one between `input_file` and `input_string` must be specified"
                )),
```

rbehjati (approved):

oak_sign/src/lib.rs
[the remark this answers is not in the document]
ipetr0v: I changed it in this PR

oak_sign/src/main.rs
ipetr0v: Done

Issue #1459, opened 2020-09-10
Use idiomatic types to represent HTTP requests in SDK


tiziano88:
e.g. https://docs.rs/http/0.2.1/http/request/struct.Request.html

cc @rbehjati

rbehjati:
Can you please be more specific?

tiziano88:
Currently, HTTP requests and responses are represented as a protobuf objects in the SDK:

https://github.com/project-oak/oak/blob/79de881763b06e83a00e80aeba740978ebd0949e/oak_services/proto/http_encap.proto#L21-L31

https://github.com/project-oak/oak/blob/79de881763b06e83a00e80aeba740978ebd0949e/oak_services/proto/http_encap.proto#L33-L41

For convenience / familiarity / compatibility with other libraries and frameworks, we should expose more idiomatic types for them instead, e.g.:

https://docs.rs/http/0.2.1/http/request/struct.Request.html

https://docs.rs/http/0.2.1/http/response/struct.Response.html

conradgrobler:
This seems like it could cause more confusion. Both of these types expose many things that we do not support (e.g. Request::connect, Request::get, Response::extensions, etc.).

It is also not clear to me how this would work with the current invocation pattern. How will these types be serialised?

rbehjati:
I agree with Conrad. I thought we needed protobuf messages here.

tiziano88:
Those methods seem fine to me, e.g. Request::connect creates a builder for a new request, but with a `CONNECT` method, and same for Request::get etc. They don't perform any operations by themselves, if that's what you meant.

Extensions may actually be useful at some point (e.g. to propagate labels in some more structured way), but for now we can just not use them.

The types will still be serialized as the existing protobuf messages (I'm not suggesting deleting those), but the SDK will convert between protobuf and idiomatic types.

This would make writing HTTP handling code (e.g. https://github.com/project-oak/oak/pull/1461/files) more approachable for developers already familiar with other Rust libraries and frameworks.

In case it is not clear, the `http` crate in Rust only defines the data types, not any HTTP related functionality.

conradgrobler:
I did not notice that the functions such as connect only create builders, so there are fewer unsupported features than I thought.

I don't think extensions would be useful for propagating Labels and we are unlikely to ever support it. If we implement this we should make sure that we raise appropriate errors when the client code uses unsupported features such as extensions.

tiziano88:
[...]

PR #1513, opened 2020-09-23
Allow headers with multiple values

### LINKED ISSUE #1458: HTTP headers should not be represented as a map

rbehjati:
Fixes #1458

# Checklist

### DIFF
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -19,7 +19,7 @@
 use log::{info, warn};
 use maplit::hashmap;
 use oak::{http::Invocation, Node, OakError};
-use oak_services::proto::oak::encap::HttpResponse;
+use oak_services::proto::oak::encap::{HeaderValue, HttpResponse};
 
 oak::entrypoint!(oak_main => |_in_channel| {
     oak::logger::init_default();
@@ -48,7 +48,7 @@ impl Node<Invocation> for StaticHttpServer {
                     body: include_bytes!("../static/index.html").to_vec(),
                     status: http::StatusCode::OK.as_u16() as i32,
                     headers: hashmap! {
-                        CONTENT_TYPE.as_str().to_string() => "text/html; charset=UTF-8".to_string().into_bytes(),
+                        CONTENT_TYPE.as_str().to_string() => HeaderValue { values: vec!["text/html; charset=UTF-8".to_string().into_bytes()] },
                     },
                 },
                 _ => HttpResponse {
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -529,25 +529,22 @@ impl OakHeaderMap {
         self.headers
             .iter()
             .flat_map(|(name, value)| {
-                value
-                    .values
-                    .iter()
-                    .map(move |val| {
+                value.values.iter().filter_map(move |val| {
+                    let name_value_pair = || -> Result<(http::header::HeaderName, http::header::HeaderValue), OakStatus>  {
                         let header_name = http::header::HeaderName::from_bytes(name.as_bytes())
                             .map_err(|err| {
                                 warn!("Error when parsing header name: {}", err);
-                                err
-                            });
+                                OakStatus::ErrInternal
+                            })?;
                         let header_value =
                             http::header::HeaderValue::from_bytes(val).map_err(|err| {
                                 warn!("Error when parsing header value: {}", err);
-                                err
-                            });
-                        (header_name, header_value)
-                    })
-                    .filter(|(name, value)| name.is_ok() && value.is_ok())
-                    // can safely call unwrap since the `Error`s are filtered out
-                    .map(|(name, value)| (name.unwrap(), value.unwrap()))
+                                OakStatus::ErrInternal
+                            })?;
+                        Ok((header_name, header_value))
+                    };
+                    name_value_pair().ok()
+                })
             })
             .collect()
     }
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -57,7 +57,7 @@ use tokio_rustls::TlsAcceptor;
 #[cfg(test)]
 pub mod tests;
 
-// Checks that prot is not reserved (i.e., is greater than 1023).
+/// Checks that port is not reserved (i.e., is greater than 1023).
 fn check_port(address: &SocketAddr) -> Result<(), ConfigurationError> {
     if address.port() > 1023 {
         Ok(())
@@ -91,7 +91,7 @@ impl hyper::server::accept::Accept for TlsServer<'_> {
             // be awoken when there is a new incoming connection, by calling `wake` on
             // the `waker` in the context `cx`.
             Poll::Ready(Some(Err(e))) => {
-                log::error!("Error when processing TLS stream: {:?}", e);
+                error!("Error when processing TLS stream: {:?}", e);
                 Poll::Pending
             }
             _ => connection,
@@ -125,7 +125,7 @@ impl HttpServerNode {
         })
     }
 
-    // Make a server, with graceful shutdown, from the given [`HttpRequestHandler`].
+    /// Make a server, with graceful shutdown, from the given [`HttpRequestHandler`].
     async fn make_server(
         &self,
         request_handler: HttpRequestHandler,
@@ -176,9 +176,9 @@ impl HttpServerNode {
                 io::Error::new(io::ErrorKind::Other, format!("Incoming failed: {:?}", err))
             })
             .and_then(move |stream| {
-                log::debug!("Received incoming TLS stream: {:?}", stream);
+                debug!("Received incoming TLS stream: {:?}", stream);
                 tls_acceptor.accept(stream).map_err(|err| {
-                    log::error!("Client-connection error: {:?}", err);
+                    error!("Client-connection error: {:?}", err);
                     io::Error::new(io::ErrorKind::Other, format!("TLS Error: {:?}", err))
                 })
             })
@@ -279,7 +279,7 @@ struct HttpRequestHandler {
 
 impl HttpRequestHandler {
     async fn handle(&self, req: Request<Body>) -> Result<Response<Body>, OakStatus> {
-        let request = to_oak_http_request(req).await;
+        let request = to_oak_http_request(req).await?;
         match get_oak_label(&request) {
             Ok(oak_label) => {
                 info!(
@@ -343,7 +343,7 @@ impl HttpRequestHandler {
     }
 }
 
-// A pair of temporary channels to pass the HTTP request and to receive the response.
+/// A pair of temporary channels to pass the HTTP request and to receive the response.
 struct Pipe {
     request_writer: oak_abi::Handle,
     request_reader: oak_abi::Handle,
@@ -433,9 +433,9 @@ impl Pipe {
     }
 }
 
-// HTTP requests can either provide JSON formatted labels or protobuf encoded labels. But exactly
-// one of these should be provided. This method checks that exactly one label is provided in a
-// header in the request and extracts it for use for further handling of the request.
+/// HTTP requests can either provide JSON formatted labels or protobuf encoded labels. But exactly
+/// one of these should be provided. This method checks that exactly one label is provided in a
+/// header in the request and extracts it for use for further handling of the request.
 fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
     let headers = (
         req.headers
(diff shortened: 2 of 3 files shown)


oak_services/proto/http_encap.proto
```diff
@@ -27,15 +27,22 @@ message HttpRequest {
   // The body of the request.
   bytes body = 3;
   // The HTTP request headers.
-  map<string, bytes> headers = 4;
+  map<string, HeaderValue> headers = 4;
```
tiziano88, line 30: Would it be simpler to just have a `Header` message that encapsulates name and value, and may be repeated? That seems to map more closely to HTTP semantics, in which repeated headers with the same name are not really related to each other. Also it seems simpler to create and pass around, and would presumably map nicely to something like https://docs.rs/hyper/0.13.8/hyper/header/struct.HeaderMap.html#impl-FromIterator%3C(HeaderName%2C%20T)%3E or other multimap-style collections in Rust. WDYT?

tiziano88 (commented):
one high level comment before I look into more details

oak_services/proto/http_encap.proto
rbehjati: We are mostly converting the headers to hyper headers. For that I don't see a huge difference between the two types. But we also directly lookup in the headers in `get_oak_label`. For that it is easier to have a map. As far as I know, multimaps are just wrappers around Maps. But I don't think `prost` will create a multimap for `repeated Header`, so in my opinion `map<string, HeaderValue>` is a more suitable type.

juliettepretot:
I'm a little bit confused as for why this is need. According to the HTTP specs it seems that headers can only be repeated in very specific circumstances?

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

Couldn't we just combine multiple header fields if they occur, and not concern ourselves with encoding them into a proto message as repeated headers?

juliettepretot (commented):
see prev comment

rbehjati:
> Couldn't we just combine multiple header fields if they occur, and not concern ourselves with encoding them into a proto message as repeated headers?

I think it is better to be consistent with `http::header::HeaderMap`, and that seems to allow multiple values for headers. The solution you suggested should work, but I am not sure if that would be easier, or if there are special cases that we'd have to consider.

tiziano88:
I agree @juliettepretot , this was also partly my point earlier; I think the most common use case is to have one value per key, and modelling headers as a list of (key, value) pairs makes that easy enough; if multiple headers with the same key are provided, then we can decide what to do with them (combine them in a multimap or whatever). And having a repeated field with (key, value) seems more straightforward to me than a map with a nested message. But anyways, since we are going to abstract this away with proper types from the `http` crate (at least in Rust), any of these options seem fine.

oak_runtime/src/node/http/mod.rs
```diff
@@ -49,13 +50,14 @@ use futures_util::{
...
 };
+use maplit::hashmap;
 use prost::Message;
 use tokio_rustls::TlsAcceptor;
 
 #[cfg(test)]
 pub mod tests;
 
-/// Checks that prot is not reserved (i.e., is greater than 1023).
+// Checks that prot is not reserved (i.e., is greater than 1023).
```
juliettepretot, line 60: nit (and precedes this PR) but
```suggestion
// Checks that port is not reserved (i.e., is greater than 1023).
```

oak_runtime/src/node/http/mod.rs
```diff
@@ -466,13 +438,17 @@ impl Pipe {
...
 fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
     let headers = (
-        req.headers.get(oak_abi::OAK_LABEL_HTTP_JSON_KEY),
-        req.headers.get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY),
+        req.headers
+            .get(oak_abi::OAK_LABEL_HTTP_JSON_KEY)
+            .map(|m| m.values.as_slice()),
+        req.headers
+            .get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY)
+            .map(|m| m.values.as_slice()),
```
tiziano88, line 446: On the other hand, I guess this logic will become a bit more complicated if we allowed individual (key, value) headers, since we would need to iterate through all the headers to figure out how many of each there are.

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+                    };
+
+                    (item.0.to_owned(), header_value)
+                })
+            })
+            .collect()
+    }
+}
+
+impl From<&http::header::HeaderMap<http::header::HeaderValue>> for OakHeaderMap {
```
tiziano88, line 548: Is the reference necessary in the generic type? I would have thought that some blanket impl takes care of it, and this impl may be declared without refs, though maybe it only works for `Copy` types.

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+    fn from(hyper_headers: &http::header::HeaderMap<http::header::HeaderValue>) -> Self {
+        #![allow(clippy::mutable_key_type)]
+        let mut headers = hashmap!();
```
tiziano88, line 551:
```suggestion
        let mut headers = hashmap!{};
```

juliettepretot (approved):

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    fn to_vec(&self) -> Vec<(String, http::header::HeaderValue)> {
+        self.headers
+            .iter()
+            .flat_map(|item| {
+                item.1.values.iter().map(move |val| {
+                    let header_value = match http::header::HeaderValue::from_bytes(val) {
+                        Ok(header_value) => header_value,
+                        Err(err) => {
+                            log::warn!("Error when mapping response: {}", err);
```
tiziano88, line 536:
```suggestion
                            log::warn!("Error when parsing header value: {}", err);
```

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+}
+
+impl OakHeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    fn to_vec(&self) -> Vec<(String, http::header::HeaderValue)> {
+        self.headers
+            .iter()
+            .flat_map(|item| {
+                item.1.values.iter().map(move |val| {
```
tiziano88, line 532: I'm curious, why is the `move` necessary? Does it cause the converted value to reuse the underlying bytes buffer this way, instead of copying it?

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+impl From<&http::header::HeaderMap<http::header::HeaderValue>> for OakHeaderMap {
+    fn from(hyper_headers: &http::header::HeaderMap<http::header::HeaderValue>) -> Self {
+        #![allow(clippy::mutable_key_type)]
+        let mut headers = hashmap!();
+        for (header_name, header_value) in hyper_headers {
+            headers
+                .entry(header_name)
+                .or_insert_with(Vec::new)
+                .push(header_value.as_bytes().to_vec())
+        }
```
tiziano88, line 557: I think this can be simplified by relying on https://docs.rs/http/0.2.1/http/header/struct.HeaderMap.html#impl-FromIterator%3C(HeaderName%2C%20T)%3E .

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+        let headers = headers
+            .iter()
+            .map(|item| {
+                (
+                    item.0.to_string(),
+                    HeaderValue {
+                        values: item.1.to_owned(),
+                    },
+                )
+            })
```
tiziano88, line 567: Would something like this work?

```suggestion
            .map(|(name, values)| {
                (
                    name,
                    HeaderValue {
                        values: values.to_owned(),
                    },
                )
            })
```

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+    headers: HashMap<String, HeaderValue>,
+}
+
+impl OakHeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    fn to_vec(&self) -> Vec<(String, http::header::HeaderValue)> {
+        self.headers
+            .iter()
+            .flat_map(|item| {
```
tiziano88, line 531:
```suggestion
            .flat_map(|(name, values)| {
```

etc.

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,122 @@ impl HttpResponseIterator {
...
+}
+
+// Create an instance of Oak HttpRequest from the given hyper Request.
+async fn to_oak_http_request(req: Request<Body>) -> HttpRequest {
+    let uri = req.uri().to_string();
+    let method = req.method().as_str().to_string();
+    let headers = OakHeaderMap::from(req.headers()).headers;
+    let body = hyper::body::to_bytes(req.into_body())
+        .await
+        .expect("Error when reading request body from the connection")
```
tiziano88, line 586: Just checking: is it fine to panic if this happens? Can it not be triggered by a malformed request or something like that?

examples/http_server/module/src/lib.rs
```diff
@@ -48,7 +48,7 @@ impl Node<Invocation> for StaticHttpServer {
                     body: include_bytes!("../static/index.html").to_vec(),
                     status: http::StatusCode::OK.as_u16() as i32,
                     headers: hashmap! {
-                        CONTENT_TYPE.as_str().to_string() => "text/html; charset=UTF-8".to_string().into_bytes(),
+                        CONTENT_TYPE.as_str().to_string() => HeaderValue { values: vec!["text/html; charset=UTF-8".to_string().into_bytes()] },
```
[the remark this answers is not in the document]
juliettepretot: agreed :)

oak_runtime/src/node/http/mod.rs
juliettepretot: It should be fine for these two headers though, since neither of them is a comma seperated list, and therefore cannot be repeated.

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,131 @@ impl HttpResponseIterator {
...
+                        let header_value =
+                            http::header::HeaderValue::from_bytes(val).map_err(|err| {
+                                warn!("Error when parsing header value: {}", err);
+                                err
+                            });
+                        (header_name, header_value)
+                    })
+                    .filter(|(name, value)| name.is_ok() && value.is_ok())
+                    // can safely call unwrap since the `Error`s are filtered out
+                    .map(|(name, value)| (name.unwrap(), value.unwrap()))
```
tiziano88, line 550: I think this can be a single `filter_map` https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map in which the function is something like `|(name, value)| Some(name?, value?)`, assuming `?` works in lambdas, which I am not sure.

oak_runtime/src/node/http/mod.rs
```diff
@@ -518,21 +494,131 @@ impl HttpResponseIterator {
...
+    }
+
+    builder
+        .status(http_response.status as u16)
+        .body(Body::from(http_response.body))
+        .map_err(|err| {
+            warn!("Could not build response: {}", err);
+            OakStatus::ErrInternal
+        })
+}
```
tiziano88, line 624: Would https://doc.rust-lang.org/std/convert/trait.TryFrom.html or something similar help here?

tiziano88 (approved):

oak_runtime/src/node/http/mod.rs
tiziano88: Semantically they are equivalent, but usually in macros `[]` are used for sequences, `()` for fn calls, and `{}` for sets and blocks.

oak_runtime/src/node/http/mod.rs
rbehjati: Thanks. Done.

oak_runtime/src/node/http/mod.rs
rbehjati: This has to be done in `oak_services`. I'll do this with #1459.

oak_runtime/src/node/http/mod.rs
rbehjati: Thanks :)

PR #1518, opened 2020-09-24
Add oak_* libraries to doc generation

ipetr0v:
This change adds multiple `oak_*` libraries to doc generation.

# Checklist


tiziano88 (approved):

ipetr0v:
The PR is not ready yet.
In order to add `oak_io` to docs, we need to change:
```
[`oak::io::Sender`](../io/struct.Sender.html)
```
to just `oak::io::Sender` in:
https://github.com/project-oak/oak/blob/6b3ca281f2b45f6453016880c8792c8e37d956a6/proto/handle.proto#L29-L34
Because it tries to resolve relative paths while being compiled by different crates.

tiziano88:
Thanks, that seems like a reasonable change to me in fact, right?

conradgrobler (approved):

proto/handle.proto
```diff
@@ -28,14 +28,14 @@ extend google.protobuf.FieldOptions {
 
 // Sender handle for an Oak channel.
 // This type is sent over the wire, but in types generated by prost it is replaced with
-// [`oak::io::Sender`](../io/struct.Sender.html).
+// `oak::io::Sender`.
```
tiziano88, line 31: So I thought the whole point was to make this a standard link like ``[`oak::io::Sender`]``, why does that not work now?

tiziano88 (approved):

proto/handle.proto
ipetr0v: It cannot resolve it:
```
 --> /opt/my-project/oak_io/target/debug/build/oak_io-4bc20a416904bf4a/out/oak.handle.rs:3:6
  |
3 | /// [`oak::io::Sender`].
  |      ^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
```

conradgrobler: `oak_io` doesn't reference the SDK, so it cannot be resolved even though the SDK re-exports the types. In `oak_io` it should be able to resolve something like `crate::Sender` or `super::Sender`. I don't know whether it would cause problems in the cases where handle.proto is imported in other proto files compiled in other crates.

ipetr0v: `crate::Sender` works fine, but there are problems, if we want to use `GrpcInvocation` in one of the examples:
```
Found invalid urls in /opt/my-project/examples/target/doc/trusted_database_handler/proto/oak/invocation/index.html:
	Linked file at path /opt/my-project/examples/target/doc/trusted_database_handler/proto/oak/invocation/oak_io::Receiver does not exist!
	Linked file at path /opt/my-project/examples/target/doc/trusted_database_handler/proto/oak/invocation/oak_io::Sender does not exist!
```

PR #1520, opened 2020-09-25
Restructure NI defs to use low projections. NI Proofs using low proj steps.

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/EvAugSemantics.v
+++ b/experimental/ni_coq/vfiles/EvAugSemantics.v
@@ -46,6 +46,8 @@ so the call is a piece of state that probably does not matter at the moment)
  and a downgrade event.
 *)
 
+Definition trace := list (state * event_l).
+
 Declare Scope aug_scope.
 Local Open Scope aug_scope.
 Notation "n '--->' msg":= (EvL (OutEv msg) n.(nlbl)) (at level 10) : aug_scope.
@@ -59,37 +61,32 @@ Definition head_st (t: trace) :=
         | (s', _)::_ => Some s'
     end.
 
-Inductive step_node_ev (id: node_id): call -> trace -> trace -> Prop :=
-    | SWriteChan (t: trace) s n han msg s' t:
-        head_st t = Some s ->
+Inductive step_node_ev (id: node_id): call -> state -> state -> event_l -> Prop :=
+    | SWriteChan s n han msg s':
         s.(nodes) .[?id] = Some n ->
         step_node id (WriteChannel han msg) s s' ->
-        step_node_ev id (WriteChannel han msg) t
-            ((s',  n ---> msg) :: t)
-    | SReadChan (t: trace) s n han chan msg s' t:
-        head_st t = Some s ->
+        step_node_ev id (WriteChannel han msg) s s' (n ---> msg)
+    | SReadChan s n han chan msg s':
         s.(nodes) .[?id] = Some n ->
         step_node id (ReadChannel han) s s' ->
         msg_is_head chan msg ->
-        step_node_ev id (ReadChannel han) t ((s', n <--- msg) :: t)
-    | SCreateChan (t: trace) s n lbl s' t:
+        step_node_ev id (ReadChannel han) s s' (n <--- msg)
+    | SCreateChan s n lbl s':
             (* It seems clear that no event is needed since nodes only observe
             * contents of channels indirectly via reads *)
-        head_st t = Some s ->
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateChannel lbl) s s' ->
-        step_node_ev id (CreateChannel lbl) t ((s', n --- ) :: t)
-    | SCreateNode (t: trace) s n lbl h s' t:
+        step_node_ev id (CreateChannel lbl) s s' (n --- )
+    | SCreateNode s n lbl h s':
             (* model observation that a node is created ?? *)
-        head_st t = Some s ->
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateNode lbl h) s s' ->
-        step_node_ev id (CreateNode lbl h) t ((s', n ---) :: t)
-    | SInternal (t: trace) s n s' t:
-        head_st t = Some s ->
+        step_node_ev id (CreateNode lbl h) s s' (n ---)
+    | SInternal s n s':
         s.(nodes) .[?id] = Some n ->
         step_node id Internal s s' ->
-        step_node_ev id Internal t ((s', n ---) :: t).
+        step_node_ev id Internal s s' (n ---).
+
 
 Definition trace_upd_head_state (t: trace) (s: state) :=
     match t with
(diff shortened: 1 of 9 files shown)


aferr:
The main point of this change is to both to make the security definition provable (while still ensuring security) and also to restructure the proofs. The main change is to:
-  move away from a statement of the theorem that uses a low-equivalence relation directly and instead use one based on low projections.
- write a set of lemmas about taking steps to/from the low projections of states, and use these to prove the single step theorem.
- the system-wide step function is also extended with a system-wide NOP 

As side-effects:
- These changes are fairly sweeping, so many of the old theorems are invalidated
- The definitions of step functions in EvAugSemantics are rewritten to be clearer/cleaner/better-separated:
        - The single-node function is now (nid -> call -> state -> state -> event) rather than (nid -> call -> trace -> trace)
        which makes it closer to the step function that does not emit events
        - A separate relation builds traces using the individual steps on states. 

The heart of the problem with the previous definition and proof structure is in the theorem about a single step of noninterference for an individual node. An unusual aspect of this proof compared to one for a programming language is that for a language, the single step theorem would compare two executions (each with different values for secrets) for a single particular command that is used in both executions, but in this case the command executed by a node might be a secret itself. The fact that the command is a secret shows up in the old low-equivalence definition: when the label of the node does not flow to the observer, it relates any two nodes with the same label (but possibly differing commands). This presents a problem in the old single-step theorem (possibilistic_ni_1step_node) which we were attempting to do by cases on the command in the first execution, and then by cases on whether the label of the node that was executed in the first exectution flowsTo the observer or not. In the flowsTo case, the node referred to by the chosen node id is the equal in both executions. But in the not-flows-to case, we know nothing about what the node looks like in the other execution (its contents are a secret). So the command executed by the node could be anything, even an invalid command that cannot take a step. So this old theorem with the old security definitions is not true.

This PR attempts to fix this problem by using low projections in the definition of security and a new proof structure. The low projections are functions of type: {A: Type} -> label -> A -> A, that removes all the secrets (above the label) from the thing of type A. Examples of things are nodes, channels, and events. The new proof structure no longer has the possibilistic_ni_1step_node theorem, but instead uses a new set of theorems about steps to/from low projections over states (step_implies_lowproj_steps_leq, low_proj_steps_implies_leq_step) in the proof for possibilistic_ni_1step (which is about a system-wide step rather than a single-node step). The (new) proof for the lemma possibilistic_ni_1step is complete (aside from necessitating some obviously true lemmas about transitivity of low-equivalences).

An intuition as to how this (should) fix rather than just move the problem is that now the potential issue involving unobservable steps  might show up in the low_proj_steps_implies_leq_step lemma. However, when the step is a command run by a secret node, the low-projection will: (1) only modify secret state, and (2) only emit NOP-events (the low projection of a secret event is a NOP). So for execution 2, we can choose the state and event that results from a system-wide NOP because this should have equivalent behavior as the secret step.

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -59,6 +61,34 @@ Definition head_st (t: trace) :=
         | (s', _)::_ => Some s'
     end.
 
+Inductive step_node_ev (id: node_id): call -> state -> state -> event_l -> Prop :=
```
jadephilipoom, line 64: This change seems like an excellent idea -- it's more intuitive with the single-step based on state-steps instead of trace-steps, and more intuitive usually means easier proofs! (Except, of course, when it doesn't.)

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -90,6 +120,7 @@ Inductive step_node_ev (id: node_id): call -> trace -> trace -> Prop :=
         s.(nodes) .[?id] = Some n ->
         step_node id Internal s s' ->
         step_node_ev id Internal t ((s', n ---) :: t).
+*)
```
jadephilipoom, line 123: Did you mean to leave this commented out, or to delete it?

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -104,21 +135,30 @@ Definition head_set_call t id c: trace :=
         | (s, e) :: t' => (s_set_call s id c, e) :: t'
     end.
 
-Inductive step_system_ev: trace -> trace -> Prop :=
-    | ValidStep id n c c' s t s' t':
-        head_st t  = Some s ->
-        head_st t' = Some s' ->
+Inductive step_system_ev: state -> state -> event_l -> Prop :=
+    | SystemStep s ell: step_system_ev s s (EvL NilEv ell)
```
jadephilipoom, line 139: Kind of a nitpick, but it looks like this is basically a skip -- why call it `SystemStep` instead of something like `SystemSkip` (matching the runtime model as well)?

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -104,21 +135,30 @@ Definition head_set_call t id c: trace :=
...
 Inductive step_system_ev_multi: trace -> trace -> Prop :=
     | multi_system_ev_refl t t':
-        step_system_ev t t' ->
+        step_system_ev_t t t' ->
         step_system_ev_multi t t'
     | multi_system_ev_tran t1 t2 t3:
-        step_system_ev t2 t3 ->
+        step_system_ev_t t2 t3 ->
         step_system_ev_multi t1 t2 ->
         step_system_ev_multi t1 t3.
```
jadephilipoom, line 163: This is possibly a larger refactor than you want to go for, but I think now that the previous single-step definitions are defined on states you could explicitly construct traces using lists in a single inductive, using an initial state. To be more concrete, I'm thinking something like:
```
Inductive TracePath : state -> trace -> Prop :=
| EmptyTrace : forall (init : state), TracePath init []
| StepTrace :
  forall (init s s' : state) (e : event) (t : trace),
    TracePath init t ->
    head_st t = Some s ->
    step_system_ev s s' e ->
    TracePath init ((s', e) :: t)
```
This would allow for easy induction on traces, because if you do induction on the trace and you have a `TracePath _ (_ :: _)` in context, doing inversion on it will give you only the `cons` case.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -73,6 +73,42 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+    (chan_state_low_eq ell s1.(chans) s2.(chans)) ->
+    (state_low_eq ell s1 s2).
+Proof.
+Admitted. (* WIP *)
+
+Theorem eq_nodes_have_eq_lbls: forall n1 n2,
+    n1 = n2 -> (nlbl n1) = (nlbl n2).
+    congruence.
+Proof.
+Qed.
```
jadephilipoom, line 93: It seems unlikely that this bit will compile! Do you mean `Admitted`?

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -73,6 +73,42 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+    - inversion H1. congruence.
+Qed.
+
+(* TODO: possibly specializing this for the FlowsTo case 
+    where (nlbl n1 ) flowsTo ell is still useful with the new proof structure.
+    This might get used in the proofs for either of the theorems about stepping
+    with low projections
+*)
+(*
 Theorem possibilistic_ni_1step_node: forall ell id t1 t2 s1 s2 n1 n2 t1',
```
jadephilipoom, line 112: Is this supposed to be commented out or deleted? I tend to prefer deleting old versions of code and pulling them from git history if I need to restore them, just to avoid having to remove it in unrelated PRs later.

jadephilipoom (approved):
LGTM! I can't claim to completely understand the IFC implications, but from a proof-structure standpoint I think this is likely to help the possibilistic NI proofs.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: Ooops. I meant to delete it. Thanks for catching this.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: Oops. I meant to call it something like skip. It's fixed and I also gave the other constructor a better name, as well as disambiguating these constructors from the similar ones in RuntimeModel.

experimental/ni_coq/vfiles/EvAugSemantics.v
aferr: I do like this idea. I think the make-or-break next steps are probably in the single step parts of the proof. So I think I will leave the definitions for constructing traces as-is for now, make progress on the single step bits, then probably look at refactoring the trace definitions as you suggested.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Oddly, the the compiler is perfectly happy with "congruence. Proof. Qed.", though I think I wish it wasn't. It's fixed now.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Good point. It's deleted now.

PR #1521, opened 2020-09-25
Coq CI

jadephilipoom:
This adds a Github Action that runs the `ni_coq` `make`, compiling all the proofs. The job will trigger only on:`
- changes to files ending in `.v`
- changes to files whose names end in `_CoqProject`
- changes to `experimental/ni_coq/Makefile`

With those restrictions, it shouldn't run if anything other than the Coq code changes (and as you can probably see from the many commit messages, I tested this quite a bit on my fork!).

The action I found to do the Coq build requires a `.opam` file so it can try to install the project as a local opam package, so I added a minimalistic but roughly accurate one for the `ni_coq` proofs. I currently set the CI to test on Coq 8.11 and 8.12, but we can change that to just 8.11 or expand it to include the dev version as well.

# Checklist

      construction.

### DIFF
--- a/.github/workflows/coq.yaml
+++ b/.github/workflows/coq.yaml
@@ -4,17 +4,17 @@ on:
   push:
     branches: [main]
     paths:
-      - "**.v"
-      - "**coq.yaml"
-      - "**_CoqProject"
-      - "experimental/ni_coq/Makefile"
+      - '**.v'
+      - '**coq.yaml'
+      - '**_CoqProject'
+      - 'experimental/ni_coq/Makefile'
   pull_request:
     branches: [main]
     paths:
-      - "**.v"
-      - "**coq.yaml"
-      - "**_CoqProject"
-      - "experimental/ni_coq/Makefile"
+      - '**.v'
+      - '**coq.yaml'
+      - '**_CoqProject'
+      - 'experimental/ni_coq/Makefile'
 
 jobs:
   coq:
--- a/.github/workflows/coq.yaml
+++ b/.github/workflows/coq.yaml
@@ -2,7 +2,7 @@ name: Compile Coq proofs
 
 on:
   push:
-    branches: [main, coq-ci] # TODO: remove coq-ci; this is just for debugging 
+    branches: [main]
     paths:
       - "**.v"
       - "**coq.yaml"
--- a/.github/workflows/coq.yaml
+++ b/.github/workflows/coq.yaml
@@ -13,10 +13,10 @@ on:
 jobs:
   coq:
     strategy:
-    matrix:
-      coq_version:
-        - '8.11'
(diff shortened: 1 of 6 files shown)


experimental/ni_coq/oak-ni-coq.opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "aferr@google.com"
+authors:"Google LLC and contributors"
+synopsis: "Non-interference proofs for Project Oak"
```
aferr, line 4: Nit: noninterference (with no hyphen) is the more common spelling in the literature. (I realize the wikipedia article does hyphenate it). The [original paper](https://www.cs.purdue.edu/homes/ninghui/readings/AccessControl/goguen_meseguer_82.pdf) also does not spell it with a hyphen.

experimental/ni_coq/oak-ni-coq.opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "aferr@google.com"
```
aferr, line 2: Jade, do you want to also be listed as a maintainer? (Only if you want).

aferr (approved):

daviddrysdale (approved):
Thanks for setting this up!

(BTW, looks like there's some layout thing that needs fixing in `coq.yaml` to make `prettier` happy)

.github/workflows/coq.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - "**_CoqProject"
+      - "experimental/ni_coq/Makefile"
+
+jobs:
+  coq:
+    strategy:
+      matrix:
+        coq_version:
+          - '8.11'
+          - '8.12'
```
tiziano88, line 25: Out of curiosity, why not just choose one version?

tiziano88 (approved):

experimental/ni_coq/oak-ni-coq.opam
jadephilipoom: All the opam packages I looked at as examples seem to have only one name (I'm not sure if this is a rule or just convention). I put yours since you're the main project architect, but if you don't want to be on the hook for opam maintenance we can change it. Since this opam package configuration is just a dummy for the build system, I think no one is likely to receive maintenance requests!

.github/workflows/coq.yaml
jadephilipoom: Coq versions change pretty frequently, so I think it's useful to have a warning if the next upgrade is going to break something. The README claims 8.11 is supported, so we definitely need a check for that, but 8.12 is the latest stable version and we'll probably have to upgrade to it eventually, so a check for that version also seemed helpful.

PR #1527, opened 2020-09-28
Nflows case for low_proj_steps_implies_leq_step

aferr:
This includes a proof of low_proj_steps_implies_leq_step for the case in
which a NodeStep is taken, and the particular node executing the step has a
label that is not visible to the observer. This was the case that seemed most
likely to determine whether or not the proof strategy will work
out, so the proof suggests it will.

Some new admitted theorems are introduced.

Some cleanup related to the placement of theorems was done.

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/lib.rs
+++ b/oak_runtime/src/lib.rs
@@ -1093,9 +1093,17 @@ impl Runtime {
         Ok(match result {
             None => None,
             Some(ReadStatus::NeedsCapacity(z, c)) => Some(NodeReadStatus::NeedsCapacity(z, c)),
-            Some(ReadStatus::Success(msg)) => Some(NodeReadStatus::Success(
-                self.node_message_from(msg, node_id),
-            )),
+            Some(ReadStatus::Success(msg)) => {
+                let message = self.node_message_from(msg, node_id);
+
+                self.introspection_event(EventDetails::MessageDequeued(MessageDequeued {
+                    node_id: node_id.0,
+                    channel_id: half.get_channel_id(),
+                    acquired_handles: message.handles.clone(),
+                }));
+
+                Some(NodeReadStatus::Success(message))
+            }
         })
     }
 


experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
-        (chan_low_eq ell ch1 ch2).
+Theorem state_loweq_to_deref_node: forall ell s1 s2 id n1,
+    (nodes s1).[? id] = Some n1 ->
+    (state_low_eq ell s1 s2) ->
+    exists n2,
+        (nodes s2).[? id] = Some n2 /\
+        (node_low_eq ell n1 n2).
 Proof.
-Admitted. (* WIP // TODO *)
+Admitted. (* WIP. For this one I think we need functional extensionality *)
```
aferr, line 132: Do I need functional extensionality here or is there another way. If I need it, how can I use FE safely?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
+    id' <> id ->
+    (state_upd_node id n s).(nodes).[?id'] = s.(nodes).[?id'].
+Proof.
+    intros. eapply upd_neq. congruence.
 Qed.
 
-Theorem flows_uncons_proj: forall ell n,
+(*---------------------------------------------------------------------------*)
+(* Low Projections *)
+(*---------------------------------------------------------------------------*)
```
jadephilipoom, line 35: Not essential to this PR, but have you considered using `Section` to organize related code instead of banner comments?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
     reflexivity.
 Qed.
 
-Theorem state_leq_and_flowsto_to_node_eq: forall ell s1 s2 id n1 n2,
-    (nodes s1).[? id] = Some n1 ->
-    (nodes s2).[? id] = Some n2 ->
-    (state_low_eq ell s1 s2) -> 
-    (nlbl n1 <<L ell) ->
-    n1 = n2.
+Theorem state_low_proj_idempotent: forall ell s,
```
jadephilipoom, line 52: Naming nitpick: this isn't exactly idempotence, but `state_low_proj_idempotent'` is -- maybe rename the latter to `state_low_proj_idempotent` without the `'` and call this one `state_low_proj_low_eq`?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
 Proof.
-Admitted. (* WIP // TODO *)
+    intros. unfold state_low_proj in *. cbn in H0.  unfold node_state_low_proj in H0.
+    destruct (nodes s id) eqn: E in H0. 
+        - (* some *) 
+        replace n0 with n in *.
+        erewrite nflows_node_proj in H0. inversion H0. auto. auto.
+        replace (nodes s id) with ((nodes s).[? id]) in E by auto.
+            (* why on earth do I need this, it's just notation ? *)
+            (* Do I need hint rewrites or something ? *)
```
jadephilipoom, line 98: It could be that there's an implicit type you're converting here. If you do the replace without `by auto` and do `Set Printing All` on the subgoal that it leaves, do you see *exactly* the same thing on either side of `=`?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
-Theorem leq_node_updates_preserve_node_state_leq:
-    forall ell s1 s2 n1 n2 id,
-    node_state_low_eq ell (nodes s1) (nodes s2) ->
-    node_low_eq ell n1 n2 ->
-    node_state_low_eq ell
-        (nodes (state_upd_node id n1 s1)) (nodes (state_upd_node id n2 s2)).
+Definition symmetric {A: Type} (r: A -> A -> Prop) := 
+    forall a1 a2, (r a1 a2) -> (r a2 a1).
+
+Theorem state_low_eq_trans: forall ell, transitive (state_low_eq ell).
```
jadephilipoom, line 113: If you use the definitions from standard library [RelationClasses](https://coq.inria.fr/library/Coq.Classes.RelationClasses.html), you can use tactics like `reflexivity` and `transitivity` on goals with these definitions. It's pretty nifty.
```suggestion
Global Instance state_low_eq_trans: Transitive (state_low_eq) | 10.
```
It'll compile without the `| 10`, but I've been told it's a good idea to always put that there for global RelationClasses instances so that when Coq knows it's looking for a proof that a relation has one of the relation properties but doesn't know what the relation is, it'll try `eq` first. For instance, if the goal is `?R x x`, Coq should try to resolve it with `eq_refl`, not an instance of `Reflexive state_low_eq`.

experimental/ni_coq/vfiles/NIUtilTheorems.v
jadephilipoom: What makes you say you think you need it here? Are you running into a particular subgoal inside the WIP version that seems to require it?

The safest way to use FE is to prove it individually for every definition over which you require it (e.g. Theorem `foo_fun_ext : forall f g, (forall x, f x = g x) -> foo f = foo g`). You can also import an axiom for it, but that gets complicated theoretically; there's a reason it's not included as one of Coq's standard axioms, and you have to be careful not to compromise soundness.

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: Actually, on second thought I don't think FE make sense here. Though I do have a question for you about proving this theorem. At some point I have premises like:

H: (nodes s1).[? id] = Some n1
H2: (fun id : node_id =>
      LETOPT n <== nodes s1 id IN Some (node_low_proj ell n)) =
     (fun id : node_id =>
      LETOPT n <== nodes s2 id IN Some (node_low_proj ell n))

and I need to show

exists n2 : node, (nodes s2).[? id] = Some n2 /\ node_low_eq ell n1 n2

Is there a way to apply the particular node_id called "id" to both sides of H2?

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: That sounds very handy! What does `| 10` mean? Is it like the numbered annotations that go with Notation sometimes?

I think it needs to be like 
```
Global Instance state_low_eq_trans: forall ell, Transitive (state_low_eq ell) | 10.
```
because the extra argument needs to be applied before it's a relation. Given that low_eq's are A -> B -> B -> Prop instead of just B -> B -> Prop, do we need to do anything special to take advantage of the builtin tactics? Or is writing it like the above enough? Also, is it any better if we change the "ell's" in these definitions to be implicit so that it seems more like a B -> B -> Prop most of the time?

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: Done now.

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -104,24 +104,28 @@ Qed.
...
 
-Theorem event_low_eq_trans: forall ell, transitive (event_low_eq ell).
+Global Instance event_low_eq_refl: forall ell, Reflexive (event_low_eq ell) | 10.
 Proof.
 Admitted. (* WIP *)
 
+Global Instance event_low_eq_trans: forall ell, Transitive (event_low_eq ell) | 10.
+Admitted. (* WIP *)
+
+Global Instance event_low_eq_sym: forall ell, Symmetric (event_low_eq ell) | 10.
```
jieungkim, line 128: For the curiosity, do you have a specific reason to use "Global Instance" instead of "Theorem" in here?

aferr: @jadephilipoom  ? Maybe this is needed for compatibility with the builtin tactics like reflexivity and transitivity ?

experimental/ni_coq/vfiles/GenericMap.v
```diff
@@ -83,8 +83,7 @@ Notation "x '|->' v" := (pg_update pg_empty x v)
 * deal with laziness of me *)
 Notation "m '.[' k '<-' v ']'" := (pg_update m k v)
   (at level 100) : map_scope.
-Definition fnd {KT: KeyT}{VT: Type} (m: (pg_map KT VT)) k :=  (m k).
-Notation "m '.[?' k ']'" := (fnd m k) 
+Notation "m '.[?' k ']'" := (m k)
```
aferr, line 86: Hmmmm.... actually this change seems to be a horrible idea because then coq will use this notation absoulutely everywhere a function application happens. Having the extra "fnd" function gets around that.

I'll undo this in the next PR (unless this turns out to be too annoying).

experimental/ni_coq/vfiles/NIUtilTheorems.v
jieungkim: I guess... my point is quite similar to Jade. Is there a reason that you do not use 

"forall n, LETOPT n <== nodes s1 id IN Some (node_low_proj ell n))  = LETOPT n <== nodes s2 id IN Some (node_low_proj ell n))" instead of using lambda in the goal directly?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
-    (state_upd_node id n s).(nodes).[? id] = Some n.
+Theorem node_low_eq_to_lbl_eq: forall ell n1 n2,
+    (node_low_eq ell n1 n2) ->
+    (n1.(nlbl) = n2.(nlbl)).
 Proof.
-    intros. eapply upd_eq.
+    intros. inversion H.
+    assert ( nlbl (node_low_proj ell n1) = nlbl (node_low_proj ell n2)).
+    rewrite H1. reflexivity.
+    rewrite node_projection_preserves_lbl in H0.
```
jieungkim, line 147: rewrite !node_projection_preserves_lbl in H0.

will do the same thing with those two lines.

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -13,118 +13,155 @@ Import RecordSetNotations.
...
+    inversion 3. 
+    - 
+        exfalso. rewrite <- H3 in H. inversion H.
+    - 
+        assert (xs = s1). {
+            assert (head_st ((xs, xe) :: t0 ) = Some xs).
+            reflexivity. congruence.
+        }
+        assert (ys = s2). {
+            assert (head_st ((ys, ye) :: t3 ) = Some ys).
```
jieungkim, line 167: assert (head_st ((ys, ye) :: t3 ) = Some ys) by reflexivity.
            congruence.

It's a matter of preference in my sense though...

experimental/ni_coq/vfiles/NIUtilTheorems.v
jieungkim: Oh. I saw the comments that mentioned about this point. Thanks for the clarification!.

experimental/ni_coq/vfiles/NIUtilTheorems.v
jadephilipoom: I think writing it the way you showed above, with `ell` forall-quantified and plugged in, should be enough.

Here's the relevant Coq manual link for `Instance`, which talks about `| num`: https://coq.inria.fr/distrib/current/refman/addendum/type-classes.html#coq:cmd.instance

Essentially, the number is the "priority", which determines which instances `auto` and `eauto` try first when solving a goal. The default is 0, which is the highest priority, but then it would take precedence over the transitive property of `eq`, which you probably want to use first to avoid having automation insert `low_eq` where it doesn't belong.

experimental/ni_coq/vfiles/NIUtilTheorems.v
jadephilipoom: Yep, you're exactly right that it's for the builtin tactics!

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: I think I don't quite follow. Do you mean why is the premise H2 not closer to what you wrote? If so, it comes from the definition of node_state_low_eq which is in LowEquivalences.v. (It changed recently, so it's possible you saw an old definition of node_state_low_eq).

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: Done now

experimental/ni_coq/vfiles/NIUtilTheorems.v
aferr: Done now

experimental/ni_coq/vfiles/NIUtilTheorems.v
jadephilipoom: > Is there a way to apply the particular node_id called "id" to both sides of H2?

You could always do something like
```coq
match type of H2 with
?f = ?g => assert (f id = g id) by (rewrite H2; reflexivity)
end.
```

jieungkim: @aferr Thanks for the clarification.

PR #1528, opened 2020-09-28
Use idiomatic HTTP types in the SDK

### LINKED ISSUE #1459: Use idiomatic types to represent HTTP requests in SDK

rbehjati:
Fixes #1459 

# Checklist

### DIFF
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -17,9 +17,7 @@
 //! Simple example starting an Oak Application serving a static page over HTTP.
 
 use log::{info, warn};
-use maplit::hashmap;
-use oak::{http::Invocation, Node, OakError};
-use oak_services::proto::oak::encap::{HeaderValue, HttpResponse};
+use oak::{http::Invocation, Node, OakError, OakStatus};
 
 oak::entrypoint!(oak_main => |_in_channel| {
     oak::logger::init_default();
@@ -37,35 +35,21 @@ impl Node<Invocation> for StaticHttpServer {
     fn handle_command(&mut self, invocation: Invocation) -> Result<(), OakError> {
         let request = invocation.receive()?;
 
-        info!("Handling a request to {}.", request.uri);
+        info!("Handling a request to {}.", request.uri());
 
-        // Workaround for https://rust-lang.github.io/rust-clippy/master/index.html#borrow_interior_mutable_const.
-        static CONTENT_TYPE: http::header::HeaderName = http::header::CONTENT_TYPE;
-
-        let response = match request.uri.parse::<http::Uri>() {
-            Ok(uri) => match uri.path().as_ref() {
-                "/" => HttpResponse {
-                    body: include_bytes!("../static/index.html").to_vec(),
-                    status: http::StatusCode::OK.as_u16() as i32,
-                    headers: hashmap! {
-                        CONTENT_TYPE.as_str().to_string() => HeaderValue { values: vec!["text/html; charset=UTF-8".to_string().into_bytes()] },
-                    },
-                },
-                _ => HttpResponse {
-                    body: "not found".to_string().into_bytes(),
-                    status: http::StatusCode::NOT_FOUND.as_u16() as i32,
-                    headers: hashmap! {},
-                },
-            },
-            Err(err) => {
-                warn!("could not parse URI: {}", err);
-                HttpResponse {
-                    body: "invalid URI".to_string().into_bytes(),
-                    status: http::StatusCode::BAD_REQUEST.as_u16() as i32,
-                    headers: hashmap! {},
-                }
-            }
+        let response = match request.uri().path().as_ref() {
+            "/" => http::response::Builder::new()
+                .status(http::StatusCode::OK)
+                .header(http::header::CONTENT_TYPE, "text/html; charset=UTF-8")
+                .body(include_bytes!("../static/index.html").to_vec()),
+            _ => http::response::Builder::new()
+                .status(http::StatusCode::NOT_FOUND)
+                .body("not found".to_string().into_bytes()),
         };
+        let response = response.map_err(|err| {
+            warn!("Could not build response: {}", err);
+            OakError::OakStatus(OakStatus::ErrInternal)
+        })?;
 
         let res = invocation.send(&response);
         invocation.close_channels();
--- a/oak_runtime/src/node/http/mod.rs
+++ b/oak_runtime/src/node/http/mod.rs
@@ -38,8 +38,8 @@ use oak_io::{
     handle::{ReadHandle, WriteHandle},
     OakError, Receiver, Sender,
 };
-use oak_services::proto::oak::encap::{HeaderValue, HttpRequest, HttpResponse};
-use std::{collections::HashMap, io, net::SocketAddr, pin::Pin};
+use oak_services::proto::oak::encap::{HeaderMap, HttpRequest, HttpResponse};
+use std::{io, net::SocketAddr, pin::Pin};
 use tokio::{
     net::{TcpListener, TcpStream},
     sync::oneshot,
@@ -50,7 +50,6 @@ use futures_util::{
     future::TryFutureExt,
     stream::{StreamExt, TryStreamExt},
 };
-use maplit::hashmap;
 use prost::Message;
 use tokio_rustls::TlsAcceptor;
 
@@ -299,7 +298,7 @@ impl HttpRequestHandler {
                         OakStatus::ErrInternal
                     })?;
 
-                response.to_hyper_response()
+                response.try_into_hyper_response()
             }
             Err(OakStatus::ErrInvalidArgs) => http::response::Builder::new()
                 .status(StatusCode::BAD_REQUEST)
@@ -438,12 +437,16 @@ impl Pipe {
 /// header in the request and extracts it for use for further handling of the request.
 fn get_oak_label(req: &HttpRequest) -> Result<Label, OakStatus> {
     let headers = (
-        req.headers
-            .get(oak_abi::OAK_LABEL_HTTP_JSON_KEY)
-            .map(|m| m.values.as_slice()),
-        req.headers
-            .get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY)
-            .map(|m| m.values.as_slice()),
+        req.headers.as_ref().and_then(|map| {
+            map.headers
+                .get(oak_abi::OAK_LABEL_HTTP_JSON_KEY)
+                .map(|m| m.values.as_slice())
+        }),
+        req.headers.as_ref().and_then(|map| {
+            map.headers
+                .get(oak_abi::OAK_LABEL_HTTP_PROTOBUF_KEY)
+                .map(|m| m.values.as_slice())
+        }),
     );
 
     match headers {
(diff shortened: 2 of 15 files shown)


examples/http_server/module/src/lib.rs
```diff
@@ -37,37 +35,26 @@ impl Node<Invocation> for StaticHttpServer {
...
-                }
-            }
+        info!("Handling a request to {}.", request.uri());
+
+        let response = match request.uri().path().as_ref() {
+            "/" => http::response::Builder::new()
+                .status(http::StatusCode::OK)
+                .header(
+                    http::header::CONTENT_TYPE,
+                    "text/html; charset=UTF-8".to_string().into_bytes(),
```
tiziano88, line 45: Is this conversion still necessary? https://docs.rs/http/0.2.1/http/response/struct.Builder.html#examples-3 the examples seem to suggest it is not.

examples/http_server/module/src/lib.rs
```diff
@@ -37,37 +35,26 @@ impl Node<Invocation> for StaticHttpServer {
...
+                .status(http::StatusCode::NOT_FOUND)
+                .body("not found".to_string().into_bytes()),
         };
+        let response = response.map_err(|err| {
+            warn!("Could not build response: {}", err);
+            OakError::OakStatus(OakStatus::ErrInternal)
+        })?;
 
-        let res = invocation.send(&response);
+        let res = invocation.send(response);
```
tiziano88, line 57: Is it expected that the invocation would consume response by value? I would think a reference is sufficient.

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+        let mut builder = Request::builder()
+            .method(self.method.as_bytes())
+            .uri(self.uri);
+        if let Some(headers) = self.headers {
+            for (header_name, header_value) in headers.to_vec() {
+                builder = builder.header(header_name, header_value);
+            }
+        }
+        builder.body(self.body).map_err(|err| {
+            warn!("Could not create body from vector: {}", err);
```
tiziano88, line 47: I don't think an error here means there is a problem with the body, in fact most likely it will be in the headers, which need to follow specific rules and escaping etc. So I think the message should be about creating the request, not just the body.

Also see https://docs.rs/http/0.2.1/http/response/struct.Builder.html#method.body:

> This function may return an error if any previously configured argument failed to parse or get converted to the internal representation. For example if an invalid head was specified via header("Foo", "Bar\r\n") the error will be returned when this function is called rather than when header was called.

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+            OakStatus::ErrInternal
+        })
+    }
+}
+
+impl From<http::header::HeaderMap<http::header::HeaderValue>> for HeaderMap {
+    fn from(http_headers: http::header::HeaderMap<http::header::HeaderValue>) -> Self {
+        #![allow(clippy::mutable_key_type)]
```
tiziano88, line 55: Should we do something about this, or is it fine to keep ignoring it?

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+            .collect();
+
+        HeaderMap { headers }
+    }
+}
+
+impl HeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    pub fn to_vec(&self) -> Vec<(http::header::HeaderName, http::header::HeaderValue)> {
```
tiziano88, line 82: I'm guessing this may as well return an `impl Iterator`, and avoid the `collect` at the end, so that everything is still lazily evaluated and may be optimized based on what the caller actually needs?

sdk/rust/oak/src/http/mod.rs
```diff
@@ -21,24 +21,35 @@ use crate::{
...
 };
+use http::{Request, Response};
 use log::warn;
-pub use oak_services::proto::oak::encap::{HttpRequest, HttpResponse};
+pub use oak_services::proto::oak::encap::{HeaderMap, HttpRequest, HttpResponse};
+use std::convert::TryInto;
 
 pub type Invocation = crate::proto::oak::invocation::HttpInvocation;
 
+// This implementation provides an interface for sending requests and receiving responses, using the
```
tiziano88, line 31: Doc comment?

sdk/rust/oak/src/lib.rs
```diff
@@ -17,6 +17,8 @@
 //! SDK functionality that provides idiomatic Rust wrappers around the
 //! underlying Oak platform functionality.
 
+#![feature(result_flattening)]
```
tiziano88, line 20: Interesting, I didn't know about this feature, is it necessary because of some of the new code?

tiziano88 (approved):
Looks great, thanks!

conradgrobler (approved):

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn from(response: Response<Vec<u8>>) -> Self {
+        HttpResponse {
+            body: response.body().to_owned(),
+            status: response.status().as_u16() as i32,
+            headers: Some(HeaderMap::from(response.headers().to_owned())),
+        }
+    }
+}
+
+// Generally directly implementing `TryInto` is discouraged. But we cannot implement `TryFrom` for
```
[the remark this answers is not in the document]
rbehjati: I expected it to be disallowed, due to the rule that says that you cannot implement external traits on external types. But implementing `TryFrom` for `Response` works!

examples/http_server/module/src/lib.rs
tiziano88: The most generic is perhaps `Request<&[u8]>` though, right? i.e. an owned request whose body is a reference to some bytes. Anyways either way is fine, in practice we don't do anything else with the request nor the body after this, so even the pass by value version is ok.

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+            .collect();
+
+        HeaderMap { headers }
+    }
+}
+
+impl HeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    pub fn to_vec(
```
tiziano88, line 79: Probably should rename the method now

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+            })
+            .collect();
+
+        HeaderMap { headers }
+    }
+}
+
+impl HeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
+    // be easily iterated over when populating the headers in the hyper Response builder.
```
tiziano88, line 78: Doc comment.

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+                )
+            })
+            .collect();
+
+        HeaderMap { headers }
+    }
+}
+
+impl HeaderMap {
+    // Flatten the headers into a vector of tuples (String, http::header::HeaderValue) that can
```
tiziano88, line 77: Not a vector any more.

oak_services/src/http/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    // be easily iterated over when populating the headers in the hyper Response builder.
+    pub fn to_vec(
+        &self,
+    ) -> impl Iterator<Item = (http::header::HeaderName, http::header::HeaderValue)> + '_ {
```
tiziano88, line 81: Maybe this should just be an `IntoIter` impl: https://doc.rust-lang.org/std/iter/trait.IntoIterator.html (note that consumes the original value though). up to you.

oak_runtime/src/node/http/mod.rs
```diff
@@ -439,11 +438,21 @@ impl Pipe {
...
         req.headers
-            .get(oak_abi::OAK_LABEL_HTTP_JSON_KEY)
-            .map(|m| m.values.as_slice()),
+            .as_ref()
+            .map(|map| {
+                map.headers
+                    .get(oak_abi::OAK_LABEL_HTTP_JSON_KEY)
+                    .map(|m| m.values.as_slice())
+            })
+            .flatten(),
```
tiziano88, line 447: Would it work to use `filter_map` on line 442, and avoid this flattening here?

tiziano88 (approved):

oak_services/src/http/mod.rs
rbehjati: Thanks. Fixed.

oak_services/src/http/mod.rs
rbehjati: Done! :)

oak_services/src/http/mod.rs
rbehjati: Renamed to `iter`.

oak_services/src/http/mod.rs
rbehjati: I like the idea, but I could not get it to work. I can try again in a separate PR.

oak_runtime/src/node/http/mod.rs
rbehjati: `Option` does not have a `filter_map`. (my earlier answer to the question about `#![feature(result_flattening)]` was misleading. Please see my updated response :))

sdk/rust/oak/src/lib.rs
rbehjati: Sorry, I was sloppy in my answer! This is needed for the `flatten` in the `Invocation::receive` in `sdk/rust/oak/src/http/mod.rs`.

sdk/rust/oak/src/http/mod.rs
```diff
@@ -21,24 +21,35 @@ use crate::{
...
+        let request = self
+            .receiver
             .as_ref()
             .ok_or(OakError::OakStatus(OakStatus::ErrBadHandle))?
-            .receive()
+            .receive();
+
+        request
+            .map(|req| req.try_into().map_err(crate::OakError::OakStatus))
+            .flatten()
```
tiziano88, line 44: FYI, if `filter_map` works here, it is usually more idiomatic.

oak_runtime/src/node/http/mod.rs
tiziano88: Sorry I meant `flat_map`, and anyways I thought it was an Iter. For Option, `and_then` should work, and is more idiomatic.

rbehjati: Thanks.`and_then` works.

sdk/rust/oak/src/http/mod.rs
rbehjati: I think only `Iter` has `flat_map`, I cannot find `flat_map` for `Result`.

tiziano88: `flat_map` for Result and Option is called `and_then`

tiziano88: In case you are interested, the reason that all those functions "feel" related is that they are actually are instances of Monads, and `flat_map` / `and_then` is the "bind" operator. See http://learnyouahaskell.com/a-fistful-of-monads .

rbehjati: Oh! That helps. Thanks :)

examples/http_server/module/src/lib.rs
rbehjati: I'll keep `From<&Response<Vec<u8>>>` for now.

PR #1530, opened 2020-09-28
Add async channel stream

wildarch:
# Checklist

      construction.

Adds a `ChannelReadStream` type implementing [`Stream`](https://docs.rs/futures/0.3.5/futures/stream/trait.Stream.html) to the async sdk.

This PR builds on top of #1466.

For more background, please see the [RFC#01391](
https://github.com/project-oak/oak/blob/main/rfcs/01391_async_sdk.md).

### DIFF
--- a/experimental/oak_async/src/io.rs
+++ b/experimental/oak_async/src/io.rs
@@ -23,6 +23,7 @@ use core::{
     pin::Pin,
     task::{Context, Poll},
 };
+use futures::stream::Stream;
 use log::debug;
 use oak::{
     io::{Decodable, Message},
@@ -37,7 +38,7 @@ pub struct ChannelRead<T: Decodable> {
 }
 
 impl<T: Decodable> ChannelRead<T> {
-    pub(crate) fn new(handle: ReadHandle) -> ChannelRead<T> {
+    fn new(handle: ReadHandle) -> ChannelRead<T> {
         ChannelRead {
             reader_id: with_executor(|e| e.new_id()),
             handle,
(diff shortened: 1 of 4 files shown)


wildarch:
@tiziano88 and @daviddrysdale when reviewing, please note that only the last commit contains new code compared to what you have already seen in #1466. Thanks! 😄

experimental/oak_async/src/io.rs
```diff
@@ -72,6 +73,32 @@ impl<T: Decodable> Drop for ChannelRead<T> {
     }
 }
 
+/// `Stream` representing a sequence of asynchronous reads from a channel.
+pub struct ChannelReadStream<T: Decodable>(
+    // Note: `Stream` could be implemented directly on the `ChannelRead` type. Unfortunately the
+    // `Future` and `Stream` extension traits have some methods that overlap, such as `map`. This
```
tiziano88, line 79: Why not just directly impl `Stream` for `ChannelRead`, and forget about `Future`?

wildarch: I expect that reading a single message from a channel (and expecting one to be sent) will be a fairly common pattern that I think is a bit clumsy if we just have a `Stream`. There are things like `next` and `into_future`, but neither of these are convenient for this purpose.

When we make async the default I think it would be reasonable to make `Receiver` implement `Stream` directly, keeping `receive_async` around to have a convenient method for a `Future`. I am wondering if there could ever be a case where 'conflating' a receiver and `Stream` into one type could lead to issues, let me know if you can think of anything. It seems like they could become one and the same.

Having separate types might also help with type safety, especially since we have methods in gRPC that can be streaming or non-streaming. Having two different types could help express that better.

experimental/oak_async/src/io.rs
```diff
@@ -72,6 +73,32 @@ impl<T: Decodable> Drop for ChannelRead<T> {
     }
 }
 
+/// `Stream` representing a sequence of asynchronous reads from a channel.
+pub struct ChannelReadStream<T: Decodable>(
+    // Note: `Stream` could be implemented directly on the `ChannelRead` type. Unfortunately the
```
rbehjati, line 78:
```suggestion
    // Note: `Stream` could not be implemented directly on the `ChannelRead` type. Unfortunately the
```

experimental/oak_async/src/io.rs
```diff
@@ -100,4 +132,8 @@ impl<T: Decodable + Send> ReceiverAsync for oak::io::Receiver<T> {
     fn receive_async(&self) -> ChannelRead<Self::Message> {
         ChannelRead::new(self.handle)
     }
+
+    fn receive_stream(&self) -> ChannelReadStream<Self::Message> {
```
rbehjati, line 136: Could this be a default implementation in `ReceiverAsync`?

experimental/oak_async/tests/tests.rs
```diff
@@ -285,6 +288,63 @@ fn many_reads_parallel() {
     assert_eq!(data.len(), 1000);
 }
 
+// Stream functionality relies heavily on the implementation of `Future`, so we only need minimal
+// tests for the `Stream` impl.
+#[test]
+fn stream_immediate() {
```
rbehjati, line 294: Perhaps give this a more descriptive name if possible.

rbehjati (approved):
LGTM. I have only a few minor comments/questions.

experimental/oak_async/src/io.rs
wildarch: Aah, the phrasing here is a bit confusing. This is possible, it would just be very un-ergonomic to use. I will try to explain that a bit better.

experimental/oak_async/src/io.rs
wildarch: Yes! Thanks 😄

experimental/oak_async/tests/tests.rs
wildarch: Done.

PR #1534, opened 2020-09-29
Add a flag for rerunning server for each client

ipetr0v:
This change adds a `rerun_server_for_each_client` flag for rerunning server for each client.

# Checklist
  - [ ] I have checked that these tests are run by [Cloudbuild](/cloudbuild.yaml)

### DIFF
--- a/examples/private_set_intersection/example.toml
+++ b/examples/private_set_intersection/example.toml
@@ -1,4 +1,5 @@
 name = "private_set_intersection"
+rerun_server_for_each_client = true
 
 [applications]
 
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -553,6 +553,12 @@ fn run_example_server(
     )
 }
 
+/// Assigns `false` as a default for values deserialized with `serde`.
+/// https://serde.rs/attr-default.html
+fn default_bool() -> bool {
+    false
+}
+
 #[derive(serde::Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 struct Example {
@@ -563,6 +569,8 @@ struct Example {
     backend: Option<Executable>,
     applications: HashMap<String, Application>,
     clients: HashMap<String, Executable>,
+    #[serde(default = "default_bool")]
+    rerun_server_for_each_client: bool,
 }
 
 #[derive(serde::Deserialize, Debug)]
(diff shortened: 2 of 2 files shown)


examples/private_set_intersection/example.toml
```diff
@@ -1,4 +1,5 @@
 name = "private_set_intersection"
+rerun_server_for_each_client = true
```
tiziano88, line 2: Why is this not under `[server]`?

runner/src/main.rs
```diff
@@ -563,6 +569,8 @@ struct Example {
     backend: Option<Executable>,
     applications: HashMap<String, Application>,
     clients: HashMap<String, Executable>,
+    #[serde(default = "default_bool")]
```
tiziano88, line 572: Does `#[serde(default)]` not work here?

runner/src/main.rs
```diff
@@ -650,10 +658,39 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
+                        .clients
+                        .iter()
+                        .filter(|(name, _)| match opt.build_client.client_variant.as_str() {
+                            ALL_CLIENTS => true,
+                            client => *name == client,
+                        })
+                        .map(|(name, client)| Step::WithBackground {
+                            name: "background server".to_string(),
+                            background: run_example_server(
+                                &opt.build_server,
```
tiziano88, line 674: Is this basically a slightly modified version of `run_clients` above? If so, how would future maintainers realise this without a comment or some other indication of what's going on here?

tiziano88 (commented):
TBH I am not sure this is the right way to go, what actual server would require being restarted every time in order to work properly? If the issue is showing how PSI can work with multiple sets of clients, I think the proper solution would be to allow them to specify a run / bucket / set id that they provide at each invocation, so that multiple sets of clients may reuse the same server instance. WDYT?

rbehjati:
I agree with @tiziano88 that this should be fixed in the example code. I think in a realistic scenario a single instance of the Oak application should be able to handle any number of requests from any number of clients (but I am not sure what is the best/correct way to do this). @ipetr0v I'd suggest that you temporarily comment out one of the clients in #1464, to allow you to merge that PR, and then modify the PSI application to handle multiple clients in a separate PR.

PR #1535, opened 2020-09-30
Lowproj step WriteChannel FlowsTo case

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/EvAugSemantics.v
+++ b/experimental/ni_coq/vfiles/EvAugSemantics.v
@@ -62,27 +62,27 @@ Definition head_st (t: trace) :=
     end.
 
 Inductive step_node_ev (id: node_id): call -> state -> state -> event_l -> Prop :=
-    | SWriteChan s n han msg s':
+    | SWriteChanEv s n han msg s':
         s.(nodes) .[?id] = Some n ->
         step_node id (WriteChannel han msg) s s' ->
         step_node_ev id (WriteChannel han msg) s s' (n ---> msg)
-    | SReadChan s n han chan msg s':
+    | SReadChanEv s n han chan msg s':
         s.(nodes) .[?id] = Some n ->
         step_node id (ReadChannel han) s s' ->
         msg_is_head chan msg ->
         step_node_ev id (ReadChannel han) s s' (n <--- msg)
-    | SCreateChan s n lbl s':
+    | SCreateChanEv s n lbl s':
             (* It seems clear that no event is needed since nodes only observe
             * contents of channels indirectly via reads *)
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateChannel lbl) s s' ->
         step_node_ev id (CreateChannel lbl) s s' (n --- )
-    | SCreateNode s n lbl h s':
+    | SCreateNodeEv s n lbl h s':
             (* model observation that a node is created ?? *)
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateNode lbl h) s s' ->
         step_node_ev id (CreateNode lbl h) s s' (n ---)
-    | SInternal s n s':
+    | SInternalEv s n s':
         s.(nodes) .[?id] = Some n ->
         step_node id Internal s s' ->
         step_node_ev id Internal s s' (n ---).
--- a/experimental/ni_coq/vfiles/EvAugSemantics.v
+++ b/experimental/ni_coq/vfiles/EvAugSemantics.v
@@ -78,7 +78,6 @@ Inductive step_node_ev (id: node_id): call -> state -> state -> event_l -> Prop 
         step_node id (CreateChannel lbl) s s' ->
         step_node_ev id (CreateChannel lbl) s s' (n --- )
     | SCreateNodeEv s n lbl h s':
-            (* model observation that a node is created ?? *)
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateNode lbl h) s s' ->
         step_node_ev id (CreateNode lbl h) s s' (n ---)
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -71,6 +71,19 @@ Proof.
     reflexivity. auto.
 Qed.
 
+Theorem chan_projection_preserves_lbl: forall ell ch,
+    (clbl (chan_low_proj ell ch)) = clbl ch.
+Proof.
+Admitted. (* WIP *)
+
+Theorem uncons_proj_chan_s: forall ell s han ch,
+    (chans (state_low_proj ell s)).[? han] = Some ch ->
+    exists ch',
+         (chans s).[? han] = Some ch' /\
+        chan_low_proj ell ch' = ch.
+Proof.
+Admitted. (* WIP *)
+
 Theorem proj_node_state_to_proj_n: forall ell s id n,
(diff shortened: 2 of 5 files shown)


experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -98,6 +154,7 @@ Proof.
         - (* some *) 
         replace n0 with n in *.
         erewrite nflows_node_proj in H0. inversion H0. auto. auto.
+        replace (nodes s id) with ((nodes s).[? id]) in E by auto.
```
jadephilipoom, line 157: You could probably make this change a lemma, which would make using it less verbose.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -109,9 +109,40 @@ Proof.
...
+                    rewrite H6 in H0. inversion H0. symmetry. eapply flows_node_proj. auto.
+                }
+                eapply SWriteChan; try congruence; try assumption;
+                    try (erewrite <- chan_projection_preserves_lbl; eassumption).
+                (* s1'' =L s2'' *)
+                (* This part looks tactic-able to me *)
+                subst. unfold s1''. eapply set_call_unwind. unfold ch'.
+                eapply state_upd_chan_unwind. eapply chan_append_unwind. 
+                eapply chan_low_proj_loweq. eapply state_upd_node_unwind.
+                reflexivity. eapply state_low_proj_loweq.
```
jadephilipoom, line 137: May or may not work:
```suggestion
                eauto using state_upd_chan_unwind, chan_append_unwind, 
                eapply chan_low_proj_loweq, eapply state_upd_node_unwind,
                state_low_proj_loweq.
```

If that works, you can make it nicer using a custom auto database. Somewhere above, not in the lemma, write:
```
Hint Resolve state_upd_chan_unwind chan_append_unwind
                eapply chan_low_proj_loweq eapply state_upd_node_unwind
                state_low_proj_loweq : unwind.
```
...and then you can just use `eauto with unwind` to get the same behavior.

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -109,9 +109,32 @@ Proof.
...
             as [n' [Hidx_n' Hproj_n']].
         destruct (n'.(nlbl) <<? ell).
-        * (* flowsto case*) (* likely by inversion on H_step_projs_s1' *)
-        (* inversion H_step_projs_s1'; assert (n0 = n) by congruence; subst. *)
-        admit. 
+        *
+            (* flowsto case*) (* likely by inversion on H_step_projs_s1' *)
+            inversion H_step_projs_s1'; assert (n0 = n) by congruence; subst.
+            + (* WriteChannel *)
+                inversion H3; subst.
```
jieungkim, line 116: If you are OK, I recommend you to make a user-defined inversion. One possible simple version might be 

Ltac inv H := inversion H; subst. 

It's quite useful because this pattern will appear in several places when we try to prove some properties by case analyses for transition rules in the future. 


It's quite useful because this pattern will appear several places when we try to prove some properties by case analyses for transition rules in the future.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -109,9 +109,32 @@ Proof.
...
+                specialize (uncons_proj_chan_s ell s han ch H8)
+                    as [ch2 [H_ch'_idx H_ch2_proj]].
+                remember (s_set_call (state_upd_chan han (chan_append ch2 msg)
+                (state_upd_node id n'0 s)) id c') as s2''.
+                exists s2'', ((node_low_proj ell n') ---> msg); repeat try split.
```
jieungkim, line 121: change "repeat try split" to "repeat split". I believe they have the same behaviors (please discard my comments when it behaves differently).

experimental/ni_coq/vfiles/PossibilisticNI.v
jieungkim: @jadephilipoom points out the similar thing that I hoped to say in here. For more explanations about hint database, Please look at [here](https://softwarefoundations.cis.upenn.edu/lf-current/Auto.html).

experimental/ni_coq/vfiles/EvAugSemantics.v
```diff
@@ -63,26 +63,27 @@ Definition head_st (t: trace) :=
...
-    | SCreateChan s n lbl s':
+    | SCreateChanEv s n lbl s':
             (* It seems clear that no event is needed since nodes only observe
             * contents of channels indirectly via reads *)
         s.(nodes) .[?id] = Some n ->
         step_node id (CreateChannel lbl) s s' ->
         step_node_ev id (CreateChannel lbl) s s' (n --- )
-    | SCreateNode s n lbl h s':
+    | SCreateNodeEv s n lbl h s':
+            (* model observation that a node is created ?? *)
```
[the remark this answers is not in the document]
aferr: Oops. I did not. Thanks. Good catch :)

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Huh. I tried both "eauto using ..." as well as making a separate hint database. I also bumped up the number after eauto (to 12). Surprisingly to me neither one of those worked even though that's pretty much an exhaustive list of all the tactics needed. Even doing "eauto using state_upd_chan_unwind" alone when that's the theorem needed doesn't change the goal. Is there any way to gain more information from coq about why this doesn't work?

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: There's invert_clean in Tactics.v. I just keep forgetting to use it. At some point, I'll make a pass to use it everywhere.

experimental/ni_coq/vfiles/PossibilisticNI.v
jadephilipoom: I'm not too terribly surprised, to be honest --  for performance reasons, `eapply` tries harder to make the goal and lemma match than `eauto` does. When you do these `eapply` calls, does the goal look *exactly* like the lemma postcondition, or does something need to be changed first (definition unfolded, two slightly-different-but-actually-equal arguments unified)? It will probably work (and help debugging) to use `Hint Extern` instead, which very directly tells `eauto` what to do:
```coq
Hint Extern 4 (state_low_eq _ (state_upd_chan ?han _ _) (state_upd_chan ?han _ _) => (eapply state_upd_chan_unwind) : unwind.
```
Relevant Coq manual entry [here](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:cmdv.hint-extern), but the tl;dr is that this tells `eauto` to try `eapply state_upd_chan_unwind` any time it sees a goal that matches the pattern `(state_low_eq _ (state_upd_chan ?han _ _) (state_upd_chan ?han _ _)`, where the underscores can be anything and the `?han` variables can be anything but have to match each other. This is probably pretty much what `eauto` is doing already, but it's a more adjustable hint, so if you inspect the proof and see that the goal actually looks slightly different right before you `eapply`, then you can change the pattern to apply the lemma more aggressively. You could even give it a pattern like `state_low_eq _ _ _`.

All of this is probably not worthwhile unless you find yourself using the same set of lemmas all over the place, but it also can come in handy so I thought I'd give you the pointer!

PR #1538, opened 2020-09-30
Event for node creation. Prune old things. A TODO

aferr:
In EvAugSemantics, an event for node creation is added. I think this is
more accurate because presumably an attacker can tell whether or not a
node gets created.

I think it still does not make sense to add an event for node creation
because node creation can only be observed indirectly (i.e., an attacker
writes code that tries to read or write from a channel to see if the channel
exists). The events for read/write will capture these potential indirect
leaks.

Things to do with changing the call in the last node to execute when
constructing traces are removed. These are no longer needed now that
there is a single step semantics on just states and traces are built
separately.

A note about connecting system_step and system_step_ev are added.

# Checklist

      construction.


jadephilipoom (approved):
LGTM!

PR #1552, opened 2020-10-03
Implement IntoIterator for HeaderMap in oak_services

rbehjati:
Ref [comment on 1528](https://github.com/project-oak/oak/pull/1528#discussion_r497370582).

### DIFF
--- a/oak_services/src/http/mod.rs
+++ b/oak_services/src/http/mod.rs
@@ -37,7 +37,7 @@ impl std::convert::TryFrom<HttpRequest> for Request<Vec<u8>> {
             .method(request.method.as_bytes())
             .uri(request.uri);
         if let Some(headers) = request.headers {
-            for (header_name, header_value) in headers.iter() {
+            for (header_name, header_value) in headers.into_iter() {
                 builder = builder.header(header_name, header_value);
             }
         }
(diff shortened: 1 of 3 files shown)


oak_services/src/http/mod.rs
```diff
@@ -73,30 +73,32 @@ impl From<http::header::HeaderMap<http::header::HeaderValue>> for HeaderMap {
...
+
+    /// Convert into an iterator over (http::header::HeaderName, http::header::HeaderValue) tuples.
+    fn into_iter(self) -> Self::IntoIter {
+        let iter = self.headers
+            .into_iter()
             .flat_map(|(name, value)| {
-                value.values.iter().filter_map(move |val| {
-                    let name_value_pair = || -> Result<(http::header::HeaderName, http::header::HeaderValue), OakStatus>  {
+                value.values.into_iter().filter_map(move |val| {
+                    let name_value_pair_closure = || -> Result<(http::header::HeaderName, http::header::HeaderValue), OakStatus>  {
```
tiziano88, line 86: What is this closure actually doing? I'm a bit confused

oak_services/src/http/mod.rs
```diff
@@ -73,30 +73,32 @@ impl From<http::header::HeaderMap<http::header::HeaderValue>> for HeaderMap {
...
                         let header_name = http::header::HeaderName::from_bytes(name.as_bytes())
                             .map_err(|err| {
                                 warn!("Error when parsing header name: {}", err);
                                 OakStatus::ErrInternal
                             })?;
                         let header_value =
-                            http::header::HeaderValue::from_bytes(val).map_err(|err| {
+                            http::header::HeaderValue::from_bytes(&val[..]).map_err(|err| {
```
tiziano88, line 93: Does `as_ref` work here?

oak_services/src/http/mod.rs
```diff
@@ -73,30 +73,32 @@ impl From<http::header::HeaderMap<http::header::HeaderValue>> for HeaderMap {
...
-impl HeaderMap {
-    /// Flatten the headers into an iterator of tuples (String, http::header::HeaderValue).
-    pub fn iter(
-        &self,
-    ) -> impl Iterator<Item = (http::header::HeaderName, http::header::HeaderValue)> + '_ {
-        self.headers
-            .iter()
+impl std::iter::IntoIterator for HeaderMap {
+    type Item = (http::header::HeaderName, http::header::HeaderValue);
+    type IntoIter = Box<dyn Iterator<Item = (http::header::HeaderName, http::header::HeaderValue)>>;
```
tiziano88, line 78: Why is the boxing and dynamic dispatch necessary? Is there no static type that works here?

e.g. https://doc.rust-lang.org/std/slice/struct.Iter.html

Or even better, does the "impl trait" syntax work here?

tiziano88 (approved):
Nice one! Though I'm a bit puzzled by the choice of type

oak_services/src/http/mod.rs
rbehjati: The actual type is `std::iter::FlatMap` with some anonymous type parameters. With `impl Trait` there is the ``impl Trait` in type aliases is unstable` error message that can be fixed by adding `#![feature(type_alias_impl_trait)]`. Should I use `impl Trait`?

oak_services/src/http/mod.rs
rbehjati: This is only used to allow using `?` in the enclosing lambda. Otherwise the code would be more verbose.

oak_services/src/http/mod.rs
rbehjati: Thanks. Done.

oak_services/src/http/mod.rs
tiziano88: Interesting, I didn't know about that feature. I don't think it's worth using it, but `FlatMap` seems a good type for this iterator, not sure about the anonymous type parameters though, do you mean they are unnameable types?

rbehjati: This is the exact type: `std::iter::FlatMap<std::collections::hash_map::IntoIter<std::string::String, proto::oak::encap::HeaderValue>, std::iter::FilterMap<std::vec::IntoIter<std::vec::Vec<u8>>, [closure@/opt/my-project/oak_services/src/http/mod.rs:85:53: 100:18 name:_]>, [closure@/opt/my-project/oak_services/src/http/mod.rs:84:23: 101:14]>`. The `FilterMap` in it makes it complicated.

tiziano88: If that's the case, it seems like an unnameable (not anonymous) type indeed (see https://internals.rust-lang.org/t/idea-making-unnameable-types-nameable/7797 ) -- more precisely, the type of the closure is unnameable, and it ends up being part of the reified `FlatMap` type.

rbehjati: Yes. That is the problem. I did not know about the term `unnameable`. Thanks.

PR #1557, opened 2020-10-05
Create 'unwind' hint db for eauto

jadephilipoom:
Follow-up to discussion here: https://github.com/project-oak/oak/pull/1535#discussion_r499647876

The reason the plain `Hint Resolve`s weren't working was that `state_upd_node_unwind` produces a `node_low_eq` subgoal, and there's no declared `Reflexive` instance for that definition. Using an explicit `Hint Extern` tells `eauto` to unfold the definition until it finds an equality underneath. So I was wrong to suspect in the previous discussion that it was an `eapply` unfolding things more aggressively than `eauto` does; it was actually `reflexivity`!

# Checklist

      construction.


aferr (approved):
That's much better. Thanks. Thanks also for the detailed comment in the other PR and for the link to the appropriate part of the ref manual.

PR #1560, opened 2020-10-06
Introspection: Add favicon

juliettepretot:
Small but I like it

this pr:
Screenshot at Oct 06 11-28-01 am

main branch:
Screenshot at Oct 05 4-00-45 pm

# Checklist

      construction.


oak_runtime/src/introspect.rs
```diff
@@ -111,20 +111,47 @@ fn find_client_file(path: &str) -> Option<(Vec<u8>, String)> {
...
+
+            Some(response)
+        }
+        "/graphvizlib.wasm" => {
+            let mut response = Response::new(Body::from(
+                include_bytes!("../introspection_browser_client/dist/graphvizlib.wasm.gz").to_vec(),
+            ));
+            let headers = response.headers_mut();
+            headers.insert(CONTENT_TYPE, "application/wasm".parse().unwrap());
+            headers.insert(hyper::header::CONTENT_ENCODING, "gzip".parse().unwrap());
```
juliettepretot, line 130: An odd Rust thing happened here: I had to pick the `CONTENT_ENCODING` inline, instead importing it at the top of the file to fix a rust warning:

image

I suspect the reason is that not all items in the match use this type? It seemed like weird behaviour.

oak_runtime/src/introspect.rs
```diff
@@ -111,20 +111,47 @@ fn find_client_file(path: &str) -> Option<(Vec<u8>, String)> {
...
+
+            Some(response)
+        }
+        "/favicon.png" => {
+            let mut response = Response::new(Body::from(
+                include_bytes!("../introspection_browser_client/dist/favicon.png").to_vec(),
+            ));
```
tiziano88, line 137: May be useful to define some helper method(s) so that you can just do something like `Some(static_file("favicon.png", "application/png"))` etc.

oak_runtime/src/introspect.rs
tiziano88: I think that's just because the import is only used if `oak_introspection_client` is on, but it is imported unconditionally. Maybe it would make sense to have a submod that gets excluded completely if the feature is not enabled, and would contain also the imports? Otherwise you would have to feature gate the imports too, or use them inline as you did here.

tiziano88 (approved):
cute! :deciduous_tree:

oak_runtime/src/introspect.rs
juliettepretot: Ah that explains it. :) Initially I ruled that possibility out since `CONTENT_TYPE` works fine, but indeed that type is used outside of the feature flagged function. 

Will implement your suggestion of the conditional import.

oak_runtime/src/introspect.rs
juliettepretot: Good point, did that :). I'm always reluctant to create abstractions, but it does make things easier here.

oak_runtime/src/introspect.rs
```diff
@@ -25,6 +25,71 @@ use prost::Message;
...
+            )),
+            "/graphvizlib.wasm" => Some(static_file(
+                "../introspection_browser_client/dist/graphvizlib.wasm.gz",
+                "application/wasm",
+                Some("gzip"),
+            )),
+            "/favicon.png" => Some(static_file(
+                "../introspection_browser_client/dist/favicon.png",
+                "image/png",
+                None,
```
tiziano88, line 80: Can we just gzip everything?

tiziano88 (approved):

oak_runtime/src/introspect.rs
juliettepretot: I did consider it, but it would reduce code complexity slightly. But realistically it only really makes sense for text based assets. For images etc it only adds CPU overhead while potentially even increasing file size. 

I'm personally ok with a few more lines of code if it means we follow web-dev best practices.

oak_runtime/src/introspect.rs
```diff
@@ -25,6 +25,71 @@ use prost::Message;
...
+    fn static_file(
+        file: &str,
+        content_type: &str,
+        content_encoding: Option<&str>,
+    ) -> Response<Body> {
+        let mut response = Response::new(Body::from(file.to_string()));
+        let headers = response.headers_mut();
+        headers.insert(CONTENT_TYPE, content_type.parse().unwrap());
+        if let Some(content_encoding_value) = content_encoding {
+            headers.insert(CONTENT_ENCODING, content_encoding_value.parse().unwrap());
```
tiziano88, line 51: Is it enough to just add this header in order to make the server actually do the compression magically? Or is there something else that I am missing? Either way, perhaps worth a comment / link to some reference?

oak_runtime/src/introspect.rs
```diff
@@ -25,6 +25,71 @@ use prost::Message;
...
+
+#[cfg(feature = "oak_introspection_client")]
+mod introspection_client {
+    use super::*;
+    use hyper::header::CONTENT_ENCODING;
+
+    fn static_file(
+        file: &str,
+        content_type: &str,
+        content_encoding: Option<&str>,
```
tiziano88, line 45: Maybe create a bool-like enum for this, e.g.

```rust
enum Compress {
  Yes,
  No,
}
```

which also helps with readability when invoking the function.

tiziano88 (approved):

oak_runtime/src/introspect.rs
juliettepretot: Good point, that's not really obvious from the code, happy to add a comment. 

The tl;dr is that the rust server only serves static files, headers can be seen as mere "metadata" instructing the browser how to parse the received files. 

The gzip compression itself takes place when assets are generated by webpack.

tiziano88: Ah thanks, my bad, I didn't notice the `.gz` suffix earlier, it all makes sense now.

oak_runtime/src/introspect.rs
juliettepretot: Mmmh, not sure if it makes sense to abstract the HTTP header with a boolean. Since compression is a property of the file (as opposed to compression being applied by the rust code) I think it makes sense to explicitly specify which type was used. Theoretically this might differ per file as well, with other compressions such as brotli existing.

juliettepretot: The readability concern imho will be addressed by the comment explaining what headers do exactly.

PR #1561, opened 2020-10-06
Introspection: (Re)add missing include_bytes! call

juliettepretot:
Fixes a bug introduced in #1560 that led to serving the string file path instead of the actual file. 😬

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/introspect.rs
+++ b/oak_runtime/src/introspect.rs
@@ -40,11 +40,11 @@ mod introspection_client {
     use hyper::header::CONTENT_ENCODING;
 
     fn static_file(
-        file: &str,
+        file: Vec<u8>,
         content_type: &str,
         content_encoding: Option<&str>,
     ) -> Response<Body> {
-        let mut response = Response::new(Body::from(file.to_string()));
+        let mut response = Response::new(Body::from(file));
         let headers = response.headers_mut();
         headers.insert(CONTENT_TYPE, content_type.parse().unwrap());
         // If the served file is compressed, add a header instructing the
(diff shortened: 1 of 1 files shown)


oak_runtime/src/introspect.rs
```diff
@@ -40,11 +40,11 @@ mod introspection_client {
     use hyper::header::CONTENT_ENCODING;
 
     fn static_file(
-        file: &str,
+        file: Vec<u8>,
```
tiziano88, line 43: Does `&[u8]` not work here? `Vec` would cause the entire file to be re-allocated for each request, which is not ideal.

tiziano88 (approved):

oak_runtime/src/introspect.rs
juliettepretot: Thanks for the heads up. :) `&'static [u8]` does work.

I had (mistakenly) assumed that accepting an array as an arg required typing out it's length (which would prohibit different files), but turns this works a-ok.

tiziano88: - `[u8; 1024]` is an array of 1024 bytes, and would get passed as the entire 1024 bytes on the stack, by value (I think).
- `&[u8]` is a pointer to a byte slice, and would get passed as a pointer (a single word) on the stack.

tiziano88: See https://doc.rust-lang.org/beta/rust-by-example/primitives/array.html

juliettepretot: makes sense :)

PR #1562, opened 2020-10-06
Automate possibilistic_ni proof and prove no_steps_from_empty

jadephilipoom:
I put together some tactics and used them to automate the `possibilistic_ni` proof. Also proves a small admit in `TraceTheorems.v`. The proof structure here is actually pretty similar to before; it's just that the tactics are handling manipulations like plugging one hypothesis into the other.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -63,6 +63,32 @@ https://www.cs.cornell.edu/andru/papers/csfw03.pdf
     ***
 *)
 
+(* TODO: maybe move to Tactics.v *)
+Local Ltac logical_simplify :=
+  repeat match goal with
+         | H : _ /\ _ |- _ => destruct H
+         | H : exists _, _ |- _ => destruct H
+         | H1 : ?P, H2 : ?P -> _ |- _ => specialize (H2 H1)
+         | H : ?x = ?x |- _ => clear H
+         end.
+
+(* Single step of [crush] *)
+Local Ltac crush_step :=
+  repeat match goal with
+         | _ => progress intros
+         | _ => progress subst
+         | _ => progress logical_simplify
+         | H : Some _ = Some _ |- _ => invert_clean H
+         | H1 : ?x = Some _, H2 : ?x = Some _ |- _ =>
+           rewrite H2 in H1; invert_clean H1
+         | _ => reflexivity
+         end.
+(* General-purpose tactic that simplifies and solves simple goals for
+   possibilistic NI. *)
+Local Ltac crush := repeat crush_step.
+
+Hint Resolve multi_system_ev_refl multi_system_ev_tran : multi.
+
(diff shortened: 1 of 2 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -287,28 +317,12 @@ Admitted. (* WIP *)
...
-        specialize (step_system_multi_extends _ (a::t1n) ltac:(eauto)) as E2.
-        specialize (possibilistic_ni_unwind_t ell t1n t2n (a::t1n) Hleq_t1n_t2n E2)
-            as [t2n' [Hs_t2n_t2n' Hleq_at1n_t2n']].
-        exists t2n'. split.
-        apply (step_system_transitive t2_init t2n t2n'). assumption.
-        constructor. assumption. assumption.
+  unfold conjecture_possibilistic_ni. crush.
+  let H := match goal with H : step_system_ev_multi _ _ |- _ => H end in
+  induction H; crush.
+  all:pose proof (possibilistic_ni_unwind_t
```
aferr, line 323: What do the "all:" annotations do? I can't seem to find the relevant part of the reference manual.

From what I can find in the ref manual "pose proof X" Is kind of like "assert (H: T) using X" where T is the type of X, right? So this will make new hypotheses, (In this case "exists t2', ....") right?

And then what do the "ltac:..." bits do? It looks like they are in the place of arguments of type (trace_low_eq ell t1 t2) and (step_system_ev_t t1 t1'). Does this do something like tell coq to use the given ltac tactic to figure out what these args should be?

aferr (approved):

experimental/ni_coq/vfiles/PossibilisticNI.v
jadephilipoom: > What do the "all:" annotations do? I can't seem to find the relevant part of the reference manual.

They run the tactic on all the focused subgoals, instead of just the first one. [Here](https://coq.inria.fr/refman/proof-engine/ltac.html#coq:tacn.goal-selector)'s the manual entry (with typical Coq manual entry logic, it was listed under `... : ...` in the tactic index, because you can also do other expressions e.g `1:` or `3-5,7:` to run tactics on only subgoal 1 or subgoals 3,4,5, and 7.

> From what I can find in the ref manual "pose proof X" Is kind of like "assert (H: T) using X" where T is the type of X, right? So this will make new hypotheses, (In this case "exists t2', ....") right?

Yes, exactly.

> And then what do the "ltac:..." bits do? It looks like they are in the place of arguments of type (trace_low_eq ell t1 t2) and (step_system_ev_t t1 t1'). Does this do something like tell coq to use the given ltac tactic to figure out what these args should be?

Yep, that's precisely it! It says "fill these arguments using the `eassumption` tactic", and because doing that for the two preconditions fills in all the variables I leave underscores for them and let Coq infer what they are.

PR #1563, opened 2020-10-07
Introspection: Persist state across links & page refreshes

juliettepretot:
Uses [session storage](https://developer.mozilla.org/en/docs/Web/API/Window/sessionStorage) to persist application state across page refreshes & links. 

Prevents state from resetting when navigating via native links etc

# Checklist

      construction.

### DIFF
--- a/oak_runtime/introspection_browser_client/components/Root/index.tsx
+++ b/oak_runtime/introspection_browser_client/components/Root/index.tsx
@@ -309,14 +309,13 @@ export default function Root() {
 
   const classes = useStyles();
 
-  // Subset of events representing the inspected point in time
-  const events = totalEvents.slice(0, presentEventIndex + 1);
-
   // Application state based of the current point in time
-  const applicationState: OakApplicationState = events.reduce(eventReducer, {
-    nodeInfos: new Map(),
-    channels: new Map(),
-  });
+  const applicationState: OakApplicationState = totalEvents
+    .slice(0, presentEventIndex + 1)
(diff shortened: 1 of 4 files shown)


oak_runtime/introspection_browser_client/components/Root/index.tsx
```diff
@@ -241,23 +241,52 @@ function useEvents() {
...
       setEvents(events);
-      setPresentEventIndex(events.length - 1);
     }
 
     loadEvents();
   }, []);
 
+  // Persist the presentEventIndex across page refreshes & links.
+  // Ref: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
+  React.useEffect(() => {
```
tiziano88, line 281: I don't really understand how these effects work in React, do you have any good reference I can read on?

tiziano88 (approved):

oak_runtime/introspection_browser_client/components/Root/index.tsx
juliettepretot: Good question! It's a complex topic, esp if you're new to React. 
The TL;DR that useEffect hooks are used to separate side effects from the rest of the component. It accepts to arguments a function (that performs a side effect) and a dependency array. Whenever an item in the dependency array changes, the effect reruns. 

I would recommend the react docs for an overview: https://reactjs.org/docs/hooks-effect.html
And https://overreacted.io/a-complete-guide-to-useeffect if you'd like to go deeper.

Issue #1522, opened 2020-09-25
experimental setup for performance testing and benchmarking


tiziano88:
We should start thinking how to approach an experimental setup that would allow us to gather performance characteristics of Oak applications.

I think to start with we can focus on a particular example (trusted database?), and on latency, throughput, memory footprint, network badwidth etc.

@ipetr0v WDYT? could you look into it and clarify the details?

tiziano88:
@ipetr0v what do you think we should measure, concretely, and how? Rust supports benchmarks (similar to tests) natively (see https://doc.rust-lang.org/1.7.0/book/benchmark-tests.html ) but I am not sure they are sufficient for our use case.

tiziano88:
This also looks potentially interesting: https://bheisler.github.io/criterion.rs/book/getting_started.html

ipetr0v:
In the case of trusted database it looks like we will need to measure the response time depending on the database size, and for now we can just measure it on a client.

tiziano88:
Sure but I don't think we want to manually create a server, manually create a client, manually create a response, manually measure the elapsed time, right? We will have to figure out a setup to reproduce these scenarios and drive them from within some Rust code.

ipetr0v:
We already have similar code in the integration tests:
https://github.com/project-oak/oak/blob/71b950c2803a3f22f69a504fca77a2c915dc8ee4/examples/trusted_database/module_0/rust/tests/integration_test.rs#L136-L143

tiziano88:
Will we not need some code to vary some parameters (e.g. DB size, request size, etc.), perform some operations until we achieve statistical significance, print out performance values, etc.? Basically the kind of stuff that benchmarking frameworks usually offer.

ipetr0v:
I'm measuring XML database parsing outside of Oak.
And also Oak runtime loading time, that includes: starting Oak, sending config and parsing database (but visually it seems that most of the time is spent parsing database)
```
Database parsing time: 0.004s for 32204 bytes
Runtime loading time: 11.917s

Database parsing time: 0.025s for 161007 bytes
Runtime loading time: 52.115s

Database parsing time: 0.050s for 322010 bytes
Runtime loading time: 102.245s
```
More than a minute, compared to less than a second outside of Oak.

ipetr0v:
Preliminary (and very experimental) results:
```
Database size: 100 entries
Database parsing time: 0.005s for 32195 bytes
Runtime loading time: 11.962s
Average raw request time: 0.004s
[...]

PR #1565, opened 2020-10-07
Add Trusted Database benchmark

ipetr0v:
This change adds an experimental Trusted Database benchmark.

Ref https://github.com/project-oak/oak/issues/1522

# Checklist
      construction.

### DIFF
--- a/examples/trusted_database/module/rust/src/database.rs
+++ b/examples/trusted_database/module/rust/src/database.rs
@@ -14,7 +14,9 @@
 // limitations under the License.
 //
 
-use crate::proto::oak::examples::trusted_database::{Location, PointOfInterest};
+use crate::proto::oak::examples::trusted_database::{
+    Location, PointOfInterest, PointOfInterestMap,
+};
 use log::{debug, error};
 use oak::OakError;
 use oak_abi::proto::oak::application::ConfigMap;
@@ -27,7 +29,7 @@ struct Database {
     #[serde(rename = "lastUpdate", default)]
     last_update: String,
     version: String,
-    #[serde(rename = "station", default)]
+    #[serde(rename = "station")]
     stations: Vec<Station>,
 }
 
@@ -41,7 +43,9 @@ struct Station {
     latitude_degrees: f32,
     #[serde(rename = "long", default)]
     longitude_degrees: f32,
+    #[serde(default)]
     installed: bool,
+    #[serde(default)]
     locked: bool,
     #[serde(rename = "installDate", default)]
     install_date: String,
@@ -57,17 +61,19 @@ struct Station {
 }
 
 /// Load an XML database from [`ConfigMap`] and parse it.
-pub fn load_database(config_map: ConfigMap) -> Result<Vec<PointOfInterest>, OakError> {
+pub fn load_database(config_map: ConfigMap) -> Result<PointOfInterestMap, OakError> {
     debug!("Loading database");
     match config_map.items.get("database") {
         Some(xml_database) => {
+            debug!("Parsing database - size: {} bytes", xml_database.len());
             let points_of_interest = parse_database(xml_database).map_err(|error| {
                 error!("Couldn't parse database: {:?}", error);
                 OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
             })?;
             debug!(
-                "Database loaded - size: {} entries",
-                points_of_interest.len()
+                "Database loaded - size: {} entries ({} bytes)",
+                points_of_interest.entries.len(),
+                std::mem::size_of_val(&points_of_interest),
             );
             Ok(points_of_interest)
         }
@@ -78,8 +84,8 @@ pub fn load_database(config_map: ConfigMap) -> Result<Vec<PointOfInterest>, OakE
     }
 }
 
-/// Parse an XML database into a vector of [`PointOfInterest`].
-pub fn parse_database(xml_database: &[u8]) -> Result<Vec<PointOfInterest>, OakError> {
+/// Parse an XML database into a [`PointOfInterestMap`].
+pub fn parse_database(xml_database: &[u8]) -> Result<PointOfInterestMap, OakError> {
     let database: Database = quick_xml::de::from_str(
         String::from_utf8(xml_database.to_vec())
             .map_err(|error| {
@@ -93,16 +99,21 @@ pub fn parse_database(xml_database: &[u8]) -> Result<Vec<PointOfInterest>, OakEr
         OakError::OakStatus(oak_abi::OakStatus::ErrInvalidArgs)
     })?;
 
-    let points_of_interest = database
-        .stations
-        .iter()
-        .map(|station| PointOfInterest {
-            name: station.name.to_string(),
-            location: Some(Location {
-                latitude_degrees: station.latitude_degrees,
-                longitude_degrees: station.longitude_degrees,
-            }),
-        })
-        .collect();
+    let points_of_interest = PointOfInterestMap {
+        entries: database
+            .stations
+            .iter()
+            .map(|station| {
+                let point_of_interest = PointOfInterest {
+                    name: station.name.to_string(),
+                    location: Some(Location {
+                        latitude_degrees: station.latitude_degrees,
+                        longitude_degrees: station.longitude_degrees,
+                    }),
+                };
+                (station.id.to_string(), point_of_interest)
+            })
+            .collect(),
+    };
     Ok(points_of_interest)
 }
--- a/examples/trusted_database/module/rust/src/handler.rs
+++ b/examples/trusted_database/module/rust/src/handler.rs
@@ -23,8 +23,9 @@
 //! Main Node.
 
 use crate::proto::oak::examples::trusted_database::{
-    ListPointsOfInterestRequest, ListPointsOfInterestResponse, Location, PointOfInterest,
-    TrustedDatabase, TrustedDatabaseCommand, TrustedDatabaseDispatcher,
+    GetPointOfInterestRequest, GetPointOfInterestResponse, ListPointsOfInterestRequest,
+    ListPointsOfInterestResponse, Location, PointOfInterestMap, TrustedDatabase,
+    TrustedDatabaseCommand, TrustedDatabaseDispatcher,
 };
 use log::{debug, error, warn};
 use oak::{grpc, io::ReceiverExt};
@@ -35,11 +36,32 @@ const EMPTY_DATABASE_ERROR: &str = "Empty database";
 
 /// Oak Node that contains a copy of the database.
 pub struct TrustedDatabaseHandlerNode {
-    points_of_interest: Vec<PointOfInterest>,
+    points_of_interest: PointOfInterestMap,
 }
 
 /// A gRPC service implementation for the Private Information Retrieval example.
 impl TrustedDatabase for TrustedDatabaseHandlerNode {
+    // Find Point Of Interest based on id.
+    fn get_point_of_interest(
+        &mut self,
+        request: GetPointOfInterestRequest,
+    ) -> grpc::Result<GetPointOfInterestResponse> {
+        debug!("Received request: {:?}", request);
+        match self.points_of_interest.entries.get(&request.id) {
+            Some(point) => {
+                debug!("Found Point Of Interest: {:?}", point);
+                Ok(GetPointOfInterestResponse {
+                    point_of_interest: Some(point.clone()),
+                })
+            }
+            None => {
+                let err = grpc::build_status(grpc::Code::Internal, &EMPTY_DATABASE_ERROR);
+                error!("{:?}", err);
+                Err(err)
+            }
+        }
+    }
+
     /// Find the nearest Point Of Interest based on linear search in the database.
     fn list_points_of_interest(
         &mut self,
@@ -51,7 +73,7 @@ impl TrustedDatabase for TrustedDatabaseHandlerNode {
             warn!("{:?}", err);
             err
         })?;
-        let nearest_point = self.points_of_interest.iter().fold(
+        let nearest_point = self.points_of_interest.entries.values().fold(
             (None, f32::MAX),
             |(current_closest_point, current_closest_point_distance), point| {
                 let point_location = point
@@ -120,7 +142,7 @@ oak::entrypoint!(handler_oak_main => |in_channel| {
     let receiver = command.invocation_receiver.expect("Couldn't receive gRPC invocation receiver");
 
     // Run event loop and handle incoming invocations.
-    let node = TrustedDatabaseHandlerNode { points_of_interest: command.points_of_interest };
+    let node = TrustedDatabaseHandlerNode { points_of_interest: command.points_of_interest.expect("No database entries") };
     let dispatcher = TrustedDatabaseDispatcher::new(node);
     let invocation_receiver = receiver.receiver.expect("Empty gRPC invocation receiver");
     // The event loop only runs once because the `Main` Node sends a single invocation.
--- a/examples/trusted_database/module/rust/src/handler.rs
+++ b/examples/trusted_database/module/rust/src/handler.rs
@@ -32,6 +32,7 @@ use oak::{grpc, io::ReceiverExt};
 
 // Error messages.
 const NO_LOCATION_ERROR: &str = "Location is not specified";
+const ID_NOT_FOUND_ERROR: &str = "ID not found";
 const EMPTY_DATABASE_ERROR: &str = "Empty database";
 
 /// Oak Node that contains a copy of the database.
@@ -55,7 +56,7 @@ impl TrustedDatabase for TrustedDatabaseHandlerNode {
                 })
             }
             None => {
-                let err = grpc::build_status(grpc::Code::Internal, &EMPTY_DATABASE_ERROR);
+                let err = grpc::build_status(grpc::Code::NotFound, &ID_NOT_FOUND_ERROR);
                 error!("{:?}", err);
                 Err(err)
             }
--- a/examples/trusted_database/module/rust/src/lib.rs
+++ b/examples/trusted_database/module/rust/src/lib.rs
@@ -60,11 +60,11 @@ use log::{debug, error};
 use oak::{
     grpc, io::SenderExt, proto::oak::invocation::GrpcInvocationReceiver, CommandHandler, OakError,
 };
-use proto::oak::examples::trusted_database::{PointOfInterest, TrustedDatabaseCommand};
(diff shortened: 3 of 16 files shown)


experimental/benchmark/src/database.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[serde(rename = "removalDate", default)]
+    removal_date: String,
+    temporary: bool,
+    #[serde(rename = "nbBikes", default)]
+    number_of_bikes: u32,
+    #[serde(rename = "nbEmptyDocks", default)]
+    number_of_empty_docks: u32,
+    #[serde(rename = "nbDocks", default)]
+    number_of_docks: u32,
+}
```
tiziano88, line 60: If you are re-creating the example instead of using the existing one, why carry over all this complexity? Just create key / value blobs of pseudo-random data.

ipetr0v: I'm using the existing example, these fields are just for generating the data, that can (should be) parsed by the example.
I was planning to move this database to smth like `trusted_database/common`

tiziano88: Then all this stuff should just be under `trusted_database/benchmark` I guess? if it's specific to that example anyways

ipetr0v: It's still experimental, so I thought I'll put it in `experimental` for now

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "Trusted Database Client")]
+pub struct Opt {
+    #[structopt(long, help = "Database size (number of items)")]
+    database_size: usize,
+}
+
+#[tokio::main(core_threads = 4)]
+async fn main() -> anyhow::Result<()> {
```
rbehjati, line 39: Would be nice to have a high-level documentation here explaining the experiment; in particular mentioning the difference between the `Backend` and the `TestRuntime`.

experimental/benchmark/src/backend.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    async fn list_points_of_interest(
+        &self,
+        _request: Request<ListPointsOfInterestRequest>,
+    ) -> Result<Response<ListPointsOfInterestResponse>, Status> {
+        unimplemented!();
+    }
+}
+
+pub struct TestBackend {
```
rbehjati, line 72: nit: I think something like `TonicBackend` or `VanillaBackend` is more expressive. Similarly, I think `OakBackend` is a better name for `TestRuntime`.

You could also have a `Backend` trait with a `send_request` function, and two implementations for it, one with Tonic and one with Oak. But this is super-nit :) 

Perhaps also add a brief comment here explaining what this struct does.

examples/trusted_database/proto/trusted_database.proto
```diff
@@ -30,15 +30,25 @@ message PointOfInterest {
...
 }
 
 message ListPointsOfInterestResponse {
-  PointOfInterest point_of_interest = 2;
+  PointOfInterest point_of_interest = 1;
 }
 
 service TrustedDatabase {
+  // Find Point Of Interest based on id.
+  rpc GetPointOfInterest(GetPointOfInterestRequest) returns (GetPointOfInterestResponse);
```
rbehjati, line 51: This seems to have been added just for the sake of the benchmark, and is not part of the actual functionality of a TrustedDatabase. If this is the case, it would be nice to mention it in the comment.

Even better, if it is possible, define it as a separate service, for instance `TrustedDatabaseBenchmark`, and implement the generated trait for `TrustedDatabaseHandlerNode`, and `TrustedDatabaseService ` in the benchmark backend. But I don't know if it is possible to implement two services for `TrustedDatabaseHandlerNode`.

rbehjati (approved):
Looks good to me. But more documentation could be useful, especially if this code is meant to be maintained or be used as a replication package for a potential future paper.

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    // Measure raw request durations.
+    let iterations: u16 = 10;
+    let mut total_raw_request_time = 0;
+    for _ in 0..iterations {
+        let request_start = Instant::now();
+        let _ = backend.send_request("0").await?;
+        total_raw_request_time += request_start.elapsed().as_millis();
+    }
+    let raw_average_request_time = (total_raw_request_time as f64) / (iterations as f64) / 1000.0;
```
rbehjati, line 68: Could not you just measure the time the entire iteration takes, instead of measuring the time for each request and then summing them?

examples/trusted_database/module/rust/src/database.rs
```diff
@@ -79,7 +84,7 @@ pub fn load_database(config_map: ConfigMap) -> Result<Vec<PointOfInterest>, OakE
 }
 
 /// Parse an XML database into a vector of [`PointOfInterest`].
```
tiziano88, line 86: Update comment.

examples/trusted_database/module/rust/src/database.rs
```diff
@@ -57,10 +61,11 @@ struct Station {
 }
 
 /// Load an XML database from [`ConfigMap`] and parse it.
-pub fn load_database(config_map: ConfigMap) -> Result<Vec<PointOfInterest>, OakError> {
+pub fn load_database(config_map: ConfigMap) -> Result<HashMap<String, PointOfInterest>, OakError> {
```
tiziano88, line 64: In the return type, what are keys and values of the map?

examples/trusted_database/module/rust/src/handler.rs
```diff
@@ -23,23 +23,46 @@
...
+use std::collections::HashMap;
 
 // Error messages.
 const NO_LOCATION_ERROR: &str = "Location is not specified";
 const EMPTY_DATABASE_ERROR: &str = "Empty database";
 
 /// Oak Node that contains a copy of the database.
 pub struct TrustedDatabaseHandlerNode {
-    points_of_interest: Vec<PointOfInterest>,
+    points_of_interest: HashMap<String, PointOfInterest>,
```
tiziano88, line 40: Please describe what keys and values are.

examples/trusted_database/module/rust/src/lib.rs
```diff
@@ -59,10 +59,11 @@ use database::load_database;
 use log::{debug, error};
 use oak::{grpc, io::SenderExt, proto::oak::invocation::GrpcInvocationReceiver, Node, OakError};
 use proto::oak::examples::trusted_database::{PointOfInterest, TrustedDatabaseCommand};
+use std::collections::HashMap;
 
 /// Oak Node that contains an in-memory database.
 pub struct TrustedDatabaseNode {
-    points_of_interest: Vec<PointOfInterest>,
+    points_of_interest: HashMap<String, PointOfInterest>,
```
tiziano88, line 66: Explain key / values.

examples/trusted_database/proto/trusted_database_command.proto
```diff
@@ -23,5 +23,5 @@ import "oak_services/proto/grpc_invocation.proto";
 
 message TrustedDatabaseCommand {
   oak.invocation.GrpcInvocationReceiver invocation_receiver = 1;
-  repeated oak.examples.trusted_database.PointOfInterest points_of_interest = 2;
+  map<string, oak.examples.trusted_database.PointOfInterest> points_of_interest = 2;
```
tiziano88, line 26: What is the key?

Would it make sense to just define this type once and for all in a proto, document it there, and use it everywhere else, instead of having so many (undocumented) copies?

experimental/benchmark/src/backend.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::collections::HashMap;
+use tokio::task::JoinHandle;
+use tonic::{
+    metadata::MetadataValue,
+    transport::{Channel, Server},
+    Request, Response, Status,
+};
+
+pub struct TrustedDatabaseService {
+    points_of_interest: HashMap<String, PointOfInterest>,
```
tiziano88, line 38: What is the key?

experimental/benchmark/src/database.rs
```diff
@@ -0,0 +1,123 @@
...
+
+pub struct Database {
+    pub entries: HashMap<String, PointOfInterest>,
```
tiziano88, line 63: What is the key?

tiziano88 (commented):
Is there a high level description of this setup? I am not sure I understand whether it is meant to use the existing trusted database example logic, or reimplement it in a different way, and what the backend does, and what needs to be trusted.

ipetr0v:
> Is there a high level description of this setup? I am not sure I understand whether it is meant to use the existing trusted database example logic, or reimplement it in a different way, and what the backend does, and what needs to be trusted.

I added a short description in the beginning of the `main.rs` file.

What do you mean `what needs to be trusted`? Since it's an experiment, nothing is supposed to be trusted in this particular example.

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Trusted Database Benchmark.
+//!
+//! This benchmark measures average gRPC request time to a database that is either:
+//! - Implemented as part of the Trusted Database Oak application (using `oak::OakApplication`)
+//! - Implemented as a `tonic` gRPC server without using Oak (using `vanilla::VanillaApplication`)
```
rbehjati, line 21: I know I suggested vanilla, but after seeing the reports the code generates, I think `raw` is a better name. Just as a suggestion, up to you :)

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+//!
+//! Benchmark randomly generates (with 0 seed) a database and sends it both to Oak and Vanilla
+//! application. Then it send `REQUEST_NUMBER` of requests to each applications and computes the
```
rbehjati, line 24:
```suggestion
//! application. Then it sends `REQUEST_NUMBER` of requests to each application and computes the
```

rbehjati (approved):
Thanks :)

experimental/benchmark/src/main.rs
ipetr0v: Done

experimental/benchmark/src/main.rs
ipetr0v: Done

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "Trusted Database Client")]
+pub struct Opt {
+    #[structopt(long, help = "Database size (number of items)")]
+    database_size: usize,
+}
+
+// Port for running a gRPC server for the raw application.
+const VANILLA_APP_PORT: u16 = 8888;
```
tiziano88, line 62: Rename to something more specific?

experimental/benchmark/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    // Start Oak runtime.
+    info!(
+        "Loading Runtime: Sending database - size: {} entries ({} bytes)",
+        database.points_of_interest.entries.len(),
+        serialized_database.as_bytes().len()
+    );
+    let oak_loading_start = Instant::now();
+    let mut oak_app = OakApplication::start(&serialized_database).await;
+    let oak_loading_time = (oak_loading_start.elapsed().as_millis() as f64) / 1000.0;
```
tiziano88, line 103: Why is everything converted back and forth from millis? I think it's simpler to just keep these around as Durations, and just convert to seconds when printing (or even just print them with their default format).

experimental/benchmark/src/application/native.rs
```diff
@@ -0,0 +1,136 @@
...
+        debug!("Received request: {:?}", request);
+        match self.points_of_interest.entries.get(&request.get_ref().id) {
+            Some(point) => {
+                debug!("Found Point Of Interest: {:?}", point);
+                Ok(Response::new(GetPointOfInterestResponse {
+                    point_of_interest: Some(point.clone()),
+                }))
+            }
+            None => {
+                let err = tonic::Status::new(tonic::Code::Internal, "Empty database");
```
tiziano88, line 58: Incorrect error?

experimental/benchmark/src/application/native.rs
```diff
@@ -0,0 +1,136 @@
...
+                error!("{:?}", err);
```
tiziano88, line 59: This should probably be a warn, since it can easily be triggered by a client.

experimental/benchmark/src/application/native.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+pub struct NativeApplication {
+    _handle: JoinHandle<()>,
+    client: TrustedDatabaseClient<tonic::transport::channel::Channel>,
+}
+
+impl NativeApplication {
+    pub async fn start(database: &Database, port: u16) -> Self {
+        info!("Running backend database");
```
tiziano88, line 80: Is this accurate? I thought the backend term is used in a different context.

experimental/benchmark/src/application/native.rs
```diff
@@ -0,0 +1,136 @@
...
+            Ok(request)
+        })
+    }
+}
+
+#[async_trait]
+impl Application for NativeApplication {
+    async fn send_request(&mut self, id: &str) -> Result<(), tonic::Status> {
+        let request = Request::new(GetPointOfInterestRequest { id: id.to_string() });
+        self.client.get_point_of_interest(request).await?;
```
tiziano88, line 133: Perhaps repeat the comment about discarding the request because irrelevant to the benchmark (though TBH it seems simpler to just make this method return the actual response to the caller, and just ignore it there)

experimental/benchmark/src/application/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub mod oak;
+
+use async_trait::async_trait;
+
+/// Trait for sending test requests. Returns `()` since the value of the request is not needed for
+/// current benchmark implementation.
+#[async_trait]
+pub trait Application: std::marker::Send {
+    async fn send_request(&mut self, id: &str) -> Result<(), tonic::Status>;
+}
```
tiziano88, line 27: The abstraction here seems a bit odd, I would expect this to be an ApplicationClient or similar, and the application server to be managed separately (which it is anyways, since there are no methods on this trait to start and stop the actual application). Anyways up to you.

experimental/benchmark/src/application/native.rs
```diff
@@ -0,0 +1,136 @@
...
+    async fn list_points_of_interest(
+        &self,
+        _request: Request<ListPointsOfInterestRequest>,
+    ) -> Result<Response<ListPointsOfInterestResponse>, Status> {
+        unimplemented!();
+    }
+}
+
+pub struct NativeApplication {
+    _handle: JoinHandle<()>,
```
tiziano88, line 74: What's the point of this field? Does the spawned task get killed if this goes out of scope? Either way, please add a comment.

experimental/benchmark/src/database.rs
```diff
@@ -0,0 +1,134 @@
...
+                }),
+            };
+            entries.insert(format!("{}", id), entry);
+        }
+        Self {
+            points_of_interest: PointOfInterestMap { entries },
+        }
+    }
+
+    pub fn serialize(&self) -> String {
```
tiziano88, line 88: I'm confused as to what's going on here, it seems that it is serializing and deserializing the db, and also taking some timing and printing it out, but returning the serialized version of the XML file, is that correct? Could you add a comment, especially since it is so coupled with various functionality?

experimental/benchmark/src/database.rs
```diff
@@ -0,0 +1,134 @@
...
+            .replace("<Station>", "<station>")
+            .replace("</Station>", "</station>");
+
+        // Measure database parsing time.
+        let start = Instant::now();
+        let parsed_database: StationDatabase =
+            quick_xml::de::from_str(&database).expect("Couldn't deserialize database");
+        let duration = start.elapsed();
+        info!(
+            "Database parsing time: {:?} for {} entries ({} bytes)",
```
tiziano88, line 126: How is this information useful? AFAICT it is always done on the host natively, or am I missing something? how do you get to run this in Oak via Wasm?

tiziano88 (approved):

examples/trusted_database/proto/trusted_database.proto
```diff
@@ -30,15 +30,32 @@ message PointOfInterest {
...
+  map<string, oak.examples.trusted_database.PointOfInterest> entries = 1;
+}
+
+message GetPointOfInterestRequest {
+  string id = 1;
+}
+
+message GetPointOfInterestResponse {
+  PointOfInterest point_of_interest = 1;
+}
```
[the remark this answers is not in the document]
ipetr0v: I think right now we are encapsulating responses in all of the examples

experimental/benchmark/src/database.rs
ipetr0v: Added a comment

experimental/benchmark/src/application/native.rs
ipetr0v: Updated
Using `handle` to stop the `tokio` task

experimental/benchmark/src/application/mod.rs
tiziano88: I understand, but it is an abstraction over a client, not over the application.

experimental/benchmark/src/application/native.rs
tiziano88: I don't think the handle stops the task in any way, it just waits for it to stop.

experimental/benchmark/src/database.rs
tiziano88: But I don't understand, isn't this code always running on the host natively? How / when does it get compiled to Wasm? Apologies if I am missing something obvious.

ipetr0v: Yes, it measures the native XML parsing time.
And `main.rs` has code for measuring Oak runtime loading time.

experimental/benchmark/src/application/native.rs
ipetr0v: Done

experimental/benchmark/src/application/mod.rs
ipetr0v: Renamed to `ApplicationClient`

PR #1567, opened 2020-10-08
Ensure HandleDestroyed event fires before ChannelDestroyed

juliettepretot:
Previously there was bug in which the channel was dropped before the `HandleDestroyed` event fired. This led to an oddly ordered list of events, in which handles referenced channels that no longer existed. 

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/lib.rs
+++ b/oak_runtime/src/lib.rs
@@ -411,27 +411,18 @@ impl Runtime {
         let mut node_infos = self.node_infos.write().unwrap();
         let node_info = node_infos.get_mut(&node_id).expect("Invalid node_id");
 
-        let event_details = HandleDestroyed {
-            node_id: node_id.0,
-            handle,
-            channel_id: node_info
-                .abi_handles
-                .get(&handle)
-                .ok_or(OakStatus::ErrBadHandle)?
-                .get_channel_id(),
-        };
+        match node_info.abi_handles.remove(&handle) {
+            Some(half) => {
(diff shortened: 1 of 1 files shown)


oak_runtime/src/lib.rs
```diff
@@ -411,27 +411,18 @@ impl Runtime {
...
-
-        let result = node_info
-            .abi_handles
-            .remove(&handle)
-            .ok_or(OakStatus::ErrBadHandle)
-            .map(|_half| ());
-
-        if result.is_ok() {
-            self.introspection_event(EventDetails::HandleDestroyed(event_details))
-        };
```
juliettepretot, line 432: Previously the `ChannelHalf` (and associated channel) was dropped before firing the event

oak_runtime/src/lib.rs
```diff
@@ -411,27 +411,18 @@ impl Runtime {
...
-                .get(&handle)
-                .ok_or(OakStatus::ErrBadHandle)?
-                .get_channel_id(),
-        };
-
-        let result = node_info
-            .abi_handles
-            .remove(&handle)
-            .ok_or(OakStatus::ErrBadHandle)
-            .map(|_half| ());
```
juliettepretot, line 428: Imho using an explicit match is also more readable than combining `ok_or` & `map` to convert the option returned by `remove` to the desired `Result`

daviddrysdale (approved):

Issue #1067, opened 2020-06-02
Web-based Oak Runtime emulator


tiziano88:
At a basic level, the Oak Runtime runs WebAssembly modules, and exposes to them [functions](https://github.com/project-oak/oak/blob/master/docs/abi.md#host-functions) that allow them to interoperate with the Runtime itself and with other Nodes / Channels. Since browsers already have a WebAssembly engine built-in, and quite advanced ones at that, and also have debugging capabilities for WebAssembly modules when running it, it would be neat to have essentially an emulator / parallel implementation of the Oak Runtime for the browser.

To start with, it should be able to load an Oak [ApplicationConfiguration](https://github.com/project-oak/oak/blob/master/oak/proto/application.proto) protobuf file, which contains the list of Wasm modules, plus some metadata about how to create the first Wasm node out of that. For the web version, the runtime would basically just instantiate the first Wasm module in a native browser Wasm context, and expose to it the necessary host functions (initially just stubs, to make sure it actually loads, then gradually implement them with the actual functionality). For instance, once the runtime supports creating new nodes, then we can allow creating a log node, and see some output from the running Oak Application to the JS console of the browser. Then we can implement a web UI that allows populating a protobuf (or just parse it from JSON), and simulate sending that to the loaded application.

If this all works out, then we can even consider promoting this as to a first-class programming model Oak. We should consider the implications of Information Flow Control in the context of the browser at that point, but for the time being I would be inclined to just ignore any IFC-related functionality (just allow channel and nodes to talk freely to each other, and ignore labels).

juliettepretot:
Thanks for writing this up! It makes a lot more sense now. :) 

I quite like it, not just because of its utility for debugging, but also because writing an emulator seems like a pretty friendly way to incrementally learn more about the Oak architecture.

juliettepretot:
Yeah would love to work on that

tiziano88:
Awesome! How about we gather some requirements, either here or in a separate doc, about technology to use (is Typescript part of the plan?), milestones (with expected functionality at each of them).

BTW, it occurred to me that the current Rust Oak Runtime could itself be compiled to Wasm, and be the foundation of the emulator, rather than re-implementing it from scratch in another language. If that works, it may massively simplify the scope of this task, in that at least the inner working of the Runtime emulator would rely on the actual code of the Runtime itself.

cc @anghelcovici

tiziano88:
cc @daviddrysdale

tiziano88:
[...]

PR #1569, opened 2020-10-08
Initial web emulator

tiziano88:
Very barebones, only stubs all Oak ABI calls instead of doing anything
useful with them.

Demo:
https://screencast.googleplex.com/cast/NTc0OTA3NTU3ODE5MTg3Mnw1NmE2ZGUyNi1lMw

Ref: #1067

# Checklist

      construction.

### DIFF
--- a/experimental/web/index.html
+++ b/experimental/web/index.html
@@ -1,0 +1,178 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8" />
+    <title>Oak emulator</title>
+</head>
+
+<body>
+    <script type="module">
+        import Vue from 'https://cdn.jsdelivr.net/npm/vue@2.6.12/dist/vue.esm.browser.js'
+        function init() {
+            var app = new Vue({
+                el: '#app',
+                data: {
+                    // Exports of the loaded Module.
+                    exports: [],
+                    // Import of the loaded Module.
+                    imports: [],
+                    // Oak ABI calls trace.
+                    trace: [],
+                    // Loaded module.
+                    module: null,
+                    // Loaded module instance.
+                    instance: null,
+                },
+                methods: {
+                    // Read the specified file and load it as a Wasm module.
+                    readFile: function (e) {
+                        const file = e.target.files[0];
+                        if (!file) {
+                            console.log('no file selected');
+                            return;
+                        }
+                        const reader = new FileReader();
+                        reader.onload = this.loadModule;
+                        reader.readAsArrayBuffer(file);
+                    },
+                    // Parse the provided ArrayBuffer as a Wasm module and load it as an Oak module,
+                    // providing the necessary imports.
+                    loadModule: async function (e) {
+                        const contents = e.target.result
+                        console.log('file loaded');
+                        this.module = await WebAssembly.compile(contents);
+                        this.exports = WebAssembly.Module.exports(this.module);
+                        this.imports = WebAssembly.Module.imports(this.module);
+                        await this.instantiate();
+                    },
+                    instantiate: async function () {
+                        const statusOk = 1;
+                        // Provide a mock implementation of some of the Oak ABI functions.
+                        // Mostly these just log their argument to the trace, and return a
+                        // successful status without actually doing anything.
+                        const importObject = {
+                            oak: {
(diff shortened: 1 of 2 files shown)


experimental/web/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8" />
+    <title>Oak emulator</title>
+</head>
+
+<body>
+    <script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js"></script>
```
juliettepretot, line 10: Interesting that vue recommends this as the default way to import it, I wonder what happens when they release a breaking version change. 

Personally I'd recommend to import it as a module, which is versioned & avoids polluting the global namespace.

This would go into the module tag below:

```
import Vue from 'https://cdn.jsdelivr.net/npm/vue@2.6.12/dist/vue.esm.browser.js'
```

via https://vuejs.org/v2/guide/installation.html#CDN

experimental/web/index.html
```diff
@@ -0,0 +1,152 @@
...
+                    // providing the necessary imports.
+                    loadModule: async function (e) {
+                        const contents = e.target.result
+                        console.log('file loaded');
+                        this.module = await WebAssembly.compile(contents);
+                        this.exports = WebAssembly.Module.exports(this.module);
+                        this.imports = WebAssembly.Module.imports(this.module);
+                        // Bind this to a variable so that it can be referred to from the
+                        // implementation of the Oak ABI functions.
+                        const v = this;
```
juliettepretot, line 49: This shouldn't be needed, as arrow functions use a "lexical" this that refers to the outer scope.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

experimental/web/index.html
```diff
@@ -0,0 +1,152 @@
...
+                        const statusOk = 1;
+                        // Provide a mock implementation of some of the Oak ABI functions.
+                        // Mostly these just log their argument to the trace, and return a
+                        // successful status without actually doing anything.
+                        const importObject = {
+                            oak: {
+                                wait_on_channels: (buf, count) => {
+                                    const ret = statusOk;
```
juliettepretot, line 57: What is `ret` a shorthand for?

juliettepretot (approved):

experimental/web/index.html
tiziano88: Done, thanks!

experimental/web/index.html
tiziano88: Ah thanks, that explains some other weird behaviour I noticed elsewhere when switching from `function` to `=>`, in which `this` points to the wrong thing. I thought the two were equivalent.

experimental/web/index.html
tiziano88: "return", but happy to change to something else if there is any convention?

tiziano88 (commented):
Thanks Juliette!

BTW, the reason why I went with Vue is that it seemed simpler than React to get something self-contained in a single page; a limitation of loading this in Chrome without a web server is that it cannot access other files, even in the same directory, so it's not possible to split js or css files out, and Vue seems to fit in better (though it is the first time I use it, so I'm not even sure I'm holding it right). I would be more than happy for you to rebuild all of this from scratch properly, I just wanted to convince myself that it is technically possible at least (which looks like it is), before letting others chase this :)

experimental/web/index.html
juliettepretot: Oh that makes sense. :) Afaik there isn't a JS convention, but I'd suggest `result` (we often use that in the Rust runtime) or maybe `returnValue`. 

My personal nitty opinion, but imho, except for exceedingly common cases, the time savings of abbreviated var names rarely justify friction introduced when reading them.

juliettepretot:
@tiziano88 Vue's a totally fair choice for this use-case, imho it works quite well. React is great for larger scale apps, but Vue's much nimbler. But yeah happy to refactor it too if we want to invest in it further.

experimental/web/index.html
tiziano88: renamed to `status`, I think here it makes sense to be more specific :)

daviddrysdale (approved):
Neat!

PR #1570, opened 2020-10-09
Add an option to compile release Wasm in tests

ipetr0v:
This change adds an option to compile release Wasm in tests (necessary for benchmarking).

### DIFF
--- a/sdk/rust/oak_tests/src/lib.rs
+++ b/sdk/rust/oak_tests/src/lib.rs
@@ -64,16 +64,6 @@ pub fn compile_rust_wasm(
     }
 
     Command::new("cargo")
-        // .args(&[
-        //     "build",
-        //     &format!(
-        //         "--target-dir={}",
-        //         target_dir.to_str().expect("invalid target dir")
-        //     ),
-        //     &format!("{}", if let Profile::Release = profile { "--release" } else { "" }),
-        //     "--target=wasm32-unknown-unknown",
-        //     &format!("--manifest-path={}", cargo_path),
-        // ])
         .args(args)
         .env_remove("RUSTFLAGS")
         .spawn()
--- a/sdk/rust/oak_tests/src/lib.rs
+++ b/sdk/rust/oak_tests/src/lib.rs
@@ -33,25 +33,48 @@ use tonic::{
     Request,
 };
 
+pub enum Profile {
+    Release,
+    Debug,
+}
+
 // TODO(#544): re-enable unit tests of SDK functionality
 
 /// Uses cargo to compile a Rust manifest to Wasm bytes.
-pub fn compile_rust_wasm(cargo_path: &str, module_wasm_file_name: &str, release: bool) -> anyhow::Result<Vec<u8>> {
+pub fn compile_rust_wasm(
+    cargo_path: &str,
+    module_wasm_file_name: &str,
+    profile: Profile,
+) -> anyhow::Result<Vec<u8>> {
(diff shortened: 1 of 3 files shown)


sdk/rust/oak_tests/src/lib.rs
```diff
@@ -36,7 +36,7 @@ use tonic::{
 // TODO(#544): re-enable unit tests of SDK functionality
 
 /// Uses cargo to compile a Rust manifest to Wasm bytes.
-pub fn compile_rust_wasm(cargo_path: &str, module_wasm_file_name: &str) -> anyhow::Result<Vec<u8>> {
+pub fn compile_rust_wasm(cargo_path: &str, module_wasm_file_name: &str, release: bool) -> anyhow::Result<Vec<u8>> {
```
tiziano88, line 39: This is not very readable when invoked, could you create something like

```rust
enum Profile {
  Release,
  Debug,
}
```

tiziano88 (approved):

sdk/rust/oak_tests/src/lib.rs
ipetr0v: Done

ipetr0v:
Updated, because of the:
```
error: Found argument '' which wasn't expected, or isn't valid in this context
```

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -33,24 +33,38 @@ use tonic::{
...
+        format!(
+            "--target-dir={}",
+            target_dir.to_str().expect("invalid target dir")
+        ),
+        "--target=wasm32-unknown-unknown".to_string(),
+        format!("--manifest-path={}", cargo_path),
+    ];
+    if let Profile::Release = profile {
+        args.push("--release".to_string());
+    }
```
tiziano88, line 64: This should probably be a `match`, so that if we add more values to the enum, the compiler will warn us that we need to handle all cases here too.

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -33,24 +33,38 @@ use tonic::{
...
+    Release,
+    Debug,
+}
+
 // TODO(#544): re-enable unit tests of SDK functionality
 
 /// Uses cargo to compile a Rust manifest to Wasm bytes.
-pub fn compile_rust_wasm(cargo_path: &str, module_wasm_file_name: &str) -> anyhow::Result<Vec<u8>> {
+pub fn compile_rust_wasm(
+    cargo_path: &str,
```
tiziano88, line 45: Not this PR, but could you rename this to `manifest_path`? It is not a path to cargo.

tiziano88 (approved):

sdk/rust/oak_tests/src/lib.rs
ipetr0v: Done
https://github.com/project-oak/oak/pull/1571

sdk/rust/oak_tests/src/lib.rs
ipetr0v: Done

sdk/rust/oak_tests/src/lib.rs
tiziano88: FYI I just meant that the issue was not introduced by this PR, not that it needs to be fixed in a separate PR.

ipetr0v: Got it
I thought you meant a separate PR

PR #1573, opened 2020-10-09
Parse protobuf messages in web emulator

tiziano88:
This uses the https://github.com/protobufjs/protobuf.js library and
loads the proto descriptors from GitHub.

Ref #1067

# Checklist

      construction.

### DIFF
--- a/experimental/web/index.html
+++ b/experimental/web/index.html
@@ -6,8 +6,10 @@
   </head>
 
   <body>
+    <script src="https://raw.githack.com/protobufjs/protobuf.js/v6.10.1/dist/protobuf.js"></script>
     <script type="module">
       import Vue from 'https://cdn.jsdelivr.net/npm/vue@2.6.12/dist/vue.esm.browser.js';
+
       function init() {
         var app = new Vue({
           el: '#app',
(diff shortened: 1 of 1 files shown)


experimental/web/index.html
```diff
@@ -150,9 +170,8 @@
                       ) +
                       ') -> ' +
                       status +
-                      '\n  label: [' +
-                      label +
-                      ']';
+                      '\n  label: ' +
```
juliettepretot, line 173: It would probably make sense to write these as tagged string literals for readability. Eg `hi ${name}. That way we can omit the `+`s to join strings, and new lines are supported without `\n`

experimental/web/index.html
```diff
@@ -150,9 +170,8 @@
...
+                      JSON.stringify(label);
```
juliettepretot, line 174: It seems safer to convert the message instance to a normal object via `.toObject` before stringifying it, to make sure we don't stringify fields related to the functioning of protobuf.js.

experimental/web/index.html
```diff
@@ -6,8 +6,10 @@
   </head>
 
   <body>
+    <script src="https://raw.githack.com/protobufjs/protobuf.js/v6.10.1/dist/protobuf.js"></script>
```
juliettepretot, line 9: I've used this unofficial protobuf lib before for the node.js example before, and while it's convenient due dynamically reading proto files, I found it to be somewhat buggy. 

For instance it sometimes omits number fields with the value `0`. (Common JS pitfall, where people check whether a field exists via checking for truthyness instead of undefinedness.

experimental/web/index.html
```diff
@@ -46,14 +48,30 @@
...
-              const statusOk = 1;
+              // Load proto files from GitHub. We cannot refer to local files unless they are served
+              // via a web server first.
+              // This means that if protobuf definitions are modified locally, this script will not
+              // be able to use the updated definitions.
+              const root = await protobuf.load([
+                'https://raw.githubusercontent.com/project-oak/oak/main/oak_abi/proto/application.proto',
+                'https://raw.githubusercontent.com/project-oak/oak/main/oak_abi/proto/label.proto',
+                'https://raw.githubusercontent.com/project-oak/oak/main/oak_abi/proto/oak_abi.proto',
+              ]);
```
juliettepretot, line 59: Hmmm I see why you went that route, but it's worth mentioning that the overhead of avoiding a web server quickly starts outweighing the overhead of a web server. 

In this case adding one would be simple, we can use npx to load an webserver on demand & use to serve the file. 

A bash file to serve this would look like so
```sh
#!/bin/bash
# Navigate to the project root
cd ../../
# Serve the web emulator
npx servor ./ ./experimental/web/index.html
```

juliettepretot (approved):
Doesn't seem terrible per se, we use protobuf.js to parse labels, and then stringify them for tracing. 

I would advise to use a local web server though, imho loading files from the GitHub repo feels like a lot of overhead.

experimental/web/index.html
tiziano88: Done.

experimental/web/index.html
tiziano88: It seems fine to just stringify those objects, they even do it in their own doc: https://www.npmjs.com/package/protobufjs#using-the-js-api
In practice I don't seem to get other fields in the stringified representation anyways.

experimental/web/index.html
tiziano88: Indeed, or maybe I could convince someone (:eyes:) to rewrite this stuff properly at some point...
For now I'll leave it out of this PR anyways, but let's chat about it offline.

experimental/web/index.html
tiziano88: Good to know, yes I only used this because it can parse proto files at runtime, but I expect a proper version of this app would just pre-compile them from sources anyways.

experimental/web/index.html
juliettepretot: All good then :)

experimental/web/index.html
juliettepretot: SGTM

PR #1575, opened 2020-10-11
Finish proof of step_implies_lowproj_steps_leq

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -93,6 +93,18 @@ Local Ltac crush := repeat crush_step.
 
 Hint Resolve multi_system_ev_refl multi_system_ev_tran : multi.
 
+
+(* Hints for [eauto with unwind] *)
+Hint Resolve state_upd_chan_unwind chan_append_unwind chan_low_proj_loweq
+    chan_low_proj_idempotent state_upd_node_unwind set_call_unwind 
+    state_low_proj_loweq : unwind.
+Hint Extern 4 (node_low_eq _ _ _) => reflexivity : unwind.
+Hint Extern 4 (chan_low_eq _ _ _) => reflexivity : unwind.
+(* meant to be case where we have (cleq ch (proj ch) ) and want to swap order *)
+Hint Extern 4 (chan_low_eq _ _ (chan_low_proj _ _)) => symmetry : unwind.
+Hint Extern 4 (state_low_eq _ _ (state_low_proj _ _)) => symmetry : unwind.
+
+
 Definition is_init(t: trace) := length t = 1.
 
 Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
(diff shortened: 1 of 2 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -165,13 +165,112 @@ Proof.
...
+                assert (Hnproj: n = (node_low_proj ell n)) by
+                    repeat (eapply flows_node_proj || symmetry|| congruence ).
+                assert (Hcproj: (chan_low_proj ell ch) = ch) by
+                    eauto using flows_chan_proj, ord_trans.
+                pose proof (state_nidx_to_proj_state_idx ell _ _ _
+                    ltac:(eauto)) as Hn_idx_s1proj.
+                eexists. exists (n <--- msg).
+                split. econstructor.
+                eauto. eauto.
+                replace (node_low_proj ell n) with n by auto; auto.
```
jadephilipoom, line 178: Could this be a lemma? Then you could `rewrite !node_low_proj_id in * by auto`, which might be a little cleaner than repeated calls to `replace`.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -165,13 +165,112 @@ Proof.
...
+                replace n with (node_low_proj ell n) by auto; auto.
+                auto.
+                replace (handle_fresh (state_low_proj ell s1))
+                    with (handle_fresh s1)
+                    by (symmetry; eapply proj_pres_handle_fresh); auto.
+                unfold s1''. subst.
+                eapply set_call_unwind. eapply state_upd_node_unwind.
+                reflexivity. eapply state_upd_node_unwind. reflexivity.
+                eapply state_upd_chan_unwind. reflexivity.
+                symmetry. eapply state_low_proj_loweq.
```
jadephilipoom, line 224: Does `eauto with unwind` work here? (You might need to add `state_low_proj_loweq` to its list of hints.)

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/PossibilisticNI.v
jadephilipoom: Oh, actually I think `flows_node_proj` is the lemma, right?

aferr: Right, flows_node_proj is the lemma. I use the lemma to make Hnproj and then do all these extra replaces. What you suggested looks cleaner.

What does " ! " do in rewrite? I can't find it in the docs too easily.

jadephilipoom: It's like regex `+`; means "rewrite one or more times". The [manual entry for rewrite](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:tacn.rewrite) very briefly mentions this syntax and other variations like `?` (which is "zero or more times").

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: I'm using eauto with unwind now. It doesn't quite go as far as one would hope for the WriteChannel case. It seems to get stuck where it needs to apply one of either chan_low_proj_loweq or chan_low_proj_idempotent. Both are in the hint database now, though.

jadephilipoom: When in doubt, you can always add a `Hint Extern` with the goal pattern you see and the lemma you want to apply (not blocking for this PR, though)!

Issue #1537, opened 2020-09-30
Add `bucket_id` to PSI example


ipetr0v:
Since currently Private Set Intersection example blocks new set submissions after someone requests the intersection value, it doesn't support multiple client variants running in the CI.

So we can add a `bucket_id` that will represent each specific intersection.

PR #1576, opened 2020-10-12
Add set ID to Private Set Intersection

ipetr0v:
This change:
- Adds a set ID to each PSI request
- Re-enables PSI C++ client

Fixes https://github.com/project-oak/oak/issues/1537


examples/private_set_intersection/example.toml
```diff
@@ -16,10 +16,11 @@ additional_args = [
 
 [clients]
 rust = { Cargo = { cargo_manifest = "examples/private_set_intersection/client/rust/Cargo.toml" }, additional_args = [
+  "--set-id=rust",
```
tiziano88, line 19: Perhaps add a comment somewhere around here explaining why these have different ids, and what they mean, etc.

examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -49,32 +49,46 @@ oak::entrypoint!(oak_main => |_in_channel| {
...
+    values: HashSet<String>,
     /// Number of contributed private sets.
     set_count: u64,
     /// The intersection is locked and new contributions are discarded.
     locked: bool,
 }
 
+#[derive(Default)]
+struct Node {
+    /// Map from set ID to `SetIntersection`.
```
tiziano88, line 63: Could you explain a bit more here? e.g. that this allows multiple sets of clients to compute their own intersections, also explain what the security characteristics are, in particular I think currently just specifying a different set id does not guarantee that data from different clients is kept separate, right? That would require some more sophisticated labelling.

examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -49,32 +49,46 @@ oak::entrypoint!(oak_main => |_in_channel| {
...
+    sets: HashMap<String, SetIntersection>,
+}
+
 impl PrivateSetIntersection for Node {
     fn submit_set(&mut self, req: SubmitSetRequest) -> grpc::Result<()> {
-        if self.locked {
+        let mut current_set = self
+            .sets
+            .entry(req.set_id)
+            .or_insert(SetIntersection::default());
```
tiziano88, line 72:
```suggestion
            .or_default();
```

examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -49,32 +49,46 @@ oak::entrypoint!(oak_main => |_in_channel| {
...
+        if current_set.set_count < SET_THRESHOLD {
+            let submitted_set = req.values.iter().cloned().collect::<HashSet<_>>();
+            current_set.values = if current_set.values.is_empty() {
+                submitted_set
+            } else {
+                current_set
+                    .values
+                    .intersection(&submitted_set)
+                    .cloned()
+                    .collect()
```
tiziano88, line 90: Where is the code that increments the sample count now?

If it's missing, how can it work?

tiziano88 (approved):

examples/private_set_intersection/module/rust/src/lib.rs
ipetr0v: Done

examples/private_set_intersection/module/rust/src/lib.rs
ipetr0v: Done

PR #1578, opened 2020-10-12
Finish low_proj_steps_implies... + ..._unwind_t; tactics

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -77,8 +77,10 @@ Local Ltac logical_simplify :=
          end.
 
 Local Ltac step_econstruct :=
-    match goal with
+    repeat match goal with
         | H: _ = ncall _ |- _ => progress rewrite <- H
+    end;
+    lazymatch goal with
         | |- step_system_ev _ _ _ => econstructor; eauto
         | |- step_node_ev _ _ _ _ _  => econstructor; eauto
         | |- step_system _ _ => econstructor; eauto
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -422,8 +422,18 @@ Theorem possibilistic_ni_unwind_t: forall ell t1 t2 t1',
     (step_system_ev_t t2 t2') /\
     (trace_low_eq ell t1' t2')).
 Proof.
-    (* Should make use of possibilistic_ni_1step *)
-Admitted. (* WIP *)
+    intros. 
+    inversion H; crush.
+    -  (* starts from empty trace *)
+        exfalso. eapply no_steps_from_empty; eauto.
+    - (* non-empty trace *)
+        inversion H0; inversion H4; crush.
+        pose proof (possibilistic_ni_1step _ s ys s' _
+            ltac:(eauto) ltac:(eauto)) as [s2' [e2 [H2step [Hsleq Heleq]]]].
+        eexists ((s2', e2) :: (ys, ye) :: t3); repeat split; crush.
+        econstructor; crush; eauto. 
+        econstructor; crush; eauto.
+Qed.
 
 Theorem possibilistic_ni: conjecture_possibilistic_ni.
 Proof.
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -76,12 +76,22 @@ Local Ltac logical_simplify :=
          | H : ?x = ?x |- _ => clear H
          end.
 
+Local Ltac step_econstruct :=
+    match goal with
+        | H: _ = ncall _ |- _ => progress rewrite <- H
+        | |- step_system_ev _ _ _ => econstructor; eauto
+        | |- step_node_ev _ _ _ _ _  => econstructor; eauto
+        | |- step_system _ _ => econstructor; eauto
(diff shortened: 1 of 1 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -298,33 +298,69 @@ Proof.
...
-                eauto with unwind.
+                crush. unfold s1'', ch'. eauto with unwind.
             + (* ReadChannel *)
-                admit.
+                inversion H3; crush.
+                pose proof (uncons_proj_chan_s _ _ _ _ ltac:(eauto))
+                    as [ch2 [H_ch'_idx H_ch2_proj]].
+                unfold n'0 in *; crush.
+                unfold s1'' in *.
+                rewrite flows_node_proj in *; eauto; crush.
```
jadephilipoom, line 318: I'm seeing this line in every case; is it possible to do this rewrite before the inversion splits it into cases? Or do you need one of the preconditions you get from inversion?

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -298,33 +298,69 @@ Proof.
...
-                rewrite Heqs2''. eapply SystemEvStepNode; eauto; subst. 
-                replace (ncall n') with (ncall (node_low_proj ell n')) by
-                    (erewrite flows_node_proj; try eassumption; try congruence).
-                rewrite <- H1. econstructor. erewrite flows_node_proj; assumption.
-                assert (n' = n). { 
-                    rewrite H6 in H0. inversion H0. symmetry. eapply flows_node_proj. auto.
-                }
-                eapply SWriteChan; try congruence; try assumption;
-                    try (erewrite <- chan_projection_preserves_lbl; eassumption).
+                repeat try ((econstructor; eauto) || rewrite <- H1).
```
jadephilipoom, line 309: `repeat try` is a bit dangerous. `repeat` keeps going until it gets a tactic failure, and `try` never fails. I'm guessing that this tactic solves the subgoal, because otherwise you'd get an infinite loop! I think the behavior you want here is just `repeat`, which would run the tactic over and over until it fails, and then stop.

Also, this is a great place to put a `repeat match goal` statement, which will prevent the tactic from running `econstructor` on an unexpected goal or relying on the autogenerated `H1` name. It would look something like:
```suggestion
                repeat match goal with
                           | |- step_node _ _ _ _ => econstructor; eauto
                           | H : <whatever H1 looks like, can have underscores> |- _ => rewrite <-H
                           end.
```
As an added advantage, this pattern makes it easier to see what's going on without stepping through the proof interactively, because it specifies the kind of goal or hypothesis you're expecting.

jadephilipoom (approved):
Nice! The proofs here are very clean and concise.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Since it looks like (econstruct; eauto) happens all over the place, I made a tactic following your suggestion and put it in crush. See what you think.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Good point. I put it in a semicolon after the inversion that splits into cases on the call. I also tried to cleanup the proofs in general a bit more.

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -76,12 +76,22 @@ Local Ltac logical_simplify :=
          | H : ?x = ?x |- _ => clear H
          end.
 
+Local Ltac step_econstruct :=
+    match goal with
+        | H: _ = ncall _ |- _ => progress rewrite <- H
```
jadephilipoom, line 81: This tactic is a great idea! I have a couple suggestions to make it even better :)

Assuming the intent here is to first rewrite any `_ = ncall _` hypotheses and *then* try to call `econstructor` (and you maybe don't want the rewrite if the `econstructor` doesn't work),  you could consider making this structure explicit:

```suggestion
    repeat match goal with
              | H: _ = ncall _ |- _ => progress rewrite <- H
              end;
        lazymatch goal with
```
...and then, you can change the match that performs `econstructor` into a `lazymatch`, which will give you better error messages here. The difference between `lazymatch` and `match` is that a `match` tries *all* clauses that pattern-match and fails with the uninformative `no matching clauses for match` none of them succeed. But `lazymatch` finds the *first* clause with a pattern that matches and fails entirely if the associated tactic fails, passing on the error message.

For example, if your goal was `step_system_ev x y z`, but `econstructor` on this goal failed with the message `cannot unify 'x' with 'foo'`, then `match goal with |- step_system_ev  _ _ _ => econstructor end` would give you `no matching clauses for match`, and `lazymatch goal with |- step_system_ev _ _ _ => econstructor end` would give you `cannot unify 'x' with 'foo'`. The distinction makes a big difference when debugging!

So you use `match` when you want to try all the clauses (useful especially for tactics like `crush`, when you want to just try a bunch of tactics in succession and use whichever succeeds first), and `lazymatch` if you expect a given goal to match exactly one clause of the match, and want to see the error if it fails.

jadephilipoom (commented):
Awesome idea to make the separate tactic! It's cool to see that just the one pattern replaced so much code in the proofs.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: That does sound like it will make debugging better. Thanks also for the very informative explanation as always!

PR #1581, opened 2020-10-13
Apply all constructors first to automatically construct new state

jadephilipoom:
Based on conversation with @aferr, this change implements a tactic that applies all the system/node constructors before any of the generated subgoals are solved. That way, we can use `eexists` instead of manually repeating the new state based on the constructor, which should help the proof keep pace with future changes. ~Also, it seems to help out `eauto with unwind`, which now solves the subgoal it was struggling with before!~ ~(Edit: never mind, CI says no so fixing)~ (Edit again: I accidentally deleted a `Hint Extern`, now it works again!)

# Checklist

      construction.


aferr (approved):
LGTM!

Issue #1452, opened 2020-09-09
Implement downgrading privilege for user principals authenticated over gRPC / HTTP


tiziano88:
Once the challenge-response mechanism is implemented (as per #1357), the Oak Runtime should then use the result of the authentication step to determine the downgrading privilege of the client connecting to it.

The most elegant way of doing this would be to create a sort of "virtual" node per invocation that has exactly the downgrading privilege corresponding to the provided authenticated client.

For instance:

- client connects over gRPC
- client obtains challenge from Oak Runtime over gRPC
- client signs challenge with its private key
- client sends data to the Application over gRPC
  + data is labelled with an arbitrary label, which is independent of the authentication
  + additionally, the client provides the signed challenge as an extra metadata field, to prove its own identity (and does this for any subsequent invocation over the same connection)
- Runtime receives the request, validates the signed challenge (if any is provided)
- if one is provided and it matches the public key, then it creates a new pseudo-Node dedicated to that invocation, with the privilege to declassify the principal corresponding to the public key of the provided signed challenge (more details are needed here, I'm handwaving for now)

rbehjati:
In the case of the HttpServerNode this requires replacing the `Pipe` with the virtual node as follows: 
1. After receiving a request from the user, the HttpServerNode creates a channel for communication with the virtual node.
2. The HttpServerNode creates the virtual node, setting its privilege to the user's identity.
3. The HttpServerNode sends a NodeMessage, containing the user's request as the data, the `invocation_channel` (between the HttpServerNode and the actual server node) as one handle, and an `outer_response_writer` as another handle. 
4. The virtual node creates `request_reader` and `response_writer` channels with correct tags as in #1428 (this is currently done inside `Pipe`):
https://github.com/project-oak/oak/blob/a6e62945aa4ad0fdf16c038f081faf9f471c0a63/oak_runtime/src/node/http/mod.rs#L349-L354

5. The virtual node send the request over the `request_reader` channel.
6. The virtual node then bundles the the read-half of the `request_reader`, and the write half of the `response_writer` in an invocation and sends that over the `invocation_channel` to the actual server node. 
7. The virtual node waits for receiving the response on the read-half of the `response_writer`, declassifies it and sends a NodeMessage with the HTTP response, back to the HttpServerNode on the `outer_response_writer` channel to the HTTP server p-N. 

@tiziano88 Does this make sense? This basically doubles the number of channels that we create per request.

rbehjati:
[...]

Issue #1428, opened 2020-09-03
Set the user's identity as the confidentiality tag in invocation channels for HTTP and gRPC server nodes


rbehjati:
Each incoming gRPC/HTTP invocation is represented by a message containing two Channel handles: a "request receiver" and a "response sender". 

Currently the "response sender" gets the public_untrusted label. However, as explained in the [docs](https://github.com/project-oak/oak/blob/main/docs/concepts.md#grpc-and-user-labels), it should have a confidentiality component that is set based on the identity of the user. This is needed to avoid any accidental data leaks.

tiziano88:
@rbehjati I believe your changes already fix this, so assigning to you, but let me know if that's not the case.

rbehjati:
Yes. But partially. I'll update the description here, and can create a separate issue to track setting the user identity as the integrity label. Do we have a top-level issue for integrity labels?

rbehjati:
Closing this now, but the work on identity-related integrity labels is going to be part of IFC v1, and #1715.

Issue #1357, opened 2020-08-14
Implement support for generic challenge-response style authentication


conradgrobler:
At the moment we use bearer tokens to do pseudonymous authentication. It would be better not to pass secret tokens or keys to the server to do authentication, seeing that a leaked secret allows impersonation. Challenge-response protocols allow for authentication without sending secrets to the server, but cannot be implemented by just attaching metadata to each call due to the interactive nature.

Mutually authenticated TLS connections with client certificates provide an option for authenticating clients, but it is somewhat inconvenient and requires a root of trust for the client certificates (typicall the subject of the certificate is seen as the identity, rather than using the public key as an identifier).

A more generic option is to provide a gRPC-based challenge response authentication mechanism that allows a client to 'upgrade' an existing connection with an identity after the TLS connection has been established.

The idea is roughly

- The client establishes a TLS connection to the server
- The client asks for a challenge via a gRPC call
  - In the OpenID Connect case this includes the OIDC server details
  - For a pseudonymous signature-based approach this will just be a random byte string
- The client produces a response and sends it to the server via another gRPC call to attach an identity to the current connection
  - For OIDC this is the ID token
  - For a pseudonymous signature-based approach the client will sign the challenge and send the signature and public key. The public key will act as an identifier for the user if the signature is validated.
- All future calls on the connection will then have the privilege associated with the assoicated identity
- It should be possible for a client to attach multiple identities to a single connection by repeating the process

A prerequisite for this is #1020 (supporting per-connection state on the gRPC server node), which has been on hold for a while. The server must be able to associate the challenge sent to the client and the received identity with the current connection.

tiziano88:
An alternative approach may be for the client to just sign the content of each outgoing request (including headers and body).

Advantages:

- no need to keep state on the server
- can sign each request differently if necessary
- provides non-repudiation for requests (server can store the payload and signature as proof of the request actually coming from the client with that public key)

Disadvantages:

- vulnerable to replay attacks (though only by the other endpoint of the TLS connection, so probably not a big deal)
- more logic on the client
- may not work with streaming requests

tiziano88:
BTW in any case let's make sure that we use the same signing / verification logic that @ipetr0v is using for the wasm module signatures

ipetr0v:
We also can implement [Kerberos](https://tools.ietf.org/html/rfc4120) like client authentication (token distribution server). And this server could be a part of Oak Runtime, just on different port.
[...]

PR #1586, opened 2020-10-14
Use challenge response for authentication

rbehjati:
Summary of the changes:
- Adds a `user` module, which defines a virtual node (#1452) named `UserNode`, to `oak_runtime/node/http`. An instance of `UserNode` is created for every incoming HTTP request. This node gets the identity of the user as its declassification and endorsement privileges. The functionality of `UserNode` is mostly copied from `HttpServerNode`, in particular, the function `inject_http_request`. In `HttpServerNode` the function `inject_http_request` is updated to instead create the `UserNode` instance and forward the HTTP request and the required invocation channels to it. 
  - Something similar is needed for gRPC. 
  - Does this node need to be destroyed when the handling of the request is completed, or is the Runtime smart enough to remove the node itself? 
- Adds a simple implementation of the challenge response (#1357), using a fixed challenge phrase (oak-challenge)
  - The signed challenge is added as an HTTP header in the examples and unit tests, but for now the code assumes that the signed challenge is optional. Not sure if this a valid assumption though.
  - New protobuf messages `SignedChallenge`, `UserIdentityTag`, `UserNodeConfiguration`, and `OuterHttpInvocation` are added.
- The `UserNode` with its privilege uses the user's identity to correctly set the labels on the invocation channels used for the interaction with the Oak node (#1428).
- Tests are updated accordingly.
- [Documentation](https://github.com/project-oak/oak/blob/main/docs/concepts.md#grpc-and-user-labels) will be updated in a separate PR. 

Here is an attempt to visualize this change: 

before

after

# Checklist

  - [x] I have written/updated tests that cover the code changes.
      construction.


oak_runtime/src/node/http/user.rs
```diff
@@ -0,0 +1,303 @@
...
+        "user-node"
+    }
+
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        startup_handle: oak_abi::Handle,
+        _notify_receiver: oneshot::Receiver<()>,
+    ) {
+        let _unit_result = self.try_run(runtime, startup_handle);
```
conradgrobler, line 177: This seems weird that `try_run` returns a result, but is only used in one place where the result is ignored.

oak_runtime/src/node/http/user.rs
```diff
@@ -0,0 +1,303 @@
...
+        })?;
+
+        // Send the request to the Oak node and get a handle to the channel containing the response.
+        let response_reader =
+            self.inject_http_request(runtime.clone(), outer_http_invocation.clone())?;
+        let response_receiver = crate::io::Receiver::new(ReadHandle {
+            handle: response_reader,
+        });
+        let response = response_receiver.receive(&runtime).map_err(|err| {
+            error!(
```
conradgrobler, line 146: Optional: I personally don't like this pattern of using `map_err` to do logging. I think it is cleaner if mapper functions like `map_err` do not produce side-effects. One option might be to define an error type that contains the message and it's severity. This can then be logged as eitehr a warning or error at the higher level where the result is processed.

oak_runtime/src/node/http/user.rs
```diff
@@ -0,0 +1,303 @@
...
+    privilege: NodePrivilege,
+}
+
+impl UserNode {
+    pub fn new(node_name: &str, config: UserNodeConfiguration) -> Result<Self, ConfigurationError> {
+        let mut privilege_tags = HashSet::new();
+        let tag = match config.privilege.is_empty() {
+            true => None,
+            false => Some(oak_abi::label::tag::Tag::UserIdentityTag(UserIdentityTag {
+                public_key: config.privilege,
```
conradgrobler, line 57: The naming seems confusing: using `privilege` in the proro to represent a public key in the config, but later it is used to represent the entire node privilege.

oak_runtime/src/node/http/user.rs
```diff
@@ -0,0 +1,303 @@
...
+        let request_reader_label = Label {
+            confidentiality_tags: request_label.confidentiality_tags.clone(),
+            integrity_tags: privilege
+                .can_endorse_integrity_tags
+                .clone()
+                .into_iter()
+                .collect(),
+        };
+        let response_writer_label = Label {
+            confidentiality_tags: privilege
```
conradgrobler, line 209: Should this check whether user identity tag was included in the original label? If the user did not include the identity in the original request, this could unnecessarily increase the confidentiality.

oak_runtime/src/node/http/user.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /// Sends an instance of [`HttpInvocation`] to the Oak node. The [`HttpInvocation`] instance
+    /// contains two request-specific channels, one that allows the Oak node to read the request,
+    /// and one on which the Oak node sends the response.
+    fn send_invocation(
+        &self,
+        runtime: &RuntimeProxy,
+        invocation_channel: oak_abi::Handle,
+    ) -> Result<(), ()> {
```
conradgrobler, line 262: Result<(),()> does not seem like a particularly useful type as it does not really convey much information and the results do not seem to be used anywhere.

conradgrobler (commented):
I was not involved in the original design discussions, so I am unsure on the background. Could you please add some information on why these label calculations are done in a separate node rather than in the HTTP server pseudo-node (similar to the current gRPC pseudo-node implementation)?

Is this just to make a logical separation more clear in the code, or is there a technical reason that would make this approach more secure (or some other technical advantage)?

rbehjati:
> I was not involved in the original design discussions, so I am unsure on the background. Could you please add some information on why these label calculations are done in a separate node rather than in the HTTP server pseudo-node (similar to the current gRPC pseudo-node implementation)?

According to the [docs](https://github.com/project-oak/oak/blob/main/docs/concepts.md#grpc-and-user-labels), the integrity component of the `request receiver` channel must be set to the identity of the user. However, the HTTP server pseudo-Node cannot create a channel with a non-empty integrity component. The user node has the privilege that allows it to create such channels. Does the gRPC implementation actually set the integrity of the `request receiver` channel to the user's identity? Do we have examples that show that it works? 

I had a discussion with @tiziano88 and realized that the `OuterHttpInvocation` that the HTTP server pseudo-node sends to the UserNode needs to be changed, as currently it publicly exposes the request. 

The other issue is the one that I am experimenting with in #1614. If the `request label` is non-empty, the Oak node won't be able to read the request. This means that neither the before approach nor the new approach work!

conradgrobler:
> According to the [docs](https://github.com/project-oak/oak/blob/main/docs/concepts.md#grpc-and-user-labels), the integrity component of the `request receiver` channel must be set to the identity of the user. However, the HTTP server pseudo-Node cannot create a channel with a non-empty integrity component. The user node has the privilege that allows it to create such channels. Does the gRPC implementation actually set the integrity of the `request receiver` channel to the user's identity? Do we have examples that show that it works?

Integrity label support is not currently implemented for the gRPC server node, but it should be possible to do so. Seeing that the gRPC and HTTP server pseudo-nodes are part of the runtime and therefore the TCB they should in effect have the "top" integrity label. Therefore if we had a way of representing "top" as a label, it would resolve the issue. That seems to me to be a cleaner approach than creating a new node that gets assigned per-user integrity by the runtime. In both cases the runtime must be trusted to create the appropriate label, but in the case of the extra node there is significant overhead without clear security or safety improvements.

> I had a discussion with @tiziano88 and realized that the `OuterHttpInvocation` that the HTTP server pseudo-node sends to the UserNode needs to be changed, as currently it publicly exposes the request.

I agree.

> The other issue is the one that I am experimenting with in #1614. If the `request label` is non-empty, the Oak node won't be able to read the request. This means that neither the before approach nor the new approach work!

Any user-specific label on the request requires a "router" pattern where the router node (public) would create a new node with the right label and forward the invocation to the new node which should handle the request and produce the response.

rbehjati:
> Integrity label support is not currently implemented for the gRPC server node, but it should be possible to do so. Seeing that the gRPC and HTTP server pseudo-nodes are part of the runtime and therefore the TCB they should in effect have the "top" integrity label. Therefore if we had a way of representing "top" as a label, it would resolve the issue. That seems to me to be a cleaner approach than creating a new node that gets assigned per-user integrity by the runtime. In both cases the runtime must be trusted to create the appropriate label, but in the case of the extra node there is significant overhead without clear security or safety improvements.

What do you mean by "top" integrity label? Currently, the default choice for labelling the new nodes and channels is `public_untrusted`. Are you suggesting that is should be changed to `public_fully_trusted` (for those nodes and channels that belong to the TCB)?

I agree that that would be a more elegant solution, but I don't know what are its implications for our security guarantees, and NI proofs.

 > Any user-specific label on the request requires a "router" pattern where the router node (public) would create a new node with the right label and forward the invocation to the new node which should handle the request and produce the response.

Currently (with `public_untrusted` being the default label, and the label of the router node), this would not work in general. We may need intermediate nodes with specific privileges to be able to create user/request-specific Oak nodes with the right labels.

conradgrobler:
> What do you mean by "top" integrity label? Currently, the default choice for labelling the new nodes and channels is `public_untrusted`. Are you suggesting that is should be changed to `public_fully_trusted` (for those nodes and channels that belong to the TCB)?

Yes.

> I agree that that would be a more elegant solution, but I don't know what are its implications for our security guarantees, and NI proofs.

I don't think it is a problem. The TCB is fully trusted, so to explicitly assign a "fully_trusted" integrity label to parts of it should be fine.

> > Any user-specific label on the request requires a "router" pattern where the router node (public) would create a new node with the right label and forward the invocation to the new node which should handle the request and produce the response.
> 
> Currently (with `public_untrusted` being the default label, and the label of the router node), this would not work in general. We may need intermediate nodes with specific privileges to be able to create user/request-specific Oak nodes with the right labels.

I don't think intermediate nodes will help. The untrusted intial node will not be able to create channels to communicate with the more trusted intermediate nodes (and probably not be able to create these nodes either). If there is some exception that allows it, it breaks IFC (or implicitly grants the router "fully_trusted" integrity). The router node also needs to be "public_fully_trusted" to function correctly. As an aside, this is part of the reason for the configuration-based router node RFC: a well-reviewed, reusable node that can act in a fully trusted way.

Integrity tags in general are currently still problematic. E.g. we don't have a way to create the initial Wasm node with any level of trusted integrity (apart from perhaps its own Wasm hash), so it severely limits the propagation of integrity labels through the system.

rbehjati:
> I don't think intermediate nodes will help. The untrusted intial node will not be able to create channels to communicate with the more trusted intermediate nodes (and probably not be able to create these nodes either).

The intermediate nodes won't be more trusted, but they'd have extra privileges. I don't think we have any IFC rules that restrict the privileges of a node. At least currently, while we still have separate privileges instead of robust declassification and transparent endorsement.

conradgrobler:
> The intermediate nodes won't be more trusted, but they'd have extra privileges. I don't think we have any IFC rules that restrict the privileges of a node. At least currently, while we still have separate privileges instead of robust declassification and transparent endorsement.

Good point. This is the case currently. But it seems that is still does not solve the problem. We can't currently assign arbitrary privilege to any nodes outside of the TCB, so these nodes must be generic nodes in the TCB. How will they decide whether or not they should create the more trusted nodes? If they always just create the nodes that they are told to create, it is equivalent to just granting the privilege to the original node that creates the intermediate node. Then just granting "fully_trusted" privilege to the original node is a simpler, but security-wise equivalent solution.

rbehjati:
> Then just granting "fully_trusted" privilege to the original node is a simpler, but security-wise equivalent solution.

Good point :)

rbehjati:
@conradgrobler @daviddrysdale @ipetr0v @tiziano88 Following the [discussions on Slack](https://project-oak.slack.com/archives/CHE9E13C3/p1603458166142800) I think it is best to abandon this PR for now. If I leave out the integrity labels, then it would be possible to implement the first increment of the challenge-response without the virtual `UserNode` (or a node with `fully_trusted` integrity label). So, the PR could split into two.

tiziano88:
I agree, I think this may be split even further:

- implement proper privilege for the bearer token auth (though this may be throwaway work if we decide to remove it)
- implement challenge / response mechanism, and assign privilege correctly (only for confidentiality labels)
- separately look into integrity labels (I think this is not a priority for now)
- separately look into splitting nodes per user (or we can discuss on an issue / slack thread if necessary)
- fix tests related to assigning labels based on HTTP headers (already started in https://github.com/project-oak/oak/pull/1614)

oak_abi/proto/label.proto
```diff
@@ -82,3 +83,11 @@ message TlsEndpointTag {
   // using the set of Certificate Authorities (CA) that are available to it.
   string authority = 1;
 }
+
+// Policies related to user identification.
+message UserIdentityTag {
```
tiziano88, line 88: As I am rewriting the chat app to use labels, it turns out that we use public keys to represent things other than users (e.g. chat rooms), so I think we should use a more generic terminology, so perhaps just use `PublicKeyIdentityTag` or similar?

rbehjati: Thanks. I'll change the name.

oak_runtime/src/node/http/user.rs
ipetr0v: Will `anyhow::Context` fit here?

oak_abi/proto/label.proto
rbehjati: I have changed the name in my new PR,  but I now wonder if `PublicKeyIdentityTag` still corresponds to the `user` sub-lattice.  I guess there should be a one-to-one correspondence between the tag types and the principals. Or do we consider `chat room` to have the same principal nature as users?

rbehjati (commented):
Thanks for the reviews. I have created a new PR (https://github.com/project-oak/oak/pull/1652) that leaves out setting the user's identity in the integrity label of the request channel. I have applied the comments in the new PR.

rbehjati:
Replaced with #1652

PR #1591, opened 2020-10-16
Async service generator

wildarch:
# Checklist

      construction.

Example of using the new Async service code can be found at: https://github.com/wildarch/oak/blob/async_service_with_chat/examples/chat/module/rust/src/lib.rs.

### DIFF
--- a/experimental/oak_async/src/grpc.rs
+++ b/experimental/oak_async/src/grpc.rs
@@ -217,6 +217,7 @@ impl<T: Decodable> Stream for GrpcRequestStream<T> {
     }
 }
 
+// A gRPC invocation decomposed into the method, request stream and response writer.
 // This should only be used by the Oak service generator. Not a stable API.
 #[doc(hidden)]
 pub struct DecomposedInvocation {
--- a/experimental/oak_async/src/grpc.rs
+++ b/experimental/oak_async/src/grpc.rs
@@ -159,10 +159,8 @@ pub struct GrpcRequestStream<T: Decodable> {
 // on to the handler.
 impl<T: Decodable> Unpin for GrpcRequestStream<T> {}
 
-// This should only be used by the Oak service generator. Not a stable API.
-#[doc(hidden)]
 impl GrpcRequestStream<()> {
-    pub fn new(inner: ChannelReadStream<GrpcRequest>) -> GrpcRequestStream<()> {
+    fn new(inner: ChannelReadStream<GrpcRequest>) -> GrpcRequestStream<()> {
         GrpcRequestStream {
             inner: inner.peekable(),
             is_drained: false,
@@ -171,6 +169,8 @@ impl GrpcRequestStream<()> {
     }
 
     // Casts this stream into the given message type.
+    // This should only be used by the Oak service generator. Not a stable API.
+    #[doc(hidden)]
     pub fn into_type<T: Decodable>(self) -> GrpcRequestStream<T> {
         GrpcRequestStream {
             inner: self.inner,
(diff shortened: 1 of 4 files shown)


experimental/oak_async/src/grpc.rs
```diff
@@ -129,3 +145,116 @@ impl<T> Drop for MultiWriter<T> {
...
+        })?
+        .as_ref()
+        // `peek` only gives us a reference to the error, and `OakError` cannot implement `Clone`,
+        // so we need to do a poor man's clone here.
+        .map_err(|e| clone_error(e))?;
+    let method_name = req1.method_name.clone();
+    Ok((method_name, requests, response_writer))
+}
+
+// Poor man's clone for `OakError`
```
wildarch, line 252: I might be able to get around this by consuming `requests` on the error path, hopefully that will allow getting rid of this ugly hack 😅

wildarch: That turned out to be easier than expected, sorted!

experimental/oak_async/src/lib.rs
```diff
@@ -19,4 +19,4 @@ pub mod grpc;
 mod io;
 
 pub use executor::block_on;
-pub use io::{ChannelRead, ChannelReadStream, ReceiverAsync};
+pub use io::{run_command_loop, ChannelRead, ChannelReadStream, ReceiverAsync};
```
rbehjati, line 22: Why is this needed?

oak_utils/src/lib.rs
```diff
@@ -163,6 +163,97 @@ impl prost_build::ServiceGenerator for OakServiceGenerator {
     }
 }
 
+struct AsyncServiceGenerator;
+
+impl prost_build::ServiceGenerator for AsyncServiceGenerator {
```
rbehjati, line 168: Is it possible to add a simple example that uses this? It would be nice to see the output of the generator.

wildarch: I have a separate branch that modifies the chat example to use the async service generator: https://github.com/wildarch/oak/blob/async_service_with_chat/examples/chat/module/rust/src/lib.rs.

I would prefer not to add it to this PR yet, to keep the changes small and self-contained.

experimental/oak_async/src/lib.rs
wildarch: The `io` module is not public, so we need to manually export the function here.

experimental/oak_async/src/grpc.rs
```diff
@@ -129,3 +145,116 @@ impl<T> Drop for MultiWriter<T> {
...
+}
+
+// Unpin is needed to be able to peek at the first request before passing it
+// on to the handler.
+impl<T: Decodable> Unpin for GrpcRequestStream<T> {}
+
+// This should only be used by the Oak service generator. Not a stable API.
+#[doc(hidden)]
+impl GrpcRequestStream<()> {
+    pub fn new(inner: ChannelReadStream<GrpcRequest>) -> GrpcRequestStream<()> {
```
rbehjati, line 165: Can this be non-public?

experimental/oak_async/src/grpc.rs
```diff
@@ -129,3 +145,116 @@ impl<T> Drop for MultiWriter<T> {
...
+                    };
+                    T::decode(&message)
+                })
+            })
+        })
+    }
+}
+
+// This should only be used by the Oak service generator. Not a stable API.
+#[doc(hidden)]
```
rbehjati, line 221: Adding a new struct, with specific field names, for the return type might be overkill, but I think it is still useful to have a bit of documentation here, explaining what this function produces.

experimental/oak_async/src/grpc.rs
```diff
@@ -129,3 +145,116 @@ impl<T> Drop for MultiWriter<T> {
...
+pub async fn invocation_to_requests_and_writer(
```
rbehjati, line 222: > 
```suggestion
pub async fn invocation_to_requests_and_response_writer(
```

experimental/oak_async/src/grpc.rs
```diff
@@ -129,3 +145,116 @@ impl<T> Drop for MultiWriter<T> {
...
+    })?;
+    let mut requests = GrpcRequestStream::new(receiver.receive_stream());
+    let requests_pinned = core::pin::Pin::new(&mut requests);
+    let sender = invocation.sender.ok_or_else(|| {
+        error!("Invocation did not have a sender");
+        OakError::OakStatus(OakStatus::ErrInvalidArgs)
+    })?;
+    let response_writer = ChannelResponseWriter::new(sender);
+
+    let req1 = match requests_pinned
```
rbehjati, line 237: `req1` is only used for getting the `method_name` from it. I think it is more readable to merge this and the next statement, and only declare `method_name`.

experimental/oak_async/src/io.rs
```diff
@@ -128,10 +128,31 @@ pub trait ReceiverAsync {
...
+    T: Decodable + Send,
+    F: FnOnce(ChannelReadStream<T>) -> R,
+    R: Future<Output = ()> + 'static,
+{
+    match crate::block_on(handler(receiver.receive_stream())) {
+        Ok(()) => {}
+        Err(OakStatus::ErrTerminated) => {
+            info!("Received termination status, terminating command loop");
+        }
+        Err(e) => panic!("Command loop received non-termination error: {:?}", e),
```
rbehjati, line 156: In the other `run_command_loop` in the SDK, we log an error and return in case of an error. Does it make sense to do the same thing here as well instead of using `panic!`?

oak_utils/src/lib.rs
rbehjati: Thanks for the link :)

rbehjati (approved):
LGTM. Thanks!

experimental/oak_async/src/grpc.rs
wildarch: Aah yes, this can actually by private now. Thanks for spotting!

experimental/oak_async/src/io.rs
wildarch: I think it makes sense to have the same behaviour for both, will change to `error!` instead. Thanks!

experimental/oak_async/src/grpc.rs
wildarch: I think it's a good idea, done!

PR #1592, opened 2020-10-17
Use anyhow for error handling in CommandHandler

### LINKED ISSUE #1580: Make `CommandLoop` return a generic error instead of only `OakError`
Current version:

https://github.com/project-oak/oak/blob/694d7aeccabe73eb77aaaadaf341fb45727dcd07/sdk/rust/oak/src/lib.rs#L430

This is only used as part of the command loop, in which case any error is simply printed out, and the loop continues at the next iteration:

https://github.com/project-oak/oak/blob/694d7aeccabe73eb77aaaadaf341fb45727dcd07/sdk/rust/oak/src/lib.rs#L516-L518

I think it would make things simpler and more idiomatic if the return type were just `anyhow::Result<()>`, to mimic what may be returned from a `main` function.

tiziano88:
Fixes #1580

# Checklist

      construction.

### DIFF
--- a/sdk/rust/oak/src/lib.rs
+++ b/sdk/rust/oak/src/lib.rs
@@ -425,9 +425,9 @@ pub fn set_panic_hook() {
 ///
 /// It has a single method for handling commands, which are [`Decodable`](crate::io::Decodable)
 /// objects that are received via the single incoming channel handle which is passed in at Node
-/// creation time. The return value is only used for logging in case of failure.
+/// creation time. The return value is only used for logging in case of error.
 pub trait CommandHandler<T: crate::io::Decodable> {
-    fn handle_command(&mut self, command: T) -> Result<(), crate::OakError>;
+    fn handle_command(&mut self, command: T) -> anyhow::Result<()>;
 }
 
 /// Run a command loop on the provided [`CommandHandler`]:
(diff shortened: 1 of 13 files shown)


sdk/rust/oak/src/lib.rs
```diff
@@ -425,9 +425,9 @@ pub fn set_panic_hook() {
 ///
 /// It has a single method for handling commands, which are [`Decodable`](crate::io::Decodable)
 /// objects that are received via the single incoming channel handle which is passed in at Node
-/// creation time. The return value is only used for logging in case of failure.
+/// creation time. The return value is only used for logging in case of error.
 pub trait CommandHandler<T: crate::io::Decodable> {
-    fn handle_command(&mut self, command: T) -> Result<(), crate::OakError>;
+    fn handle_command(&mut self, command: T) -> anyhow::Result<()>;
```
ipetr0v, line 430: Does `anyhow` propagate outside Wasm?

ipetr0v (approved):

sdk/rust/oak/src/lib.rs
tiziano88: Wasm does not know anything about anyhow or any other error, all the handling is done in our own code, which is what this PR implements.

PR #1597, opened 2020-10-19
Update various versions of tools in Dockerfile

tiziano88:


rbehjati (approved):

third_party/roughenough/src/sign.rs
```diff
@@ -41,10 +41,7 @@ impl Verifier {
 
     pub fn verify(&self, expected_sig: &[u8]) -> bool {
         let public_key = signature::UnparsedPublicKey::new(&signature::ED25519, &self.pubkey);
-        match public_key.verify(&self.buf, expected_sig) {
-            Ok(_) => true,
-            _ => false,
-        }
+        public_key.verify(&self.buf, expected_sig).ok()
```
tiziano88, line 44: @conradgrobler FYI

oak_derive/tests/visit.rs
```diff
@@ -131,9 +131,13 @@ mod enums {
 fn assert_visit<T: HandleVisit>(mut t: T, count: usize) {
     let mut counter = 0;
 
-    t.visit(|_| {
-        counter += 1;
-    });
+    // TODO(#1599): Re-enable when the underlying issue is fixed.
+    #[allow(unused_must_use)]
```
wildarch, line 135: Can you use `let _ = ` here instead? See also: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

tiziano88: Done.

PR #1603, opened 2020-10-20
Fix state projection defs; finish unobs_step lemma

aferr:
# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -79,6 +79,7 @@ Theorem proj_pres_nid_fresh: forall ell s,
 Proof.
 Admitted.
 
+
 Definition idempotent {A: Type} (f: A -> A) := forall a, f (f a) = f a.
 
 Theorem node_low_proj_idempotent: forall ell, idempotent (node_low_proj ell).
@@ -169,17 +170,12 @@ Admitted.
 
 Theorem proj_node_state_to_proj_n: forall ell s id n,
     ((nodes (state_low_proj ell s)).[? id] = Some n) ->
-    n.(nlbl) <<L ell ->
     exists n',
         ((nodes s).[? id] = Some n') /\
         (node_low_proj ell n') = n.
 Proof.
-    intros.
-    inversion H. unfold node_state_low_proj in H2. 
-    Set Printing All. unfold fnd in H2.
-    destruct (nodes s id) eqn:E. 
-    - exists n0. split. eauto. congruence.
-    - admit. (* This case is not true *)
+    (* XXX *)
+    (* This is no longer true, I think *)
 Admitted.
 
 Theorem node_projection_preserves_flowsto: forall ell s id n n',
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -79,7 +79,6 @@ Theorem proj_pres_nid_fresh: forall ell s,
 Proof.
 Admitted.
 
-
 Definition idempotent {A: Type} (f: A -> A) := forall a, f (f a) = f a.
 
 Theorem node_low_proj_idempotent: forall ell, idempotent (node_low_proj ell).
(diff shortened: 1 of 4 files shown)


aferr:
@tiziano88 you're here because of https://github.com/project-oak/oak/issues/1197

experimental/ni_coq/vfiles/RuntimeModel.v
```diff
@@ -193,7 +193,16 @@ Inductive step_node (id: node_id): call -> state -> state -> Prop :=
...
+        ch.(clbl) <<L n.(nlbl) -> 
+            (* label of channel flowsTo label of caller. 
+                checks that reading the message is safe
+             *)
+        n.(nlbl) <<L ch.(clbl) -> 
+            (* label of caller flowsTo label of ch. This check is less intuitive than the above.
+            However, reads are destructive because the channel is popped, so the calling node
+            is effectively doing a write to the channel. Any node that does a subsequent read
+            can detect whether or not the channel has been read because its contents may have 
+            been changed. This check rules out this more subtle leak *)
```
tiziano88, line 205: Just to confirm, I think this also implies that e.g. a secret node may not be able to directly read from a public channel, right?

aferr: That is correct! (The reason is ReadChannel effectively also does a write back into the channel by popping the channel).

aferr: In this system, a public node can still send a message to a secret node by having the secret node hold the read handle to a secret channel and having the public node send to that (secret) channel.

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -71,6 +79,7 @@ Theorem proj_pres_nid_fresh: forall ell s,
 Proof.
 Admitted.
 
+
```
jadephilipoom, line 82: nit: extra line break added

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -148,14 +148,80 @@ Definition conjecture_possibilistic_ni := forall ell t1_init t2_init t1n,
...
+    intros. inversion H0; (split; [ inversion H4 | ] ); subst_lets; crush.
+    - (* WriteChannel; states loweq *) 
+        assert ( ~ (clbl ch <<L ell) ) by eauto using ord_trans.
+        eapply state_low_eq_trans.
+        eapply (state_upd_node_unobs _ _ _ _ (node_del_rhans (rhs msg) n)
+            ltac:(eauto) ltac:(eauto) ltac:(eauto)).
+        eapply (state_upd_chan_unobs _
+            (state_upd_node id (node_del_rhans (rhs msg) n) s)
+            _ _ (chan_append ch msg)
+            ltac:(eauto) ltac:(eauto) ltac:(eauto)).
```
jadephilipoom, line 166: If you're up for it, this looks like another place where an `eauto with ...` hint database would help!

jadephilipoom (approved):

tiziano88 (approved):

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: That would be better. I did try doing `eauto using <...>` and listed the ..._unobs theorems in a few places, but that didn't directly work. I think with that really simple way of doing it, coq struggles to come up with both:
- the particular choice of states to use in the transitivity theorem
- the values of some of the arguments to the _unobs theorems (like the new node / new channel)

Though maybe it works better with a Hint that shows coq exactly how to apply the theorem by matching on the goal / context ?

If you have time and you're interested to do better with the automation, please do.

jadephilipoom: Hmm, yeah, it might require some more fancy techniques like `Hint Extern` or `Hint Transparent`. Fair enough!

aferr: Actually Jade, I think the change in the definitions from this PR causes proj_node_state_to_proj_n in NIUtils to no longer be true. I am struggling a bit to come up with a way to fix the proof of low_proj_steps_implies_leq_step so that it no longer needs this theorem (or to sufficiently special-case proj_node_state_to_proj_n). If you happen to have time to take a look at it, I'd appreciate it.

jadephilipoom: Okay, I'll take a look. I'm a little swamped until tomorrow evening, but should be able to take a look Friday or early next week if that's a reasonable timeframe for you.

PR #1604, opened 2020-10-20
Add Rust Oak client library

ipetr0v:
This change adds a Rust `oak_client` library.

Ref https://github.com/project-oak/oak/issues/1097


oak_abi/src/label/mod.rs
```diff
@@ -54,6 +54,25 @@ impl crate::proto::oak::label::Label {
...
+    /// Returns a WebAssembly module Ed25519 public key label.
+    /// https://ed25519.cr.yp.to
+    pub fn public_key(public_key: &[u8]) -> Self {
+        let public_key_tag = web_assembly_module_signature_tag(&public_key);
+        Label {
+            confidentiality_tags: vec![public_key_tag],
+            integrity_tags: vec![],
+        }
+    }
+
```
tiziano88, line 75: I don't think these methods should be here (or exist at all). AFAICT they are just wrapping a tag into a label by using it as confidentiality component only. How about a `confidentiality_label` method (outside of `Label`, as top-level, like the existing `web_assembly_module_signature_tag` & similar methods). Also the naming is quite vague, and inconsistent with the existing methods about tags, or is there a reason for that?

oak_client/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        .tls_config(tls_config)
+        .context("Couldn't create TLS configuration")?
+        .connect()
+        .await
+        .context("Couldn't connect to Oak Application")
+}
+
+/// Intercepts gRPC messages and adds `label` as a gRPC metadata.
+pub struct Interceptor {
+    label: Vec<u8>,
```
tiziano88, line 43: Why not keep this as type Label?

oak_client/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+}
```
tiziano88, line 44: Add a TODO somewhere to deal with authentication too?

oak_client/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        ClientTlsConfig::new().ca_certificate(Certificate::from_pem(root_tls_certificate.to_vec()));
+    Channel::builder(uri.clone())
+        .tls_config(tls_config)
+        .context("Couldn't create TLS configuration")?
+        .connect()
+        .await
+        .context("Couldn't connect to Oak Application")
+}
+
+/// Intercepts gRPC messages and adds `label` as a gRPC metadata.
```
tiziano88, line 41: Also mention that it adds the same label to each request.

tiziano88 (approved):

oak_client/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl Into<tonic::Interceptor> for Interceptor {
+    fn into(self) -> tonic::Interceptor {
+        let interceptor = move |mut request: Request<()>| {
+            request.metadata_mut().insert_bin(
+                oak_abi::OAK_LABEL_GRPC_METADATA_KEY,
+                MetadataValue::from_bytes(self.label.as_ref()),
+            );
+            Ok(request)
```
tiziano88, line 65:
```suggestion
            Ok(())
```

oak_client/src/lib.rs
ipetr0v: Because encoding the label returns an error. And `Into<Interceptor>` couldn't return an error.

tiziano88: That seems worth a comment in the file, right?

oak_client/src/lib.rs
ipetr0v: Done

oak_client/src/lib.rs
ipetr0v: Done

oak_client/src/lib.rs
ipetr0v: Done

oak_abi/src/label/mod.rs
ipetr0v: I thought since we already have `public_untrusted` method, we just could add convenience functions for simple labels.

tiziano88: `public_untrusted` is like `default` or `empty`, that's why it is defined on `Label`. But I think individual label constructors can just be regular functions outside of it.

ipetr0v: Done, added:
```rust
/// Convenience function for creating labels with a single confidentiality tag.
pub fn confidentiality_label(tag: Tag) -> crate::proto::oak::label::Label {
    Label {
        confidentiality_tags: vec![tag],
        integrity_tags: vec![],
    }
}
```

rbehjati (approved):
LGTM :)

PR #1609, opened 2020-10-21
Delete old version of Trusted Information Retrieval

ipetr0v:
This change deletes the old version of Trusted Information Retrieval example.

This version is being deleted because it was replaced by Trusted Database example which is simpler and doesn't require additional labels (on top of user tokens).


tiziano88:
Can you add a bit of context to the commit message as to why this is being deleted? :)

ipetr0v:
Done

tiziano88 (approved):

PR #1610, opened 2020-10-21
Replace tokio::fs::read with std::fs::read

ipetr0v:
This change replaces `tokio::fs::read` with `std::fs::read`, because `tokio::fs::read` is an unnecessary complication in these particular examples.

### DIFF
--- a/examples/aggregator/backend/src/main.rs
+++ b/examples/aggregator/backend/src/main.rs
@@ -68,12 +68,10 @@ async fn main() -> anyhow::Result<()> {
     env_logger::init();
     let opt = Opt::from_args();
 
-    let private_key = tokio::fs::read(&opt.grpc_tls_private_key)
-        .await
-        .context("Couldn't load private key")?;
-    let certificate = tokio::fs::read(&opt.grpc_tls_certificate)
-        .await
-        .context("Couldn't load certificate")?;
+    let private_key =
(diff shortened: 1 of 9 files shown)


tiziano88:
Rather than the "what" (which is pretty obvious looking at the changes), could the commit explain the "why" of the change? e.g. to make things simpler / faster / more elegant / whatever

juliettepretot:
I actually think there's some value in read operations being async, enabling joining of the futures to await them together

ipetr0v:
In these particular cases `read` just loads keys at the start of clients, so it's not necessary to be `async`

tiziano88:
To clarify: I know (or at least can guess) why you are making this change (since I suggested it in the past), but I am just suggesting to add this context to the commit message, since other people may wonder why this change was made.

tiziano88:
> I actually think there's some value in read operations being async, enabling joining of the futures to await them together. Even if we might not always do that, async seems like a reasonable default.

In this case, no one will ever perform this operation asynchronously, so this is just annoying for functions that are not already async, since now everything needs to be turned into async just for the sake of reading a file. So I think it makes sense to stick to sync.

tiziano88 (approved):

juliettepretot:
Makes sense, if we can otherwise turn these functions synchronous this seems reasonable. :) I assume we'll then follow this up with PRs to make these functions synchronous?

juliettepretot (approved):

PR #1614, opened 2020-10-22
Adds a new HTTP test with a non-empty request label

rbehjati:


oak_runtime/src/node/http/tests.rs
```diff
@@ -73,6 +73,28 @@ fn init_logger() {
...
+    init_logger();
+
+    // Start a runtime with an HTTP server node, and a thread simulating an Oak node to respond to
+    // HTTP requests.
+    let mut http_server_tester = HttpServerTester::new(2524);
+    let client_with_valid_tls = create_client(LOCAL_CA);
+
+    // Send an HTTPS request, and check that response has StatusCode::OK
+    let label = confidentiality_label(tls_endpoint_tag("localhost"));
+    let resp = send_request(client_with_valid_tls, "https://localhost:2524", label).await;
```
rbehjati, line 87: @tiziano88 The non-empty request label is set here, and the test passes. However, if a non-empty label is used in line 241 when creating the invocation channel for communication with the Oak node, then there will be a PermissionDenied error.

oak_runtime/src/node/http/tests.rs
```diff
@@ -186,8 +233,12 @@ fn create_communication_channel(runtime: &RuntimeProxy) -> (oak_abi::Handle, oak
     // exactly one handle in it.
     //
     // Create a channel for receiving invocations to pass to the HTTP server pseudo-Node.
+
+    // A non-empty label here results in PermissionDenied error.
+    // let label = confidentiality_label(tls_endpoint_tag("localhost"));
```
rbehjati, line 238: A non-empty label here results in a PermissionDenied error.

rbehjati (commented):
@tiziano88 This is the weird test that I mentioned.

oak_runtime/src/node/http/tests.rs
tiziano88: Ok I did some debugging (it took me a while!) but I think the solution is quite simple :) the simulated node never actually tries to read the request! It just receives the invocation, and basically throws away the request and just writes a response back!

https://github.com/project-oak/oak/blob/1dd7520f071f01a7cbae77e7534842e59ccfb229/oak_runtime/src/node/http/tests.rs#L218-L231

I think the tests would be much more realistic and useful if the simulated node actually replies back with the same content of the request (kind of like echo), or something derived from it. Could you try that and let me know if it explains the issue?

rbehjati: OMG 😃  Thanks! Actually this test does not use a real Oak node (the simulated node is not registered in the runtime). So it still allows reading the request. I changed the Rust client in the `http_server` example to use a non-empty request label. Now it gives an `ErrPermissionDenied` error. I guess this means that we have to rethink some things, in particular #1586.

How is it working for gRPC? Do we have gRPC tests or examples that use non-empty request labels?

tiziano88: > (the simulated node is not registered in the runtime). So it still allows reading the request

Not sure I understand why that is the case though

tiziano88: re: gRPC, it may be that we only have abitests for it (https://github.com/project-oak/oak/blob/1dd7520f071f01a7cbae77e7534842e59ccfb229/examples/abitest/module_0/rust/src/lib.rs), and I am not sure these exercise labels yet

rbehjati: It has to be changed :) It felt like a simpler way to quickly create an Oak node when I first wrote these tests. I guess the right way to do this is to use `build_wasm` and `node_create`.

rbehjati: I had a quick look at the abitests, and they don't seem to exercise the labels.

tiziano88: I think we should have a better way of writing the logic of a node for testing that does not involve compiling to wasm, do we not have something like that for other tests already?

rbehjati: I agree. Using `build_wasm` (by which I actually meant `oak_tests::compile_rust_wasm`) does not seem to be a good approach, especially for tests in the `oak_runtime`. 
Perhaps `TestNode`:
https://github.com/project-oak/oak/blob/caa91c86f4f2b9d451d7a8bae9b56d82375f4b3e/oak_runtime/src/tests.rs#L45-L66

That is used inside the `run_node_body`, which seems to be duplicating a lot of the `node_create` logic.

tiziano88: Yes I think that logic should be refactored and reused if possible.

rbehjati: I agree. If we make `Runtime` parametric on `NodeFactory`, then we could use test-specific node factories. I think that would be an easier way, and would not require duplicating any internal `Runtime` logic.

tiziano88: That's a nice idea, though it may be complicated to plumb the different NodeFactory impl all the way through. I wonder whether we should just have a method on Runtime to instantiate an arbitrary node for testing, if we don't have one already.

rbehjati: Yes. That is another alternative. I am playing with these ideas, to see which one works better :)

rbehjati: @tiziano88 What do you think about #1619? It allows replacing `ServerNodeFactory` with a test-specific node factory. If you think this make sense, I can try and change `TestNode` in `oak_runtime/src/tests.rs` to use a similar approach.

PR #1617, opened 2020-10-23
Change signature of `HandleVisit` impl function to a fold.

### LINKED ISSUE #1599: unused closure in handle visit usage
A [new version](https://github.com/project-oak/oak/pull/1597/files) of the Rust compiler now makes this an error:

```
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ error: unused closure that must be used
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║    --> /opt/my-project/oak_io/src/handle.rs:129:5
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     |
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 129 | /     msg.visit(|handle: &mut Handle| {
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 130 | |         handles.push(*handle);
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 131 | |         *handle = oak_abi::INVALID_HANDLE;
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 132 | |     });
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     | |_______^
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     |
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     = note: `-D unused-must-use` implied by `-D warnings`
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     = note: closures are lazy and do nothing unless called
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ error: unused closure that must be used
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║    --> /opt/my-project/oak_io/src/handle.rs:171:5
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     |
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 171 | /     msg.visit(|handle| {
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 172 | |         if let Some(to_inject) = handles.next() {
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 173 | |             *handle = *to_inject;
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 174 | |         } else {
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 175 | |             result = Err(OakError::ProtobufDecodeError(None));
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 176 | |         }
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ 177 | |     });
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     | |_______^
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     |
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║     = note: closures are lazy and do nothing unless called
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ error: aborting due to 2 previous errors
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║
tests ❯ cargo clippy ❯ ./oak_runtime/Cargo.toml ║ error: could not compile `oak_io`
```

I guess [the type of `visit` is a bit awkward](https://github.com/project-oak/oak/pull/1108#discussion_r440454021) in order to support recursion, perhaps it should be revisited.

wildarch:
# Checklist

      construction.

Fixes #1599.

### DIFF
--- a/oak_derive/src/lib.rs
+++ b/oak_derive/src/lib.rs
@@ -61,7 +61,7 @@ fn struct_impls(name: &Ident, data: &syn::DataStruct) -> TokenStream {
                 #body
             }
 
-            fn fold<B, F: FnMut(B, &mut ::oak_io::Handle) -> B>(&mut self, init: B, mut f: F) -> B {
+            fn fold<B>(&mut self, init: B, f: fn(B, &mut ::oak_io::Handle) -> B) -> B {
                 #fold_body
             }
         }
@@ -81,7 +81,7 @@ fn enum_impls(name: &Ident, data: &syn::DataEnum) -> TokenStream {
                 }
             }
 
-            fn fold<B, F: FnMut(B, &mut ::oak_io::Handle) -> B>(&mut self, init: B, mut f: F) -> B {
+            fn fold<B>(&mut self, init: B, f: fn(B, &mut ::oak_io::Handle) -> B) -> B {
                 match self {
                     #(
                         #name::#fold_variants,
@@ -172,7 +172,7 @@ fn accessors_fold<T: ToTokens>(accessors: &[T]) -> TokenStream {
     quote! {
         let accumulator = init;
         #(
-            let accumulator = #accessors.fold(accumulator, &mut f);
+            let accumulator = #accessors.fold(accumulator, f);
         )*
         accumulator
     }
--- a/oak_derive/src/lib.rs
+++ b/oak_derive/src/lib.rs
@@ -52,17 +52,12 @@ fn struct_impls(name: &Ident, data: &syn::DataStruct) -> TokenStream {
             .collect(),
         Fields::Unit => Vec::new(),
     };
-    let body = accessors_visit(&accessors);
-    let fold_body = accessors_fold(&accessors);
+    let body = accessors_fold(&accessors);
 
     quote! {
         impl ::oak_io::handle::HandleVisit for #name {
-            fn visit<F: FnMut(&mut ::oak_io::Handle)>(&mut self, visitor: F) -> F {
-                #body
-            }
-
             fn fold<B>(&mut self, init: B, f: fn(B, &mut ::oak_io::Handle) -> B) -> B {
-                #fold_body
+                #body
             }
         }
     }
@@ -70,21 +65,12 @@ fn struct_impls(name: &Ident, data: &syn::DataStruct) -> TokenStream {
 
 fn enum_impls(name: &Ident, data: &syn::DataEnum) -> TokenStream {
     let variants: Vec<TokenStream> = data.variants.iter().map(variant_impl).collect();
-    let fold_variants: Vec<TokenStream> = data.variants.iter().map(fold_variant_impl).collect();
     quote! {
         impl ::oak_io::handle::HandleVisit for #name {
-            fn visit<F: FnMut(&mut ::oak_io::Handle)>(&mut self, visitor: F) -> F {
-                match self {
-                    #(
-                        #name::#variants,
-                    )*
-                }
-            }
-
             fn fold<B>(&mut self, init: B, f: fn(B, &mut ::oak_io::Handle) -> B) -> B {
                 match self {
                     #(
-                        #name::#fold_variants,
+                        #name::#variants,
                     )*
                 }
             }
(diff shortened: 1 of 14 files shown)


oak_derive/src/lib.rs
```diff
@@ -52,11 +52,11 @@ fn struct_impls(name: &Ident, data: &syn::DataStruct) -> TokenStream {
             .collect(),
         Fields::Unit => Vec::new(),
     };
-    let body = accessors_visit(&accessors);
+    let body = accessors_fold(&accessors);
 
     quote! {
         impl ::oak_io::handle::HandleVisit for #name {
-            fn visit<F: FnMut(&mut ::oak_io::Handle)>(&mut self, visitor: F) -> F {
+            fn fold<B>(&mut self, init: B, f: fn(B, &mut ::oak_io::Handle) -> B) -> B {
```
tiziano88, line 59: This seems a bit too restrictive, it can only use fn pointers, but not Fn traits. Why not go with Fn trait instead? (BTW, no need to go all the way to FnMut, unlike iter::fold, I think we can assume that the receiver is just an immutable ref for now)

oak_derive/tests/visit.rs
```diff
@@ -127,14 +126,9 @@ mod enums {
     }
 }
 
-// Asserts that `t` is visited exactly `count` times when calling `HandleVisit::visit`.
-fn assert_visit<T: HandleVisit>(mut t: T, count: usize) {
-    let mut counter = 0;
+// Asserts that `t` is visited exactly `expected` times when calling `HandleVisit::fold`.
```
tiziano88, line 129:
```suggestion
// Asserts that `t` is visited exactly `expected` times when calling [`HandleVisit::fold`].
```

oak_io/src/handle.rs
```diff
@@ -82,17 +82,16 @@ impl From<Handle> for WriteHandle {
...
 /// impl HandleVisit for Thing {
-///     fn visit<F: FnMut(&mut Handle)>(&mut self, mut visitor: F) -> F {
-///         visitor(&mut self.handle);
-///         visitor
+///     fn fold<B>(&mut self, init: B, mut f: fn(B, &mut Handle) -> B) -> B {
+///         f(init, &mut self.handle)
 ///     }
 /// }
 /// ```
 pub trait HandleVisit {
```
tiziano88, line 90: Random thought, but would it make sense to just implement `Iterator` instead? Then we would get `fold` and other things for free. Conceptually it kind of makes sense for a type to return an iterator of handles, then the caller can do whatever it wants with it.

I guess this may be either `trait HandleVisit : Iterator<Item=Handle> {}`, or something with an extra layer of indirection, e.g. `trait HandleVisit { fn handles(&self) -> impl Iterator<Item=Handle> }` if we want to avoid clashes with `Iterator` that may already be on the type.

oak_io/src/handle.rs
```diff
@@ -167,20 +162,20 @@ pub fn extract_handles<T: HandleVisit>(msg: &mut T) -> Vec<Handle> {
...
+    msg.fold(Ok(handles.iter()), |handles, handle| {
+        let mut handles = handles?;
+        let to_inject = handles.next().ok_or(OakError::ProtobufDecodeError(None))?;
+        *handle = *to_inject;
+        Ok(handles)
+    })
+    // Check that there are no remaining handles
+    .and_then(|mut remaining_handles| {
+        if remaining_handles.next().is_some() {
+            Err(OakError::ProtobufDecodeError(None))
```
tiziano88, line 174: Could you log something meaningful here? As it is, if this happens I expect we will spend hours trying to debug it :)

tiziano88 (approved):

oak_derive/src/lib.rs
wildarch: I tried doing this with `Fn` instead, but that unfortunately does not work with recursion. Since the closure can't be copied or cloned I have to pass down a reference to the closure, which ends up generating `&&&&&...closure` types. Since fn pointers are `Copy`, that does work. I had a similar problem in the original implementation, which is why the `visit` method would return the visitor again instead, hence the weird type signature.

oak_io/src/handle.rs
wildarch: It's an interesting idea! I wonder how that will work with recursion though, we might need some boxing there to make it work.

Note we would need `Iterator<Item=&mut Handle>` here, since we want to be able to modify the handles.

PR #1619, opened 2020-10-23
Allow creating Oak nodes for testing

rbehjati:
# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/lib.rs
+++ b/oak_runtime/src/lib.rs
@@ -27,6 +27,7 @@ use crate::{
     channel::{with_reader_channel, with_writer_channel, Channel},
     message::Message,
     metrics::Metrics,
+    node::Node,
     proto::oak::introspection_events::{
         event::EventDetails, ChannelCreated, Direction, Event, HandleCreated, HandleDestroyed,
         MessageDequeued, MessageEnqueued, NodeCreated, NodeDestroyed,
@@ -1158,10 +1159,30 @@ impl Runtime {
     /// This method is defined on [`Arc`] and not [`Runtime`] itself, so that
     /// the [`Arc`] can clone itself and be included in a [`RuntimeProxy`] object
     /// to be given to a new Node instance.
-    fn node_create(
+    fn node_create_and_register(
         self: Arc<Self>,
         node_id: NodeId,
         config: &NodeConfiguration,
+        label: &Label,
+        initial_handle: oak_abi::Handle,
+    ) -> Result<(), OakStatus> {
+        // This only creates a Node instance, but does not start it.
+        let instance = self.node_factory.create_node(config).map_err(|err| {
+            warn!("could not create node: {:?}", err);
+            OakStatus::ErrInvalidArgs
+        })?;
+
+        // Register the instance within the `Runtime`.
+        self.node_register(node_id, instance, &config.name, label, initial_handle)
+    }
+
+    /// Registers the given [`Node`] instance within the [`Runtime`]. The registration fails if the
+    /// labels violate the IFC rules.
+    fn node_register(
+        self: Arc<Self>,
+        node_id: NodeId,
+        instance: Box<dyn Node>,
+        node_name: &str,
         label: &Label,
         initial_handle: oak_abi::Handle,
     ) -> Result<(), OakStatus> {
@@ -1181,13 +1202,7 @@ impl Runtime {
 
         let new_node_proxy = self.clone().proxy_for_new_node();
         let new_node_id = new_node_proxy.node_id;
-        let new_node_name = format!("{}({})", config.name, new_node_id.0);
-
-        // This only creates a Node instance, but does not start it.
-        let instance = self.node_factory.create_node(config).map_err(|err| {
-            warn!("could not create node: {:?}", err);
-            OakStatus::ErrInvalidArgs
-        })?;
+        let new_node_name = format!("{}({})", node_name, new_node_id.0);
 
         let node_privilege = instance.get_privilege();
         let node_type = instance.node_type();
--- a/oak_runtime/src/node/http/tests.rs
+++ b/oak_runtime/src/node/http/tests.rs
@@ -23,14 +23,50 @@ use oak_abi::{
         NodeConfiguration,
     },
 };
-use std::{fs, option::Option, thread::JoinHandle};
+use std::fs;
 
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
 static GCP_CA: &str = "../examples/certs/gcp/ca.pem";
 
+/// A simple Oak node that responds with 200 (OK) to every request it receives
+struct OakNodeForTest;
+
+impl crate::node::Node for OakNodeForTest {
+    fn node_type(&self) -> &'static str {
+        "test"
+    }
+    fn run(
+        self: Box<Self>,
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        _notify_receiver: oneshot::Receiver<()>,
+    ) {
+        // Get invocation message that contains the response_writer handle.
+        let invocation_receiver = Receiver::<HttpInvocation>::new(ReadHandle { handle });
+
+        while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            let request = invocation
+                .receiver
+                .unwrap()
+                .receive(&runtime)
+                .expect("Could not receive the request");
+
+            let resp = HttpResponse {
+                body: request.body,
+                status: http::status::StatusCode::OK.as_u16() as i32,
+                headers: request.headers,
+            };
+            invocation
+                .sender
+                .expect("Empty sender on invocation.")
+                .send(resp, &runtime)
+                .unwrap();
+        }
+    }
+}
+
 struct HttpServerTester {
     runtime: RuntimeProxy,
-    oak_node_simulator_thread_handle: Option<JoinHandle<()>>,
 }
 
 impl HttpServerTester {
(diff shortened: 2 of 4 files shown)


oak_runtime/src/node/http/tests.rs
```diff
@@ -20,17 +20,71 @@ use oak_abi::{
...
+    server_node_factory: crate::node::ServerNodeFactory,
+}
+
+impl crate::node::NodeFactory<NodeConfiguration> for TestNodeFactory {
+    fn create_node(
+        &self,
+        node_configuration: &NodeConfiguration,
+    ) -> Result<Box<dyn Node>, ConfigurationError> {
+        match &node_configuration.config_type {
+            Some(ConfigType::WasmConfig(_config)) => Ok(Box::new(OakNodeForTest {})),
```
tiziano88, line 76: Am I understanding correctly that this means that whenever a wasm node is requested, this will actually create a test node, regardless of the actual configuration? And for any other node type, it just passes it through to the ServerNodeFactory? If so, it seems quite hacky, and I think we will have to change it again if we wanted to use it to test anything that does rely on Wasm nodes at all.

BTW my intention with the NodeFactory trait was actually to push it up one level and make it an explicit generic type of `Runtime` itself, so that in test we may instantiate a `Runtime<TestNodeFactory>` that may have a separate implementation just for test. Would that make things simpler for you?

oak_runtime/src/node/http/tests.rs
```diff
@@ -20,17 +20,71 @@ use oak_abi::{
...
+            invocation
+                .sender
+                .expect("Empty sender on invocation.")
+                .send(resp, &runtime)
+                .unwrap();
+        }
+    }
+}
+
+struct TestNodeFactory {
```
tiziano88, line 66: This needs way more documentation if we decide to keep it, but if we do keep it it may make sense to model this as a wrapper around the inner factory:

```rust
struct TestNodeFactory<T: NodeFactory> {
  inner: T,
}

enum TestNodeConfiguration {
  Test(Box<dyn Node>),
  Inner(NodeConfiguration),
}

impl NodeFactory<TestNodeConfiguration> for TestNodeFactory<T> {
  ...
}
```

or something like that.

I think this actually would allow to write more realistic tests overall, since the logic for creating the initial node graph would also be run from an actual (test) node, rather than by the runtime itself.

WDYT?

oak_runtime/src/node/http/tests.rs
```diff
@@ -20,17 +20,71 @@ use oak_abi::{
...
 };
-use std::{fs, option::Option, thread::JoinHandle};
+use std::fs;
 
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
 static GCP_CA: &str = "../examples/certs/gcp/ca.pem";
 
+struct OakNodeForTest {}
+
+impl crate::node::Node for OakNodeForTest {
```
tiziano88, line 33: I assume we will want to generalize this so that it can be reused in other places, we only need to provide the body of the node for each particular test, but hopefully not an entire new factory and node type every time.

oak_runtime/src/node/http/tests.rs
rbehjati: I totally agree :)

oak_runtime/src/node/http/tests.rs
rbehjati: I totally agree with making `NodeFactory` an explicit generic type for `Runtime`. That would require making `RuntimeProxy` and all the pseudo-nodes and a few other things generic too. But that would allow defining a `TestNodeFactory` with a test-specific `NodeConfiguration` like the one you suggested above. So, let me try this one now... it is going to be a much bigger change.

tiziano88: I agree it is a big task, though I think there a few more complications down the line, e.g. that wasm nodes will have to be compiled targeting a specific NodeConfiguration, etc., which may not be a priority for us to solve now.

For now, what about adding a method to the current NodeFactory to allow creating arbitrary nodes (by providing a function) for test? I think that is fine, because other nodes do not get direct access to the NodeFactory instance anyways.

rbehjati: I think that would work, but in addition to that, we need something like `create_test_node` in the `Runtime` to call that new method in `NodeFactory`. We should be able to compile these additional methods only for test, right?

tiziano88: Either that, or we need to create the NodeFactory first, and in tests return a reference to that in addition to the Runtime reference.

Anyways, a `Runtime::create_test_node` method seems fine; if we can compile it only for test then it's even better, but even if not, it's not a big deal, nothing apart from the runtime itself (or its creator) will be able to invoke it, right?

tiziano88: Not sure whether `#[cfg(test)]` can be applied to methods.

rbehjati: > nothing apart from the runtime itself (or its creator) will be able to invoke it, right? 
True. And it will not be exposed in the SDK. 

> Not sure whether #[cfg(test)] can be applied to methods.
I thought it is possible. I'll try.

tiziano88: Another thought: perhaps it would make sense to actually expose a `Runtime::create_node` method that is invoked with the result of the NodeFactory object. In fact, I think the current code does something similar already, it just needs to be refactored slightly. No additional method for testing should be necessary.

oak_runtime/src/node/http/tests.rs
rbehjati: With the new approach, I am not sure if generalising it has much value. I think each test module can define a set of test Oak nodes, and provide the implementation for the `run` method.

rbehjati (commented):
@tiziano88 I have now refactored `Runtime::node_create` and added a `Runtime::register_node_instance`. A public method with the same name is added to `RuntimeProxy` and is only compiled for `test` targets. 

WDYT?

oak_runtime/src/proxy.rs
```diff
@@ -180,6 +183,33 @@ impl RuntimeProxy {
         result
     }
 
+    /// See [`Runtime::register_node_instance`].
```
tiziano88, line 186: Perhaps mention that this exposed to facilitate testing etc.

oak_runtime/src/lib.rs
```diff
@@ -1164,6 +1165,26 @@ impl Runtime {
...
+        self.register_node_instance(node_id, instance, config, label, initial_handle)
+    }
+
+    /// Registers the given [`Node`] instance within the [`Runtime`]. The registration fails if the
+    /// labels violate the IFC rules.
+    fn register_node_instance(
+        self: Arc<Self>,
+        node_id: NodeId,
+        instance: Box<dyn Node>,
+        config: &NodeConfiguration,
```
tiziano88, line 1185: Why is the config needed here? Conceptually, the node has already been created by the factory, or passed by test code, so I don't expect this to be here.

oak_runtime/src/lib.rs
```diff
@@ -1164,6 +1165,26 @@ impl Runtime {
...
+            OakStatus::ErrInvalidArgs
+        })?;
+
+        // Register the instance within the `Runtime`.
+        self.register_node_instance(node_id, instance, config, label, initial_handle)
+    }
+
+    /// Registers the given [`Node`] instance within the [`Runtime`]. The registration fails if the
+    /// labels violate the IFC rules.
+    fn register_node_instance(
```
tiziano88, line 1181: Maybe the other method should be called `node_create_and_register` and this `node_register`, or some other combination of names that are consistent with each other.

oak_runtime/src/lib.rs
```diff
@@ -1164,6 +1165,26 @@ impl Runtime {
...
+            warn!("could not create node: {:?}", err);
+            OakStatus::ErrInvalidArgs
+        })?;
+
+        // Register the instance within the `Runtime`.
+        self.register_node_instance(node_id, instance, config, label, initial_handle)
+    }
+
+    /// Registers the given [`Node`] instance within the [`Runtime`]. The registration fails if the
+    /// labels violate the IFC rules.
```
tiziano88, line 1180: Now that we have this method, do we still need `node_configure_instance`? I think that was just there to help with testing too, but I think this new method is overall a much better abstraction, so we should be able to change the previous test to just use this, and delete / inline the other one?

oak_runtime/src/node/http/tests.rs
```diff
@@ -23,14 +23,49 @@ use oak_abi::{
...
     },
 };
-use std::{fs, option::Option, thread::JoinHandle};
+use std::fs;
 
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
 static GCP_CA: &str = "../examples/certs/gcp/ca.pem";
 
+/// A simple Oak node that responds with 200 (OK) to every request it receives
+struct OakNodeForTest {}
```
tiziano88, line 32:
```suggestion
struct OakNodeForTest;
```

tiziano88 (commented):
Nice, I like this pattern!

rbehjati (commented):
Thanks for the review :)

oak_runtime/src/lib.rs
tiziano88: Got it, sounds good.

BTW I was hoping that the "implicit" initial node can also be removed now? Or is it still needed for anything? I guess it is used in order to send the config map to the initial node, but I think now it should be possible to use the new method you just added to create a proper node to do that (without having to go through NodeFactory). WDYT? Could you create an issue if you think it makes sense?

https://github.com/project-oak/oak/blob/27f580699a84e78d7b36c4883ef77b590950c636/oak_runtime/src/proxy.rs#L78-L84

oak_runtime/src/node/http/tests.rs
```diff
@@ -40,29 +75,22 @@ impl HttpServerTester {
...
-        }));
-
-        HttpServerTester {
-            runtime,
-            oak_node_simulator_thread_handle: oak_node_simulator_thread,
-        }
+        // Create an Oak node that responds with 200 (OK) to every request it receives.
+        runtime
+            .register_node_instance(
+                Box::new(OakNodeForTest {}),
```
tiziano88, line 81: Does this work?

```suggestion
                Box::new(OakNodeForTest),
```

oak_runtime/src/node/http/tests.rs
```diff
@@ -23,14 +23,49 @@ use oak_abi::{
...
+        runtime: RuntimeProxy,
+        handle: oak_abi::Handle,
+        _notify_receiver: oneshot::Receiver<()>,
+    ) {
+        // Get invocation message that contains the response_writer handle.
+        let invocation_receiver = Receiver::<HttpInvocation>::new(ReadHandle { handle });
+
+        while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            let request = invocation.receiver.unwrap().receive(&runtime);
+            info!("received request: {:?}", request);
```
tiziano88, line 49: It may be interesting to do something with the request, at least to have some level of confidence it is received correctly, e.g. send back any headers and body, similar to an echo server. Otherwise we may again have issues in which we don't fully "look inside" what is coming in, and we risk missing catching bugs related to HTTP serialization.

tiziano88 (approved):
Thanks!

oak_runtime/src/node/http/tests.rs
rbehjati: Done. It is now including the request body and the headers in the response.

Issue #1620, opened 2020-10-26
Refactor keys in `oak_sign`


ipetr0v:
Right now `KeyBundle` struct has both `private_key` and a `key_pair` fields, where in fact `private_key` already contains a Public Key because it's encoded with PKCS#8 v2 format:
https://github.com/project-oak/oak/blob/27f580699a84e78d7b36c4883ef77b590950c636/oak_sign/src/lib.rs#L34-L39

We need to refactor it and delete `KeyBundle`.

Originated from: https://project-oak.slack.com/archives/CHE9E13C3/p1603722973180400

tiziano88:
I suggest just calling it `KeyPair` and have `pkcs_8` and `parsed` fields (or something like that). AFAICT the only reason for this is that there is no method to go from a parsed key to its PKCS#8 representation, so ideally that should be clearly explained in the comments.

tiziano88:
Since you are at it, could you please also impl Clone for it? It will be useful in tests. Probably it needs to re-parse it from the pkcs8 bytes.

PR #1625, opened 2020-10-27
Refactor oak sign key pair

ipetr0v:
This change:
- Refactors `oak_sign::KeyBundle` to `oak_sign::KeyPair`
- Adds more comments
- Implements `Clone` for `oak_sign::KeyPair` (necessary for tests)

Fixes https://github.com/project-oak/oak/issues/1620

### DIFF
--- a/oak_sign/src/lib.rs
+++ b/oak_sign/src/lib.rs
@@ -18,7 +18,7 @@ use anyhow::Context;
 use pem::Pem;
 use ring::{
     rand,
-    signature::{self, Ed25519KeyPair, KeyPair},
+    signature::{self, Ed25519KeyPair},
 };
 use std::{
     collections::HashMap,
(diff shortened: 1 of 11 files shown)


oak_sign/src/lib.rs
```diff
@@ -31,44 +31,59 @@ pub const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
...
-    key_pair: Ed25519KeyPair,
+/// Convenience struct that encapsulates `ring::signature::Ed25519KeyPair`.
+pub struct KeyPair {
+    /// PKCS#8 v2 encoded private key and public key.
+    /// https://tools.ietf.org/html/rfc5958
+    ///
+    /// The encoded version is kept because a `parsed` key pair doesn't contain a method for
+    /// extracting private key bytes.
+    /// https://briansmith.org/rustdoc/ring/signature/struct.Ed25519KeyPair.html
+    encoded_key_pair: Vec<u8>,
```
tiziano88, line 42: Why not call this field pkcs_8 something? There is nothing else that can be done with it without knowing its format anyways, so this seems better than just a comment.

oak_sign/src/lib.rs
```diff
@@ -31,44 +31,59 @@ pub const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
...
+    /// Parsed PKCS#8 v2 key pair representation.
+    parsed_key_pair: Ed25519KeyPair,
 }
 
-impl KeyBundle {
+impl Clone for KeyPair {
+    fn clone(&self) -> Self {
+        KeyPair::parse(&self.encoded_key_pair).expect("Couldn't clone key pair")
```
tiziano88, line 49: Could you add a comment to explain why clone is implemented this way?

oak_sign/src/lib.rs
```diff
@@ -31,44 +31,59 @@ pub const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
...
             .ok()
             .context("Couldn't generate key pair")?;
-        KeyBundle::parse(private_key.as_ref())
+        KeyPair::parse(private_key.as_ref())
     }
 
     /// Parses a Ed25519 key pair from bytes.
-    pub fn parse(private_key: &[u8]) -> anyhow::Result<KeyBundle> {
-        let key_pair = Ed25519KeyPair::from_pkcs8(private_key)
+    pub fn parse(private_key: &[u8]) -> anyhow::Result<KeyPair> {
```
tiziano88, line 64: Could you clarify that the bytes are pkcs8, ideally both in the param name, and in the comment? Otherwise it seems impossible for anyone to use this method (or maybe it should not be public?)

oak_sign/src/lib.rs
```diff
@@ -31,44 +31,59 @@ pub const PUBLIC_KEY_TAG: &str = "PUBLIC KEY";
...
             .context("Couldn't parse generated key pair")?;
         Ok(Self {
-            private_key: private_key.to_vec(),
-            key_pair,
+            encoded_key_pair: private_key.to_vec(),
+            parsed_key_pair,
         })
     }
 
     pub fn private_key(&self) -> Vec<u8> {
```
tiziano88, line 74: Is this method what `parse` reads back? Could you add a comment and ideally also some tests that encode and decode a key?

And also please add some other tests of various functionality, since this crate is more and more used for other purposes too.

oak_sign/src/lib.rs
```diff
@@ -81,7 +96,7 @@ pub struct SignatureBundle {
 
 impl SignatureBundle {
     /// Signs a SHA-256 hash of the `input` using `private_key`.
-    pub fn create(input: &[u8], key_pair: &KeyBundle) -> anyhow::Result<SignatureBundle> {
+    pub fn create(input: &[u8], key_pair: &KeyPair) -> anyhow::Result<SignatureBundle> {
```
tiziano88, line 99: Personally I would just expose `sign` and `verify` as top-level methods in this module, *not* on `SignatureBundle` itself. There is really not much about `SignatureBundle` doing the work here, I think it would read better as `oak_sign::sign` and `oak_sign::verify`.

oak_sign/src/lib.rs
ipetr0v: Done
Also added tests

oak_sign/src/tests.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[test]
+fn key_pair_test() {
+    let generate_result = KeyPair::generate();
+    assert_matches!(generate_result, Ok(_));
+    let key_pair = generate_result.unwrap();
+
+    let pkcs8_key_pair = key_pair.pkcs8_key_pair();
+    let parse_result = KeyPair::parse(&pkcs8_key_pair);
+    assert_matches!(parse_result, Ok(_));
```
tiziano88, line 31: This does not check that the resulting key is actually the same as the original one, right?

oak_sign/src/tests.rs
```diff
@@ -0,0 +1,49 @@
...
+//
+
+use assert_matches::assert_matches;
+use oak_sign::{get_sha256_hex, KeyPair, SignatureBundle};
+
+const TEST_DATA: &str = "Test data";
+const TEST_SHA256: &str = "e27c8214be8b7cf5bccc7c08247e3cb0c1514a48ee1f63197fe4ef3ef51d7e6f";
+
+#[test]
+fn key_pair_test() {
```
tiziano88, line 24: Give this a more precise name, AFAICT this is testing serialization and deserialization, is that correct?

oak_sign/src/tests.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[test]
+fn signature_test() {
+    let key_pair = KeyPair::generate().expect("Couldn't create key pair");
+    let create_result = SignatureBundle::create(TEST_DATA.as_bytes(), &key_pair);
+    assert_matches!(create_result, Ok(_));
+    let signature = create_result.unwrap();
+
+    let verify_result = signature.verify();
+    assert_matches!(verify_result, Ok(_));
```
tiziano88, line 42: also verify an invalid signature, or create two signature with different keys and check only the corresponding ones check correctly

tiziano88 (approved):

oak_sign/src/tests.rs
ipetr0v: `Eq` is not implemented for `Ed25519KeyPair`

oak_sign/src/tests.rs
ipetr0v: Added an incorrect signature

oak_sign/src/tests.rs
tiziano88: But you could re-serialize to pkcs8 and compare that, I guess? or impl Eq for KeyPair, similar to how you did Clone.

oak_sign/src/tests.rs
tiziano88: Cool thanks! Though personally I still think something like this would read nicely:

```rust
let key_0 = ...
let key_1 = ...

let sign_0 = ...
let sign_1 = ...

assert_eq!(true, verify(sign_0, key_0))
assert_eq!(false, verify(sign_0, key_1))
assert_eq!(false, verify(sign_1, key_0))
assert_eq!(true, verify(sign_1, key_1))
```

tiziano88 (approved):
LG, thanks!

oak_sign/src/tests.rs
ipetr0v: Done

Issue #1615, opened 2020-10-22
assign debug name to channels


tiziano88:
Trying to debug oak application is often quite complicated due to the interleaving of channel operations. This is made harder by the fact that channel handle are randomised, therefore even running the same piece of code twice in a row results in a different trace of execution.

It may be useful to assign a friendly (not necessarily unique) name to channels at creation time, that is shown in traces instead of (or in addition to) handle numbers, to facilitate debugging. It may also be reflected in the introspection server graph.

I think nodes already have a similar mechanism (I should double check), but it is worth making sure that there is a coherent API for assigning names to both channels and nodes, and that debug names for both are shown consistently in log statements.

Both nodes and channels have unique ids already (which are incrementally assigned and therefore short and predictable), so perhaps we can combine the debug name with the id to obtain something that is still unique, e.g. a channel with "name" `invocation-sender` and id 2123 may be logged as `invocation-sender-2123` (and similarly for nodes).

for reference, here is an example of what a trace looks like now:

```
[2020-10-22T21:46:37Z DEBUG oak_runtime::channel] create new Channel object with ID 2
[2020-10-22T21:46:37Z DEBUG oak_runtime] NodeId(1): new ABI handle 2367277874138833998 maps to Channel 2 WRITE
[2020-10-22T21:46:37Z DEBUG oak_runtime] NodeId(1): new ABI handle 18203277921471555918 maps to Channel 2 READ
[2020-10-22T21:46:37Z DEBUG oak_runtime::proxy] NodeId(1): channel_create(Label { confidentiality_tags: [Tag { tag: Some(TlsEndpointTag(TlsEndpointTag { authority: "localhost1" })) }], integrity_tags: [] }) -> Ok((2367277874138833998, 18203277921471555918))
[2020-10-22T21:46:37Z DEBUG oak_runtime::proxy] NodeId(1): channel_create(Label { confidentiality_tags: [], integrity_tags: [] })
[2020-10-22T21:46:37Z DEBUG oak_runtime::channel] create new Channel object with ID 3
[2020-10-22T21:46:37Z DEBUG oak_runtime] NodeId(1): new ABI handle 3161086210843448291 maps to Channel 3 WRITE
[2020-10-22T21:46:37Z DEBUG oak_runtime] NodeId(1): new ABI handle 2333144199131668344 maps to Channel 3 READ
[...]

PR #1632, opened 2020-10-28
Add `name` to `channel_create`

conradgrobler:
This change adds a debug name field to channels.

This is a step towards #1615.

# Checklist

      construction.


oak_runtime/src/channel.rs
```diff
@@ -280,6 +288,7 @@ impl Channel {
             reader_count: AtomicU64::new(0),
             waiting_threads: Mutex::new(HashMap::new()),
             label: label.clone(),
+            name: name.to_owned(),
```
juliettepretot, line 291: For node names, we presently combine the passed name with the id. 

https://github.com/project-oak/oak/blob/ab7e7f032359e0233ef3f859a9193733a9895372/oak_runtime/src/lib.rs#L1207

Should we do the same for channels?

docs/abi.md
```diff
@@ -181,9 +181,10 @@ If writing to the specified channel would violate
 
 ### `channel_create`
 
-Creates a new unidirectional Channel assigning the label specified by `param[2]`
-and `param[3]` to the newly created Channel, and returns the Channel handles for
-its read and write halves as output parameters in `param[0]` and `param[1]`.
+Creates a new unidirectional Channel assigning the name specified by `param[2]`
+and `param[3]` and the label specified in `param[4]` and `param[5]` to the newly
+created Channel, and returns the Channel handles for its read and write halves
+as output parameters in `param[0]` and `param[1]`.
```
tiziano88, line 187: Could you mention that the name may be empty, does not need to be unique, and is only used for debugging?

docs/programming-oak.md
```diff
@@ -488,8 +488,8 @@ to the room:
 [embedmd]:# (../examples/chat/module/rust/src/lib.rs Rust /.*self\.rooms\.entry\(/ /\}\);$/)
 ```Rust
                 let channel = self.rooms.entry(label.clone()).or_insert_with(|| {
-                    let (wh, rh) =
-                        oak::io::channel_create(&label).expect("could not create channel");
+                    let (wh, rh) = oak::io::channel_create("Room initial channel", &label)
```
tiziano88, line 491: I suspect we can just omit "channel" from any channel name, it should be pretty obvious that they are channels from the context in which they are printed / displayed (hopefully).

oak_runtime/src/channel.rs
tiziano88: I think we should too.

tiziano88 (approved):

docs/abi.md
conradgrobler: Added a paragraph to explain this.

docs/programming-oak.md
conradgrobler: Removed "channel" from the names everywhere.

oak_runtime/src/channel.rs
conradgrobler: Done.

sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -236,7 +236,7 @@ where
 {
     // Create a new channel for request message delivery.
     let (req_sender, req_receiver) =
-        crate::io::channel_create::<GrpcRequest>(&Label::public_untrusted())
+        crate::io::channel_create::<GrpcRequest>("gRPC request", &Label::public_untrusted())
```
rbehjati, line 239: What is the difference between this and 
```
 .channel_create("gRPC request", &label)
```
in `oak_runtime/src/node/grpc/server/mod.rs`? Does one of these need a different name?

rbehjati (approved):
Nice! :)

juliettepretot (approved):

sdk/rust/oak/src/grpc/mod.rs
conradgrobler: I believe this is the equivalent that can used by a Wasm node to send gRPC invocations out to the gRPC client pseudo node. I think that it is fine for both to have the same name as they would be used in different contexts, which should make clear whether it is a gRPC client or server pseudo node.

Issue #1631, opened 2020-10-28
Add a new Tag type representing top (or wildcard) privilege 


rbehjati:
Following the [discussion on Slack](https://project-oak.slack.com/archives/CHE9E13C3/p1603799158194600), we need a way to allow certain trusted nodes (e.g., gRPC and HTTP server nodes) to be able to declassify use-specific content. This is equivalent to having user-specific intermediate nodes to do that. 

Ideally, we should have separate Top tags for each sub-lattice (i.e., separate top tags for user, domain, and wasm tags). However in the first step, we can have a universal top privilege that allows declassifying any combination of the other types of tags. Moreover, this can first be implemented for confidentiality tags. Later on, this should be evolved into separate tags for each sub-lattice, and extended to integrity tags.

Issue #1301, opened 2020-07-30
Forbid declassifying reads and endorsing writes in IFC


tiziano88:
from https://project-oak.slack.com/archives/CHE9E13C3/p1595439459314800

Currently, a Node may use its full privilege for both reads and writes, but I suspect this leads to leaks (though I don't have a concrete example; @aferr suggested one [here](https://project-oak.slack.com/archives/CHE9E13C3/p1595440408318600?thread_ts=1595439459.314800&cid=CHE9E13C3)), or at the very least is unnecessary / inelegant.

I suggest we limit the privilege to allow:

- declassifying writes
- endorsing reads

and disallow the rest.

If / when we come up with a use case that actually requires using privilege in a different way, we can then perhaps start a discussion to re-add this functionality in a way that we feel does not cause leaks. The Coq model @aferr is developing should also allow us to model this behaviour and have a definite result of whether it is actually sound.

conradgrobler:
Looking at this in more detail, I think that we should only allow downgrading (both declassification and endorsement) on writes rather than allowing declassification on writes and endorsement on reads. An endorsing read implies that the node endorses the data before it has seen it.

@tiziano88 WDYT?

tiziano88:
Uhmm I am not sure, to me it still feels that a node should use its privilege when it decides to read from a less trusted input, rather than when it writes to a more trusted output.

aferr:
If the goal is to be able to prove robust declassification it should be fine to allow downgrading of both kinds during both reads or writes. (Assuming also that privilege replaced with using integrity to control declassification and vice versa).

Separately from hyperproperties:
Even though an endorsing read means the node does the endorsement before it knows what the data does, it does at least mean that because a downgrading version of the read call is done instead of a non-downgrading one, that the trusted code is expecting untrusted data. So the code running in the trusted node could sanitize the untrusted data after it gets it. I think as long as long as downgrading reads are separated from normal ones, accidentally endorsing things should be limited.

I think it would be too limiting to prevent BOTH kinds of downgrades during reads, though maybe it is still possible to write useful applications if just declassification (but not endorsement) is allowed during reads. Where possible, I think its best to make these kinds of decisions based on security properties (which would mean allowing either kind of downgrade on either reads or writes).

conradgrobler:
Should we then close this issue if implementing it will not improve any security properties?

tiziano88:
[...]

PR #1633, opened 2020-10-28
Add top privilege tag

rbehjati:
Ref #1631

### DIFF
--- a/oak_abi/proto/label.proto
+++ b/oak_abi/proto/label.proto
@@ -43,6 +43,7 @@ message Tag {
     WebAssemblyModuleTag web_assembly_module_tag = 2;
     WebAssemblyModuleSignatureTag web_assembly_module_signature_tag = 3;
     TlsEndpointTag tls_endpoint_tag = 4;
+    Top top_tag = 5;
   }
 }
 
@@ -82,3 +83,8 @@ message TlsEndpointTag {
   // using the set of Certificate Authorities (CA) that are available to it.
   string authority = 1;
 }
+
+// Message representing top element of the principal lattice. It can be used as a confidentiality
+// tag to represent the top-secret level, or as a privilege to represent infinite downgrading
+// privilege.
+message Top {}
--- a/oak_abi/src/label/mod.rs
+++ b/oak_abi/src/label/mod.rs
@@ -66,8 +66,16 @@ impl crate::proto::oak::label::Label {
         // The target label must have (compared to the self label):
         // - same or more confidentiality tags
         // - same or fewer integrity tags
-        self_confidentiality_tags.is_subset(&other_confidentiality_tags)
-            && other_integrity_tags.is_subset(&self_integrity_tags)
+        // A special case is the `top` tag that may represent the set of all confidentiality tags,
+        // or the set of all integrity tags. Therefore, to be more accurate:
+        // - the target label must be `top`, or have same or more confidentiality tags than the self
+        //   label, and
+        // - the self label must be `top`, or the target label must have same or fewer integrity
+        //   tags than the self label
+        (other_confidentiality_tags.contains(&top())
+            || self_confidentiality_tags.is_subset(&other_confidentiality_tags))
+            && (self_integrity_tags.contains(&top())
+                || other_integrity_tags.is_subset(&self_integrity_tags))
     }
 }
 
@@ -124,4 +132,10 @@ pub fn tls_endpoint_tag(authority: &str) -> Tag {
             authority: authority.to_string(),
         })),
     }
+}
+
+/// Convenience function for creating the top tag.
+pub fn top() -> Tag {
+    let tag = Some(crate::proto::oak::label::tag::Tag::TopTag(Top {}));
+    Tag { tag }
 }
--- a/oak_runtime/src/lib.rs
+++ b/oak_runtime/src/lib.rs
@@ -39,7 +39,7 @@ use itertools::Itertools;
 use log::{debug, error, info, trace, warn};
 use node::NodeFactory;
 use oak_abi::{
-    label::{Label, Tag},
+    label::{top, Label, Tag},
     proto::oak::application::{ApplicationConfiguration, ConfigMap, NodeConfiguration},
     ChannelReadStatus, OakStatus,
 };
@@ -227,6 +227,19 @@ impl NodePrivilege {
         Self {
             can_declassify_confidentiality_tags,
             can_endorse_integrity_tags,
+        }
+    }
+
+    /// Return the infinite privilege.
+    ///
+    /// A Node with this privilege can downgrade any data regardless of its label. It should only
+    /// be used by the trusted pseudo-nodes.
+    pub(crate) fn top_privilege() -> Self {
+        let mut top_tag = HashSet::new();
+        top_tag.insert(top());
+        NodePrivilege {
+            can_declassify_confidentiality_tags: top_tag.clone(),
+            can_endorse_integrity_tags: top_tag,
         }
     }
 }
@@ -575,9 +588,19 @@ impl Runtime {
     fn get_node_downgraded_label(&self, node_id: NodeId, initial_label: &Label) -> Label {
         // Retrieve the set of tags that the node may downgrade.
         let node_privilege = self.get_node_privilege(node_id);
-        Label {
+        let node_has_top_privilege = node_privilege
+            .can_declassify_confidentiality_tags
+            .contains(&top());
+
+        let confidentiality_tags = if node_has_top_privilege {
+            // Remove all the confidentiality tags if the node has the `top` privilege.
+            // TODO(#1631): When we have a separate top for each sub-lattice, this check should be
+            // done separately for each sub-lattice, removing only the tags belonging to that
+            // sub-lattice.
+            vec![]
+        } else {
             // Remove all the confidentiality tags that the Node may declassify.
-            confidentiality_tags: initial_label
+            initial_label
                 .confidentiality_tags
                 .iter()
                 .filter(|t| {
@@ -586,7 +609,10 @@ impl Runtime {
                         .contains(t)
                 })
                 .cloned()
-                .collect(),
+                .collect()
+        };
+        Label {
+            confidentiality_tags,
             // Add all the integrity tags that the Node may endorse.
             integrity_tags: initial_label
                 .integrity_tags
--- a/oak_runtime/src/node/grpc/server/mod.rs
+++ b/oak_runtime/src/node/grpc/server/mod.rs
@@ -22,7 +22,7 @@ use crate::{
     metrics::Metrics,
     node::{
         grpc::{codec::VecCodec, to_tonic_status},
-        ConfigurationError, Node,
+        ConfigurationError, Node, NodePrivilege,
     },
     proto::oak::invocation::{GrpcInvocation, GrpcInvocationSender},
     RuntimeProxy,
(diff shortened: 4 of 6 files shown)


oak_abi/src/label/mod.rs
```diff
@@ -66,8 +66,18 @@ impl crate::proto::oak::label::Label {
         // The target label must have (compared to the self label):
         // - same or more confidentiality tags
         // - same or fewer integrity tags
-        self_confidentiality_tags.is_subset(&other_confidentiality_tags)
-            && other_integrity_tags.is_subset(&self_integrity_tags)
+        (other_confidentiality_tags.contains(&Tag::universal_top_tag())
```
rbehjati, line 69: Updated this check for completeness, but I am not sure if we ever want to use the top tag as part of a label (perhaps when we switch to robust declassification and transparent endorsement?). Currently, I am only using it to set the privilege of HTTP and gRPC server nodes.

rbehjati (commented):
@tiziano88 @conradgrobler I am not entirely happy with the naming. Let me know if you have better suggestions :)

oak_abi/proto/label.proto
```diff
@@ -82,3 +83,8 @@ message TlsEndpointTag {
   // using the set of Certificate Authorities (CA) that are available to it.
   string authority = 1;
 }
+
+// Message representing top-secret. Declaring this as an enum results in conflicting `derive`
```
tiziano88, line 87: This is not yet a confidentiality label, it is just a principal that may be used as part of a label, so top-secret is not accurate, I would just mention that this is the top element of the principal lattice, and perhaps that it may be used in a confidentiality label to represent the top-secret level, and also as privilege to represent infinite downgrading privilege.

oak_abi/proto/label.proto
```diff
@@ -82,3 +83,8 @@ message TlsEndpointTag {
   // using the set of Certificate Authorities (CA) that are available to it.
   string authority = 1;
 }
+
+// Message representing top-secret. Declaring this as an enum results in conflicting `derive` 
+// rules between the ones that Prost automatically generates for enums, and the ones that are
```
tiziano88, line 88: I am not sure I understand the sentence about enum, could you clarify? Are you talking about a Rust enum?

oak_abi/src/label/mod.rs
tiziano88: LGTM, @aferr should take a look too, please update the comment as well.

oak_abi/src/label/mod.rs
```diff
@@ -66,8 +66,18 @@ impl crate::proto::oak::label::Label {
...
+            || self_confidentiality_tags.is_subset(&other_confidentiality_tags))
+            && (self_integrity_tags.contains(&Tag::universal_top_tag())
+                || other_integrity_tags.is_subset(&self_integrity_tags))
+    }
+}
+
+impl Tag {
+    /// Convenience function for creating top privilege tag that can declassify all types of tags.
```
tiziano88, line 77: I think this comment mixes privilege and tags a bit, AFAICT it is just creating an instance of the top tag. Also I would just call it top, universal_top seems just another name for the same thing, and it is already called top elsewhere (and in literature)

oak_abi/src/label/mod.rs
```diff
@@ -66,8 +66,18 @@ impl crate::proto::oak::label::Label {
...
+    pub fn universal_top_tag() -> Tag {
```
tiziano88, line 78: Why do we need a method for this particular tag but not others? Or do we already have methods for the others elsewhere? I seem to remember we have constructors for various types of labels (though perhaps not tags).

oak_runtime/src/lib.rs
```diff
@@ -229,6 +229,19 @@ impl NodePrivilege {
             can_endorse_integrity_tags,
         }
     }
+
+    /// Return the infinite privileged.
```
tiziano88, line 233:
```suggestion
    /// Return the infinite privilege.
```

oak_runtime/src/lib.rs
```diff
@@ -229,6 +229,19 @@ impl NodePrivilege {
...
+    ///
+    /// A Node with this privilege can declassify any data regardless of its label. It should only
```
tiziano88, line 235:
```suggestion
    /// A Node with this privilege can downgrade any data regardless of its label. It should only
```

oak_runtime/src/lib.rs
```diff
@@ -575,9 +588,15 @@ impl Runtime {
     fn get_node_downgraded_label(&self, node_id: NodeId, initial_label: &Label) -> Label {
         // Retrieve the set of tags that the node may downgrade.
         let node_privilege = self.get_node_privilege(node_id);
-        Label {
-            // Remove all the confidentiality tags that the Node may declassify.
-            confidentiality_tags: initial_label
+
+        // Remove all the confidentiality tags that the Node may declassify.
```
tiziano88, line 592: @aferr is there a more elegant way of computing a downgraded label in general, perhaps using meet and join operator from the underlying principal lattice, if we had them?

oak_runtime/src/node/grpc/server/mod.rs
```diff
@@ -126,6 +126,10 @@ impl Node for GrpcServerNode {
         "grpc-server"
     }
 
+    fn get_privilege(&self) -> NodePrivilege {
+        NodePrivilege::top_privilege()
```
tiziano88, line 130: Could you add a comment about why this needs this privilege?

tiziano88 (commented):
Thanks!

oak_abi/proto/label.proto
rbehjati: I wanted to have it as an enum. Prost converts this to a Rust enum, but then there is conflict in the derived annotation: 
```
error[E0119]: conflicting implementations of trait `std::cmp::Eq` for type `proto::oak::label::TopTag`:
   --> /opt/my-project/oak_runtime/target/debug/build/oak_abi-941eb48ba258a197/out/oak.label.rs:111:10
    |
111 | #[derive(Eq, Hash)]
    |          ^^ conflicting implementation for `proto::oak::label::TopTag`
...
114 | #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
    |                                         -- first implementation here
    |
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

It is related to this issue: https://github.com/danburkert/prost/issues/332.

Removed it from the comment.

oak_abi/src/label/mod.rs
rbehjati: Done.

oak_abi/src/label/mod.rs
rbehjati: I included `universal` in the name to distinguish it from the case where we'll have separate tops for each sub-lattice. I have changed it to `top` for now.

oak_abi/src/label/mod.rs
rbehjati: We have methods for creating other types of tags, but they are not members of `Tag`. They are ordinary functions. `top` is a special value, and felt more like `Label::public_untrusted`, so I had it as a method. Now changed it to be a function too.

oak_runtime/src/node/grpc/server/mod.rs
rbehjati: Done.

oak_runtime/src/tests.rs
```diff
@@ -316,6 +316,51 @@ fn create_channel_with_more_confidential_label_from_public_node_with_privilege_o
...
+            };
+
+            {
+                // Writing to a more confidential Channel is always allowed.
+                let result = runtime.channel_write(write_handle, message.clone());
+                assert_eq!(Ok(()), result);
+            }
+
+            {
+                // Reading from a more confidential Channel is allowed because of the privilege.
```
conradgrobler, line 354: I am planning on implementing a change (see #1301) that will stop this from working. A more accurate test that will continue working in future would be that a non-public node is able to write to a less confidential (e.g. public) channel.

oak_runtime/src/node/grpc/server/mod.rs
```diff
@@ -126,6 +126,12 @@ impl Node for GrpcServerNode {
         "grpc-server"
     }
 
+    fn get_privilege(&self) -> NodePrivilege {
+        // This node needs to have `top` privilege to be able to declassify data tagged with a
```
conradgrobler, line 130: nit: I think it would be even clearer to say something like "... to declassify data tagged with any arbitrary user identities"

conradgrobler (approved):

oak_runtime/src/lib.rs
aferr: If we go with the NMIFC rules for downgrading, meet/join would be closer to how its done in that paper and the related ones. With privilege as a separate thing from integrity, it might work out similarly if you just take the type rule from that paper and replace the integrity projection with the downgrade privilege. I haven't fully thought this out though. 

I think the way you have it here is perfectly fine, though.

oak_runtime/src/lib.rs
```diff
@@ -575,9 +588,15 @@ impl Runtime {
...
+        // TODO(#1631): Perform this check for each sub-lattice.
+        let confidentiality_tags = match node_privilege
+            .can_declassify_confidentiality_tags
+            .contains(&top())
+        {
+            true => vec![],
+            false => initial_label
```
aferr, line 599: I'm not a Rust programmer, so I might just not know Rust conventions. For me this would be slightly more intuitive if it was an if/else expression to check for top and then a match within the else (not top) branch.

The way you have it already is fine, though.

aferr (approved):

oak_runtime/src/tests.rs
rbehjati: This is a copy of another test (`create_channel_with_more_confidential_label_from_public_node_with_privilege_ok`) except that it uses the top privilege. I think it is better to update both when #1301 is done. 

I think the test you described requires creating an additional node with a more confidential label, which is a more complicated scenario than the tests here. Also, I just noticed that several assertions fail in these tests, but they all still return `ok` (so the CI builds don't fail). I will fix the tests separately.

oak_runtime/src/lib.rs
rbehjati: Thanks. Changed it to an if-else expression.

oak_runtime/src/lib.rs
rbehjati: Do you have a link to the paper?

oak_runtime/src/lib.rs
```diff
@@ -575,9 +588,16 @@ impl Runtime {
...
-            // Remove all the confidentiality tags that the Node may declassify.
-            confidentiality_tags: initial_label
+
+        // Remove all the confidentiality tags that the Node may declassify.
+        // TODO(#1631): Perform this check for each sub-lattice.
+        let confidentiality_tags = if node_privilege
+            .can_declassify_confidentiality_tags
+            .contains(&top())
+        {
+            vec![]
```
tiziano88, line 598: I think this is also worth a comment, it seems a pretty crucial step and we don't want our future selves to accidentally get this wrong when changing things around.

oak_runtime/src/node/grpc/server/mod.rs
tiziano88: I would have also mentioned that this is because we don't have a more restrictive privilege that corresponds to just the user tags for now, anyways I think we have discussed it enough in issues and slack, so perhaps fine to omit.

tiziano88 (approved):

oak_runtime/src/lib.rs
rbehjati: Done.

oak_runtime/src/node/grpc/server/mod.rs
rbehjati: Added a to-do comment.

PR #1635, opened 2020-10-28
Add an HTTP test with non-empty label

rbehjati:
* Adds a new test for HttpServerNode
* Uses a router node in the http tests to allow serving requests with
  non-empty request labels
* Fixes HttpServerNode to use base64 protobuf headers for request labels

### DIFF
--- a/oak_runtime/src/node/http/tests.rs
+++ b/oak_runtime/src/node/http/tests.rs
@@ -15,9 +15,13 @@
 //
 
 use super::*;
+use crate::{
+    io::{ReceiverExt, SenderExt},
+    node::Node,
+};
 use maplit::hashmap;
 use oak_abi::{
-    label::Label,
+    label::{confidentiality_label, tls_endpoint_tag, Label},
     proto::oak::application::{
         node_configuration::ConfigType, ApplicationConfiguration, HttpServerConfiguration,
         NodeConfiguration,
@@ -28,12 +32,12 @@ use std::fs;
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
 static GCP_CA: &str = "../examples/certs/gcp/ca.pem";
 
-/// A simple Oak node that responds with 200 (OK) to every request it receives
-struct OakNodeForTest;
+/// A router node that creates a per-request [`EchoNode`] for each incoming request.
+struct RouterNode;
 
-impl crate::node::Node for OakNodeForTest {
+impl Node for RouterNode {
     fn node_type(&self) -> &'static str {
-        "test"
+        "test-router"
     }
     fn run(
         self: Box<Self>,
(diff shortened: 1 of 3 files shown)


oak_runtime/src/node/http/tests.rs
```diff
@@ -45,6 +46,58 @@ impl crate::node::Node for OakNodeForTest {
         let invocation_receiver = Receiver::<HttpInvocation>::new(ReadHandle { handle });
 
         while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            // Create a request-specific EchoNode, with a privilege that allows the node to
+            // declassify the request.
+            let privilege = NodePrivilege {
+                can_declassify_confidentiality_tags: hashset! {tls_endpoint_tag("localhost")},
+                can_endorse_integrity_tags: hashset! {},
+            };
```
tiziano88, line 54: The fact that we can create nodes with arbitrary privilege made me realise that we should probably make sure that this enforced by the runtime, I created https://github.com/project-oak/oak/issues/1636 to track.

oak_runtime/src/node/http/tests.rs
```diff
@@ -45,6 +46,58 @@ impl crate::node::Node for OakNodeForTest {
         let invocation_receiver = Receiver::<HttpInvocation>::new(ReadHandle { handle });
 
         while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            // Create a request-specific EchoNode, with a privilege that allows the node to
+            // declassify the request.
+            let privilege = NodePrivilege {
+                can_declassify_confidentiality_tags: hashset! {tls_endpoint_tag("localhost")},
```
tiziano88, line 52: This does not seem to create a request-specific node, or better, it seems to create the same exact node for each request, regardless of the label of the request, right? It would make more sense to actually create it based on the label of the request. Which also applies to line 71 below, in which it looks like the newly created node is actually being created as public, instead of matching the request label, so that should also be changed I think.

oak_runtime/src/node/http/tests.rs
```diff
@@ -151,17 +214,59 @@ async fn test_https_server_does_not_terminate_after_a_bad_request() {
...
+    // HTTP requests.
+    let mut http_server_tester = HttpServerTester::new(2528);
+    let client_with_valid_tls = create_client(LOCAL_CA);
+
+    // Send an HTTPS request, and check that response has StatusCode::OK
+    let resp = send_request(
+        client_with_valid_tls,
+        "https://localhost:2528",
+        confidentiality_label(tls_endpoint_tag("localhost")),
+    )
```
tiziano88, line 262: As I mentioned in the other comment, this only works because the label is exactly the one that is used above, but it would be nice to show that multiple labels are declassified. WDYT?

oak_runtime/src/node/http/tests.rs
rbehjati: Are you suggesting that the node should not have the privilege, but should have the same label as the request label? 
In that case, it will be able to read the request but not write the response. Should the response channel be created with the request label too?

oak_runtime/src/node/http/tests.rs
rbehjati: IIUC, only public nodes can create new nodes. With this restriction in privilege, we will never be able to create any node with a privilege unless the initial public node has the top privilege. Is this the assumption?

oak_runtime/src/node/http/tests.rs
rbehjati: I am not sure that this should be the purpose of the HTTP tests. But I agree that the label should be retrieved from the request. I'll fix that.

oak_runtime/src/node/http/tests.rs
```diff
@@ -15,25 +15,37 @@
...
 use std::fs;
 
 static LOCAL_CA: &str = "../examples/certs/local/ca.pem";
 static GCP_CA: &str = "../examples/certs/gcp/ca.pem";
 
-/// A simple Oak node that responds with 200 (OK) to every request it receives
-struct OakNodeForTest;
+fn get_label(runtime: &RuntimeProxy, handle: Handle) -> Option<Label> {
+    match runtime
+        .get_serialized_channel_label(handle, std::usize::MAX)
```
tiziano88, line 35: Maybe we should have a method that gets the actual label, without serializing and deserializing, and without having to pass capacity, which is irrelevant, could you create an issue to track?

oak_runtime/src/node/http/tests.rs
```diff
@@ -15,25 +15,37 @@
...
+        .unwrap()
```
tiziano88, line 36: I think this may be something like

```rust
...unwrap()
  .and_then(|label| Label::decode(label.as_ref()))
```

or something like that, up to you.

oak_runtime/src/node/http/tests.rs
```diff
@@ -15,25 +15,37 @@
...
+    {
+        Success(label) => Some(Label::decode(&label[..]).unwrap()),
```
tiziano88, line 38: Regardless of my other suggestion, the unwrap and re-wrap in Some seems odd, maybe just call `.ok()` on the decoded value? Up to you

oak_runtime/src/node/http/tests.rs
```diff
@@ -45,22 +57,95 @@ impl crate::node::Node for OakNodeForTest {
...
 
         while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            // Compare the request and response labels. If the echo-node (which gets the
+            // request-label) can write to the response-channel, let the `EchoNode` send the
+            // response. Otherwise, the RouterNode should send the response to the caller.
+            let request_label =
+                get_label(&runtime, invocation.clone().receiver.unwrap().handle.handle).unwrap();
+            let response_label =
+                get_label(&runtime, invocation.clone().sender.unwrap().handle.handle).unwrap();
+            let echo = request_label.flows_to(&response_label);
```
tiziano88, line 67: (optional) Maybe s/echo/can_reply/, here and in the struct?

tiziano88, line 67: Nice logic!

oak_runtime/src/node/http/tests.rs
```diff
@@ -45,22 +57,95 @@ impl crate::node::Node for OakNodeForTest {
         let invocation_receiver = Receiver::<HttpInvocation>::new(ReadHandle { handle });
 
         while let Ok(invocation) = invocation_receiver.receive(&runtime) {
+            // Compare the request and response labels. If the echo-node (which gets the
+            // request-label) can write to the response-channel, let the `EchoNode` send the
+            // response. Otherwise, the RouterNode should send the response to the caller.
+            let request_label =
+                get_label(&runtime, invocation.clone().receiver.unwrap().handle.handle).unwrap();
```
tiziano88, line 64: Reading the label from the receiver object seems more idiomatic than doing it via low-level handles.

Edit: actually I'm not sure whether that's implemented natively, or just in the SDK.

tiziano88 (approved):

oak_runtime/src/node/http/tests.rs
tiziano88: Ah yes, makes sense, thanks!

PR #1637, opened 2020-10-28
Implement Decodable and Encodable for Either type

tiziano88:
The `Either` type comes from the https://docs.rs/either/1.6.1/either/
crate instead of being re-implemented from scratch, as that has useful
methods already defined on it, and it is pretty lightweight.

This may be used as foundation for more interesting patterns related to
multiplexing several message types on the same channel, e.g. init and
command messages.

# Checklist

      construction.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -1143,6 +1143,7 @@ dependencies = [
 name = "oak_io"
 version = "0.1.0"
 dependencies = [
+ "either",
  "oak_abi",
  "oak_derive",
  "oak_utils",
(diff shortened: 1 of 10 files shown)


ipetr0v (approved):

ipetr0v:
Do you think we will ever need to support `Enum` also (variant number of parameters, compared to two in `Either`)?

tiziano88:
> Do you think we will ever need to support `Enum` also (variant number of parameters, compared to two in `Either`)?

Presumably that will require some code generation, basically implementing something like serde::Serialize, I don't think we want to go there yet, but let me know if you have any ideas.

PR #1641, opened 2020-10-29
Install `rust-analyzer` in docker image

tiziano88:
Also configure VSCode dev container to use it, so that it is guaranteed
to work with the Rust version we use, and also fix settings so that it
works well with protobuf imports.

# Checklist

      construction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -250,6 +250,13 @@ ARG udeps_location=https://github.com/est31/cargo-udeps/releases/download/${udep
 RUN curl --location ${udeps_location} | tar --extract --gzip --directory=${install_dir} --strip-components=2 ./${udeps_dir}/cargo-udeps
 RUN chmod +x ${install_dir}/cargo-udeps
 
+# Install rust-analyzer
+# https://github.com/rust-analyzer/rust-analyzer
+ARG rust_analyzer_version=2020-10-26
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -250,6 +250,13 @@ ARG udeps_location=https://github.com/est31/cargo-udeps/releases/download/${udep
 RUN curl --location ${udeps_location} | tar --extract --gzip --directory=${install_dir} --strip-components=2 ./${udeps_dir}/cargo-udeps
 RUN chmod +x ${install_dir}/cargo-udeps
 
+# Install rust-analyzer
+# https://github.com/rust-analyzer/rust-analyzer
+ARG rust_analyzer_version=2020-10-26
```
rbehjati, line 255: I guess only few of us use this, and it is not relevant for the CI. Is it possible to instead have a separate Docker file for development that uses this Dockerfile, but adds a layer for rust-analyzer?

rbehjati (approved):

Dockerfile
tiziano88: It is technically possible but I don't think it's worth the extra complexity IMO, we would have to deal with multiple images, including caching and pushing them.

rbehjati: Not everyone would use the development image. Also, we would not need to push it. It is just something that we'd use locally. No?

tiziano88: I wonder if multi stage docker images would help here, I'll look into it separately.

PR #1644, opened 2020-10-29
Add channel names to the rust introspection server

juliettepretot:
# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/channel.rs
+++ b/oak_runtime/src/channel.rs
@@ -121,6 +121,12 @@ impl ChannelHalf {
         self.channel.id
     }
 
+    /// Get the ID of the underlying channel.  For debugging/introspection
+    /// purposes.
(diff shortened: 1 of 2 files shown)


oak_runtime/src/channel.rs
```diff
@@ -121,6 +121,12 @@ impl ChannelHalf {
         self.channel.id
     }
 
+    /// Get the ID of the underlying channel.  For debugging/introspection
+    /// purposes.
+    pub fn get_channel_name(&self) -> String {
```
conradgrobler, line 126: nit: Does this need to be a `String`? Would `&str` not be sufficient?

conradgrobler (approved):

oak_runtime/src/channel.rs
juliettepretot: good point, will make it that

Issue #1634, opened 2020-10-28
Create labeled nodes in Aggregator example


ipetr0v:
Currently Aggregator example creates an Aggregator Node with a public label which performs declassifying reads from client requests.

We need to change its behavior in a following way:
- Create an Aggregator node labeled with its SHA256 hash
- Check that client requests are labeled with Aggregator's SHA256 hash
- Perform a declassifying write when sending aggregated data to the backend server

Originated from https://project-oak.slack.com/archives/CHE9E13C3/p1603883481220500

ipetr0v:
cc @tiziano88

tiziano88:
Cc @conradgrobler

tiziano88:
Increasing priority as I think we should get this right sooner rather than later.

PR #1646, opened 2020-10-29
Check request labels in Aggregator example

ipetr0v:
This change:
- Adds Router Node to Aggregator example
- Checks client request labels in Aggregator example

Fixes https://github.com/project-oak/oak/issues/1634


tiziano88:
Not sure I follow this PR, I thought all incoming requests are expected to be with the same label, so we can probably just check in the current "main" node, basically here:

https://github.com/project-oak/oak/blob/7dca63b78edc241e19807191cb2a42396a51dc62/examples/aggregator/module/rust/src/lib.rs#L246-L251

ipetr0v:
I just thought to refactor it also, since we start using `Router` pattern everywhere - so it all will look similar.

tiziano88:
Got it, thanks, so the previous main node is now the router node (more or less), right?

BTW I think https://github.com/project-oak/oak/pull/1647 (if submitted before this) would simplify things a bit.

ipetr0v:
> Got it, thanks, so the previous main node is now the router node (more or less), right?
Correct

> BTW I think #1647 (if submitted before this) would simplify things a bit.
Sure, I'll rebase on it after it's merged

tiziano88:
FYI #1647 is merged. Also I think #1653 might help with this PR, feel free to take a look at that and comment there too.

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,159 +35,142 @@ pub mod proto {
...
+    // TODO(#1357): Verify that the label corresponds to exactly a single `WebAssemblyModuleTag`.
+    // (label.confidentiality_tags.len() == 1)
+    //     && (if let Some(tag) = &label.confidentiality_tags[0].tag {
+    //         match tag {
+    //             tag::Tag::GrpcTag(_) => true,
+    //             _ => false,
+    //         }
+    //     } else {
+    //         false
+    //     })
```
tiziano88, line 72: Why can this not be implemented now? AFAICT it does not depend on challenge response or any missing functionality.

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,159 +35,142 @@ pub mod proto {
...
-            }
-        }
-        Ok(())
-    }
+impl Router {
+    fn create(config: &Config) -> anyhow::Result<Self> {
+        // Create a gRPC client pseudo-Node with a less restrictive label than the current Node.
+        // In particular, the confidentiality component of the current Node label includes the
+        // "aggregator" WebAssembly hash, which is declassified as part of the gRPC client
+        // Node creation. This is only allowed if the current Node actually has the
```
tiziano88, line 91: Is this accurate? AFAICT the Router node has public label.

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,159 +35,142 @@ pub mod proto {
...
+        let (aggregator_sender, aggregator_receiver) = oak::io::channel_create::<
+            Either<AggregatorInit, oak::grpc::Invocation>,
+        >(
+            "Aggregator init", &Label::public_untrusted()
+        )
+        .context("Couldn't create channel to aggregator node")?;
+        oak::node_create(
+            "aggregator",
+            &oak::node_config::wasm("app", "aggregator"),
+            &Label::public_untrusted(),
```
tiziano88, line 120: This one will have to change too, right? It needs to have the wasm hash label, and will declassify when writing out to the backend.

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,159 +35,142 @@ pub mod proto {
...
+            aggregator_receiver.handle,
+        )
+        .expect("Couldn't create aggregator node");
+        oak::channel_close(aggregator_receiver.handle.handle).expect("Couldn't close receiver channel");
+
+        // Send the initialization message to Aggregator Node containing a gRPC server invocation
+        // receiver and a gRPC client invocation sender.
+        debug!("Sending the initialization message to handler Node");
```
tiziano88, line 128: Does handler == aggregator == Aggregator? If so, could you make the messages and comments more consistent?

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,159 +35,142 @@ pub mod proto {
...
-}
-
-/// Create initialization message for Aggregator Node.
-fn create_init_message(sender: &oak::io::Sender<GrpcInvocation>) -> AggregatorInit {
-    AggregatorInit {
-        grpc_client_invocation_sender: Some(GrpcInvocationSender {
-            sender: Some(sender.clone()),
-        }),
-    }
+    pub grpc_client_connect_address: String,
```
tiziano88, line 173: I think this is the backend server address, right? If so, could you name it that way?

ipetr0v:
Currently blocked because `Router` needs to be initialized, but it can only have a gRPC channel receiver.

Originated from: https://github.com/project-oak/oak/pull/1653#discussion_r515148501

examples/aggregator/module/rust/src/lib.rs
tiziano88: Just checking: do you mean this corresponds to the TODO above, right?

examples/aggregator/config.toml
```diff
@@ -1 +1,3 @@
 grpc_server_listen_address = "[::]:8080"
+grpc_client_connect_address = "https://localhost:8888"
+aggregator_module_hash = "05493131b2041ee7fbbcb41758133b420862c57374d318454646f51014afa838"
```
tiziano88, line 3: I guess this was the only way of avoiding the infinite regress of the hash including itself as part of the hash calculation? :) Apart from splitting the aggregator into a separate wasm module (which BTW I guess would be the correct way of doing this, since currently this hash includes all the nodes, including the public ones, so if you change a single string in the router node, for instance, this hash will also change).

tiziano88:
> Currently blocked because `Router` needs to be initialized, but it can only have a gRPC channel receiver.
> 
> Originated from: [#1653 (comment)](https://github.com/project-oak/oak/pull/1653#discussion_r515148501)

We should probably implement the Init pattern that @wildarch suggested (can be a special case for now, like Either) to unblock this?

tiziano88:
I just hacked this together: https://github.com/project-oak/oak/pull/1659/files I think it would solve this case (and in general, the initialization pattern), do you think that is sufficient @ipetr0v ? also cc @wildarch .

examples/aggregator/config.toml
ipetr0v: I think we had this pattern in TIR (separate modules for nodes) but then decided to merge them together.
Not sure that we need separate modules in this example, since this example is very simple right now.

ipetr0v: And also it would require to make separate modules in all examples, where we want to check incoming modules

tiziano88: It would only be necessary for nodes that perform declassification based on wasm hash or signatures. For others, there is no point in splitting them.

ipetr0v:
Unblocked by https://github.com/project-oak/oak/pull/1675

examples/aggregator/proto/handler_init.proto
```diff
@@ -16,13 +16,13 @@
 
 syntax = "proto3";
 
-package oak.examples.aggregator_init;
+package oak.examples.handler_init;
```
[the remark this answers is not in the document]
ipetr0v: Done

examples/aggregator/config.toml
ipetr0v: Should we add a function that returns hash and public key of a module, for these purposes?

ipetr0v: And we probably wouldn't need to split modules (not sure yet)

ipetr0v: Will divide modules as a separate PR

examples/aggregator/module/rust/tests/integration_test.rs
```diff
@@ -49,7 +49,11 @@ async fn test_aggregator() {
         "oak_main",
         ConfigMap {
             items: hashmap! {
-                "config".to_string() => br#"grpc_server_listen_address = "[::]:8080""#.iter().cloned().collect()
+                "config".to_string() => br#"
+                    grpc_server_listen_address = "[::]:8080"
+                    backend_server_address = "https://localhost:8888"
+                    aggregator_module_hash = "49480d98a87569ba39b0b29bc739c9f863502de73c964a8e54f6d720045de2c3"
```
conradgrobler, line 55: The hash is repreated in many places. Perhaps add documentation somewhere to explain how to get the new hash and all the different places where the hash needs to be updated if the aggregator example needs to be rebuilt.

examples/aggregator/proto/BUILD
```diff
@@ -44,3 +44,22 @@ cc_grpc_library(
     well_known_protos = True,
     deps = [":aggregator_cc_proto"],
 )
+
+proto_library(
```
conradgrobler, line 48: Is this required? The proto files are only used by the rust code on the server side, not by any clients.

conradgrobler (commented):
The CI is failing. It seems that the request receiver on the invocation has an empty label for some reason.

examples/aggregator/module/rust/tests/integration_test.rs
ipetr0v: Done

ipetr0v:
CI is failing because we need labeled channels in `oak_tests`
Done in https://github.com/project-oak/oak/pull/1706 (will rebase this PR once it's merged)

conradgrobler (approved):

examples/aggregator/module/rust/src/lib.rs
```diff
@@ -35,221 +36,49 @@ pub mod proto {
...
+        let config: Config =
+            toml::from_slice(&command.items.get("config").expect("Couldn't find config"))
+                .context("Couldn't parse TOML config file")?;
+        info!("Parsed config: {:?}", config);
+
+        let router_init_sender =
+            oak::io::entrypoint_node_create::<Router>("router", &Label::public_untrusted(), "app")
+                .context("Couldn't create router node")?;
+        let router_command_sender =
+            oak::io::send_init(router_init_sender, command, &Label::public_untrusted())?;
```
tiziano88, line 74: Am I reading this correctly that it is sending the entire configmap to the router node? That seems excessive, also the config in general may contain other details that are irrelevant for the router node. AFAICT the only reason for passing it is in order to create the gRPC client node (which I think should be created here) and in order to pass the wasm module hash (which I think should be hardcoded here, though that's not possible yet because the app is still in a single wasm module for now). Maybe leave a TODO for the latter for now? Though I think the former could be fixed now already.

tiziano88 (approved):

examples/aggregator/module/rust/src/router.rs
```diff
@@ -0,0 +1,137 @@
...
+                self.handler_command_sender
+                    .send(&command)
+                    .context("Couldn't send invocation to worker node")
+            }
+            _ => {
+                anyhow::bail!("Received malformed gRPC invocation");
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
ipetr0v: I think, specifically for this it could be a wrapper around Channel, that checks privileges (probably does the same thing as Channel in Runtime), but does this in SDK.
Since having SDK (non-Runtime) errors was the only reason behind having this piece of code.

examples/aggregator/module/rust/src/router.rs
```diff
@@ -0,0 +1,137 @@
...
+        // let label = Label {
+        //     confidentiality_tags: vec![Tag {
+        //         tag: Some(tag::Tag::TlsEndpointTag(TlsEndpointTag {
+        //             certificate_subject_alternative_name: "google.com".to_string(),
+        //         })),
+        //     }],
+        //     integrity_tags: vec![],
+        // };
+        let grpc_client_invocation_sender = oak::grpc::client::init(&config.backend_server_address)
+            .expect("Couldn't create gRPC client");
```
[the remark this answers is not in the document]
ipetr0v: Handler is also created inside `Router`. Will probably create all nodes in the Main node.
But this will leave `Router` only the check of incoming labels (which later on, I think, should be a part of SDK).

examples/aggregator/module/rust/src/lib.rs
ipetr0v: Adding a TODO(https://github.com/project-oak/oak/issues/1731)

examples/aggregator/client/android/cpp/client.cc
```diff
@@ -46,8 +46,16 @@ JNIEXPORT void JNICALL Java_com_google_oak_aggregator_MainActivity_createChannel
...
+  // Label corresponding to the hash of the WebAssembly module that implements the aggregator logic.
+  //
+  // The intention is that only such module will be able to declassify the data sent over gRPC.
+  // When a new version of the aggregator module is "released", this value needs to be changed to
+  // reflect the hash of the new module.
+  //
+  // The particular value corresponds to the hash on the `aggregator.wasm` line in
+  // https://github.com/project-oak/oak/blob/hashes/reproducibility_index.
+  oak::label::Label label = oak::WebAssemblyModuleHashLabel(
+      absl::HexStringToBytes("e3e58c7838e3b51d38c9ad978094ead1dfb4e32eac72424afb5b493118c5381e"));
```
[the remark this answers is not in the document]
ipetr0v: Done

examples/aggregator/config.toml
```diff
@@ -1 +1,3 @@
 grpc_server_listen_address = "[::]:8080"
+backend_server_address = "https://localhost:8888"
+aggregator_module_hash = "49480d98a87569ba39b0b29bc739c9f863502de73c964a8e54f6d720045de2c3"
```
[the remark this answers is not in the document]
ipetr0v: Correct, I will remove this as part of https://github.com/project-oak/oak/issues/1731

Issue #1639, opened 2020-10-28
init / command pattern for Oak nodes


tiziano88:
It is relatively common (and familiar for programmers) to rely on a mental model in which objects are created via some sort of constructor (let us assume also passing some initial data), and messages are sent to instances (and / or methods are invoked) in a command / event loop.

Currently this pattern is not easy to achieve in Oak, because both the initial message and the subsequent commands must be sent to the same node, but each channel only deals with a single type of message, therefore we need to manually create an extra initialization message that contains the initialization data plus a handle to the channel on which the node will receive subsequent commands:

https://github.com/project-oak/oak/blob/c3b500eb9c54e84f7d88f079c1c14f33e995687a/examples/aggregator/module/rust/src/lib.rs#L154-L178

I am hoping that #1637 and #1638 may help come up with a better approach, but I haven't thought out the details yet.

For instance, we may try to move towards the following pattern:

```rust
struct Worker {
  init_data: InitData,
}

// Implement a gRPC service definition trait.
impl Chat for Worker {
  fn subscribe(&mut self, req: SubscribeRequest) -> grpc::Result<SubscribeResponse> {
    ...
  }

  fn send_message(&mut self: req: SendMessageRequest) -> grpc::Result<SendMessageResponse> {
    ...
  }
}

// Similar to CommandHandler, but has separate types for init and command.
impl oak::CommandNode<InitData, oak::grpc::Invocation> for Worker {
  fn create(init_data: InitData) -> Self {
    Self {
      init_data,
    }
  }

  fn handle_command(&mut self, command: oak::grpc::Invocation) -> anyhow::Result<()> {
    // Instead of wrapping the struct in `ChatDispatcher::new` as we would do currently, we leave this explicit dispatch call here.
    // This avoids an extra level of nesting.
    ChatDispatcher::dispatch(self, command)
  }
}

// I am imagining a more powerful macro than the one we currently have, which creates the entrypoint function
// based on the name of the struct, receives messages from the incoming channel, creates an instance of
// Worker when the first InitData message is received, and then sends commands to it. It may use an Either type
// such as the one implemented in #1637 to discriminate between the two message types, and effectively
// acts as a state machine, rejecting multiple init messages if they arrive after the first one.
oak_entrypoint_command_node!(Worker);
```

@project-oak/core @wildarch thoughts?

tiziano88:
[...]

PR #1647, opened 2020-10-29
Use Either pattern to initialize aggregator node

tiziano88:
# Checklist

      construction.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -73,6 +73,7 @@ dependencies = [
  "aggregator_common",
  "aggregator_grpc",
  "assert_matches",
+ "either",
  "env_logger",
  "itertools",
  "log",
--- a/examples/aggregator/module/rust/Cargo.toml
+++ b/examples/aggregator/module/rust/Cargo.toml
@@ -10,6 +10,7 @@ crate-type = ["cdylib", "rlib"]
 
 [dependencies]
 aggregator_common = { path = "../../common" }
+either = "*"
 itertools = "*"
 log = "*"
 oak = "=0.1.0"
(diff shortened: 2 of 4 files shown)


tiziano88:
Cc @wildarch

wildarch:
As mentioned in https://github.com/project-oak/oak/issues/1639#issuecomment-718848474, could we use something like this instead:
```
// Needs to `impl` `Encodable`, `Decodable`
struct InitWrapper<InitData, Command> {
  init: InitData,
  command_receiver: Receiver<Command>,
}
```

I feel like that better encodes the fact that the node receives init data, followed by a stream of commands. To me `Either` suggests that the message types can alternate.

Of course this is all semantics and mostly SDK internal, so I'm also happy to go with `Either` if you prefer 😄 .

tiziano88:
> As mentioned in [#1639 (comment)](https://github.com/project-oak/oak/issues/1639#issuecomment-718848474), could we use something like this instead:
> 
> ```
> // Needs to `impl` `Encodable`, `Decodable`
> struct InitWrapper<InitData, Command> {
>   init: InitData,
>   command_receiver: Receiver<Command>,
> }
> ```
> 
> I feel like that better encodes the fact that the node receives init data, followed by a stream of commands. To me `Either` suggests that the message types can alternate.
> 
> Of course this is all semantics and mostly SDK internal, so I'm also happy to go with `Either` if you prefer  .

Makes sense, let's go ahead with Either for now, and I'll try your suggested approach right after, I think it may require a bit more crafting of structs to get right, or it may just be an almost drop-in replacement in the best case.

ipetr0v (approved):

PR #1651, opened 2020-10-29
Add label to ReceiverExt and SenderExt in the runtime

### LINKED ISSUE #1649: Add a method to Sender and Receiver in the oak_io for retrieving the label
Sender and Receiver in the SDK currently have a convenience method for retrieving the label. We need something similar to this for the Sender and Receiver in oak_io. For simplicity, consider adding a new `get_channel_label` method to the Runtime similar to `get_serialized_channel_label`.

Ref [comment on #1635](https://github.com/project-oak/oak/pull/1635/files#r514370168).

rbehjati:
Fixes #1649

### DIFF
--- a/oak_runtime/src/io.rs
+++ b/oak_runtime/src/io.rs
@@ -18,7 +18,7 @@
 
 use crate::RuntimeProxy;
 use log::{error, info};
-use oak_abi::{ChannelReadStatus, OakStatus};
+use oak_abi::{label::Label, ChannelReadStatus, OakStatus};
 use oak_io::{Decodable, Encodable, OakError};
 pub use oak_io::{Receiver, Sender};
 
@@ -29,6 +29,9 @@ pub trait ReceiverExt<T> {
 
     /// Waits, reads and decodes a message from the [`Receiver::handle`].
     fn receive(&self, runtime: &RuntimeProxy) -> Result<T, OakError>;
+
+    /// Gets the label associated with this receiver.
+    fn label(&self, runtime: &RuntimeProxy) -> Result<Label, OakError>;
 }
 
 impl<T: Decodable> ReceiverExt<T> for Receiver<T> {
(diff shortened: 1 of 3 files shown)


oak_runtime/src/io.rs
```diff
@@ -29,6 +29,9 @@ pub trait ReceiverExt<T> {
 
     /// Waits, reads and decodes a message from the [`Receiver::handle`].
     fn receive(&self, runtime: &RuntimeProxy) -> Result<T, OakError>;
+
+    /// Get the label associated with this receiver.
+    fn label(&self, runtime: &RuntimeProxy) -> Result<Label, OakError>;
```
tiziano88, line 34: It seems this could just return `Result<Label, OakStatus>`, which is strictly more precise? Not a big deal anyways, I guess we don't do much with the error anyways.

oak_runtime/src/io.rs
```diff
@@ -29,6 +29,9 @@ pub trait ReceiverExt<T> {
 
     /// Waits, reads and decodes a message from the [`Receiver::handle`].
     fn receive(&self, runtime: &RuntimeProxy) -> Result<T, OakError>;
+
+    /// Get the label associated with this receiver.
```
tiziano88, line 33:
```suggestion
    /// Gets the label associated with this receiver.
```

tiziano88 (approved):
Thanks for the quick fix!

oak_runtime/src/lib.rs
```diff
@@ -690,6 +690,19 @@ impl Runtime {
         serialize_label(self.get_node_privilege(node_id).into(), capacity)
     }
 
+    /// Returns the [`Label`] associated with the channel handle.
+    fn get_channel_label(
```
conradgrobler, line 694: Perhaps reuse this function in `get_serialized_channel_label` rather than repeating the same logic?

conradgrobler (approved):

oak_runtime/src/io.rs
rbehjati: I agree that `OakError` does not add anything here, but all the other methods in `ReceiverExt` and `SenderExt` use `OakError` as the error type. I am using `OakError` to be consistent.

oak_runtime/src/lib.rs
rbehjati: Thanks. Done.

PR #1653, opened 2020-10-29
Implement planner + router + worker pattern

tiziano88:
Introduce macro for setting up a Node with a CommandHandler instance.

Ref #1639

# Checklist

      construction.

### DIFF
--- a/examples/chat/module/rust/src/lib.rs
+++ b/examples/chat/module/rust/src/lib.rs
@@ -18,7 +18,7 @@ use anyhow::Context;
 use log::{info, warn};
 use oak::{
     grpc,
-    io::{Receiver, ReceiverExt, Sender, SenderExt},
+    io::{ReceiverExt, Sender, SenderExt},
     Label,
 };
 use oak_abi::proto::oak::application::ConfigMap;
@@ -29,6 +29,29 @@ use std::collections::HashMap;
 mod proto {
     include!(concat!(env!("OUT_DIR"), "/oak.examples.chat.rs"));
 }
+
+oak::entrypoint_command_handler!(main => Main);
+
+/// Main entrypoint of the chat application.
+///
+/// This node is in charge of creating the other top-level nodes, but does not process any request.
+struct Main;
+
+impl oak::CommandHandler<ConfigMap> for Main {
+    fn handle_command(&mut self, _command: ConfigMap) -> anyhow::Result<()> {
+        let grpc_channel =
+            oak::grpc::server::init("[::]:8080").expect("could not create gRPC server pseudo-Node");
+        oak::node_create(
+            &oak::node_config::wasm("app", "router"),
+            &Label::public_untrusted(),
+            grpc_channel.handle,
+        )
+        .expect("could not create router node");
+        Ok(())
+    }
+}
+
+oak::entrypoint_command_handler!(router => Router::default());
 
 /// A node that routes each incoming gRPC invocation to a per-room worker node (either pre-existing,
 /// or newly created) that can handle requests with the label of the incoming request.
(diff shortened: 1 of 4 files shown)


tiziano88:
Personally I like this pattern, even if it introduces an extra `main` node whose job is just setting up the other nodes (including a separate `router` node, instead of merging the two together).

It is also similar to (or at least it moves in the direction of) RFC #1440.

cc @wildarch @project-oak/core

examples/chat/module/rust/src/lib.rs
```diff
@@ -30,6 +30,29 @@ mod proto {
...
+/// Main entrypoint of the chat application.
+///
+/// This node is in charge of creating the other top-level nodes, but does not process any request.
+struct Main;
+
+impl oak::CommandHandler<ConfigMap> for Main {
+    fn handle_command(&mut self, _command: ConfigMap) -> anyhow::Result<()> {
+        let grpc_channel =
+            oak::grpc::server::init("[::]:8080").expect("could not create gRPC server pseudo-Node");
+        oak::node_create(
```
conradgrobler, line 44: Do you have thoughts on how this pattern could be adapted to support the case where the router node needs initial configuration information (e.g. routing decisions are based on info received in the config map)?

conradgrobler (approved):
I think this works well for simple cases and looks clean. I am not sure how we can adapt this pattern for more complex cases where the router or workers need to be initialised with configuration data.

examples/chat/module/rust/src/lib.rs
tiziano88: Yes, that should be doable with the Either type I mentioned elsewhere, more on that in a future PR.

conradgrobler: I was more thinking about channels rather than types. The current implementation sends the grpc invocation channel to the router as the initial channel, so it does not have a channel over which it could send the init message to the router. It could create another init channel, and include the grpc invocation channel handle in that message, but then we are back to the router node needing to read messages from two different channels.

examples/chat/module/rust/src/lib.rs
```diff
@@ -30,6 +30,29 @@ mod proto {
...
+            &oak::node_config::wasm("app", "router"),
+            &Label::public_untrusted(),
+            grpc_channel.handle,
+        )
+        .expect("could not create router node");
+        Ok(())
+    }
+}
+
+oak::entrypoint_command_handler!(router => Router::default());
```
ipetr0v, line 54: What if someone would want to initialize Router - for example pass parameters with gRPC address to which to connect through a gRPC client?

sdk/rust/oak/src/lib.rs
```diff
@@ -539,3 +539,36 @@ macro_rules! entrypoint {
...
+/// struct Main;
+///
+/// impl oak::CommandHandler<ConfigMap> for Main {
+///     fn handle_command(&mut self, command: ConfigMap) -> anyhow::Result<()> {
+///         // ...
+/// #        unimplemented!()
+///     }
+/// }
+///
+/// oak::entrypoint_command_handler!(oak_main => Main);
```
ipetr0v, line 563: I still think that having a regular `main` function will be more obvious and dev friendly rather then having to define multiple structs, like `Main` and `Router` in addition to structs for all the nodes.

And also devs need to treat these new structs as Nodes (since they are created via `node_create`), which increases the number of nodes even in the simple application.

sdk/rust/oak/src/lib.rs
```diff
@@ -539,3 +539,36 @@ macro_rules! entrypoint {
...
     };
 }
+
+/// Similar to [`entrypoint`], but offers a less verbose API in case in which the handler is a
+/// [`CommandHandler`] instance. In this case, the type of incoming messages is inferred from the
+/// generic type associated with the handler instance, and only the entrypoint name needs to be
+/// specified.
+///
+/// The generated implementation of the body of this node sets up logging, and then immediately
+/// calls [`run_command_loop`] on the provided handler.
```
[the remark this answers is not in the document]
tiziano88: That's why I explicitly mentioned in the comment. Do you have suggestions on how to make it more obvious?

examples/chat/module/rust/src/lib.rs
tiziano88: I am working on that, anyways that's an orthogonal issue, right? That's why we are having the discussion on using `Either` or a tuple.

ipetr0v: If so, them why is separate `Main` node necessary, could Router just be the main node?

sdk/rust/oak/src/lib.rs
tiziano88: Do you mean just for the main node, make that a function instead of a struct? My point is that, since it really is a node, we should treat it as a node. I guess it's like in Java even main is inside a class, since everything is a class. Also I think that the planner node @conradgrobler is exploring works better if there is already an assumption that the entry point looks like a node (which it is). Basically I think special casing the main node and make it look different from other nodes, when it actually is the same, is actually more confusing.

sdk/rust/oak/src/lib.rs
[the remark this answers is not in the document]
ipetr0v: I think people wouldn't use logger in production, so they probably would want to have an explicit logger creation.

examples/chat/module/rust/src/lib.rs
tiziano88: That's what I am trying to avoid. I would like each node to correspond to a loop that only handles one type of commands, to keep things organized. This has the advantage that the main node only sets up other nodes, and does not try to read from other channel at some point. Also it means that the main node may be replaced by something more declarative based on a config if we want.

ipetr0v: So right now I need to send init from `Main` to `Router`, right?

tiziano88: Yes, using this patter we would have to do that.

ipetr0v: Also, since we provide `Router` with a gRPC server channel, how can we send an `init` message to it? Since gRPC channel already has a defined message type.

tiziano88: I am hoping Algebraic Data Types, specifically `Either` (or `Pair`), would help here. Haven't fully thought it out though.

ipetr0v: I don't think `Either` would help here.
`Main` creates both `gRPC` and `Router` nodes, and while creating `Router` it provides it with `gRPC` channel (so `gRPC` and `Router` are connected). But the `init` message should be sent from `Main` to `Router`, so it would need an additional channel for this.

tiziano88: You are right, actually I think that @wildarch 's suggestion would work better here: https://github.com/project-oak/oak/issues/1639#issuecomment-718848474

PR #1655, opened 2020-10-30
Experimental support for Encodable and Decodable on top of serde and HandleVisit

wildarch:
# Checklist

      construction.

This demonstrates how we can support a serde-based encoding of messages that is suitable for inter-node communication.

My thinking is we can add different serialization formats later on as needed, and eventually 'mainline' this as an `oak_serde` crate. Users could specify the serialization types they need using feature flags.


experimental/serde_encoding/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use oak::io::{Encodable, Receiver};
+
+    #[derive(Clone, Serialize, Deserialize, HandleVisit, Debug, PartialEq)]
+    struct Test {
+        a: String,
+        b: Receiver<()>,
```
tiziano88, line 38: Does it work with multiple fields having handles? That's the trick case (though I think your code should work)

wildarch: I think this is already covered in generic `HandleVisit` tests such as https://github.com/project-oak/oak/blob/main/oak_derive/tests/visit.rs, but I suppose it doesn't hurt to add a few here as well.

Before I do that though, maybe we should discuss if we actually want to try and eventually merge this PR into `main`? Do you think there is a need for supporting serde based formats, i.e. things like bincode, json etc., and do you think this is a good way to go about it?

tiziano88: Yes I think as we move towards apps with more nodes, it is handy to have a way of specifying inter-node messages in Rust, without having to go all the way to protobuf, as long as they are only used internally. So something like this (or even simpler) would be super useful, and would help reduce the custom logic around serializing enums and structs, such as:

https://github.com/project-oak/oak/blob/1d84fbaebb689d8e858e941841c1147aaf68e381/oak_io/src/lib.rs#L52-L138

experimental/serde_encoding/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+pub struct Bincoded<T>(pub T);
+
+impl<T: Clone + Serialize + HandleVisit> Encodable for Bincoded<T> {
+    fn encode(&self) -> Result<Message, OakError> {
+        Ok(Message {
+            bytes: bincode::serialize(&self.0).map_err(|_| OakError::ProtobufEncodeError(None))?,
+            handles: oak::handle::extract_handles(&mut self.0.clone()),
+        })
+    }
+}
```
tiziano88, line 69: Would it not be possible to have a generic fn to which we can pass a specific serialization fn (e.g. `bincode::serialize` or `serde_json::to_string`) and it runs that in parallel to the handle extraction? As long as that fn returns something that can be treated as a `&[u8]` it should be fine, right? Then we would not need wrapper types for each serialization mechanism.

oak_io/src/serde.rs
```diff
@@ -0,0 +1,54 @@
...
+        S: Serializer,
+    {
+        serializer.serialize_unit()
+    }
+}
+
+impl<'de, T: Encodable + Decodable + HandleVisit> Deserialize<'de> for Sender<T> {
+    fn deserialize<D: Deserializer<'de>>(_: D) -> Result<Self, D::Error> {
+        Ok(Sender::default())
+    }
```
tiziano88, line 53: This would benefit from some extra documentation, in particular clarifying that they are not meant to be serialized by themselves, but rather via the encode and decode methods, and a bit of a high level description (and perhaps some simple test cases) to show what happens if they are serialized with serde (e.g. handle numbers are lost on the round trip conversion).

tiziano88 (commented):
Neat!

In order to validate this, it would be nice to also convert at least one example (e.g. https://github.com/project-oak/oak/blob/main/examples/aggregator/proto/aggregator_init.proto) to use this instead of the specific proto types for inter-node communication. Do you think you could add it to this PR, perhaps as a separate commit, but to submit together?

experimental/serde_encoding/src/lib.rs
wildarch: That's not a bad idea! Since we're keeping this mostly internal and we don't need to mix serializer that'll probably work fine, thanks!

oak_io/src/serde.rs
wildarch: I have tried to make this clear in the top-level module comment, would you prefer to have it on the `impl`s themselves?

I would link to `serde_encoding` here, but I don't think we allow for cross-crate documentation links? 
Adding test cases seems a bit overkill to me, especially since we would need to add a crate like `bincode` to the test dependencies to write those tests.

experimental/serde_encoding/src/lib.rs
wildarch: On second thought, we might actually need to keep the wrapper types around. The reason is that types like `InitWrapper` expect their inner types to be `Encodable + Decodable`, so a type like `InitWrapper<Bincoded<MySerdeStruct>, _>` is valid, but `InitWrapper<MySerdeStruct, _>` is not. This also holds the other way around, you can't have `MySerdeStruct<GrpcInvocation>` and have it be `Serializable + Deserialize` *or* `Encodable + Decodable`, but something like `MySerdeStruct<ProtoSerde<GrpcInvocation>>` would work (`ProtoSerde` would encode the proto as a `Vec<u8>`).

We could try and mitigate this by automatically deriving `Serialize + Deserialize` for all generated protos, but I'm not sure I'm a fan of that.

I'm starting to think the combination of serde and protocol buffers might introduce a lot of sharp edges, what do you think?

PR #1657, opened 2020-10-30
Make command an associated type of CommandHandler

tiziano88:
This makes sense because there is only at most one implementation of
`CommandHandler` for each underlying type. Also allows cleaning up some
type bounds on a few methods.

# Checklist

      construction.

### DIFF
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -35,7 +35,8 @@ oak::entrypoint!(oak_main<ConfigMap> => |_receiver| {
 /// should be modified with care!
 pub struct StaticHttpServer;
 
-impl CommandHandler<Invocation> for StaticHttpServer {
+impl CommandHandler for StaticHttpServer {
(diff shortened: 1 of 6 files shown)


examples/http_server/module/src/lib.rs
```diff
@@ -35,7 +35,8 @@ oak::entrypoint!(oak_main<ConfigMap> => |_receiver| {
 /// should be modified with care!
 pub struct StaticHttpServer;
 
-impl CommandHandler<Invocation> for StaticHttpServer {
+impl CommandHandler for StaticHttpServer {
+    type Command = Invocation;
```
rbehjati, line 39: Perhaps add a blank line here to follow the same convention in the rest of the PR.

rbehjati (approved):
Nice!

examples/http_server/module/src/lib.rs
tiziano88: Done.

PR #1659, opened 2020-10-30
Define InitWrapper message

tiziano88:
# Checklist

      construction.

### DIFF
--- a/oak_io/src/lib.rs
+++ b/oak_io/src/lib.rs
@@ -83,14 +83,108 @@ impl<L: Decodable, R: Decodable> Decodable for Either<L, R> {
     }
 }
 
+/// A wrapper struct that holds an init message, plus the handle of a channel from which to read
+/// command messages. This is useful for patterns in which an Oak node needs some data at
+/// initialization time, but then handles commands of a different type and possibly coming from a
+/// different channel, to be processed in an event-loop pattern.
+#[derive(Clone, Debug, PartialEq)]
+pub struct InitWrapper<Init, Command: Decodable> {
+    init: Init,
+    command_receiver: Receiver<Command>,
+}
+
+/// Implementation of [`Encodable`] for [`InitWrapper`] that encodes the handle of the command
+/// receiver channel in the first handle of the resulting [`Message`].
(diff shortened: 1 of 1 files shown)


oak_io/src/lib.rs
```diff
@@ -83,14 +83,105 @@ impl<L: Decodable, R: Decodable> Decodable for Either<L, R> {
...
+}
+
+/// Implementation of [`Encodable`] for [`InitWrapper`] that encodes the handle of the command
+/// receiver channel in the first handle of the resulting [`Message`].
+impl<Init: Encodable, Command: Encodable + Decodable> Encodable for InitWrapper<Init, Command> {
+    fn encode(&self) -> Result<Message, OakError> {
+        let mut init_message = self.init.encode()?;
+        init_message
+            .handles
+            .insert(0, self.command_receiver.handle.handle);
```
ipetr0v, line 102: Is `0` added here because we encode an `Either`?
Please add a comment

oak_io/src/lib.rs
```diff
@@ -83,14 +83,105 @@ impl<L: Decodable, R: Decodable> Decodable for Either<L, R> {
     }
 }
 
+/// A wrapper struct that holds an init message, plus the handle of a channel from which to read
+/// command messages. This is useful for patterns in which an Oak node needs some data at
+/// initialization time, but then handles commands of a different type in an event-loop pattern.
```
ipetr0v, line 88: Does it mean that `Init` and `Commands` are now sent in separate channels?

ipetr0v (approved):

oak_io/src/lib.rs
tiziano88: The 0 is just the position in which to insert (i.e. at the beginning). Added comment to clarify.

oak_io/src/lib.rs
tiziano88: Yes, clarified. The point is that you can have a router node initialize the worker node with some information, and then send it a channel (coming from somewhere else) from which to read commands (e.g. gRPC invocations). I think this should help with the aggregator and trusted database pattern, let's try and use it and see if it makes things better.

tiziano88 (commented):
Thanks for the quick review.

PR #1661, opened 2020-11-02
Update versions to avoid [RUSTSEC-2020-0059] and [RUSTSEC-2020-0060]

rbehjati:


tiziano88:
I am not sure how / why this solves the issues, and why it is adding non-wildcard dependency versions, but I trust this was the least invasive way to unbreak the build; we can discuss that separately after this is submitted what we should be doing in general about deps perhaps?

tiziano88 (approved):

rbehjati:
I had to specify latests versions for the affected crates, because `cargo update` updated too many things and caused even more conflicts. I think it is because I had to run `cargo update` for each workspace separately. I am surprised that despite having the dependabot, we are still behind the latest for so many crates. 

Apparently with `cargo update` it is possible to update only specific packages, but I did not know that!

rbehjati:
Now I have changed the specific versions back to wildcards (in the `.toml` files).

tiziano88:
got it, yes `cargo update` is a bit obscure, but glad you found out how to selectively upgrade only the selected deps!

PR #1662, opened 2020-11-02
Implement type safe wasm entrypoint

tiziano88:
This allows creating Wasm nodes in a type safe way, returning a channel
of the correct type for the particular node instance, as long as that
instance implements the corresponding traits.

Ref #1639

# Checklist

      construction.

### DIFF
--- a/sdk/rust/oak/src/io/mod.rs
+++ b/sdk/rust/oak/src/io/mod.rs
@@ -49,6 +49,20 @@ pub fn node_create<T: Encodable + Decodable>(
     Ok(sender)
 }
 
+/// Creates a node and corresponding inbound channel of the same type, for nodes that are
+/// instantiated via the [`crate::entrypoint_command_handler`] macro.
+pub fn entrypoint_node_create<T: crate::WasmEntrypoint + crate::CommandHandler>(
+    name: &str,
+    label: &Label,
+    wasm_module_name: &str,
+) -> Result<Sender<T::Command>, OakStatus>
+where
+    T::Command: Encodable + Decodable,
+{
+    let node_config = &crate::node_config::wasm(wasm_module_name, T::ENTRYPOINT_NAME);
+    node_create(name, label, node_config)
+}
+
 /// Map a non-OK [`OakStatus`] value to the nearest available [`std::io::Error`].
 ///
 /// Panics if passed an `OakStatus::Ok` value.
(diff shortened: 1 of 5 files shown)


daviddrysdale (approved):

sdk/rust/oak/src/io/mod.rs
```diff
@@ -49,6 +49,20 @@ pub fn node_create<T: Encodable + Decodable>(
     Ok(sender)
 }
 
+/// Creates a node and corresponding inbound channel of the same type, for nodes that are
+/// instantiated via the [`crate::entrypoint_command_handler`] macro.
```
ipetr0v, line 53: Could there be a situation where `impl ::oak::WasmEntrypoint` is defined in a different file from where `entrypoint_node_create` is called?
In this case it would be hard to debug such error

tiziano88: that trait is not meant to be implemented manually, and even if it were, you would get a regular error from Rust telling you that it's implemented twice, and where the implementations are, so I think it's fine

ipetr0v: I meant that `entrypoint_command_handler` will be called in a different crate (and thus the implementation)

tiziano88: I think even in that case, normal `impl trait` visibility rules should still apply, although I am not 100% sure it's what you are referring to

sdk/rust/oak/src/lib.rs
```diff
@@ -449,6 +449,17 @@ pub fn run_command_loop<N: CommandHandler, R: Iterator<Item = N::Command>>(
     }
 }
 
+/// Trait implemented by structs that correspond to entrypoint of Wasm nodes.
+///
+/// This statically ties together the Wasm entrypoint name (i.e. the name of the Wasm exported
+/// function) to the associated constant of this trait.
+///
+/// This trait is automatically implemented for types used as part of [`entrypoint_command_handler`]
+/// entrypoint definitions.
```
wildarch, line 458: I think it would be useful to link to the `entrypoint_node_create` here in the docs, right now it is hard to find where this trait is actually being used.

sdk/rust/oak/src/io/mod.rs
```diff
@@ -49,6 +49,20 @@ pub fn node_create<T: Encodable + Decodable>(
...
+pub fn entrypoint_node_create<T: crate::WasmEntrypoint + crate::CommandHandler>(
+    name: &str,
+    label: &Label,
+    wasm_module_name: &str,
```
wildarch, line 57: `entrypoint_node_create` only works for entrypoints defined in the same module, right? It seems a bit silly that devs have to specify the current module name manually.

Could we expose a global like `oak::MODULE_NAME` that contains the name of the current module somehow? I suppose we would need to read that from the app manifest?

PR #1667, opened 2020-11-03
Ignore older duplicate versions in deny.toml

conradgrobler:
Ignoring the older version rather than the newer version should mean that cargo-deny should start warning when the old version is no longer used so that we can then clean it up.

# Checklist

      construction.


rbehjati (approved):
Makes sense. Thanks.

PR #1672, opened 2020-11-03
Prove step_implies_lowproj_steps_leq

aferr:
Notably, the way channel append works when writing to a channel needed
to be changed in order to make this provable. Because it is possible for
a public node to send a message to a secret channel, and because a
public observer cannot distinguish between a handle that points to a
secret channel and a handle that points to nothing (becuase the channel
was deleted for example), the write call cannot check whether or not the
handle is allocated because this would cause a leak. This potential leak made
step_implies_lowproj unprovable without changing the semantics.

The change is in chan_append_labeled and the semantics of the write call.
If the write call is to a handle that points to nothing, the write has no
effect and no error is thrown. See chan_append_labeled.

For similar reasons, the checks for "fresh" node ids and handles had to
be removed, though I think these don't correspond to anything in the
real implementation anyway.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -16,6 +16,29 @@ Require Import Coq.Classes.RelationClasses.
 
 Local Open Scope map_scope.
 
+Section misc.
+Theorem can_split_node_index: forall s id n ell,
+    (nodes s).[? id] = {| obj := Some n; lbl := ell |} ->
+    (obj (nodes s).[? id] = Some n) /\
+    (lbl (nodes s).[? id] = ell).
+Proof.
+    intros. split; destruct ((nodes s).[? id]).
+    - unfold RuntimeModel.obj. inversion H. reflexivity.
+    - unfold RuntimeModel.lbl. inversion H. reflexivity.
+Qed.
+
+Theorem can_split_chan_index: forall s han ch ell,
+    (chans s).[? han] = {| obj := Some ch; lbl := ell|} ->
+    (obj (chans s).[? han] = Some ch) /\
+    (lbl (chans s).[? han] = ell).
+Proof.
+    intros. split; destruct ((chans s).[? han]).
+    - unfold RuntimeModel.obj. inversion H. reflexivity.
+    - unfold RuntimeModel.lbl. inversion H. reflexivity.
+Qed.
+
+End misc.
+
 Section low_projection.
 
 Theorem flows_labeled_proj {A: Type}: forall ell (x: @labeled A),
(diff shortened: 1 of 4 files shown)


aferr:
@tiziano88 : In this commit I had to change the semantics of the write call in the model, but I think the implementation actually already works this way. l When a node makes a write call to a handle that points to nothing (for example because the channel got deleted already) does it just fail silently (that is, no error is thrown and the call is essentially a NOP)?

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -16,6 +16,29 @@ Require Import Coq.Classes.RelationClasses.
...
+Qed.
+
+Theorem can_split_chan_index: forall s han ch ell,
+    (chans s).[? han] = {| obj := Some ch; lbl := ell|} ->
+    (obj (chans s).[? han] = Some ch) /\
+    (lbl (chans s).[? han] = ell).
+Proof.
+    intros. split; destruct ((chans s).[? han]).
+    - unfold RuntimeModel.obj. inversion H. reflexivity.
+    - unfold RuntimeModel.lbl. inversion H. reflexivity.
```
jadephilipoom, line 37:
```suggestion
    unfold RuntimeModel.obj; destruct ((chans s).[? han]);
      inversion 1; split; reflexivity.
```

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -229,6 +231,8 @@ Admitted.
 Qed.
 *)
 
+
+
```
jadephilipoom, line 235: nit: extra whitespace

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -249,45 +253,73 @@ Proof.
...
                     ltac:(eauto)) as Hn_idx_s1proj;
                 (erewrite flows_labeled_proj in Hn_idx_s1proj; eauto);
-                inversion H3; crush.
+                inversion H3; crush; subst_lets.
+                all: try replace n0 with n in * by
+                (pose proof (can_split_node_index _ _ _ _ ltac:(eauto));
+                    logical_simplify; congruence). 
+            (*
+                I can't quite get this to work. Coq complains
+                about s1 not being found in environment
```
jadephilipoom, line 262: In which line?

jadephilipoom (approved):

tiziano88:
> @tiziano88 : In this commit I had to change the semantics of the write call in the model, but I think the implementation actually already works this way. When a node makes a write call to a handle that points to nothing (for example because the channel got deleted already) does it just fail silently (that is, no error is thrown and the call is essentially a NOP)?

Actually I think in the current implementation we do signal to the sender whether the channel is orphaned, which I remember we talked about in the past is actually a leak and we should fix it. I believe @daviddrysdale put some thought into this already, but I don't remember what the conclusion was.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: It would error in the line that does:

   | H: (nodes s1).[? id] = _ |- _ =>

I thought the problem might have been that for the internal call the state gets called s1', so I tried doing all: try rename ... . I also tried replacing the s1 with an underscore, but then it doesn't match ever. The tactic did work in the WriteCall case, but it doesn't work if I put it in an "all:" at the front with a try.

aferr:
Argh. I had a commit to take Jade's suggestions, and I thought I pushed it before merging, but I did not. I'll put it into the next PR.

PR #1675, opened 2020-11-03
Allow creating node entrypoint with init data

tiziano88:
Introduce a `WithInit` trait implemented by structs that correspond to
nodes that require receiving initialization data over their inbound
channel, before accepting command from the wrapped channel.

Currently there are two entrypoint-related macros, one for structs that
impl `Default` and do not require init data, and the new one for structs
that impl `WithInit`, which expect an `InitWrapper` message. It may be
possible to unify them in the future with some more macro magic, or we
may just decide to make everything impl `WithInit`, using `()` as init
in case the node does not require init data, but for now I think the
separation makes sense.

Ref #1639

# Checklist

      construction.

### DIFF
--- a/examples/hello_world/module/rust/src/lib.rs
+++ b/examples/hello_world/module/rust/src/lib.rs
@@ -18,30 +18,64 @@ pub mod proto {
     include!(concat!(env!("OUT_DIR"), "/oak.examples.hello_world.rs"));
 }
 
+use either::Either;
 use log::info;
-use oak::{grpc, io::ReceiverExt};
+use oak::grpc;
 use oak_abi::{label::Label, proto::oak::application::ConfigMap};
+use oak_io::Sender;
 use proto::{HelloRequest, HelloResponse, HelloWorld, HelloWorldDispatcher};
 
-oak::entrypoint!(oak_main<ConfigMap> => |_receiver| {
-    oak::logger::init_default();
-    let node = Node {
-        translator: grpc::client::Client::new(
+#[derive(Default)]
+struct Main;
+
+oak::entrypoint_command_handler!(oak_main => Main);
+
+impl oak::CommandHandler for Main {
+    type Command = ConfigMap;
+
+    fn handle_command(&mut self, _command: Self::Command) -> anyhow::Result<()> {
+        let translator_sender_result = oak::io::node_create::<grpc::Invocation>(
             "translator",
+            &Label::public_untrusted(),
             &oak::node_config::wasm("translator", "oak_main"),
-            &Label::public_untrusted()
-        )
-        .map(translator_common::TranslatorClient),
-    };
-    let dispatcher = HelloWorldDispatcher::new(node);
-    let grpc_channel =
-        oak::grpc::server::init("[::]:8080").expect("could not create gRPC server pseudo-Node");
-    oak::run_command_loop(dispatcher, grpc_channel.iter());
-});
+        );
+        let handler_init_sender = oak::io::entrypoint_node_create::<HelloWorldDispatcher<Node>>(
+            "handler",
+            &Label::public_untrusted(),
+            "app",
+        )?;
+        let handler_command_sender = oak::io::send_init(
+            handler_init_sender,
+            translator_sender_result
+                .map(Either::Right)
+                .unwrap_or(Either::Left(())),
+            &Label::public_untrusted(),
+        )?;
+        oak::grpc::server::init_with_sender("[::]:8080", handler_command_sender)?;
+        Ok(())
+    }
+}
 
 struct Node {
     translator: Option<translator_common::TranslatorClient>,
 }
+
+impl oak::WithInit for Node {
+    // Since `Option<T>` is not directly `Encodable` and `Decodable`, we emulate it via a
+    // `Either<(), T>`, which is isomorphic to it.
+    type Init = Either<(), Sender<grpc::Invocation>>;
+
+    fn create(init: Self::Init) -> Self {
+        Node {
+            translator: init
+                .right()
+                .map(|invocation_sender| grpc::client::Client { invocation_sender })
+                .map(translator_common::TranslatorClient),
+        }
+    }
+}
+
+oak::entrypoint_command_handler_init!(node => HelloWorldDispatcher<Node>);
 
 impl Node {
     fn translate(&self, text: &str, from_lang: &str, to_lang: &str) -> Option<String> {
--- a/sdk/rust/oak/src/io/mod.rs
+++ b/sdk/rust/oak/src/io/mod.rs
@@ -51,16 +51,33 @@ pub fn node_create<T: Encodable + Decodable>(
 
 /// Creates a node and corresponding inbound channel of the same type, for nodes that are
 /// instantiated via the [`crate::entrypoint_command_handler`] macro.
-pub fn entrypoint_node_create<T: crate::WasmEntrypoint + crate::CommandHandler>(
+pub fn entrypoint_node_create<T: crate::WasmEntrypoint>(
     name: &str,
     label: &Label,
     wasm_module_name: &str,
-) -> Result<Sender<T::Command>, OakStatus>
-where
-    T::Command: Encodable + Decodable,
-{
+) -> Result<Sender<T::Message>, OakStatus> {
     let node_config = &crate::node_config::wasm(wasm_module_name, T::ENTRYPOINT_NAME);
     node_create(name, label, node_config)
+}
+
+/// Sends an init message over the provided [`Sender`], which is consumed by this method, and
+/// returns a [`Sender`] for subsequent commands.
+///
+/// Useful for sending initial data to nodes instantiated via the
+/// [`crate::entrypoint_command_handler_init`] macro.
+pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
+    sender: Sender<oak_io::InitWrapper<Init, Command>>,
+    init: Init,
+    command_channel_label: &Label,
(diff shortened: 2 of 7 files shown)


tiziano88:
@conradgrobler it is interesting to note that the `WithEntrypoint` trait starts to look (intentionally) more and more like a tiny manifest file for a node, in that it specifies its wasm entrypoint name, as well as the type of messages it handles.

tiziano88:
cc @ipetr0v I think this is a similar pattern to what the aggregator would look like by using the new API for creating nodes etc. Let me know if you think there is anything missing that you think the aggregator example would nedd.

tiziano88:
cc @wildarch in case you have any comments on the Rust API or the high-level design

examples/hello_world/module/rust/src/lib.rs
```diff
@@ -18,31 +18,65 @@ pub mod proto {
...
-});
+        );
+        let handler_init_sender = oak::io::entrypoint_node_create::<HelloWorldDispatcher<Node>>(
+            "handler",
+            &Label::public_untrusted(),
+            "app",
+        )?;
+        let handler_command_sender = oak::io::send_init(
+            handler_init_sender,
+            translator_sender_result
```
conradgrobler, line 49: nit: I know it is basically the same, but this feels clunky and does not read easily. It might be worth implementing `Encodable` and `Decodable` for `Option<T>`, especially if this (having a sender that might or might not exist) is a pattern that might be happen regularly.

conradgrobler (approved):

examples/hello_world/module/rust/src/lib.rs
tiziano88: I agree, I will look into it next.

tiziano88: (though I don't think it will happen often TBH)

sdk/rust/oak/src/io/mod.rs
```diff
@@ -51,18 +51,35 @@ pub fn node_create<T: Encodable + Decodable>(
...
 /// Creates a node and corresponding inbound channel of the same type, for nodes that are
 /// instantiated via the [`crate::entrypoint_command_handler`] macro.
-pub fn entrypoint_node_create<T: crate::WasmEntrypoint + crate::CommandHandler>(
+pub fn entrypoint_node_create<T: crate::WasmEntrypoint>(
     name: &str,
     label: &Label,
     wasm_module_name: &str,
-) -> Result<Sender<T::Command>, OakStatus>
-where
-    T::Command: Encodable + Decodable,
```
ipetr0v, line 60: Why don't we need `Encodable + Decodable` ?

sdk/rust/oak/src/lib.rs
```diff
@@ -593,10 +609,33 @@ macro_rules! entrypoint_command_handler {
...
+            type Message = <$handler as ::oak::CommandHandler>::Command;
+        }
+    };
+}
+
+/// Similar to [`entrypoint_command_handler`], but for nodes that expect to receive initialization
+/// data via an [`oak_io::InitWrapper`] instance over the inbound channel before processing
+/// subsequent commands.
+#[macro_export]
+macro_rules! entrypoint_command_handler_init {
```
ipetr0v, line 626: I think it can replace `entrypoint_command_handler` and just check whether Init message is `()` and thus don't send it

sdk/rust/oak/src/lib.rs
```diff
@@ -452,12 +453,27 @@ pub fn run_command_loop<N: CommandHandler, R: Iterator<Item = N::Command>>(
...
+    /// Name of the WebAssembly exported function that corresponds to the entrypoint of this Node.
     const ENTRYPOINT_NAME: &'static str;
+
+    /// Type of message that the Node accepts on its inbound channel.
+    type Message: Encodable + Decodable;
+}
+
+/// Trait implemented by structs that require an initialization step by receiving data over the
+/// inbound channel (usually wrapped within a [`oak_io::InitWrapper`] message).
+pub trait WithInit {
```
ipetr0v, line 470: Will devs need to use this trait by themselves or it will just be automatically used in existing macros?

sdk/rust/oak/src/io/mod.rs
```diff
@@ -51,18 +51,35 @@ pub fn node_create<T: Encodable + Decodable>(
...
+///
+/// Useful for sending initial data to nodes instantiated via the
+/// [`crate::entrypoint_command_handler_init`] macro.
+pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
+    sender: Sender<oak_io::InitWrapper<Init, Command>>,
+    init: Init,
+    command_channel_label: &Label,
+) -> Result<Sender<Command>, oak_io::OakError> {
+    let (command_sender, command_receiver) =
+        channel_create::<Command>("command sender", command_channel_label)?;
```
ipetr0v, line 74: Does it mean that devs don't have control over naming this channel?

examples/hello_world/module/rust/src/lib.rs
```diff
@@ -18,31 +18,65 @@ pub mod proto {
...
+}
 
 struct Node {
     translator: Option<translator_common::TranslatorClient>,
 }
 
+impl oak::WithInit for Node {
+    // Since `Option<T>` is not directly `Encodable` and `Decodable`, we emulate it via a
+    // `Either<(), T>`, which is isomorphic to it.
+    type Init = Either<(), Sender<grpc::Invocation>>;
```
ipetr0v, line 66: Should we use `InitWrapper` instead of `Either`, so it will just be a struct with `.init` rather that `left()` or `right()` ?

ipetr0v (approved):

sdk/rust/oak/src/io/mod.rs
tiziano88: The bounds are now already in `WasmEntrypoint`.

sdk/rust/oak/src/io/mod.rs
tiziano88: Correct, though they can always use the underlying channel_create if they really want. My guess is that the naming of inbound channels is not that important in general for debugging, it should be clear enough. But let's see how things go once we use the names in more places for logging / debugging.

sdk/rust/oak/src/lib.rs
tiziano88: The macros use it, but nothing stops devs to use it manually if they want to, it's effectively just a constructor. In general I don't think it's very useful to manually create a struct which is normally part of a node anyways.

sdk/rust/oak/src/lib.rs
tiziano88: Thanks, that's an interesting idea! basically if init is () then it would mean that the channel is actually a receiver of command instances, right? Not sure it's possible to express this as a single type though, nor how to make the macro match on it. I'll leave it out for now, but please feel free to try it out in a separate PR!

examples/hello_world/module/rust/src/lib.rs
tiziano88: The point is the optionality, not the init aspect. How would `InitWrapper` help?

ipetr0v: I just thought that we already have a struct for sending init messages alongside commands:
https://github.com/project-oak/oak/blob/bd8ecb84ab8a4788243a73d5a2daf39782bd71f6/oak_io/src/lib.rs#L90-L94

We might use it everywhere

tiziano88: I am not sure I follow.

`InitWrapper` contains one non-optional value, plus a receiver for other values.

What we need here is only a way of passing a handle, or nothing at all.

What is your suggestion for how to fill in the associated type to make it work the way it does now?

```rust
impl oak::WithInit for Node {
    type Init = InitWrapper<grpc::Invocation, grpc::Invocation>;
```

something like this?

tiziano88: Also note that `InitWrapper` is a product type (i.e. a struct), but `Either` is a sum type (i.e. an enum), so I don't think there is a way of using them isomorphically (but let me know if I am mistaken).

ipetr0v: Ah, so this message gets wrapped in the `InitWrapper` already. I think I just got confused that multiple usages of the word `Init`

ipetr0v: But I still don't get why `Either` even though `left()` is never used?

tiziano88: Correct, InitWrapper wraps Init + Receiver<Command>, then Init is handles by this trait, and Command is handled by CommandHandler.

Either<(), T> is isomorphic to Option<T> (if you think about it, Option is defined even in Rust as either None (i.e. `()`) or Some(T) (i.e. `T` itself), I am using it because Option<T> is not Encodable / Decodable. Makes sense?

ipetr0v: In this (isomorphic) case: why would you need `Option` if `None` is never used ?

tiziano88: It is used below:

```rust
    fn create(init: Self::Init) -> Self {
        Node {
            translator: init // Either<(), T>
                .right() // Option<T>
                .map(|invocation_sender| grpc::client::Client { invocation_sender }) // Option<T>::map
                .map(translator_common::TranslatorClient),
        }
    }
```

I added some comments to clarify what the types are in the chain, does it help?

tiziano88: The entire chain will evaluate to `None` if the value was `Either::Left(())`.

PR #1681, opened 2020-11-03
Implement main / handler pattern for more examples

### LINKED ISSUE #1679: missing instructions on how to update the private set intersection example when the Wasm module is modified
Not sure if actually missing or I am looking in the wrong place.

@ipetr0v do you know if we have anything like this? Or do people just do it from scratch every time?

tiziano88:
Fix #1679

# Checklist

      construction.

### DIFF
--- a/examples/private_set_intersection/README.md
+++ b/examples/private_set_intersection/README.md
@@ -1,0 +1,17 @@
+# Private Set Intersection
+
+## To update this example
+
+1. Build the example, including the Wasm module
+1. Push the module to GS via
+   `./scripts/push-example -e private_set_intersection`
+1. Fix the URL and hash in [`./oak_app_manifest.toml`](./oak_app_manifest.toml)
+1. Sign the module with the test key that is checked in the repository (only for
+   test / debug use):
+
+   ```bash
+   ./scripts/oak_sign sign \
+     --private-key=examples/keys/ed25519/test.key \
+     --input-file=examples/private_set_intersection/bin/private_set_intersection.wasm \
+     --signature-file=examples/private_set_intersection/private_set_intersection.sign
+   ```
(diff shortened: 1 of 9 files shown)


examples/private_set_intersection/README.md
```diff
@@ -0,0 +1,16 @@
+# Private Set Intersection
```
ipetr0v, line 1: Should we move these instructions to `examples/README.md` so they could apply to all examples?

examples/private_set_intersection/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+1. Build the example, including the Wasm module
+1. Push the module to GS via
+   `./scripts/push-example -e private_set_intersection`
+1. Fix the URL and hash in [`./oak_app_manifest.toml`](./oak_app_manifest.toml)
+1. Sign the module with
+
+   ```bash
+   ./scripts/oak_sign sign \
+     --private-key=examples/keys/ed25519/test.key \
```
ipetr0v, line 13: I think we need an explicit comment stating that here we use a Test key that is only used for debugging

ipetr0v (approved):

examples/private_set_intersection/README.md
tiziano88: I think it needs more work to make them applicable to the other examples (do any other examples actually rely on signatures and uploading modules?), perhaps let's do it in a separate PR.

examples/private_set_intersection/README.md
tiziano88: Done.

examples/private_set_intersection/README.md
ipetr0v: Agree
Currently we only have Aggregator and Private Set Intersection that use remotely downloaded modules

PR #1685, opened 2020-11-04
Fix WriteChannel case for low_proj_steps_implies_leq_step

jadephilipoom:
A cleanup of work from a pair-programming session with @aferr. I changed the format of `state_upd_{node,chan}_unwind` because it was helpful for `eauto with unwind` to know what to plug in, but it would also work with the original version to do `eapply state_upd_node_unwind with (n1:=Labeled _ _ _); [ reflexivity .. | ]`.

I tried proving the new versions to make sure they were provable, but I don't think they are -- and I don't think the old ones are either, actually, without functional extensionality. The issue is that `low_eq` expects Leibniz equality (`eq`), but `nodes` and `chans` are both functions, so unless the two are constructed in exactly the same way they will not be Leibniz-equal. Was this addressed somehow in the previous version of the code, before labels were introduced? If not, the `low_eq` definition should probably be changed (I made a suggestion in a comment).

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI.v
@@ -124,8 +124,8 @@ Hint Resolve multi_system_ev_refl multi_system_ev_tran : multi.
 
 (* Hints for [eauto with unwind] *)
 Hint Resolve state_upd_chan_unwind chan_append_unwind chan_low_proj_loweq
-    chan_low_proj_idempotent state_upd_node_eq_unwind set_call_unwind 
-    state_upd_chan_eq_unwind state_low_proj_loweq 
+    chan_low_proj_idempotent state_upd_node_unwind set_call_unwind
+    state_upd_chan_unwind state_low_proj_loweq
     state_upd_chan_labeled_unwind
     state_chan_append_labeled_unwind: unwind.
 Hint Extern 4 (node_low_eq _ _ _) => reflexivity : unwind.
@@ -341,6 +341,33 @@ Proof.
             congruence.
 Qed.
 
+Lemma separate_lableled {A} (o : option A) (l : level) (x : labeled) :
+  obj x = o -> lbl x = l -> x = Labeled _ o l.
+Proof. destruct x; cbn; congruence. Qed.
+
+Ltac separate_hyp T :=
+  repeat match goal with
+         | H : ?s = Labeled T ?o ?l |- _ =>
+           assert (obj s = o /\ lbl s = l) by (rewrite H; tauto);
+           clear H; logical_simplify
+         end.
+Ltac separate_goal := apply separate_lableled.
+
+Lemma invert_chans_state_low_proj_flowsto ell lvl s han :
+  lvl <<L ell ->
+  lvl <<L (chans (state_low_proj ell s)).[? han].(lbl) ->
+  lvl <<L (chans s).[? han].(lbl).
+Proof.
+  destruct s.
+  repeat match goal with
+         | _ => progress cbn [state_low_proj
+                               RuntimeModel.chans RuntimeModel.lbl ]
+         | _ => progress cbv [low_proj chan_state_low_proj fnd]
+         | _ => destruct_match
+         | _ => tauto
+         end.
+Qed.
+
 Theorem low_proj_steps_implies_leq_step: forall ell s s1' e1,
     (step_system_ev (state_low_proj ell s) s1' e1) ->
     (exists s2' e2,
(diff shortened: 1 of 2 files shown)


aferr:
You're right, the state update unwind theorems are not provable like this without FE. (And they also were not provable before the new labeled thing. Before the [change to using low-projections](https://github.com/project-oak/oak/pull/1520/files) in the low-equivalence definitions, FE might not have been needed maybe (?). ). I figured the proofs would need to use FE, but it's also possible to break soundness with FE if it isn't used carefully, right? With the change to the LEQ defs you suggested, FE should not be needed at all, right? If so, that sounds better.

jadephilipoom:
> I figured the proofs would need to use FE, but it's also possible to break soundness with FE if it isn't used carefully, right?

Yes. And since you're not going to use these functions in a way that depends on their bodies, and you're using custom equalities anyway, it shouldn't be _too_ difficult to avoid using it.

> With the change to the LEQ defs you suggested, FE should not be needed at all, right?

Correct! And since everything is in terms of LEQ anyway, it will hopefully not be a huge change.

aferr (approved):

jadephilipoom:
Looks like this needs a nontrivial rebase on top of #1672; will fix and resubmit.

aferr:
> Looks like this needs a non-trivial rebase on top of #1672; will fix and resubmit.

Sorry about that! I hope it's not too annoying, I maybe should have let you merge this one first.

jadephilipoom:
No worries, it actually wasn't that bad in the end! I re-requested review just for a sanity check on the new version.

aferr (approved):

aferr:
Go ahead and submit this, but continuing discussion:

Actually, the proof structure that I have relies on having a definition of state low-projection (which is a function that ultimately produces a new state) in addition to just having a low-equivalence on states (a relation on states). For example, there are the lemmas about taking steps to/from low-projections of states. I think the definition you proposed to avoid using FE gives us a relation, but then we lose the definition of low projection. 

Another nearby answer could be to define low-equivalence on states as you proposed (and not based on a definition of low projection), but also:
- still separately define low projection on states
- write a lemma that says if the low projection of s1 = low proj of s2, then s1 low-eq s2.
but I think proving that lemma would still require FE, so I think it might still be better to just use the definitions as they already are than to use this nearby answer.

With the definitions as they are, is it likely to be safe to use FE to prove them?

jadephilipoom:
> Actually, the proof structure that I have relies on having a definition of state low-projection (which is a function that ultimately produces a new state) in addition to just having a low-equivalence on states (a relation on states). For example, there are the lemmas about taking steps to/from low-projections of states. I think the definition you proposed to avoid using FE gives us a relation, but then we lose the definition of low projection.

I don't think the definition of low projection needs to change, does it? I proposed:
```coq
Definition low_eq ell x y :=
  (forall id, nodes (low_proj_ell x) id = nodes (low_proj ell y) id)
￼   /\ (forall han, chans (low_proj_ell x) han = chans (low_proj ell y) han)
```

It's still stated in terms of `low_proj`, and the definition of `low_proj` would stay exactly the same -- the change would be only that proofs using `low_eq` would need to reason about functions that _behave the same_, as opposed to functions that literally _are_ the same.

> * write a lemma that says if the low projection of s1 = low proj of s2, then s1 low-eq s2.
>   but I think proving that lemma would still require FE, so I think it might still be better to just use the definitions as they already are than to use this nearby answer.

I don't think that lemma would require FE, actually -- it would to go in the other direction, but `s1 = low_proj s2` is a stronger statement than `low_eq s1 s2`. Isn't this similar to `forall (f g : A -> B), f = g -> forall x, f x = g x`?

aferr:
> It's still stated in terms of low_proj,

Right, this is a low-equivalence relation defined on states (i.e. `level -> state -> state -> Prop`) defined in terms of the low_proj definition on Labeled things. But to keep the same proof structure (for example to have a lemma like `step_implies_lowproj_steps_leq`) we also need another low projection definition on states (that is something of type `level -> state -> state`).

With this definition of low-equivalence on states, we don't have the `level -> state -> state` definition needed to write some of the key lemmas anymore.

aferr:
(And sorry, I re-read my reply 2 comments up and think it was confusingly written, so edited it just now).

aferr:
Nope, sorry I see what you're saying now. I'm just being dumb. Hope I wasn't too frustrating.

jadephilipoom:
Don't worry, not frustrating at all! Glad we converged :+1:

Issue #1678, opened 2020-11-03
unclear error message when running private set intersection example with wrong module signature


tiziano88:
When running the private set intersection with an invalid / outdated module signature, the client does not get much information of what is failing and why:

```
examples ❯ private_set_intersection ❯ run ❯ background server ❯ run clients ❯ rust ❯ run ║ [2020-11-03T22:36:59Z INFO  psi_client_bin] Connecting to Oak Application: https://localhost:8080/
examples ❯ private_set_intersection ❯ run ❯ background server ❯ run clients ❯ rust ❯ run ║ Error: Couldn't submit set
examples ❯ private_set_intersection ❯ run ❯ background server ❯ run clients ❯ rust ❯ run ║
examples ❯ private_set_intersection ❯ run ❯ background server ❯ run clients ❯ rust ❯ run ║ Caused by:
examples ❯ private_set_intersection ❯ run ❯ background server ❯ run clients ❯ rust ❯ run ║     status: InvalidArgument, message: "could not process gRPC request", details: [], metadata: MetadataMap { headers: {} }
```

cc @ipetr0v

ipetr0v:
Will add a check for incoming requests similar to https://github.com/project-oak/oak/issues/1674
Since currently request just fails at the level of Runtime

ipetr0v:
And it will probably require to add a Router node to this example also

PR #1697, opened 2020-11-05
Check request labels in Private Set Intersection

ipetr0v:
This change:
- Adds Router node
- Checks incoming request labels

Fixes https://github.com/project-oak/oak/issues/1678


examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -38,105 +38,98 @@ pub mod proto {
...
-            Some(ref mut set) => {
-                set.locked = true;
-                Ok(GetIntersectionResponse {
-                    values: set.values.iter().cloned().collect(),
-                })
+    fn handle_command(&mut self, command: oak::grpc::Invocation) -> anyhow::Result<()> {
+        match (&command.receiver, &command.sender) {
+            (Some(receiver), Some(sender)) => {
+                let label = receiver.label()?;
+                let grpc_response_writer = oak::grpc::ChannelResponseWriter::new(sender.clone());
```
conradgrobler, line 112: Should this not happen inside the if statement?

conradgrobler (approved):

examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -38,105 +38,98 @@ pub mod proto {
...
+    proto::oak::application::ConfigMap,
 };
-use std::collections::{HashMap, HashSet};
+use oak_services::proto::google::rpc;
+use proto::PrivateSetIntersectionDispatcher;
 
-#[derive(Default)]
-struct Main;
+// Base64 encoded Ed25519 public key corresponding to Wasm module signature.
+const PUBLIC_KEY_BASE64: &str = "f41SClNtR4i46v2Tuh1fQLbt/ZqRr1lENajCW92jyP4=";
```
tiziano88, line 57: Isn't this just https://github.com/project-oak/oak/blob/main/examples/keys/ed25519/test.pub ? If so, please mention it here, otherwise future maintainers will have no idea where it's coming from and what it represents.

Also, would it be possible to just use `include_bytes` or similar and refer directly to it? (probably it's not worth it though)

examples/private_set_intersection/module/rust/src/lib.rs
```diff
@@ -38,105 +38,98 @@ pub mod proto {
...
-                    values: set.values.iter().cloned().collect(),
-                })
+    fn handle_command(&mut self, command: oak::grpc::Invocation) -> anyhow::Result<()> {
+        match (&command.receiver, &command.sender) {
+            (Some(receiver), Some(sender)) => {
+                let label = receiver.label()?;
+                let grpc_response_writer = oak::grpc::ChannelResponseWriter::new(sender.clone());
+                // Check if request label is valid in the context of Private Set Intersection
+                // application.
+                if label != self.public_key_label {
```
tiziano88, line 115: I think in general this should check `flows_to` rather than equality. In particular, if someone wanted to send public data, that should be allowed. WDYT?

tiziano88 (approved):

PR #1702, opened 2020-11-06
Use protobuf for machine_learning example

tiziano88:
I think we just implemented this manually in the early days before we
had protobuf support.

### DIFF
--- a/examples/machine_learning/module/rust/src/lib.rs
+++ b/examples/machine_learning/module/rust/src/lib.rs
@@ -17,12 +17,20 @@
 // This model was inspired by
 // https://github.com/AtheMathmo/rusty-machine/blob/master/examples/naive_bayes_dogs.rs .
 
-use log::{error, info, warn};
+pub mod proto {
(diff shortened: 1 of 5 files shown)


examples/machine_learning/module/rust/src/lib.rs
```diff
@@ -181,107 +189,102 @@ struct Node {
...
+            training_matrix,
+            target_matrix,
+            test_matrix,
+            test_animals,
+        });
+        Ok(MlResponse::default())
+    }
+    fn learn(&mut self, _req: MlLearn) -> grpc::Result<MlResponse> {
+        info!("Training model");
+        //self.model = NaiveBayes::<naive_bayes::Gaussian>::new();
```
ipetr0v, line 209: Why is this commented out? Is there a TODO

ipetr0v: hm, it was commented out already

ipetr0v (approved):

PR #1707, opened 2020-11-06
Add challenge response support to the gRPC server node

rbehjati:
# Checklist

      construction.

### DIFF
--- a/examples/trusted_database/client/rust/src/main.rs
+++ b/examples/trusted_database/client/rust/src/main.rs
@@ -19,7 +19,10 @@
 use anyhow::{ensure, Context};
 use log::info;
 use oak_abi::label::Label;
-use oak_client::{create_tls_channel, interceptors::label::LabelInterceptor};
+use oak_client::{
+    create_tls_channel,
+    interceptors::{auth::AuthInterceptor, combine, label::LabelInterceptor},
+};
 use structopt::StructOpt;
 use tonic::Request;
 use trusted_database_client::proto::{
@@ -72,8 +75,13 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context("Couldn't create TLS channel")?;
     let label = Label::public_untrusted();
-    let interceptor =
+    let label_interceptor =
         LabelInterceptor::create(&label).context("Couldn't create gRPC interceptor")?;
+
+    let key_pair = oak_sign::KeyPair::generate()?;
+    let auth_interceptor = AuthInterceptor::create(key_pair);
+
+    let interceptor = combine(label_interceptor, auth_interceptor);
     let mut client = TrustedDatabaseClient::with_interceptor(channel, interceptor);
 
     let request = Request::new(ListPointsOfInterestRequest {
--- a/oak_runtime/src/node/grpc/server/mod.rs
+++ b/oak_runtime/src/node/grpc/server/mod.rs
@@ -29,7 +29,11 @@ use crate::{
 };
 use hyper::service::Service;
 use log::{debug, error, info, trace, warn};
-use oak_abi::{label::Label, proto::oak::application::GrpcServerConfiguration, OakStatus};
+use oak_abi::{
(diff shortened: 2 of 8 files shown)


rbehjati:
@tiziano88 @ipetr0v I am trying to add the support for challenge-response based authentication to the gRPC server node. I want to add some tests and examples. What is the easiest way to add a gRPC client that includes a signed challenge in the requests it sends. Should I add a new example or is there an example that could be meaningfully updated? 

@daviddrysdale Is it possible to have a test with such a client in the abitest, without adding a proto definition for a new gRPC service?

tiziano88:
> @tiziano88 @ipetr0v I am trying to add the support for challenge-response based authentication to the gRPC server node. I want to add some tests and examples. What is the easiest way to add a gRPC client that includes a signed challenge in the requests it sends. Should I add a new example or is there an example that could be meaningfully updated?

I am not sure adding a new example to check only this behaviour is worth it, but I think all of the existing examples that rely on gRPC (e.g. the trusted database) must be changed to use the correct per-key label and corresponding authentication, right? Ideally abstracted in a Rust client API.

rbehjati:
@tiziano88  I agree. I was thinking of adding something to the Interceptor: 
https://github.com/project-oak/oak/blob/f99a22928deda49522738f226bb30312cb886c1d/oak_client/src/lib.rs#L43-L49

Does that make sense? 

I think updating all the examples at this stage would make the PR too big, especially since I guess we want identity to be an optional part of the request (so the examples should continue to work without it). I think for this PR it is best to update only one example to make sure that the functionality works correctly. Do you have a suggestion for which example to use for that?

ipetr0v:
Do we need to authentify clients in all examples?

> I was thinking of adding something to the Interceptor:

Agree with modifying the `Interceptor`

I think right now we only need to authentify clients in the Trusted Database example, because it creates `Handler` nodes per each client (which process its requests). And right now it's just public:
https://github.com/project-oak/oak/blob/b5c644940a3bb7a9d861865ca0b498659e950478/examples/trusted_database/module/rust/src/lib.rs#L81-L87

rbehjati:
> Do we need to authentify clients in all examples?

I don't think so. At least not at this stage. 

Thanks for the suggestion @ipetr0v. I'll look into the Trusted Database example then.

oak_runtime/src/node/grpc/server/mod.rs
```diff
@@ -332,6 +336,80 @@ fn get_oak_label(metadata_map: &MetadataMap) -> Result<Label, OakLabelError> {
...
+///
+/// An error is returned if more than one signed challenge is specified by the caller, or in case
+/// the provided signature cannot be verified.
+fn get_user_identity_label(metadata_map: &MetadataMap) -> Result<Label, OakIdentityError> {
+    let signature = metadata_map
+        .get_all_bin(oak_abi::OAK_SIGNED_CHALLENGE_GRPC_METADATA_KEY)
+        .iter()
+        .collect::<Vec<_>>();
+    if signature.is_empty() {
+        info!("no signed challenge provided.",);
```
ipetr0v, line 370: nit: extra `,` in the end

examples/trusted_database/client/rust/src/main.rs
```diff
@@ -72,8 +75,13 @@ async fn main() -> anyhow::Result<()> {
...
         .context("Couldn't create TLS channel")?;
     let label = Label::public_untrusted();
-    let interceptor =
+    let label_interceptor =
         LabelInterceptor::create(&label).context("Couldn't create gRPC interceptor")?;
+
+    let key_pair = oak_sign::KeyPair::generate()?;
+    let auth_interceptor = AuthInterceptor::create(key_pair);
+
+    let interceptor = combine(label_interceptor, auth_interceptor);
```
ipetr0v, line 84: Could you please add tests for the new logic into abitest? Maybe in the next commit

ipetr0v (approved):

PR #1709, opened 2020-11-06
Simplify generated code for gRPC dispatcher

tiziano88:
Introduce a separate macro to implement the dispatching logic on the
same type that implements the server trait, instead of introducing the
extra indirection through the dispatcher type, which also does not work
in case in which the dispatcher is defined in a different crate.

Ref #1691

# Checklist

      construction.


examples/abitest/module_0/rust/src/lib.rs
```diff
@@ -178,13 +178,14 @@ pub extern "C" fn frontend_oak_main(_in_handle: u64) {
...
         let node = FrontendNode::new();
-        let dispatcher = OakAbiTestServiceDispatcher::new(node);
         let grpc_channel =
             oak::grpc::server::init("[::]:8080").expect("could not create gRPC server pseudo-Node");
-        oak::run_command_loop(dispatcher, grpc_channel.iter());
+        oak::run_command_loop(node, grpc_channel.iter());
     });
 }
 
+oak::impl_dispatcher!(impl FrontendNode : OakAbiTestServiceDispatcher);
```
rbehjati, line 187: Just curious, why is this syntax chosen?

oak_utils/src/lib.rs
```diff
@@ -131,46 +131,18 @@ impl prost_build::ServiceGenerator for OakServiceGenerator {
...
+            impl #dispatcher_name {
+                pub fn server_method<T: #service_name>(method_name: &str) -> Option<&dyn #oak_package::grpc::Invocable<T>> {
                     match method_name {
                         #(#dispatcher_invocable_methods,)*
                         _ => None
                     }
                 }
             }
 
-            impl <T: #service_name + Default> Default for #dispatcher_name<T> {
```
rbehjati, line 150: I don't know why this one and `WithInit` were needed before, but should not they be implemented for `CommandHandler` or `ServerNode` now?

rbehjati (approved):
LGTM.

examples/abitest/module_0/rust/src/lib.rs
tiziano88: Kind of randomly, I wanted something that reads naturally, but I agree it's not particularly clear right now. Anyways we can (and should) easily change it later on, it should be easy to search & replace, so please feel free to suggest better alternatives.

oak_utils/src/lib.rs
tiziano88: No, the beauty of the new approach is that there are no wrappers needed, so it's enough to just implement them on the actual struct that implements the server trait, we don't need to forward them.

examples/abitest/module_0/rust/src/lib.rs
rbehjati: I am actually fine with this syntax. I guess the documentation on the macro is good enough for now.

PR #1723, opened 2020-11-10
Update prometheus version

rbehjati:
Fixes #1450 and #1419.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -276,6 +276,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
 
 [[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
 name = "chat"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


rbehjati:
The second commit updates prometheus to version 0.10.0. Not sure if there is a better way to do this, or perhaps we want to ignore version 0.10.0 for now, as discussed under #1419?

tiziano88:
What are the ignores for? Different versions of the same dep? Or something else?

rbehjati:
> What are the ignores for? Different versions of the same dep? Or something else?

Yes. Different versions of the same dep.

tiziano88:
> > What are the ignores for? Different versions of the same dep? Or something else?
> 
> Yes. Different versions of the same dep.

If we end up just ignoring them anyways, should we just relax the restrictions in the first place?

rbehjati:
> If we end up just ignoring them anyways, should we just relax the restrictions in the first place?

I have been wondering about the same thing for a while :) I think `cargo-deny` will inform us when an ignored version is no longer present in the Cargo.lock... but I think when that happens the only thing we have to do is to update `deny.toml`. So, I really don't see the benefit of this check.

conradgrobler:
> > > What are the ignores for? Different versions of the same dep? Or something else?
> > 
> > 
> > Yes. Different versions of the same dep.
> 
> If we end up just ignoring them anyways, should we just relax the restrictions in the first place?

I guess for now we don't really care about multiple versions of the same crate as long as everything builds, so we can just set the config for `multiple-versions` to `allow`. In future, when we expect external developers to use the SDK we might care more about this.

conradgrobler (approved):

tiziano88:
@rbehjati I think the point of that check was so that we would have to review a smaller set of dependencies, but I don't think we are doing much reviewing at the moment. 

@conradgrobler I don't think this is particularly relevant to the SDK, it's mostly about the runtime and oak_loader, right?

conradgrobler:
> @conradgrobler I don't think this is particularly relevant to the SDK, it's mostly about the runtime and oak_loader, right?

I think the SDK matters if many people use it because they would have to depend on everything that the SDK depends on. The effect could be multiplied by the number of people using it.

tiziano88:
but the SDK is not in the TCB, I don't expect we will have to put the same amount of effort in reviewing its dependencies, right?

conradgrobler:
> but the SDK is not in the TCB, I don't expect we will have to put the same amount of effort in reviewing its dependencies, right?

We won't, but many of the developers would hopefully review their own code and be concerned about vulnerabilities etc in the dependencies.

PR #1728, opened 2020-11-11
Add HTTP client pseudo-node

rbehjati:
The change list includes
- Some refactoring in `oak_runtime::node`
  - moved the `HttpServerNode` from `http` to the `http::server` submodule
  - added a new submodule `http::util` that contains the struct `Pipe`. It is used both in the server and the tests
  - replaced `grpc::invocation` with a generic `invocation` submodule that is used both in the `GrpcClientNode` and the `HttpClientNode`.
- An HTTP client pseudo-node similar to the gRPC client pseudo-node
- Adds a test of the new HTTP client pseudo-node
- Includes a few minor irrelevant fixes

# Checklist

      construction.


oak_runtime/src/node/http/util.rs
```diff
@@ -0,0 +1,151 @@
+//
```
rbehjati, line 1: Moved common structs from `server.rs` to this submodule. These structs are also used in the tests for the `HttpClientNode`.

oak_runtime/src/node/invocation.rs
```diff
@@ -0,0 +1,142 @@
+//
```
rbehjati, line 1: Almost entirely copied from `oak_runtime/src/node/grpc/invocation.rs`. It is used both in `HttpClientNode` and `GrpcClientNode`.

rbehjati (commented):
It may be easier to review this change commit by commit.

oak_loader/src/options.rs
```diff
@@ -326,6 +332,23 @@ fn get_root_tls_certificate_or_default(opt: &Opt) -> anyhow::Result<String> {
...
+        Some(certificate_path) => {
+            if !cfg!(feature = "oak_debug") {
+                // Unreachable: it should not be possible to specify the flag without `oak_debug`.
+                anyhow::bail!(
+                    "Specifying `root-tls-certificate` requires the `oak_debug` feature."
+                );
+            };
+            Ok(certificate_path.to_owned())
+        }
+        None => Ok("certs/roots.pem".to_string()),
```
tiziano88, line 348: These root certs don't seem to end up being statically compiled in the executable binary, what stops someone from just replacing this file and pointing to an arbitrary self-signed certs at runtime? Is this also what we are doing for the gRPC client pseudo-node? (if so, I think that must be fixed there too)

cc @conradgrobler

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,280 @@
...
+            invocation,
+        }
+    }
+
+    async fn handle(&self) -> Result<(), OakError> {
+        let response = HttpResponse {
+            body: vec![],
+            status: 400,
+            headers: None,
+        };
```
tiziano88, line 232: What does this response represent? Why are all its fields hardcoded? I guess this is not the real method that sends responses back over the invocation channel, or am I missing something?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,280 @@
...
+                "Couldn't read HTTP request from the invocation: {:?}",
+                error
+            );
+            error
+        })?;
+        debug!("Incoming HTTP request: {:?}", request);
+
+        let mut hyper_request_builder = hyper::Request::builder()
+            .method(&request.method[..])
+            .uri(request.uri);
```
tiziano88, line 174: I am not sure I understand why there is a request URI here, but there is also one that gets specified at node creation time, and how they are related to each other, and to the node privilege. Could you point me to the relevant code?

tiziano88 (commented):
Thanks Razieh, a couple high level comments

oak_runtime/src/node/http/client.rs
rbehjati: Oops! Thanks for catching it! Fixed. I'll add more checks to the test too.

oak_runtime/src/node/http/client.rs
rbehjati: This is something that I needed to discuss :) 
I am not sure what is the best way to send the requests to the HTTP client. I think to follow the same pattern as in the gRPC client, we could have the base URI in the HTTP client, and each request would only contain a relative URI. 
The other alternative is to let the incoming HTTP requests to have the absolute URI, but check those URIs against the base URI in the HTTP client. I think generally we want this kind of checks to be done more strictly, or via the IFC. So, perhaps the first approach is the better one. The only problem I see with that is that it might be confusing that sometimes we use the `uri` in the `HttpRequest` to be the absolute URI, but in this case it has to be the relative URI. WDYT?

If we can go for the second alternative, we could perhaps start the HTTP client with an allow-list of authorities (instead of having a dedicated client per authority as we have for gRPC). Incoming requests come with absolute URIs, but are only handled if the authority matches one of the authorities in the allow-list. With this approach, we'd probably need disjunctions too.

oak_loader/src/options.rs
```diff
@@ -302,7 +302,13 @@ fn create_http_config(opt: &Opt) -> anyhow::Result<oak_runtime::HttpConfiguratio
     };
 
     match oak_runtime::tls::TlsConfig::new(http_tls_certificate_path, http_tls_private_key_path) {
-        Some(tls_config) => Ok(oak_runtime::HttpConfiguration { tls_config }),
+        Some(tls_config) => {
+            let root_tls_certificate = get_root_tls_certificate_path_or_default(&opt).ok();
+            Ok(oak_runtime::HttpConfiguration {
+                tls_config,
+                http_client_root_tls_certificate: root_tls_certificate,
```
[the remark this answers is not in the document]
rbehjati: This is no longer applicable.

rbehjati (commented):
Thanks for the comments @tiziano88.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,291 @@
...
+
+        // Forward the request to the external HTTP service and wait for the response.
+        let resp = self
+            .http_client
+            .request(hyper_request)
+            .await
+            .map_err(|error| {
+                error!("Error while awaiting response: {}", error);
+                // TO-DO: Retry instead?
+                send_error(
```
conradgrobler, line 193: nit: this seems quite inconsistent. If the body parsing fails above, it won't send an error back, and only logs a warning. Here it logs an error and sends an error response back (which logs another error). At the point where the response is handled, another warning is logged. What is the criteria for logging errors vs warning and sending an error response back vs not?

oak_runtime/src/node/http/client.rs
conradgrobler: For consistency with the gRPC client, I would favour only accepting relative URIs. The gRPC method is effectively a relative URI.

conradgrobler (approved):

oak_runtime/src/node/http/client.rs
tiziano88: Accepting relative URIs in the request would probably not be very ergonomic for callers (also error prone). I suggest allowing the caller to specify the full URI, but then the node would check whether the schema is HTTPS, and whether the host matches the one defined in the node configuration.

Thinking more long-term, I actually think that we should get rid of the hardcoded host in the configuration eventually. An HTTP client pseudo-node may be created with an arbitrary label (and no config), and at creation time we would chech whether the label is either public or a TLS cert one (for now let's ignore connectives).

If the HTTP client node is created with a public label, then it will allow arbitrary requests to any URL for any host. Of course only nodes that are public or can declassify to public will be able to use it, but I think that's probably an easy win that covers a lot of nodes that just fetch some external resource at startup.

If the HTTP client node is created with a TLS cert label, then at runtime it will check whether the host of the incoming request matches the label (and privilege), and only proceed if the node is able to effectively declassify the request before sending it out.

conradgrobler: If the HTTP client node automatically gets the privilege to declassify based on its assigned label, it is equivalent to just adding a TLS endpoint Top tag and always giving that privilege to the HTTP client node. I think the current approach of restricting the authority is safer. If a single node needs to connect to multiple different servers, we could perhaps change the HTTP client node configuration to have a list of authorities (rather than just a single one).

conradgrobler (approved):

oak_runtime/src/node/http/client.rs
rbehjati: That is an interesting observation :) I prefer a node with  theTop privilege and an allowlist of authorities. It gives more flexibility.

rbehjati (commented):
I have reverted some of the earlier changes that added two variants of the HttpClientNode to allow both HTTP and HTTPS. It turned out the HTTPS client can be used to serve both HTTP and HTTPS requests.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+            invocation.send_error(code, msg, runtime);
+        };
+
+        // Receive a request from the invocation channel.
+        let request = match invocation.receive_request(&runtime) {
+            Ok(request) => request,
+            Err(err) => {
+                send_error(
+                    format!("{:?}", err),
+                    http::StatusCode::INTERNAL_SERVER_ERROR,
```
tiziano88, line 176: Not sure I understand where these errors end up, does it mean that a caller would get these errors as if they are coming from the remote server, even though they are actually coming from this pseudo-node implementation? (I don't think that needs to change in this PR anyways, but if that's the case we may want to think how to better propagate real HTTP errors from internal node errors at some point perhaps).

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+
+/// Struct that represents an HTTP client pseudo-Node.
+pub struct HttpClientNode {
+    /// Pseudo-Node name.
+    node_name: String,
+    /// HTTP client to allow re-use of connection across multiple method invocations. It can be
+    /// used to send either HTTP or HTTPS requests.
+    http_client: HyperClient,
+    /// The authority corresponding to this client, or none in case of a public client.
+    authority: Option<String>,
```
tiziano88, line 48: Does this also mean that this is None if the scheme was HTTP, or the value of the authority in the protobuf if it was HTTPS? That seems somewhat inconsistent, for instance below I would expect to see a check on both scheme and authority, but only authority is checked, which I guess is because the scheme has been already taken into account?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// must also match the label with which this node is being created. If the node is configured
+    /// to handle URIs with the `HTTP` scheme, then it should be labelled public. In this case,
+    /// the authority could be empty, meaning that the node can handle all HTTP requests. In
+    /// addition, a non-empty `root_ca` is always required.
+    pub fn new(node_name: &str, config: HttpClientConfiguration, root_ca: Vec<u8>) -> Option<Self> {
+        if !validate_config(&config, root_ca.is_empty()) {
+            return None;
+        }
+        let http_client = create_client(root_ca).ok()?;
+        let node_privilege = get_privilege(config.authority.clone());
```
tiziano88, line 123: What happens to `config.scheme` here? is it only used for checking? If so, why not just check the authority for non-emptiness, which is also what the rest of the code here assumes anyways?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /// Process an HTTP invocation for an external HTTP service.
+    async fn process_invocation(&mut self, runtime: &RuntimeProxy, invocation: &Invocation) {
+        // TODO(#1717): Update HTTP client metrics
+
+        // A closure to ease sending and logging the errors
+        let send_error = |err, code, msg| {
+            warn!("{}: {}", msg, err);
+            invocation.send_error(code, msg, runtime);
```
tiziano88, line 167: Would it make sense to make this method return `Result` or something, and have the logging and sending the error response one level up? It would also allow more idiomatic error handling in this code, instead of the various early returns.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+                    "Failed to convert the request body",
+                );
+                return;
+            }
+        };
+
+        // Check that the `host` in the request URI matches the authority supported by the node, if
+        // any.
+        if let Some(authority) = self.authority.clone() {
+            if let Some(req_authority) = hyper_request.uri().authority() {
```
tiziano88, line 208: This seems to allow requests with no authority set, regardless of the authority of the node. Is that expected? What happens if such request is sent? I guess it fails inside hyper? If so, maybe it would be simpler to make this check more exhaustive at this stage and catch it early?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+            invocation.close(&runtime);
+        }
+    }
+
+    /// Process an HTTP invocation for an external HTTP service.
+    async fn process_invocation(&mut self, runtime: &RuntimeProxy, invocation: &Invocation) {
+        // TODO(#1717): Update HTTP client metrics
+
+        // A closure to ease sending and logging the errors
+        let send_error = |err, code, msg| {
```
tiziano88, line 165: Could you add type annotations to some of these args? In particular, I thought err would be an error, but actually it seems it is a string when this is actually called?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+                    http::StatusCode::INTERNAL_SERVER_ERROR,
+                    "Failed to convert the request body",
+                );
+                return;
+            }
+        };
+
+        // Check that the `host` in the request URI matches the authority supported by the node, if
+        // any.
+        if let Some(authority) = self.authority.clone() {
```
tiziano88, line 207: Conceptually, it feels like this check should happen as early as possible (as for most checks), especially since it's only based on the URI of the request. Everything else should happen only if this check was successful. For now it is not a big difference, but I am worried in the future the logic may grow, and some side-effects may accidentally sneak in before the validation is performed.

tiziano88 (commented):
Thanks Razieh!

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+        ResponseHandler {
+            runtime,
+            response,
+            invocation,
+        }
+    }
+
+    async fn handle(&mut self) -> Result<(), OakError> {
+        let headers = Some(HeaderMap::from(self.response.headers().to_owned()));
+        let status = self.response.status().as_u16() as i32;
```
ipetr0v, line 261: Why do we need it a signed?

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+        let body = self.response.body_mut();
+        let body = match to_bytes(body).await {
+            Ok(bytes) => bytes.to_vec(),
+            Err(err) => {
+                warn!(
+                    "Error when converting the response body into bytes: {:?}",
+                    err
+                );
+                vec![]
```
ipetr0v, line 270: Is it ok to continue working with an empty body, if it wasn't empty and parsing returned an error?

ipetr0v (approved):

oak_runtime/src/lib.rs
```diff
@@ -130,6 +129,8 @@ pub struct SignatureTable {
 pub struct HttpConfiguration {
     /// TLS identity to use for all HTTP Server Nodes.
     pub tls_config: crate::tls::TlsConfig,
+    /// Path to the root TLS certificate to use for all HTTP Client Nodes.
```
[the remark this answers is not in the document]
rbehjati: It is actually a vector. I have updated the comment. It was wrong. But I don't think I can use `Certificate` here. `Certificate` is from `tonic`. However, here we need something that can be used with `hyper_rustls`. We could perhaps have our own `Certificate` type, consisting of only a vector (or a string), and then converting that to each specific type for `tonic` or `hyper_rustls`.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+            hashset! { oak_abi::label::tls_endpoint_tag(&authority.as_str()) },
+            hashset! { oak_abi::label::tls_endpoint_tag(&authority.as_str()) },
+        )
+    }
+}
+
+/// Checks the validity of the configuration:
+/// - A non-empty TLS root certificate is provided
+/// - If the scheme is HTTPS, a non-empty authority is specified in the config
+fn validate_config(config: &HttpClientConfiguration, empty_root_ca: bool) -> bool {
```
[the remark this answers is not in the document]
rbehjati: Yeah! Perhaps. Although the actual value of `root_ca` is not needed inside this function. I might still need to change this function based on the comment from @tiziano88 below.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+impl HttpClientNode {
+    /// Creates a new `HttpClientNode` that can either handle HTTP requests or HTTPS requests. To
+    /// handle HTTPS requests a non-empty authority in the `config` is required. This configuration
+    /// must also match the label with which this node is being created. If the node is configured
+    /// to handle URIs with the `HTTP` scheme, then it should be labelled public. In this case,
+    /// the authority could be empty, meaning that the node can handle all HTTP requests. In
+    /// addition, a non-empty `root_ca` is always required.
+    pub fn new(node_name: &str, config: HttpClientConfiguration, root_ca: Vec<u8>) -> Option<Self> {
```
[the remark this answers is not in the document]
rbehjati: I guess this could be fixed consistently for all nodes in a separate PR. If there is a consensus about this convention, I can create an issue for it :)

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// Main loop that handles HTTP invocations from the `handle`, sends HTTP requests to an
+    /// external HTTP service and writes HTTP responses back to the invocation channel.
+    async fn handle_loop(&mut self, runtime: RuntimeProxy, handle: Handle) -> Result<(), OakError> {
+        // Create a [`Receiver`] to receive the HTTP request.
+        let receiver = Receiver::<Invocation>::new(ReadHandle { handle });
+        loop {
+            debug!("Waiting for HTTP invocation");
+            // Read an HTTP invocation from the [`Receiver`].
+            let invocation = receiver.receive(&runtime).map_err(|error| {
+                match error {
```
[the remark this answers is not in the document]
rbehjati: hmm... I copied this part completely from the `GrpcClientNode`. I guess there is some wisdom about logging `channel closure` at the INFO level.

oak_runtime/src/node/http/client.rs
rbehjati: To match the type in the `prost`-generated type (`HttpResponse`).

oak_runtime/src/node/http/client.rs
rbehjati: Perhaps not. Originally, I assumed that we want to always send a response, unless sending the response itself fails. But in this case the empty body would be inconsistent with the status. Changed it to return an error instead. Thanks :)

rbehjati (commented):
Thanks for the comments. I think I have addressed all the comments now. Except for the comment on `validate_config` (now renamed to `is_config_valid`). Could you please have a look at my comment there @tiziano88? 
Also, please see my comment about the type of `http_client_root_tls_certificate` (@tiziano88 @ipetr0v).

oak_abi/proto/application.proto
```diff
@@ -121,14 +122,33 @@ message RoughtimeClientConfiguration {
...
+  // node's label should match the `authority`.
+  UriScheme scheme = 1;
+  // The authority representing an external HTTPS server that this client can send requests too. The
+  // authority should match the label of the HTTP Client pseudo-Node.
+  string authority = 2;
+}
+
+// Enum representing different the scheme that an HTTP client pseudo-Node is allowed to handle.
+enum UriScheme {
+  HTTP = 0;
```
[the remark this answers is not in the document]
rbehjati: As discussed, made `HTTPS` the default.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,327 @@
...
+/// - A non-empty TLS root certificate is provided
+/// - If the scheme is HTTPS, a non-empty authority is specified in the config
+fn validate_config(config: &HttpClientConfiguration, empty_root_ca: bool) -> bool {
+    if empty_root_ca {
+        error!("HTTP client pseudo-node: No root TLS certificate provided.");
+    }
+    if config.scheme == UriScheme::Https as i32 && config.authority.is_empty() {
+        error!("HTTP client pseudo-node: No authority specified.");
+    }
+    !empty_root_ca && (config.scheme != UriScheme::Https as i32 || !config.authority.is_empty())
```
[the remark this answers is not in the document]
rbehjati: Changed it to return `Result<(), ConfigurationError>`.

oak_runtime/src/node/http/client.rs
```diff
@@ -0,0 +1,346 @@
...
+        response: http::Response<hyper::Body>,
+        invocation: &'a Invocation,
+    ) -> Self {
+        ResponseHandler {
+            runtime,
+            response,
+            invocation,
+        }
+    }
+
```
tiziano88, line 276: I'm not sure I understand why ResponseHandler is a struct, it seems it may just be a top-level function, like `async fn handle_response(runtime, response, invocation)`? Anyways fine to keep as it is if you think it makes sense.

oak_runtime/src/node/http/tests.rs
```diff
@@ -507,3 +527,226 @@ async fn send_request(
...
+    let resp = result_value.unwrap();
+    assert_eq!(resp.status, 200);
+    runtime.runtime.stop();
+}
+
+/// Creates an HTTP client pseudo-node in the given Runtime.
+fn create_client_node(
+    runtime: &RuntimeProxy,
+    authority: String,
+    scheme: UriScheme,
```
tiziano88, line 618: It may read better if scheme came first, I was confused when reading the invocation and didn't see that HTTPS was after the domain name.

oak_runtime/src/node/http/tests.rs
```diff
@@ -507,3 +527,226 @@ async fn send_request(
...
+        .unwrap();
+
+    // Wait for the test Node to complete execution before terminating the Runtime.
+    let result_value = result_receiver.recv().expect("test node disconnected");
+
+    assert!(result_value.is_ok());
+    let resp = result_value.unwrap();
+    // The test uses a self signed certificate. So, we expect an error that is converted to a 500
+    // response in the HTTP client node.
+    assert_eq!(resp.status, 500);
```
tiziano88, line 572: https://docs.rs/http/0.2.1/http/status/struct.StatusCode.html#associatedconstant.INTERNAL_SERVER_ERROR

oak_runtime/src/node/http/tests.rs
```diff
@@ -507,3 +527,226 @@ async fn send_request(
...
+            oak_node_init_receiver,
+        )
+        .unwrap();
+
+    // Wait for the test Node to complete execution before terminating the Runtime.
+    let result_value = result_receiver.recv().expect("test node disconnected");
+
+    assert!(result_value.is_ok());
+    let resp = result_value.unwrap();
+    assert_eq!(resp.status, 200);
```
tiziano88, line 610: https://docs.rs/http/0.2.1/http/status/struct.StatusCode.html#associatedconstant.OK

oak_runtime/src/node/http/tests.rs
```diff
@@ -507,3 +527,226 @@ async fn send_request(
...
+                "Failed to close initial inbound channel {}: {:?}",
+                handle, err
+            );
+        }
+
+        // create request.
+        // NOTE: Method is case sensitive.
+        let request = HttpRequest {
+            uri: self.uri.clone(),
+            method: "GET".to_string(),
```
tiziano88, line 721: https://docs.rs/http/0.2.1/http/method/struct.Method.html#associatedconstant.GET

tiziano88 (approved):

PR #1729, opened 2020-11-13
Rewrite hello_world example using async std.

wildarch:
# Checklist

      construction.

Also includes a patch to `ChannelResponseWriter`'s behaviour on `close` and a tiny bugfix to async code generation. I can spin these off into separate PRs if preferred.

If you prefer to keep all examples sync for now, we can also keep this PR around just for testing.

### DIFF
--- a/sdk/rust/oak/src/grpc/mod.rs
+++ b/sdk/rust/oak/src/grpc/mod.rs
@@ -117,14 +117,14 @@ impl ChannelResponseWriter {
 
     /// Close out the gRPC method invocation with the given final result.
     pub fn close(&self, result: Result<()>) -> std::result::Result<(), OakError> {
-        // Build a final GrpcResponse message wrapper and serialize it into the
-        // channel.
-        let mut grpc_rsp = GrpcResponse::default();
-        grpc_rsp.last = true;
         if let Err(status) = result {
+            // Build a final GrpcResponse message wrapper and serialize it into the
+            // channel.
+            let mut grpc_rsp = GrpcResponse::default();
+            grpc_rsp.last = true;
             grpc_rsp.status = Some(status);
+            self.sender.send(&grpc_rsp)?;
         }
-        self.sender.send(&grpc_rsp)?;
         self.sender.close()?;
         Ok(())
     }
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -618,11 +618,13 @@ dependencies = [
  "assert_matches",
  "either",
  "env_logger",
+ "futures",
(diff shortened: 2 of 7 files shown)


sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -117,14 +117,14 @@ impl ChannelResponseWriter {
 
     /// Close out the gRPC method invocation with the given final result.
     pub fn close(&self, result: Result<()>) -> std::result::Result<(), OakError> {
-        // Build a final GrpcResponse message wrapper and serialize it into the
-        // channel.
-        let mut grpc_rsp = GrpcResponse::default();
-        grpc_rsp.last = true;
         if let Err(status) = result {
+            // Build a final GrpcResponse message wrapper and serialize it into the
```
wildarch, line 121: To me these changes seem reasonable, please let me know if this causes problems somewhere else (e.g. if anyone relies on `last` always being set on the final message).

tiziano88: This may be a separate change, could you split it into a separate PR, so it's not blocked on the rest of the review?

tiziano88:
Thanks for the changes @wildarch . I have been thinking more about async support lately, and while initially I was more keen that we may want to make everything async in the future, I am now that convinced any more.

Currently with the new command-loop style API, most of the read operations from incoming channels are abstracted in SDK helper functions, which I think ends up being a simpler API to implement, even though it is sync. Also we don't really do reads concurrently from a single node. In fact the common pattern that is emerging is that a new handler node is spawned for each request, and since nodes implicitly run in parallel, that ends up being the main way of achieving concurrency in oak, while still maintaining single-thread blocking individual nodes.

In this particular example, the situation is complicated by the fact that async traits are not yet stabilized, so this seems a bit of a step back from implementing an auto-generated server trait (though personally I don't mind this style in general).

WDYT?

wildarch:
You make a good point @tiziano88, in some sense the async version can be more verbose, and there are plenty of cases where async does not add a lot, at least in the examples.

That said, I think that there are plenty of good reasons to write async nodes (stream routing and gRPC client side streaming come to mind), and preventing separate sync and async worlds seems worthwhile to me. Assuming we do not want to drop async support altogether, I would consider eventually going async-only the logical outcome.

I think that with some effort we can make the async std as easy to use as the sync version. If that requires async traits, those will be stabilized at some point (see areweasyncyet.rs), and until then we could use https://github.com/dtolnay/async-trait. Personally I quite like the stream based API the async std currently has, but that is probably just author bias 😜

PR #1734, opened 2020-11-17
Stop using node privilege by default

### LINKED ISSUE #1670: Add `with_privilege` versions of node and channel function to ABI
At the moment operations such as `channel_read` will automatically try to apply the privilege of the node. This means that it is not always obvious whether privilege is being used or not. The following functions should be added to the ABI to make clear that privilege is being used:

- `channel_read`
- `channel_write`
- `node_create`
- `channel_create`

The following might also need `with_privilege` versions:

- `wait_on_channels`
- `channel_close`

Question: should the explicity privilege to use be passed to the `with_privilege` versions, or should the signature just be the same as the original, but making it explicit that privilege is being used?

conradgrobler:
This PR includes the following changes:

- Removed automatic usage of node privilege when creating nodes and channels, waiting on, reading from or writing to channels
- Removed the need for declassifying reads by making the HTTP and gRPC server pseudo_nodes "top secret"
- Only added explicit `with_downgrade` versions of functions to the `RuntimeProxy` where it is currently required
  - `node_create`
  - `channel_create`
  - `channel_write`
- Added associated `with_downgrade` functions to `oak_runtime`'s `SenderExt`
- Only added explicit `with_downgrade` version of function to the ABI where it is currently needed:
  - `node_create`
  - `channel_create`
  - `channel_write`
- Only added explicity `with_downgrade` versions of functions to the SDK where it is currently needed
  - `node_create`
  - `channel_create`
  - `channel_write`
- Added associated `with_downgrade` version to the SDK's `SenderExt`
- Updated documentation to reflect ABI and SDK changes

Fixes #1670 

# Checklist

      construction.


docs/abi.md
```diff
@@ -179,6 +179,11 @@ If writing to the specified channel would violate
 - `param[4]: u32`: Source handle array count
 - `result[0]: u32`: Status of operation
 
+### `channel_write_with_privilege`
+
+The same as [`channel_write`], except that it uses the current node's
```
tiziano88, line 184: I don't think this works as a link, does it?

oak_abi/src/lib.rs
```diff
@@ -139,6 +139,22 @@ extern "C" {
...
     ) -> u32;
 
+    /// Write a message to a channel using the current node's privilege.
+    ///
+    /// Write `size` bytes of data from `buf`, together with `handle_count` handles from
+    /// the space at `handle_buf`, to the channel identified by `handle`.
+    ///
+    /// Returns the status of the operation, as an [`OakStatus`] value.
+    ///
+    /// [`OakStatus`]: crate::OakStatus
```
tiziano88, line 149: May be simpler to maintain by just linking to `channel_write` in the comment and explaining the difference.

oak_runtime/src/io.rs
```diff
@@ -84,26 +81,32 @@ pub trait SenderExt<T> {
     /// Encodes and sends a message to the [`Sender::handle`].
     fn send(&self, message: T, runtime: &RuntimeProxy) -> Result<(), OakError>;
 
+    /// Encodes and sends a message to the [`Sender::handle`] using the current Node's privilege.
+    fn send_with_privilege(&self, message: T, runtime: &RuntimeProxy) -> Result<(), OakError>;
```
tiziano88, line 85: Alternatively perhaps `send_with_downgrade` or something like that? (also at the ABI level)

Especially since we are planning to at some point remove the explicit privilege and use integrity as declassification privilege effectively, so I think downgrade is more accurate.

@aferr do you have a preference either way?

oak_runtime/src/lib.rs
```diff
@@ -340,6 +340,13 @@ pub enum LabelReadStatus {
     NeedsCapacity(usize),
 }
 
+/// Indicator whether an operation is executed using the Node's privilege or without it.
+#[derive(Clone, Copy, Debug)]
+enum OperationPrivilege {
+    None,
+    UsePrivilege,
+}
```
tiziano88, line 348: Similarly to my previous comment, maybe just `Downgrading { No, Yes }` or similar?

oak_runtime/src/lib.rs
```diff
@@ -1254,6 +1299,7 @@ impl Runtime {
         node_name: &str,
         label: &Label,
         initial_handle: oak_abi::Handle,
+        use_privilege: OperationPrivilege,
```
tiziano88, line 1302: Not sure I follow what this means in this context, could you expand the comment?

oak_runtime/src/node/http/tests.rs
```diff
@@ -377,13 +377,12 @@ fn create_server_node(runtime: &RuntimeProxy, port: u32) -> oak_abi::Handle {
         })),
     };
 
+    // TODO(#1631): When we have a separate top for each sub-lattice, this should be changed to
+    // the top of the `user` sub-lattice.
```
tiziano88, line 381:
```suggestion
    // the top of the "public key identity" sub-lattice.
```

I think. See

https://github.com/project-oak/oak/blob/9fd16d81c0594264eb9bc0939b3a0a0aff80448d/oak_abi/proto/label.proto#L88-L94

sdk/rust/oak/src/lib.rs
```diff
@@ -294,6 +345,42 @@ pub fn node_create(
     result_from_status(status as i32, ())
 }
 
+/// Uses the calling node's label-downgrading privilege to creates a new Node running the
+/// configuration identified by `config_name`, running the entrypoint identified by
+/// `entrypoint_name` (for a Web Assembly Node; this parameter is ignored when creating a
+/// pseudo-Node), with the provided `label`, and passing it the given handle.
```
tiziano88, line 351: The wording seems quite different from that of the channel_create_with_privilege, could you make them a bit more similar? Also I think it may be simpler to maintain here too if the _with_privilege functions just link to the regular functions, and explain only the delta in functionality.

oak_runtime/src/proxy.rs
```diff
@@ -283,28 +340,74 @@ impl RuntimeProxy {
...
+            .channel_read(self.node_id, read_handle, OperationPrivilege::None);
         debug!(
             "{:?}: channel_read({}) -> {:?}",
             self.node_name, read_handle, result
         );
         result
     }
 
+    /// Calls [`Runtime::channel_read`] using the Node's privilege.
+    pub fn channel_read_with_privilege(
```
daviddrysdale, line 393: Wasn't there some talk of only allowing declassifying-writes, not declassifying-reads? @tiziano88 do you recall whether there was a conclusion?

oak_runtime/src/lib.rs
```diff
@@ -733,25 +741,34 @@ impl Runtime {
...
+
+                trace!(
+                    "{:?}: downgraded source label: {:?}?",
+                    node_name,
+                    downgraded_source_label
+                );
+                downgraded_source_label
+            }
+            OperationPrivilege::None => source_label.to_owned(),
+        };
```
daviddrysdale, line 769: Could this chunk be encapsulated into a common `get_effective_label()` helper method?  It looks like similar code is repeated a few times.

oak_runtime/src/node/wasm/mod.rs
```diff
@@ -634,6 +667,86 @@ impl WasmInterface {
...
+                    "{}: fetch_message(): Unable to read handles from guest memory: {:?}",
+                    self.pretty_name, err
+                );
+                OakStatus::ErrInvalidArgs
+            })?;
+
+        let handles: Vec<u64> = raw_handles
+            .chunks(8)
+            .map(|bytes| LittleEndian::read_u64(bytes))
+            .collect();
```
daviddrysdale, line 746: Given how neatly you've encapsulated things, maybe put this chunk into a `fetch_handles()` helper too?

daviddrysdale (approved):

docs/abi.md
conradgrobler: Updated to include anchor link.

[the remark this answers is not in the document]
conradgrobler: Done.

oak_abi/src/lib.rs
```diff
@@ -139,6 +139,22 @@ extern "C" {
         handle_count: u32,
     ) -> u32;
 
+    /// Write a message to a channel using the current node's privilege.
```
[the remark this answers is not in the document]
conradgrobler: Done.

oak_abi/src/lib.rs
conradgrobler: Done.

oak_runtime/src/io.rs
conradgrobler: Changed to use `with_downgrade` throughout

oak_runtime/src/lib.rs
conradgrobler: Done.

oak_runtime/src/lib.rs
conradgrobler: Done.

oak_runtime/src/lib.rs
conradgrobler: Added a comment.

oak_runtime/src/node/http/tests.rs
conradgrobler: Changed to "identity sub-lattice", seeing that we might support other identities in future.

oak_runtime/src/node/wasm/mod.rs
conradgrobler: Done.

oak_runtime/src/proxy.rs
conradgrobler: There was a discussion in #1301 and the outcome was that it was not required for the security properties we want and might be needed in future. It is not available in the ABI and SDK at the moment, with the idea that we can add it later if we need it. I just added this to the proxy to support two existing tests in the runtime.

sdk/rust/oak/src/lib.rs
conradgrobler: Done.

oak_runtime/src/lib.rs
```diff
@@ -629,6 +637,20 @@ impl Runtime {
         node_privilege.downgrade_label(initial_label)
     }
 
+    /// Returns the effective label for `initial_label` in the context of the Node, taking into
+    /// acount whether the downgrade privilege should be applied.
+    fn get_node_effective_label(
```
tiziano88, line 642: Nice! Though IIUC the label is not always a node label, right?

sdk/rust/oak/src/logger/mod.rs
```diff
@@ -44,7 +44,7 @@ impl Log for OakChannelLogger {
             level: map_level(record.level()) as i32,
             message: format!("{}", record.args()),
         };
-        match self.log_sender.send(&log_msg) {
+        match self.log_sender.send_with_downgrade(&log_msg) {
```
tiziano88, line 47: This does not feel right, why is every log statement performing a downgrade? This means that if the node has sufficient privilege to declassify to public, it will be able to actually log, which is probably not the intention.

To clarify: I expect that non-public nodes will be able to log in debug (where the log node is top secret) but not in production (where the log node is public).

tiziano88 (approved):
Just one change to make. Logging must never downgrade behind the scenes, downgrading is a very powerful operation and must always be very explicitly requested by the developer.

sdk/rust/oak/src/logger/mod.rs
conradgrobler: Removed the downgrade privilege and stopped panicking when the send is blocked by IFC.

oak_runtime/src/lib.rs
conradgrobler: The "node" part was to indicate that it is in the context of the node's privilege (when used), but agree that it is confusing. Renamed.

sdk/rust/oak/src/grpc/server.rs
```diff
@@ -52,18 +52,18 @@ pub fn init_with_sender(
...
-        &config,
-    ) {
-        Ok(s) => s,
-        Err(e) => {
-            let _ = invocation_sender.close();
-            return Err(e);
-        }
-    };
+    let init_sender =
+        match crate::io::node_create::<GrpcInvocationSender>("grpc_server", &top_label, &config) {
```
rbehjati, line 60: Just to clarify... previously the gRPC and the HTTP server nodes where automatically doing declassifying reads using their privileges. Is it right? 
With this change, do the `GrpcServerNode` and `HttpServerNode` still need to have "top" as their privilege?

conradgrobler: Yes, they previously did declassifying reads which they no longer need to do because they are now top secret. But now they do declassifying writes to send the invocations from a top secret node to less secret nodes, so they still need the privilege.

rbehjati: That is right. Thanks.

PR #1742, opened 2020-11-18
Add forward_invocation function

ipetr0v:
This change adds `forward_invocation` function that checks invocation receiver label before forwarding it and is used in `Router` Nodes.

### DIFF
--- a/sdk/rust/oak/src/io/mod.rs
+++ b/sdk/rust/oak/src/io/mod.rs
@@ -17,7 +17,9 @@
 //! Wrappers for Oak SDK types to allow their use with [`std::io`].
 
 use crate::OakStatus;
+use anyhow::Context;
 use oak_abi::label::Label;
+use oak_services::proto::google::rpc;
 use std::io;
 
 mod receiver;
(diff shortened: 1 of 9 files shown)


ipetr0v:
In this commit I didn't use `forward_invocation` in Chat application - since it will be an exception (almost all other applications will just rely on `forward_invocation`, while Chat and maybe Trusted Database require specific types of labels).

sdk/rust/oak/src/io/mod.rs
```diff
@@ -87,6 +89,47 @@ pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
...
+) -> anyhow::Result<()> {
+    match (&invocation.receiver, &invocation.sender) {
+        (Some(request_receiver), Some(response_sender)) => {
+            let request_label = request_receiver
+                .label()
+                .context("Couldn't read request label")?;
+            let sender_label = crate::channel_label_read(invocation_sender.handle.handle)
+                .context("Couldn't read invocation sender label")?;
+            // Check if request label is valid in the context of invocation sender.
+            if !request_label.flows_to(&sender_label) {
```
tiziano88, line 110: This may read better as a if / else, and avoid the early return with `bail`.

sdk/rust/oak/src/io/mod.rs
```diff
@@ -87,6 +89,47 @@ pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
...
+                anyhow::bail!("Invalid request label: {:?}", request_label);
+            }
+
+            // Forward invocation through invocation sender.
+            invocation_sender
+                .send(&invocation)
+                .context("Couldn't forward invocation")
+        }
+        _ => {
+            anyhow::bail!("Received malformed invocation");
```
tiziano88, line 128: Maybe just return `Err(anyhow::anyhow!("..."))`, the early return is not necessary here.

sdk/rust/oak/src/io/mod.rs
```diff
@@ -87,6 +89,47 @@ pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
...
+pub fn forward_invocation(
+    invocation: crate::grpc::Invocation,
+    invocation_sender: &Sender<crate::grpc::Invocation>,
+) -> anyhow::Result<()> {
+    match (&invocation.receiver, &invocation.sender) {
+        (Some(request_receiver), Some(response_sender)) => {
+            let request_label = request_receiver
+                .label()
+                .context("Couldn't read request label")?;
+            let sender_label = crate::channel_label_read(invocation_sender.handle.handle)
```
tiziano88, line 107: Why does `invocation_sender.label()` not work here?

sdk/rust/oak/src/io/mod.rs
```diff
@@ -87,6 +89,47 @@ pub fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
...
+            if !request_label.flows_to(&sender_label) {
+                let grpc_response_writer =
+                    crate::grpc::ChannelResponseWriter::new(response_sender.clone());
+                grpc_response_writer
+                    .close(Err(crate::grpc::build_status(
+                        rpc::Code::InvalidArgument,
+                        "Invalid request label",
+                    )))
+                    .context("Couldn't send error response back")?;
+                anyhow::bail!("Invalid request label: {:?}", request_label);
```
tiziano88, line 119: No need to `bail`, just return `Err(anyhow::anyhow!(...))`.

tiziano88 (approved):
Nice!

sdk/rust/oak/src/io/mod.rs
ipetr0v: Done

sdk/rust/oak/src/io/mod.rs
ipetr0v: Done

sdk/rust/oak/src/io/mod.rs
ipetr0v: Done

PR #1743, opened 2020-11-18
Delete machine_learning and running_average examples

tiziano88:
They do not show anything interesting any more, so they end up just
being lazily updated and increase CI latency.


ipetr0v:
PR race condition with https://github.com/project-oak/oak/pull/1745
Why don't we need `running_average`? Is it because it is too similar to `private_set_intersection`?

tiziano88:
> PR race condition with #1745

Will rebase

> Why don't we need `running_average`? Is it because it is too similar to `private_set_intersection`?

Yes it's a less interesting version of that basically, right?

tiziano88:
PTAL now, thanks

ipetr0v (approved):

Issue #1722, opened 2020-11-10
Partially Secret Labels


aferr:
Roughly, move to ABI Design 2 here:
https://docs.google.com/document/d/1pd1NfpROHjxmtRZaA9pfAC3zYhhEZ_xRfDU2gyptMg0/edit#

aferr:
@conradgrobler @daviddrysdale @tiziano88 

Just for my own agenda setting, does making this part of v1 mean:
- "We definitely want to use an ABI design closer to 2, but we won't implement that until v1"
or
- "We think we might want to use ABI design 2, and we'lll re-evaluate that when we're working on v1"

On my end, it would be much easier to stick with one set of definitions than trying to keep both alive in the code at once since these definitions are pretty close to the root of the tree of definitions. I'd be inclined to stick with the one you'll wind up with eventually rather than what you have now.

conradgrobler:
My understanding is that it was more the second option (ie "We think we might want to use ABI design 2, and we'lll re-evaluate that when we're working on v1")

aferr:
Got it, thanks.

PR #1744, opened 2020-11-19
Split low projection definitions + model

aferr:
The two different definitions of low projection and two different models
discussed in (issue #1722)[https://github.com/project-oak/oak/issues/1722]
are both alive in the code up to some smaller theorems.

The code redundancy is not ideal at the moment, so if these are really
meant to stay alive forever, then that should be improved at some point.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/NIUtilTheorems.v
+++ b/experimental/ni_coq/vfiles/NIUtilTheorems.v
@@ -80,13 +80,6 @@ Definition node_low_proj_idempotent := @labeled_low_proj_idempotent node.
 Definition chan_low_proj_idempotent := @labeled_low_proj_idempotent channel.
 Definition event_low_proj_idempotent := @labeled_low_proj_idempotent event.
 
-(*
-Theorem state_low_proj_idempotent: forall ell, idempotent (state_low_proj ell).
-Proof.
-    (* this one is not used and looks like it needs FE to me *)
-Admitted.
-*)
-
 (* 
 Note that this theorem is not true for the definition where labels are
 partially secret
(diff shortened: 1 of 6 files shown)


experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -86,74 +85,39 @@ Proof.
...
+        rewrite H4 in H2.
+        contradiction.
     - (* CreateNode *)
-        admit.
+        (* there are no unobservable node creations, these only happen
+        from public*)
+        pose proof (bot_is_bot ell).
+        rewrite H4 in H2.
+        contradiction.
+    (* I get this situation where:
```
aferr, line 114: @jadephilipoom 

If you happen to have time to see what's going on here, I'd appreciate it.

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -51,29 +59,48 @@ Qed.
...
 Qed.
   
 Definition node_low_proj_idempotent := @labeled_low_proj_idempotent node.
 Definition chan_low_proj_idempotent := @labeled_low_proj_idempotent channel.
 Definition event_low_proj_idempotent := @labeled_low_proj_idempotent event.
 
+(*
 Theorem state_low_proj_idempotent: forall ell, idempotent (state_low_proj ell).
 Proof.
+    (* this one is not used and looks like it needs FE to me *)
```
aferr, line 86: Note to self to delete this because it turned out this theorem is just not used anywhere.

experimental/ni_coq/vfiles/NIUtilTheorems.v
```diff
@@ -386,5 +404,6 @@ Proof.
         (chans.[? h0]).
         reflexivity. symmetry. apply upd_neq; auto. 
 Qed.
+*)
```
jadephilipoom, line 407: Delete this commented-out theorem, if it's unused?

experimental/ni_coq/vfiles/PossibilisticNI.v
```diff
@@ -37,12 +37,11 @@ Proof.
          | _ => destruct_match
          | _ => tauto
          end.
-    intros.
     (*
     Note: I think this is not true with the low-projection def
     where labels are partially secret
     *)
```
jadephilipoom, line 43: Is this note still accurate? The `Admitted` is replaced with `Qed`, so presumably the lemma is correct?

experimental/ni_coq/vfiles/PossibilisticNI.v
jadephilipoom: Do the extra goals show up if you write `Unshelve`?

jadephilipoom (approved):

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: It is still accurate, but thanks for double-checking. PossibilisticNI is using the definition of low projection that is label-preserving rather than the one where the labels of secret objects get mapped into top. For the label-preserving definition this theorem is still correct. The other def is kept partially alive along with some smaller theorems about it since we might change to this other def later.

experimental/ni_coq/vfiles/PossibilisticNI.v
aferr: Thanks :) that fixed it.

PR #1745, opened 2020-11-19
Use labels in Running Average

ipetr0v:
This change updates Running Average example to use Router pattern and signature labels.


ipetr0v:
Closed because this example was deleted in https://github.com/project-oak/oak/pull/1743

PR #1746, opened 2020-11-19
Refactor nodes & channels to use debug_id

juliettepretot:
Previously both the `Node` and `Channel` struct stored a `name` property. It is supplied via the ABI as non unique string. When setting it in the struct however it is combined with  the `id` to form a unique debug_id (eg `grpc_node(12)`, `invocation_channel(1)`).

As pointed out by @tiziano88 in https://github.com/project-oak/oak/pull/1698#discussion_r525347971 it can make the code a bit hard to read, since it isn't quite clear when the name refers to the non-unique name, and when it referred to the uniquified version.

This fixes this by storing the `name` as as it was supplied via the Oak ABI. A method `get_debug_id` then is used to combine the name with the id, forming a the debug_id used in logs. 

Ref #1615

# Checklist

      construction.


oak_runtime/src/channel.rs
```diff
@@ -76,14 +76,22 @@ pub struct Channel {
...
+    name: String,
 
     /// Weak reference to the Runtime used for sending introspection events.
     runtime_weak: Weak<Runtime>,
 }
 
+impl Channel {
+    /// Returns a unique debug_id used to identify the node in the debug output,
+    /// consisting out of the provided [`ChannelId`], and the node's name.
+    fn get_debug_id(&self, channel_id: ChannelId) -> String {
```
tiziano88, line 90: I'm confused, why is this both a method on Channel, but also expects a ChannelId to be passed in?

oak_runtime/src/lib.rs
```diff
@@ -1301,28 +1319,21 @@ impl Runtime {
...
         let initial_handle = new_node_proxy
             .runtime
             .new_abi_handle(new_node_proxy.node_id, reader);
 
         info!(
             "{:?}: start node instance {:?} of type {} with privilege {:?}",
-            self.get_node_name(node_id),
-            new_node_name,
+            self.get_node_debug_id(node_id),
+            NodeInfo::construct_debug_id(node_name, new_node_id),
```
tiziano88, line 1331: It does not seem great that this logic is replicating the actual logic that assigns the debug id to the actual channel object. Is it not possible to get the debug id from the newly created channel?

oak_runtime/src/proxy.rs
```diff
@@ -150,7 +150,9 @@ impl RuntimeProxy {
             self.channel_create("Initial", &Label::public_untrusted())?;
         debug!(
             "{:?}: created initial channel ({}, {})",
-            self.node_name, write_handle, read_handle,
+            NodeInfo::construct_debug_id(&self.node_name, self.node_id),
```
tiziano88, line 153: Would it make sense to instead have a method to fetch the channel object from the Runtime, and call get_debug_id on it, rather than duplicate this logic here in various places? I am not sure whether it is possible, but worth looking into

Even if not, it seems that this may just be encapsulated as a `get_debug_id` method on this RuntimeProxy object itself, or is there any reason for not doing that?

oak_runtime/src/channel.rs
```diff
@@ -76,14 +76,22 @@ pub struct Channel {
...
+        format!("{}({})", &self.name, channel_id)
```
rbehjati, line 91: Just curious... why is this in a separate `impl Channel` block than the one below with the `new` method?

oak_runtime/src/graph.rs
```diff
@@ -79,7 +79,7 @@ impl Runtime {
                     &mut s,
                     r###"    {} [label="{}" URL="{}"]"###,
                     node_id.dot_id(),
-                    node_info.name,
+                    NodeInfo::construct_debug_id(&node_info.name, *node_id),
```
rbehjati, line 82: Is it possible to replace this call (and the other calls in this file) with `node_info.get_debug_id(*node_id)`?

oak_runtime/src/lib.rs
```diff
@@ -154,18 +154,18 @@ struct NodeStopper {
 
 impl NodeStopper {
     /// Sends a notification to the Node and joins its thread.
-    fn stop_node(self) -> thread::Result<()> {
-        let node_name = &self.node_name;
+    fn stop_node(self, node_id: NodeId) -> thread::Result<()> {
+        let node_debug_id = NodeInfo::construct_debug_id(&self.node_name, node_id);
```
rbehjati, line 158: optional: Perhaps add a `get_debug_id` function to `NodeStopper` to be more consistent with `NodeInfo` and `Channel`. With this change, it probably makes sense to change `NodeInfo::construct_debug_id` into a normal function outside `NodeInfo`.

oak_runtime/src/lib.rs
```diff
@@ -739,7 +757,7 @@ impl Runtime {
         // the Channel Label).
         let downgraded_source_label = self.get_node_downgraded_label(node_id, source_label);
         let target_label = self.get_node_label(node_id);
-        let node_name = self.get_node_name(node_id);
+        let node_name = self.get_node_debug_id(node_id);
```
rbehjati, line 760: Nit! Could be fixed in the rest of the file too.
```suggestion
        let node_debug_id = self.get_node_debug_id(node_id);
```

oak_runtime/src/proxy.rs
rbehjati: Would not `self.runtime.get_node_debug_id(self.node_id)` work here?  (maybe it is not yet added to `node_infos`?)

I think having a `get_debug_id` function here makes more sense, especially since the logic in `get_debug_id` is very simple, and does not use anything from `NodeInfo`.

rbehjati (commented):
Just a few comments that could help the code look more consistent throughout :)

oak_runtime/src/channel.rs
juliettepretot: We don't currently store the `id` in either the `Channel` or `Node` struct. Instead nodes/channnels are stored in a hashmap with the id as a key. 

Do you think I should add the `id` to the struct, eliminating the need to pass it for this method?

oak_runtime/src/channel.rs
juliettepretot: Overlooked the other implementation block, will move it there :)

oak_runtime/src/graph.rs
juliettepretot: Would be, will do! :)

oak_runtime/src/lib.rs
juliettepretot: good point, done!

oak_runtime/src/lib.rs
juliettepretot: In this case yes. :) I wasn't quite sure if this log was after the node was already registered, but it is. Updated to use the `self.get_node_debug_id(new_node_id)` method.

juliettepretot:
@rbehjati RE your last comment (GitHub doesn't let me reply to it): That line happens before the node is created & registered with the runtime. 

However I'm happy to add a `get_debug_id` method to the proxy. :)

oak_runtime/src/proxy.rs
juliettepretot: That line happens before the node is created & registered with the runtime (via `self.node_create`), hence we can't fetch it yet.

However I'm added a get_debug_id method to the proxy. :)

oak_runtime/src/lib.rs
juliettepretot: Added a `get_debug_id` to `NodeStopper`. :) How would you go about moving `construct_debug_id` to its own function? 

I suppose I could share it across `Channel`s and `Node`s via:

```
enum Id {
    NodeId,
    ChannelId,
}

/// Returns a unique debug_id consisting out of the provided name and [`NodeId`].
fn construct_debug_id(name: &str, id: Id) -> String {
    // ...
}
```

oak_runtime/src/lib.rs
```diff
@@ -208,6 +217,19 @@ struct NodeInfo {
     node_stopper: Option<NodeStopper>,
 }
 
+impl NodeInfo {
+    /// Returns a unique debug_id used to identify the node in the debug output,
+    /// consisting out of the provided [`NodeId`], and the node's name.
```
rbehjati, line 222:
```suggestion
    /// consisting of the provided [`NodeId`], and the node's name.
```
Or is `consist out of` standard usage?

rbehjati (approved):
Thanks :)

oak_runtime/src/channel.rs
rbehjati: `Channel` does have an `id`. So I have suggested to change `self.channel.get_debug_id(self.channel.id)`  to `self.channel.get_debug_id()`.

But we don't have a struct for `Node` (it is a trait), and `node_id` is assigned to a node when it is registered in the runtime.

oak_runtime/src/lib.rs
rbehjati: I am not too worried about `Channel`. But yeah, `construct_debug_id` does not have to be inside `NodeInfo`. I don't think the signature needs to be changed.

oak_runtime/src/channel.rs
juliettepretot: Makes sense! Overlooked that, will rm that argument for nodes. :)

oak_runtime/src/lib.rs
```diff
@@ -183,7 +189,10 @@ impl std::fmt::Debug for NodeStopper {
 }
 
 struct NodeInfo {
-    /// Name for the Node in debugging output.
+    /// The name for the node.
```
conradgrobler, line 192: nit: s/node/Node/ in all new/updated comments for consistency with the rest of the file.

conradgrobler (approved):

PR #1748, opened 2020-11-19
Use Router in Trusted Database

ipetr0v:
This change updates Trusted Database example to use Router pattern.

Ref https://github.com/project-oak/oak/issues/1066

### DIFF
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -407,13 +407,17 @@ associated type to
 or manually reading from the initial `Receiver`:
 
 <!-- prettier-ignore-start -->
-[embedmd]:# (../examples/trusted_database/module/rust/src/lib.rs Rust /oak::entrypoint/ /.*let config_map =.*/)
+[embedmd]:# (../examples/aggregator/module/rust/src/lib.rs Rust /impl oak::CommandHandler/ /.*context.*/)
 ```Rust
-oak::entrypoint!(oak_main<ConfigMap> => |receiver: Receiver<ConfigMap>| {
-    let log_sender = oak::logger::create().unwrap();
-    oak::logger::init(log_sender, log::Level::Debug).unwrap();
+impl oak::CommandHandler for Main {
+    type Command = ConfigMap;
 
-    let config_map = receiver.receive().expect("Couldn't read config map");
+    fn handle_command(&mut self, command: ConfigMap) -> anyhow::Result<()> {
+        let log_sender = oak::logger::create()?;
+        oak::logger::init(log_sender.clone(), log::Level::Debug)?;
+        let config: Config =
+            toml::from_slice(&command.items.get("config").expect("Couldn't find config"))
+                .context("Couldn't parse TOML config file")?;
 ```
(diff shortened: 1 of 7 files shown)


docs/programming-oak.md
```diff
@@ -407,13 +407,17 @@ associated type to
 or manually reading from the initial `Receiver`:
```
tiziano88, line 407: Could you update this doc? I think we don't need to mention the initial Receiver any more, all nodes are expected to impl CommandHandler now.

examples/trusted_database/module/rust/src/router.rs
```diff
@@ -0,0 +1,73 @@
...
+            .receiver
+            .clone()
+            .context("Couldn't get receiver")?
+            .label()
+            .context("Couldn't get label")?;
+        let init = TrustedDatabaseInit {
+            log_sender: Some(self.log_sender.clone()),
+            points_of_interest: Some(self.points_of_interest.clone()),
+        };
+        let handler_invocation_sender =
```
tiziano88, line 64: Maybe add some comments to explain the logic here.

examples/trusted_database/module/rust/src/router.rs
```diff
@@ -0,0 +1,73 @@
...
+    io::{ReceiverExt, Sender, SenderExt},
+    CommandHandler,
+};
+use oak_services::proto::oak::log::LogMessage;
+
+/// Oak Router Node that contains an in-memory database.
+#[derive(Default)]
+pub struct Router {
+    log_sender: Sender<LogMessage>,
+    points_of_interest: PointOfInterestMap,
```
tiziano88, line 33: May be simpler to just store the init message here, since it's the same for router and handler node, right? up to you

examples/trusted_database/module/rust/src/router.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl CommandHandler for Router {
+    type Command = grpc::Invocation;
+
+    fn handle_command(&mut self, invocation: Self::Command) -> anyhow::Result<()> {
+        let label = invocation
+            .receiver
+            .clone()
```
tiziano88, line 56: Why does the receiver need to be cloned?

tiziano88 (approved):

examples/trusted_database/module/rust/src/router.rs
ipetr0v: Done
Using `.as_ref()`

examples/trusted_database/module/rust/src/router.rs
ipetr0v: Added a comment for `impl CommandHandler for Router`

docs/programming-oak.md
ipetr0v: Deleted `or manually reading from the initial Receiver`

rbehjati (approved):
LGTM! I'll update the example with authentication logic after you merge this change. 

I believe in the description of the PR `Running Average` should be changed to `Trusted Database` :)

PR #1752, opened 2020-11-20
Give top privilege to logger in debug builds

### LINKED ISSUE #1751: Set log node declassification privilege to "top" in `oak_debug` builds
This will allow applications running in debug mode to create the logger as with top secret confidentiality, which would allow all nodes to successfully log during debug runs.

The SDK should also be changed to try to create the log node as top secret (which would fail in non-debug builds) and fall back to creating a public log node if that does not work.

conradgrobler:
Fixes #1751 

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/node/logger.rs
+++ b/oak_runtime/src/node/logger.rs
@@ -18,7 +18,7 @@
 
 use crate::{
     io::{Receiver, ReceiverExt},
-    RuntimeProxy,
+    NodePrivilege, RuntimeProxy,
 };
 use log::{error, info, warn};
 use oak_abi::OakStatus;
@@ -44,6 +44,12 @@ impl LogNode {
 impl super::Node for LogNode {
     fn node_type(&self) -> &'static str {
         "logger"
+    }
+
+    #[cfg(feature = "oak_debug")]
+    fn get_privilege(&self) -> NodePrivilege {
+        // Allow the logger Node to declassify log messages in debug builds only.
+        NodePrivilege::top_privilege()
     }
 
     /// Main execution loop for the logging pseudo-Node just waits for incoming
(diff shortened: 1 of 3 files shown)


oak_runtime/src/node/logger.rs
```diff
@@ -46,6 +46,12 @@ impl super::Node for LogNode {
         "logger"
     }
 
+    #[cfg(feature = "oak_debug")]
+    fn get_privilege(&self) -> NodePrivilege {
+        // Allow the logger Node to declassify log messages in debug builds only.
+        NodePrivilege::top_privilege()
+    }
```
tiziano88, line 53: Would it make sense to explicitly have get_privilege returning the empty privilege here for non-debug? Or do you think it's sufficiently clear that it is the default? Maybe just mention the default?

sdk/rust/oak/src/logger/mod.rs
```diff
@@ -82,9 +85,20 @@ pub fn init(log_sender: Sender<LogMessage>, level: Level) -> Result<(), SetLogge
 }
 
 pub fn create() -> Result<Sender<LogMessage>, OakStatus> {
-    crate::io::node_create(
+    match crate::io::node_create(
```
tiziano88, line 88: (optional) Alternative more functional way:

```rust
for label in [top, public] {[
  match node_create(...) {
    Err(PermissionDenied) => continue,
    Err(err) => return Err(err),
    Ok(n) => return Ok(n),
  }
}
Err(...)
```

tiziano88 (approved):

oak_runtime/src/node/logger.rs
conradgrobler: Added an explict implementation to return the empty privilege for non-debug builds.

sdk/rust/oak/src/logger/mod.rs
conradgrobler: Done.

PR #1753, opened 2020-11-20
Allow multiple versions of the same crate in cargo-deny

### LINKED ISSUE #1270: Cargo-deny: Stop skipping duplicate crate checks once older versions are no longer required
- `rusty-machine` depends on older versions of `rand`
- `tonic`, `jsonwebtoken' and `rustls` depend on an older version of `base64`
- `spinners` depends on older versions of
  - `ansi_term`
  - `quote`
  - `syn`
  - `unicode-xid`

conradgrobler:
We currently allow multiple versions anyway by just adding exceptions and then removing it later when no longer needed. This change means that we don't have to keep doing this maintenance for now while the versions are not very stable. We can add the checks in later for specific crates if exact versions become important.

Closes #1270 

# Checklist

      construction.

### DIFF
--- a/examples/deny.toml
+++ b/examples/deny.toml
@@ -20,50 +20,8 @@ ignore = [
 
 # Deny multiple versions unless explicitly skipped.
 [bans]
-multiple-versions = "deny"
+multiple-versions = "allow"
 wildcards = "allow"
-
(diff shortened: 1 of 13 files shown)


rbehjati (approved):
Thanks :)

Issue #1529, opened 2020-09-28
Case study: local-only WiFi password QR code generator


tiziano88:
Spawned from https://github.com/project-oak/oak/issues/1492 .

Case study:

Random developer wants to implement an app to generate a QR code from a WiFi password.

Looking around, there are currently a number of such apps available, including web-based and native, e.g.:

- https://qifi.org/
- https://www.qr-code-generator.com/solutions/wifi-qr-code/
- https://qrcode.tec-it.com/en/wifi
- https://coda.io/@rod/wifi-qr-code-generator

Some of them have some text trying to justify that it is safe to enter passwords there, for instance, from https://qifi.org/
:

> About
Ever wanted to create a cool QR code for your guests? But never wanted to type in your WiFi credentials into a form that submits them to a remote webserver to render the QR code? QiFi for the rescue! It will render the code in your browser, on your machine, so the WiFi stays as secure as it was before (read the code if you do not trust text on the internet :-))!
>
> If you use the Save-button to store a code, this is still secure, as the data is stored in HTML5 localStorage and is never transmitted to the server (in contrast to cookie-based solutions).
>
> Don't trust your browser either? Just pipe the string WIFI:S:<SSID>;T:<WPA|WEP|>;P:<password>;; through the QR code generator of your choice after reading the documentation.

User wants to use such app but because because it requires entering their own WiFi password, they want to be sure that such app cannot send anything over to a remote server.

Note that this is not even specific to web apps; we do this all the time by downloading binaries (e.g. games, text editors, etc.) and running them locally, under the assumption that they behave correctly. In some cases it is possible to download the source and either inspect it or compile it locally in order to have some guarantees, but in practice this tends not to happen, and if it does happen requires inspecting the entirety of the code (including third party dependencies).

I would like to think more how this example, which is arguably much simpler than many other Oak use cases, can actually be solved.

In particular, it seems that this should be solved completely locally, and with no reliance on a remote server or remote attestation between client and server.

If we imagine that the logic (including rendering) of the app is contained in a single Wasm file, then this seems solvable by introducing a web-based version of the Oak Runtime which is hosted on a well-known domain (e.g. https://oak.google.com) which is trusted **by assumption** by users once and for all (i.e. does not depend on the particular implementation of the third party app that will run on top of it). 

For instance, the Oak host page may contain a bit of logic roughly equivalent to the Oak Runtime, plus additional logic for rendering DOM efficiently.

[...]

PR #1754, opened 2020-11-20
Update chrome extension to use Content-Security-Policies

juliettepretot:
Ref: #1492, #1529

This PR updates the chrome-extension significantly. Instead of blocking requests after entering an opt-in "Oak Mode", it now relies on [Content-Security-Policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) to enforce that data cannot be leaked. 

Websites that implement a secure CSP (no network calls, prevents navigation, cookies, localStorage etc) are shown to users as safe with a green icon, other websites are shown as red. See the readme for more info. 

In the future this can be extended to allow communications with verified Oak servers, imposing this CSP on websites (likely breaking them), or even sharing confidential data with "green" websites without permission dialogs. 

It seems this could come in handy for the UI of our hackathon project. 

This work is based on my investigations of CSP, and discussions with @tiziano88, and @seefeldb (who suggested using the sandbox mechanism to prevent leaks via navigation).

Note that although we couldn't find significant bypasses to leak data (apart from attacks relying on local proximity via sound/bluetooth), I invite you to play around with it and try to break it. 

[https://github.com/juliettepretot/csp-restrictive-serve](https://github.com/juliettepretot/csp-restrictive-serve) could come in handy as a playground, it serves a page with the required CSP and can be used to try various JS APIs.

# Checklist

      construction.


chrome_extension/README.md
```diff
@@ -3,32 +3,36 @@
...
+This enforcement happens via the
+[Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)
+of
+`sandbox allow-scripts; default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline'`.
+It prevents calling servers or navigating to other pages.
+
+Depending on whether a given website complies with this CSP or not, the
+extension will show either a red or green icon.
+
+<img src="icon-red.png" alt="drawing" width="32" height="32"/>
```
juliettepretot, line 20: Using HTML tags for the images to show the icon in a smaller (32x32px instead 128x128px) format.

tiziano88 (approved):
amazing stuff!

ipetr0v (approved):

PR #1756, opened 2020-11-20
Use Router and labels in Translator

ipetr0v:
This change adds Router pattern and client labels to Translator example.
Translator creates individual Handler Nodes for each client request and assigns client labels to them.

Ref https://github.com/project-oak/oak/pull/1745

### DIFF
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -98,7 +98,7 @@ struct Handler;
 impl Translator for Handler {
     fn translate(&mut self, req: TranslateRequest) -> grpc::Result<TranslateResponse> {
         info!(
-            "attempt to translate '{}' from {} to {}",
+            "Attempt to translate '{}' from {} to {}",
             req.text, req.from_lang, req.to_lang
         );
         let mut rsp = TranslateResponse::default();
@@ -108,7 +108,7 @@ impl Translator for Handler {
                     "fr" => "MONDES".to_string(),
                     "it" => "MONDI".to_string(),
                     _ => {
-                        info!("output language {} not found", req.to_lang);
+                        info!("Output language {} not found", req.to_lang);
                         return Err(grpc::build_status(
                             grpc::Code::NotFound,
                             "Output language not found",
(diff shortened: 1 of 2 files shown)


ipetr0v:
This example will also require client authentication labels.

tiziano88:
I wasn't expecting the translator or hello world examples to have user identity labels, but if we are going that way perhaps we can remove even more of the other examples? Basically this also becomes something similar to the trusted database example, right?

ipetr0v:
Yes, the application structure (router + client label) is similar.

tiziano88 (approved):

docs/programming-oak.md
```diff
@@ -187,17 +187,17 @@ impl oak::CommandHandler for Main {
...
             &Label::public_untrusted(),
             "app",
             LogInit {
                 log_sender: Some(log_sender),
             },
         )
-        .context("could not create handler node")?;
-        oak::grpc::server::init_with_sender("[::]:8080", handler_sender)
-            .context("could not create gRPC server pseudo-Node")?;
+        .context("Couldn't create handler node")?;
```
rbehjati, line 198:
```suggestion
        .context("Couldn't create router node")?;
```
I guess?

examples/translator/module/rust/src/lib.rs
```diff
@@ -33,21 +37,61 @@ impl oak::CommandHandler for Main {
...
             &Label::public_untrusted(),
             "app",
             LogInit {
                 log_sender: Some(log_sender),
             },
         )
-        .context("could not create handler node")?;
-        oak::grpc::server::init_with_sender("[::]:8080", handler_sender)
-            .context("could not create gRPC server pseudo-Node")?;
+        .context("Couldn't create handler node")?;
```
rbehjati, line 48:
```suggestion
        .context("Couldn't create router node")?;
```

rbehjati (approved):

examples/translator/module/rust/src/lib.rs
ipetr0v: Done

docs/programming-oak.md
ipetr0v: Done

PR #1757, opened 2020-11-20
Consistently pass the certificate as byte vector in the configurations

### LINKED ISSUE #1747: Consistently represent certificates as a byte vector
Ref [comment on #1728](https://github.com/project-oak/oak/pull/1728/files#r525353632)

rbehjati:
Fixes #1747

### DIFF
--- a/oak_loader/src/options.rs
+++ b/oak_loader/src/options.rs
@@ -23,7 +23,7 @@ use log::debug;
 use oak_abi::proto::oak::application::{ApplicationConfiguration, ConfigMap};
 use oak_runtime::{
     auth::oidc_utils::{parse_client_info_json, ClientInfo},
-    config::load_certificate,
+    tls::Certificate,
     SignatureTable,
 };
 use oak_sign::SignatureBundle;
@@ -218,7 +218,7 @@ fn create_grpc_config(opt: &Opt) -> anyhow::Result<oak_runtime::GrpcConfiguratio
         Some(path) => read_to_string(path).context("Couldn't read gRPC TLS certificate")?,
         None => return Err(anyhow!("No gRPC TLS certificate file provided.")),
     };
-    let root_tls_certificate = get_root_tls_certificate_or_default(&opt)?;
+    let root_tls_certificate = get_root_tls_certificate_or_default(&opt).ok();
     let oidc_client_info = get_oidc_client_info(&opt)?;
 
     let grpc_config = oak_runtime::GrpcConfiguration {
@@ -226,10 +226,7 @@ fn create_grpc_config(opt: &Opt) -> anyhow::Result<oak_runtime::GrpcConfiguratio
             &grpc_tls_certificate,
             &grpc_tls_private_key,
         )),
-        grpc_client_root_tls_certificate: Some(
-            load_certificate(&root_tls_certificate)
-                .map_err(|()| anyhow!("Couldn't parse TLS certificate"))?,
-        ),
+        grpc_client_root_tls_certificate: root_tls_certificate,
         oidc_client_info,
     };
 
@@ -303,8 +300,7 @@ fn create_http_config(opt: &Opt) -> anyhow::Result<oak_runtime::HttpConfiguratio
 
     match oak_runtime::tls::TlsConfig::new(http_tls_certificate_path, http_tls_private_key_path) {
         Some(tls_config) => {
-            let root_tls_certificate =
-                get_root_tls_certificate_or_default(&opt).map(|s| s.as_bytes().to_vec())?;
+            let root_tls_certificate = get_root_tls_certificate_or_default(&opt).ok();
             Ok(oak_runtime::HttpConfiguration {
                 tls_config,
                 http_client_root_tls_certificate: root_tls_certificate,
@@ -316,10 +312,12 @@ fn create_http_config(opt: &Opt) -> anyhow::Result<oak_runtime::HttpConfiguratio
     }
 }
 
-/// If `oak_debug` is enabled, read root TLS certificate from the specified file. Otherwise, return
-/// the default root TLS certificate from the embedded byte array.
-fn get_root_tls_certificate_or_default(opt: &Opt) -> anyhow::Result<String> {
-    match &opt.root_tls_certificate {
+/// If `oak_debug` is enabled, reads root TLS certificate from the specified file into a byte array.
+/// Otherwise, loads the default root TLS certificate from the embedded byte array.
+/// Parses the byte array into a [`Certificate`], or returns an error if the byte array does not
+/// represent a valid PEM formatted certificate.
+fn get_root_tls_certificate_or_default(opt: &Opt) -> anyhow::Result<Certificate> {
+    let certificate_bytes = match &opt.root_tls_certificate {
         Some(certificate_path) => {
             if !cfg!(feature = "oak_debug") {
                 // Unreachable: it should not be possible to specify the flag without `oak_debug`.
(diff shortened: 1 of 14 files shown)


oak_loader/src/options.rs
```diff
@@ -348,11 +343,8 @@ fn get_oidc_client_info(opt: &Opt) -> anyhow::Result<Option<ClientInfo>> {
 }
 
 /// Gets the default root TLS certificates from the embedded byte array.
-fn get_default_root_tls_certs() -> anyhow::Result<String> {
-    let result = std::str::from_utf8(include_bytes!("certs/roots.pem"))
-        .context("could not read embedded PEM-encoded root TLS certificates as a UTF8 string")?
-        .to_owned();
-    Ok(result)
+fn get_default_root_tls_certs() -> Vec<u8> {
+    include_bytes!("certs/roots.pem").to_vec()
```
rbehjati, line 347: The `TLS certificate` fields in the configurations are now `Vec<u8>`. So I changed it here too. I am not sure how important the check for conversion to a UTF8 string is. But this change list moves all such checks into the `GrpcClientNode::new` (basically inside `load_certificate`). Do we generally want to have these checks earlier, or is it ok to postpone them until the actual node creation?

conradgrobler (approved):

oak_loader/src/options.rs
```diff
@@ -318,7 +313,7 @@ fn create_http_config(opt: &Opt) -> anyhow::Result<oak_runtime::HttpConfiguratio
 
 /// If `oak_debug` is enabled, read root TLS certificate from the specified file. Otherwise, return
 /// the default root TLS certificate from the embedded byte array.
```
tiziano88, line 315: Could you clarify what format the resulting certificate is in? Is it PEM? base64 encoded?

oak_runtime/src/node/grpc/client.rs
```diff
@@ -85,6 +85,8 @@ impl GrpcClientNode {
         // We compute the node privilege once and for all at start and just store it, since it does
         // not change throughout the node execution.
         let node_privilege = grpc_client_node_privilege(&uri);
+        let root_tls_certificate = crate::config::load_certificate(&root_tls_certificate_bytes)
+            .map_err(|_err| ConfigurationError::InvalidNodeConfiguration)?;
```
tiziano88, line 89: Also log the error?

tiziano88 (approved):

oak_loader/src/options.rs
rbehjati: Added a call to `rustls::internal::pemfile::certs`, just for checking the validity of the certificate. But the resulting certificate object is discarded. The same check happens inside `load_certificate`, which I think can be left there as well, since that function might be called independently too.

oak_runtime/src/node/grpc/client.rs
rbehjati: I don't think that is needed. The error (`_err`) does not have any specific content (its type is`()`), and `load_certificate` already logs an error.

oak_loader/src/options.rs
rbehjati: Done.

rbehjati (commented):
Thanks for the comments :)

oak_loader/src/options.rs
```diff
@@ -348,11 +348,22 @@ fn get_oidc_client_info(opt: &Opt) -> anyhow::Result<Option<ClientInfo>> {
...
+}
+
+/// Check that input byte vector represents a valid PEM formatted certificate.
+fn validate_certificate(certificate_bytes: &[u8]) -> anyhow::Result<()> {
+    let mut cursor = std::io::Cursor::new(certificate_bytes);
+
+    // We don't need the resulting certificate, but just check that a valid certificate can be
+    // created from the given byte array. `rustls` doesn't specify certificate parsing errors:
+    // https://docs.rs/rustls/0.17.0/rustls/internal/pemfile/fn.certs.html
+    rustls::internal::pemfile::certs(&mut cursor).map_err(|_err| {
```
tiziano88, line 362: You can capture what you just said in words via an irrefutable pattern, which also shows the type, and will fail to match if the type of the error changes in the future:

```suggestion
    rustls::internal::pemfile::certs(&mut cursor).map_err(|()| {
```

oak_loader/src/options.rs
```diff
@@ -348,11 +348,22 @@ fn get_oidc_client_info(opt: &Opt) -> anyhow::Result<Option<ClientInfo>> {
...
-    let result = std::str::from_utf8(include_bytes!("certs/roots.pem"))
-        .context("could not read embedded PEM-encoded root TLS certificates as a UTF8 string")?
-        .to_owned();
-    Ok(result)
+fn get_default_root_tls_certs() -> Vec<u8> {
+    include_bytes!("certs/roots.pem").to_vec()
+}
+
+/// Check that input byte vector represents a valid PEM formatted certificate.
+fn validate_certificate(certificate_bytes: &[u8]) -> anyhow::Result<()> {
```
tiziano88, line 356: So why not make this function actually parse the cert as the rustls type, and pass that around instead of `Vec<u8>`? Or if that type is not good for some reason, create your own wrapper `ValidatedCert` or similar? Not sure it's worth it, but I think it would make the code more readable, since the current `Result<()>` return type is a bit weird.

see https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

oak_runtime/src/config.rs
```diff
@@ -51,11 +51,15 @@ pub fn configure_and_run(config: RuntimeConfiguration) -> Result<Arc<Runtime>, O
...
+        error!("The certificate must be a valid non-empty byte vector");
+        return Err(());
+    }
+    let mut cursor = std::io::Cursor::new(certificate_bytes);
     // `rustls` doesn't specify certificate parsing errors:
     // https://docs.rs/rustls/0.17.0/rustls/internal/pemfile/fn.certs.html
     rustls::internal::pemfile::certs(&mut cursor)?;
 
-    Ok(Certificate::from_pem(certificate))
+    Ok(Certificate::from_pem(certificate_bytes))
```
tiziano88, line 64: I haven't looked at it closely, but why is essentially the same logic present in both oak_loader and oak_runtime, with slightly different error handling?

oak_loader/src/options.rs
rbehjati: Created our own `Certificate` type. The difference between `rustls::Certificate` and `tonic::Certificate` is that the former parses a PEM file into a vector of certificates, while the latter parses a PEM file into a single Certificate object. With our internal type, we can ensure that the certificate that we pass around is a valid certificate; and can convert it into each of the destination formats when needed.

oak_loader/src/options.rs
rbehjati: Of course! Thanks :)

oak_runtime/src/config.rs
rbehjati: The checks are now moved to `oak_runtime::tls::Certificate::parse`.

oak_runtime/src/tls.rs
```diff
@@ -14,20 +14,54 @@
...
+    fn from(certificate: Certificate) -> Self {
+        tonic::transport::Certificate::from_pem(certificate)
+    }
+}
 
 /// Represents TLS identity to use for HTTP server pseudo-nodes.
 #[derive(Default, Clone)]
 pub struct TlsConfig {
-    certs: Vec<Certificate>,
+    certs: Vec<rustls::Certificate>,
```
tiziano88, line 64: Could / should this be the newly created Certificate type too?

tiziano88 (approved):
Nice one!

oak_runtime/src/tls.rs
rbehjati: I think it is better to keep the `rustls` version here. The other field in `TlsConfig` is `rustls::PrivateKey`, and these fields are used to create other `rustls` objects. I have qualified `PrivateKey` with `rustls` too, for better readability.

PR #1758, opened 2020-11-20
Use Router and labels in Hello World

ipetr0v:
This change adds Router pattern and client labels to Hello World example.
Hello World creates individual Handler Nodes for each client request and assigns client labels to them.

Ref #1745

### DIFF
--- a/examples/hello_world/module/rust/src/lib.rs
+++ b/examples/hello_world/module/rust/src/lib.rs
@@ -26,18 +26,23 @@ pub mod proto {
     }
 }
 
-use log::info;
-use oak::grpc;
-use oak_abi::{label::Label, proto::oak::application::ConfigMap};
-use oak_services::proto::oak::log::LogInit;
-use proto::oak::examples::hello_world::{
-    HelloRequest, HelloResponse, HelloWorld, HelloWorldDispatcher, Init,
+pub mod handler;
+
+use anyhow::Context;
+use handler::Handler;
+use oak::{
+    grpc,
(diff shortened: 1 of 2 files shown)


ipetr0v:
This example will also require client authentication labels.

tiziano88 (approved):

examples/hello_world/module/rust/src/lib.rs
```diff
@@ -26,146 +26,100 @@ pub mod proto {
...
+impl CommandHandler for Router {
+    type Command = grpc::Invocation;
 
-impl Handler {
-    fn translate(&self, text: &str, from_lang: &str, to_lang: &str) -> Option<String> {
-        let client = self.translator.as_ref()?;
-        translator_common::translate(client, text, from_lang, to_lang)
-    }
-}
+    fn handle_command(&mut self, invocation: Self::Command) -> anyhow::Result<()> {
```
rbehjati, line 89: Perhaps add a brief comment here. Does it create two nodes with the same label?

rbehjati (approved):

examples/hello_world/module/rust/src/lib.rs
ipetr0v: Comment was added above `impl`
https://github.com/project-oak/oak/pull/1758/files#diff-68dc93d24ee4fa90a8be1ebc02c4b778eabcc2ae2898d5e6de9fa7773f313b35R85

PR #1761, opened 2020-11-23
Use GitHub Actions for CI

tiziano88:
This is in addition to the GCP run for the time being, and not blocking
for any submission, but it should give a faster feedback for some of the
  steps since they are run in parallel.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,52 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  runner:
+    runs-on: ubuntu-latest
+    # We use the same job template, but parametrize the actual command to be passed to the runner
+    # binary using the matrix strategy, so that we get the commands running in parallel.
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          - check-format
+          - run-cargo-deny
+          - run-cargo-udeps
+          - build-server --server-variant=base
+          - build-server --server-variant=logless
+          - run-tests
+          - run-tests-tsan
+          - run-examples --application-variant=rust
+          - run-examples --application-variant=cpp
+          - run-examples --application-variant=rust --example-name=hello_world
+            --build-docker
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      # Copied from https://github.com/jens-maus/RaspberryMatic/blob/ea6b8ce0dd2d53ea88b2766ba8d7f8e1d667281f/.github/workflows/ci.yml#L34-L40
+      - name: Free disk space
+        run: |
+          df --human-readable
+          sudo swapoff --all
+          sudo rm --force /swapfile
+          sudo apt clean
+          docker rmi $(docker image ls --all --quiet)
+          df --human-readable
+
+      # Build Docker image, caching from the latest version from the remote repository.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+          - run-examples --application-variant=cpp
+          - run-examples --application-variant=rust --example-name=hello_world
+            --build-docker
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      # Copied from https://github.com/jens-maus/RaspberryMatic/blob/ea6b8ce0dd2d53ea88b2766ba8d7f8e1d667281f/.github/workflows/ci.yml#L34-L40
+      - name: free disk space
```
daviddrysdale, line 35: super-nit: "Free disk space" to match capitalization of other steps

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+          - run-cargo-deny
+          - run-cargo-udeps
+          - build-server --server-variant=base
+          - build-server --server-variant=logless
+          - run-tests
+          - run-tests-tsan
+          - run-examples --application-variant=rust
+          - run-examples --application-variant=cpp
+          - run-examples --application-variant=rust --example-name=hello_world
+            --build-docker
```
daviddrysdale, line 28: Does YAML automatically join split lines somehow?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+        cmd:
+          - check-format
+          - run-cargo-deny
+          - run-cargo-udeps
+          - build-server --server-variant=base
+          - build-server --server-variant=logless
+          - run-tests
+          - run-tests-tsan
+          - run-examples --application-variant=rust
+          - run-examples --application-variant=cpp
```
daviddrysdale, line 26: The way that GitHub displays individual steps makes it hard to see which step is which among these &ndash; I guess there's no way around that?

daviddrysdale (approved):
Is #1762 going to mean that these CI tests are going to fail with some regularity?  (I.e. should we hold off merging this until #1762 is fixed?)

.github/workflows/ci.yaml
tiziano88: Done.

.github/workflows/ci.yaml
tiziano88: Yes, in fact this line was split by prettier or some other linter.

.github/workflows/ci.yaml
tiziano88: Not that I can think of, anyways it seems ok to me, these do show up in parentheses next to the step name.

tiziano88 (commented):
> Is #1762 going to mean that these CI tests are going to fail with some regularity? (I.e. should we hold off merging this until #1762 is fixed?)

Yes it may mean that these fail, but they are not blocking submission for now (they are just informational), and in fact it may be good to see that the tests stop failing after #1762 is fixed if that's the case.

.github/workflows/ci.yaml
daviddrysdale: image

tiziano88: Uhmm god it. Anyways it may be possible to specify objects with {name, command} fields in the matrix and extract the two individually, I may try that separately but also not sure it's worth it.

PR #1763, opened 2020-11-23
Add SDK support for HTTP client pseudo-node

rbehjati:
# Checklist

      construction.

### DIFF
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -16,8 +16,9 @@
 
 //! Simple example starting an Oak Application serving a static page over HTTP.
 
-use log::{info, warn};
-use oak::{http::Invocation, CommandHandler, Label, OakError, OakStatus};
+use anyhow::Context;
+use log::info;
+use oak::{http::Invocation, CommandHandler, Label};
 use oak_abi::proto::oak::application::ConfigMap;
 use oak_services::proto::oak::log::LogInit;
 
@@ -41,7 +42,7 @@ impl oak::CommandHandler for Main {
             },
         )
         .expect("could not create handler node");
-        oak::http::init_with_sender("[::]:8080", http_handler_sender)
+        oak::http::server::init_with_sender("[::]:8080", http_handler_sender)
             .expect("could not create HTTP server pseudo-Node");
         Ok(())
     }
@@ -49,7 +50,8 @@ impl oak::CommandHandler for Main {
 
 oak::entrypoint_command_handler_init!(http_handler => StaticHttpHandler);
 
-/// A simple HTTP handler that responds with `OK` (200) to every request sent to `/`, and with
+/// A simple HTTP handler that responds with `OK` (200) to every request sent to `/`; with a
+/// response from `http://www.google.com` for every request sent to `/test_google`; and with
 /// `NOT_FOUND` (400) to any other request. It is used in the `abitest`. So its functionality
 /// should be modified with care!
 #[derive(Default)]
@@ -76,15 +78,26 @@ impl CommandHandler for StaticHttpHandler {
             "/" => http::response::Builder::new()
                 .status(http::StatusCode::OK)
                 .header(http::header::CONTENT_TYPE, "text/html; charset=UTF-8")
-                .body(include_bytes!("../static/index.html").to_vec()),
+                .body(include_bytes!("../static/index.html").to_vec())
+                .context("Could not build response"),
+            "/test_google" => {
+                // create a public HTTP client pseudo-node
+                let http_client = oak::http::client::init("")
+                    .context("Could not create HttpClient pseudo-node")?;
+                let request = http::Request::builder()
+                    .method(http::Method::GET)
+                    .uri("http://www.google.com")
+                    .body(vec![])
+                    .context("Could not build request")?;
+                http_client
+                    .send_request(request, &Label::public_untrusted())
+                    .context("Could not get response")
+            }
             _ => http::response::Builder::new()
                 .status(http::StatusCode::NOT_FOUND)
-                .body("not found".to_string().into_bytes()),
-        };
-        let response = response.map_err(|err| {
-            warn!("Could not build response: {}", err);
-            OakError::OakStatus(OakStatus::ErrInternal)
-        })?;
+                .body("not found".to_string().into_bytes())
+                .context("Could not build response"),
+        }?;
 
         let res = invocation.send(&response);
         invocation.close_channels();
--- a/sdk/rust/oak/src/http/client.rs
+++ b/sdk/rust/oak/src/http/client.rs
@@ -1,0 +1,84 @@
+//
+// Copyright 2020 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{
+    http::{util::Pipe, Invocation},
+    OakError,
+};
+use http::{Request, Response};
+use oak_abi::label::{confidentiality_label, tls_endpoint_tag, Label};
+use oak_io::Sender;
+use oak_services::proto::oak::encap::HttpRequest;
+use std::convert::TryInto;
+
+/// Initializes an HTTP client pseudo-Node serving requests that have the given authority in their
+/// URIs. If the authority is empty, the node is public and can handle any incoming request.
+/// Otherwise, the node is created with the `authority` as its confidentially tag. In this case, the
+/// client can only serve HTTPS requests to the services indicated by the `authority`.
+///
+/// Returns an [`HttpClient`].
+pub fn init(authority: &str) -> Result<HttpClient, OakError> {
+    let config = crate::node_config::http_client(authority);
+    let label = if authority.is_empty() {
+        Label::public_untrusted()
+    } else {
+        confidentiality_label(tls_endpoint_tag(authority))
+    };
+    let sender = crate::io::node_create::<Invocation>("http_client", &label, &config)?;
+    Ok(HttpClient {
+        sender,
+        authority: authority.to_string(),
+    })
+}
+
+pub struct HttpClient {
+    sender: Sender<Invocation>,
+    authority: String,
+}
+
+impl HttpClient {
+    /// Sends the given request via the HTTP client pseudo-node. The `response_label` should match
+    /// the label of the Oak Node.
+    pub fn send_request(
+        &self,
+        request: Request<Vec<u8>>,
+        response_label: &Label,
+    ) -> Result<Response<Vec<u8>>, OakError> {
+        // Use the authority of the client node as the label of the request channel.
+        let request_label = if self.authority.is_empty() {
+            Label::public_untrusted()
+        } else {
+            confidentiality_label(tls_endpoint_tag(&self.authority))
+        };
+        // create a pipe with a request and a response channel
+        let pipe = Pipe::new(&request_label, response_label)?;
+
+        // insert the request into the request channel
+        pipe.insert_message(HttpRequest::from(request))?;
+
+        // wrap the request and response channels in an invocation, and send it to the HTTP client
+        // pseudo-node
+        pipe.send_invocation(&self.sender)?;
+
+        // wait for the response to be received
+        let rsp = pipe.get_response()?;
+
+        // close all the channels in the pipe
+        pipe.close();
+
+        rsp.try_into().map_err(OakError::OakStatus)
+    }
+}
--- a/sdk/rust/oak/src/http/mod.rs
+++ b/sdk/rust/oak/src/http/mod.rs
@@ -14,20 +14,22 @@
 // limitations under the License.
 //
 
-//! Functionality to help Oak Nodes create HTTP server pseudo-Nodes.
+//! Functionality to help Oak Nodes create HTTP server and HTTP client pseudo-Nodes.
 
 use crate::{
-    io::{Receiver, ReceiverExt, Sender, SenderExt},
-    proto::oak::invocation::HttpInvocationSender,
+    io::{ReceiverExt, SenderExt},
     OakError, OakStatus,
 };
 use http::{Request, Response};
 use log::warn;
-use oak_abi::label::Label;
 pub use oak_services::proto::oak::encap::{HeaderMap, HttpRequest, HttpResponse};
 use std::convert::TryInto;
 
 pub type Invocation = crate::proto::oak::invocation::HttpInvocation;
+
+pub mod client;
+pub mod server;
+mod util;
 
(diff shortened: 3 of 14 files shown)


sdk/rust/oak/src/http/mod.rs
```diff
@@ -70,61 +72,3 @@ impl Invocation {
         };
     }
 }
-
-/// Initializes an HTTP server pseudo-Node listening on the provided address.
```
rbehjati, line 74: Moved to `http/server.rs`.

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,103 @@
+//
```
rbehjati, line 1: Copied from `oak_runtime/src/node/http/util.rs`, but uses the SDK version of some of the types.

sdk/rust/oak/src/http/client.rs
```diff
@@ -0,0 +1,84 @@
...
+    sender: Sender<Invocation>,
+}
+
+impl HttpClient {
+    pub fn send_request(
+        &self,
+        request: Request<Vec<u8>>,
+        request_label: &Label,
+    ) -> Result<Response<Vec<u8>>, OakError> {
+        let pipe = Pipe::new(&request_label, &Label::public_untrusted())?;
```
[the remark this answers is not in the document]
conradgrobler: As discussed, I think the request channel should have the same label as the http client node, and the response channel should have the same label as the node sending the request to the http client node.

conradgrobler (approved):

examples/http_server/module/src/lib.rs
```diff
@@ -76,15 +78,26 @@ impl CommandHandler for StaticHttpHandler {
...
+                let http_client = oak::http::client::init("")
+                    .context("Could not create HttpClient pseudo-node")?;
+                let request = http::Request::builder()
+                    .method(http::Method::GET)
+                    .uri("http://www.google.com")
+                    .body(vec![])
+                    .context("Could not build request")?;
+                http_client
+                    .send_request(request, &Label::public_untrusted())
+                    .context("Could not get response")
```
tiziano88, line 94: FYI Creating a new node from another node is not very realistic (most of the examples now create all the nodes in the main node). Also it seems this example is not really exercising TLS label, but given how it is written already it could easily do that too, right? For instance by handling data labelled with https:google.com TLS label.

sdk/rust/oak/src/http/client.rs
```diff
@@ -0,0 +1,87 @@
...
+    let method = req.method().as_str().to_string();
+    let headers = Some(HeaderMap::from(req.headers().to_owned()));
+    let body = req.into_body();
+
+    HttpRequest {
+        uri,
+        method,
+        body,
+        headers,
+    }
```
tiziano88, line 86: Isn't there a From or Into conversion between these types? If not, should there be?

sdk/rust/oak/src/http/client.rs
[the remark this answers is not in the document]
tiziano88: +1, definitely not public.

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// A pair of temporary channels to pass the HTTP request to the Oak Node and receive the response.
+pub struct Pipe {
+    pub request_writer: Sender<HttpRequest>,
+    pub request_reader: Receiver<HttpRequest>,
+    pub response_writer: Sender<HttpResponse>,
+    pub response_reader: Receiver<HttpResponse>,
+}
+
+impl Pipe {
```
tiziano88, line 34: Personally I'm not convinced this type makes things any simpler, it seems to just couple the API of the underlying senders and receivers.

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn new(request_label: &Label, response_label: &Label) -> Result<Self, OakError> {
+        // Create a channel for passing HTTP requests to the HTTP Client node.
+        let (request_writer, request_reader) =
+            crate::io::channel_create_with_downgrade("HTTP request", request_label)
+                .map_err(OakError::OakStatus)?;
```
tiziano88, line 39: Is this map_err necessary? I thought OakError already implements `From<OakStatus>`, or something like that.

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub request_reader: Receiver<HttpRequest>,
+    pub response_writer: Sender<HttpResponse>,
+    pub response_reader: Receiver<HttpResponse>,
+}
+
+impl Pipe {
+    pub fn new(request_label: &Label, response_label: &Label) -> Result<Self, OakError> {
+        // Create a channel for passing HTTP requests to the HTTP Client node.
+        let (request_writer, request_reader) =
+            crate::io::channel_create_with_downgrade("HTTP request", request_label)
```
tiziano88, line 38: Off the top of my head I cannot convince myself that downgrade is necessary here, could you add a comment perhaps?

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,103 @@
...
+        &self,
+        invocation_channel: &oak_io::Sender<HttpInvocation>,
+    ) -> Result<(), OakError> {
+        // Create an invocation containing request-specific channels.
+        let invocation = HttpInvocation {
+            receiver: Some(self.request_reader.clone()),
+            sender: Some(self.response_writer.clone()),
+        };
+        invocation_channel.send_with_downgrade(&invocation)
+    }
```
tiziano88, line 70: The existence of this method and the cloning it performs suggests to me that this Pipe is not the most appropriate abstraction. It seems what we really need is the dual of an Invocation, which has sender and receiver swapped. Then we could create an invocation together with its dual, send the invocation to the other node, and use the dual to read and write to the invocation channels. WDYT?

rbehjati (commented):
Thanks for the comments.

sdk/rust/oak/src/http/util.rs
```diff
@@ -0,0 +1,114 @@
...
+            http_invocation,
+            request_writer,
+            response_reader,
+        })
+    }
+
+    /// Inserts the incoming HTTP request in the `request channel` part of the `HttpInvocation`.
+    pub fn insert_message(&self, request: HttpRequest) -> Result<(), OakError> {
+        // Put the HTTP request message inside the per-invocation request channel.
+        self.request_writer.send_with_downgrade(&request)
```
tiziano88, line 57: Just checking: are all these downgrades necessary? If so, it would be good to have a comment, downgrading is really a special operation, and we should be very explicit about it now that we have separate ABI calls.

tiziano88 (approved):
Thanks!

sdk/rust/oak/src/http/client.rs
```diff
@@ -0,0 +1,84 @@
...
+        // insert the request into the request channel
+        pipe.insert_message(HttpRequest::from(request))?;
+
+        // wrap the request and response channels in an invocation, and send it to the HTTP client
+        // pseudo-node
+        pipe.send_invocation(&self.sender)?;
+
+        // wait for the response to be received
+        let rsp = pipe.get_response()?;
+
```
juliettepretot, line 78: Pardon my naiveté, I'm not super familiar with the discussions around implementing an async Oak API. 

But given that this waits for a network call, I'm surprised it's synchronous. Is it because the underlying `response_reader.receive()` call is blocking the thread?

juliettepretot (approved):
That was nice & interesting to review

sdk/rust/oak/src/http/client.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+pub struct HttpClient {
+    sender: Sender<Invocation>,
+    authority: String,
+}
+
+impl HttpClient {
+    /// Sends the given request via the HTTP client pseudo-node. The `response_label` should match
+    /// the label of the Oak Node.
```
tiziano88, line 54: There seems to be no enforcement of this "should", which means that a node may just set that to public, send a request to some endpoint that reflects back the request as-is (or returns data which is as sensitive as the request, anyways), and this would have the effect of declassifying any data labelled with this label, right?

It seems to me that the response label should just always be set to be the same of the request label, but I haven't thought about it enough yet.

sdk/rust/oak/src/http/util.rs
tiziano88: If a node wanted to send a message to an HTTP client node while downgrading, it should be very explicit about it. Definitely we should not hardcode downgrading operations "just in case" a node may need it.

sdk/rust/oak/src/http/client.rs
tiziano88: I think that makes sense actually. Imagine there is some data labelled with TLS:google.com (i.e. intended to be sent to google.com). If we allow the caller to set the label of the response, it can just set it to public, send the data to google.com, e.g. to the search page, and then the result page (which of course also contains the query itself) would be public, which is effectively a declassification! So whatever we end up doing, this kind of attack must be prevented. I think the simplest thing is to set the label of the response to be the same as the request, which intuitively means that the data is not declassified in the round trip over HTTP / gRPC. WDYT?

sdk/rust/oak/src/http/client.rs
tiziano88: Yes that is effectively a wait followed by a read, so overall blocking.

sdk/rust/oak/src/http/client.rs
tiziano88: I guess because you have removed the downgrades from Pipe now this is actually implicitly enforced, so actually setting this label to public when the HTTP client node is non-public would result in the response being lost. Is that correct?

tiziano88: (either way, this seems worth a test case, to show that an HTTP round trip may not be used to declassify data)

tiziano88: For instance in the /test_google endpoint, if it used an actual non-public label, I would expect that the response label cannot be set to public (though this cannot be tested currently because the label is public).

rbehjati: > I think the simplest thing is to set the label of the response to be the same as the request, which intuitively means that the data is not declassified in the round trip over HTTP / gRPC. WDYT?

In that case, the SDK (in `send_request`) should set the label on both the request and response channels based on the `authority` of the HTTP client node. The Oak node must be created with the same label too. So, I think in the general case it means that we need something like the router/virtual node pattern here; to create a separate Oak Node for each authority. 

> so actually setting this label to public when the HTTP client node is non-public would result in the response being lost. Is that correct?

Yes, I think if the HTTP client node is non-public, then it won't be able to write into a public channel. I noticed that we don't have a test for this. It should be added to the runtime tests in a separate PR. 

> For instance in the /test_google endpoint, if it used an actual non-public label, I would expect that the response label cannot be set to public (though this cannot be tested currently because the label is public).

Currently, the runner runs the examples with the `root_tls_certificate` set to `./examples/certs/local/ca.pem`. So, we cannot test this scenario (a non-public HTTP client with a non-empty authority) in the `http_server` example.

tiziano88: > > I think the simplest thing is to set the label of the response to be the same as the request, which intuitively means that the data is not declassified in the round trip over HTTP / gRPC. WDYT?
> 
> In that case, the SDK (in `send_request`) should set the label on both the request and response channels based on the `authority` of the HTTP client node. The Oak node must be created with the same label too. So, I think in the general case it means that we need something like the router/virtual node pattern here; to create a separate Oak Node for each authority.

I think just setting the label to be the same between request and responses fixes all these issues. If it's public, then there is no issue, they can both be public. If it has a TLS label, then the request will have that label, but also the response, which intuitively is as confidential as the request (in general). If the application handles multiple TLS labels, it needs to create multiple nodes, one per authority. That seems sensible, regardless of what we do about the response label.

> > so actually setting this label to public when the HTTP client node is non-public would result in the response being lost. Is that correct?
> 
> Yes, I think if the HTTP client node is non-public, then it won't be able to write into a public channel. I noticed that we don't have a test for this. It should be added to the runtime tests in a separate PR.

SG

> > For instance in the /test_google endpoint, if it used an actual non-public label, I would expect that the response label cannot be set to public (though this cannot be tested currently because the label is public).
> 
> Currently, the runner runs the examples with the `root_tls_certificate` set to `./examples/certs/local/ca.pem`. So, we cannot test this scenario (a non-public HTTP client with a non-empty authority) in the `http_server` example.

I don't remember the details, isn't that root TLS cert for the HTTP *server* part?

rbehjati: Created #1775 that updates the HTTP client tests in the runtime. Everything is either public or has the tls endpoint as its label. The Oak node needs to have a privilege too. Would it be able to create channels without downgrading since it is not public? I guess some of those downgrades are needed in the SDK too. 

> I don't remember the details, isn't that root TLS cert for the HTTP server part?

Does not seem like it. We have `grpc_tls_certificate` and `http_tls_certificate` for the server nodes. If `root_tls_certificate` is left empty, then the `oak_loader` uses the default root certificate. But I am not sure what that is.  Is that actually a valid certificate? Perhaps @conradgrobler knows.

conradgrobler: The "default root certificate" is actually many root certificates. It is generated from from the list of CA root certificates trusted by the Android Open Source project. See https://github.com/project-oak/oak/tree/main/oak_loader/src/certs

tiziano88: IIRC that is because we want to use TLS but connecting to servers for testing, e.g. aggregator backend. Therefore, it seems it is not possible to have both regular root CAs and custom CAs, we should just make the test use a custom local back-end over TLS instead of going out to google.com.

rbehjati: With HTTPS requests we need a domain name in the uri. So, a local backend won't work. We could perhaps use `oakexamples.dev`. But for that we'll need to have an HTTP server with a self-signed certificate on GCP, similar to what we currently have for the aggregator example with a gRPC server.

tiziano88: I think we can self-sign a cert for localhost or even 127.0.0.1, isn't that the point of providing a custom CA for testing?

conradgrobler: There is a problem with connecting to 127.0.0.1 over TLS, as the web pki crate that we are using does not support validating certificates for an IP. It should still be fine to use localhost for the connections.

tiziano88: BTW maybe we should consider making the logic add the root CA specified on the command line to the ones that we load from the file, rather than replace all of them. It may make this kind of test easier.

rbehjati: In #1779 I have replaced the request to google.com with a request to localhost, and the HTTP client pseudo-node now has a non-public label. But it is not possible to make this example work without at least one node that has the privilege to declassify responses from the HTTP client node. I think, the only node that makes sense to have that kind of privilege is a cryptographic node that encrypts and declassifies the response. So the changes in #1779 should be postponed until we have that pseudo node. Or are the any other solutions?

tiziano88: Not sure I follow the argument, is it that because a crypto node would leak some information then we can use it to bypass IFC rules?

tiziano88: (also should this discussion be on #1779 maybe?)

PR #1767, opened 2020-11-24
Create gRPC interceptor for public key auth

tiziano88:
This is not used until #1707, but it will be useful in order to test the
logic in that PR.

# Checklist

      construction.


oak_client/src/interceptors/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl<A: Interceptor, B: Interceptor> Interceptor for CombinedInterceptor<A, B> {
+    fn process(&self, request: Request<()>) -> Result<Request<()>, Status> {
+        let request = self.interceptor_a.process(request)?;
+        self.interceptor_b.process(request)
+    }
+}
+
+impl<A: Interceptor, B: Interceptor> Into<tonic::Interceptor> for CombinedInterceptor<A, B> {
```
rbehjati, line 53: Is not it possible to implement `From` here? It is advised to implement `From` instead of `Into`.

oak_client/src/interceptors/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+pub mod auth;
+pub mod label;
+
+/// Trait for interceptors, to facilitate combining until
+/// https://github.com/hyperium/tonic/issues/499 offers a better solution.
+pub trait Interceptor: Send + Sync + 'static {
+    fn process(&self, request: Request<()>) -> Result<Request<()>, Status>;
+}
+
+pub struct CombinedInterceptor<A: Interceptor, B: Interceptor> {
```
rbehjati, line 31: Perhaps add a comment? 
Looks like the order matters, and the second interceptor won't be processed if the first one returns an error (in case an interceptor has a side effect, although that is probably discouraged).

oak_client/src/interceptors/tests.rs
```diff
@@ -0,0 +1,77 @@
...
+        );
+        Ok(request)
+    }
+}
+
+#[test]
+fn test_combine_different_keys() {
+    let combined = super::combine(
+        TestInterceptor {
+            metadata_to_append: ("foo_0", "bar_0"),
```
rbehjati, line 39: How about "key" and "value" instead of "foo" and "bar"?

rbehjati (approved):
LGTM!

oak_client/src/interceptors/mod.rs
tiziano88: Good point, thanks.

oak_client/src/interceptors/mod.rs
tiziano88: No, `tonic::Interceptor` is a foreign type, so I cannot implement traits on it.

oak_client/src/interceptors/tests.rs
tiziano88: Done.

tiziano88 (commented):
Thanks!

PR #1770, opened 2020-11-25
Update Rust deps across the workspaces

tiziano88:
# Checklist

      construction.

### DIFF
--- a/examples/aggregator/backend/Cargo.toml
+++ b/examples/aggregator/backend/Cargo.toml
@@ -14,7 +14,8 @@ futures-util = "*"
 log = "*"
 prost = "*"
 structopt = "*"
-tokio = { version = "*", features = [
+# Pinned to 0.2 because of tonic: https://github.com/hyperium/tonic/blob/master/tonic/Cargo.toml
+tokio = { version = "0.2", features = [
   "fs",
   "macros",
   "signal",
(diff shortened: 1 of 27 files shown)


examples/aggregator/backend/Cargo.toml
```diff
@@ -17,6 +17,7 @@ structopt = "*"
 tokio = { version = "*", features = [
   "fs",
   "macros",
+  "rt-multi-thread",
```
conradgrobler, line 20: From the CI errors, it seems like this might be needed in a few other examples where there are rust clients using tokio.

conradgrobler (approved):

examples/aggregator/backend/Cargo.toml
```diff
@@ -14,7 +14,8 @@ futures-util = "*"
 log = "*"
 prost = "*"
 structopt = "*"
-tokio = { version = "*", features = [
+# Pinned to 0.2 because of tonic: https://github.com/hyperium/tonic/blob/master/tonic/Cargo.toml
+tokio = { version = "0.2", features = [
```
tiziano88, line 18: @conradgrobler there is an issue now if some libraries use tokio 0.2 and the tokio runner is 0.3 they will not work together and fail at runtime. Is there anything in cargo deny that can be done so that at least for this dep we stick to a single version?

conradgrobler: I think the only option is to go back to denying multiple versions and ignoring the ones that we cannot fix.

PR #1775, opened 2020-11-26
Fix labels in HTTP client tests

rbehjati:


oak_runtime/src/node/http/tests.rs
```diff
@@ -689,6 +698,9 @@ impl Node for ClientTesterNode {
         // testing.
         NodeIsolation::Sandboxed
     }
+    fn get_privilege(&self) -> NodePrivilege {
+        super::client::get_privilege(self.authority.clone())
+    }
```
tiziano88, line 703: IIUC this node is just sending requests to the actual HTTP client node, is that correct? (BTW it would be good to add a comment to this struct). If so, why does it need any privilege at all? And why the privilege of the HTTP client node itself? It does not seem representative of an actual Oak application.

tiziano88: meta-point: I am not sure it is worth re-implementing all this logic natively in Rust to try and emulate what the Wasm SDK and logic does. Would it be simpler to just use a Wasm node here?

rbehjati: The privilege is needed to be able to create a `Pipe` between this node and the HTTP client node. In the `http_example`, the problem can be solved by moving the creation of the `Pipe` to the Planner node. Since this is a test of the functionality of the HTTP client node, and not the IFC, I think it is ok to avoid introducing that kind of complexity here. I have added a comment that explains why the privilege is needed. 

However, even in the case of the `http_example`, I don't think there is a way to make the example work without a node with declassifying privilege.

rbehjati: > Would it be simpler to just use a Wasm node here?

Do we have an example that shows how a Wasm node can be used in Runtime tests? I think the test is more readable and easier to maintain if we have all the code in one place.

tiziano88 (approved):

PR #1777, opened 2020-11-26
Finish NIUtils.v proofs

aferr:
The definition of fresh_han and fresh_chan needed to be changed so
that the labels of unused objects is public. For the other definition
where labels are partially secret, I think this would need to stay
"top". Either way, fresh_han and fresh_chan are not really parts of the
implementation; in the implementation I think these objects are just
genuinely newly created objects.

There's also some fixup leftover from the last PR :) 

# Checklist

      construction.


jadephilipoom (approved):
LGTM!

Issue #1739, opened 2020-11-18
Design a solution for making use of privilege in gRPC generated code explicit


conradgrobler:
The gRPC generated code currently automatically uses the node's downgrade privilege to write the response back (in the case of server code) or send the request (in the case of client code). In order to be more explicit about when privilege is being used, we need to design a way for node implementations that based on gRPC code generation to declare whether privilege is being used or not.

tiziano88:
How many (and which ones) of the current examples rely on declassification from a gRPC handler node?

I can think of two options to approach the problem:

- augment the return type of gRPC methods to include a bool / enum of whether to use the node privilege to declassify
- expose sender / writer objects (even for unary gRPC methods) with methods for writing a message with or without declassification

conradgrobler:
There are currently two:

- Aggregator: to send a success response to the client, and to report the aggregated data via the gRPC client
- Private set intersection: to send responses to the clients

tiziano88:
Thanks, makes sense, basically those that use their Wasm declassification privilege.

Issue #1674, opened 2020-11-03
Appropriate label for the aggregator example blocks gRPC responses back to the client.


conradgrobler:
The aggregator handler node should have a label that contains both a `WebAssemblyModuleTag` and `TlsEndpointTag`. The node has the privilege to declassify the `WebAssemblyModuleTag`, but not the `TlsEndpointTag`. This is Ok for using the gRPC client pseudo node to report the aggregate results to the external service, as it would have the privilege to declassify this.

The problem is that the aggregator handler node cannot write to the response channel on the gRPC server invocation to let the client know that the data submission was successful, as it has a "public" label.

tiziano88:
**I think** (but I am not 100% sure) this is WAI.

Any response that is sent back to the client must be based on public information, or information that can be declassified directly back to the client. In particular, it should not depend on data labelled for a TLS endpoint, since that data is really only declassifiable to that TLS endpoint, no matter whether it originally came from the client. I guess in this case a disjunction label would help, but it seems too much of a loosening compared to really labelling to the TLS endpoint. I think @ipetr0v also suggested using disjunction labels for this case.

conradgrobler:
It is WAI in terms of IFC, but it means that we cannot currently implement the appropriate labels. If we implement it, the success response can not be sent back to the client and the client just keeps waiting. We need to think of a way to implement a "fire and forget" type message from the client so that it does not depend on the server responding to continue.

tiziano88:
I think the chat example has a similar (though simpler) pattern: 

https://github.com/project-oak/oak/blob/2fbd54a89589385ec007a1011b68a49112eabdd5/examples/chat/module/rust/src/lib.rs#L96-L127

It sends a message back to the caller from a node (in that case the router node) that has not seen the request data. Not sure how well this maps to the aggregator example though, I think @ipetr0v started doing a similar refactoring for the aggregator, maybe that helps.

conradgrobler:
Something similar can be done, but it would mean that the aggregator handler node would need to be reimplemented as it cannot use the SDK-generated gRPC handling code (which would try to also send back a response over the response channel).

PR #1779, opened 2020-11-27
Use a non-public HTTP client node in the `http_server` example

rbehjati:


rbehjati:
In response to https://github.com/project-oak/oak/pull/1763#discussion_r532637035: 
I think a crypto node is the only choice because the Oak nodes (Wams nodes) cannot have a privilege, and the other node we have is the HTTP client pseudo-Node, which as we discussed before should not declassify the response, and make it public.

In theory a crypto node encrypts the information, and therefore can share the encrypted information publicly. However, there is still a problem. To give the crypto node the privilege to declassify data from `google.com` or `localhost::8080` then the label it has should match that authority, but I guess it makes more sense for it to have a privilege that matches the private key it uses for the encryption. So, I am not sure even a crypto node would work!

conradgrobler:
> In theory a crypto node encrypts the information, and therefore can share the encrypted information publicly. However, there is still a problem. To give the crypto node the privilege to declassify data from `google.com` or `localhost::8080` then the label it has should match that authority, but I guess it makes more sense for it to have a privilege that matches the private key it uses for the encryption. So, I am not sure even a crypto node would work!

I don't think a crypto node resolves the issue. It would presumably only declassify the data if the data is encrypted, which means that no other node would be able to decrypt the response to send it back to the client. If the crypto node decrypts the data, it should add the old label back, so the declassification is undone.

There is currently a similar problem with the gRPC client code (see https://github.com/project-oak/oak/issues/1739 and https://github.com/project-oak/oak/issues/1674). The only reason that the aggregator example works at the moment is because we don't give it the appropriate label and automatically use the privilege. The data is only labeled with the hash of the Wasm aggregator node (rather than the conjunction of the hash and the TLS endpoint for the external gRPC service). The Wasm node uses its Wasm hash privilege automatically to create the invocation channels for the gRPC client node and to send success responses back to the client app.

tiziano88:
Let's take a step back: what is it that you are trying to test exactly?

rbehjati:
@conradgrobler Yes. gRPC client node has the same problem, but I think the problem with the `http_server` is even simpler than the one in the aggregator example (#1674). I am not sure how disjunctions would solve the problem. We need a node that can declassify and remove `TlsEndpointTag`. I am not sure what kind of node is a good candidate for doing such a declassification.

@tiziano88 Currently, the `http_server` example uses a public HTTP client pseudo-node. So, everything works fine. Following the discussions on #1763 (https://github.com/project-oak/oak/pull/1763#discussion_r530724584), I am trying to change it to a non-public HTTP client that handles only HTTPS requests sent to `localhost:8080`. The problem is that the Oak node that sends the requests to the HTTP client pseudo-node is either unable to read the response, or unable to send the response to any other node (and the outside world, for example via an HTTP server pseudo-node). This is happening because none of the nodes can declassify the data.

tiziano88:
> @conradgrobler Yes. gRPC client node has the same problem, but I think the problem with the `http_server` is even simpler than the one in the aggregator example (#1674). I am not sure how disjunctions would solve the problem. We need a node that can declassify and remove `TlsEndpointTag`. I am not sure what kind of node is a good candidate for doing such a declassification.
> 
> @tiziano88 Currently, the `http_server` example uses a public HTTP client pseudo-node. So, everything works fine. Following the discussions on #1763 ([#1763 (comment)](https://github.com/project-oak/oak/pull/1763#discussion_r530724584)), I am trying to change it to a non-public HTTP client that handles only HTTPS requests sent to `localhost:8080`. The problem is that the Oak node that sends the requests to the HTTP client pseudo-node is either unable to read the response, or unable to send the response to any other node (and the outside world, for example via an HTTP server pseudo-node). This is happening because none of the nodes can declassify the data.

This is expected, it will not be able to declassify the response, since the response may contain data which is as classified as the request. In debug mode, it may be able to log it, because the log node is top secret, but in "production" it will just deal with the response, but not be able to declassify to public. I don't think trying to rely on side channels (e.g. from a crypto node) is a satisfactory solution to this (in fact that is something we should try and prevent as much as possible when implementing such node).

tiziano88:
Just to clarify intuitively why that must be the case: imagine some data labelled with "TLS:google.com", e.g. a query string from a user. The expectation is that it can only be sent to google.com. If we "forget" about the "TLS:google.com" label on the response, then it is trivially possible to declassify any such data just by sending it to the google.com search page, get the search results (which include the query), and extract the original data from there. Therefore the label didn't actually make the data only flow to google.com, it made it flow to public.

The expectation is that a response to an HTTP request is as confidential as the request.

examples/http_server/module/src/lib.rs
```diff
@@ -80,28 +181,166 @@ impl CommandHandler for StaticHttpHandler {
...
+            invocation_sender: init.invocation_sender.unwrap(),
+        }
+    }
+}
+
+impl CommandHandler for RedirectHandler {
+    type Command = Invocation;
+
+    fn handle_command(&mut self, invocation: Invocation) -> anyhow::Result<()> {
+        let request = invocation.clone().receiver.unwrap().receive()?;
```
conradgrobler, line 228: I thought the intention was that the client will send a request with a TLS endpoint label. At the moment, the redirect handler is public. If the client sends a redirect request with a TLS endpoint label, the redirect handler will not be able to receive the request on this channel and it will fail here.

I think the public handler that receives the invocation should split the invocation. It should send the request receiver channel handle to a node with the right TLS endpoint label (where the request inside the channel can be received)m together with the new invocation channel for the HTTP client node request.

The public node should respond immediately with "success" on the public response channel, otherwise the client will hang as it will never receive a response. The public node will never know whether the redirect request was successful, so can just always return success.

ipetr0v (approved):

examples/http_server/module/src/lib.rs
rbehjati: I have pushed a change that splits the invocation (if I understood correctly). 

For a non-public request label to work, I think we have to use the router pattern. Basically, the `StaticHttpHandler` should create a separate node for each request label. Is that right? Did not we earlier say that only the planner/Main node should create new nodes?

conradgrobler: I don't hink we need a full router pattern. The router pattern usually creates one node per label. As all requests will have the same label, one node with the non-public label should be enough.

But there should still be a public node that works a bit like a router node, without creating a new node for every request:

- It should receive the invocations from the HTTP server
- it should create 2 channels for the http client node invocation
- it should send the invocation to the http client node
- It should send the following to the non-public node
  - the request reader channel from the server invocation
  - the other sides of the channels for the invocation that was sent to the http client node (request writer and response reader)
- it should respond with a success message on the response writer it received in the server invocation and close this response writer

The non-public node should then:

- Read the request from the server invocation using the request receiver it received
- Translate that request into a request that it writes on the request writer for the client invocation
- Wait for a response from the client using the response reader from the client invocation
- Read the response (and ignore it)
- Close the response reader

Nodes can only be created by public nodes at the moment, so either the initial node can create all the nodes and channels that are needed, or it can create a dedicated planner node with a public label which then creates the other nodes and all the channels used for the invocations.

oak_services/proto/http_invocation.proto
```diff
@@ -33,3 +33,9 @@ message HttpInvocation {
 message HttpInvocationSender {
   oak.handle.Sender sender = 1 [(oak.handle.message_type) = ".oak.invocation.HttpInvocation"];
 }
+
+message HttpPipe {
+  HttpInvocation http_invocation = 1;
+  oak.handle.Sender request_sender = 2 [(oak.handle.message_type) = ".oak.encap.HttpRequest"];
+  oak.handle.Receiver response_receiver = 3 [(oak.handle.message_type) = ".oak.encap.HttpResponse"];
```
tiziano88, line 40: This looks like the "dual invocation" that I mentioned a while ago; it is basically an invocation in which sender and receiver are swapped. I think that should be its own type though, and have a method that creates both ends of the invocation and return them to the caller. Maybe a refactoring for the future anyways (especially as I cannot think of a good name for this object -- maybe `Invocation` vs `InvocationProducer`?)

sdk/rust/oak/src/http/util.rs
```diff
@@ -89,6 +94,30 @@ impl Pipe {
     }
 }
 
+impl Into<HttpPipe> for Pipe {
```
tiziano88, line 97: I must say I am getting confused with all these pipe-like structs :) I guess they are necessary, but we definitely need how to unify or make them more understandable at some point.

oak_runtime/src/node/http/server.rs
```diff
@@ -555,6 +555,7 @@ impl HttpResponseReceiver {
         let response_receiver = crate::io::Receiver::<HttpResponse>::new(ReadHandle {
             handle: self.response_reader,
         });
+        info!("Trying to read the response...");
```
tiziano88, line 558: Just checking: are these logs intentional?

examples/http_server/module/src/lib.rs
```diff
@@ -80,28 +176,150 @@ impl CommandHandler for StaticHttpHandler {
...
+        // Send the request-receiver channel, together with the new invocation source for
+        // communication with the HTTP client pseudo-Node, to the RedirectHelper node in a
+        // RedirectInvocation.
+        let helper_invocation = RedirectInvocation {
+            request_receiver: Some(invocation.clone().receiver.unwrap()),
+            http_invocation_source: Some(HttpInvocationSource {
+                sender: Some(request_writer),
+                receiver: Some(response_reader),
+            }),
+        };
```
tiziano88, line 242: This pattern is what I was talking about in another comment: now we could have a single method (with 2 generic type params) that creates an invocation (2 channels) and returns both "sides" of the invocation (the invocation itself and the "source"). Perhaps a refactoring for another PR, but could you create an issue to track?

tiziano88 (approved):

conradgrobler (approved):

examples/http_server/module/src/lib.rs
rbehjati: Created #1832.

oak_runtime/src/node/http/server.rs
rbehjati: This was intentional, but perhaps not quite consistent with the rest of the logs. Removed.

PR #1781, opened 2020-11-29
Finish Unwind.v except set_call_unwind

aferr:
I think set_call_unwind needs to be weakened.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/Unwind.v
+++ b/experimental/ni_coq/vfiles/Unwind.v
@@ -9,7 +9,8 @@ From OakIFC Require Import
     ModelSemUtils
     LowEquivalences
     NIUtilTheorems
-    Tactics.
+    Tactics
+    Unfold.
 From RecordUpdate Require Import RecordSet.
 Import RecordSetNotations.
 Local Open Scope map_scope.
@@ -60,6 +61,28 @@ Proof.
     simpl in *. erewrite upd_neq; auto. erewrite upd_neq; auto.
 Qed.
 
+Ltac label_cases :=
+    match goal with
+        | l1: level, l2: level, ell: level |- _ =>
+            (* l1 not found in environment here ?? *)
+            (* (if l1 <<? ell then _ else _) = 
+            (if l2 <<? ell then _ else _) => *)
+            destruct (l1 <<? ell), (l2 <<? ell)
+    end.
+
+Ltac unwind_crush_step2 :=
+    lazymatch goal with
+        | H: {| obj := _; lbl := _ |} = _ |- _ => inversion H; subst
+        | H: Some _ = Some _ |- _ => inversion H
+        | _ => idtac
+    end.
+
+Ltac unwind_crush :=
+    repeat match goal with
+        (* | _ => progress label_cases *) (* this causes coq to get stuck *)
+        | _ => progress unwind_crush_step2
+    end.
+
 Theorem chan_append_unwind: forall ell ch1 ch2 ch1obj ch2obj msg,
     chan_low_eq ell ch1 ch2 ->
     ch1.(obj) = Some ch1obj ->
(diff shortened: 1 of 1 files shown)


experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -68,43 +91,114 @@ Theorem chan_append_unwind: forall ell ch1 ch2 ch1obj ch2obj msg,
...
+        specialize (H0 han). autounfold with loweq in *.
+        destruct (chans han), (chans0 han). simpl in *. 
+        destruct (lbl <<? ell), (lbl0 <<? ell) eqn:E.
+        all: unwind_crush. 
+        all: try congruence.
+        (* not sure what happened here ...*)
+        destruct obj; destruct obj0; (rewrite E; auto).
+    * (* <> *)
+    subst. erewrite upd_neq. erewrite upd_neq.
+    all: eauto.
```
jadephilipoom, line 194: Very nitpicky, but since the `<>` case is pretty short, I'd remove the need for two cases here by doing:
```coq
destruct (dec_eq_h han0 han); [ | subst; erewrite !upd_neq by eauto; eauto ].
```

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -60,6 +61,28 @@ Proof.
     simpl in *. erewrite upd_neq; auto. erewrite upd_neq; auto.
 Qed.
 
+Ltac label_cases :=
+    match goal with
+        | l1: level, l2: level, ell: level |- _ =>
+            (* l1 not found in environment here ?? *)
```
jadephilipoom, line 67: Try `Set Printing All` -- is there maybe a coercion around `l1`?

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -60,6 +61,28 @@ Proof.
...
+            (* (if l1 <<? ell then _ else _) = 
+            (if l2 <<? ell then _ else _) => *)
+            destruct (l1 <<? ell), (l2 <<? ell)
+    end.
```
jadephilipoom, line 71: Maybe you were working around the problem you mention in the comment here, but if what you want is just a thing that destructs `<<?`, I'd rewrite it as 
```coq
Ltac label_cases :=
    lazymatch goal with
    | l: level, ell: level |- context [l <<? ell] =>
            destruct (l <<? ell)
    end.
```
...so that it works when you have only one level equality check in context, and doesn't keep destructing when the check has already been destructed.

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -60,6 +61,28 @@ Proof.
...
+Ltac unwind_crush_step2 :=
+    lazymatch goal with
+        | H: {| obj := _; lbl := _ |} = _ |- _ => inversion H; subst
+        | H: Some _ = Some _ |- _ => inversion H
+        | _ => idtac
+    end.
+
+Ltac unwind_crush :=
+    repeat match goal with
+        (* | _ => progress label_cases *) (* this causes coq to get stuck *)
```
jadephilipoom, line 82: Does it still get stuck if there's a `|- context [l <<? ell]` check in `label_cases`? Maybe you're getting repeated `destruct`s on the same equality.

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -68,43 +91,114 @@ Theorem chan_append_unwind: forall ell ch1 ch2 ch1obj ch2obj msg,
...
+    unfold state_low_eq, state_upd_node_labeled.
+    intros; split; logical_simplify.
+    Focus 2. auto.
+    destruct s1, s2. cbv [RecordSet.set State.nodes] in *. simpl.
+    intros. do 2 rewrite node_state_proj_index_assoc.
+    destruct (dec_eq_nid id nid). 
+        * (* = *)
+            subst. do 2 erewrite upd_eq. eauto. 
+        * (* <> *)
+            erewrite upd_neq. erewrite upd_neq. all: eauto.
```
jadephilipoom, line 149:
```suggestion
            erewrite !upd_neq. all: eauto.
```

experimental/ni_coq/vfiles/Unwind.v
```diff
@@ -68,43 +91,114 @@ Theorem chan_append_unwind: forall ell ch1 ch2 ch1obj ch2obj msg,
...
-Admitted.
+    unfold state_low_eq, state_upd_chan_labeled.
+    intros; split; logical_simplify. auto.
+    destruct s1, s2. cbv [RecordSet.set State.chans] in *. simpl.
+    intros. do 2 rewrite chan_state_proj_index_assoc.
+    destruct (dec_eq_h h han). 
+        * (* = *)
+        subst. do 2 erewrite upd_eq. eauto.
+        * (* <> *)
+        erewrite upd_neq. erewrite upd_neq. all: eauto.
```
jadephilipoom, line 166:
```suggestion
        erewrite !upd_neq. all: eauto.
```

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/Unwind.v
aferr: That does make sense as to why it wouldn't terminate. Hmmm.. I also got an error that l is not found in the environment around the "context [l ... ]" part. I tried also doing "Set Printing All" like you suggested, but I don't get any new info in the IDE before the error. If I try cutting out the part that throws the error I also don't see any more info. If I try replacing "l" with "_" I get an error that "ell" is not found in the environment instead. I don't think there should be a coercion with "ell", since that's used as a variable name often. 

Does coq usually let you bind a variable | here and then let you use it |- here? I know you can use it => here. (Apologies if that's a confusing way to ask that question).

jadephilipoom: Ah, sorry, I wasn't reading closely enough. No, Coq usually doesn't let you reference a variable in quite this way; you have to get around it with a double match. Something like
```
Ltac label_cases :=
    match goal with
        | l1: level, l2: level, ell: level |- _ =>
            lazymatch goal with
            | (if l1 <<? ell then _ else _) = 
              (if l2 <<? ell then _ else _) =>
              destruct (l1 <<? ell), (l2 <<? ell)
           end
    end.
```

aferr: That was it! Awesome, thanks.

Issue #1731, opened 2020-11-16
Split Aggregator Wasm module


ipetr0v:
After finishing https://github.com/project-oak/oak/pull/1646 we will start checking labels of the incoming messages (if they have Aggregator Wasm module hash). But in order to do this, we need this hash to be present in the Aggregator code.

Currently this problem is solved via sending hash as a part of `ConfigMap`, but in the future we need to split these modules into `Router` and `Handler` modules, and hardcode the `Handler`'s hash in the `Router`.

ipetr0v:
The only problem with splitting modules and hardcoding SHA-256 is that Oak tests also compile Wasm modules which result in different SHA-256 hashes.

ipetr0v:
I order to allow both tests and release versions to use the same module (but with different hashes) and also prevent constant re-hardcoding of module hashes - I can add the following function to ABI:
```rust
// `oak::io`
fn module_privilege_read(wasm_module_name: &str) -> Result<Label, OakStatus> {
    ...
    oak_abi::module_privilege_read(...)
    ...
}

```
Similar to existing `node_privilege_read`.

This will also work for Signature labels.

WDYT @tiziano88 ?

tiziano88:
I'm not sure I see how this helps. The problem is the same from the client point of view: the client needs to know the label to assign upfront. So I don't think having a special case to solve it in an Oak application helps much. Or am I missing something?

ipetr0v:
The main problem is - the same module is compiled for release and for tests, resulting in having different hashes. This is problematic if we want to hardcode hashes in `Main` modules.
Second problem (more of an inconvenience) we always need to re-hardcode hashes once we updated an application. This also could result in the avalanche of changes if multiple modules use such hardcoded hashes.

tiziano88:
> The main problem is - the same module is compiled for release and for tests, resulting in having different hashes. This is problematic if we want to hardcode hashes in `Main` modules.

Maybe we should only pick one. The point is that they are in fact different modules, so I don't think we should try and work around this.

> Second problem (more of an inconvenience) - we always need to re-hardcode hashes once we updated an application. This also could result in the avalanche of changes if multiple modules use such hardcoded hashes.

As I said, this is the same problem that a client using these labels would have, so solving it for ourselves is actually counterproductive. If we cannot solve it for ourselves, how do we expect clients to deal with it?

PR #1783, opened 2020-12-01
Update Wasm module generation in oak_tests

ipetr0v:
This change updates Wasm module generation in `oak_tests` in order to generate the same modules as in `runner` (i.e. now they have same SHA256 hashes).

Ref https://github.com/project-oak/oak/issues/1731

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -191,9 +191,14 @@ fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
                     "--target=wasm32-unknown-unknown",
                     &format!("--manifest-path={}", cargo_manifest),
                     // Use a fixed target directory, because `--target-dir` influences SHA256 hash
-                    // of Wasm module. This directory should also be synchronized with
+                    // of Wasm module.  Target directory should also be synchronized with
                     // `--target-dir` used in [`oak_tests::compile_rust_wasm`] in order to have
                     // same SHA256 hashes.
+                    //
+                    // This directory is separate from `examples/target` because it is used by
+                    // `cargo test`, which also executes [`oak_tests::compile_rust_wasm`] and thus
+                    // runs `cargo build` inside it. It may lead to errors, since dependencies may
+                    // be recompiled by `cargo build` and `cargo test` will fail to continue.
                     &format!("--target-dir={}", {
                         let mut target_dir = PathBuf::from(cargo_manifest);
                         target_dir.pop();
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -190,6 +190,16 @@ fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
                     "--release",
                     "--target=wasm32-unknown-unknown",
                     &format!("--manifest-path={}", cargo_manifest),
+                    // Use a fixed target directory, because `--target-dir` influences SHA256 hash
+                    // of Wasm module. This directory should also be synchronized with
+                    // `--target-dir` used in [`oak_tests::compile_rust_wasm`] in order to have
+                    // same SHA256 hashes.
+                    &format!("--target-dir={}", {
+                        let mut target_dir = PathBuf::from(cargo_manifest);
+                        target_dir.pop();
+                        target_dir.push("target");
+                        target_dir.to_str().expect("Invalid target dir").to_string()
+                    }),
                     &format!("--out-dir=examples/{}/bin", example_name),
                 ],
             ),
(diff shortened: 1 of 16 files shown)


ipetr0v:
I didn't make `oak_tests` and `runner` use the same function, because:
- `runner` shouldn't be included in any module
- `runner` uses `Step` objects and a complex infrastructure for running commands

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -46,42 +47,35 @@ pub fn compile_rust_wasm(
...
+        format!("--out-dir={}", out_dir),
     ];
-    let profile_str = match profile {
-        Profile::Release => {
-            args.push("--release".to_string());
-            "release".to_string()
-        }
-        Profile::Debug => "debug".to_string(),
-    };
+    if profile == Profile::Release {
```
tiziano88, line 62: Should these be kept in sync with runner? if so, could you add a comment here and also in runner?

tiziano88 (approved):

sdk/rust/oak_tests/src/lib.rs
ipetr0v: Added a comment

ipetr0v:
Redone the PR: now both test and `runner` versions are compiled in `examples/{}/module/rust/target`

ipetr0v:
Also temporary removed external modules to check if CI Docker compiles with the same SHA256 hashes.

tiziano88:
> Also temporarily removed external modules to check if CI Docker compiles with the same SHA256 hashes.

Great! if it works, please leave it with the local files instead of downloading modules from GCP every time.

runner/src/main.rs
```diff
@@ -190,6 +190,16 @@ fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
...
                     &format!("--manifest-path={}", cargo_manifest),
+                    // Use a fixed target directory, because `--target-dir` influences SHA256 hash
+                    // of Wasm module. This directory should also be synchronized with
+                    // `--target-dir` used in [`oak_tests::compile_rust_wasm`] in order to have
+                    // same SHA256 hashes.
+                    &format!("--target-dir={}", {
+                        let mut target_dir = PathBuf::from(cargo_manifest);
+                        target_dir.pop();
+                        target_dir.push("target");
+                        target_dir.to_str().expect("Invalid target dir").to_string()
```
tiziano88, line 201: Isn't this the same as not specifying any target dir?

ipetr0v: Without specifying the `--target-dir` it would be `examples/target` and not `examples/{}/module/rust/target`

tiziano88: But wouldn't it be the same for both runner and tests, at least?

ipetr0v: If we will use `examples/target` then `integration_test.rs` also will be occupy the same directory with Wasm modules.
It leads to errors because we are running `cargo build` inside `cargo test`, and they are using the same target directory.

tiziano88: Got it, that seems the most important thing to include in the comment then, isn't it? (sorry if I missed it)

ipetr0v:
Doesn't look like CI is building the same Wasm modules:
https://pantheon.corp.google.com/cloud-build/builds/0bfa8477-c1bf-48f6-9f60-d49f6f7e46da?project=oak-ci

tiziano88:
> Doesn't look like CI is building the same Wasm modules:
> https://pantheon.corp.google.com/cloud-build/builds/0bfa8477-c1bf-48f6-9f60-d49f6f7e46da?project=oak-ci

Do either of them match the hashes in https://github.com/project-oak/oak/blob/hashes/reproducibility_index ?

ipetr0v:
Hm, interesing observation:
`aggregator.wasm` hash is the same on both of my machines:
```
53d38904ac029db2f25d8a976ee4bf9060cd01c80e48e2c10d9cfcbedbec50de
```
But `private_set_intersection_handler.wasm` hash is different:
```
08ac29c2e8fc7de860750457f1e97decc3f0089711e5efd574453085706b0a42
```
vs
```
67e9f599357de7b81f775116b05645f79bda4f8647e0f240300b6dde463707b2
```

ipetr0v:
And the only difference between these files is:
```
--- (func $_ZN4core3ptr13drop_in_place17h05792fa346fda189E.llvm.1918291159313483383 (type 1) (param i32)
+++ (func $_ZN4core3ptr13drop_in_place17h05792fa346fda189E.llvm.17173589550728216195 (type 1) (param i32)
```
Which is just duplicated in different places.

runner/src/main.rs
ipetr0v: Done

PR #1784, opened 2020-12-01
Finish set_call_unwind (And NI proof!)

aferr:
The proof of the last theorem is clearly a mess, though.

# Checklist

      construction.

### DIFF
--- a/experimental/ni_coq/vfiles/Unwind.v
+++ b/experimental/ni_coq/vfiles/Unwind.v
@@ -191,12 +191,73 @@ Proof.
         destruct obj; destruct obj0; (rewrite E; auto).
 Qed.
 
+Theorem set_call_none_is_nop: forall s id c,
+    obj ((nodes s) id) = None ->
+    s_set_call s id c = s.
+Proof.
+    autounfold with semutils. unfold fnd. intros.
+    (* why doesn't subst work here? sad/fake_news. *)
+    rewrite H. congruence.
+Qed.
+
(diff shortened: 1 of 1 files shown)


jadephilipoom (approved):
LGTM! But I see only `set_call_unwind` and a couple small lemmas -- did you mean to include an NI proof, or did you mean that `set_call_unwind` is the last admitted lemma for the NI proof?

aferr:
This WAS the last admitted theorem in the first iteration of the proof. (There are some admitted theorems in LowProjPS_Thms.v but those are not used in the main proof. That file is meant to be useful in case we adopt 
[issue 1722](https://github.com/project-oak/oak/issues/1722) )

There's still more to do with these proofs, though. There are ways to make the model better. I probably should also do more of a thorough comparison of this and the implementation again. There are also better variants of NI we could be proving.

aferr:
Tactics and cleaner proofs would still help make the rest of the proofs easier.

jadephilipoom:
I see. Pretty big milestone, then -- congratulations!

If you haven't found it already, you can use `Print Assumptions` to see any admits that a lemma depends on.

aferr:
Oh nice I haven't seen that. I ran `Print Assumptions possibilistic_ni` with so much anxiety! I just got:
```
Axioms:
Parameters.p : params
Parameters.lat : Lattice level
```

which I think are reasonable, still.

jadephilipoom:
Yep, seems like those are just parameters that you've quantified over.

PR #1785, opened 2020-12-01
Enable colorized output in GitHub actions logs

tiziano88:
The current logic is not quite right; the thing that is missing from
GitHub action is the interactive STDIN, not the TTY itself. Fixing this
makes most commands detect the TTY correctly, which produces colored
outputs in logs etc.


juliettepretot (approved):
LGTM :)

PR #1789, opened 2020-12-02
Split Aggregator example

ipetr0v:
This change:
- Splits Aggregator Wasm module into 2 new modules: Main and Handler
- Removes sending SHA256 hash as `ConfigMap`

Fixes https://github.com/project-oak/oak/issues/1731


examples/aggregator/main_module/rust/tests/integration_test.rs
```diff
@@ -51,23 +64,15 @@ async fn submit_sample(
...
-    let module_config = format!(
-        "grpc_server_listen_address = \"[::]:8080\"\n\
-         backend_server_address = \"https://localhost:8888\"\n\
-         aggregator_module_hash = \"{}\"",
-        hex::encode(&wasm_module_hash)
-    );
+    let module_config = "\
+        grpc_server_listen_address = \"[::]:8080\"\n\
+        backend_server_address = \"https://localhost:8888\"\n\
+    ";
```
tiziano88, line 70: could you use a raw string here?

examples/aggregator/example.toml
```diff
@@ -11,7 +11,8 @@ manifest = "examples/aggregator/oak_app_manifest.toml"
 out = "examples/aggregator/bin/aggregator.oak"
 
 [applications.rust.modules]
-module = { Cargo = { cargo_manifest = "examples/aggregator/module/rust/Cargo.toml" } }
+main_module = { Cargo = { cargo_manifest = "examples/aggregator/main_module/rust/Cargo.toml" } }
+handler_module = { Cargo = { cargo_manifest = "examples/aggregator/handler_module/rust/Cargo.toml" } }
```
tiziano88, line 15: I don't think having the extra /rust/ in the path is particularly useful, can we just remove them?

ipetr0v:
Also temporarily checking if `aggregator` could be compiled in CI with the same hash.

tiziano88 (approved):

examples/aggregator/example.toml
ipetr0v: Done

examples/aggregator/main_module/rust/tests/integration_test.rs
ipetr0v: Done

examples/aggregator/main_module/tests/integration_test.rs
```diff
@@ -14,26 +14,43 @@
...
+const HANDLER_MODULE_MANIFEST: &str = "../handler_module/Cargo.toml";
+
+const MAIN_MODULE_NAME: &str = "app";
+const HANDLER_MODULE_NAME: &str = "handler";
 const ENTRYPOINT_NAME: &str = "oak_main";
 
+// SHA-256 hash of Handler module.
+// TODO(865): Remove and compute SHA-256 hash once reproducibility is fixed.
+const HANDLER_SHA256_HASH: &str =
+    "93539428d76d71d1812267013880b0aafcb40cbcc42b49d5dca511422a33690f";
```
tiziano88, line 45: Do we have a test with the wrong hash label?

tiziano88 (approved):

examples/aggregator/main_module/tests/integration_test.rs
ipetr0v: I just have sent it to test if CI for aggregator produces the same hash.
But it looks like it does.
I think we need to fix reproducibility before submitting this PR

examples/aggregator/handler_module/src/lib.rs
```diff
@@ -14,6 +14,30 @@
 // limitations under the License.
 //
 
+//! Aggregator example for Project Oak.
```
rbehjati, line 17: Is not it better to have this documentation in the `main_module`?

rbehjati (approved):
LGTM!

examples/aggregator/handler_module/src/lib.rs
ipetr0v: Done

PR #1790, opened 2020-12-03
Prevent false positives for pages whose CSP cannot be checked

juliettepretot:
# Checklist

      construction.

### DIFF
--- a/chrome_extension/background.js
+++ b/chrome_extension/background.js
@@ -19,10 +19,18 @@
 const GREEN_ICON_PATH = './icon-green.png';
 const RED_ICON_PATH = './icon-red.png';
 
-const isUsingDesiredCspPerTab = new Map();
+// Map of tabs that are applying the desired CSP. The key is the `tabId`, the
+// value is the specific `url` that applies them. Keeping track of the URL is
+// necessary since administrators can block access to the `chrome.webRequest`
+// API used to check CSP on a per page basis. Without checking that the URL
+// matches, a user could navigate from a secure page to a potentially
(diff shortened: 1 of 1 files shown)


chrome_extension/background.js
```diff
@@ -19,21 +19,31 @@
 const GREEN_ICON_PATH = './icon-green.png';
 const RED_ICON_PATH = './icon-red.png';
 
-const isUsingDesiredCspPerTab = new Map();
+// Map of tabs that are applying the desired CSP. The key is the `tabId`, the
+// value is the specific `url` that applies them. Keeping track of the URL is
+// necessary since administrators can block access to the `chrome.webRequest`
+// API used to check CSP on a per page basis. Without checking that the URL
+// matches, a user could navigate from a secure page to a potentially
```
conradgrobler, line 26: How does the user navigate to a different page?

Is it only if the user types a URL in the address bar, or can a user click on a link to also navigate or is this blocked by the CSP?

conradgrobler (approved):
Interesting case.

chrome_extension/background.js
juliettepretot: For this case form of navigation works, clicking a bookmark, typing a URL etc. As long as it is in the same tab.

tiziano88 (approved):

chrome_extension/background.js
juliettepretot: RE disabling in page navigation, it prevents some, unfortunately not all. I created an issue to tackle this: https://github.com/project-oak/oak/issues/1792

Issue #1795, opened 2020-12-04
Write a simple handler node


rbehjati:
Similar to what we already have in `fetch`.

PR #1799, opened 2020-12-07
Add empty Treehouse Oak Application

ipetr0v:
Ref https://github.com/project-oak/oak/issues/1795

### DIFF
--- a/experimental/Cargo.lock
+++ b/experimental/Cargo.lock
@@ -2126,6 +2126,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "treehouse"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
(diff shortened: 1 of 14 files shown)


rbehjati:
Since the example is added to `experimental` can we run it using the `run-examples` command? Or it would not be possible to run it with that command anyways?

ipetr0v:
No, only using:
```bash
cargo build --manifest-path=experimental/treehouse/application/module/Cargo.toml
```

ipetr0v:
Ah, no
Looks like `runner` can run it without any changes:
```bash
clear; ./scripts/runner --logs run-examples --example-name=treehouse
```

rbehjati:
Interesting!

PR #1819, opened 2020-12-10
Fix Wasm module reproducibility

ipetr0v:
This change:
- Prevents Docker from depending on the username
- Makes Wasm modules reproducibly buildable inside Docker on different machines (by making rust library paths have the same user)

Fixes https://github.com/project-oak/oak/issues/865

### DIFF
--- a/scripts/docker_build
+++ b/scripts/docker_build
@@ -11,12 +11,10 @@ source "$SCRIPTS_DIR/common"
 # https://github.com/googleapis/google-cloud-cpp/blob/a186208b79d900b4ec71c6f9df3acf7638f01dc6/ci/kokoro/docker/build.sh#L147-L152
 readonly DOCKER_UID="${UID:-0}"
 readonly DOCKER_GID="$(id -g)"
-readonly DOCKER_USER="${USER:-root}"
 
 docker build \
   --cache-from="$DOCKER_IMAGE_NAME:latest" \
   --tag="$DOCKER_IMAGE_NAME:latest" \
-  --build-arg=USERNAME="$DOCKER_USER" \
   --build-arg=USER_UID="$DOCKER_UID" \
   --build-arg=USER_GID="$DOCKER_GID" \
   . 1>&2
--- a/Dockerfile
+++ b/Dockerfile
@@ -281,9 +281,12 @@ RUN make depend && make -j"$(nproc)"&& make install_sw install_ssldirs
 ENV PKG_CONFIG_ALLOW_CROSS 1
 ENV OPENSSL_STATIC 1
 
+# We use the `docker` user in order to maintain library paths on different
+# machines and to make Wasm modules reproducible.
+ARG USERNAME=docker
+
 # Placeholder args that are expected to be passed in at image build time.
 # See https://code.visualstudio.com/docs/remote/containers-advanced#_creating-a-nonroot-user
-ARG USERNAME=user-name-goes-here
 ARG USER_UID=1000
 ARG USER_GID=${USER_UID}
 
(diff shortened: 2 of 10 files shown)


tiziano88:
@ipetr0v if the issue is only on GCB (and not on GH actions) , should we just delete the GCB pipeline? I think keeping only one makes sense long term anyways

tiziano88:
> @ipetr0v if the issue is only on GCB (and not on GH actions) , should we just delete the GCB pipeline? I think keeping only one makes sense long term anyways

@ipetr0v ping

ipetr0v:
It looks like GH actions also hang and don't succeed

ipetr0v:
So, github CI can reproducibly build Wasm modules now.
Turned out that changing Docker users and adjusting file permissions affects github CI, because it doesn't run Docker containers as a `root` but as a `runner` instead.

So in order to have a reproducible CI - we need to drop GCB and only use github CI.

ipetr0v:
But IIUC it's only possible to remove GCB simultaneously for all PRs.

tiziano88:
Is the problem with GCB that it still runs the containers as root?

ipetr0v:
Yes
And making it run as a non-root user still adds permission problems

tiziano88:
> But IIUC it's only possible to remove GCB simultaneously for all PRs.

Yes I can change the repo settings to make the GH actions check mandatory, but it will be a global change.

I think that's fine though, as long as we are fine with GH actions as our primary CI. Is there anything left that is not currently checked on GH actions? I think the git diff stuff may not be there. Is GH actions stable enough, or is it more flaky than GCB, in your experience?

ipetr0v:
> Is GH actions stable enough, or is it more flaky than GCB, in your experience?

To be honest GH actions do seem more flaky than GCB.

tiziano88:
> > Is GH actions stable enough, or is it more flaky than GCB, in your experience?
> 
> To be honest GH actions do seem more flaky than GCB.

it's true, do you have an idea why?

ipetr0v:
Not sure
Sometimes I've seen that it showed `This check failed` without logs.

ipetr0v:
Created a PR that updates github CI to match GCB: https://github.com/project-oak/oak/pull/1888
So we could disable GCB in favor of reproducibility

tiziano88 (approved):

scripts/docker_build
```diff
@@ -11,7 +11,9 @@ source "$SCRIPTS_DIR/common"
 # https://github.com/googleapis/google-cloud-cpp/blob/a186208b79d900b4ec71c6f9df3acf7638f01dc6/ci/kokoro/docker/build.sh#L147-L152
 readonly DOCKER_UID="${UID:-0}"
 readonly DOCKER_GID="$(id -g)"
-readonly DOCKER_USER="${USER:-root}"
+# We use the `oak` Docker user in order to maintain library paths on different
+# machines and to make Wasm modules reproducible.
+readonly DOCKER_USER="oak"
```
tiziano88, line 16: Maybe `'docker'` would be a more common value for the username.

tiziano88 (approved):

scripts/docker_build
tiziano88: In fact, why keep this variable if it's always constant? I think we can remove this and also the corresponding ARG from the Docker image, right?

tiziano88 (approved):

scripts/docker_build
ipetr0v: Agree
But why removing the ARG from Dockerfile? We could just change it to `ARG USERNAME=docker`
https://github.com/project-oak/oak/blob/68b4de822f56b40beab4d498e3a778ff98940191/Dockerfile#L286-L295

tiziano88: Yeah makes sense, it is already used in a couple of places.

ipetr0v: Done

Issue #565, opened 2020-02-06
Implement ECDSA / EdDSA authentication and labelling system


tiziano88:
To replace the current bearer token mechanism.

A possible implementation would be to extend the `GrpcTag` message with a `ecdsa_public_key` field, which clients may use to label data with the literal public key of the intended recipient(s).

https://github.com/project-oak/oak/blob/1528e03883bf23e727f858e1342f335ddf041b06/oak/proto/policy.proto#L44-L57

On the other side, the gRPC pseudo-node would authenticate users by requiring a proof of possession of the corresponding private key, in order to be able to declassify data labelled with a public key.

Note that the private key never leaves the client device, and in fact it may be desirable to perform the signature scheme in hardware (via a TPM or secure element) as much as possible, for production use cases.

Note that we already have a dependency on `boringssl`, so we can leverage that for any crypto-related operations.

Also note that Asylo may support a native way of authenticating users based on PKI at some point, but it's unclear how / when.

For the time being, we may rely on a non-interactive proof of possession of the private key, for instance by having the client sign a server-provided challenge, or, failing that, a fixed value concatenated with a timestamp.

See https://keybase.io/docs/api/1.0/call/login for a vaguely related auth scheme, though it is only for inspiration, I am not suggesting we would be able to depend on it or share any code with it.

tiziano88:
@rbehjati I think this is just a dup / slightly different version (perhaps explained in a slightly more general way) of what you are already working on, could you dedup against the current one, and see if there is anything interesting in addition to what we already have elsewhere? Thanks

rbehjati:
I am closing this, since the challenge-response based authentication (#1357) which is currently under implementation, replaces the bearer token and provides more or less the same functionality described here.

PR #1820, opened 2020-12-10
Remove GrpcTag

rbehjati:
Now that we have the `PublicKeyIdentityTag`, the `GrpcTag`, and hmac bearer token can be removed. 

Ref #565


rbehjati:
I am not quite comfortable with removing all that cpp code. Do we need something similar for `PublicKeyIdentityTag` instead?

oak_abi/proto/label.proto
```diff
@@ -39,7 +39,6 @@ message Label {
 // See https://pdos.csail.mit.edu/papers/flume-sosp07.pdf section 3.1.
 message Tag {
   oneof tag {
-    GrpcTag grpc_tag = 1;
     WebAssemblyModuleTag web_assembly_module_tag = 2;
```
conradgrobler, line 42: It might be worth renumbering these to start at 1, seeing that we don't have any historic stored data that must be parsed. Or alternatively mark 1 as reserved. See https://developers.google.com/protocol-buffers/docs/proto?hl=en#reserved.

conradgrobler (approved):

conradgrobler:
> I am not quite comfortable with removing all that cpp code. Do we need something similar for `PublicKeyIdentityTag` instead?

I think it would be good to add support for the PublicKeyIdentityTag to the common cpp code (label.h and label.cpp) to keep the proto defintions and cpp helpers in sync.

I don't think there is a need to add special handling for public key identity tags in application_client.h.

I also noticed that authorization_bearer_token_metadata.h and authorization_bearer_token_metadata.cc seems to be unused after these changes, so could probably also be deleted.

conradgrobler:
> I don't think there is a need to add special handling for public key identity tags in application_client.h.

Thinking about it a bit more, I guess it would need to provide support for adding the authentication response token as a metadata field as part of the gRPC challenge-response authentication implementation.

oak_abi/src/label/tests.rs
```diff
@@ -57,8 +57,8 @@ fn serialize_deserialize() {
 
 #[test]
 fn label_flow() {
-    let tag_0 = authorization_bearer_token_hmac_tag(&[0, 0, 0]);
-    let tag_1 = authorization_bearer_token_hmac_tag(&[1, 1, 1]);
+    let tag_0 = public_key_identity_tag(vec![0, 0, 0]);
```
tiziano88, line 60: Is there a particular reason why the other fn takes a slice ref, and the new one an actual vec? May be worth changing it to take a ref as well, perhaps in a separate PR, but up to you anyways.

tiziano88 (approved):

rbehjati:
> > I don't think there is a need to add special handling for public key identity tags in application_client.h.
> 
> Thinking about it a bit more, I guess it would need to provide support for adding the authentication response token as a metadata field as part of the gRPC challenge-response authentication implementation.

I tried to add the public-key identity tag. But something is failing. I'll have to look into this later.

rbehjati:
> > I don't think there is a need to add special handling for public key identity tags in application_client.h.
> 
> Thinking about it a bit more, I guess it would need to provide support for adding the authentication response token as a metadata field as part of the gRPC challenge-response authentication implementation.

As discussed, I have created a separate issue for this: #1838

Issue #1776, opened 2020-11-26
Cargo deadlinks version is unpinned in Dockerfile 


ebeasant-arm:
Current x86 Dockerfile does not pin version of deadlines:

RUN cargo install cargo-deadlinks

Current docker image version of deadlinks:
cargo deadlinks --version
0.4.1

Should probably read something like:
ARG deadlinks_version=0.4.1
RUN cargo install --version=${deadlinks_version} cargo-deadlinks

(Note that Version 0.6.1 and above report errors in documentation)

tiziano88:
Good catch, I think we should fix it for the main Dockerfile too:

https://github.com/project-oak/oak/blob/a3e0ea9351967854f3af536f958fe59c46fa9930/Dockerfile#L218-L219

tiziano88:
(sorry, just realised you **were** talking about the main Dockerfile)

PR #1823, opened 2020-12-14
Pin deadlinks version

ipetr0v:
This change pins `cargo deadlinks` version to `0.5.0` in the Dockerfile.

Fixes https://github.com/project-oak/oak/issues/1776

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -216,7 +216,7 @@ RUN rustup component add \
   rustfmt
 
 # No binary available on Github, have to use cargo install.
-ARG deadlinks_version=0.7.0
+ARG deadlinks_version=0.5.0
 RUN cargo install --version=${deadlinks_version} cargo-deadlinks
 
 # Where to install rust tooling
--- a/Dockerfile
+++ b/Dockerfile
@@ -216,7 +216,8 @@ RUN rustup component add \
   rustfmt
 
 # No binary available on Github, have to use cargo install.
-RUN cargo install cargo-deadlinks
+ARG deadlinks_version=0.7.0
+RUN cargo install --version=${deadlinks_version} cargo-deadlinks
 
 # Where to install rust tooling
 ARG install_dir=${rustup_dir}/bin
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -216,7 +216,8 @@ RUN rustup component add \
   rustfmt
 
 # No binary available on Github, have to use cargo install.
-RUN cargo install cargo-deadlinks
+ARG deadlinks_version=0.4.1
```
tiziano88, line 219: Why not use the latest version? https://github.com/deadlinks/cargo-deadlinks/releases/tag/0.7.0

tiziano88 (approved):

Dockerfile
ipetr0v: Updated

ipetr0v: The latest version shows errors in every single file, and it doesn't look like it's a legitimate error:
```
tests ❯ cargo doc ╔════════════════════════
tests ❯ cargo doc ║ ════╡ stdout ╞════
tests ❯ cargo doc ║ Found invalid urls in trusted_database_client_bin/struct.Opt.html:
tests ❯ cargo doc ║ 	Fragment #tymethod.from_clap at trusted_database_client_bin/struct.Opt.html does not exist!
tests ❯ cargo doc ║ Found invalid urls in chat/struct.Router.html:
tests ❯ cargo doc ║ 	Fragment #tymethod.create at chat/struct.Router.html does not exist!
tests ❯ cargo doc ║ 	Fragment #tymethod.handle_command at chat/struct.Router.html does not exist!
tests ❯ cargo doc ║ Found invalid urls in chat/struct.Main.html:
tests ❯ cargo doc ║ 	Fragment #tymethod.handle_command at chat/struct.Main.html does not exist!
tests ❯ cargo doc ║ Found invalid urls in chat/struct.Room.html:
tests ❯ cargo doc ║ 	Fragment #tymethod.invoke at chat/struct.Room.html does not exist!
tests ❯ cargo doc ║ 	Fragment #tymethod.create at chat/struct.Room.html does not exist!
tests ❯ cargo doc ║ Found invalid urls in chat/proto/struct.SendMessageRequest.html:
```

ipetr0v: Will try to check different versions to see, which one starts this.

ipetr0v: Version `0.5.0` is the latest, that works

Issue #1792, opened 2020-12-03
Chrome Extension: Prevent top-level navigation


juliettepretot:
The chrome extension currently relies on the `sandbox` CSP to prevent data leaks from navigation. By my understanding of the the spec it should prevent all navigations of the top level browsing context:

> This flag prevents content from navigating browsing contexts other than the sandboxed browsing context itself (or browsing contexts further nested inside it), auxiliary browsing contexts (which are protected by the sandboxed auxiliary navigation browsing context flag defined next), and the top-level browsing context 

https://html.spec.whatwg.org/multipage/origin.html#sandboxed-navigation-browsing-context-flag

However upon investigation it seems that if the sandboxed page is itself in the top-level browsing context (as is the case enforced by the Oak chrome extension), browsers permit navigating to other pages in the top level browsing context (Tested in chrome). Unless I misread the spec, this seems like a bug.

### The navigate-to CSP (not yet possible)
This would prevent navigations, but it's a working draft, and not yet implemented: https://www.chromestatus.com/feature/6457580339593216

### Wrap the page in an iFrame (possible but not that nice)
This prevents top level navigation, and in-frame navigation can be prevented with the frame-src CSP, only allowing inline sources. This solves the issue nicely, while we wait for the navigate-to CSP to be implemented by browsers. 

Unfortunately though, to verify the proper implementation of this (ensuring the top-level context is just an iframe wrapper), the chrome extension would need to read the source-code of the page, for which unfortunately there currently is no chrome API for. 

The only solution to this is to make the chrome extension wrap the page in an iframe itself, say after the user clicks its icon to activate it. It's not nearly as nice (since now the extension doesn't just validate CSPs, but requires user action to work), but it appears to be the only viable solution. As a silver lining, this potentially restores the opt-in functionality of the original extension, allowing the page to load fronts/images/etc from external sources before entering "secure" mode.

juliettepretot:
Resolved with #1824

PR #1824, opened 2020-12-14
Update chrome extension to prevent top level navigation

juliettepretot:
# Checklist

      construction.

ref #1792

### DIFF
--- a/chrome_extension/README.md
+++ b/chrome_extension/README.md
@@ -17,7 +17,7 @@ This enforcement works by using two mechanisms in conjunction:
   to prevent from leaking data through navigation.
 - It applies the
   [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)
-  of `sandbox allow-scripts; default-src 'unsafe-inline';` to prevents network
+  of `sandbox allow-scripts; default-src 'unsafe-inline';` to prevent network
   calls, and stop leaks via cookies/localStorage/postMessage/etc.
 
 To load a website into this secure context, the user clicks the red extension
--- a/chrome_extension/README.md
+++ b/chrome_extension/README.md
@@ -3,36 +3,49 @@
 The Oak Chrome Extension is a proof of concept to show how Oak may be used as
 part of web applications.
 
-At its current state, the extension detects a website can leak information
-entered into it to outside servers. For example it can ensure that a website
-performing base64 encoding does so locally only, and does not leak the entered
-information.
+In its current state, the extension can load a given website into a secure
+sandbox that prevents it from leaking information entered into it. For example
+it can enforce that a website performing base64 encoding does so locally only.
 
-This enforcement happens via the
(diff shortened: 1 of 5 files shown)


conradgrobler (approved):
Nice.

chrome_extension/README.md
```diff
@@ -3,36 +3,49 @@
...
+
+- It renders the source code of the website in an iframe, utilizing the
+  `frame-src: 'unsafe-inline'`
+  [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)
+  alongside the
+  [sandbox attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox)
+  to prevent from leaking data through navigation.
+- It applies the
+  [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)
+  of `sandbox allow-scripts; default-src 'unsafe-inline';` to prevents network
```
tiziano88, line 20:
```suggestion
  of `sandbox allow-scripts; default-src 'unsafe-inline';` to prevent network
```

chrome_extension/README.md
```diff
@@ -3,36 +3,49 @@
...
-
-Depending on whether a given website complies with this CSP or not, the
-extension will show either a red or green icon.
+In its current state, the extension can load a given website into a secure
+sandbox that prevents it from leaking information entered into it. For example
+it can enforce that a website performing base64 encoding does so locally only.
+
+This enforcement works by using two mechanisms in conjunction:
+
+- It renders the source code of the website in an iframe, utilizing the
```
tiziano88, line 12: I don't remember what the conclusion was when we talked about it, but AFAICT in principle now this does not need to be a Chrome extension any more, it may just be a static web page sitting somewhere, right? Or is that not possible because of CORS or other restrictions?

tiziano88 (approved):

chrome_extension/README.md
juliettepretot: Initially I tried to do this with a websites that uses a restricted iframe to load any website. The problem is that CSP cannot do exact matching, it excludes the query path of any URL. Eg `?foo=bar`. Hence we cannot completely eliminate in-frame navigation by. setting the `frame-src` CSP to the URL of specific website. This extension circumvents that by inlining the source code for the website directly and restrict navigation with `frame-src 'unsafe-inline'`. 

A website could theoretically do that too, but due to CORS restrictions it won't work for many websites at all since requests to other websites won't be allowed. 

I could however make this a bookmarklet, a JavaScript bookmark that can be run in any browser.

juliettepretot:
@googlebot I fixed it.

chrome_extension/README.md
juliettepretot: Another solution for a website only approach is to use a service worker instead of CSP to block requests, but I'll need to further investigate

Issue #1496, opened 2020-09-18
Add a typed version of `channel_create`  in the Runtime


rbehjati:
We need a typed version of `channel_create` similar to the one in the `sdk`:
https://github.com/project-oak/oak/blob/8f03093723535d6291a365859ae9a8d850b66986/sdk/rust/oak/src/io/mod.rs#L31
in the Runtime, possibly in `oak_runtime::io`.

PR #1834, opened 2021-01-06
Add typed channel_create in Oak Runtime

ipetr0v:
This change:
- Adds typed `channel_create` and `channel_create_with_downgrade` to `oak_runtime`
- Updates gRPC server and HTTP tests with new functions

Fixes https://github.com/project-oak/oak/issues/1496

### DIFF
--- a/oak_runtime/src/io.rs
+++ b/oak_runtime/src/io.rs
@@ -113,3 +113,24 @@ impl<T: Encodable> SenderExt<T> for Sender<T> {
             .map_err(|err| err.into())
     }
 }
+
+/// Creates a new channel for transmission of [`Encodable`] and [`Decodable`] types.
+pub fn channel_create<T: Encodable + Decodable>(
+    runtime: &RuntimeProxy,
+    name: &str,
+    label: &Label,
+) -> Result<(Sender<T>, Receiver<T>), OakStatus> {
+    let (wh, rh) = runtime.channel_create(name, label)?;
+    Ok((Sender::<T>::new(wh.into()), Receiver::<T>::new(rh.into())))
+}
+
+/// Uses the current node's label-downgrading privilege to create a new channel for transmission of
+/// [`Encodable`] and [`Decodable`] types.
+pub fn channel_create_with_downgrade<T: Encodable + Decodable>(
+    runtime: &RuntimeProxy,
+    name: &str,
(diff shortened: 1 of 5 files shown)


conradgrobler (approved):

oak_runtime/src/node/grpc/server/mod.rs
```diff
@@ -514,41 +514,37 @@ impl GrpcInvocationHandler {
...
+
         // The channel containing the response is created with the label containing the identity of
         // the user.
-        let (response_writer, response_reader) = self
-            .runtime
-            .channel_create_with_downgrade("gRPC response", identity_label)
-            .map_err(|err| {
-                warn!("could not create gRPC response channel: {:?}", err);
-            })?;
+        let (response_sender, response_receiver) = channel_create_with_downgrade::<GrpcResponse>(
```
rbehjati, line 525: Does it make sense to make a similar change for `HttpServerNode`? Channel creation is actually done in the `Pipe`:
https://github.com/project-oak/oak/blob/2bdae2477e5f0cb368b8cbbf6033dc8493e21a26/oak_runtime/src/node/http/util.rs#L59

oak_runtime/src/io.rs
```diff
@@ -113,3 +113,25 @@ impl<T: Encodable> SenderExt<T> for Sender<T> {
...
+}
+
+/// Uses the current node's label-downgrading privilege to create a new channel for transmission of
+/// [`Encodable`] and [`Decodable`] types.
+pub fn channel_create_with_downgrade<T: Encodable + Decodable>(
+    runtime: &RuntimeProxy,
+    name: &str,
+    label: &Label,
+) -> Result<(Sender<T>, Receiver<T>), OakStatus> {
+    let (wh, rh) = runtime.channel_create_with_downgrade(name, label)?;
```
rbehjati, line 135: Would it work if instead `RuntimeProxy::channel_create_with_downgrade` was changed to be typed? Although I think that change would require updating all the calls to `channel_create_with_downgrade`.

rbehjati (approved):

oak_runtime/src/io.rs
ipetr0v: I think we still rely a lot on these low level functions. For example in tests.

PR #1839, opened 2021-01-13
Add wait_on_channels_with_downgrade to SDK

ipetr0v:
This change adds:
- `wait_on_channels_with_downgrade` to ABI, SDK andRuntime

Fixes https://github.com/project-oak/oak/issues/1736

### DIFF
--- a/examples/abitest/module_0/rust/src/lib.rs
+++ b/examples/abitest/module_0/rust/src/lib.rs
@@ -292,16 +292,20 @@ impl OakAbiTestService for FrontendNode {
             (Self::test_channel_chain_recovered, Count::Unchanged),
         );
         tests.insert(
-            "WaitOnChannelsRaw",
-            (Self::test_channel_wait_raw, Count::Unchanged),
-        );
-        tests.insert(
             "WaitOnChannels",
             (Self::test_channel_wait, Count::Unchanged),
         );
         tests.insert(
             "WaitOnChannelsOrphan",
             (Self::test_channel_wait_orphan, Count::Unchanged),
+        );
+        tests.insert(
+            "WaitOnChannelsWithDowngrade",
+            (Self::test_channel_wait_with_downgrade, Count::Unchanged),
+        );
+        tests.insert(
+            "WaitOnChannelsRaw",
+            (Self::test_channel_wait_raw, Count::Unchanged),
         );
         tests.insert("NodeCreate", (Self::test_node_create, Count::Unchanged));
         tests.insert(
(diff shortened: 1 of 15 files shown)


ipetr0v:
Currently is blocked by https://github.com/project-oak/oak/pull/1837 since it implements `abitest` signing.

tiziano88:
just checking: do we need this for anything in particular (and similarly for read_with_downgrade)?

ipetr0v:
Examples currently don't use these functions.
This is a remainder logic necessary for IFCv0.

tiziano88:
> Examples currently don't use these functions.
> This is just remaining logic necessary for IFCv0.

I don't remember it being necessary for anything in particular, definitely not for IFCv0, in fact https://github.com/project-oak/oak/issues/1736 is added to IFCv1, but I am not even sure we should do it at all unless we have a use case for it, or do you have anything in mind?

tiziano88:
(BTW similar argument for https://github.com/project-oak/oak/pull/1837)

ipetr0v:
This PR just completes the `with_downgrade` SDK that has already been started (`write_with_downgrage`, `node_create_with_downgrade`).
So that we have a full SDK

ipetr0v:
Also this PR adds `wait_with_downgrade` that should be user by `Receiver::receive_with_downgrade` instead of just `wait`:
https://github.com/project-oak/oak/blob/490a3e8491900aefda4d43aa3167623643c27e12/sdk/rust/oak/src/io/receiver.rs#L164-L167

examples/abitest/module_0/rust/src/lib.rs
```diff
@@ -1624,6 +1628,41 @@ impl FrontendNode {
...
+        oak::channel_write(private_out_channel, &test_msg, &[])
+            .expect("could not write to channel");
+        expect_eq!(
+            vec![ChannelReadStatus::PermissionDenied],
+            status_convert(oak::wait_on_channels_with_downgrade(&[private_in_channel]))?
+        );
+        expect_eq!(Ok(()), oak::channel_close(private_out_channel.handle));
+        expect_eq!(Ok(()), oak::channel_close(private_in_channel.handle));
+
+        Ok(())
```
conradgrobler, line 1663: Perhaps add a todo to check the case where it is different from the normal `wait_on_channels` once the ABI tests can support wasm hash labels.

sdk/rust/oak/src/lib.rs
```diff
@@ -128,6 +128,32 @@ pub fn wait_on_channels(handles: &[ReadHandle]) -> Result<Vec<ChannelReadStatus>
     }
 }
 
+/// The same as [`wait_on_channels`](#method.wait_on_channels), but also applies the current Node's
+/// downgrade privilege when checking IFC restrictions.
+pub fn wait_on_channels_with_downgrade(
+    handles: &[ReadHandle],
+) -> Result<Vec<ChannelReadStatus>, OakStatus> {
+    let mut space = new_handle_space(handles);
```
conradgrobler, line 136: Not sure whether it is worth it, but you could consider extracting the logic into a util function to avoid the duplication of quite complex logic.

conradgrobler (approved):

PR #1840, opened 2021-01-13
Use anyhow in HttpServerNode

### LINKED ISSUE #1693: Use anyhow in HttpServerNode
Originated from the following comments: 
https://github.com/project-oak/oak/pull/1652#discussion_r516849374
and 
https://github.com/project-oak/oak/pull/1586/files#r510714975

rbehjati:
Fixes #1693.

### DIFF
--- a/oak_runtime/src/node/http/server.rs
+++ b/oak_runtime/src/node/http/server.rs
@@ -21,14 +21,11 @@
 
 use crate::{
     io::ReceiverExt,
-    node::{
-        http::util::{HttpError, Pipe},
-        ConfigurationError, Node, NodePrivilege,
-    },
+    node::{http::util::Pipe, ConfigurationError, Node, NodePrivilege},
     proto::oak::invocation::HttpInvocationSender,
     RuntimeProxy,
 };
-use core::task::{Context, Poll};
+use core::task::Poll;
 use futures_util::stream::Stream;
 use http::{request::Request, response::Response};
 use hyper::{
@@ -50,6 +47,7 @@ use tokio::{
 };
 use tokio_rustls::server::TlsStream;
 
+use anyhow::{anyhow, Context};
 use futures_util::{
     future::TryFutureExt,
     stream::{StreamExt, TryStreamExt},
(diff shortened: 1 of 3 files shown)


oak_runtime/src/node/http/server.rs
```diff
@@ -320,43 +310,16 @@ impl HttpRequestHandler {
...
-                    .status(StatusCode::BAD_REQUEST)
-                    .body(Body::from("Could not verify user's identity."))
-                    .map_err(|e| {
-                        HttpError::HttpResponse(format!("Could not create response: {}", e))
-                    })
-            }
-            Err(http_err) => {
-                warn!("Internal server error: {:?}", http_err);
+            Err(err) => {
+                warn!("Error when handling the request: {}", err);
```
rbehjati, line 318: Based on the changes here, I am no longer certain that using anyhow is a good idea, as we cannot distinguish between a bad request and other types of errors. However, maybe the http responses here are not that important after all, since in most cases they are not delivered to the caller. WDYT @tiziano88?

oak_runtime/src/node/http/server.rs
```diff
@@ -612,5 +532,5 @@ fn to_hyper_response(http_response: HttpResponse) -> Result<Response<Body>, Http
     builder
         .status(http_response.status as u16)
         .body(Body::from(http_response.body))
-        .map_err(|err| HttpError::HttpResponse(format!("Could not build response: {}", err)))
+        .map_err(|err| anyhow!("Could not build response: {}", err))
```
tiziano88, line 535: Does `context` work for this and other cases in which we want to propagate the original error?

oak_runtime/src/node/http/server.rs
tiziano88: What is the effect of this change? That everything will be returned as `StatusCode::INTERNAL_SERVER_ERROR`, even if it is another type of error? If so, that seems indeed too coarse. Or are some more particular errors still being propagated precisely?

tiziano88:
Is this the same as #1841 ? cc @ipetr0v

oak_runtime/src/node/http/server.rs
rbehjati: After a closer look at the code (thanks to [comment](https://github.com/project-oak/oak/issues/1693#issuecomment-759619180) from @ipetr0v), I think it is correct to always send a bad request response, since `get_oak_label` only returns an error if something is wrong with the label. I have changed the code to return `StatusCode::BAD_REQUEST`.

oak_runtime/src/node/http/server.rs
rbehjati: Yes. Fixed.

rbehjati:
> Is this the same as #1841 ? cc @ipetr0v

Yes. I have updated this one to include more `context` like Ivan has in #1841.

ipetr0v (approved):

PR #1843, opened 2021-01-22
Allow passing borrowed handles to SDK functions.

wildarch:
This is currently not needed because handles implement `Copy`, but once
they are made linear, using borrowed handles will be the only way to use
these functions without giving up ownership of the handle.

# Checklist

      construction.

Related issue: #1686

### DIFF
--- a/sdk/rust/oak/src/lib.rs
+++ b/sdk/rust/oak/src/lib.rs
@@ -18,6 +18,7 @@
 //! underlying Oak platform functionality.
 
 use byteorder::WriteBytesExt;
+use core::borrow::Borrow;
 use log::{debug, error, warn};
 use oak_abi::proto::oak::application::NodeConfiguration;
 use oak_io::{Decodable, Encodable};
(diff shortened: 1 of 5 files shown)


wildarch:
Failed test looks like a flake:

```
Get https://registry-1.docker.io/v2/debian/snapshot/manifests/buster-20201012: received unexpected HTTP status: 502 Bad Gateway
```

sdk/rust/oak/src/lib.rs
```diff
@@ -67,11 +68,11 @@ pub mod proto {
 
 // Build a chunk of memory that is suitable for passing to oak_abi::wait_on_channels,
 // holding the given collection of channel handles.
-fn new_handle_space(handles: &[ReadHandle]) -> Vec<u8> {
+fn new_handle_space<R: Borrow<ReadHandle>>(handles: &[R]) -> Vec<u8> {
```
tiziano88, line 71: Is this just temporary until we modify all the call places to use actual borrows, or do you think we will want to keep it after that? I think it does not look very nice, so I hope we can get rid of it at some point (perhaps worth a TODO + issue?)

tiziano88 (approved):

sdk/rust/oak/src/lib.rs
wildarch: Once linear handles are fully implemented and the default, then this can be simplified again. I'll add a TODO and link to the linear handles issue.

PR #1844, opened 2021-01-22
Downgrading Security Specification

aferr:
This is a candidate specification of a security condition that does permit downgrades and ensures that the only way around the information flow labels is through these downgrades. There is just a partial specification of the model that supports downgrades that includes just one call so that the proof effort can focus on the proof structure and the model can be scaled up later.

# Checklist

      construction.


experimental/ni_coq/vfiles/EscapeHatchCondition_def.v
```diff
@@ -0,0 +1,41 @@
...
+    Events
+    LowEquivalences
+    ModelTypes
+    TraceLowEq.
+
+
+Definition eh_t := @trace (state * down_l * event_l).
+Definition is_init(t: eh_t) := length t = 1.
+
+Definition d_in_trace_elt d (t_elt: (state * down_l * event_l)) :=
```
aferr, line 17: @jadephilipoom 

This file has a WIP specification of a downgrading security condition loosely inspired by delimited release and TNIsec, but it doesn't typecheck yet. I am wondering if there is a way to actually specify this in coq without introducing another decidable equality.

The problem is with d_in_trace_elt and traces_corresp. I need some way of specifying when a trace of type `list (state * down_l * event_l)` matches a sequence of type `list down_l`. I think just dropping the downgrade lists from the forall and producing them from t1n/t2n using `map _.2` doesn't quite work since the theorem needs to say that `dt1 =L dt2` before the `exists t2n` bit.

Downgrades generate a new kind of event that is used to specify another low-observable input (even though they do deliberately contain high values). Roughly the security condition says that if two initial states t1_init[0] and t2_init[0] look the same to level L AND there are sequences of downgrade events dt1 dt2 that look the same to L then the possible behaviors beginning from t1_init[0] and using downgrades dt1 will look the same to L as the possible behaviors beginning from t2_init[0] and using downgrades dt2.

(As a side point, I am considering cutting out the "events" because the only event that actually matters is the "InEv" which is used to keep track of the messages that a node receives from a channel. The other events are captured by the state, and it seems like it could be reasonable to just pack "most recently read messsage" into the node state.)

aferr: Actually, I think the latest small change might work just fine (and possibly my not figuring it out initially was a bit silly :). I am not sure if the extra equality on lists is really needed or if it would actually be just fine to do "=" on the two lists directly. (?)

experimental/ni_coq/vfiles/EscapeHatchCondition_def.v
```diff
@@ -0,0 +1,42 @@
...
+    Events
+    LowEquivalences
+    ModelTypes
+    TraceLowEq.
+
+
+Definition eh_t := @trace (state * down_l * event_l).
+Definition is_init(t: eh_t) := length t = 1.
+
+Inductive list_eq {A: Type}: list A -> list A -> Prop :=
```
aferr, line 17: Do I need this or is it just fine to do "=" on two lists directly?

experimental/ni_coq/vfiles/EscapeHatchCondition_def.v
jadephilipoom: It's fine to just do `=`! The standard library even has a decider for this equality that you can use (`list_eq_dec`).

experimental/ni_coq/vfiles/TraceLowEq.v
```diff
@@ -114,3 +114,42 @@ Inductive stut_trace_low_eq {s_leq: @low_eq_t state}{e_leq: @low_eq_t event_l}:
         event_low_eq ell xe ye ->
         state_low_eq ell xs ys ->
         stut_trace_low_eq ell ((xs, xe)::t1) ((ys, ye)::t2).
+
+(*============================================================================
+* Low Equivalences for escape hatch condition
+* ==========================================================================*)
+(* These are used for downgrading conditions involving *)
+
+(* this is just "map ._2" *)
```
jadephilipoom, line 123: It might make proofs easier to define this using `map`, then; you'd have access to standard library proofs instead of having to make every lemma about this definition inductive.

jadephilipoom (approved):
LGTM!

experimental/ni_coq/vfiles/TraceLowEq.v
aferr: Thanks. I wasn't sure which would make proofs easier. Is there also a standard library implementation of "._2" for tuples that are not just pairs? I couldn't find it.

jadephilipoom: No, there isn't. Either `map (fun '(s,d,e) => d) t` or `map (fun sde => snd (fst sde)) t` should be fine, with the tradeoff that the first is more readable and the second is destructed more easily in proofs.

PR #1845, opened 2021-01-22
Add defense-in-depth support

### LINKED ISSUE #1029: defense-in-depth configuration of `oak_loader`

rbehjati:
* Updates the runtime to take into account the permissions configuration
  when creating the nodes.
* Adds a new input flag to `oak_loader` for permissions. This is only
  required if `oak_debug` is disabled.
* Updates runner with a path to a permissions `.toml` file
* Updates the documentation with info about permissions file
* Updates gcp scripts accordingly

Fixes #1029

# Checklist

      construction.

### DIFF
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -368,20 +368,52 @@ using the following script:
 ./scripts/push_example -e hello_world
 ```
 
+### Creating a Permissions File
+
+A permissions file is a `.toml` file provided by the host owner, as an
+additional layer of defense. This file specifies the features of Oak that are
+permitted for the applications running on the host. In particular, using this
+file, the host owner can enable or disable gRPC or HTTP connections. In
+addition, it is possible to specify an allowlist of external gRPC or HTTP
+authorities (in the `[userinfo@]host[:port]` format) that the applications can
+connect to over TLS. Interaction with all other authorities is prohibited.
+Connections to insecure HTTP servers are allowed only if explicitly enabled via
+the `allow_insecure_http_egress` flag.
+
+Here is an example of a permissions file:
+
+```toml
+allow_grpc_server_nodes = true
+allow_http_server_nodes = true
+allow_log_nodes = true
+allow_insecure_http_egress = true
+allow_egress_https_authorities = ["localhost:8080", "localhost:8888"]
+
+```
+
+All permissions are denied by default. So, an empty permissions file is the most
+restrictive one.
+
 ### Starting the Oak Application
 
 The Oak Application is then started using the Oak Loader:
 
 ```bash
-./oak_loader/target/x86_64-unknown-linux-musl/release/oak_loader --application=./examples/hello_world/bin/hello_world.oak
+./oak_loader/target/x86_64-unknown-linux-musl/release/oak_loader \
+  --application=./examples/hello_world/bin/hello_world.oak \
+  --permissions=./examples/permissions/permissions.toml
 ```
+
+Providing a permissions file via the `permissions` flag is only needed for
+Logless server releases.
 
 The Oak Loader will launch an [Oak Runtime](concepts.md#oak-runtime), and this
 Runtime will check the provided Wasm module(s) and application configuration.
 Assuming everything is correct (e.g. the Nodes all have a main entrypoint and
 only expect to link to the Oak [host functions](abi.md#host-functions)), the Oak
-Runtime opens up the gRPC port specified by the Application Configuration. This
-port is then used by clients to connect to the Oak Application.
+Runtime opens up the gRPC port specified by the Application Configuration, if
+creating a gRPC connection is permitted in the permissions file. This port is
+then used by clients to connect to the Oak Application.
 
 ### Configuring the Oak Application
 
@@ -678,8 +710,13 @@ framework via the Oak Runtime:
 #[tokio::test(core_threads = 2)]
 async fn test_translate() {
     let _ = env_logger::builder().is_test(true).try_init();
+    let permissions = oak_runtime::permissions::PermissionsConfiguration {
+        allow_grpc_server_nodes: true,
+        allow_log_nodes: true,
+        ..Default::default()
+    };
 
-    let runtime = oak_tests::run_single_module_default(MODULE_WASM_FILE_NAME)
+    let runtime = oak_tests::run_single_module_default(MODULE_WASM_FILE_NAME, permissions)
         .expect("Unable to configure runtime with test wasm!");
 
     let (channel, interceptor) = oak_tests::public_channel_and_interceptor().await;
@@ -731,7 +768,11 @@ like in the following example:
 #[tokio::test(core_threads = 2)]
 async fn test_say_hello() {
     let _ = env_logger::builder().is_test(true).try_init();
-
+    let permissions = oak_runtime::permissions::PermissionsConfiguration {
+        allow_grpc_server_nodes: true,
+        allow_log_nodes: true,
+        ..Default::default()
+    };
     let runtime_config = oak_tests::runtime_config_wasm(
         hashmap! {
             MAIN_MODULE_NAME.to_owned() => oak_tests::compile_rust_wasm(MAIN_MODULE_MANIFEST, MAIN_MODULE_WASM_FILE_NAME, oak_tests::Profile::Release).expect("Couldn't compile main module"),
@@ -740,6 +781,7 @@ async fn test_say_hello() {
         MAIN_MODULE_NAME,
         MAIN_ENTRYPOINT_NAME,
         ConfigMap::default(),
+        permissions,
         oak_runtime::SignatureTable::default(),
     );
     let runtime = oak_runtime::configure_and_run(runtime_config)
--- a/examples/permissions/permissions.toml
+++ b/examples/permissions/permissions.toml
@@ -1,0 +1,11 @@
+allow_grpc_server_nodes = true
+allow_http_server_nodes = true
+allow_log_nodes = true
+allow_insecure_http_egress = true
+allow_egress_authorities = [
+  "localhost:7878",
+  "test.invalid:9999",
+  "localhost:8888",
+  "localhost:8080",
+  "localhost:7856",
+]
--- a/oak_loader/src/options.rs
+++ b/oak_loader/src/options.rs
@@ -23,6 +23,7 @@ use log::debug;
 use oak_abi::proto::oak::application::{ApplicationConfiguration, ConfigMap};
 use oak_runtime::{
     auth::oidc_utils::{parse_client_info_json, ClientInfo},
+    permissions::PermissionsConfiguration,
     tls::Certificate,
     SignatureTable,
 };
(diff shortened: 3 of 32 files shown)


oak_runtime/src/node/mod.rs
```diff
@@ -131,24 +147,32 @@ impl NodeFactory<NodeConfiguration> for ServerNodeFactory {
                 Ok(Box::new(crypto::CryptoNode::new(node_name)))
             }
             Some(ConfigType::LogConfig(LogConfiguration {})) => {
-                Ok(Box::new(logger::LogNode::new(node_name)))
+                if self.permissions_configuration.allow_log_nodes {
```
rbehjati, line 150: I think these checks could be moved into the `new` functions. I can think of arguments for both approaches. WDYT @tiziano88?

docs/programming-oak.md
```diff
@@ -368,20 +368,47 @@ using the following script:
...
+### Creating a Permissions File
+
+A permissions file is a `.toml` file provided by the host owner, as an
+additional layer of defense. This file specifies the features of Oak that are
+permitted for the applications running on the host. In particular, using this
+file, the host owner can enable or disable gRPC or HTTP connections. In
+addition, it is possible to specify an allowlist of external gRPC or HTTP
+services that the applications can connect to. Interaction with all other
+external servers will be prohibited. Connections to insecure HTTP servers will
+be allowed regardless of the allowlist! However, Oak's IFC guarantees that only
```
rbehjati, line 380: I added this note here, but I think the permissions file should allow blocking insecure HTTP connections altogether. So, I think we want the Runtime to block insecure HTTP connections by default. But if a host owner wants it to be allowed, then a special token, e.g., `HTTP`, should be included in `allow_http_egress_addresses`. WDYT?

docs/programming-oak.md
```diff
@@ -368,20 +368,47 @@ using the following script:
...
+`public` data will be sent over to insecure external services.
```
ipetr0v, line 381: Also, can Oak applications request their permission list?
But I'm not sure if it's really necessary. Probably it would make sense, if applications are supposed to work with different permission combinations, and not fail.

examples/permissions/permissions.toml
```diff
@@ -0,0 +1,10 @@
+allow_grpc_server_nodes = true
+allow_http_server_nodes = true
+allow_log_nodes = true
+allow_introspection = true
+allow_grpc_egress_addresses = [
+  "https://localhost:7878",
+  "https://test.invalid:9999",
+  "https://localhost:8888"
+]
+allow_http_egress_addresses = ["localhost:8080", "localhost:7856"]
```
ipetr0v, line 10: Should HTTP and gRPC be the same, since gRPC is built on top of HTTP?
WDYT

oak_loader/src/options.rs
```diff
@@ -136,13 +143,17 @@ pub fn create_runtime_config() -> anyhow::Result<oak_runtime::RuntimeConfigurati
         } else {
             None
         },
-        introspect_port: if cfg!(feature = "oak_debug") && !opt.no_introspect {
+        introspect_port: if cfg!(feature = "oak_debug")
+            && !opt.no_introspect
+            && permissions_config.allow_introspection
```
ipetr0v, line 148: I think its a non-obvious change that is also hidden inside the code.
Should we have two separate permissions files, and choose between them in `runner` based on `oak_debug` ?

oak_loader/src/options.rs
```diff
@@ -314,3 +325,15 @@ fn create_app_config(opt: &Opt) -> anyhow::Result<ApplicationConfiguration> {
     Ok(ApplicationConfiguration::decode(app_config_data.as_ref())
         .context("could not parse application configuration")?)
 }
+
+/// Parse permissions configuration into an instance of [`PermissionsConfiguration`]
+fn create_permissions_config(opt: &Opt) -> anyhow::Result<PermissionsConfiguration> {
+    let permissions_config_data =
+        read(&opt.permissions).context("could not read permissions configuration")?;
+    if let Ok(permissions) = toml::from_str(std::str::from_utf8(permissions_config_data.as_ref())?)
```
ipetr0v, line 333: I think `.context` works with `toml` errors:
https://github.com/project-oak/oak/blob/1b6d0867c45c4f4ff95eb3c07881904b4e31b813/sdk/rust/oak_app_build/src/main.rs#L253-L254

oak_runtime/src/node/mod.rs
```diff
@@ -177,31 +211,46 @@ impl NodeFactory<NodeConfiguration> for ServerNodeFactory {
...
-                    .clone()
-                    .expect("no HTTP configuration provided to Oak Runtime")
-                    .http_client_root_tls_certificate
-                    .expect("no root TLS certificate provided to Oak Runtime");
-                Ok(Box::new(http::client::HttpClientNode::new(
-                    node_name,
-                    config.clone(),
-                    http_client_root_tls_certificate,
-                )?))
+                if config.authority.is_empty()
```
ipetr0v, line 231: Whats does empty authority mean?

rbehjati (commented):
Thanks for the review @ipetr0v.

ipetr0v (approved):

docs/programming-oak.md
```diff
@@ -368,20 +368,46 @@ using the following script:
...
+### Creating a Permissions File
+
+A permissions file is a `.toml` file provided by the host owner, as an
+additional layer of defense. This file specifies the features of Oak that are
+permitted for the applications running on the host. In particular, using this
+file, the host owner can enable or disable gRPC or HTTP connections. In
+addition, it is possible to specify an allowlist of external gRPC or HTTP
+services that the applications can connect to. Interaction with all other
+external servers will be prohibited. Connections to insecure HTTP servers will
+be allowed only if explicitly enabled via the `allow_insecure_http_egress` flag.
```
tiziano88, line 380: Does this mean that even if an http URL is present in `allow_egress_addresses`, it is not actually honored, if this flag is off? It seems unnecessary, since those URLs already have a schema?

oak_runtime/src/permissions.rs
```diff
@@ -20,29 +20,43 @@
...
+    /// Whether to enable creating an HTTP client for insecure HTTP connection.
     #[serde(default)]
-    allow_grpc_egress_addresses: Vec<String>,
+    pub allow_insecure_http_egress: bool,
+
+    /// List of the allowed URIs that the application can connect to via a gRPC or an HTTP client.
+    /// Only GrpcClientNodes or HttpClientNodes that are configured to connect to URIs (or
+    /// authorities) in this list can be created.
+    #[serde(default)]
+    pub allow_egress_addresses: Vec<String>,
```
tiziano88, line 51: Are those full URIs? Or prefixes? Or just hosts? Or authorities? What is the relationship between this and the HTTP egress flag, since these URIs already have a schema AFAICT?

docs/programming-oak.md
rbehjati: This flag concerns insecure HTTP clients, where the authority is empty. We decided earlier that it does not make sense to provide an authority in case of an HTTP client for insecure connections. The URIs/authorities in `allow_egress_addresses` will only be used for gRPC and secure HTTP connections. So, the list will still be honored, even if `allow_egress_addresses` is off.

oak_runtime/src/permissions.rs
rbehjati: Currently, the addresses are `scheme + authority`. But actually, it only works if the scheme is `https`. Perhaps it is better to list the authorities here, instead of the URIs. I have this change in my new commit.  I've also updated the comments with more clarification. 

There is no relation between this list and the `allow_insecure_http_egress` flag. Previously, we decided that we would not distinguish between different insecure HTTP hosts, when creating HttpClient nodes. So it does not make sense to have them in the allow list here.

rbehjati (commented):
Thanks for the comments @tiziano88. I've added clarifications and fixes.

oak_loader/src/options.rs
tiziano88: `oak_debug` is not a flag, it is a compile-time feature, it is "provided" by whoever compiles the code. The idea is that enabling `oak_debug` results in a version of Oak that does *not* enforce IFC (maybe we should rename the feature to `unsafe_oak_debug` or something scarier), while defense-in-depth permissions should result in a safe execution always. In particular, if introspection leaks data bypassing IFC, then it should probably only be available in `oak_debug` (and at that point perhaps there is not much point in also having it in the permission file).

cc @daviddrysdale

rbehjati: Thanks for the clarification @tiziano88. 
If there is no flag for disabling introspection in the permissions file, then defense-in-depth permissions will not result in a safe execution if the code is compiled with `oak_debug`. Let's say someone, by mistake, deploys a version that is compiled with `oak_debug`. With a flag for introspection in the permissions file, there is still a way to disable introspection. So, I think it makes sense to have it in the permissions file. The same applies to the log nodes too.

tiziano88: If someone deploys a debug version of Oak to production, we have much bigger problems :) since that could literally just log everything out.

tiziano88: I'm not against this change, but the way it is currently structured seems confusing, since multiple flags (some compile time, some run time) control the same underlying functionality in slightly different ways (e.g. logging, introspection). I think we should have a clearer explanation of how these things interact.

rbehjati: Oh! Good! I have always been wondering about the logs!! Now, I see that we actually disable logging when not compiled with `oak_debug`, and the log node sends to stderr. If the permissions-file is meant to be used with non-debug versions, then I agree that there is no point for having a permission for introspection.

rbehjati: Pushed a new change that removes `allow_introspection`, and limits the use of the permissions file to logless servers (those compiled with `oak_debug` disabled). WDYT?

oak_runtime/src/node/mod.rs
tiziano88: Actually an alternative option would be to check the permission upfront by passing the NodeConfiguration object to a function (rather than just the specific gRPC configs). 

```rust
if !self.permission_configuration.allowed_creation(&node_configuration) {
  return Err(...);
}
match &node_configuration.config_type {
  ...
}
```

tiziano88 (approved):
Thanks!

rbehjati (commented):
Thanks for the review @tiziano88 :)

PR #1850, opened 2021-01-29
Add linear-handles feature to oak_io

wildarch:
# Checklist

      construction.

The feature enables what I hope will eventually be the default behaviour for handles. Roughly speaking, this does 3 things:
- Make handles non-`Copy`
- Add a `Clone` impl that calls `oak_abi::handle_clone`
- Add a `Drop` impl that calls `oak_abi::channel_close`

This builds on #1782.
Related issue: #1686. It is step (3) in https://github.com/project-oak/oak/issues/1686#issuecomment-765351586.

### DIFF
--- a/examples/abitest/module_0/rust/src/lib.rs
+++ b/examples/abitest/module_0/rust/src/lib.rs
@@ -888,7 +888,7 @@ impl FrontendNode {
         oak::channel_write(init_wh, &[], &[result_wh.handle])
             .expect("Failed to write result handle");
 
-        // the linear_handles module should return a single result message (without handles), its
+        // The linear_handles module should return a single result message (without handles), its
         // body a string containing the test result.
         let mut buf = Vec::new();
         let mut handles = Vec::new();
--- a/examples/abitest/module_0/rust/src/lib.rs
+++ b/examples/abitest/module_0/rust/src/lib.rs
@@ -50,6 +50,7 @@ const FRONTEND_MODULE_NAME: &str = "frontend_module";
 const BACKEND_MODULE_NAME: &str = "backend_module";
 const BACKEND_ENTRYPOINT_NAME: &str = "backend_oak_main";
 const STORAGE_NAME_PREFIX: &str = "abitest";
+const LINEAR_HANDLES_MODULE_NAME: &str = "linear_handles_module";
 
 // Distinct listening addresses to avoid port-in-use errors
 const HTTP_ADDR: &str = "[::]:8383";
@@ -233,6 +234,10 @@ impl OakAbiTestService for FrontendNode {
         tests.insert(
             "HandleCloneRaw",
             (Self::test_handle_clone_raw, Count::Unchanged),
+        );
+        tests.insert(
+            "LinearHandles",
+            (Self::test_linear_handles, Count::Unchanged),
         );
         tests.insert(
             "ChannelReadRaw",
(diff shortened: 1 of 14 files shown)


rbehjati (approved):
LGTM. But can we have tests for it at this stage?

oak_io/src/handle.rs
```diff
@@ -59,6 +61,67 @@ impl From<Handle> for WriteHandle {
...
 
+/// Provides implementations of `Clone` and `Drop` for handle types, based on the `handle_clone` and
+/// `channel_close` ABI syscalls, respectively.
+#[cfg(feature = "linear-handles")]
+mod linear_handles {
+    use super::*;
+    use crate::OakStatus;
+    use log::trace;
+
+    /// Creates a new distinct handle using `oak_abi::clone_handle`
```
tiziano88, line 72: Could you make these actual links?

oak_io/src/handle.rs
```diff
@@ -59,6 +61,67 @@ impl From<Handle> for WriteHandle {
...
+            drop_handle(self.handle);
+        }
+    }
+
+    fn clone_handle(handle: Handle) -> Handle {
+        let mut cloned_handle = 0;
+        let result = OakStatus::from_i32(unsafe {
+            oak_abi::handle_clone(handle, &mut cloned_handle as *mut Handle) as i32
+        })
+        .expect("handle_clone returned invalid oak status");
```
tiziano88, line 109: Could you print the status too? (or does it get printed anyways?)

oak_io/src/handle.rs
```diff
@@ -59,6 +61,67 @@ impl From<Handle> for WriteHandle {
...
+        if result != OakStatus::Ok {
+            panic!("Failed to clone handle: {:?}", result);
+        }
+        cloned_handle
+    }
+
+    fn drop_handle(handle: Handle) {
+        // The channel may have already been closed, or it may be invalid due to extracting it for
```
tiziano88, line 117: I think we should check for invalid handle here (and also in clone) and deal with that case explicitly, whether it is about ignoring it, or producing a more precise error message.

tiziano88 (approved):

oak_io/src/handle.rs
wildarch: It gets printed already: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

> Panics if the value is an Err, with a panic message including the passed message, **and the content of the Err**.

oak_io/src/handle.rs
wildarch: Sure, maybe we can ignore `INVALID_HANDLE` in `drop`, and `panic!` with a more helpful message in `clone`?

wildarch:
> LGTM. But can we have tests for it at this stage?

I think unit tests would not add much value here, but I will try and set up some tests in `abitest`

oak_io/src/handle.rs
tiziano88: Sure, but I don't know if the Err contains the status that failed to be converted :)

wildarch:
I had to jump through a few hoops to set up ABI tests, mainly because we cannot enable the `linear-handles` feature in the existing modules of the abitest. I have added a new module that gets started and called by `module_0` to get around that. It adds quite a bit of code, but it seems like it is worth it: I actually discovered a bug in the previous code. 
The `channel_close` in the `Drop` impl was inside a `trace!` macro call, which meant that it was only called if a logger was set up (that would have been very hard to debug in an application, whoops!).

examples/abitest/module_0/rust/src/lib.rs
```diff
@@ -858,6 +863,45 @@ impl FrontendNode {
...
+
+        let (result_wh, result_rh) =
+            oak::channel_create("linear_handle_result", &Label::public_untrusted()).unwrap();
+
+        // The linear_handles module expects to read a single init message with exactly one handle:
+        // a write handle to return the result message in.
+        oak::channel_write(init_wh, &[], &[result_wh.handle])
+            .expect("Failed to write result handle");
+
+        // the linear_handles module should return a single result message (without handles), its
```
rbehjati, line 887: nit:
```suggestion
        // The linear_handles module should return a single result message (without handles), its
```

examples/abitest/module_linear_handles/rust/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    // Channel is now orphaned
+    assert_eq!(channel_read(&r), Err(OakStatus::ErrChannelClosed));
+}
+
+#[no_mangle]
+pub extern "C" fn linear_handles_oak_main(init_handle: u64) {
+    let _ = std::panic::catch_unwind(|| {
+        // Set up a panic handler that reports errors to the parent node
+        std::panic::set_hook(Box::new(move |panic_info| {
```
rbehjati, line 79: Just checking: Are you defining this hook instead of using [`set_panic_hook` in the SDK](https://github.com/project-oak/oak/blob/a5dcfe403e2a0b4d6273c70d6973b57955f14dff/sdk/rust/oak/src/lib.rs#L476) because the `SDK` does not support linear-handles, and so cannot be used in this test?

examples/abitest/module_linear_handles/rust/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+fn test_drop_closes_handle() {
+    let (w, r) = channel_create().expect("Failed to create channel");
+    assert_eq!(channel_read(&r), Err(OakStatus::ErrChannelEmpty));
+
+    // Explicitly drop the handle
+    core::mem::drop(w);
+
+    // Channel is now orphaned
```
rbehjati, line 71: Would the channel be closed if there is a clone of the write handle?

rbehjati (approved):
Nice. Thanks!

oak_io/src/handle.rs
wildarch: Oops, `from_i32` returns an `Option`, not a `Result`. Thanks for pointing this out, will fix! 😄

examples/abitest/module_linear_handles/rust/src/lib.rs
wildarch: Yes, that's exactly right 😄

examples/abitest/module_linear_handles/rust/src/lib.rs
wildarch: That specific channel would not be closed at that point if there were a clone of the write handle. That cloned handle could still be used to write data into the channel. The channel is closed only once all writers are dropped (and the channel is empty).

wildarch:
Another wrinkle: when running checks like clippy, cargo compiles the whole examples workspace in one go, and with apparently a single feature configuration for dependency in the tree. This caused the `linear-handles` feature to be enabled for other examples, failing to compile the SDK used there.
I've now added an additional `no-linear-handles` feature which suppresses the `linear-handles` feature, i.e. if both are enabled, linear handles are not available.
This is only an issue when running the workspace level checks, when compiling individual crates there is no problem.

The `no-linear-handles` feature is a bit hacky, but it's the best solution I could come up with. This should all be much nicer once we land linear handles support in the SDK.

PR #1856, opened 2021-02-01
Use user's identity as the request label in the Trusted Database example

rbehjati:
Leftover from #1707! I had not realized the request label was meant to be set to the user's identity!


tiziano88 (approved):
Great, thanks!

PR #1857, opened 2021-02-01
Refactor JS client to remove `@grpc/proto-loader` dependency 

juliettepretot:
Removes the dependency on `@grpc/proto-loader`. 
`@grpc/proto-loader` is an (imho cursed) abstraction around `protobufjs`. It's supposed to aid when using `@grpc/grpc-js` with `protobufjs`, however due it's flawed architecture it mostly just makes things a lot harder to work with. (see comment below).

Also allows us to finally replace callbacks with promises 

# Checklist

      construction.


examples/hello_world/client/nodejs/index.js
```diff
@@ -17,31 +17,24 @@
...
 async function main() {
-  const [oaklabelDefinition, helloWorldDefinition] = await Promise.all([
-    // We use two different libs with a similar APIs to load proto files,
-    // which due to the design of the `@grpc/grpc-js` lib.
-    // `@grpc/proto-loader` is a wrapper around `protobufjs` that adds
-    // functionality required for working with `grpc-js`, but omits others.
-    // Hence we use grpcProtoLoader for gRPC services, and protobufjs
-    // for all other protos.
-    protobufjs.load(OAK_LABEL_PROTO_PATH),
-    grpcProtoLoader.load(SERVICE_PROTO_PATH),
```
juliettepretot, line 39: Stumbled upon this when revisiting this code to add challenge-response auth. Imho this makes things quite confusing, so I really wanted to fix it.

juliettepretot: Unrelated to this PR, but imho the architecture of `@grpc/proto-loader` should probably be changed to just export the underlying helper functions, instead of needlessly wrapping a whole protobufjs instance. That is if the abstraction is even needed, which imho it probably isn't.

examples/hello_world/client/nodejs/index.js
```diff
@@ -53,31 +46,34 @@ async function main() {
...
-    // Callback invoked with the response.
-    (error, response) => {
-      if (error) {
-        console.error(error);
-        process.exit(1);
-      } else {
-        console.log(response.reply);
-        process.exit(0);
-      }
+  const HelloWorld = helloWorldService.create(
```
tiziano88, line 56: (nit) is the convention to start identifiers with upper or lower case letter?

examples/hello_world/client/nodejs/index.js
```diff
@@ -53,31 +46,34 @@ async function main() {
...
+    (method, requestData, callback) => {
+      grpcClient.makeUnaryRequest(
+        `/oak.examples.hello_world.HelloWorld/${method.name}`,
+        (arg) => arg,
+        (arg) => arg,
+        requestData,
+        getGrpcMetadata(),
+        callback
+      );
```
tiziano88, line 65: I'm not sure I understand what's going on here, is this the normal way of making a gRPC call? The `helloWorldService.create` sounds a bit awkward, and also not sure what the `(arg) => arg` things are.

tiziano88 (approved):

examples/hello_world/client/nodejs/index.js
juliettepretot: Generally only constructors should be capitalized, this was a bit inconsistent indeed, thx for catching. :)

examples/hello_world/client/nodejs/index.js
juliettepretot: Good point, I added comments to better explain what is going on here.

tiziano88 (approved):

juliettepretot:
(forced pushed to rebase CI fix from main onto this branch)

Issue #1859, opened 2021-02-02
Investigate creating a JS client equivalent to oak_sign


juliettepretot:
At the moment in non-Rust clients we have to manually create the client keys, and sign the challenge response via the `oak_sign` CLI. It's an extra layer of complexity, and means we have to hardcode the challenge in the Oak runtime (atm it's just `"oak-challenge"`) thereby compromising the auth security of challenge-response auth, and allowing for replay attacks. 

A native JS way of doing this would be nice. 

Related to #1853, #1855, #1851

tiziano88:
FYI: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto

juliettepretot:
implemented :]

tiziano88:
Does the current implementation also work from browser, or only in Node.js?

juliettepretot:
All the juicy bits, yes. :) The one exception is the hash function. The API is different for Node.js and browsers.

It's easy to adjust:

https://github.com/project-oak/oak/blob/3deab43be26dff143a5f61ca0ab8ab21b7a94e93/examples/hello_world/client/nodejs/oakSign.js#L45

s/`createHash('sha256').update(input).digest()`/`await globalThis.crypto.subtle.digest('sha256', input)`/

If you think it'd make good, I can add a small browser client to the `hello_world` example. :)

PR #1870, opened 2021-02-04
Implement challenge reponse signing in JS

juliettepretot:
Uses https://github.com/paulmillr/noble-ed25519, a dependency free ed25519 JS implementation. 
The [WebCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) doesn't yet support ed25519, but `noble-ed25519` uses it under the hood when possible. 

One noteworthy thing is that the Node.js built in crypto lib technically does support `ed25519`, however for a reason I haven't yet understood, it appears to work with 44byte public keys, and doesn't work with the runtime. 

Thinking about building an example that works in the browser and via http after.  

# Checklist

      construction.

### DIFF
--- a/examples/hello_world/client/nodejs/index.js
+++ b/examples/hello_world/client/nodejs/index.js
@@ -17,6 +17,7 @@
 const fs = require('fs');
 const protobufjs = require('protobufjs');
 const grpc = require('@grpc/grpc-js');
+const oakSign = require('./oakSign.js');
 
 const REPOSITORY_ROOT = `${__dirname}/../../../..`;
 const CERT_PATH = `${REPOSITORY_ROOT}/examples/certs/local/ca.pem`;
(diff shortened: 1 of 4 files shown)


tiziano88 (approved):
Very nice, I like how minimal it ends up being! Does it actually work with the Oak Runtime? i.e. does everything (public key, signatures, etc.) line up correctly even in terms of serialization? If so, that's great news!

juliettepretot:
@tiziano88 Yeah it does :).
The only part I'm not yet happy with is that we hardcode the challenge, instead of dynamically fetching it from the Oak runtime, thereby opening us up to replay attacks. 

However, it works the same way in our rust clients, a separate issue from the JS implementation.

juliettepretot:
xref #1859

examples/hello_world/client/nodejs/index.js
```diff
@@ -28,42 +29,19 @@ const SERVICE_PROTO_PATH = `${REPOSITORY_ROOT}/examples/hello_world/proto/hello_
...
-  'base64'
-);
+const OAK_CHALLENGE = 'oak-challenge';
 
 async function main() {
-  const protos = await protobufjs.load([
-    OAK_LABEL_PROTO_PATH,
-    OAK_IDENTITY_PROTO_PATH,
-    SERVICE_PROTO_PATH,
+  const [protos, keyPair] = await Promise.all([
```
rbehjati, line 35: I know absolutely nothing about JS... so I am curious, could not you leave the declaration of `protos`  as before, and add a separate declaration for `keyPair`? I guess it has something to do with `Promise.all`. What does it do? Why is it needed?

tiziano88: I would imagine if you do one promise after the other it would work, but the second one would not start until the first one is resolved, this looks like it starts both in parallel, so it resolves when both of them resolve. (disclaimer: I also don't know much about JS, but I guess this is the same for any language that has async support, including Rust)

rbehjati:
> However, it works the same way in our rust clients, a separate issue from the JS implementation.

Yes. Using a dynamically generated challenge is the next step of the increment in #1357. We have decided to postpone it, until there is need for it.

examples/hello_world/client/nodejs/index.js
rbehjati: Makes sense. Thanks.

juliettepretot: Yup, what @tiziano88 said :)

PR #1871, opened 2021-02-04
Prove partially-secret PNI up through unwind

bMacSwigg:
Partially-secret means a node can only see labels that flow to its
label (i.e. are at most as secret as the node is), so labels are
partially secret. This commit makes progress towards proving
possibilistic non-interference with a partially-secret runtime model.

Much of the added content is redundant and might be refactorable. For
simplicity, I've just created new PS files rather than spending too
much effort trying to refactor the existing files to support both
runtime models.

# Checklist

      construction.


bMacSwigg:
This PR looks much meatier than it actually is, since many of the PS files are very near copies of the original non-PS files. Git doesn't seem to support explicitly marking a file as branched/copied from another file. Please let me know if you would like me to annotate what I've changed or otherwise split this up to make the changes clearer (e.g. I could create a PR that just copies these files directly first, then a follow-up PR that changes them).

experimental/ni_coq/vfiles/TraceLowEqPS.v
```diff
@@ -0,0 +1,40 @@
...
+(* This is a copy of TraceLowEq but for partially-secret labels.
+   This version only replicates the possibilistic NI definition.
+*)
+
+(* This is a straightforward definition of trace low-equivalence
+    Roughly, it says that 
+    t1 =L t2 <-> forall i, t1[i] =L t2[i].
+*)
+
+Inductive trace_low_eq {s_leq: @low_eq_t state}{e_leq: @low_eq_t event_l}:
```
aferr, line 33: Are you sure you need multiple copies of this file? This one takes arguments that specify the specific definitions used for low-equivalences over states and events to hopefully work for exactly the proof you're working on. I think it should be possible to use TraceLowEq.v (not PS) in the top-level security condition, so I think you don't need this one.

experimental/ni_coq/vfiles/UnwindPS.v
```diff
@@ -0,0 +1,286 @@
...
+    inversion H0; subst;
+    pose proof (ord_anti ell top ltac:(eauto) ltac:(eauto));
+    congruence.
+    - specialize (H0 nid).
+    cbv [node_state_low_proj low_proj fnd State.lbl 
+        State.obj] in *.
+    simpl in *. erewrite upd_neq; auto. erewrite upd_neq; auto.
+Qed.
+
+Ltac label_cases :=
```
aferr, line 70: I don't think anything about these tactics specific to either of the definitions. Would it work if these were taken out of the two Unwind{PS}.v variants and put into Tactics.v (or some other file that is common to both?)

experimental/ni_coq/vfiles/TraceTheoremsPS.v
```diff
@@ -0,0 +1,89 @@
...
+            invert_clean H
+          | H : _ :: _ = _ :: _ |- _ =>
+            invert_clean H
+          end
+        | congruence
+        | destruct_match ].
+(* General-purpose tactic that simplifies and solves simple goals. *)
+Local Ltac crush := repeat crush_step.
+
+Theorem no_steps_from_empty: forall t,
```
aferr, line 30: One way to refactor this file and the old version could be to make the definitions of these theorems take the semantics of choice as an argument (e.g. here it's step_system_ev_t from the PS file). Then we could state the theorem for each of the two semantics by just plugging it in. For the cases where the proof body is more than just a tactic or two, the body could be made a new tactic and that tactic could just be repeated for each version of the body.

(I don't think doing this is crucial for this PR though)

experimental/ni_coq/vfiles/UnwindPS.v
```diff
@@ -0,0 +1,286 @@
...
+    autounfold with semutils. unfold fnd. intros.
+    rewrite H. congruence.
+Qed.
+
+(* STOPPED HERE *)
+
+Theorem set_call_unwind: forall ell id c s1 s2,
+    state_low_eq ell s1 s2 ->
+    state_low_eq ell (s_set_call s1 id c) (s_set_call s2 id c).
+Proof.
```
aferr, line 219: This is one where it looks like the proof could be simplified (though not crucial for this PR).

experimental/ni_coq/vfiles/LowProjPS_Thms.v
```diff
@@ -228,12 +233,53 @@ Proof. congruence. Qed.
...
 Proof. congruence. Qed.
 
-End low_equivalence.
+Hint Unfold obj lbl fnd: structs.
 
-Hint Unfold state_low_eq chan_state_low_eq node_state_low_eq event_low_eq
-    chan_low_eq node_low_eq low_eq state_low_proj chan_state_low_proj
-    node_state_low_proj event_low_proj chan_low_proj node_low_proj
-    low_proj: loweq.
+Lemma state_low_eq_implies_node_lookup_eq ell s1 s2 :
```
aferr, line 238: It might be useful to factor some of these definitions so that they take each of the state_lowe_eq and each of the low projection definitions as arguments. Then state and prove each of the variants separately, but using the same body where possible. Where the body is actually the same, the body could be a tactic and the tactic just gets called twice.

(Not crucial for this PR)

aferr (commented):
Would you mind also leaving notes about the following in both comments in the code and on the PR about:

- which theorems are true with one set of definitions but not the other?
- which theorems have proofs that are different between the two versions?

Even if we don't have a good way of refactoring yet, I think tracking that could help refactor. It might also be useful/interesting detail to have if/when we reach a point where we are putting a paper together.

experimental/ni_coq/vfiles/TraceLowEqPS.v
bMacSwigg: Nope, you're right. Removed this.

bMacSwigg:
I removed some unnecessary changes:
- Got rid of TraceLowEqPS, which was an exact copy of TraceLowEq and didn't serve any purpose
- Removed the NIUtilTheorems import in UnwindPS, since LowProjPS_Thms and Unfold together are enough now. This let me get rid of the `LowProjPS_Thms.state_upd_node_unobs` specifying that I was doing.

I also added some comments that explain better what the differences are for each file. In summary:
- EvAugSemanticsPS: Only changed the import (to RuntimeModelPS), otherwise identical
- LowProjPS_Thms: For each of the theorems added, they have an identical statement & body, but they use the PS `state_low_proj` and the PS autounfold Hints. Specifically, these are: `state_low_eq_implies_node_lookup_eq`, `state_low_eq_implies_chan_lookup_eq`, `node_state_proj_index_assoc`, `node_state_proj_index_assoc_form2`, `chan_state_proj_index_assoc`, `chan_state_proj_index_assoc_form2`, and `state_upd_node_unobs`.
- Tactics: Added or modified necessary Tactics for PS versions
- TraceTheoremsPS: Only changed the imports to RuntimeModelPS and EvAugSemanticsPS, otherwise identical
- UnwindPS: Added top_is_top_co to help with PS theorems `state_upd_chan_unwind` and `state_chan_append_labeled_unwind`

experimental/ni_coq/vfiles/TraceTheoremsPS.v
bMacSwigg: Do we think that both versions of the runtime model will be sticking around? If so then it makes sense to do that refactor, but if not then I think it might just be more churn to factor all of these definitions out, then plug them back in again once we get rid of the non-PS model.

For making the proof body its own tactic: Although the proof bodies generally look identical, I suspect that Coq sees them as distinct because they use different things that happen to be named the same (both for the base semantics, and for intermediate theorems that have been proven separately for the different semantics). I'm open to trying this approach, but I wouldn't be surprised if Coq objects/can't figure out which model is being referred to when the Tactic is written.

experimental/ni_coq/vfiles/UnwindPS.v
bMacSwigg: Good point, done.

experimental/ni_coq/vfiles/TraceTheoremsPS.v
aferr: OK I don't think this is a big deal. It's unclear that both will stick around, but they might in case we want to write a paper and have both versions in a code release. I think having the best code reuse is maybe not the biggest priority.

I don't think we'd run into issues with coq being able to differentiate different theorems. The refactor I was trying to describe was something like:

```
Definition no_steps_end_in_one semantics: forall t t',
    length t' = 1 ->
~ (semantics t t').

Ltac no_steps_end_in_one_proof :=. .... <whatever this is >

Theorem no_steps_one_regular: no_steps_end_in_one system_step_ev.
Proof. no_steps_end_in_one_proof. Qed.

Theorem no_steps_one_ps: no_steps_end_in_one system_step_ev_ps.
Proof. no_steps_end_in_one_proof. Qed.
```

At a glance, it looks like it should be possible to use the same tactic for both proofs. The actual theorem for each of the different versions gets a different name, so it seems like this should not cause problems where these get used.

But yes, let's not spend time on this right now.

aferr (approved):
LGTM!

PR #1872, opened 2021-02-05
Leaks

rbehjati:
Closes `sender` and `receiver` channels in various places, including in the examples. 

Fixes the problem of orphaned channels reported in #1866, as well as #1718. 
The problem unused/leaked nodes in #1866 will be fixed separately. For that we need a new function similar to `run_command_loop `
https://github.com/project-oak/oak/blob/61ecb0b2a07eac2f39cadee196c4463d4ce694ca/sdk/rust/oak/src/lib.rs#L564
except that it would handle only one command.

# Checklist

      construction.

### DIFF
--- a/examples/chat/module/rust/src/lib.rs
+++ b/examples/chat/module/rust/src/lib.rs
@@ -118,7 +118,7 @@ impl oak::CommandHandler for Router {
         // The router node has a public confidentiality label, and therefore cannot read the
         // contents of the request of the invocation (unless it happens to be public as well), but
         // it can always inspect its label.
-        match (&command.receiver, &command.sender) {
+        let result = match (&command.receiver, &command.sender) {
             (Some(receiver), Some(sender)) => {
                 let label = receiver.label()?;
                 let grpc_response_writer = oak::grpc::ChannelResponseWriter::new(sender.clone());
@@ -150,7 +150,9 @@ impl oak::CommandHandler for Router {
             _ => {
                 anyhow::bail!("received malformed gRPC invocation");
             }
-        }
+        };
+        command.close()?;
+        result
     }
 }
 
--- a/examples/http_server/module/src/lib.rs
+++ b/examples/http_server/module/src/lib.rs
@@ -266,7 +266,8 @@ impl CommandHandler for RedirectHelper {
     type Command = RedirectInvocation;
 
     fn handle_command(&mut self, invocation: RedirectInvocation) -> anyhow::Result<()> {
-        let request = invocation.request_receiver.unwrap().receive()?;
+        let request = invocation.request_receiver.as_ref().unwrap().receive()?;
+        invocation.request_receiver.unwrap().close()?;
         let client_invocation = invocation.http_invocation_source.unwrap();
         let uri = request.uri.parse::<http::Uri>()?;
 
--- a/oak_runtime/src/node/http/server.rs
+++ b/oak_runtime/src/node/http/server.rs
@@ -38,7 +38,10 @@ use oak_abi::{
     proto::oak::application::HttpServerConfiguration,
     OakStatus,
 };
-use oak_io::{handle::ReadHandle, OakError, Receiver};
+use oak_io::{
+    handle::{ReadHandle, WriteHandle},
+    OakError, Receiver,
+};
 use oak_services::proto::oak::encap::{HeaderMap, HttpRequest, HttpResponse};
 use std::{io, net::SocketAddr, pin::Pin};
 use tokio::{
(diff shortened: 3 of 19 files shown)


examples/private_set_intersection/main_module/rust/src/lib.rs
```diff
@@ -81,6 +82,7 @@ impl oak::WithInit for Router {
             LogInit {
                 log_sender: Some(log_sender),
             },
+            true,
```
rbehjati, line 85: This is the only place where the `keep_sender_open` flag I added needs to be `true`, and I am not sure why. 
Changing this to `false` results in an orphaned channel (`handler-in(5)`). 
Screen Shot 2021-02-10 at 18 59 25
I don't understand why this happens. I expected the following line to close the read-half of the `handler-in` channel. 
https://github.com/project-oak/oak/blob/4da10691dfe867a022d2d37c7b991d5bdaaf3673/sdk/rust/oak/src/lib.rs#L754

sdk/rust/oak/src/io/mod.rs
```diff
@@ -115,6 +116,7 @@ fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
     sender: Sender<oak_io::InitWrapper<Init, Command>>,
     init: Init,
     command_channel_label: &Label,
+    keep_sender_open: bool,
```
tiziano88, line 119: If we do need this param, could you create a dedicated self-documenting enum? Seeing `true` and `false` at each call place is not particularly self-explanatory.

sdk/rust/oak/src/io/mod.rs
```diff
@@ -123,6 +125,9 @@ fn send_init<Init: Encodable + Decodable, Command: Encodable + Decodable>(
         command_receiver,
     };
     sender.send(&init_wrapper)?;
+    if !keep_sender_open {
+        sender.close()?;
+    }
```
tiziano88, line 130: I'm not sure what the advantage of coupling this functionality here is, vs just letting the caller cloe the channel normally if wanted?

I mean, now this function does two things: send the init **and** (sometimes) close the channel, which was previously two independent steps that can be combined by the caller one after the other. If the param is just to save some typing, personally I would advise against it.

sdk/rust/oak/src/grpc/mod.rs
```diff
@@ -37,6 +37,21 @@ pub mod server;
 pub type Result<T> = std::result::Result<T, rpc::Status>;
 pub type Invocation = crate::proto::oak::invocation::GrpcInvocation;
 
+impl Invocation {
+    pub fn close(&self) {
+        self.receiver
+            .as_ref()
```
tiziano88, line 43: Why are all the `as_ref()` necessary? I'm not very familiar with it.

sdk/rust/oak/src/io/mod.rs
rbehjati: Removed. It was part of my experimentation :)

sdk/rust/oak/src/io/mod.rs
rbehjati: The caller is `entrypoint_node_create`, which also creates the `sender` channel.  The channel was never closed previously. It is now being closed in `entrypoint_node_create`.

sdk/rust/oak/src/grpc/mod.rs
rbehjati: Fixed.

examples/chat/module/rust/src/lib.rs
```diff
@@ -150,7 +150,9 @@ impl oak::CommandHandler for Router {
             _ => {
                 anyhow::bail!("received malformed gRPC invocation");
             }
-        }
+        };
+        command.close();
```
tiziano88, line 154: Is it intentional to ignore errors due to closing channels (here and elsewhere)? does it not work to just use `?` for both, in a linear fashion?

sdk/rust/oak/src/io/mod.rs
```diff
@@ -147,10 +149,20 @@ pub fn forward_invocation(
...
-                    .context("Couldn't forward invocation")
+                    .context("Couldn't forward invocation");
+                // Close the channels.
+                invocation.close();
+                result
             } else {
+                // Close the receiver channel separately.
+                invocation
+                    .receiver
+                    .as_ref()
```
tiziano88, line 162: Is this particular as_ref still needed?

oak_runtime/src/node/http/server.rs
```diff
@@ -274,6 +267,32 @@ impl Node for HttpServerNode {
     }
 }
 
+/// Reads the [`WriteHandle`] (to be used for sending new invocations) from a startup channel.
+/// Returns an error if the startup channel couldn't be read, or if the initial message is
+/// invalid (it must be an encoded [`HttpInvocationSender`]).
+fn get_invocation_channel(
+    runtime: &RuntimeProxy,
+    startup_handle: oak_abi::Handle,
```
tiziano88, line 275: Should this be of type ReadHandle, for symmetry with the other method?

tiziano88 (approved):

rbehjati (commented):
Thanks for the review @tiziano88!

examples/http_server/module/src/lib.rs
```diff
@@ -266,7 +266,8 @@ impl CommandHandler for RedirectHelper {
     type Command = RedirectInvocation;
 
     fn handle_command(&mut self, invocation: RedirectInvocation) -> anyhow::Result<()> {
-        let request = invocation.request_receiver.unwrap().receive()?;
+        let request = invocation.request_receiver.as_ref().unwrap().receive()?;
```
tiziano88, line 269: Is this `as_ref` needed? BTW I'm just asking because I don't uderstand what it does, (I don't remember using `as_ref` on anything other than strings) could you give me a summary?

sdk/rust/oak/src/io/mod.rs
```diff
@@ -103,7 +103,9 @@ pub fn entrypoint_node_create<
 ) -> Result<Sender<Command>, oak_io::OakError> {
     let node_config = &crate::node_config::wasm(wasm_module_name, T::ENTRYPOINT_NAME);
     let init_sender = node_create(name, label, node_config)?;
-    send_init(init_sender, init, label)
+    let result = send_init(&init_sender, init, label);
+    init_sender.close()?;
```
tiziano88, line 107: I understand that this is meant to close the channel even if the sending failed, but I wonder in practice when that may happen. Would it be the case if the labels are not correct, for instance? Is this fine, or should we think of some other way of doing it to make sure channels get closed? Apart from the linear handles proposal, which I guess would solve it anyways.

tiziano88 (approved):

examples/http_server/module/src/lib.rs
rbehjati: This one is needed, to borrow `invocation.request_receiver` instead of moving it. It was ok to move previously, because we did not close channel before.

sdk/rust/oak/src/io/mod.rs
rbehjati: I tried to implement `Drop` for some structs, but I could not make it to work. I cannot think of any other general solution.  Perhaps we should revisit these changes after the work on linear handles is completed.

PR #1873, opened 2021-02-08
Ignore RUSTSEC-2021-0020

rbehjati:

### DIFF
--- a/oak_runtime/src/node/http/server.rs
+++ b/oak_runtime/src/node/http/server.rs
@@ -56,6 +56,9 @@ use oak_abi::proto::oak::identity::SignedChallenge;
 use prost::Message;
 use tokio_rustls::TlsAcceptor;
 
+// Workaround for https://rust-lang.github.io/rust-clippy/master/index.html#borrow_interior_mutable_const.
+static TRANSFER_ENCODING: http::header::HeaderName = http::header::TRANSFER_ENCODING;
+
 /// Checks that port is not reserved (i.e., is greater than 1023).
 fn check_port(address: &SocketAddr) -> Result<(), ConfigurationError> {
     if address.port() > 1023 {
@@ -297,6 +300,7 @@ struct HttpRequestHandler {
 
 impl HttpRequestHandler {
     async fn handle(&self, req: Request<Body>) -> anyhow::Result<Response<Body>> {
+        let req = validate_request(req)?;
         let request = to_oak_http_request(req).await?;
         match get_oak_label(&request) {
             Ok(oak_label) => {
@@ -355,6 +359,23 @@ impl HttpRequestHandler {
             runtime: self.runtime.clone(),
             response_receiver,
         })
+    }
+}
+
+/// Check if the request contains a `TRANSFER_ENCODING` header, and reject the request in that case
+/// by returning an error.
+// TODO(#1874): Remove when tonic and hyper versions are updated.
+fn validate_request(req: Request<Body>) -> anyhow::Result<Request<Body>> {
+    if req
+        .headers()
+        .get(&TRANSFER_ENCODING.as_str().to_string())
+        .is_some()
+    {
+        Err(anyhow!(
+            "Requests containing TRANSFER_ENCODING headers are not allowed."
+        ))
+    } else {
+        Ok(req)
     }
 }
 
--- a/examples/deny.toml
+++ b/examples/deny.toml
@@ -16,10 +16,8 @@ notice = "deny"
 ignore = [
   # TODO(#1267): Remove when mio no longer depends on net2.
   "RUSTSEC-2020-0016",
-  # TODO: Remove when rusoto-* no longer depends on dirs.
-  "RUSTSEC-2020-0053",
-  # TODO: Remove when rusoto-* no longer depends on stdweb (via time).
-  "RUSTSEC-2020-0056"
+  # TODO(#1874): Remove when tonic and hyper versions are updated.
+  "RUSTSEC-2021-0020",
 ]
 
 # Deny multiple versions unless explicitly skipped.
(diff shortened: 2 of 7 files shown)


oak_runtime/src/node/http/server.rs
```diff
@@ -358,6 +362,21 @@ impl HttpRequestHandler {
     }
 }
 
+/// Check if the request contains a `TRANSFER_ENCODING` header, and reject the request in that case
+/// by returning an error.
+fn validate_request(req: &HttpRequest) -> anyhow::Result<()> {
```
rbehjati, line 367: @conradgrobler Is this the right way to reject the requests with a TRANSFER_ENCODING header?

oak_runtime/src/node/http/server.rs
```diff
@@ -358,6 +362,21 @@ impl HttpRequestHandler {
     }
 }
 
+/// Check if the request contains a `TRANSFER_ENCODING` header, and reject the request in that case
+/// by returning an error.
```
tiziano88, line 366: This can be removed once https://rustsec.org/advisories/RUSTSEC-2021-0020.html does not apply any more, right? which seems to be fixed in some newer versions already. Could you add a TODO and create an issue to remove the workaround?

oak_runtime/src/node/http/server.rs
```diff
@@ -358,6 +362,21 @@ impl HttpRequestHandler {
...
+fn validate_request(req: &HttpRequest) -> anyhow::Result<()> {
+    if req
```
tiziano88, line 368: Just checking: are we not using HTTP2 only? It seems that's not affected: https://github.com/hyperium/hyper/security/advisories/GHSA-6hfq-h8hq-87mf

oak_runtime/src/node/http/server.rs
tiziano88, line 367: How about making this accept an `HttpRequest` (by value) and also return a `Result<HttpRequest>`, so that there is no risk that the caller misses it? I think the `Result<()>` return type is always a bit too easy to miss / misuse.

oak_runtime/src/node/http/server.rs
```diff
@@ -358,6 +362,21 @@ impl HttpRequestHandler {
...
+        .headers
+        .as_ref()
+        .and_then(|map| map.headers.get(&TRANSFER_ENCODING.as_str().to_string()))
+        .is_some()
```
tiziano88, line 372: Why is all this necessary as opposed to just

```rust
req.headers.get(TRANSFER_ENCODING).is_some()
```

?

oak_runtime/src/node/http/server.rs
```diff
@@ -56,6 +56,9 @@ use oak_abi::proto::oak::identity::SignedChallenge;
 use prost::Message;
 use tokio_rustls::TlsAcceptor;
 
+// Workaround for https://rust-lang.github.io/rust-clippy/master/index.html#borrow_interior_mutable_const.
+static TRANSFER_ENCODING: http::header::HeaderName = http::header::TRANSFER_ENCODING;
```
conradgrobler, line 60: Is it needed to define a constant in the code here that points directly to a constant in another library?

oak_runtime/src/node/http/server.rs
rbehjati: Yes, it is already fixed in `hyper:0.14.4`, but `tonic` is not updated. Do we have an issue for tracking tonic's version? Because we are already behind.

oak_runtime/src/node/http/server.rs
conradgrobler: It looks right, apart from seeming a bit more complex than needed as @tiziano88 suggested below.

oak_runtime/src/node/http/server.rs
rbehjati: In HttpServerNode we are using HTTP/1.1.

oak_runtime/src/node/http/server.rs
conradgrobler: +1. I would even consider doing this validation in `to_oak_http_request` seeing that it is already returning a result. This would also mean that you can use the http library's headermap to check for existence of the header before mapping it to the proto headermap.

rbehjati: I like it. Thanks. 
I also changed it to take the original request as input. But I am keeping it separate from `to_oak_http_request` to be explicit that a validation step is also happening.

oak_runtime/src/node/http/server.rs
rbehjati: Yes. Because of clippy's `borrow_interior_mutable_const` error. I could not find a better solution.

oak_runtime/src/node/http/server.rs
rbehjati: Done! Changed the signature of `validate_request` to take the original `http::Request` instead of our internal representation, in which `headers` is `Option<HeaderMap>`.

oak_runtime/src/node/http/server.rs
tiziano88: I don't think we do, we just update manually or when dependabot decides to :)

tiziano88 (approved):

oak_runtime/src/node/http/server.rs
rbehjati: Created an issue for it to be able to add a TODO comment.

oak_runtime/src/node/http/server.rs
tiziano88: @conradgrobler note that strictly speaking this is just a static binding, not a const. It is "instantiating" the const value once, which otherwise gets inlined in every call place (hence the warning, since instances of const items are always distinct).

tiziano88 (approved):

PR #1884, opened 2021-02-12
Add linear-handles feature to SDK.

wildarch:
# Checklist

      construction.

### DIFF
--- a/sdk/rust/oak/Cargo.toml
+++ b/sdk/rust/oak/Cargo.toml
@@ -5,6 +5,11 @@ authors = ["Tiziano Santoro <tzn@google.com>"]
 edition = "2018"
 license = "Apache-2.0"
 
+[features]
+default = ["no-linear-handles"]
+no-linear-handles = ["oak_io/no-linear-handles"]
+linear-handles = ["oak_io/linear-handles"]
+
 [dependencies]
 anyhow = "*"
 byteorder = "*"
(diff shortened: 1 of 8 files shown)


sdk/rust/oak/Cargo.toml
```diff
@@ -5,14 +5,19 @@ authors = ["Tiziano Santoro <tzn@google.com>"]
 edition = "2018"
 license = "Apache-2.0"
 
+[features]
+default = ["no-linear-handles"]
+no-linear-handles = ["oak_io/no-linear-handles"]
```
rbehjati, line 10: Are both `no-linear-handles` and `linear-handles` still needed now that the features are added to `SDK`?

rbehjati (approved):
LGTM!

sdk/rust/oak/Cargo.toml
wildarch: I'm afraid we do. Some of the examples are not compatible with the 'linear-handles' feature (I think it's just abitest and injection), and will cause a build failure for the examples workspace without having a `no-linear-handles` feature to cancel out the `linear-handles` one.

Once those examples are fixed, we can get rid of this feature 😄

rbehjati: I see. Thanks for the clarification :)

PR #1888, opened 2021-02-16
Update github CI

ipetr0v:
This change updates github CI actions to match Google Cloud Build.
Required for https://github.com/project-oak/oak/pull/1819 in order to disable GCB CI and completely move to github CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,6 +9,7 @@ on:
 jobs:
   runner:
     runs-on: ubuntu-latest
+    timeout-minutes: 120
     # We use the same job template, but parametrize the actual command to be passed to the runner
     # binary using the matrix strategy, so that we get the commands running in parallel.
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -50,5 +51,52 @@ jobs:
           docker pull gcr.io/oak-ci/oak:latest
           docker build --pull --cache-from=gcr.io/oak-ci/oak:latest --tag=gcr.io/oak-ci/oak:latest .
 
+      # Init .git repository used by check_generated.
+      # Workaround for https://github.com/GoogleCloudPlatform/cloud-builders/issues/236
```
tiziano88, line 55: So is this also an issue for GH Actions?

.github/workflows/ci.yaml
```diff
@@ -50,5 +51,52 @@ jobs:
...
       - name: Run command
         run: ./scripts/docker_run ./scripts/runner ${{ matrix.cmd }}
+
+      - name: Generate root CA certs
+        run: ./scripts/docker_run ./scripts/generate_root_ca_certs
+
+      # Check whether any of the previous steps resulted in file diffs that were not checked in or
+      # ignored by git.
+      - name: Git check diff
+        run: ./scripts/docker_run ./scripts/git_check_diff
```
tiziano88, line 71: LG as long as this step actually works, please double check that it does fail if any auto-generated file does not match.

tiziano88 (approved):

scripts/git_check_diff
```diff
@@ -4,5 +4,8 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 # shellcheck source=scripts/common
 source "$SCRIPTS_DIR/common"
 
+echo TEST > ./examples/translator/proto/translator.pb.go
```
ipetr0v, line 7: Checked that the CI fails if any auto-generated file does not match

.github/workflows/ci.yaml
ipetr0v: Done

.github/workflows/ci.yaml
ipetr0v: Deleted

PR #1889, opened 2021-02-16
Fix default permissions file

ipetr0v:


runner/src/main.rs
```diff
@@ -509,11 +509,11 @@ fn run_example_server(
             // TODO(#396): Add `--oidc-client` support.
             format!("--application={}", application_file),
             match opt.server_variant {
-                ServerVariant::Base => format!("--permissions={}", permissions_file),
+                ServerVariant::Base | ServerVariant::Experimental => format!("--permissions={}", permissions_file),
```
rbehjati, line 512: Does it work? Experimental uses the `oak_unsafe` feature:
https://github.com/project-oak/oak/blob/87632c651af99ff49815e1da35185c1341d97e80/runner/src/main.rs#L323

But the `permissions` option is only available when `oak_unsafe` is disabled.

ipetr0v: hm,
Didn't know `experimental` used `oak_unsafe`
Last time I checked - it didn't

ipetr0v: Does it mean - we shouldn't use permissions with `experimental` at all ?

rbehjati: I guess you can remove `oak_unsafe` from `experimental`. I added it when I made the default set of features empty (before that `oak_unsafe` was included by default). If you remove `oak_unsafe`, I think you have to remove `oak_introspection_client` too. I think `experimental` should work without `oak_unsafe` if you don't need introspection. 

(Actually, I think `oak_introspection_client` does not make sense without `oak_unsafe`. So, maybe https://github.com/project-oak/oak/blob/87632c651af99ff49815e1da35185c1341d97e80/oak_runtime/Cargo.toml#L15 should be change; possibly in a separate PR.)

ipetr0v: I remember it didn't work without `oak_introspection_client`, but I need to re-check

rbehjati (approved):
Thanks!

runner/src/main.rs
```diff
@@ -320,7 +320,7 @@ fn build_server(opt: &BuildServer) -> Step {
                             // If building in coverage mode, use the default target from the host, and build
                             // in debug mode.
                             ServerVariant::Coverage => vec!["--features=oak_unsafe,oak_introspection_client".to_string()],
-                            ServerVariant::Experimental => vec!["--features=oak_attestation,oak_unsafe,oak_introspection_client".to_string(),
+                            ServerVariant::Experimental => vec!["--features=oak_attestation".to_string(),
```
rbehjati, line 323: `abitest` needs `oak_unsafe`, because it uses introspection functionality. So, if you don't want to have `oak_unsafe` with `experimental`, it cannot be the default server variant for running examples in the CI builds.

Why is the default set to `experimental`?

tiziano88:
FWIW, I think we should have the "default" feature set to correspond to what we would need to compile if we want a "production-ready" version of Oak. Then separately we can have an "experimental" feature that includes all the other features that either are not ready for production yet (e.g. linear handles), or are only intended for testing (e.g. logging from any node).

rbehjati:
> FWIW, I think we should have the "default" feature set to correspond to what we would need to compile if we want a "production-ready" version of Oak. Then separately we can have an "experimental" feature that includes all the other features that either are not ready for production yet (e.g. linear handles), or are only intended for testing (e.g. logging from any node).

I think I agree with this. But then it won't be possible to use a permissions file with `experimental`. For the CI, I suggest we add a new step to check `Base + permissions`.

ipetr0v:
> FWIW, I think we should have the "default" feature set to correspond to what we would need to compile if we want a "production-ready" version of Oak. Then separately we can have an "experimental" feature that includes all the other features that either are not ready for production yet (e.g. linear handles), or are only intended for testing (e.g. logging from any node).

I'll change the default server back to `base` in a separate PR.

ipetr0v:
Reverting this PR to only change the `permissions.toml` file, since running permissions in `abitest` shows errors:
```
I0217 11:34:42.345198 28588 abitest.cc:89]     Details: Error: abitest/module_0/rust/src/lib.rs:2477: expectation failed: `(left == right)`
  left: `0`,
 right: `5`
...
I0217 11:34:42.345508 28588 abitest.cc:86] [ FAIL ] GrpcClientUnaryMethod
I0217 11:34:42.345522 28588 abitest.cc:89]     Details: Error: abitest/module_0/rust/src/lib.rs:2433: expectation failed: `(left == right)`
  left: `Err(Status { code: 5, message: "Service connection failed", details: [] })`,
 right: `Ok(GrpcTestResponse { text: "test" })`
I0217 11:34:42.345557 28588 abitest.cc:86] [ FAIL ] LoggerPseudoNodePrivilege
I0217 11:34:42.345569 28588 abitest.cc:89]     Details: Error: abitest/module_0/rust/src/lib.rs:3170: expectation failed: `(left == right)`
  left: `Ok(())`,
 right: `Err(ErrPermissionDenied)`
```

PR #1890, opened 2021-02-16
Centralize node privilege assignment

tiziano88:
Note to reviewers: please double check I am not mixing up privileges for
the various node types, thanks!

# Checklist

      construction.

### DIFF
--- a/oak_runtime/src/node/mod.rs
+++ b/oak_runtime/src/node/mod.rs
@@ -20,6 +20,7 @@ use crate::{
     permissions::PermissionsConfiguration, NodePrivilege, RuntimeProxy, SecureServerConfiguration,
     SignatureTable,
 };
+use log::warn;
 use oak_abi::proto::oak::application::{
     node_configuration::ConfigType, ApplicationConfiguration, CryptoConfiguration,
     LogConfiguration, NodeConfiguration,
@@ -62,11 +63,6 @@ pub trait Node: Send {
         handle: oak_abi::Handle,
         notify_receiver: oneshot::Receiver<()>,
     );
-
-    /// Gets the privilege associated with the Node.
-    fn get_privilege(&self) -> NodePrivilege {
-        NodePrivilege::default()
-    }
 }
 
 /// Indication of the level of isolation of a node.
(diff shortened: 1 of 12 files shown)


oak_runtime/src/node/mod.rs
```diff
@@ -215,17 +258,34 @@ impl NodeFactory<NodeConfiguration> for ServerNodeFactory {
...
+///
+/// The reason the privilege is returned here and not by the node itself is that it should be
+/// determined by the [`NodeFactory`] instance, so that an untrusted Node may not be able to
+/// single-handedly increase its own privilege. In this codebase all the Nodes are defined alongside
+/// each other, so this does not make much difference, but if Nodes were loaded dynamically or from
+/// untrusted libraries, then it would be more obvious that we would still want to keep the
+/// privilege assignment here.
+pub struct CreatedNode {
+    pub instance: Box<dyn Node>,
+    pub privilege: NodePrivilege,
```
conradgrobler, line 286: Should this be public? Is the plan to allow modification of the node's privilege at a later time after node creation?

conradgrobler (approved):
I think the logic looks fine. I am not clear on the motiviation behind the change. Is it addressing an existing problem, or to allow some future use case that need is?

oak_runtime/src/node/mod.rs
tiziano88: There is no such plan, anyways it needs to be passed around, and this makes it more convenient. This may change in a follow up PR though.

tiziano88:
> I think the logic looks fine. I am not clear on the motiviation behind the change. Is it addressing an existing problem, or to allow some future use case that need is?

Nothing blocking, but it is addressing the issue that there should be a separation between assigning the privilege and the implementation of a node, otherwise there is not much point in having different privileges for different node types. Also I am trying to extract the core logic of the runtime, and node creation and privilege assignment end up being separate pieces that will be provided by specific implementation, but it does not make much sense if each node type declares its own privilege.

Long term, I think we should be able to have a pluggable mechanism for adding new node types, without worrying that a node maintained by someone else in some other crate is able to elevate its own privilege.

PR #1891, opened 2021-02-18
Improve the use of features with server variants in the runner

rbehjati:
* `kebab-case` seems to be more conventional for feature names, so the feature names are updated
* Merges `kms` and `experimental` server variants in the runner
* Adds doc comments for server variants
* Adds a new step to `run_ci` and github CI for running the `hello_world` example with the base server and a permissions file.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -271,7 +271,7 @@ fn build_server(opt: &BuildServer) -> Step {
                 ),
             }],
             match opt.server_variant {
-                ServerVariant::Unsafe | ServerVariant::Coverage | ServerVariant::Kms | ServerVariant::Experimental => vec![Step::Single {
+                ServerVariant::Unsafe | ServerVariant::Coverage | ServerVariant::Experimental => vec![Step::Single {
                     name: "build introspection browser client".to_string(),
                     command: Cmd::new("npm",
                                       vec![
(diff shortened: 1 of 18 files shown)


runner/src/main.rs
```diff
@@ -305,22 +305,18 @@ fn build_server(opt: &BuildServer) -> Step {
                                 format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET)),
                                 "--release".to_string(),
                             ],
-                            ServerVariant::NoIntrospectionClient => vec!["--features=oak_unsafe".to_string(),
+                            ServerVariant::NoIntrospectionClient => vec!["--features=oak-unsafe".to_string(),
```
tiziano88, line 308: This seems incorrect?

runner/src/main.rs
```diff
@@ -305,22 +305,18 @@ fn build_server(opt: &BuildServer) -> Step {
...
-                                format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET)),
-                                "--release".to_string(),
-                            ],
-                            ServerVariant::Kms => vec!["--features=oak_unsafe,oak_introspection_client,awskms,gcpkms".to_string(),
+                            ServerVariant::Unsafe => vec!["--features=oak-unsafe,oak-introspection-client".to_string(),
                                 format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET)),
                                 "--release".to_string(),
                             ],
                             // If building in coverage mode, use the default target from the host, and build
                             // in debug mode.
```
tiziano88, line 317: or rather, s/debug/unsafe/?

runner/src/main.rs
```diff
@@ -305,22 +305,18 @@ fn build_server(opt: &BuildServer) -> Step {
                                 format!("--target={}", opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET)),
```
tiziano88, line 305: Is `--no-default-features` still needed here (1 line above)?

tiziano88 (approved):

runner/src/main.rs
rbehjati: Not strictly. I've kept it for the emphasis.

runner/src/main.rs
rbehjati: I think with `NoIntrospectionClient`, we still want to allow logs (therefore `oak-unsafe` is needed), without having to install npm, which is needed for the introspection client. Without `oak-unsafe`, this server variant will be the same as base.

runner/src/main.rs
rbehjati: Done.

ipetr0v (approved):

runner/src/main.rs
tiziano88: Up to you, though it seems misleading to me, since default features is already empty (and at a first read this seems to suggest it is not).

runner/src/main.rs
tiziano88: So perhaps this and Base could be merged in the same match arm at some point?

rbehjati: `Base` does not log anything, but `NoIntrospectionClient` does. Are you suggesting that we don't need the `NoIntrospectionClient` variant?

tiziano88: Ah my bad, I got confused by the various layers of negation :) I think it would be simpler if the name of the variants relates directly to the enabled features, rather than a mix of enabled and disabled features. e.g. unsafe, unsafeWithIntrospection, etc. But also actually, why do we need unsafeWithIntrospection if we already have experimental? I thought the point of the experimental variant is to enable all the features. I guess it may help debugging issues with specific features anyways?

rbehjati: That is a good question. Actually I don't know what are the exact use cases for each the variants. I agree with you that we probably only need the following:
- Base
- Experimental (with all `oak_loader` features enabled)
- ExperimentalWithoutIntrospectionClient (similar to Experimental, but without introspection client)
- Coverage (We probably want all the features enabled for it, but it is not currently the case. It is different from Experimental, because in addition to the features it also needs other flags for running coverage)

For debugging specific features it is probably best to also have a way to enable individual features in the `runner` commands, instead of defining new server variants. 

WDYT?

Issue #1854, opened 2021-02-01
Linear handles tracking issue


tiziano88:
Here is my game plan for making handles linear types:
1. Add handle_clone syscall (#1782)
2. Allow SDK functions (channel_read, channel_write) to use borrowed handles instead of owned ones (#1843)
3. Add `linear-handles` feature to `oak_io` that optionally makes handle types defined there linear (#1850)
4. Add `linear-handles` feature to SDK for compatibility with the `oak_io` feature (#1884)
5. Migrate individual example modules to use the new feature
6. Migrate runtime to use linear handles
7. Remove the features and make their behaviour the default

Of course we could also do all of this at once (e.g. https://github.com/wildarch/oak/tree/handle_clone_drop_impls), but I suspect that would be a very tough PR to land, because it touches so much of the codebase.

_Originally posted by @wildarch in https://github.com/project-oak/oak/issues/1686#issuecomment-765351586_

tiziano88:
Pulling this into its own issue as there are quite a few moving parts already.

I wonder if we should have a more detailed RFC, discussing pros / cons, alternatives considered, and what effect it has on the current codebase.

It seems to me that #1850 is a good proof of concept already that shows that it is feasible and quite elegant.

cc @daviddrysdale

wildarch:
Thanks for this, I agree it makes sense to spin this off into a separate issue 😄 .

I can write up an RFC for this if that is appreciated, but to me it seems quite clear-cut that linear handles are an improvement over the current approach: they correspond to Rust's ownership model and prevent leaks. Are there many alternative ideas to consider here? Keeping the current system would be one, I'm not sure if there are others.

It would have quite an impact on the codebase though, maybe that in itself would be a good reason to write up an RFC? I'm not sure.

tiziano88:
Perhaps no need for an actual RFC, but could you describe roughly what the behaviour would be for read and write handles, for instance? e.g. I would expect that only write handles can be cloned, is that what you are thinking too already? Also what happens when a message with handles is cloned? I guess all the handles inside it are cloned too, is that something that will come naturally, or will it require additional logic in the code generation or elsewhere?

wildarch:
The major changes for read and write handles would be:
- No more `Copy` impl. SDK functions such as `channel_read` should take a `&ReadHandle` instead of an owned `ReadHandle`, to avoid unnecessary cloning.
[...]

PR #1892, opened 2021-02-19
Migrate injection example to use linear-handles feature.

wildarch:
# Checklist

      construction.

Tracking issue: #1854

### DIFF
--- a/sdk/rust/oak/src/io/sender.rs
+++ b/sdk/rust/oak/src/io/sender.rs
@@ -43,12 +43,14 @@ impl<T: Encodable> SenderExt<T> for Sender<T> {
     fn send(&self, t: &T) -> Result<(), OakError> {
         let message = t.encode()?;
         crate::channel_write(&self.handle, &message.bytes, &message.handles)?;
+        drop_handles(message);
         Ok(())
     }
 
     fn send_with_downgrade(&self, t: &T) -> Result<(), OakError> {
         let message = t.encode()?;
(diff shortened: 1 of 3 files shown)


sdk/rust/oak/src/io/sender.rs
```diff
@@ -43,12 +43,14 @@ impl<T: Encodable> SenderExt<T> for Sender<T> {
     fn send(&self, t: &T) -> Result<(), OakError> {
         let message = t.encode()?;
         crate::channel_write(&self.handle, &message.bytes, &message.handles)?;
+        drop_handles(message);
```
rbehjati, line 46: I am not sure why the handles need to be dropped. Are not they borrowed or moved?

rbehjati (approved):
LGTM, thanks.

sdk/rust/oak/src/io/sender.rs
wildarch: During encoding we extract the raw handle values from the message, and replace them with `INVALID_HANDLE`: https://github.com/project-oak/oak/blob/main/oak_io/src/handle.rs#L202.

When this message is dropped, all handle values are invalid, so no handles are dropped. Instead we need to manually drop the raw handle values here. I hope that makes sense?

rbehjati: Yes. Thanks. I had overlooked a `clone` :)

rbehjati (approved):

Issue #1861, opened 2021-02-02
Check TEE quotes in X.509 certificates


ipetr0v:
After finishing https://github.com/project-oak/oak/issues/1860 we need to implement the ability to analyze X.509 certificates and specifically check that TEE quotes (saved as an extension) are correct.

*Note*: it looks like [tonic](https://docs.rs/tonic/0.4.0/tonic) doesn't provide functions for analyzing server certificates. There is a [peer_certs](https://docs.rs/tonic/0.4.0/tonic/struct.Request.html?search=#method.peer_certs) function, but it only intercepts certificates sent in the requests, so it doesn't allow to get server certificates.

PR #1893, opened 2021-02-24
Add Rust Proxy Attestation client

ipetr0v:
This commit adds a Proxy Attestation client with a custom TLS certificate verifier, that checks TEE quotes in X.509 TEE extensions.

Fixes https://github.com/project-oak/oak/issues/1861

### DIFF
--- a/examples/proxy_attestation/client/rust/src/main.rs
+++ b/examples/proxy_attestation/client/rust/src/main.rs
@@ -24,7 +24,10 @@ use anyhow::Context;
 use http::uri::Uri;
 use log::info;
 use oak_abi::label::Label;
-use oak_client::{create_tls_channel, interceptors::label::LabelInterceptor};
+use oak_client::{
+    attestation::create_attested_tls_channel, create_tls_channel,
+    interceptors::label::LabelInterceptor,
+};
 use oak_proxy_attestation::proto::{
     proxy_attestation_client::ProxyAttestationClient, GetRootCertificateRequest,
 };
@@ -35,7 +38,7 @@ use structopt::StructOpt;
 use tonic::{transport::Channel, Request};
 
 #[derive(StructOpt, Clone)]
-#[structopt(about = "Private Set Intersection Client")]
+#[structopt(about = "Proxy Attestation Client")]
 pub struct Opt {
     #[structopt(
         long,
@@ -56,6 +59,11 @@ pub struct Opt {
     proxy_root_tls_certificate: String,
 }
 
+// TODO(#1867): Add remote attestation support.
+const TEST_TEE_MEASUREMENT: &str = "Test TEE measurement";
+// Example message expected from the Oak application.
+const EXAMPLE_MESSAGE: &str = "Example message";
+
 /// Create gRPC client for Proxy Attestation Service.
 async fn create_proxy_client(
     uri: &Uri,
@@ -72,9 +80,10 @@ async fn create_proxy_client(
 async fn create_application_client(
     uri: &Uri,
     root_tls_certificate: &[u8],
+    tee_measurement: &[u8],
 ) -> anyhow::Result<ExampleApplicationClient<Channel>> {
     info!("Connecting to Oak application: {:?}", uri);
-    let channel = create_tls_channel(uri, root_tls_certificate)
+    let channel = create_attested_tls_channel(uri, root_tls_certificate, tee_measurement)
         .await
         .context("Couldn't create TLS channel")?;
     let label = Label::public_untrusted();
@@ -103,20 +112,22 @@ async fn get_root_tls_certificate(
 }
 
 /// Connect to Oak application using root TLS certificate received from Proxy Attestation Service.
-/// TODO(#1861): Check TEE quotes in server certificates.
-async fn create_application_connection(
+async fn get_example_message(
     uri: &Uri,
     root_tls_certificate: &[u8],
-) -> anyhow::Result<()> {
-    let mut client = create_application_client(&uri, &root_tls_certificate)
+    tee_measurement: &[u8],
+) -> anyhow::Result<String> {
+    let mut client = create_application_client(&uri, &root_tls_certificate, tee_measurement)
         .await
         .context("Couldn't create gRPC client")?;
     let request = Request::new(GetExampleMessageRequest {});
-    client
+    let example_message = client
         .get_example_message(request)
         .await
-        .context("Couldn't connect to backend")?;
-    Ok(())
+        .context("Couldn't connect to backend")?
+        .into_inner()
+        .message;
+    Ok(example_message)
 }
 
 #[tokio::main]
@@ -132,7 +143,13 @@ async fn main() -> anyhow::Result<()> {
     let root_tls_certificate =
         get_root_tls_certificate(&proxy_uri, &proxy_root_tls_certificate).await?;
     info!("Received root certificate from Proxy Attestation Service");
-    create_application_connection(&app_uri, &root_tls_certificate).await?;
+    let example_message = get_example_message(
+        &app_uri,
+        &root_tls_certificate,
+        TEST_TEE_MEASUREMENT.as_bytes(),
+    )
+    .await?;
+    assert_eq!(EXAMPLE_MESSAGE, example_message);
     info!("Successfully connected to Oak application");
 
     Ok(())
--- a/experimental/attestation_common/src/lib.rs
+++ b/experimental/attestation_common/src/lib.rs
@@ -29,7 +29,7 @@ pub const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840.1.113730.1
 
 /// Placeholder implementation of TEE report for remote attestation.
 /// https://www.amd.com/system/files/TechDocs/56860.pdf#page=39
-/// 
+///
 /// TODO(#1867): Add remote attestation support and use real TEE report.
 #[derive(Deserialize, Serialize, Debug, Default, PartialEq)]
 pub struct Report {
--- a/experimental/attestation_common/src/lib.rs
+++ b/experimental/attestation_common/src/lib.rs
@@ -1,0 +1,72 @@
+//
+// Copyright 2021 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#[cfg(test)]
+mod tests;
+
+use anyhow::Context;
+use openssl::{nid::Nid, x509::X509Extension};
+use serde::{Deserialize, Serialize};
+use x509_parser::der_parser::{self, oid::Oid};
+
+// Using `NETSCAPE_COMMENT` extension since `rust-openssl` doesn't support custom
+// extensions yet.
+// https://github.com/sfackler/rust-openssl/issues/1411
+// https://www.alvestrand.no/objectid/2.16.840.1.113730.1.13.html
+pub const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840.1.113730.1.13);
+
+/// Placeholder implementation of TEE report for remote attestation.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+pub struct Report {
+    /// TEE measurement, i.e. VM hash.
+    pub measurement: Vec<u8>,
+    /// Arbitrary data to put into the TEE report.
+    pub data: Vec<u8>,
+}
+
+impl Report {
+    pub fn new(measurement: &[u8], data: &[u8]) -> Self {
+        Self {
+            measurement: measurement.to_vec(),
+            data: data.to_vec(),
+        }
+    }
+
+    pub fn from_string(input: &str) -> anyhow::Result<Self> {
+        serde_json::from_str(input).context("Couldn't deserialize TEE report")
+    }
+
+    pub fn to_string(&self) -> anyhow::Result<String> {
+        serde_json::to_string(&self).context("Couldn't serialize TEE report")
+    }
+
+    /// Return the TEE [`Report`] as a custom [`X509Extension`].
+    pub fn to_extension(&self) -> anyhow::Result<X509Extension> {
+        let report_string = self.to_string()?;
+        // [`Nid::NETSCAPE_COMMENT`] is a numerical identifier for an OpenSSL object that
+        // corresponds to an X.509 extension implementation.
+        X509Extension::new_nid(None, None, Nid::NETSCAPE_COMMENT, &report_string)
+            .context("Couldn't create X.509 extension")
+    }
+}
+
+/// Computes a SHA-256 digest of `input` and returns it in a form of raw bytes.
+pub fn get_sha256(input: &[u8]) -> Vec<u8> {
+    use sha2::digest::Digest;
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(&input);
+    hasher.finalize().to_vec()
+}
--- a/experimental/attestation_common/src/lib.rs
+++ b/experimental/attestation_common/src/lib.rs
@@ -22,26 +22,37 @@ use openssl::{nid::Nid, x509::X509Extension};
 use serde::{Deserialize, Serialize};
 use x509_parser::der_parser::{self, oid::Oid};
 
-// Using `NETSCAPE_COMMENT` extension since `rust-openssl` doesn't support custom
-// extensions yet.
+// Using `NETSCAPE_COMMENT` extension since `rust-openssl` doesn't support custom extensions yet.
 // https://github.com/sfackler/rust-openssl/issues/1411
 // https://www.alvestrand.no/objectid/2.16.840.1.113730.1.13.html
 pub const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840.1.113730.1.13);
 
 /// Placeholder implementation of TEE report for remote attestation.
-#[derive(Deserialize, Serialize, Debug, PartialEq)]
+/// https://www.amd.com/system/files/TechDocs/56860.pdf#page=39
+/// 
+/// TODO(#1867): Add remote attestation support and use real TEE report.
+#[derive(Deserialize, Serialize, Debug, Default, PartialEq)]
 pub struct Report {
+    /// Version number of this attestation report.
+    pub version: u8,
+    /// Security version number of SNP firmware.
+    pub svn: u8,
+    /// The install version of firmware.
+    pub platform_version: u8,
+    /// Arbitrary data to put into the TEE report.
+    pub report_data: Vec<u8>,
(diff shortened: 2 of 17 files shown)


oak_client/src/attestation.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
+    fn verify_server_cert(
+        &self,
+        roots: &rustls::RootCertStore,
+        certs: &[rustls::Certificate],
+        hostname: webpki::DNSNameRef,
+        ocsp: &[u8],
+    ) -> Result<rustls::ServerCertVerified, rustls::TLSError> {
+        for certificate in certs.iter() {
```
rbehjati, line 117: Are these the same as the certificates in line 61 above? Why are multiple certificates allowed here but not above?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,153 @@
...
+    // Enable HTTP/2.
+    const ALPN_H2: &str = "h2";
+    config.set_protocols(&[Vec::from(&ALPN_H2[..])]);
+
+    // Add root TLS certificate.
+    let mut cc_reader = std::io::BufReader::new(&root_tls_certificate[..]);
+    let certs = rustls::internal::pemfile::certs(&mut cc_reader)
+        .map_err(|error| anyhow!("Couldn't parse TLS certificate: {:?}", error))?;
+    if certs.len() > 1 {
+        return Err(anyhow!("More than one root certificate provided"));
```
rbehjati, line 64: Just curious: why is that a problem?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    // Enable HTTP/2.
+    const ALPN_H2: &str = "h2";
+    config.set_protocols(&[Vec::from(&ALPN_H2[..])]);
+
+    // Add root TLS certificate.
+    let mut cc_reader = std::io::BufReader::new(&root_tls_certificate[..]);
+    let certs = rustls::internal::pemfile::certs(&mut cc_reader)
+        .map_err(|error| anyhow!("Couldn't parse TLS certificate: {:?}", error))?;
+    if certs.len() > 1 {
```
rbehjati, line 63: Could `certs` be empty or is that a case that `rustls::internal::pemfile::certs` handles as error?

examples/proxy_attestation/client/rust/src/main.rs
```diff
@@ -107,8 +113,9 @@ async fn get_root_tls_certificate(
 async fn create_application_connection(
     uri: &Uri,
     root_tls_certificate: &[u8],
+    tee_quote: &str,
 ) -> anyhow::Result<()> {
-    let mut client = create_application_client(&uri, &root_tls_certificate)
+    let mut client = create_application_client(&uri, &root_tls_certificate, tee_quote)
         .await
         .context("Couldn't create gRPC client")?;
     let request = Request::new(GetExampleMessageRequest {});
```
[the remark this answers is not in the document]
ipetr0v: Done

oak_client/Cargo.toml
```diff
@@ -7,6 +7,9 @@ authors = ["Ivan Petrov <ivanpetrov@google.com>"]
 edition = "2018"
 license = "Apache-2.0"
 
+[features]
+oak_attestation = ["regex", "tokio-rustls", "x509-parser", "webpki"]
```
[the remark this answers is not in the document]
ipetr0v: Done

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,153 @@
...
+use x509_parser::{
+    der_parser::{self, der::parse_der, oid::Oid},
+    parse_x509_certificate,
+};
+
+// Using [`Nid::NETSCAPE_COMMENT`] identifier since `rust-openssl` doesn't support custom
+// extensions yet.
+// https://github.com/sfackler/rust-openssl/issues/1411
+// https://www.alvestrand.no/objectid/2.16.840.1.113730.1.13.html
+const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840.1.113730.1.13);
```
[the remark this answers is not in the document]
ipetr0v: This is very experimental. Basically all this code will change once real attestation is in place.
So this check (with regex and etc.) is just a placeholder.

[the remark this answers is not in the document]
ipetr0v: As for best practices: [OpenEnclave](https://github.com/openenclave/openenclave) for example just puts an attestation quote in the extension:

https://github.com/openenclave/openenclave/blob/1839e0b92948c875f2d77352f16540b41280f4ca/enclave/tls_cert.c#L60-L61

https://github.com/openenclave/openenclave/blob/1839e0b92948c875f2d77352f16540b41280f4ca/enclave/crypto/openssl/cert.c#L322-L325

oak_client/src/attestation.rs
ipetr0v: It's just a problem with how `rustls` parses PEM certificate buffers. It assumes that there could be multiple certificates.

oak_client/src/attestation.rs
ipetr0v: Updated to `!=`

oak_client/src/attestation.rs
ipetr0v: Because previously we provide a single root certificate created by the Attestation Proxy

ipetr0v: Updated to only accept a single cert too.

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl RemoteAttestationVerifier {
+    fn new(tee_quote: &str) -> Self {
+        Self {
+            tee_quote: tee_quote.to_string(),
+        }
+    }
+
+    fn check_tee_quote(&self, extension: &str) -> anyhow::Result<()> {
+        let pattern = r"^Quote:(.*)$";
```
tiziano88, line 96: Not sure I get what's going on here, why not use JSON or something that can be parsed properly? It does not seem it would make the code much longer, and I think it is probably more similar to what we would expect in the future? Or just make the entire field the quote, without the `Quote` prefix, for now. Or is the prefix needed for something?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn verify_server_cert(
+        &self,
+        roots: &rustls::RootCertStore,
+        certs: &[rustls::Certificate],
+        hostname: webpki::DNSNameRef,
+        ocsp: &[u8],
+    ) -> Result<rustls::ServerCertVerified, rustls::TLSError> {
+        if certs.len() != 1 {
+            return Err(rustls::TLSError::General(
+                "Only a single certificate should be provided".to_string(),
```
tiziano88, line 121:
```suggestion
                "Only a single certificate must be provided".to_string(),
```

rbehjati (approved):
LGTM, thanks!

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,156 @@
...
+            })?;
+
+        // Check that X.509 TEE extension contains a correct TEE quote.
+        self.check_tee_quote(tee_extension_value).map_err(|error| {
+            rustls::TLSError::General(format!("Incorrect TEE quote: {:?}", error))
+        })?;
+
+        // Use [`rustls::WebPKIVerifier`] to check certificate correctness.
+        let verifier = rustls::WebPKIVerifier::new();
+        verifier.verify_server_cert(roots, certs, hostname, ocsp)
```
tiziano88, line 154: So if this was not here, the standard verifier would not be checked? I thought this verifier is in addition to the built in one. Either way, could you add a comment?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,156 @@
...
+        .set_certificate_verifier(std::sync::Arc::new(RemoteAttestationVerifier::new(
+            tee_quote,
+        )));
+
+    Ok(ClientTlsConfig::new().rustls_client_config(config))
+}
+
+/// Custom X.509 certificate verifier that checks TEE quotes in the X.509 TEE extension.
+struct RemoteAttestationVerifier {
+    tee_quote: String,
```
tiziano88, line 85: I think in general we would not know the exact quote to expect, since it contains extra data and signatures and stuff. So passing the literal string here seems incorrect. I think what we need to pass is some sort of predicate / verifier for the quote itself, which for now may just be a `Fn(&str) -> QuoteVerified` or something like that. Then for now we could have a single implementation that just checks for equivalence, it can be an anonymous lambda for that.

edit: I guess this is actually what `RemoteAttestationVerifier` actually is, but I think it should be defined as a trait, allow using a lambda for the check logic, clarify what the return value means, and clarify what the parameter is (`extension` seems a bit vague; also isn't that extension always the quote? or can there be other stuff in it? also `TEE_EXTENSION_OID` suggests that this is the `tee_extension` value, so maybe just stick to this terminology, and avoid quote? or use `tee_quote_*` everywhere)

oak_client/src/attestation.rs
tiziano88: Maybe also return the actual number of certs found, it will help debugging. Could you also comment on why this is the case? Is it because we expect a self-signed cert? Or is it a single one even for CA-signed certs?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,156 @@
...
+        })?;
+
+        // Extract X.509 TEE extension.
+        let extensions = parsed_certificate.extensions();
+        let tee_extension = extensions.get(&TEE_EXTENSION_OID).ok_or_else(|| {
+            rustls::TLSError::General("Couldn't find TEE quote X.509 extension".to_string())
+        })?;
+        let tee_extension_value = parse_der(tee_extension.value)
+            .map_err(|error| {
+                rustls::TLSError::General(format!("Couldn't parse TEE extension: {:?}", error))
```
tiziano88, line 139: Inconsistent naming "TEE extension" vs "TEE quote X.509 extension" above, could you make things consistent?

experimental/attestation_common/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn to_string(&self) -> String {
+        serde_json::to_string(&self).expect("Couldn't serialize TEE quote")
+    }
+
+    /// Return the TEE [`Quote`] as a custom [`X509Extension`].
+    pub fn to_extension(&self) -> Result<X509Extension, ErrorStack> {
+        // Using [`Nid::NETSCAPE_COMMENT`] identifier since `rust-openssl` doesn't support custom
+        // extensions yet:
+        // https://github.com/sfackler/rust-openssl/issues/1411
+        X509Extension::new_nid(None, None, Nid::NETSCAPE_COMMENT, &self.to_string())
```
tiziano88, line 53: What is the difference between this and `TEE_EXTENSION_OID`? Why do we need both?

experimental/attestation_common/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Computes a SHA-256 digest of `input` and returns it in a form of raw bytes.
+pub fn get_sha256(input: &[u8]) -> Vec<u8> {
+    use sha2::digest::Digest;
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(&input);
+    hasher.finalize().to_vec()
+}
+
+/// Computes a SHA-256 digest of `bytes` and returns it in a hex encoded string.
+pub fn get_sha256_hex(bytes: &[u8]) -> String {
```
tiziano88, line 74: Why is this called `bytes`, but it is called `input` in the previous fn?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,170 @@
...
+/// [`root_tls_certificate`] should contain only a single PEM encoded certificate.
+/// https://tools.ietf.org/html/rfc1421
+pub fn create_tls_attestation_config(
+    root_tls_certificate: &[u8],
+    tee_quote: &str,
+) -> anyhow::Result<ClientTlsConfig> {
+    let mut config = rustls::ClientConfig::new();
+
+    // Enable HTTP/2.
+    const ALPN_H2: &str = "h2";
```
tiziano88, line 54: Do you have a reference for this magic string?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,170 @@
...
+            "Only a single PEM certificate must be provided, provided {}",
+            certs.len()
+        ));
+    }
+    config
+        .root_store
+        .add(&certs[0])
+        .context("Couldn't add root certificate")?;
+
+    // Set custom X.509 certificate verifier.
```
tiziano88, line 73: Could you clarify also here that this completely replaces the default verifier?

oak_client/src/attestation.rs
```diff
@@ -0,0 +1,170 @@
...
+        // let tee_quotes = re.captures(&extension).context("Couldn't find TEE quote")?;
+
+        // // `0` capture group corresponds to the whole pattern, while `1` corresponds to the first
+        // // match.
+        // // https://docs.rs/regex/1.4.3/regex/struct.Captures.html#method.get
+        // if self.tee_quote == tee_quotes.get(1).map_or("", |m| m.as_str()) {
+        //     Ok(())
+        // } else {
+        //     Err(anyhow!("Incorrect TEE quote"))
+        // }
```
tiziano88, line 107: Remove? Also the regex dep?

experimental/attestation_common/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        // extensions yet:
+        // https://github.com/sfackler/rust-openssl/issues/1411
+        X509Extension::new_nid(None, None, Nid::NETSCAPE_COMMENT, &self.to_string())
+    }
+}
+
+impl std::str::FromStr for Quote {
+    type Err = serde_json::Error;
+
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
```
tiziano88, line 60: It may be better to just have this as a method on `Quote` itself, like `to_string` is already there. Or alternatively implement `Display` for it, or something like that.

experimental/attestation_common/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl Quote {
+    pub fn new(measurement: &str, public_key: &[u8]) -> Self {
+        Self {
+            measurement: measurement.to_string(),
+            public_key_hash: get_sha256_hex(public_key),
+        }
+    }
+
+    pub fn to_string(&self) -> String {
```
tiziano88, line 44: Could you add some tests for these methods? They don't need to be very extensive, but even a simple one would be good as documentation to see what the expected output is.

experimental/attestation_common/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+// extensions yet.
+// https://github.com/sfackler/rust-openssl/issues/1411
+// https://www.alvestrand.no/objectid/2.16.840.1.113730.1.13.html
+pub const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840.1.113730.1.13);
+
+/// Placeholder implementation of TEE quote for remote attestation.
+#[derive(Deserialize, Serialize, Debug)]
+pub struct Quote {
+    pub measurement: String,
+    pub public_key_hash: String,
```
tiziano88, line 33: Shouldn't these be arbitrary bytes? Why strings? I mean, it's fine to base64 or hex encode them when converting to JSON, but I don't think the struct should contain these pretty printed versions.

tiziano88:
BTW another option would be to use "report" instead of "quote", e.g. https://pkg.go.dev/github.com/edgelesssys/ego@v0.1.0/attestation#Report . Also this has interesting fields that I think eventually we will have to have, so maybe we can reuse those names if we think they are sensible, and / or link to that for future reference.

oak_client/src/attestation.rs
ipetr0v: I think having it as a trait would require `#![feature(associated_type_bounds)]`, since we would need to implement `trait rustls::ServerCertVerifier` for this trait.

https://github.com/rust-lang/rfcs/pull/2289/files#diff-6cd727664c55412d02e308dfb2dcf5f5ee273a0c6f728160f79e68e1fab88e7bR71-R78

ipetr0v: Currently made a `Quote` a separate decodable class
And this verifier takes `tee_measurement` as an argument

experimental/attestation_common/src/lib.rs
ipetr0v: Done

experimental/attestation_common/src/lib.rs
ipetr0v: Added a comment

experimental/attestation_common/src/lib.rs
ipetr0v: Done

oak_client/src/attestation.rs
tiziano88: What I mean is that I think this class should only hold something like `Predicate<Report>` or similar (e.g. by using https://crates.io/crates/predicates , or anyways a fn from Report to bool). Then the caller can decide how to build one of those. I think this would make more sense if you expand the Report struct to be more realistic, for instance by adding some more fields, similar to https://pkg.go.dev/github.com/edgelesssys/ego@v0.1.0/attestation#Report . Then there may be different validation logic for each field or combination of fields.

experimental/proxy_attestation/src/main.rs
```diff
@@ -71,12 +71,12 @@ pub struct Opt {
...
-const TEST_TEE_QUOTE: &str = "Test TEE quote";
+// TODO(#1867): Add remote attestation support.
+const TEST_TEE_MEASUREMENT: &str = "Test TEE measurement";
 
-/// Placeholder function for collecting quotes of remotely attested TEEs.
-fn get_tee_quote(_req: &Request<GetSignedCertificateRequest>) -> String {
-    // TODO(#1867): Add remote attestation support.
-    TEST_TEE_QUOTE.to_string()
+/// Placeholder function for collecting TEE measurement of remotely attested TEEs.
+fn get_tee_measurement(_req: &Request<GetSignedCertificateRequest>) -> Vec<u8> {
```
tiziano88, line 78: I'm still not sure the types line up here. The measurement is independent of any data, right? The report, on the other hand, binds together the measurement, other HW information, the extra data, all signed with a key that is baked into the hardware. But the measurement here seems just something like MRENCLAVE, which cannot possibly depend on any other data. Again, I think having a more realistic version of Report would probably make this clearer.

experimental/attestation_common/src/tests.rs
```diff
@@ -0,0 +1,36 @@
...
+
+const TEST_TEE_MEASUREMENT: &str = "Test TEE measurement";
+const TEST_TEE_DATA: &str = "Test TEE data";
+
+#[test]
+fn test_report_serialization() {
+    let report = Report::new(TEST_TEE_MEASUREMENT.as_bytes(), TEST_TEE_DATA.as_bytes());
+
+    let serialized_report_result = report.to_string();
+    assert_matches!(serialized_report_result, Ok(_));
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/proxy_attestation/src/main.rs
ipetr0v: This function is inside the Proxy Attestation service itself, so it cannot generate reports, it can only receive them from the backend.
Will change it to `get_tee_report`, but it will probably not matter, since this report should be an artifact from the remote attestation.

oak_client/src/attestation.rs
ipetr0v: Created an issue: https://github.com/project-oak/oak/issues/1896

tiziano88 (approved):

PR #1894, opened 2021-02-25
Add TLS Attestation Service

ipetr0v:
This change:
- Adds experimental TLS Attestation Service
- Adds Corresponding backend and client code
- Updates `oak_client` to work with custom X.509 certificate verifiers


experimental/tls_attestation/src/certificate.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let mut name = X509NameBuilder::new()?;
+        name.append_entry_by_text("CN", &domain_name.to_string())?;
+        let name = name.build();
+
+        let mut builder = X509::builder()?;
+        builder.set_version(CERTIFICATE_VERSION)?;
+        let serial_number = {
+            let mut serial = BigNum::new()?;
+            serial.rand(SERIAL_NUMBER_SIZE, MsbOption::MAYBE_ZERO, false)?;
```
tiziano88, line 79: Is this actually needed? If I look at https://github.com/edgelesssys/ego/blob/8fe312f9b38f3fb9a00cedd0d749f170d7f138ae/samples/remote_attestation/server.go#L52-L62 , it seems that most fields are left out, can we do the same? Also, is their client / server interoperable with ours?

experimental/tls_attestation/src/certificate.rs
```diff
@@ -0,0 +1,121 @@
...
+pub struct Identity {
+    pub private_key: Vec<u8>,
+    pub certificate: Vec<u8>,
+}
+
+impl CertificateAuthority {
+    /// Creates a self-signed X.509 certificate and a corresponding private key.
+    /// The certificate is bound to `domain_name` and contains TEE credentials in a custom X.509
+    /// extension.
+    pub fn generate_certificate(domain_name: &str) -> anyhow::Result<Identity> {
```
tiziano88, line 67: Domain name seems irrelevant (in fact misleading) since it's always going to be a self signed cert, right?

oak_client/src/attestation.rs
```diff
@@ -37,34 +43,48 @@ pub async fn create_attested_tls_channel(
         .context("Couldn't connect to Oak Application")
 }
 
+/// Creates an HTTPS client connecting to a TEE service specified by [`uri`] that checks that X.509
+/// TEE extension contains a TEE report with expected [`tee_measurement`] and a hash of the public
+/// key used to the TLS certificate.
+pub async fn create_https_client(
+    tee_measurement: &[u8],
+) -> anyhow::Result<hyper::Client<hyper_rustls::HttpsConnector<hyper::client::HttpConnector>>> {
+    // Config doesn't need root certificates, since it accepts self-signed ones.
```
tiziano88, line 52: Why does this not need the root cert, while the gRPC version needs it?

oak_client/src/attestation.rs
```diff
@@ -37,34 +43,48 @@ pub async fn create_attested_tls_channel(
...
+    let tls_config = create_tls_attestation_config(&[], tee_measurement)
+        .context("Couldn't create TLS config")?;
+
+    let mut http = hyper::client::HttpConnector::new();
+    http.enforce_http(false);
```
tiziano88, line 57: Could you comment why this is needed?

oak_client/src/attestation.rs
```diff
@@ -37,34 +43,48 @@ pub async fn create_attested_tls_channel(
...
 /// https://tools.ietf.org/html/rfc1421
 pub fn create_tls_attestation_config(
     root_tls_certificate: &[u8],
     tee_measurement: &[u8],
-) -> anyhow::Result<ClientTlsConfig> {
+) -> anyhow::Result<rustls::ClientConfig> {
     let mut config = rustls::ClientConfig::new();
 
-    // Configure ALPN to accept HTTP/2.
+    // Configure ALPN to accept HTTP/1.1 and HTTP/2.
```
tiziano88, line 74: Why do we need http1 now?

oak_client/src/attestation.rs
```diff
@@ -91,13 +110,19 @@ impl RemoteAttestationVerifier {
...
-    fn check_tee_report(&self, extension: &str) -> anyhow::Result<()> {
+    /// Check that the [`extension`] contains [`RemoteAttestationVerifier::tee_measurement`] and
+    /// a SHA-256 digest of the [`public_key`].
+    fn check_tee_report(&self, extension: &str, public_key: &[u8]) -> anyhow::Result<()> {
         let tee_report = Report::from_string(extension).context("Couldn't parse TEE report")?;
-        if self.tee_measurement == tee_report.measurement {
-            Ok(())
-        } else {
-            Err(anyhow!("Incorrect TEE report"))
+        let public_key_hash = get_sha256(public_key);
```
tiziano88, line 117: Should there be a check of the signature of the report too? Even if it is just a placeholder, that seems quite important.

oak_client/src/attestation.rs
```diff
@@ -91,13 +110,19 @@ impl RemoteAttestationVerifier {
...
+
+        if public_key_hash != tee_report.data {
+            return Err(anyhow!("TEE report contains incorrect public key"));
```
tiziano88, line 120: Print actual / expected values (and below).

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
-        self.check_tee_report(tee_extension_value)
-            .map_err(|error| {
-                rustls::TLSError::General(format!("Incorrect TEE report: {:?}", error))
+            // Extract public key.
+            let parsed_openssl_certificate = openssl::x509::X509::from_der(&certificate.0)
+                .map_err(|error| {
+                    let error =
+                        rustls::TLSError::General(format!("Couldn't parse certificate: {}", error));
+                    warn!("{:?}", error);
+                    error
```
tiziano88, line 154: out of curiosity, what happens to these returned errors? do they get logged again?

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
+                        error
+                    ));
+                    warn!("{:?}", error);
+                    error
+                })?;
+
+            // Extract X.509 TEE extension.
+            let extensions = parsed_certificate.extensions();
+            let tee_extension = extensions.get(&TEE_EXTENSION_OID).ok_or_else(|| {
+                rustls::TLSError::General("Couldn't find X.509 TEE extension".to_string())
```
tiziano88, line 179: Also print the id of the required extension, and the found extensions (if easy enough, if not then don't worry)

oak_client/src/attestation.rs
```diff
@@ -91,13 +110,19 @@ impl RemoteAttestationVerifier {
         }
     }
 
-    fn check_tee_report(&self, extension: &str) -> anyhow::Result<()> {
+    /// Check that the [`extension`] contains [`RemoteAttestationVerifier::tee_measurement`] and
+    /// a SHA-256 digest of the [`public_key`].
```
tiziano88, line 114: Also it checks the measurement, right? Which I would expect to see here as a param too, or at least mentioned in the comment.

oak_client/src/attestation.rs
```diff
@@ -91,13 +110,19 @@ impl RemoteAttestationVerifier {
...
+    fn check_tee_report(&self, extension: &str, public_key: &[u8]) -> anyhow::Result<()> {
```
tiziano88, line 115: Is this fn defined on `self` just because it needs access to the expected report value? That does not seem good encapsulation, also since the public key is passed explicitly instead, it is also quite hard to test in isolation. I would expect a top-level method that does the checking perhaps, not tied to a struct impl.

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
-        // Extract X.509 TEE extension.
-        let extensions = parsed_certificate.extensions();
-        let tee_extension = extensions.get(&TEE_EXTENSION_OID).ok_or_else(|| {
-            rustls::TLSError::General("Couldn't find X.509 TEE extension".to_string())
-        })?;
-        let tee_extension_value = parse_der(tee_extension.value)
-            .map_err(|error| {
-                rustls::TLSError::General(format!(
-                    "Couldn't parse X.509 TEE extension: {:?}",
+        for certificate in certs.iter() {
```
tiziano88, line 137: Could you add a comment about how many certs are expected here?

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
+            // Parse X.509 certificate encoded using DER format.
```
tiziano88, line 138: The body of this for loop is quite long, could the functionality be encapsulated in a separate function? Then you could also do something more idiomatic using https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each . This also avoids all the extra logging that is now done in `map_err`, you could just log in one place.

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
+                        warn!("{:?}", error);
+                        error
+                    })?;
+                }
+                root_certificates
+            } else {
+                roots.clone()
+            };
+
+            // Add certificate correctness check.
```
conradgrobler, line 228: Perhaps move this so that it is the first check? I don't think it matters for this case, but it is usually a good idea to do general validity checks before doing specific checks. In some cases doing specific checks before general validation allows attackers to gain extra information by sending invalid or malformed inputs.

experimental/tls_attestation/src/certificate.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/// Convenience structure for creating X.509 certificates.
+/// https://tools.ietf.org/html/rfc5280
+pub struct CertificateAuthority {
+    pub key_pair: PKey<Private>,
+    pub certificate: X509,
+}
+
+pub struct Identity {
```
conradgrobler, line 58: Add doc comments, seeing that this struct is public?

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
+                .public_key()
+                .map_err(|error| {
+                    let error = rustls::TLSError::General(format!(
+                        "Couldn't extract public key: {}",
+                        error
+                    ));
+                    warn!("{:?}", error);
+                    error
+                })?
+                .public_key_to_pem()
```
conradgrobler, line 166: Would it not be better to use the raw public key bytes, or at least the DER encoded version? I believe PEM encoding can contain whitespace and still be valid, so there is no guarantee that the PEM encoding of the same bytes will be exactly the same for all libraries.

oak_client/src/attestation.rs
```diff
@@ -109,50 +134,109 @@ impl rustls::ServerCertVerifier for RemoteAttestationVerifier {
...
+                        error
+                    ));
+                    warn!("{:?}", error);
+                    error
+                })?
+                .1
+                .as_str()
+                .map_err(|error| {
+                    let error = rustls::TLSError::General(format!(
+                        "Couldn't parse TEE report value: {:?}",
```
conradgrobler, line 194: nit: This is not yet parsing the report value, just checking that the extension value is a string.

experimental/tls_attestation/src/server.rs
```diff
@@ -0,0 +1,150 @@
...
+    uri: http::Uri,
+) -> Result<Response<Body>, hyper::Error> {
+    let (mut parts, body) = request.into_parts();
+    parts.uri = uri;
+
+    let proxy_request = Request::from_parts(parts, body);
+    client.request(proxy_request).await
+}
+
+/// Run HTTPS proxy server that provides self-signed TLS certificates with TEE attestation
```
rbehjati, line 44:
```suggestion
/// Runs HTTPS proxy server that provides self-signed TLS certificates with TEE attestation
```
Perhaps also briefly describe what `domain_name` is.

oak_client/src/attestation.rs
ipetr0v: Done

oak_client/src/attestation.rs
ipetr0v: Checked that these errors are only logged once

conradgrobler (approved):
LGTM

experimental/attestation_common/src/report.rs
```diff
@@ -0,0 +1,105 @@
...
+/// https://www.amd.com/system/files/TechDocs/56860.pdf#page=39
+///
+/// TODO(#1867): Add remote attestation support and use real TEE report.
+#[derive(Deserialize, Serialize, Debug, Default, PartialEq)]
+pub struct Report {
+    /// Version number of this attestation report.
+    pub version: u8,
+    /// Security version number of SNP firmware.
+    pub svn: u8,
+    /// The install version of firmware.
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/tls_attestation/src/certificate.rs
```diff
@@ -0,0 +1,121 @@
...
+        let subject_key_identifier =
+            SubjectKeyIdentifier::new().build(&builder.x509v3_context(None, None))?;
+        builder.append_extension(subject_key_identifier)?;
+
+        let subject_alt_name = SubjectAlternativeName::new()
+            .dns(domain_name)
+            .build(&builder.x509v3_context(None, None))?;
+        builder.append_extension(subject_alt_name)?;
+
+        let tee_report = get_example_tee_report(&key_pair.public_key_to_pem()?);
```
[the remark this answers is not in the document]
ipetr0v: Done

conradgrobler (approved):

experimental/attestation_common/src/certificate.rs
```diff
@@ -0,0 +1,317 @@
...
+        let certificate = builder.build(key_pair)?;
+        Ok(certificate)
+    }
+
+    /// Generates an X509 certificate based on the certificate signing `request`.
+    /// `add_tee_extension` indicates whether to add a custom extension containing a TEE report.
+    pub fn sign_certificate(
+        &self,
+        request: X509Req,
+        add_tee_extension: AddTeeExtension,
```
rbehjati, line 119: Is there a relation between this parameter and `add_tee_extension` in the `create` method above? I mean is any combination of the values of these two parameters valid? For example, is it correct to have a root-certificate that does not have a TEE extension but use it for signing certificates that have TEE extension? 
Or actually, how is the TEE extension in the root_certificate used? (sorry if you have explained it before!)

experimental/attestation_common/src/certificate.rs
```diff
@@ -0,0 +1,317 @@
...
+            .context("Couldn't encode root certificate in PEM format")
+    }
+}
+
+/// Helper struct that implements certificate creation using `openssl`.
+struct CertificateBuilder {
+    builder: X509Builder,
+}
+
+impl CertificateBuilder {
```
rbehjati, line 169: Nice builder!

rbehjati (approved):
LGTM! Thanks.

experimental/attestation_common/src/certificate.rs
ipetr0v: `add_tee_extension` is going to be added for a machine that wants to be remotely attested.
For example, if we use self-signed certificates - we will add this parameter to the root certificate.
And if we want to use Proxy Attestation, we will sign a new certificate (for backend server) using our root one (that doesn't need the parameter since it's not going to be remotely attested). And the new certificate will be created using the `add_tee_extension` and be sent to the backend server.

experimental/attestation_common/src/certificate.rs
ipetr0v: Thanks =)

experimental/attestation_common/src/certificate.rs
rbehjati: I see. Thanks. So they are sort of mutually exclusive.

tiziano88 (approved):

examples/tls_attestation/client/rust/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+//
+
+//! Client for the TLS Attestation example.
+
+use anyhow::Context;
+use assert_matches::assert_matches;
+use log::info;
+use oak_abi::label::Label;
+use oak_client::attestation::create_attested_https_client;
```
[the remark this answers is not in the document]
ipetr0v: Agree

experimental/attestation_common/src/certificate.rs
```diff
@@ -0,0 +1,317 @@
...
+// The most significant bit is excluded because it's passed as a separate argument to:
+// https://docs.rs/openssl/0.10.33/openssl/bn/struct.BigNum.html#method.rand
+const SERIAL_NUMBER_SIZE: i32 = 159;
+const CERTIFICATE_EXPIRATION_INTERVAL_IN_DAYS: u32 = 1;
+const DEFAULT_DNS_NAME: &str = "localhost";
+
+/// Indicator whether to add a custom TEE extension to a certificate.
+#[derive(PartialEq)]
+pub enum AddTeeExtension {
+    /// Parameter contains a PEM encoded X.509 certificate that signs TEE firmware key.
```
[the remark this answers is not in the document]
ipetr0v: I'm referring to the `Vec<u8>` inside `Yes`.
It's the TEE Provider's certificate, that was created during the attestation.

Updated the comment.

experimental/attestation_common/src/certificate.rs
```diff
@@ -0,0 +1,317 @@
...
+        builder.set_name()?;
+        builder.set_public_key(key_pair)?;
+        builder.set_expiration_interval(CERTIFICATE_EXPIRATION_INTERVAL_IN_DAYS)?;
+
+        builder.add_basic_constraints_extension(true)?;
+        builder.add_key_usage_extension(true)?;
+        builder.add_subject_key_identifier_extension(None)?;
+        builder.add_subject_alt_name_extension()?;
+
+        // Bind the certificate to the TEE firmware using an X509 TEE extension.
```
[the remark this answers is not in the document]
ipetr0v: Done

PR #1895, opened 2021-02-26
Use challenge response in the chat example

### LINKED ISSUE #1851: Use the challenge-response mechanism in the Chat example
The chat example uses the C++ client for which we currently don't have signing support (#1838). It should be possible to spawn a child process running [`oak_sign` cmd tool](https://github.com/project-oak/oak/blob/9584170e0c3e1c38a1b3b1fb4bea86e1125687c1/oak_sign/src/main.rs#L17-L37) and get the signature from it.

rbehjati:
* Adds a new cpp library, `oak/common/oak_sign`, similar to the rust `oak_sign` library, with support for ed25519 signing
* Adds challenge-response support to the chat example
* Adds tink as a dependency in the bazel workspace

Fixes #1851


examples/chat/README.md
```diff
@@ -26,19 +26,15 @@ and creates a first chat room inside it:
...
-2019-10-24 10:47:20  INFO  chat.cc : 242 : Join this room with --address=localhost:8080 --room_access_token=NKsceNlg69UbcvryfzmFGnMv9qnZ0DYh6u6gJxujnPPxvHsxMehoD368sumKawVaq9WaSkzrcStoNYLvVNdzhA==
-```
-
-Another party can then join the same chat room by using these arguments:
+Another party can then join the same chat room by setting `room_secret` to a
+path to a file containing the private key:
 
 ```bash
-./scripts/runner --logs run-examples --run-server=false --example-name=chat --client-additional-args=--test=false --client-additional-args=--room_access-token=NKsceNlg69UbcvryfzmFGnMv9qnZ0DYh6u6gJxujnPPxvHsxMehoD368sumKawVaq9WaSkzrcStoNYLvVNdzhA==
+./scripts/runner --logs run-examples --run-server=false --example-name=chat --client-additional-args=--test=false --client-additional-args=--room_secret="chat-room.key"
```
tiziano88, line 37:
```suggestion
./scripts/runner --logs run-examples --run-server=false --example-name=chat --client-additional-args=--test=false --client-additional-args=--room_secret=chat-room.key
```

tiziano88: (not sure whether the quotes are necessary, it seems there is no need for double quotes anyways)

examples/chat/client/cpp/chat.cc
```diff
@@ -14,29 +14,31 @@
...
 #include "include/grpcpp/grpcpp.h"
 #include "oak/client/application_client.h"
-#include "oak/common/nonce_generator.h"
+#include "oak_abi/proto/identity.pb.h"
 
 ABSL_FLAG(bool, test, false, "Run a non-interactive version of chat application for testing");
 ABSL_FLAG(std::string, address, "localhost:8080", "Address of the Oak application to connect to");
-// TODO(#1357): Use a public / secret key pair instead of bearer token credentials.
-ABSL_FLAG(std::string, room_access_token, "", "Base64-encoded public key of the room to join");
+ABSL_FLAG(std::string, room_secret, "",
```
tiziano88, line 37: Maybe rename to `room_secret_file` or `room_secret_path` for clarity? Though even `ca_cert` below should be renamed, maybe together at some point in the future.

examples/chat/client/cpp/chat.cc
```diff
@@ -158,26 +156,25 @@ int main(int argc, char** argv) {
...
+  std::string private_key;
+
+  // If no room secret, for retreiving a private key, was provided, create a fresh private/public
+  // key pair, and store the private key in a file for later use. A secure key sharing mechanims is
+  // needed to share this file with other clients.
+  if (room_secret.empty()) {
+    private_key = oak::ApplicationClient::GenerateKeyPair();
+    oak::ApplicationClient::StorePrivateKey(private_key, "chat-room.key");
+  } else {
+    private_key = oak::ApplicationClient::LoadPrivateKey(room_secret);
```
tiziano88, line 169: Personally I would have avoided this extra logic and just required that the key is always pre-generated via oak_sign or some other mechanism, and this binary just consumes it. Otherwise now we need to keep track of the key it creates, and whether that overwrites other existing keys, and where to move the key, how to document it (the flag description does not mention the file name so it seems hard to figure out what is going to happen).

To clarify what I mean, I also see this a bit similar to how git uses a SSH key for auth, but it does not need to know how to generate and store those keys.

examples/chat/client/cpp/chat.cc
```diff
@@ -158,26 +156,25 @@ int main(int argc, char** argv) {
...
-    // label requests.
-    oak::NonceGenerator<64> generator;
-    auto room_access_token_bytes = generator.NextNonce();
-    room_access_token = std::string(room_access_token_bytes.begin(), room_access_token_bytes.end());
-    LOG(INFO) << "Join this room with --address=" << address
-              << " --room_access_token=" << absl::Base64Escape(room_access_token);
-  }
+  // Use the room's secret to sign the challenge required for authenticating the client.
+  oak::identity::SignedChallenge signed_challenge =
+      oak::ApplicationClient::Sign(private_key, "oak-challenge");
```
tiziano88, line 174: Could you link to where the magic string is coming from? And perhaps make it a const?

examples/chat/module/rust/tests/integration_test.rs
```diff
@@ -154,8 +144,13 @@ impl<'a> Chatter<'a> {
             user_handle,
             base64::encode(&room_key_pair.pkcs8_public_key())
         );
-        let (channel, interceptor) = oak_tests::public_channel_and_interceptor().await;
-        // TODO(#1357): Use key pair to authenticate this client and label requests.
+
+        // Use key pair to label requests and authenticate this client.
+        let (channel, label_interceptor) =
+            oak_tests::authenticated_channel_and_interceptor(&room_key_pair.pkcs8_public_key())
+                .await;
```
tiziano88, line 151: AFAICT this does not create an authenticated client, it does not even use the private key at all, right? The next interceptor is the one that does authentication, while this does only labelling.

BTW we should probably have a short-hand interceptor that does these two things together, I think it is a very common use case.

oak/client/application_client.h
```diff
@@ -62,6 +64,20 @@ class ApplicationClient {
     return CreateChannel(addr, channel_credentials, call_credentials);
   }
 
+  // Returns a gRPC Channel connecting to the specified address, initialised with a fixed Oak Label
+  // and the given signature.
+  static std::shared_ptr<grpc::Channel> CreateChannel(
```
tiziano88, line 69: Maybe `CreatePrivateChannel` or something like that, and clarify that this does both labelling and authentication? Also, whatever name you go for, it would be nice to have consistent naming for the Rust counterpart.

oak/client/application_client.h
```diff
@@ -101,6 +117,36 @@ class ApplicationClient {
     }
     return decoded_public_key;
   }
+
+  // Generates an ed25519 key pair, and return the private key. The public key can be derived from
```
tiziano88, line 121:
```suggestion
  // Generates an ed25519 key pair, and returns the private key. The public key can be derived from
```

oak/common/oak_sign.cc
```diff
@@ -0,0 +1,103 @@
...
+
+oak::identity::SignedChallenge sign_ed25519(const std::string& private_key,
+                                            const std::string& input_string) {
+  // Compute sha256 hash of the input string
+  std::string hash_value;
+  if (!compute_sha256_hash(input_string, hash_value)) {
+    LOG(FATAL) << "Failed to compute sha256 hash.";
+  }
+
+  // Retreive private and public keys from the input private key string
```
tiziano88, line 63:
```suggestion
  // Retrieve private and public keys from the input private key string
```

oak/common/oak_sign.cc
```diff
@@ -0,0 +1,103 @@
...
+  }
+  auto digest = digest_result.ValueOrDie();
+  unsigned char* hash = digest.data();
+  // Convert the byte array into string
+  std::stringstream ss;
+  for (unsigned int i = 0; i < digest.size(); ++i) {
+    ss << (char)hash[i];
+  }
+  hashed = ss.str();
+  return true;
```
tiziano88, line 100: Is there no util library to do this?

oak/common/oak_sign.h
```diff
@@ -0,0 +1,41 @@
...
+namespace oak {
+
+// `PRIVATE KEY` tag for reading and writing from and to PEM files.
+ABSL_CONST_INIT extern const char kPrivateKeyPemTag[];
+
+// Generates an ed25519 key pair, and return the private key. The public key can be derived from
+// the private key.
+std::string generate_ed25519_key_pair();
+
+// Stores the given private key as a base64 encoded string in a PEM file in the given path.
```
tiziano88, line 29: Does it mean that it is first PEM-encoded and then base64-encoded? Doesn't PEM already base64-encode things anyways?

oak/common/oak_sign.h
```diff
@@ -0,0 +1,41 @@
...
+void store_private_key(const std::string& private_key, const std::string& private_key_path);
+
+// Signs the sha256 hash of the message using the given private key. Returns a SignedChallenge
+// containing the signed hash and the public key corresponding to the input private key.
+oak::identity::SignedChallenge sign_ed25519(const std::string& private_key,
```
tiziano88, line 34: Is there a nicer type to pass around for the private key instead of a string? e.g. `Ed25519PrivateKey`

oak/common/utils.cc
```diff
@@ -43,5 +44,46 @@ void write_file(const std::string& data, const std::string& filename) {
   t.close();
 }
 
+std::map<std::string, std::string> read_pem(const std::string& filename) {
```
tiziano88, line 47: I haven't really read through, but does https://github.com/google/boringssl/blob/master/include/openssl/pem.h have something useful perhaps? Or some other standard library for parsing PEM files? If not, could you add some tests to this logic?

oak/common/utils.cc
```diff
@@ -43,5 +44,46 @@ void write_file(const std::string& data, const std::string& filename) {
...
+  std::ofstream out_file(filename, std::ofstream::out);
+  if (!out_file.is_open()) {
+    LOG(FATAL) << "Could not open file '" << filename << "'";
+  }
+  for (const auto& item : map) {
+    std::string header = "-----BEGIN " + item.first + "-----";
+    std::string footer = "-----END " + item.first + "-----";
+
+    out_file << header << "\n";
+    out_file << item.second << "\n";
```
tiziano88, line 82: Shouldn't this also base64 encode / decode? Otherwise it seems it is pretty easy to confuse things if the items contain the string `-----BEGIN` or similar.

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -290,3 +275,12 @@ pub async fn channel_and_interceptor(label: &Label) -> (Channel, impl Into<tonic
 pub async fn public_channel_and_interceptor() -> (Channel, impl Into<tonic::Interceptor>) {
     channel_and_interceptor(&Label::public_untrusted()).await
 }
+
+/// Build a channel and a label interceptor suitable for building a client that connects to a
+/// Runtime under test. The interceptor uses the given public key to attach a public-key identity
+/// tag to every request.
+pub async fn authenticated_channel_and_interceptor(
+    public_key: &[u8],
```
tiziano88, line 283: Any more precise type to use for the key?

rbehjati (commented):
Thanks for the review @tiziano88.

examples/chat/client/cpp/chat.cc
tiziano88: That's fine, but that's not what I was talking about. I was just saying that this app should not have to generate the key itself, it can just assume it is already generated and available as a file somewhere.

examples/chat/client/cpp/chat.cc
```diff
@@ -14,29 +14,31 @@
...
 #include "oak/client/application_client.h"
-#include "oak/common/nonce_generator.h"
+#include "oak_abi/proto/identity.pb.h"
 
 ABSL_FLAG(bool, test, false, "Run a non-interactive version of chat application for testing");
 ABSL_FLAG(std::string, address, "localhost:8080", "Address of the Oak application to connect to");
-// TODO(#1357): Use a public / secret key pair instead of bearer token credentials.
-ABSL_FLAG(std::string, room_access_token, "", "Base64-encoded public key of the room to join");
+ABSL_FLAG(std::string, room_secret, "",
+          "Path to a file containing base64-encoded private key for joining the room");
```
tiziano88, line 38: Wasn't the key in PEM format? or is it also additionally base64 encoded?

examples/chat/client/cpp/chat.cc
```diff
@@ -158,26 +156,25 @@ int main(int argc, char** argv) {
...
 
-  RoomToken room_access_token;
-  if (!absl::Base64Unescape(absl::GetFlag(FLAGS_room_access_token), &room_access_token)) {
-    LOG(FATAL) << "Failed to parse --room_access_token as base64";
+  std::string room_secret = absl::GetFlag(FLAGS_room_secret);
+  std::string private_key;
+
+  // If no room secret, for retreiving a private key, was provided, create a fresh private/public
+  // key pair, and store the private key in a file for later use.
+  // TODO(#1905): A secure key sharing mechanims is needed to share this file with other clients.
```
tiziano88, line 164:
```suggestion
  // TODO(#1905): A secure key sharing mechanism is needed to share this file with other clients.
```

oak/common/oak_sign.h
tiziano88: alright up to you, though it seems a bit odd that this function is doing the parsing every time if it were called multiple times. In fact, to me it looks like this library is not very symmetric, since it has a function to store the key but not to load it, if it had functions to both load and store, I think both of them could expose more useful types instead of just strings perhaps, and only do the parsing / serialization once.

tiziano88 (approved):
Thanks!

oak/common/oak_sign.h
rbehjati: Good point. I have now wrapped everything in a new `KeyPair` class.

rbehjati (commented):
Thanks for the review @tiziano88. I've made some changes to `oak_sign`. Could you please take another look. It might be easier to only look at the last commit.

examples/chat/client/cpp/chat.cc
rbehjati: I understand, but in the meantime, we still need to generate the key-pair for the example to work. I have reworded the TODO comment.

oak/common/oak_sign.cc
```diff
@@ -35,50 +34,48 @@ using ::google::crypto::tink::Ed25519PublicKey;
 namespace oak {
 const char kPrivateKeyPemTag[] = "PRIVATE KEY";
 
-std::string generate_ed25519_key_pair() {
-  StatusOr<Ed25519PrivateKey> key_or = Ed25519SignKeyManager().CreateKey(Ed25519KeyFormat());
+Ed25519PrivateKey parse_key(const std::string& private_key) {
+  // Retrieve private and public keys from the input private key string
```
tiziano88, line 38: Strictly speaking this is just parsing the private key, right? The public key is then derived from that. I guess it's the same, but the comment suggests that the input string contains two keys (which is also confusing by the use of PEM), but I think there is actually just one key, right?

edit: I think I'm getting to the bottom of this: in EdDSA the public key is just the suffix of the private key: https://en.wikipedia.org/wiki/EdDSA#Summary

here I am decoding both test keys in our repo from https://github.com/project-oak/oak/tree/main/examples/keys/ed25519 , and if you compare them you can see that the public one is indeed a suffix of the private one:

https://gchq.github.io/CyberChef/#recipe=PEM_to_Hex()&input=LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1GTUNBUUV3QlFZREsyVndCQ0lFSU9zUWdHSzhwOVZQTGVuMk8zREQvZlh2RjNMeXdieE0zMnI0Y2VQN1h6bEgKb1NNRElRQi9qVklLVTIxSGlManEvWk82SFY5QXR1MzltcEd2V1VRMXFNSmIzYVBJL2c9PQotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0t

https://gchq.github.io/CyberChef/#recipe=PEM_to_Hex()&input=LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KZjQxU0NsTnRSNGk0NnYyVHVoMWZRTGJ0L1pxUnIxbEVOYWpDVzkyanlQND0KLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0t

cc @ipetr0v

tiziano88 (approved):

oak/common/oak_sign.cc
rbehjati: The confusion is because what `openssl` calls `ed25519` private key is different from that in RFC 8032. This is from openssl (https://github.com/google/boringssl/blob/master/include/openssl/curve25519.h):
```
// Note that, unlike RFC 8032's formulation, our private key representation
// includes a public key suffix to make multiple key signing operations with the
// same key more efficient. The RFC 8032 private key is referred to in this
// implementation as the "seed" and is the first 32 bytes of our private key.
```
What we use and store is the 64-byte `openssl` private key. `Tink` calls it `Ed25519PrivateKey`, but it is in fact a key pair. In the Rust code we call it `KeyPair`. So I named the CPP class `KeyPair` too. 

I've clarified this in the comments. But I am open to revising the naming and the comments further :)

oak/client/application_client.h
```diff
@@ -118,30 +118,34 @@ class ApplicationClient {
...
 
-    return pem_map[kPrivateKeyPemTag];
+    std::unique_ptr<oak::KeyPair> key_pair(new oak::KeyPair(pem_map[kPrivateKeyPemTag]));
+    return key_pair;
   }
 
-  // Stores the given private key as a base64-encoded string in a PEM file in the given path.
-  static void StorePrivateKey(const std::string& private_key, const std::string& filename) {
-    oak::store_private_key(private_key, filename);
+  // Stores the PEM-encoding of the private key part of the given key pair in the given path.
```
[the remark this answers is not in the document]
tiziano88: Wait a second, are you saying that the key that the C++ generates and stores to file is not compatible with the Rust client and with oak_sign?

oak/common/oak_sign.cc
tiziano88: So many standards :) thanks for looking into it; I think it is important that we have tests to ensure that we have cross-compatibility between key format and signing schemes, across languages, since otherwise I fear things may get out of sync very quickly. WDYT?

rbehjati: I agree. I am looking into this. I'll have a closer look at the Rust code. I assumed it was the same 64-byte openssl format, but the one you shared above is longer, although it still has the public key as its suffix. I'll add more tests too.

oak/common/oak_sign.cc
```diff
@@ -0,0 +1,96 @@
...
+using ::google::crypto::tink::Ed25519PrivateKey;
+using ::google::crypto::tink::Ed25519PublicKey;
+
+namespace oak {
+const char kPrivateKeyPemTag[] = "PRIVATE KEY";
+
+// Creates an Ed25519PrivateKey instance from the input private key string. The input private key
+// should be 64 bytes, composed of a 32-byte ed25519 private key, followed by a matching 32-byte
+// public key.
+Ed25519PrivateKey parse_key(const std::string& private_key) {
```
ipetr0v, line 40: What is the format of the `private_key`? Is it in base64?

oak/common/oak_sign.h
```diff
@@ -0,0 +1,62 @@
...
+
+class KeyPair {
+ public:
+  // The input `private_key` should be a 64-byte key, composed of a 32-byte ed25519 private key,
+  // followed by the corresponding 32-byte public key.
+  // Note: This parameter is called `private_key`, although it also contains the public key, because
+  // this is what `openssl` refers to as an ed25519 private key.
+  //
+  // See https://github.com/google/boringssl/blob/master/include/openssl/curve25519.h
+  KeyPair(const std::string& private_key);
```
ipetr0v, line 41: Have you checked
Can keys generated by `ring` be passed as parameters to `tink`?

oak/common/oak_sign.h
```diff
@@ -0,0 +1,62 @@
...
+  static std::unique_ptr<KeyPair> Generate();
+
+ private:
+  Ed25519PrivateKey key_;
+  KeyPair(const Ed25519PrivateKey private_key) : key_(private_key) {}
+};
+
+// Computes the sha256 hash of the unhashed input string. Returns true if hashing is successful. In
+// this case the hashed value is stored in the second input parameter.
+bool compute_sha256_hash(const std::string& unhashed, std::string& hashed);
```
ipetr0v, line 60: Is it stored as a hexademical string?
Could you please add a comment?

ipetr0v (approved):

oak/common/oak_sign.cc
ipetr0v: Does `tink` have functions to convert keys to PKCS#8?

oak/client/application_client.h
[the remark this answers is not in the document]
ipetr0v: I was saving public keys in separate files, because they were used by the clients as labels.

oak/common/oak_sign.h
rbehjati: It is an octet string (I believe), a byte array stored in a string.

oak/common/oak_sign.cc
rbehjati: Hmmm... I am sure I replied to this question, but does not seem like GitHub saved it! 
No, Tink does not have support for PKCS#8.

oak/common/pkcs8.h
```diff
@@ -0,0 +1,84 @@
...
+ public:
+  ByteArray(const std::string& in_string);
+  ByteArray(unsigned char* in_bytes, int in_len);
+  std::string ToString();
+  std::string ToBase64();
+  static ByteArray FromBase64(const std::string& in_string);
+
+  unsigned char* bytes;
+  int len;
+};
```
tiziano88, line 40: Is there no other way of storing a byte array in absl on some other library? It seems odd to have to implement such a basic class, also this particular implementation seems like it would always leak memory, or did I miss where it gets released? Is it not possible to just use a `std::string` instead of this class?

oak/common/pkcs8.h
```diff
@@ -0,0 +1,84 @@
...
+
+  // Length of  the private key. This is needed for decoding the PKCS#8 documents.
+  int private_key_len;
+};
+
+// Encode the given private and public key, wrapped in a PrivateKeyInfo object, as a PKCS#8 encoded
+// ByteArray, using the given template.
+std::unique_ptr<ByteArray> to_pkcs8(const PrivateKeyInfo& data, const Template& pkcs8_template);
+
+// Exrtact private key info from a PKCS#8 encoded private key, using the given template.
```
tiziano88, line 72:
```suggestion
// Extract private key info from a PKCS#8 encoded private key, using the given template.
```

oak/common/pkcs8.cc
```diff
@@ -0,0 +1,138 @@
...
+// unencrypted PKCS#8 byte arrays. With this template, the 32-byte
+// private key (or seed) will be placed after the `prefix`, and the 32-byte public key will be
+// inserted at the end, resulting in |prefix|private-key|middle|public-key|. With this tempplate,
+// no additional attributes can be included in the encoded byte array. This tempplate should be used
+// in combination with `kEd25519Pkcs8Template`, which specifies the location of the split between
+// `prefix` and `middle`.
+//
+// This is copied from the Rust ring crate
+// https://github.com/briansmith/ring/blob/main/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der.
+const char kBase64Ed25519Pkcs8Templace[] = "MFMCAQEwBQYDK2VwBCIEIKEjAyEA";
```
tiziano88, line 38:
```suggestion
const char kBase64Ed25519Pkcs8Template[] = "MFMCAQEwBQYDK2VwBCIEIKEjAyEA";
```

oak/common/pkcs8.h
rbehjati: I thought it would be clearer to have byte arrays instead of strings. The other alternative is to use std::vector, but I thought that might be overkill. Replaced ByteArray with string.

oak/common/pkcs8.cc
```diff
@@ -0,0 +1,111 @@
...
+// unencrypted PKCS#8 byte arrays. With this template, the 32-byte
+// private key (or seed) will be placed after the `prefix`, and the 32-byte public key will be
+// inserted at the end, resulting in |prefix|private-key|middle|public-key|. With this template,
+// no additional attributes can be included in the encoded byte array. This template should be used
+// in combination with `kEd25519Pkcs8Template`, which specifies the location of the split between
+// `prefix` and `middle`.
+//
+// This is copied from the Rust ring crate
+// https://github.com/briansmith/ring/blob/main/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der.
+const char kBase64Ed25519Pkcs8Template[] = "MFMCAQEwBQYDK2VwBCIEIKEjAyEA";
```
tiziano88, line 38: I don't really understand why there is a single string which is only used to be split into two other strings, and also why it is base64 encoded rather than just hex-escaped. But if this makes it easier to compare this logic with the one from `ring` I guess it makes sense to keep as it is?

tiziano88 (approved):

oak/common/pkcs8.cc
rbehjati: Yes. My intention is to keep this implementation similar to the one in `ring`. Using hex-escaped would not be any better, as we still need a conversion, this time using `absl::HexStringToBytes`.

tiziano88: I mean actually hex escaped, not hex encoded.

https://stackoverflow.com/questions/45612822/how-to-properly-add-hex-escapes-into-a-string-literal/45612909

Basically the bytes should just be inlined in the resulting binary, there is no need for any runtime conversion.

rbehjati: Do you mean like this: 
```
const char kBase64Ed25519Pkcs8Template[] =
    "\x30\x53\x02\x01\x01\x30\x05\x06\x03\x2b\x65\x70\x04\x22\x04\x20\xa1\x23\x03\x21\x00";
```

The embedded null at the end gets stripped. How can that be fixed without going back to byte arrays?

tiziano88: Yes that's what I meant. I don't think the null actually gets stripped, it just gets considered as an early string terminator by whatever library it's being passed to that expects C-style strings, but here it should not be treated as such, I guess the length should be provided explicitly somewhere else (but if that's complicated perhaps it's not worth it).

tiziano88: Actually, even better: https://stackoverflow.com/a/883683

rbehjati: > Actually, even better: https://stackoverflow.com/a/883683

This is the same as the string literal above. The zero at the end is  interpreted as the end of the string when  instantiating `Template`. 

I can initialize `kBase64Ed25519Pkcs8Template` like this:
```
const std::string kBase64Ed25519Pkcs8Template = std::string(
    "\x30\x53\x02\x01\x01\x30\x05\x06\x03\x2b\x65\x70\x04\x22\x04\x20\xa1\x23\x03\x21\x00", 21);
```
(21 is the actual length of the string/byte array)
But still no constant initializer. Should I go for this? Is this actually better than the base64 version?

PR #1900, opened 2021-03-05
Run checks per crate, not workspace.

wildarch:
# Checklist

      construction.

When running things like `cargo check` and `cargo clippy` on a workspace instead of a separate crate, Cargo's behaviour is a bit weird, especially regarding crate features. All crates in the workspace will be compiled with the same feature flags, which can result in output different than if the commands were run for the crates individually, e.g.
- This PR add the `rt-multi-thread` feature to the `tokio` dependency of `experimental/split_grpc/client`. Omitting this result in an error when building the crate on its own, but that is hidden from CI because the `proxy` crate depends on `tokio` with the `full` feature enabled, which transitively enables the same feature for the entire workspace.
- With the `linear-handles` feature, we currently have to awkwardly add an explicit `no-linear-handles` feature to cancel out the effects when compiling a workspace where only some crates have it enabled, and others do not support it. See https://github.com/project-oak/oak/blob/main/oak_io/Cargo.toml#L11-L21

We can resolve these issues by always running checks on individual crates, which is what this PR does 😄

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -2228,7 +2228,6 @@ name = "proxy_attestation_example"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "base64 0.13.0",
  "hex",
  "log",
  "maplit",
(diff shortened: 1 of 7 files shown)


tiziano88 (approved):
Great, thanks!

PR #1910, opened 2021-03-12
Lattices

aferr:
I tried two representations of lattices. The only two relevant files here are:
- joinSemiLatticeADT.dl
- hyperCubeLatEncoding.dl

See also the two most recent commit messages. The first one gives a way to encode join semilattices. The second one supports lattices involving both meets and joins at the syntactic level and is still solvable in HORNSAT (though it seems slower to solve in datalog than I would have thought -- I suppose the computation time is not necessarily a big deal because it is a one-time pre-computation of all the orderings/meets/joins at the start of the system rather than a cost that would re-occur during execution).

# Checklist

      construction.

### DIFF
--- a/experimental/auth-ifc2/Makefile
+++ b/experimental/auth-ifc2/Makefile
@@ -1,0 +1,11 @@
+SOUFFLE = souffle
+SOUFFLE_ARGS = ''
+
+example:
+	$(SOUFFLE) moduleCheckExample.dl && \
+		echo "flowsTo:" && cat flowsTo.csv && \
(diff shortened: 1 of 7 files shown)


tiziano88:
Converting to draft as it hasn't been acted on for a while, but feel free to revert if / when ready for review.

aferr:
@tiziano88 I think this is just a prototype used to learn and demo some things, and it has already served that purpose. Do you want to check it in? (I have no strong opinion). In a sense, keeping it as a draft forever does preserve it in a way. There is also a Raksha PR that moves some of this over there.

PR #1911, opened 2021-03-17
Fix Aggregator hash

ipetr0v:


ipetr0v:
Closed
Since the hash was updated in https://github.com/project-oak/oak/pull/1901

PR #1913, opened 2021-03-18
Use more descriptive names for options specifying a path

rbehjati:
- Specifies `path` in option names that indicate a path, as suggested in https://github.com/project-oak/oak/pull/1895#discussion_r589695057
- Fixes a few typos

### DIFF
--- a/oak/common/pkcs8.cc
+++ b/oak/common/pkcs8.cc
@@ -35,12 +35,31 @@ namespace pkcs8 {
 //
 // This is copied from the Rust ring crate
 // https://github.com/briansmith/ring/blob/main/src/ec/curve25519/ed25519/ed25519_pkcs8_v2_template.der.
-const char kBase64Ed25519Pkcs8Template[] = "MFMCAQEwBQYDK2VwBCIEIKEjAyEA";
+//
+// This is the meaning of the template bytes interpreted as ASN.1:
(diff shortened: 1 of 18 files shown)


oak/common/pkcs8.cc
```diff
@@ -35,12 +35,31 @@ namespace pkcs8 {
...
+//             :     32-BYTE PRIVATE KEY
+//             :     }
+//   48 00   33:   [0] OCTET STRING, encapsulates {
+//             :     00
+//             :     32-BYTE PUBLIC KEY
+//             :     }
+//             :   }
+const std::string kBase64Ed25519Pkcs8Template =
+    std::string({0x30, 0x53, 0x02, 0x01, 0x01, 0x30,       0x05, 0x06, 0x03, 0x2b, 0x65,
+                 0x70, 0x04, 0x22, 0x04, 0x20, (char)0xa1, 0x23, 0x03, 0x21, 0x00});
```
tiziano88, line 57: Why does only `0xa1` need the explicit type?

tiziano88 (approved):

oak/common/pkcs8.cc
rbehjati: The error says `error: narrowing conversion of '161' from 'int' to 'char' inside { }`. I think it is because, the most significant bit is 1 in the case of `0xa1`. So an explicit conversion is needed.

PR #1916, opened 2021-03-19
Migrate Abitest to linear handles

wildarch:
# Checklist

      construction.

This is a minimal set of changes to get abitest passing with `linear-handles` turned on. We should clean up unnecessary channel close calls and merge the `linear-handles` module into module 0 in a follow-up PR.

### DIFF
--- a/examples/abitest/module_1/rust/src/lib.rs
+++ b/examples/abitest/module_1/rust/src/lib.rs
@@ -52,7 +52,7 @@ fn inner_main(in_handle: u64) -> Result<(), oak::OakStatus> {
     // be waiting on the channel.
     let mut buf = Vec::<u8>::with_capacity(2);
     let mut handles = Vec::with_capacity(2);
-    oak::channel_read(in_channel, &mut buf, &mut handles)?;
+    oak::channel_read(&in_channel, &mut buf, &mut handles)?;
     if handles.len() != 1 {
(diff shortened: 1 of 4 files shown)


wildarch:
This PR will only pass the full test suite once we land #1915, but I will send it for review already so people can have a look 😉

examples/abitest/module_1/rust/src/lib.rs
```diff
@@ -61,15 +61,15 @@ fn inner_main(in_handle: u64) -> Result<(), oak::OakStatus> {
 
     let out_channel = oak::WriteHandle { handle: out_handle };
     // Wait on 1+N read handles (starting with N=0).
-    let mut wait_handles = vec![in_channel];
+    let mut wait_handles = vec![in_channel.clone()];
```
tiziano88, line 64: Do handles need to be cloned in order to be waited on now? I thought it would be nice if refs can be passed instead

tiziano88 (approved):

examples/abitest/module_1/rust/src/lib.rs
wildarch: References work too, but in this case the other handles push onto this `Vec` need to be owned to keep them from being dropped, so I've chosen to clone this handle here for convenience.

PR #1918, opened 2021-03-22
Run multiple backends

ipetr0v:
This change:
- Adds the ability to run multiple backend servers in Oak examples
- Makes running `applications` not mandatory (for examples that do not involve Oak modules)

### DIFF
--- a/examples/aggregator/backend/src/main.rs
+++ b/examples/aggregator/backend/src/main.rs
@@ -39,6 +39,12 @@ use tonic::{
 #[derive(StructOpt, Clone)]
 #[structopt(about = "Aggregator Backend")]
 pub struct Opt {
+    #[structopt(
+        long,
+        help = "Address to listen on for the gRPC server.",
+        default_value = "[::]:8888"
+    )]
+    grpc_listen_address: String,
     #[structopt(long, help = "Private RSA key file used by gRPC server.")]
     grpc_tls_private_key: String,
     #[structopt(
@@ -75,7 +81,10 @@ async fn main() -> anyhow::Result<()> {
 
     let identity = Identity::from_pem(certificate, private_key);
 
-    let address = "[::]:8888".parse().context("Couldn't parse address")?;
+    let address = opt
+        .grpc_listen_address
+        .parse()
+        .context("Couldn't parse address")?;
     let handler = AggregatorBackend::default();
 
     info!("Starting the backend server at {:?}", address);
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -575,7 +575,8 @@ struct Example {
     #[serde(default)]
     server: ExampleServer,
     #[serde(default)]
-    backend: Option<Executable>,
+    backends: HashMap<String, Executable>,
+    #[serde(default)]
     applications: HashMap<String, Application>,
     clients: HashMap<String, Executable>,
 }
@@ -626,18 +627,6 @@ struct Executable {
 }
 
 fn run_example(opt: &RunExamples, example: &Example) -> Step {
-    let application = example
-        .applications
-        .get(opt.application_variant.as_str())
-        .expect("Unsupported application variant");
-
-    let run_server = run_example_server(
-        &opt.build_server,
-        &example.server,
-        opt.server_additional_args.clone(),
-        &application.out,
-        &opt.permissions_file,
-    );
     let run_clients = Step::Multiple {
         name: "run clients".to_string(),
         steps: example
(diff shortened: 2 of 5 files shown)


runner/src/main.rs
```diff
@@ -667,25 +655,52 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
     // clients in the foreground.
     #[allow(clippy::collapsible_if)]
     let run_backend_server_clients: Step = if opt.run_server.unwrap_or(true) {
-        let run_server_clients = if opt.build_client.client_variant != NO_CLIENTS {
-            Step::WithBackground {
-                name: "background server".to_string(),
-                background: run_server,
-                foreground: Box::new(run_clients),
+        let run_server_clients = if example.applications.is_empty() {
+            if opt.build_client.client_variant != NO_CLIENTS {
```
tiziano88, line 659: Can you flip this condition, in order to avoid keeping track of the double negation in the else branch?

runner/src/main.rs
```diff
@@ -667,25 +655,52 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
+
+            let run_server = run_example_server(
+                &opt.build_server,
+                &example.server,
+                opt.server_additional_args.clone(),
+                &application.out,
+                &opt.permissions_file,
+            );
+
+            if opt.build_client.client_variant != NO_CLIENTS {
```
tiziano88, line 678: Same here.

runner/src/main.rs
```diff
@@ -667,25 +655,52 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
+                panic!("`NO_CLIENTS` variant is not supported when no applications are provided");
             }
         } else {
-            Step::Single {
-                name: "run server".to_string(),
-                command: run_server,
+            let application = example
+                .applications
+                .get(opt.application_variant.as_str())
+                .expect("Unsupported application variant");
```
tiziano88, line 668: Maybe also print the application variant, and the list of available variants?

runner/src/main.rs
```diff
@@ -667,25 +655,52 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
-        match &example.backend {
-            Some(backend) => Step::WithBackground {
-                name: "background backend".to_string(),
-                background: run(&backend, &opt.build_client, Vec::new()),
-                foreground: Box::new(run_server_clients),
-            },
-            None => run_server_clients,
+        if example.backends.is_empty() {
+            run_server_clients
+        } else {
```
tiziano88, line 693: Why does this need special treatment? Ideally the fold should cover the empty case just fine?

runner/src/main.rs
```diff
@@ -667,25 +655,52 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
+            example
+                .backends
+                .iter()
+                .fold(run_server_clients, |backend_steps, (name, backend)| {
+                    Step::WithBackground {
+                        name: name.to_string(),
+                        background: run(&backend, &opt.build_client, Vec::new()),
+                        foreground: Box::new(backend_steps),
+                    }
```
tiziano88, line 702: Could you add a comment to explain this?

runner/src/main.rs
```diff
@@ -701,20 +716,28 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
-                        .collect(),
-                },
-                Step::Single {
-                    name: "build application".to_string(),
-                    command: build_application(&application),
-                },
-            ],
+            if example.applications.is_empty() {
+                vec![]
+            } else {
```
tiziano88, line 721: Shouldn't there be a similar check here if application variant is provided but unsupported?

runner/src/main.rs
```diff
@@ -727,12 +750,16 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
             },
-            match &example.backend {
-                Some(backend) => vec![Step::Single {
-                    name: "build backend".to_string(),
-                    command: build(&backend.target, &opt.build_client),
-                }],
-                None => vec![],
+            if example.backends.is_empty() {
+                vec![]
+            } else {
```
tiziano88, line 755: Why this special case here?

runner/src/main.rs

examples/aggregator/backend/src/main.rs
```diff
@@ -39,6 +39,12 @@ use tonic::{
 #[derive(StructOpt, Clone)]
 #[structopt(about = "Aggregator Backend")]
 pub struct Opt {
+    #[structopt(
+        long,
+        help = "Address to listen on for the gRPC server.",
+        default_value = "[::]:8888"
+    )]
+    address: String,
```
tiziano88, line 47: grpc_listen_address? Maybe create a separate issue to make flags consistent across binaries, if they are not already consistent

tiziano88 (approved):

examples/aggregator/backend/src/main.rs
ipetr0v: Done
At quick glance they are already consistent

PR #1922, opened 2021-03-24
Free up more space in GitHub actions runners

tiziano88:
Make `runner` print out commands (useful to retry a single step that is
failing).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,14 +36,22 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


rbehjati (approved):

rbehjati:
Does this also help with the flaky builds? All our builds are failing with `linking with 'cc' failed: exit code: 1` when compiling random crates.

tiziano88:
> Does this also help with the flaky builds? All our builds are failing with `linking with 'cc' failed: exit code: 1` when compiling random crates.

I don't know, if they were flaky because of missing space I guess yes, but we'll see.

PR #1924, opened 2021-03-24
[RFC] TLS Remote Attestation (Certificate Authority)

ipetr0v:
TLS Remote Attestation scheme that uses a Certificate Authority server to remotely attest backend applications.


rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+## Background
+
+A lot of existing applications that would benefit from the remote attestation
+use **TLS protocol** to create secure channels with their clients. But
+integrating the remote attestation might bring up unnecessary complications in a
+form of completely rewriting the application connection logic.
+
+Thus, instead of refactoring existing applications and implementing a new
+protocol, we propose to integrate remote attestation into TLS by
+cryptographically binding **TLS certificates** to the **TEE firmware**.
```
tiziano88, line 17: TEE firmware? I thought the point is to bind them to remote attestation?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Overview
+
+In order to allow clients to use pure TLS to connect to TEE applications we
```
tiziano88, line 21:
```suggestion
In order to allow clients to use plain TLS to connect to TEE applications we
```

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+propose a Proxy Attestation Service.
+
+Proxy Attestation Service - is a service that can be used by clients to remotely
```
tiziano88, line 24: Unclear what this means -- is it a binary? an HTTP service? a gRPC service?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+attest TEE applications. The key idea is: instead of attesting TEE applications
+themselves, clients attest the proxy, which attests TEE applications on behalf
```
tiziano88, line 26: Which proxy? I thought the point is to have a CA, so maybe we should call it CA?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+  - Backend TEE applications will still be able to run on various TEE machines,
+    since the attestation logic is implemented by the proxy
+
+## Attestation
+
+<!-- From: -->
+<!-- https://docs.google.com/drawings/d/1PUHf7NW3zV-zPz6zwTGUZI6wBVxy7yIfzxsbUbCw724 -->
+<img src="images/tls_attestation_ca.png" width="850">
+
+The key idea of Proxy Attestation Service is that it works as a **Certificate
```
tiziano88, line 50: Please clarify the terminology, currently it uses proxy, service, CA for basically the same thing, we should just pick one (I think CA is the most accurate one)

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+Authority (CA)** for backend applications. It means that after attesting backend
+applications it creates **signed TLS certificates** for them, which backend
```
tiziano88, line 52: I think "backend applications" is not particularly understandable, unless somewhere first explains what the use case is (i.e. there are applications running in TEEs, and the CA not running in the TEE, and they obtain certs from the CA at startup time by performing remote attestation to the CA, etc.)

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+applications can then use to create TLS channels with clients. These
+certificates are signed by the **root certificate**, which is also being sent to
+all the clients.
+
+The crucial part here is that clients receive the root certificate after
```
tiziano88, line 57: clients receive root cert? I thought they get their cert signed by the CA?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+remotely attesting the CA, and the CA only signs certificates for backend
```
tiziano88, line 58: Is there some protocol that could be used for this? e.g. something like what letsencrypt uses?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+applications that it attested itself. It means that clients can implicitly trust
+the backend application code if they already trust in the CA code correctness.
+
+Another important note is that after attesting a backend application, the CA
+puts a **TEE report** (which is a data structure signed by the TEE platform that
```
tiziano88, line 63:
```suggestion
embeds the **TEE report** (which is a data structure signed by the TEE platform that
```

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+applications can then use to create TLS channels with clients. These
+certificates are signed by the **root certificate**, which is also being sent to
+all the clients.
+
+The crucial part here is that clients receive the root certificate after
+remotely attesting the CA, and the CA only signs certificates for backend
+applications that it attested itself. It means that clients can implicitly trust
+the backend application code if they already trust in the CA code correctness.
+
+Another important note is that after attesting a backend application, the CA
```
tiziano88, line 62:
```suggestion
After attesting a backend application, the CA
```

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+   1. **Client** trusts that this is a correct certificate, since it was sent
+      via a secure channel created during the attestation process
+   1. **Client** needs to request the root certificate only once at startup, and
+      then the client can use it for all **Server** connections
+1. **Client** connects to the **Server** using TLS
+1. **Server** sends **Client** a TLS certificate previously signed by **CA**
+1. **Client** verifies that the TLS certificate provided by the **Server** was
+   signed by the **CA**’s root certificate
+   1. **Client** also verifies the TEE report provided in the **Server**’s
+      certificate
```
tiziano88, line 136: This is the important step, right? Should it mention how to actually verify the report? in particular, what about the public key hash that was in the TEE report?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,161 @@
...
+
+Several important notes:
+
+- Client needs to **trust** in the CA and Server implementation correctness:
+  - Main requirements are:
+    - CA should correctly attest the Server and sign right certificates
+    - Server shouldn’t leak its TLS keys
+  - In order to trust, client needs to be able to access the code of the CA and
+    the Server, and also able to **reproducibly build** binaries on their
+    machine
```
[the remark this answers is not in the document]
ipetr0v: Updated

rfcs/01924_tls_attestation_ca.md
ipetr0v: Not sure
First that comes to mind is [Kerberos](https://en.wikipedia.org/wiki/Kerberos_(protocol)), but as far as I know it doesn't distribute certificates

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+Thus, instead of refactoring existing applications and implementing a new
+protocol, we propose to integrate remote attestation into TLS by
+cryptographically binding **TLS certificates** to the **TEE firmware**.
+
+## Overview
+
+In order to allow applications to use TLS for remote attestation, we propose to
+use a **Certificate Authority (CA)** that generates signed TLS certificates for
+TEE applications. The main idea is that instead of attesting TEE applications
+themselves, clients attest the CA, which attests TEE applications on behalf of
```
conradgrobler, line 24: nit: "attest" is used inaccurately throughout. Attesting is providing proof, not validating the proof. The server attests to the software it is running and the client checks or validates the attestation.

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+### Attestation Workflow
+
+The detailed workflow of the remote attestation is represented by 2 stages.
+
+First stage corresponds to the backend application attestation, which happens
+only once when the backend application starts:
+
+1. **Server** generates a TLS key pair
+1. **Server** connects to the **CA**
+1. **CA** attests the **Server** using a remote attestation protocol
```
conradgrobler, line 108: This should be a two-way attestation using something like EKEP, so that the server is sure about it can trust the CA and the CA is sure it can trust the server.

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+   1. If the **Server**’s TLS certificate is not valid, then the **Client**
+      closes the connection and aborts the protocol
+1. Since **Client** trusts **CA** (and has its root certificate) it can now
+   trust **Server** and establish a secure connection
+
+<!-- From: -->
+<!-- https://sequencediagram.googleplex.com/view/6490702202535936 -->
+<img src="images/tls_attestation_ca_workflow_client.png" width="850">
+
+Several important notes:
```
conradgrobler, line 153: This does not address the TOCTOU issue. For the attestation to have any meaning, the server must be incapable of changing any executable code in its memory while still retaining access to private key.

rfcs/01924_tls_attestation_ca.md
dreemkiller: Using attestation to establish trust in the CA implies that the CA itself is running inside some sort of TEE environment (or perhaps using a TPM). This may not always be necessary (for example if the entity operating the client is also operating the CA, in which case they already trust it).

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+TLS certificate** for it, which a TEE application can then use to create TLS
+channels with clients. Such certificates are signed by the CA’s root
+certificate.
+
+While generating a TLS certificate, the CA embeds the **TEE report** (which is a
+data structure signed by the TEE platform that contains information identifying
+the code that is running inside the TEE) into it, thus binding the TLS
+certificate to the TEE firmware. This way clients can check what payload is
+being run inside the TEE they are connecting to.
+
```
dreemkiller, line 42: There may be value in not embedding the full **TEE report** instead providing something that is easier for a client with limited resources to parse - perhaps just certain elements of the report (I'm thinking here about a Cortex-M client, that has on the order of 10s of kilobytes of RAM/Flash)

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+- **Server**
+  - TEE application that processes requests from TLS clients
+  - Runs inside the TEE environment
+- **Client**
+  - Client application that connects to applications using TLS
+  - Runs on a user device or inside of a browser
+- **Provider**
+  - TEE platform provider that can confirm authenticity of TEE platform keys
+  - It’s important to note that the Provider is an external server (i.e. belongs
+    to AMD)
```
dreemkiller, line 92: Depending on which platforms are supported, it may not be a server. For AWS Nitro Enclaves, there is no attestation server. Instead, they have long-term keys that are used. This is, of course, a very different trust model from other platforms, and may not apply here, but it's worth considering if that model can be supported in the descriptions.

rfcs/01924_tls_attestation_ca.md
dreemkiller: Could the server just rely on the normal PKI to establish trust in the CA instead of using EKEP?

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+
+1. **Server** generates a TLS key pair
+1. **Server** connects to the **CA**
+1. **CA** attests the **Server** using a remote attestation protocol
+   corresponding to the Server’s TEE
+   1. When requesting a TEE report from the TEE firmware, **Server** also puts
+      the hash of its TLS public key into it
+   1. For example, remote attestation could be done via
+      [#1929](https://github.com/project-oak/oak/pull/1929)
+1. **Server** sends **CA** a certificate signing request
```
dreemkiller, line 114: Could the server put it's CSR in the attestation token, removing a step?

rfcs/01924_tls_attestation_ca.md
dreemkiller: As long as the server is using code/data from the original attestation hash, self-modification by the server code should still be allowed, no? (Not that it's necessarily a good idea or useful)

rfcs/01924_tls_attestation_ca.md
ipetr0v: I moved it into a different step, so it wouldn't interfere with the attestation scheme that is used between the CA and the Server.

rfcs/01924_tls_attestation_ca.md
conradgrobler: I guess it depends on the type of TEE. If it is a full virtual machine with a separate OS, the entire OS must be validated to ensure that there is no way to e.g. stop this process and start a different one after the certificate is received.

rfcs/01924_tls_attestation_ca.md
conradgrobler: Yes, depending on who is operating which parts and the existing trust relationships.

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+   1. Communication happens over a secure channel creates as part of the remote
```
rbehjati, line 115:
```suggestion
   1. Communication happens over a secure channel created as part of the remote
```

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+    - Server shouldn’t leak its TLS keys
+  - Client can gain trust by accessing the code of the CA and the Server, and
+    also by **reproducibly building** binaries on their machine
+    - Though it could be done by a third party reviewer that the Client can
+      trust
+- Client only needs to **attest the CA once** and then use its root certificate
+  for multiple connections to any Server
+- Remote attestation logic on Servers can be implemented as a **separate
+  binary**, that attests to the CA, receives a signed certificate and puts it as
+  an input to the TEE application
```
rbehjati, line 167: I don't quite understand this sentence. What kind of input is this? It sounds like the application still needs to handle this input, which to me sounds like a remote-attestation specific logic.

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+TEE applications. The main idea is that instead of attesting TEE applications
+themselves, clients attest the CA, which attests TEE applications on behalf of
+the clients. Clients can then download the CA’s **root certificate** and use it
+for connecting to TEE applications via TLS.
+
+<!-- From: -->
+<!-- https://docs.google.com/drawings/d/1PUHf7NW3zV-zPz6zwTGUZI6wBVxy7yIfzxsbUbCw724 -->
+<img src="images/tls_attestation_ca.png" width="850">
+
+When a TEE application starts, the CA remotely attests it and creates a **signed
```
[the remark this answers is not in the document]
ipetr0v: Correct. Updated the text

rfcs/01924_tls_attestation_ca.md
ipetr0v: Done

rfcs/01924_tls_attestation_ca.md
ipetr0v: The input here is the certificate.
Since the application used TLS already, it needed a certificate.
We are just changing this certificate to the one created by the Attestation CA.

rfcs/01924_tls_attestation_ca.md
```diff
@@ -0,0 +1,169 @@
...
+## Objective
+
+Remotely attest applications in Trusted Execution Environments (TEEs) over the
+TLS protocol.
+
+## Background
+
+A lot of existing applications that would benefit from the remote attestation
+use **TLS protocol** to create secure channels with their clients. But
+integrating the remote attestation might bring up unnecessary complications in a
```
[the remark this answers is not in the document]
ipetr0v: The main problem is `completely rewriting the application connection logic`. For example, a legacy application used TLS to accept client connections, but now it would require something like EKEP instead etc.
Though I haven't found a dedicated report that describes problems of upgrading legacy applications to remote attestation.

tiziano88:
Hey, is this still something we should consider merging? Sorry for the delay

ipetr0v:
I think both of the RA RFCs should be merged, since they are just RFCs.
And that they will not be lost in the Github history.

tiziano88 (approved):
I think we have explored this idea in various different ways since this PR, but it is worth checking it in so we have a record in the GitHub history.

conradgrobler (approved):

PR #1925, opened 2021-03-24
Add a simple oak-functions loader

### LINKED ISSUE #1923: Simple "Oak Functions" loader and wasm engine
Related to #1919:
Implement a simple HTTP server that spawns a new wasm engine for every request and runs a very simple input wasm module.

rbehjati:
Fixes #1923.

# Checklist

      construction.


oak_functions/loader/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[features]
+# Feature that allows logging. Should be use only for debugging purposes.
+oak-unsafe = []
+default = []
+
+[dependencies]
+anyhow = "*"
+env_logger = "*"
+http = "*"
+hyper = { version = "*", features = ["full"] }
```
tiziano88, line 19: Is there any chance we can have less than `full` here?

oak_functions/loader/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[cfg(test)]
+mod tests;
+
+/// Command line options for the Oak loader.
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(about = "Oak Functions Loader")]
+pub struct Opt {
+    #[structopt(long, default_value = "8080", help = "Server Port number.")]
+    server_port: u16,
```
tiziano88, line 31: Should this be `http_listen_address` or similar? I thought we use something like that elsewhere, also it seems more explicit.

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[derive(Clone)]
+struct WasmEngine {
+    // wasm module to be served on each invocation.
+    module: Arc<wasmi::Module>,
+}
+
+impl WasmEngine {
+    async fn handle_request(&self, req: Request<Body>) -> anyhow::Result<Response<Body>> {
```
tiziano88, line 42: Could you add a simple benchmark for this? I would be curious to see how long it actually takes to start the new wasm engine each time. Can be a separate PR.

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+use log::info;
+use std::{fs, net::SocketAddr, sync::Arc};
+
+const MAIN_FUNCTION_NAME: &str = "main";
+
+/// An HTTP server that serves the `main` function of a wasm module on each invocation.
+pub struct WasmServer {
+    /// Server address to listen for client requests on.
+    address: SocketAddr,
+    /// The wasm engine that runs the given wasm module per incoming request.
```
tiziano88, line 31:
```suggestion
    /// The Wasm engine that runs the given wasm module per incoming request.
```

and elsewhere (see https://webassembly.org/)

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+        // TODO(#1919): Get the actual response from the wasm module via ABI functions.
+        http::response::Builder::new()
+            .status(StatusCode::BAD_REQUEST)
+            .body(Body::from("Not implemented yet.\n"))
+            .context("Couldn't create response")
+    }
+}
+
+impl WasmServer {
+    /// Create a WasmServer to listen on the given port, serving the given wasm module.
```
tiziano88, line 67:
```suggestion
    /// Creates a [`WasmServer`] instance listening on the given port, serving the given wasm module.
```

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+/// An HTTP server that serves the `main` function of a wasm module on each invocation.
+pub struct WasmServer {
+    /// Server address to listen for client requests on.
+    address: SocketAddr,
+    /// The wasm engine that runs the given wasm module per incoming request.
+    wasm_engine: WasmEngine,
+}
+
+#[derive(Clone)]
+struct WasmEngine {
```
tiziano88, line 36: Should we call this `WasmHandler` or something like that? I think that makes it clearer that it is ephemeral and corresponds to a request handler.

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+            .body(Body::from("Not implemented yet.\n"))
+            .context("Couldn't create response")
+    }
+}
+
+impl WasmServer {
+    /// Create a WasmServer to listen on the given port, serving the given wasm module.
+    pub fn create(address: &str, wasm_path: &str) -> anyhow::Result<Self> {
+        let wasm_module_bytes =
+            fs::read(&wasm_path).with_context(|| format!("Couldn't read file {}", wasm_path))?;
```
tiziano88, line 70: I think it would be more decoupled if the caller (in our case main) did the file reading, and this fn directly got the bytes of the wasm module to instantiate. It will also make testing easier later on.

oak_functions/loader/src/tests.rs
```diff
@@ -0,0 +1,55 @@
...
+        WasmServer::create(&address, TEST_WASM_MODULE_PATH).expect("Couldn't create the server");
+
+    let server_fut = server.start(notify_receiver);
+    let client_fut = create_client(port, notify_sender);
+
+    let (res, _) = tokio::join!(server_fut, client_fut);
+    assert!(res.is_ok());
+}
+
+async fn create_client(port: u16, notify_sender: tokio::sync::oneshot::Sender<()>) {
```
tiziano88, line 38: This is not just creating the client, it is also making a request, right?

tiziano88 (commented):
Looks great already! Thanks!

ipetr0v (approved):

oak_functions/loader/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+    wasm_path: String,
+}
+
+/// Main execution point for the Oak Functions Loader.
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    if cfg!(feature = "oak-unsafe") {
+        env_logger::init();
+    } else {
+        eprintln!("No debugging output configured at build time");
```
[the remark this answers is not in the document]
tiziano88: there is no logger installed in this branch :)

oak_functions/loader/src/server.rs
rbehjati: I have added a benchmark test, using Rust `Bencher`. This is currently not being run by the runner. I can fix that in a separate PR. 

Here is the output: 
```
test tests::bench_wasm_handler ... bench:      25,608 ns/iter (+/- 976)
```
Is this the type of test you were thinking of?

rbehjati (commented):
Thanks for the reviews.

oak_functions/loader/Cargo.toml
tiziano88: AFAICT client is only needed for tests, so maybe it should be in dev-deps?

oak_functions/loader/src/server.rs
tiziano88: Nice! That time seems low compared to what I expected, so I guess that's good news. (I expected ~100ms roughly). Adding it to runner later on sounds good, although we should think about how to make sure we don't regress (just running the bench does not do much, it can easily climb up to 10s per request, and nothing will fail, unless we add additional logic to check that the time stays below a given threshold, e.g. 50ms for this function. Maybe Bencher already does this?

tiziano88: BTW this crate seems to do something similar to what I'm thinking: https://bheisler.github.io/criterion.rs/book/user_guide/command_line_options.html#baselines

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,121 @@
...
+use log::info;
+use std::{net::SocketAddr, sync::Arc};
+
+const MAIN_FUNCTION_NAME: &str = "main";
+
+// An ephemeral request handler with a Wasm module for handling the requests.
+#[derive(Clone)]
+pub(crate) struct WasmHandler {
+    // Wasm module to be served on each invocation.
+    module: Arc<wasmi::Module>,
```
tiziano88, line 31: Not sure I see why this needs to be an `Arc` instead of just `wasmi::Module`. Is it because of the lifetimes involved in `wasmi::ModuleInstance::new`? If so, could you add a comment?

oak_functions/loader/src/server.rs
```diff
@@ -0,0 +1,121 @@
...
+/// An HTTP server that serves the `main` function of a Wasm module on each invocation.
+pub struct WasmServer {
+    /// Server address to listen for client requests on.
+    address: SocketAddr,
+    /// The Wasm handler that runs the given Wasm module per incoming request.
+    wasm_handler: WasmHandler,
+}
+
+impl WasmServer {
+    /// Creates a [`WasmServer`] instance listening on the given port, serving the given Wasm
```
tiziano88, line 74: I don't think this actually starts listening on the given port / address, though maybe `create` and `start` could be combined in one (up to you); or is there anything useful that a non-started `WasmServer` can be used for?

oak_functions/loader/src/tests.rs
```diff
@@ -0,0 +1,78 @@
...
+use hyper::client::Client;
+
+use std::fs;
+
+const TEST_WASM_MODULE_PATH: &str = "testdata/non-oak-minimal.wasm";
+
+#[tokio::test]
+async fn test_server() {
+    let _ = env_logger::builder().is_test(true).try_init();
+    let port = 9999;
```
tiziano88, line 26: I guess at some point we will want to randomize this.

oak_functions/loader/src/tests.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    // TODO(#1919): Check the response.
+
+    notify_sender
+        .send(())
+        .expect("Couldn't send completion signal.");
+}
+
+extern crate test;
+use test::Bencher;
```
tiziano88, line 62: Group all the `use` together?

tiziano88 (approved):

oak_functions/loader/src/main.rs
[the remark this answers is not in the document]
tiziano88: `eprintln` is correct, we want to print to stderr, not stdout, right?

rbehjati (commented):
Thanks!

oak_functions/loader/src/server.rs
rbehjati: That crate looks interesting. But I think for that to work on the CI builds we have to tweak some GitHub action scripts. For now I am calling `bench` directly on the `bencher` to be able to get the results and have assertions. I have also created #1933 to track benchmarking.

oak_functions/loader/src/main.rs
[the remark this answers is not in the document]
tiziano88: Well, this line is not the "output" of the program, it is just diagnostic, the same as a `log::info` basically, right? I think in general the kind of programs we usually write don't have any output per se, so I don't expect them to ever print anything to stdout.

oak_functions/loader/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+#[bench]
+fn bench_wasm_handler(bencher: &mut Bencher) {
+    let summary = bencher.bench(|bencher| {
+        let wasm_module_bytes =
+            fs::read(TEST_WASM_MODULE_PATH).expect("Couldn't read test Wasm module");
+        let wasm_handler =
+            WasmHandler::create(&wasm_module_bytes).expect("Couldn't create the server");
+        let request = hyper::Request::builder()
+            .method(http::Method::GET)
+            .uri("http://localhost:8080")
```
tiziano88, line 73: I was a bit confused seeing this line, I thought it was doing an actual request to this endpoint, while actually I guess it is just to make the request build, right? If so, maybe change it to example.com or something more obviously unrelated to the rest of the code? up to you

oak_functions/loader/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+            .unwrap();
+        bencher.iter(|| wasm_handler.handle_request(&request));
+    });
+
+    // When running `cargo test` this benchmark test gets executed too, but `summary` will be `None`
+    // in that case. So, here we first check that `summary` is not empty.
+    if let Some(summary) = summary {
+        // We expect the `mean` time for loading the test Wasm module and running its main function
+        // to be less than 50ms.
+        assert!(summary.mean < 50_000.0);
```
tiziano88, line 84: Nice! I think this is probably enough for what we need, and it is fine if we need to increase this number at some point, at least it will be obvious that something is taking longer. (BTW it is a bit odd that `summary.mean` returns an `f64` but it's unclear what unit that is, I guess ns, but it's not written explicitly anywhere obvious when I looked around)

tiziano88 (approved):

oak_functions/loader/src/main.rs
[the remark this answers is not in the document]
rbehjati: I see. Makes sense.

oak_functions/loader/src/tests.rs
rbehjati: I agree, `example.com` is much better. Fixed. Also added a comment.

oak_functions/loader/src/tests.rs
rbehjati: Yes. I had to experiment with it a bit to figure out the unit. Added a comment.

tiziano88:
Is this PR blocked by the CI failures?

rbehjati:
> Is this PR blocked by the CI failures?

Yes. It keeps failing with `No space left on device`. Not sure how to do some diagnosis. There are no logs from the `runner` when this happens.

PR #1928, opened 2021-03-25
Run examples with no clients

ipetr0v:
This change restores the ability to run examples with no clients.
Required for `reproducibility` builds.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -658,14 +658,7 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
     #[allow(clippy::collapsible_if)]
     let run_backend_server_clients: Step = if opt.run_server.unwrap_or(true) {
         let run_server_clients = if example.applications.is_empty() {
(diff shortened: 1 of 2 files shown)


runner/src/main.rs
```diff
@@ -659,10 +659,10 @@ fn run_example(opt: &RunExamples, example: &Example) -> Step {
...
         let run_server_clients = if example.applications.is_empty() {
             if opt.build_client.client_variant == NO_CLIENTS {
-                panic!(
-                    "`{}` client variant is not supported when no applications are provided",
-                    NO_CLIENTS
-                );
+                Step::Multiple {
+                    name: "run clients (empty)".to_string(),
+                    steps: vec![],
+                }
```
tiziano88, line 665: Why is this special case necessary? Wouldn't `run_clients` be naturally empty already?

tiziano88 (approved):

runner/src/main.rs

PR #1929, opened 2021-03-25
[RFC] TLS Remote Attestation

ipetr0v:
TLS Remote Attestation scheme that uses Exported Keying Material to cryptographically bind TLS sessions to the TEE platform.


rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
+# RFC#01929: Remote Attestation via TLS
+
+## Objective
+
+Remotely attest applications in Trusted Execution Environments (TEEs) over the
+TLS protocol.
```
tiziano88, line 6: Please generalize the RFC to be about arbitrary assertions / claims, as discussed. TEE remote attestation may be one prominent example in the RFC, but the protocol must be more generic.

e.g. see https://tools.ietf.org/html/rfc5705#section-1 :

> A typical example is DTLS-SRTP [DTLS-SRTP], a key
   management scheme for the Secure Real-time Transport Protocol (SRTP)
   that uses DTLS to perform a key exchange and negotiate the SRTP
   [RFC3711] protection suite and then uses the DTLS master_secret to
   generate the SRTP keys.

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+
+Thus, instead of refactoring existing applications and implementing a new
+protocol, we propose to integrate remote attestation into TLS by
+cryptographically binding **TLS sessions** to the TEE platform.
+
+## Overview
+
+In order to allow applications to use TLS for remote attestation, we propose to
+cryptographically bind assertions to a TLS session, by deriving
+[Exported Keying Material](https://tools.ietf.org/html/rfc5705) from the TLS
```
tiziano88, line 23:
```suggestion
[Exported Keying Material](https://tools.ietf.org/html/rfc5705) from an already established TLS
```

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+- _Length_
+  - Bit length of the generated value
+
+It’s important to note that with a fixed set of parameters Exported Keying
+Material value for a given TLS session stays the same - i.e. it’s a
+deterministic algorithm. Thus if both client and server negotiate corresponding
+parameters (or they are constant), they can derive equal Exported Keying
+Material values.
+
+Exported Keying Material is prefixed with the purpose string
```
tiziano88, line 107: Unclear what this means; AFAICT it is not true that the exported keying material is prefixed with this; rather, it must be prefixed with this before being used

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+`“TLSAttestationV1”`. This purpose string is used to disambiguate the proper use
```
tiziano88, line 108: Rename the purpose to something more generic, not related to attestation.

tiziano88, line 108: Clarify whether quotes are included or not (and use straight quotes anyways).

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+of the Exported Keying Material, i.e. helps mitigate confusion where Exported
+Keying Material that is generated for one purpose is misused for another.
+
+Exported Keying Material that was derived with a specific set of parameters and
+prefixed with the purpose string will be called a **Token** throughout this
+document. Such Token is derived separately with a different set of parameters
+for any given [assertion](#assertions) type provided by the firmware.
```
tiziano88, line 115: No firmware is involved in assertions in general (e.g. for a public key assertion)

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+
+##### Parameters
+
+Because Token derivation involves a TLS session master secret, TLS server should
+guarantee **Perfect Forward Secrecy**, which is defined as follows:
```
tiziano88, line 120: citation needed

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+
+| TLS master key will not be compromised even if long-term secrets used in its exchange are compromised |
+| ----------------------------------------------------------------------------------------------------- |
+
+This means that if Token parameters are leaked, an attacker would still need the
+TLS session master secret, so they won’t be able to reconstruct the Token.
+
+In order to guarantee _Perfect Forward Secrecy_ the server needs to:
```
tiziano88, line 128: citation needed

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+  - Separate Label values for clients and servers are necessary to prevent
+    ambiguity and confusion about Label usage
+- Version
+  - Necessary to allow protocol upgrades in the future
+
+##### Context
+
+**Context** is a parameter used to derive separate Tokens for different
+[assertion](#assertions) types provided by the firmware. Context value is
+represented by an assertion type, for which a Token is derived.
```
tiziano88, line 162: In what sense it is "represented by"? I thought it **is** an assertion type? But also not sure it is defined what assertion type means.

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+      the provided **Token**
+1. Request a **TEE report** from hardware
+1. Firmware generates and signs a **TEE report**
+   1. The signed **TEE report** also gives clients information about the TEE
+      itself (its version and etc.) and about the code that is running inside
+      the TEE
+1. Combine the hardware generated **TEE report** with a certificate signed by
+   the **Provider**’s root key
+   1. This is required for checking the validity of **TEE reports** and TEE
+      platform keys that were used to sign them
```
tiziano88, line 224: This is only about TEE-related assertions, right?

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+
+Since Client and Server can exchange multiple assertions, each assertion should
+contain a **Token** that was derived using a separate Context parameter.
```
tiziano88, line 227: Not necessarily. e.g. if the server has 10 key pairs, it will bind all of the corresponding assertions to the same token, right?

ipetr0v: How would it bind all these keys to the same token?

rfcs/01929_tls_attestation.md
ipetr0v: Done

rfcs/01929_tls_attestation.md
ipetr0v: Done

rfcs/01929_tls_attestation.md
ipetr0v: It is defined further in the RFC `[assertion](#assertions)`
Updated to `Context value is equal to the assertion type`

rfcs/01929_tls_attestation.md
```diff
@@ -0,0 +1,258 @@
...
+
+Currently we are using the following assertion types:
+
+- `intel_sgx_ecdsa_0_1_report`
+- `amd_sev_snp_0_1_report`
+- `vtpm_public_key`
+- `eddsa_public_key`
+
+Assertion types that represent TEE reports also include _Trusted Computing base
+(TCB)_ versions.
```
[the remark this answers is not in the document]
ipetr0v: Removed as unnecessary

rfcs/01929_tls_attestation.md
ipetr0v: Correct, changed to optional step

tiziano88 (approved):

Issue #1897, opened 2021-03-03
Weather service with Private Information Retrieval


ipetr0v:
As a new example of PIR in Oak - we can create an Oak application for anonymous weather information retrieval.
This application will accept user requests (that contain their location) and request weather information from an external service. As a simple example - for each request it can either download the whole database or split it in chunks.

The difference from existing Trusted Database example is that this application will download the database dynamically from an external service, whereas Trusted Database already keeps the static database in memory.

For requesting weather information we can use the following service: https://openweathermap.org/bulk

ipetr0v:
I think the label design could be either of this:
- The application has a single trusted Oak Node (clients assign its hash as a label) that accepts all client requests. And clients trust it to 1) correctly declassify information by requesting an external service and 2) send the response back to the original client.
- The application creates a separate Oak Node for each client (clients still assign its hash as a label as well as client identity label) that accept requests from a specific client. And clients only trust it to 1) correctly declassify information by requesting an external service.
    - This may require having disjunction labels, because in order to allow the Node to declassify requests the hash component should not be in conjunction with client label component

cc @tiziano88

tiziano88:
I don't expect there should be any declassification happening. The label should just be that of the client public key.

tiziano88:
I suggest we implement this directly on top of #1919, it will be a good proof of concept for the Functions loader.

tiziano88:
We should figure out the format of the weather data first, and also (later on) think about how to refresh it periodically (see #1930 ).

@ipetr0v could you start thinking about this, in a way that connects with the rest of the Oak Functions Loader that @rbehjati started in #1925 ?

ipetr0v:
I think the main question right now is whether we want to have database entries as:
- Protobuf arrays
  - Binary format, will have less problems with performance
  - Hard to use by developers, since they would need to define each structure twice: in `.proto` file, and as a `struct` in Rust
    - This is needed for writing a data formatter, that downloads data and puts in into the Oak store for further use by a Wasm module
- JSON arrays
  - Similar to existing no-SQL databases (like [Redis](https://redis.io/))
  - Easy to use by developers

ipetr0v:
> For the entries (keys and values) in the database, the application can use whatever format it wants, either binary or text.

Yes, I'm referring to the entries.

> Why would developers define each struct twice? The point of protobuf is that it generates the Rust code from the proto definition, right? Or did you mean something else?

[...]

PR #1937, opened 2021-03-31
Add Weather Service example

ipetr0v:
This change adds a Weather Service example.

Fixes https://github.com/project-oak/oak/issues/1897


examples/weather_service/backend/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+    Ok(())
+}
+
+/// Loop for requesting a new database version from `url`.
+async fn update_database_loop(
+    database_url: Url,
+    application_url: Url,
+    database: Arc<Mutex<Database>>,
+) {
+    loop {
```
tiziano88, line 87: I wonder if it's simpler to just have this application be invoked at regular interval (e.g scheduled by something like cron) instead of having a loop built in here. Anyways fine to keep the loop, but I'm curious if you considered both approaches already

examples/weather_service/backend/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+                *database_entries = updated_database_entries;
+            }
+            Err(error) => error!("Error loading database: {:?}", error),
+        }
+
+        // Create a separate asynchronous task for sending the database to Oak.
+        tokio::spawn(send_database(application_url.clone(), database.clone()));
+
+        // Wait for a minute before updating the database.
+        tokio::time::delay_for(tokio::time::Duration::from_secs(60)).await;
```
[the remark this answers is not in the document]
ipetr0v: It's still a prototype, and it will be very easy to change, once we have this functionality in Oak

examples/weather_service/backend/src/main.rs
ipetr0v: Agree, but it may be hard to setup in CI
Only if we run this app once in the CI as a `backend`

examples/weather_service/module/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+anyhow = "*"
+log = "*"
+oak = "=0.1.0"
```
wildarch, line 14: Could you enable the `linear-handles` feature for this new example? We will soon switch to that as the default behaviour 😄

examples/weather_service/module/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+prost = "*"
+serde = "*"
+quick-xml = { version = "*", features = ["serialize"] }
+
+[dev-dependencies]
+assert_matches = "*"
+env_logger = "*"
+maplit = "*"
+oak_abi = "=0.1.0"
+oak_runtime = "=0.1.0"
```
wildarch, line 27: Also `linear-handles` here.

examples/weather_service/module/Cargo.toml
tiziano88: The plan is for this example to be built on the oak_functions ABI, not the current oak ABI, so I think you can ignore this :) @ipetr0v could you confirm?

ipetr0v: Yes, this example will only work on `oak_functions` and is not related to `oak_runtime`

PR #1938, opened 2021-03-31
Update runner to remove generated files after running tests

rbehjati:
Update runner to remove generated files after running tests, to avoid the `No space left on device` error in CI builds.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -86,7 +86,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             Command::RunExamples(ref opt) => run_examples(&opt),
             Command::BuildServer(ref opt) => build_server(&opt),
             Command::RunTests => run_tests(),
-            Command::RunCargoTests => run_cargo_tests(),
+            Command::RunCargoTests(ref opt) => run_cargo_tests(opt.cleanup),
             Command::RunBazelTests => run_bazel_tests(),
             Command::RunTestsTsan => run_tests_tsan(),
             Command::Format => format(),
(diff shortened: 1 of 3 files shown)


rbehjati:
@tiziano88 I could also make `run_cargo_clippy` and `run_cargo_docs` separate CI steps. WDYT?
Target files generated by `run_cargo_clippy` take up 13G. But `run_cargo_docs` adds only about 2G.

runner/src/main.rs
```diff
@@ -1379,14 +1379,41 @@ fn run_cargo_fmt(mode: FormatMode) -> Step {
...
+                                command: Cmd::new(
+                                    "cargo",
+                                    &["test", &format!("--manifest-path={}", &entry)],
+                                ),
+                            },
+                            Step::Single {
+                                name: "cleanup".to_string(),
+                                command: Cmd::new(
+                                    "rm",
+                                    &["-rf", &entry.replace("Cargo.toml", "target")],
```
tiziano88, line 1403: Is this going to delete the entire hard disk if entry contains an empty string or something like that? Seems a bit risky, maybe add some additional check just in case?

runner/src/main.rs
```diff
@@ -1379,14 +1379,41 @@ fn run_cargo_fmt(mode: FormatMode) -> Step {
...
+                        ],
+                    }
+                } else {
+                    Step::Single {
+                        name: entry.clone(),
+                        command: Cmd::new(
+                            "cargo",
+                            &["test", &format!("--manifest-path={}", &entry)],
+                        ),
+                    }
```
tiziano88, line 1415: I think this (or at least the Cmd part) is the same in both, right? So maybe extract to a separate variable.

tiziano88 (approved):

tiziano88:
> @tiziano88 I could also make `run_cargo_clippy` and `run_cargo_docs` separate CI steps. WDYT?
> Target files generated by `run_cargo_clippy` take up 13G. But `run_cargo_docs` adds only about 2G.

Wait, are the files generated by clippy not reused when running tests? If they are not, then I guess we should figure out why. Maybe different flags to cargo?

Also possibly related: https://github.com/project-oak/oak/issues/1162

runner/src/main.rs
rbehjati: Hopefully not :D 
Added some checks :)

runner/src/main.rs
rbehjati: Made a closure that takes the name of the step as an argument.

rbehjati:
@tiziano88 The flag I added to `run-cargo-tests` has made it a new step. Could you please change the config so that `run-cargo-tests` is no longer required. Instead we have to add `run-cargo-tests --cleanup` as a required step.

tiziano88:
> @tiziano88 The flag I added to `run-cargo-tests` has made it a new step. Could you please change the config so that `run-cargo-tests` is no longer required. Instead we have to add `run-cargo-tests --cleanup` as a required step.

Done

PR #1942, opened 2021-04-01
Update hashes

### LINKED ISSUE #1940: Hashes differ after enabling sccache
As part of #1165 , it looks that all the hashes of binaries and wasm modules changed, likely because of the different set of flags passed to rustc: 

https://github.com/project-oak/oak/pull/1165#issuecomment-811819238

Also probably in the same PR I also accidentally made it so that errors produced by runner are ignored in CI; I think the issue is with this script:

https://github.com/tiziano88/oak/blob/daa0669a19d32c39cd5e09130e4e56f63093c668/scripts/runner#L1-L5

Because by default bash will just keep executing all the commands and only report the status code of the last command in the script! We should at least set `set -o errexit`, or, better, include the `/scripts/common` script.

tiziano88:
In #1940 the volume mount point was changed, which results in different
Wasm module hashes.

Also the runner script was not actually failing correctly, which is now
fixed.

Fix #1940

# Checklist

      construction.


tiziano88:
@rbehjati if this works, I'll submit it instead of #1941

rbehjati (approved):
Merging this instead of #1941 sounds like a good idea.

PR #1943, opened 2021-04-01
Add runner command for building oak-functions loader

rbehjati:
# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -58,7 +58,7 @@ impl WasmHandler {
         // TODO(#1919): Get the actual response from the Wasm module via ABI functions.
         http::response::Builder::new()
             .status(StatusCode::BAD_REQUEST)
-            .body(Body::from("Not implemented yet.\n"))
+            .body(Body::from("Welcome to Oak Functions!\n"))
             .context("Couldn't create response")
     }
 }
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -292,15 +292,8 @@ fn build_server(opt: &BuildServer) -> Step {
                 _ => vec![]
             },
             vec![
-                build_server_step("oak_loader", &opt.server_rust_toolchain, opt.server_rust_target.as_deref().unwrap_or(DEFAULT_SERVER_RUST_TARGET), match opt.server_variant {
-                    ServerVariant::Base => "",
-                    ServerVariant::NoIntrospectionClient => "oak-unsafe",
-                    ServerVariant::Unsafe => "oak-unsafe,oak-introspection-client",
-                    // If building in coverage mode, use the default target from the host, and build
-                    // in unsafe (debug) mode.
-                    ServerVariant::Coverage => "oak-unsafe,oak-introspection-client",
-                    ServerVariant::Experimental => "oak-attestation,awskms,gcpkms,oak-unsafe,oak-introspection-client",
-                }, &if opt.server_variant == ServerVariant::Coverage {
+                build_rust_binary("oak_loader", opt, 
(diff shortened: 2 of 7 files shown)


oak_functions/loader/src/server.rs
```diff
@@ -58,7 +58,7 @@ impl WasmHandler {
         // TODO(#1919): Get the actual response from the Wasm module via ABI functions.
         http::response::Builder::new()
             .status(StatusCode::BAD_REQUEST)
-            .body(Body::from("Not implemented yet.\n"))
+            .body(Body::from("Welcome to Oak Functions!\n"))
```
tiziano88, line 61: :tada:

runner/src/main.rs
```diff
@@ -348,6 +349,55 @@ fn build_server(opt: &BuildServer) -> Step {
...
+                command: Cmd::new_with_env(
+                    "cargo",
+                    spread![
+                        ...match &opt.server_rust_toolchain {
+                            // This overrides the toolchain used by `rustup` to invoke the actual
+                            // `cargo` binary.
+                            // See https://github.com/rust-lang/rustup#toolchain-override-shorthand
+                            Some(server_rust_toolchain) => vec![format!("+{}", server_rust_toolchain)],
+                            None => vec![],
+                        },
```
tiziano88, line 374: Could this cmd be shared with the one for building the other server? perhaps in a method parametrized with the cargo manifest file and the features to enable?

tiziano88 (approved):

runner/src/main.rs
rbehjati: I think this can be done easily if I use `ServerVariants` for `build_functions_server` too, instead of using `FunctionsServerVariant`. I am currently using `FunctionsServerVariant` because not all `ServerVariants` are relevant for oak-functions loader. What do you think about using `ServerVariants` for `build_functions_server` too?

tiziano88: I think the structure as you implemented it makes more sense than using the variant. I was just talking about sharing part of this logic so that if we add more flags to build rust binaries in general, they will apply to both.

rbehjati: I tried to refactor this into a function.

runner/src/main.rs
```diff
@@ -290,64 +291,107 @@ fn build_server(opt: &BuildServer) -> Step {
...
+                ),
+            }],
+        ]
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>(),
+    }
+}
+
+fn build_server_step(
```
tiziano88, line 359: This is basically a `build_rust_binary` function, right?

runner/src/main.rs
```diff
@@ -290,64 +291,107 @@ fn build_server(opt: &BuildServer) -> Step {
...
+                    None => vec![],
+                },
+                "build".to_string(),
+                format!("--manifest-path={}/Cargo.toml", manifest_dir),
+                format!("--out-dir={}/bin", manifest_dir),
+                // `--out-dir` is unstable and requires `-Zunstable-options`.
+                "-Zunstable-options".to_string(),
+                ...if !features.is_empty() {
+                    vec![format!("--features={}", features),
+                        format!("--target={}", server_rust_target), "--release".to_string(),
```
tiziano88, line 385: Why is this repeated twice in both branches of this if? It seems that --target and --release are always there, so I don't see why they need to be conditional on anything.

tiziano88 (approved):

runner/src/main.rs
rbehjati: Done.

runner/src/main.rs
rbehjati: Fixed.

Issue #1934, opened 2021-03-29
Scripts to deploy an Oak Function on Google Cloud Run


tiziano88:
Similar to https://github.com/project-oak/oak/blob/main/scripts/deploy_example_gcp , but specific for Oak Functions.

PR #1945, opened 2021-04-06
Split base and example images for Oak Functions

tiziano88:
Ref #1934

# Checklist

      construction.

### DIFF
--- a/oak_functions/Dockerfile
+++ b/oak_functions/Dockerfile
@@ -1,12 +1,8 @@
 FROM gcr.io/distroless/static@sha256:d2b0ec3141031720cf5eedef3493b8e129bc91935a43b50562fbe5429878d96b
 
 ARG oak_functions_loader_path=/oak_functions_loader
-ARG wasm_module_path=/module.wasm
 
 COPY ./loader/target/x86_64-unknown-linux-musl/release/oak_functions_loader ${oak_functions_loader_path}
-
-# TODO(#1934): Separate the base loader image from that of individual examples.
-COPY ./loader/testdata/non-oak-minimal.wasm ${wasm_module_path}
 
 # We cannot use ARG values as part of ENTRPOINT, so we have to repeat them here.
 # https://docs.docker.com/engine/reference/builder/#environment-replacement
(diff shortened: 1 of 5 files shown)


oak_functions/loader/testdata/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/oak-ci/oak-functions@sha256:98fe36d3538ebff72d809ef6f5d1c38e9b8835569de45c1d1540ba65486a6c0f
```
rbehjati, line 1: I think it is better to have a separate `oak_functions/examples` folder with examples in it. `testdata` should be only for testing really. But we can fix this later. Perhaps add a TODO somewhere?

oak_functions/Dockerfile
```diff
@@ -1,17 +1,15 @@
...
-
 # We cannot use ARG values as part of ENTRPOINT, so we have to repeat them here.
 # https://docs.docker.com/engine/reference/builder/#environment-replacement
 #
 # Normally we would use the $PORT env variable here, but since we do not have a shell available in
 # the Docker image, we hardcode its default value (8080) here.
 # https://cloud.google.com/run/docs/reference/container-contract#port
 ENTRYPOINT ["/oak_functions_loader", "--http-listen-port=8080", "--wasm-path=/module.wasm"]
+
+# Individual applications must copy their own Wasm module to /module.wasm.
```
rbehjati, line 15: I think some of these comments could go to a README file, but maybe later.

rbehjati (approved):
LGTM!

oak_functions/loader/testdata/Dockerfile
tiziano88: I hope as soon as we have another example (perhaps the weather one) we can move things around and it will make more sense. Basically we can have a docker image for the example and remove this one.

oak_functions/Dockerfile
tiziano88: I agree about the README, though this particular comment probably belongs here in the image itself even after we have the README.

PR #1946, opened 2021-04-07
Add HTTPS Attestation Service

ipetr0v:
This change adds an HTTPS Attestation Service example.

Ref https://github.com/project-oak/oak/pull/1929


examples/https_attestation/backend/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+anyhow = "*"
+http = "*"
+hyper = "*"
+env_logger = "*"
+futures = "*"
+futures-core = "*"
+futures-util = "*"
+log = "*"
+structopt = "*"
+# Pinned to 0.2 because of tonic: https://github.com/hyperium/tonic/blob/master/tonic/Cargo.toml
```
conradgrobler, line 18: This should no longer be needed, and might even cause some issues, after the upgrade of Tonic.

examples/https_attestation/client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+log = "*"
+oak_abi = "=0.1.0"
+oak_attestation_common = "*"
+oak_client = { version = "=0.1.0", features = ["oak-attestation"] }
+prost = "*"
+rustls = "*"
+serde_json = "*"
+structopt = "*"
+# Pinned to 0.2 because of tonic: https://github.com/hyperium/tonic/blob/master/tonic/Cargo.toml
+tokio = { version = "0.2", features = ["fs", "macros", "sync", "stream"] }
```
conradgrobler, line 26: As mentioned earlier, this should no longer be pinned.

examples/https_attestation/client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+tonic = { version = "*", features = ["tls"] }
+regex = { version = "*" }
+tokio-rustls = { version = "*", features = ["dangerous_configuration"] }
```
conradgrobler, line 29: Perhaps worth a comment on why we need to enable this feature?

experimental/attestation_common/src/keying_material.rs
```diff
@@ -0,0 +1,146 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::report::{AttestationInfo, Report};
+use anyhow::{anyhow, Context};
+use serde::{Deserialize, Serialize};
+
+const KEYING_MATERIAL_PURPOSE: &str = "TLSAttestationV1";
+const KEYING_MATERIAL_CLIENT_LABEL: &str =
```
conradgrobler, line 22: Doesn't this influence the keying material? If the client and server use different labels I think they will generate different secrets, which means they do not agree on a shared secret. Or am I missing the point of how this will be used?

examples/https_attestation/backend/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+tokio = { version = "0.2", features = [
+  "fs",
+  "macros",
+  "signal",
+  "sync",
+  "stream"
```
conradgrobler, line 24: This feature no longer exists in the latest version of tokio. The associated functionality has been moved to a separate tokio_stream create.

experimental/attestation_common/src/keying_material.rs
ipetr0v: Correct, labels should be known by both participants in advance and even be registered with IANA.
In this implementation both client and server will export Keying Material with both labels: one to send, and one to check in the response.

examples/https_attestation/client/Cargo.toml
ipetr0v: Done

examples/https_attestation/backend/Cargo.toml
ipetr0v: Done

examples/https_attestation/backend/Cargo.toml
ipetr0v: Done

examples/https_attestation/client/Cargo.toml
ipetr0v: Done

examples/https_attestation/backend/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct Opt {
+    #[structopt(
+        long,
+        help = "Address to listen on for the HTTP server.",
+        default_value = "[::]:8081"
+    )]
+    http_listen_address: String,
+}
+
+// Example message expected from the Oak application.
```
tiziano88, line 41: Not sure which Oak application, or why it is "expected". Isn't think just the message that this server returns to its client?

experimental/attestation_common/src/keying_material.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Clone, Serialize, Deserialize, PartialEq)]
+pub struct KeyingMaterial {
+    /// Value used to disambiguate the proper use of the `KeyingMaterial`
+    purpose: String,
+    /// Pseudo-random value that is uniquely exported for an individual TLS session.
+    value: Vec<u8>,
+}
+
+impl KeyingMaterial {
+    /// Export `KeyingMaterial` for a given TLS `session`.
```
tiziano88, line 91:
```suggestion
    /// Exports `KeyingMaterial` for a given TLS `session`.
```

experimental/attestation_common/src/keying_material.rs
```diff
@@ -0,0 +1,146 @@
...
+        session
+            .export_keying_material(
+                &mut buffer,
+                label.as_bytes(),
+                Some(assertion_type.as_bytes()),
+            )
+            .context("Couldn't export keying material")?;
+        use log::info;
+        info!(
+            "Server Keying material: {:?} (len: {})",
```
tiziano88, line 113: Isn't this function shared between client and server?

experimental/attestation_common/src/keying_material.rs
```diff
@@ -0,0 +1,146 @@
...
+            buffer,
+            buffer.len()
+        );
+        Ok(Self {
+            purpose: KEYING_MATERIAL_PURPOSE.to_string(),
```
tiziano88, line 118: What's the point of this? Could you add a comment? And BTW, shouldn't we use a purpose string as prefix to the context value / assertion type?

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,239 @@
...
+use tokio::net::TcpListener;
+use tokio_rustls::TlsAcceptor;
+
+#[derive(Clone)]
+enum ConnectionStatus {
+    Attested,
+    NotAttested,
+}
+
+/// HTTPS proxy that terminates TLS connections and proxies corresponding HTTP requests to the
```
tiziano88, line 34: Is this proxy run by the client or the server?

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,239 @@
...
+        (&hyper::Method::POST, "/report") => {
+            // Verify client assertion.
+            if let Err(error) = proxy.verify_client_assertion(request).await {
+                warn!("Incorrect client assertion: {:?}", error);
+                return Ok(Response::builder()
+                    .status(hyper::StatusCode::UNAUTHORIZED)
+                    .body(Body::from("Incorrect client assertion"))
+                    .unwrap());
+            }
+            proxy.set_attested_connection_status();
```
tiziano88, line 135: I don't understand what this state is for; is it trying to enforce that the client calls at least once the /report endpoint? Even if it just calls it and throws away the response without reading it? I guess it's good for compatibility, but it's not really a guarantee of anything, is it?

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,239 @@
...
+enum ConnectionStatus {
+    Attested,
+    NotAttested,
+}
+
+/// HTTPS proxy that terminates TLS connections and proxies corresponding HTTP requests to the
+/// `backend_uri`.
+/// It checks client assertions and also generates sends back server assertions in
+/// order to perform remote attestation, if receives an HTTP POST request on `/report`.
+/// A separate `Proxy` instance is created for each individual TLS session.
```
tiziano88, line 38: IIUC this is an HTTP server that intercepts all incoming HTTP request, and intercepts those to the /report path and provides remote attestation to the client, and forwards all other requests to the backend server. Is this correct? If so, could you explain some of this in the comment?

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(Clone)]
+struct Proxy {
+    /// URI of the backend TEE application to proxy HTTP requests to.
+    pub backend_uri: http::Uri,
+    /// `hyper::Client` used to connect to the backend TEE application.
+    pub client: Client<hyper::client::HttpConnector>,
```
tiziano88, line 44: AFAICT there is nothing that enforces that the TEE application is really in any TEE at all; in fact FWIW this proxy may be proxying out to a random server on the internet, right? Could you clarify in the comments? Also I think in practice we would want a unix domain socket on the same machine, rather than a network connection, right?

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Status of the current TLS session.
+    /// `Proxy` doesn't proxy HTTP requests to the backend TEE application, while status is
+    /// `ConnectionStatus::NotAttested`.
+    connection_status: Arc<Mutex<ConnectionStatus>>,
+}
+
+impl Proxy {
+    fn new(
+        backend_uri: http::Uri,
+        client: Client<hyper::client::HttpConnector>,
```
tiziano88, line 58: Is this client pointing to the same `backend_uri`? Or am I missing something?

experimental/https_attestation/src/server.rs
ipetr0v: Updated

conradgrobler (approved):

experimental/https_attestation/src/server.rs
```diff
@@ -0,0 +1,247 @@
...
+                warn!("Client connection error: {:?}", error);
+                std::io::Error::new(std::io::ErrorKind::Other, error)
+            });
+            yield stream.await;
+        }
+    };
+    futures_util::pin_mut!(incoming_tls_stream);
+
+    let client = Client::new();
+    while let Some(tls_stream) = incoming_tls_stream
```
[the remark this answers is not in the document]
rbehjati: Is the content of the `incoming_tls_stream` another stream? Or should it be `Some(tls_message)`?

experimental/attestation_common/src/keying_material.rs
```diff
@@ -0,0 +1,142 @@
...
+    "EXPERIMENTAL Google Confidential Computing Client Attestation 1.0";
+const KEYING_MATERIAL_SERVER_LABEL: &str =
+    "EXPERIMENTAL Google Confidential Computing Server Attestation 1.0";
+const KEYING_MATERIAL_LENGTH: usize = 128;
+// TODO(#2048): Implement multiple assertion types and assertion negotiation.
+const ASSERTION_TYPE: &str = "amd_sev_snp_report";
+
+/// Assertion used for remote attestation.
+#[derive(Serialize, Deserialize)]
+pub struct Assertion {
```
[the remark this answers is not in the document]
ipetr0v: In the future, when we will have multiple assertion types, there will be a global `Assertion` trait.
And this assertion will just become one of them.

experimental/https_attestation/src/main.rs
```diff
@@ -0,0 +1,141 @@
...
+// Load private key from file.
+fn load_private_key(filename: &str) -> anyhow::Result<rustls::PrivateKey> {
+    let private_key_file = std::fs::File::open(filename)
+        .map_err(|error| anyhow!("Couldn't open file {}: {}", filename, error))?;
+    let mut reader = std::io::BufReader::new(private_key_file);
+
+    let private_keys = rustls::internal::pemfile::rsa_private_keys(&mut reader)
+        .map_err(|error| anyhow!("Couldn't parse private key: {:?}", error))?;
+    if private_keys.len() != 1 {
+        return Err(anyhow!(
```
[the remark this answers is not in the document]
rbehjati: I think you don't need an explicit return now:
```rust
        anyhow::bail!(
            "Incorrect number of private keys provided: {}, expected 1",
            private_keys.len()
        )
```

rbehjati (approved):
LGTM! Thanks.

tiziano88 (approved):

PR #1953, opened 2021-04-08
Distributed Runtime

rbehjati:
# Checklist

      construction.


tiziano88:
I would leave the URI idea for later (even though I suggested it), since it complicates reasoning about linearity and liveness of handles, and introduces this concept of local vs remote nodes. 

I think to start with, just using a model similar to what you had in your diagram would be simpler and more elegant; there are no "remote" channels, just channels that terminate in the remote connector pseudo-node, which is labelled as top secret and has the ability to declassify arbitrary data; when it receives a message on one of the channels it manages, it simply sends it out natively over gRPC to the remote party, with an identifier of which channel it came from.

WDYT?

Issue #1935, opened 2021-03-29
Oak Functions Wasm ABI


tiziano88:
Let's use this issue to finalize the desired ABI at the Wasm level.

cc @rbehjati @ipetr0v 

From a hypothetical Rust SDK we would expect to expose the following high-level functions:

```rust
fn read_request() -> Result<Vec<u8>> {}
fn write_response(response: &[u8]) -> Result<()> {}
fn storage_get_item(key: &[u8]) -> Result<Vec<u8>> {}
```

So perhaps we can start from those and work backwards to the Wasm ABI.

We should also consider how much / whether to share with the existing Oak ABI. For instance, `read_request` and `write_response` may just map to the existing `channel_read` and `channel_write`, even if they may have a slightly different semantics, and only bytes (no handles) being moved around.

rbehjati:
I don't see a benefit in sharing with the existing Oak ABI. Since we are not going to use the existing `Runtime` nor the `WasmNode`, it would be more straightforward and more maintainable to have direct ABI counterparts for the SDK functions listed above.

tiziano88:
The benefit would be that the Oak functions ABI and SDK would be a strict subset of the regular Oak ones, so regular Oak could just include them instead of re-implementing them twice. Anyways, the ABI is pretty minimal so I think it's fine to keep them separate if you prefer.

ipetr0v:
Does it mean that the regular Oak Runtime should also implement a read-only data store?

ipetr0v:
> I don't see a benefit in sharing with the existing Oak ABI. Since we are not going to use the existing Runtime nor the WasmNode, it would be more straightforward and more maintainable to have direct ABI counterparts for the SDK functions listed above.

I think we also could try to reuse the same SDK that we use in regular Oak modules (e.g. `Receiver` and `Sender`)

ipetr0v:
Also, it we plan to have a "one-shot" functions, why do we need `fn read_request()` ?
Should we just pass the request to the `main` functions instead?

Also, by the same logic, instead of having a `fn write_response()` function, we can just return data from `main`.

tiziano88:
> Does it mean that the regular Oak Runtime should also implement a read-only data store?

For now the priority is to work on the Functions loader, let's not worry about back-porting functionality to the regular runtime.

tiziano88:
> Also, it we plan to have a "one-shot" functions, why do we need `fn read_request()` ?
> Should we just pass the request to the `main` functions instead?
> 
[...]

PR #1954, opened 2021-04-09
Encapsulate Wasm state for Oak Functions loader

tiziano88:
Make functions more async-friendly.

Make tests actually check responses.

Ensure benchmarks actually exercise the Wasm invocation. Note that now
the time is around 400ms, which seems more realistic than the previous
50ms, I suspect some (or all) of the Wasm execution was being optimized
away earlier.

Ref #1935

# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -32,6 +32,49 @@ pub(crate) struct WasmHandler {
     module: Arc<wasmi::Module>,
 }
 
+/// Encapsulates the state of a Wasm invocation for a single user request.
+struct WasmState {
+    instance: wasmi::ModuleRef,
+    #[allow(dead_code)]
+    memory: wasmi::MemoryRef,
+    #[allow(dead_code)]
+    request_bytes: Vec<u8>,
+    response_bytes: Vec<u8>,
+}
+
+impl WasmState {
+    fn new(module: &wasmi::Module, request_bytes: Vec<u8>) -> anyhow::Result<WasmState> {
+        // TODO(#1919): Make request body available to the Wasm module via ABI functions.
+        // TODO(#1919): Get the actual response from the Wasm module via ABI functions.
+        let instance = wasmi::ModuleInstance::new(module, &wasmi::ImportsBuilder::default())
+            .context("failed to instantiate Wasm module")?
+            .assert_no_start();
+        let memory = instance
+            .export_by_name("memory")
+            .context("could not find Wasm `memory` export")?
+            .as_memory()
+            .cloned()
+            .context("could not interpret Wasm `memory` export as memory")?;
+        Ok(WasmState {
+            instance,
+            memory,
+            request_bytes,
+            response_bytes: "Welcome to Oak Functions\n".as_bytes().to_vec(),
+        })
+    }
+
+    fn invoke(&mut self) {
+        let result = self
+            .instance
+            .invoke_export(MAIN_FUNCTION_NAME, &[], &mut wasmi::NopExternals);
+        info!(
+            "{:?}: Running Wasm module completed with result: {:?}",
+            std::thread::current().id(),
+            result
+        );
+    }
+}
+
 impl WasmHandler {
     pub(crate) fn create(wasm_module_bytes: &[u8]) -> anyhow::Result<Self> {
         let module = wasmi::Module::from_buffer(&wasm_module_bytes)?;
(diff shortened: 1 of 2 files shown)


conradgrobler (approved):

oak_functions/loader/src/server.rs
```diff
@@ -32,6 +32,48 @@ pub(crate) struct WasmHandler {
     module: Arc<wasmi::Module>,
 }
 
+/// Encapsulates the state of a Wasm invocation for a single user request.
+struct WasmState {
+    instance: wasmi::ModuleRef,
+    #[allow(dead_code)]
+    memory: wasmi::MemoryRef,
```
rbehjati, line 39: Why is memory needed as a separate filed in the state? It can already be extracted from `instance`.

oak_functions/loader/src/server.rs
```diff
@@ -40,35 +82,28 @@ impl WasmHandler {
...
-            "{:?}: Running Wasm module completed with result: {:?}",
-            std::thread::current().id(),
-            result
-        );
-
+    async fn handle_invoke(&self, req: Request<Body>) -> anyhow::Result<Response<Body>> {
+        let request_bytes = hyper::body::to_bytes(req.into_body()).await.unwrap();
+        let mut wasm_state = WasmState::new(&self.module, request_bytes.to_vec())?;
+        wasm_state.invoke();
         // TODO(#1919): Get the actual response from the Wasm module via ABI functions.
```
rbehjati, line 103: nit: Move this comment to `WasmState` too.

oak_functions/loader/src/tests.rs
```diff
@@ -68,23 +72,29 @@ fn bench_wasm_handler(bencher: &mut Bencher) {
...
-        bencher.iter(|| wasm_handler.handle_request(&request));
+        bencher.iter(|| {
+            let request = hyper::Request::builder()
+                .method(http::Method::POST)
+                // We don't really send the request. So the hostname can be anything. It is only
+                // needed for building a valid request.
+                .uri("http://example.com/invoke")
+                .body(hyper::Body::empty())
+                .unwrap();
+            let resp = tokio::runtime::Runtime::new()
```
rbehjati, line 83: I think the purpose of this benchmark was to just measure the loading of the wasm module. This is now processing the request and producing the response too. Is this the intended behaviour?

rbehjati (approved):
I only have a few minor comments.

oak_functions/loader/src/server.rs
tiziano88: It is a similar patter to the Oak Runtime. When we implement the ABI on this struct, it will be simpler if the memory is already available as a separate field.

oak_functions/loader/src/tests.rs
tiziano88: We want it to measure something meaningful from the point of view of a user of the Oak Functions Loader, so I think the response processing time is what matters (though we may want more precise smaller benches too for individual things like loading Wasm module at some point? not sure that's super useful at this stage by itself).

Also I don't think the purpose of the bench was to measure the loading of the wasm module, since it was already measuring the `handle_request` method, right? Or, if that was the intention, perhaps the naming was not quite accurate?

oak_functions/loader/src/server.rs
tiziano88: Done.

oak_functions/loader/src/tests.rs
tiziano88: In fact, I don't think it was measuring only the loading of the module, it was definitely also measuring the invocation of the Wasm exported function (since they are both in `handle_request`). I don't think it was even possible to measure Wasm module loading without the actual invocation, or did I misunderstand your comment?

rbehjati: The actual invocation (via `invoke_export`) does not do much in the case of this minimal Wasm module, and the response was hard-coded. So, I think the time that was measured previously was mostly just loading the module (even though this was done via the call to `handle_request`). I agree that if the intention is to measure loading of the module, some refactoring and renaming would be needed. 
But I am fine with this test if actually benchmarking the response processing time is intended.

tiziano88: What I don't understand is why does it take so much longer now? There should be the same amount of processing going on, right? That's why I suspect some part of the execution was just being optimized out earlier for some reason. Or maybe creating the tokio executor also takes longer?

I am also still not 100% sure I understand your point (sorry, just want to make sure I am not overlooking something):

> So, I think the time that was measured previously was mostly just loading the module (even though this was done via the call to handle_request)

Isn't this the same now? AFAICT the same exact Wasm module is loaded **and** invoked, both before and after this PR.

tiziano88: Oops, I just realised I was super wrong!

The overhead is almost entirely due to creating the tokio runtime inside the benchmark iteration! I now moved it out and the bench time went down to 25ms! All my assumptions and conclusions were incorrect :)

rbehjati: Nice. I did not know creating the runtime takes that much time.

Issue #1956, opened 2021-04-09
Hide "Parsed manifest files" output


ipetr0v:
Should we hide (for example into `debug` output) the following output:
https://github.com/project-oak/oak/blob/b6bc1597b8bb9a470702de6477e7b01fbbe33e8b/runner/src/main.rs#L160

Because it ends up in a large wall of text in the console
```
[...]

PR #1957, opened 2021-04-09
Delete 'Parsed manifest files'

ipetr0v:
This change deletes the "Parsed manifest files" output from `runner`.

Fixes https://github.com/project-oak/oak/issues/1956


ipetr0v:
Since we don't use `RUST_LOG` in `runner` deleted it all together.

tiziano88 (approved):

PR #1958, opened 2021-04-12
Add oak_functions ABI and SDK

### LINKED ISSUE #1935: Oak Functions Wasm ABI

rbehjati:
Fixes #1935


oak_functions/abi/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Type, constant and Wasm host function definitions for the Oak-Functions application
+//! binary interface (ABI).
+
+// Status values exchanged as i32 values across the WasmInterface.
+#[derive(Debug)]
+pub enum OakStatus {
```
tiziano88, line 22: Worth marking as `#[repr(u8)]` instead of manually converting to / from numbers?

oak_functions/examples/hello_world/module/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Oak functions hello world example.
+//!
+//! To build and run manually:
+//! 1. Build the wasm module:
```
tiziano88, line 20: In order to be valid markdown, there should be newlines before paragraphs and code blocks, and also the code blocks should be indented if they are part of the bullet point lists.

oak_functions/loader/src/server.rs
```diff
@@ -14,65 +14,207 @@
...
     instance: wasmi::ModuleRef,
-    #[allow(dead_code)]
-    memory: wasmi::MemoryRef,
-    #[allow(dead_code)]
-    request_bytes: Vec<u8>,
-    response_bytes: Vec<u8>,
+    abi: WasmInterface,
 }
 
 impl WasmState {
```
tiziano88, line 177: FYI my intention behind this WasmState struct was to implement ModuleImportResolver and Externals directly on it.

oak_functions/abi/src/lib.rs
rbehjati: I am not sure how exactly this helps. I still need to implement `from_i32`. But an automatic conversion would be nice. I noticed, I had already missed some values.

oak_functions/examples/hello_world/module/src/lib.rs
rbehjati: Thanks. I added newlines and indentation. Does it look like a valid markdown now?

oak_functions/loader/src/server.rs
rbehjati: I did think of it, but when creating a `wasmi::ModuleInstance` we have to pass in a resolver). Since this is happening in the `WasmState::new` function, there is no `self` to pass as the resolver, if `ModuleImportResolver` and `Externals` were to be implemented for `WasmState`.

tiziano88: Would something like this work? (I guess probably not, since the references would overlap, but maybe there is a way to make it work).

```rust
let wasm_state = WasmState { ... };
let instance = ModuleInstance::new(module, ...&wasm_state);
let wasm_state.instance = instance;
Ok(wasm_state)
```

If not anyways, then I think we can probably remove the `WasmState` wrapper, since (in my mind) that was its only purpose.

oak_functions/abi/src/lib.rs
tiziano88: You are right, I think that only allows converting from enum to number, but not the other way around. There are some crates that do the conversion both ways, but not sure it's worth introducing them for now. Also perhaps after https://github.com/project-oak/oak/pull/1959 is merged we can actually use a protobuf enum for this, like for the regular Oak ABI?

oak_functions/examples/hello_world/module/src/lib.rs
tiziano88: I think it's still missing the newlines before each code block, and also the code blocks themselves need to be indented (not the code inside the blocks). Does it render correctly currently? BTW I think this would be more maintainable and readable as an actual .md file in the directory, rather than here, since we don't tend to use the generated Rust documentation much, WDYT?

oak_functions/examples/hello_world/module/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+use oak_functions as sdk;
+
+#[cfg(test)]
+mod tests;
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn main() {
+    // A panic in the Rust module code cannot safely pass through the FFI
+    // boundary, so catch any panics here and drop them.
+    // https://doc.rust-lang.org/nomicon/ffi.html#ffi-and-panics
```
tiziano88, line 56: I'm actually not 100% sure this is correct, we should double check at some point. I mean, it is correct for real FFI (e.g. C / Rust), but I doubt that Wasm is considered like an FFI.

oak_functions/examples/hello_world/module/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        // Read the request
+        let request_body = sdk::read_request_body().expect("Couldn't read request body.");
+
+        // Write the response
+        let response_body = format!(
+            "Welcome to Oak functions!\nRequest length: {}.\n",
+            request_body.len()
+        );
+
+        let _res = sdk::write_response_body(&response_body);
```
tiziano88, line 69: Why not `expect` here too?

oak_functions/examples/hello_world/module/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        sdk::set_panic_hook();
+
+        // Read the request
+        let request_body = sdk::read_request_body().expect("Couldn't read request body.");
+
+        // Write the response
+        let response_body = format!(
+            "Welcome to Oak functions!\nRequest length: {}.\n",
+            request_body.len()
+        );
```
tiziano88, line 67: Maybe make this a classic hello world (since that's the name of the example), get the name as input and print `Hello $name!` to output?

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+// limitations under the License.
+//
+
+//! SDK functionality that provides idiomatic Rust wrappers around the
+//! underlying Oak functions platform functionality.
+
+use log::{debug, error};
+use oak_functions_abi::OakStatus;
+
+// Read and return the user request.
```
tiziano88, line 23:
```suggestion
/// Reads and returns the user request.
```

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+pub fn read_request_body() -> Result<Vec<u8>, OakStatus> {
+    let mut buf = Vec::with_capacity(1024);
+    read_request(&mut buf)?;
+    Ok(buf)
+}
+
+/// Read the user request into the buffer.
```
tiziano88, line 30: Mention that it reallocates the buffer if needed?

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+fn read_request(buf: &mut Vec<u8>) -> Result<(), OakStatus> {
+    // Try reading the request twice: first with provided vectors, making
```
tiziano88, line 32:
```suggestion
    // Try reading the request twice: first with provided vector, making
```

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+fn write_response(buf: &[u8]) -> Result<(), OakStatus> {
+    let status = unsafe { oak_functions_abi::write_response(buf.as_ptr(), buf.len()) };
+    result_from_status(status as i32, ())
+}
+
+/// Convert a status returned from a host function call to a `Result`.
+///
+/// The status is interpreted as an int representing an `OakStatus` enum value.
+///
+/// If the status is `OK` then return the provided value as `Result::Ok`, otherwise return the
```
tiziano88, line 94:
```suggestion
/// If the status is [`OakStatus::Ok`] then returns the provided value as [`Result::Ok`], otherwise returns the
```

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+//
+
+//! Test utilities to help with unit testing of Oak-Functions SDK code.
+
+use anyhow::Context;
+use log::info;
+
+use std::{path::PathBuf, process::Command};
+
+/// Uses cargo to compile a Rust manifest to Wasm bytes.
```
tiziano88, line 24: Is there any specific oak functions logic here? AFAICT it is identical to the existing one?

tiziano88 (commented):
Thanks, looking great!

rbehjati (commented):
Thanks for the review @tiziano88. I'll create a few issues to track the remaining pieces, and I can add TODO comments for them.
 
I think we need the following: 
- An example similar to abitest to properly test the ABI functions
- Commands in the `runner` for running `oak_functions` examples in the CI 
- A new benchmark test for the `hello_world` example that actually uses the ABI functions
- We probably want to read and write the entire request and response, with the headers too not just the body. 

WDYT?

oak_functions/abi/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    ///
+    /// Returns the status of the operation, as an [`OakStatus`] value.
+    ///
+    /// [`ErrBufferTooSmall`]: crate::OakStatus::ErrBufferTooSmall
+    /// [`OakStatus`]: crate::OakStatus
+    pub fn read_request(buf: *mut u8, size: usize, actual_size: *mut u32) -> u32;
+
+    /// Write the response.
+    ///
+    /// Stores `size` bytes of data from `buf` into the WebAssembly interface. This can then be
```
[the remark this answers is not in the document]
tiziano88: Yes I know, I just think the comment should say that explicitly :)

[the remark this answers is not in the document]
rbehjati: Ah! Ok :D
So no abitest needed? ;)

oak_functions/abi/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// The Oak-Functions ABI primarily consists of a collection of Wasm host functions in the
+// "oak_functions" module that are made available to WebAssembly modules running as Oak-Functions
+// workloads.
+// See https://rustwasm.github.io/book/reference/js-ffi.html
+#[link(wasm_import_module = "oak_functions")]
+extern "C" {
+    /// Read the user request.
+    ///
```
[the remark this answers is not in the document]
tiziano88: I also was just looking for a comment to explain this :) I think the answer is just that if invoked multiple times it returns the same request every time.

oak_functions/abi/src/lib.rs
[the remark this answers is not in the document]
tiziano88: Up to you, I think it may still make for a nice test case for the ABI, it can come later anyways

oak_functions/examples/hello_world/module/src/tests.rs
```diff
@@ -0,0 +1,66 @@
...
+use oak_functions_loader::server::create_and_start_server;
+
+const WASM_MODULE_NAME: &str = "hello_world.wasm";
+const EXPECTED_RESPONSE: &str = "Welcome to Oak functions!\nRequest length: 0.\n";
+
+#[tokio::test]
+async fn test_server() {
+    let _ = env_logger::builder().is_test(true).try_init();
+    let port = 9999;
+    let address = format!("[::]:{}", port);
```
tiziano88, line 26: See https://github.com/project-oak/oak/pull/1960/files for a nicer way of expressing special addresses (like `[::]` and `[::1]`).

oak_functions/examples/hello_world/module/src/tests.rs
```diff
@@ -0,0 +1,66 @@
...
+use hyper::client::Client;
+use oak_functions_loader::server::create_and_start_server;
+
+const WASM_MODULE_NAME: &str = "hello_world.wasm";
+const EXPECTED_RESPONSE: &str = "Welcome to Oak functions!\nRequest length: 0.\n";
+
+#[tokio::test]
+async fn test_server() {
+    let _ = env_logger::builder().is_test(true).try_init();
+    let port = 9999;
```
tiziano88, line 25: Declare as const? See https://github.com/project-oak/oak/pull/1960/files

oak_functions/examples/hello_world/module/src/tests.rs
```diff
@@ -0,0 +1,66 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use hyper::client::Client;
+use oak_functions_loader::server::create_and_start_server;
+
+const WASM_MODULE_NAME: &str = "hello_world.wasm";
+const EXPECTED_RESPONSE: &str = "Welcome to Oak functions!\nRequest length: 0.\n";
```
tiziano88, line 20: I think this needs to be updated, right?

oak_functions/examples/hello_world/module/src/tests.rs
```diff
@@ -0,0 +1,66 @@
...
+    let (res, _) = tokio::join!(server_fut, client_fut);
+    assert!(res.is_ok());
+}
+
+async fn start_client(port: u16, notify_sender: tokio::sync::oneshot::Sender<()>) {
+    let client = Client::new();
+    let request = hyper::Request::builder()
+        .method(http::Method::POST)
+        .uri(format!("http://localhost:{}/invoke", port))
+        .body(hyper::Body::empty())
```
tiziano88, line 50: Update?

oak_functions/loader/Cargo.toml
```diff
@@ -7,6 +7,14 @@ authors = ["Razieh Behjati <razieh@google.com>"]
 edition = "2018"
 license = "Apache-2.0"
 
+[lib]
+name = "oak_functions_loader"
+path = "src/lib.rs"
```
tiziano88, line 12: Do these paths need to be specified? AFAICT they are the default values: https://doc.rust-lang.org/cargo/guide/project-layout.html

oak_functions/loader/src/server.rs
```diff
@@ -149,3 +291,63 @@ pub async fn create_and_start_server(
...
+    // oak_functions/abi/src/lib.rs
+    let (index, sig) = match field_name {
+        "read_request" => (
+            READ_REQUEST,
+            wasmi::Signature::new(
+                &[
+                    ABI_USIZE, // dest
+                    ABI_USIZE, // dest_capacity
+                    ABI_USIZE, // actual_length_addr (out)
+                ][..],
```
tiziano88, line 310: I'm not entirely sure what the `[..]` does, what is the error without it?

oak_functions/loader/src/server.rs
```diff
@@ -14,65 +14,207 @@
...
             .as_memory()
-            .cloned()
-            .context("could not interpret Wasm `memory` export as memory")?;
-        Ok(WasmState {
-            instance,
-            memory,
-            request_bytes,
-            response_bytes: "Welcome to Oak Functions\n".as_bytes().to_vec(),
-        })
+            .cloned();
```
tiziano88, line 191: I think this should still try to unwrap / context the memory, and if it fails it should return early. It does not matter that it returns an Option and you need an Option, really here you should be setting the Option to Some(memory), and if the memory is not there it should stop as soon as possible, not fail later on with a confusing error message.

oak_functions/loader/src/server.rs
```diff
@@ -14,65 +14,207 @@
...
+const ABI_USIZE: ValueType = ValueType::I32;
+
+/// `WasmInterface` holds runtime values for a particular execution instance of Wasm, handling a
+/// single user request. The methods here correspond to the ABI host functions that allow the Wasm
+/// module to exchange the request and the response with the Oak functions server. These functions
+/// translate values between Wasm linear memory and Rust types.
+struct WasmInterface {
+    request_bytes: Vec<u8>,
+    response_bytes: Vec<u8>,
+    memory: Option<wasmi::MemoryRef>,
```
tiziano88, line 50: This Optional field and the two-step construction process does not seem very elegant, I hope we can make things more streamlined in the future :)

oak_functions/sdk/oak_functions/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "oak_functions"
+version = "0.1.0"
+authors = ["Razieh Behjati <razieh@google.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dependencies]
+oak_functions_abi = "0.1.0"
```
tiziano88, line 10: Please use relative paths when referring to local crates. I know there are overrides somewhere in the manifest files, but those tend to break easily, especially if someone at some point registers an actual crate with this name on crates.io.

oak_functions/sdk/oak_functions/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "oak_functions"
+version = "0.1.0"
+authors = ["Razieh Behjati <razieh@google.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
tiziano88, line 8: Remove.

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,141 @@
+//
+// Copyright 2019 The Project Oak Authors
```
tiziano88, line 2:
```suggestion
// Copyright 2021 The Project Oak Authors
```

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+    error!("unreachable code reached");
+    Err(OakStatus::ErrInternal)
+}
+
+/// Write the response body.
+///
+/// Multiple calls to this function will replace the earlier responses. Only the last response that
+/// is written will be kept and returned to the user.
+pub fn write_response_body(response_body: &str) -> Result<(), OakStatus> {
```
tiziano88, line 87: I don't think we should have two versions, `&[u8]` should be sufficient, in general the body is not a valid UTF8 string.

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: Yes I think it would be good, I'm sure I would definitely forget to update one or the other if I made any changes :)

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+/// system, so these values would usually be converted (via a cast) to `i32` by callers.
+pub fn result_from_status<T>(status: i32, val: T) -> Result<T, OakStatus> {
+    match OakStatus::from_i32(status) {
+        Some(OakStatus::Ok) => Ok(val),
+        Some(status) => Err(status),
+        None => Err(OakStatus::Unspecified),
+    }
+}
+
+/// Install a panic hook that logs [panic information].
```
tiziano88, line 113: Where does it log it? AFAICT there is no logging at all from functions, or are you suggesting we should add it?

tiziano88 (approved):
> Thanks for the review @tiziano88. I'll create a few issues to track the remaining pieces, and I can add TODO comments for them.
> 
> I think we need the following:
> 
> * An example similar to abitest to properly test the ABI functions

:heavy_check_mark: 

> * Commands in the `runner` for running `oak_functions` examples in the CI

:heavy_check_mark: 

> * A new benchmark test for the `hello_world` example that actually uses the ABI functions

:heavy_check_mark: (it may just replace the other benchmark, I don't think we need both)

> * We probably want to read and write the entire request and response, with the headers too not just the body.

I don't think so, we should intentionally only pass the body of the request, and provide back the body of the response. The function body does not even know whether these are HTTP or gRPC requests or someone wrote something on a piece of paper that is encrypted with a key that is bound to the TEE. In fact the latter is somewhat similar to what I hope we will get to eventually, in which most of the network stack is outside of the TEE, only some encryption / decryption happens inside, possibly with a simple transport mechanism.

> * Refactor common non-oak specific functionality into a separate crate that both `oak` and `oak_functions` crates can use.

:heavy_check_mark: 

> 
> WDYT?

oak_functions/examples/hello_world/README.md
```diff
@@ -0,0 +1,38 @@
...
+       --target=wasm32-unknown-unknown \
+       --manifest-path=oak_functions/examples/hello_world/module/Cargo.toml \
+       --out-dir=oak_functions/examples/hello_world/bin
+   ```
+
+1. Build loader with debugging enabled if needed:
+
+   ```shell
+   cargo build --manifest-path=./oak_functions/loader/Cargo.toml \
+       --release
```
conradgrobler, line 23: Missing a `\`?

conradgrobler (approved):

oak_functions/examples/hello_world/README.md
rbehjati: Fixed.

rbehjati (commented):
Thanks for the reviews :)

oak_functions/sdk/oak_functions/Cargo.toml
tiziano88: I think I've been specifying it explicitly, since I got into issues otherwise. We can do another pass at some point, but definitely this is not the only place in which it would be done.

oak_functions/loader/src/server.rs
```diff
@@ -14,65 +14,212 @@
...
 const MAIN_FUNCTION_NAME: &str = "main";
 
-// An ephemeral request handler with a Wasm module for handling the requests.
-#[derive(Clone)]
-pub(crate) struct WasmHandler {
-    // Wasm module to be served on each invocation. `Arc` is needed to make `WasmHandler`
-    // cloneable.
-    module: Arc<wasmi::Module>,
+/// Wasm host function index numbers for `wasmi` to map import names with. This numbering is not
+/// exposed to the Wasm client.  See https://docs.rs/wasmi/0.6.2/wasmi/trait.Externals.html
```
ipetr0v, line 31: very nit. =)
Double space before `See`

oak_functions/loader/src/server.rs
```diff
@@ -149,3 +296,63 @@ pub async fn create_and_start_server(
     );
     Ok(())
 }
+
+/// A resolver function, mapping Oak host function names to an index and a type signature.
+fn oak_resolve_func(
```
ipetr0v, line 301: Why does this function have an `oak_` prefix?
Should it just be a `resolve_function` ?

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+//! SDK functionality that provides idiomatic Rust wrappers around the
+//! underlying Oak functions platform functionality.
+
+use log::{debug, error};
+use oak_functions_abi::OakStatus;
+
+/// Reads and returns the user request.
+///
+/// This function is idempotent. Multiple call to this function all return the same value.
+pub fn read_request_body() -> Result<Vec<u8>, OakStatus> {
```
ipetr0v, line 26: Should this function be renamed to `read_request` (as comment calls it)
And `read_request` to `read_request_util` ?

ipetr0v (approved):

oak_functions/abi/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+//
+
+//! Type, constant and Wasm host function definitions for the Oak-Functions application
+//! binary interface (ABI).
+
+// Status values exchanged as i32 values across the WasmInterface.
+#[derive(Debug)]
+pub enum OakStatus {
+    Unspecified = 0,
+    // Success.
```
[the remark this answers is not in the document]
rbehjati: Done.

oak_functions/loader/src/server.rs
rbehjati: I think it is good to have different names. Now to be more accurate, I changed it to `oak_functions_resolve_func`, since the module name is `oak_functions` :D

oak_functions/sdk/oak_functions/src/lib.rs
rbehjati: Yes. Done.

oak_functions/sdk/oak_functions/Cargo.toml
rbehjati: Done.

Issue #1930, opened 2021-03-25
Side loading data for lookup in Oak Functions Loader


tiziano88:
Initially we will implement the Oak Functions Loader so that it loads all the data for lookup at startup from a file, and that stays immutable for the lifetime of the loader itself.

In the not-so-distant future though we need to figure a way to have the data periodically refreshed. I think there are two 
fundamental approaches to consider:

## push

An external service periodically pushes a new version of the dataset (over HTTP or gRPC) to the loader.

An advantage of this approach is that it can only update a subset of the dataset if desired.

## pull

The loader periodically calls another server (over HTTP or gRPC) and downloads a new version of the entire dataset, which replaces the previous one.

An advantage of this approach is that there is no need for the loader to authenticate incoming requests to update the data.

---

Are there other approaches to consider? Which one would we want to go with? I think so far the pull approach seems simpler to me, especially because it does not need authentication and authorization for the updates.

@rbehjati @ipetr0v

rbehjati:
I also think that pull is simpler, but perhaps this is a decision that should be made for each application depending on the size of the data and frequency of the changes. I think we could start with pull and then add support for push later if needed.

tiziano88:
SG, what format should we use for the key / value list of entries? In Google we would normally use an [sstable](https://www.quora.com/What-is-an-SSTable-in-Googles-internal-infrastructure) for this, not sure what the open source equivalent is.

tiziano88:
FTR, we discussed that a viable approach could be to pre-process the data into the format of choice, and then serve it statically (e.g. on [Google Cloud Storage](https://cloud.google.com/storage)), and the Oak Functions Loader would periodically retrieve it from there behind a fixed URL. This decouples the format and preprocessing of the source data to a separate pipeline / process that does not need to be embedded in the Oak Functions Loader itself.

tiziano88:
Some potentially relevant references:

- https://cloud.google.com/datastore/docs/export-import-entities
- https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases
- https://github.com/google/leveldb/blob/master/doc/log_format.md

tiziano88:
I think we should just define our own protobuf format for a key / value holder, and use https://docs.rs/prost/0.7.0/prost/fn.decode_length_delimiter.html to iteratively read subsequent values from a buffer that contains such records back-to-back.

The format can be as simple as

```proto
message Entry {
  bytes key = 1;
  bytes value = 2;
}
```

[...]

PR #1959, opened 2021-04-12
Allow loading lookup data in Oak Functions loader

tiziano88:
Not actually used in the server implementation yet.

Ref #1930

# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/src/main.rs
+++ b/oak_functions/loader/src/main.rs
@@ -18,7 +18,8 @@
 #![feature(test)]
 
 use anyhow::Context;
-use std::fs;
+use prost::Message;
+use std::{collections::HashMap, fs, sync::RwLock};
 use structopt::StructOpt;
 
 mod server;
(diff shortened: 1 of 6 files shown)


oak_functions/loader/src/main.rs
```diff
@@ -43,6 +48,87 @@ pub struct Opt {
...
+            .expect("could not lock entries for read")
+            .get(key)
+            .cloned()
+    }
+}
+
+fn load_lookup_entries<B: bytes::Buf>(
+    lookup_data_buffer: B,
+) -> anyhow::Result<HashMap<Vec<u8>, Vec<u8>>> {
+    let mut lookup_data_buffer = lookup_data_buffer;
```
conradgrobler, line 124: Why is this needed?

conradgrobler (approved):

rbehjati (approved):
LGTM!

oak_functions/loader/src/main.rs
tiziano88: It's just a rebinding to allow for mutable methods to be invoked on it. The variable is passed by value already, so this is just a detail of the implementation of this method, it's not making things "more" mutable, since the value is already owned by this function.

PR #1968, opened 2021-04-14
Implement custom logger

### LINKED ISSUE #1955: Selective logging for Oak Functions Loader
Currently we have a way of enabling or disabling logging completely via a feature:

https://github.com/project-oak/oak/blob/46cf7e38d998c60ae4c1b1d65e929951deefa2c6/oak_functions/loader/src/main.rs#L51-L55

This is not great because when disabled, there is no way to log anything at all, even if we would consider it safe (e.g. incoming request size, etc.).

We should think of some better way of handling this, in a way that does not allow third party dependencies to accidentally log sensitive data along the way (e.g. by setting `RUST_LOG=trace`).

conradgrobler:
This change uses features to switch off all logging via the `log` crate and adds an explicit logger that differentiates between unsafe an public logging.

Fixes #1955

# Checklist

      construction.


oak_functions/loader/src/logger.rs
```diff
@@ -0,0 +1,71 @@
...
+/// Writing of potentially sensitive content will be ignored unless the "oak-unsafe" feature is
+/// enabled.
+pub(crate) struct Logger {
+    max_level: LevelFilter,
+}
+
+impl Logger {
+    /// Creates a new logger with the specified maximum `LevelFilter`.
+    #[allow(dead_code)]
+    pub(crate) fn new(max_level: LevelFilter) -> Self {
```
tiziano88, line 32: Should we just stick to `pub` instead of `pub(crate)` throughout? I don't think it is providing any meaningful advantage, and if we ever decide to use this for something else we would have to change them to `pub` anyways.

oak_functions/loader/src/logger.rs
```diff
@@ -0,0 +1,71 @@
...
+            self.log(level, message);
+        }
+    }
+
+    /// Logs a message that contains only public, non-sensitive content at the specified `Level`.
+    ///
+    /// All code that uses this function should be inspected to ensure that the message can never
+    /// contain any sensitive or non-public information.
+    #[allow(dead_code)]
+    pub(crate) fn log_public(self: &Self, level: Level, message: &str) {
```
tiziano88, line 51: Not very actionable, but it seems a bit odd that, if we were annotating these functions with `unsafe`, actually `log_public` would be `unsafe` and `log_unsafe` would not be. I think it's fine as it is, but it makes me think whether we should rethink the terminology or the API.

rbehjati: Perhaps rename `log_unsafe` to `log_sensitive`?

oak_functions/loader/src/logger.rs
conradgrobler: Done.

oak_functions/loader/src/logger.rs
conradgrobler: Done. Also renamed feature to "oak-insecure" seeing that "unsafe" has a specific meaning in Rust.

oak_functions/loader/src/logger.rs
```diff
@@ -0,0 +1,73 @@
...
+                level,
+                message,
+            );
+        }
+    }
+}
+
+impl Default for Logger {
+    fn default() -> Self {
+        Logger::new(LevelFilter::Info)
```
tiziano88, line 71: Since this also gets used in tests, I would make it log at least at Debug, or have a separate constructor `for_test` or similar that actually does that (and can be exposed just for testing).

oak_functions/loader/src/logger.rs
```diff
@@ -0,0 +1,73 @@
...
+use chrono::{SecondsFormat, Utc};
+use log::{Level, LevelFilter};
+use std::default::Default;
+
+/// A simple logger that splits logging between writing logs that contain only public, non-sensitive
+/// content and writing logs that could potentially contain sensitive content.
+///
+/// Writing of potentially sensitive content will be ignored unless the "oak-insecure" feature is
+/// enabled.
+#[derive(Clone)]
```
tiziano88, line 25: Why is it cloned? I would expect that only references to the logger get passed around, or is that too complicated given the lifetime requirements of the current code?

tiziano88 (approved):

oak_functions/loader/src/logger.rs
conradgrobler: I initially used and `Arc<Logger>` everywhere, but I thought cloning will be more efficient as the struct contains only the log level filter. I am unsure how a single logger's lifetime will interact with the various async functions, but I will try it out and see.

rbehjati:
Does this mean that the logs in the `sdk` are useless now? 
Also, do we plan to use this logger in the examples too?

oak_functions/loader/src/logger.rs
conradgrobler: The lifetime is problematic in `create_and_start_server` with the `WasmHandler` clones in the closures. The compiler does not believe that `graceful_server.await` will outlive all `WasmHandler` instances. I tried using static lifetime annotations in various places, but that just moved the problem elsewhere.

I could use references everywhere except in the WasmHandler where it would use clones, but I am not sure whether it would be any better than just always cloning. Alternatively I could use something like `lazy_static!` to create a static singleton instance, but that feels a lot like moving back towards ambient authority.

conradgrobler:
> Does this mean that the logs in the `sdk` are useless now?

I think the logs in the `sdk` would not have worked anyway, seeing that the `sdk` code is running inside a Wasm sandbox.

> Also, do we plan to use this logger in the examples too?

If we want to log from Wasm code, we would need to provide an ABI function. That in turn could use `log_sensitive`, so that the logging would only work when the `oak-insecure` feature is enabled.

rbehjati:
> If we want to log from Wasm code, we would need to provide an ABI function. That in turn could use `log_sensitive`, so that the logging would only work when the `oak-insecure` feature is enabled.

I think an ABI function would be useful (in a separate PR).

oak_functions/loader/src/logger.rs
conradgrobler: I undid the rename of the feature, as it was referenced in the runner etc. Could be done in future in a separate PR.

conradgrobler:
> I think an ABI function would be useful (in a separate PR).

I added #1977 to track this.

PR #1969, opened 2021-04-14
Use protobuf for OakStatus

rbehjati:
Ref: https://github.com/project-oak/oak/pull/1958#discussion_r612414795


tiziano88:
Where is the proto file? :eyes:

oak_functions/abi/proto/abi.proto
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak_functions_abi;
```
tiziano88, line 19: We already have https://github.com/project-oak/oak/tree/main/oak_functions/proto which

- is in a different folder
- has a different package name
- is built as part of a different crate

It seems we should just unify all these things?

rbehjati: Yes. I was not sure where is the right place for this proto file. The `oak_abi` crate has its own `proto` folder. Why is that one not merged with the top `proto` folder?

tiziano88: I think it's just historical reasons. For Oak Functions, I think we can use either `oak_functions/proto` or `oak_functions/abi/proto`, perhaps the first one is better since it's not strictly speaking just part of the ABI? OTOH you could consider the key value stuff as part of the ABI too in a way... not a big deal anyways, we can easily move them around later.

rbehjati: Moved it to `oak_functions/proto`, since there is no good reason to have a separate `proto` file.

oak_functions/abi/build.rs
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+extern crate prost_build;
+
+fn main() {
+    prost_build::compile_protos(&["../proto/abi.proto"], &["../proto"])
+        .expect("Proto compilation failed");
+}
```
tiziano88, line 22: Now this is built twice as part of the loader though, right? I think we should remove the other build.rs file and just use the compiled protos from this ABI crate everywhere, WDYT?

tiziano88 (approved):

oak_functions/abi/build.rs
rbehjati: I removed the `build.rs` in the `loader`, but was it really being built twice? I thought the one in the `loader` only built `lookup_data.proto`.

tiziano88: You are correct, it was not being built, but my point is that we should build all the protos in one build.rs script, either the ABI one or something else just for protos.

rbehjati: Makes sense. So, now everything is being built in the `abi`.

tiziano88 (approved):
Great!

Issue #1933, opened 2021-03-29
Improve benchmarking for oak functions loader


rbehjati:
We currently have a simple benchmark test for oak_functions loader. The runner should be updated to run the benchmark as a step in the CI builds. Moreover, it is likely that we'll need more advanced benchmarking. 
See also https://github.com/project-oak/oak/pull/1925#discussion_r601739674

Related to #1919.

rbehjati:
Consider using the `hello_world` example in the benchmark.
See also: https://github.com/project-oak/oak/pull/1958#pullrequestreview-634991559

PR #1970, opened 2021-04-14
Use hello_world as benchmark

rbehjati:
Ref #1933

### DIFF
--- a/oak_functions/examples/hello_world/module/src/tests.rs
+++ b/oak_functions/examples/hello_world/module/src/tests.rs
@@ -18,7 +18,7 @@ use oak_functions_loader::server::create_and_start_server;
 use std::net::{Ipv6Addr, SocketAddr};
 
 const WASM_MODULE_NAME: &str = "hello_world.wasm";
-const EXPECTED_RESPONSE: &str = "Hello Harry Potter!\n";
+const EXPECTED_RESPONSE: &str = "Hello World!\n";
 const OAK_FUNCTIONS_SERVER_PORT: u16 = 9001;
 
 #[tokio::test]
@@ -48,7 +48,7 @@ async fn start_client(port: u16, notify_sender: tokio::sync::oneshot::Sender<()>
     let request = hyper::Request::builder()
         .method(http::Method::POST)
         .uri(format!("http://localhost:{}/invoke", port))
-        .body(hyper::Body::from("Harry Potter"))
+        .body(hyper::Body::from("World"))
         .unwrap();
     let resp = client
         .request(request)
--- a/oak_functions/loader/Cargo.toml
+++ b/oak_functions/loader/Cargo.toml
@@ -37,3 +37,4 @@ wasmi = "*"
 [dev-dependencies]
 hyper = { version = "*", features = ["client"] }
 tempfile = "*"
+test_utils = { path = "../sdk/test_utils" }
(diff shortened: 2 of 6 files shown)


oak_functions/loader/src/tests.rs
```diff
@@ -26,6 +26,8 @@ use std::net::{Ipv6Addr, SocketAddr};
 use test::Bencher;
 
 const TEST_WASM_MODULE_PATH: &str = "testdata/non-oak-minimal.wasm";
+const HELLO_WORLD_MODULE_PATH: &str = "testdata/hello_world.wasm";
```
rbehjati, line 29: I could also use this in the `test_server` test. WDYT?

oak_functions/loader/src/tests.rs
```diff
@@ -94,7 +96,11 @@ fn bench_wasm_handler(bencher: &mut Bencher) {
         // to be less than a fixed threshold.
         // Note: `summary.mean` is in nanoseconds, even though it is not explicitly documented in
         // https://doc.rust-lang.org/test/stats/struct.Summary.html.
-        assert!(summary.mean < 50_000.0, "elapsed time: {}ns", summary.mean);
+        assert!(
+            summary.mean < 1_000_000.0,
```
rbehjati, line 100: Now it takes more than 600ns.

tiziano88: :grimacing: why is that? There's not much happening, is it?

oak_functions/loader/src/tests.rs
tiziano88: Why does it get compiled to testdata? That seems odd, given it's not checked-in.

(edit): I just noticed it is checked in, that is not great, I think it should just compile it to some target / bin folder and pick it up from there, like we do for the rest of the examples.

oak_functions/loader/src/tests.rs
rbehjati: This one is actually reading the request and writing the response. I suppose that is taking some time. The whole thing is still less than 1ms. Is that bad?

tiziano88: Ah you are right, I thought it was taking 1s now! Never mind, all good! :)

rbehjati: Yeah! I made that mistake before too :D

oak_functions/loader/Cargo.toml
```diff
@@ -34,6 +34,7 @@ wasmi = "*"
 [dev-dependencies]
 hyper = { version = "*", features = ["client"] }
 tempfile = "*"
+test_utils = { version = "0.1.0", path = "../sdk/test_utils" }
```
tiziano88, line 37:
```suggestion
test_utils = { path = "../sdk/test_utils" }
```

oak_functions/examples/hello_world/module/src/tests.rs
```diff
@@ -48,7 +48,7 @@ async fn start_client(port: u16, notify_sender: tokio::sync::oneshot::Sender<()>
     let request = hyper::Request::builder()
         .method(http::Method::POST)
         .uri(format!("http://localhost:{}/invoke", port))
-        .body(hyper::Body::from("Harry Potter"))
+        .body(hyper::Body::from("World"))
```
tiziano88, line 51: So serious now :)

tiziano88 (approved):

oak_functions/loader/src/tests.rs
rbehjati: Thanks. Done.

oak_functions/examples/hello_world/module/src/tests.rs
rbehjati: Indeed :D

PR #1972, opened 2021-04-14
Deploy hello world to Cloud Run

tiziano88:
At least it will do something interesting :)

Ref #1934

# Checklist

      construction.

### DIFF
--- a/scripts/deploy_oak_functions_loader
+++ b/scripts/deploy_oak_functions_loader
@@ -20,17 +20,22 @@ docker build \
 
 docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
-# Build and push an individual example application built on Oak Functions.
+# Build and push the `hello_world` example application built on Oak Functions.
 
-readonly FUNCTIONS_EXAMPLE_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-functions-example'
+cargo -Zunstable-options build --release \
+  --target=wasm32-unknown-unknown \
+  --manifest-path=./oak_functions/examples/hello_world/module/Cargo.toml \
+  --out-dir=./oak_functions/examples/hello_world/bin
+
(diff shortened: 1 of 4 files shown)


scripts/deploy_oak_functions_loader
```diff
@@ -20,17 +20,22 @@ docker build \
 
 docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
-# Build and push an individual example application built on Oak Functions.
+# Build and push the `hello_world` example application built on Oak Functions.
 
-readonly FUNCTIONS_EXAMPLE_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-functions-example'
+cargo -Zunstable-options build --release \
```
rbehjati, line 25: There is a `TODO(#1943)` comment on line 10 of this file. Can that be removed now?
(Actually, I think I should have removed it when I merged #1943)

scripts/deploy_oak_functions_loader
```diff
@@ -20,17 +20,22 @@ docker build \
 
 docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
-# Build and push an individual example application built on Oak Functions.
```
rbehjati, line 23: Should this file be moved to `oak_functions/examples/hello_world` since it is very `hello_world` specific? Or perhaps it should be made parametric to be usable for other examples too, in a separate PR.

rbehjati (approved):
Nice! Thanks.

scripts/deploy_oak_functions_loader
tiziano88: Yes I'll do another PR to replace commands with the corresponding runner versions.

scripts/deploy_oak_functions_loader
tiziano88: Yeah I was thinking about that too, currently each example would need its own Dockerfile anyways, so I'm not sure how much value there is in parametrizing the script only. I need to read more on how to create docker images in different ways perhaps, then it may be easier to parametrize everything, either with scripts or in runner.

PR #1976, opened 2021-04-14
Refresh lookup data at periodic interval

tiziano88:
- Rename `load_lookup_entries` to `parse_lookup_entries`
- Put the `LookupData` instance behind an `Arc`
- Start a separate `tokio` task to refresh data periodically
- Add more logging at various points

Ref #1930

# Checklist

      construction.


oak_functions/loader/src/main.rs
```diff
@@ -190,12 +215,18 @@ async fn main() -> anyhow::Result<()> {
...
-    let lookup_data = LookupData::new_empty(&config.lookup_data_url);
+    let lookup_data = Arc::new(LookupData::new_empty(&config.lookup_data_url));
     if !config.lookup_data_url.is_empty() {
-        lookup_data
-            .refresh()
-            .await
-            .context("could not load lookup data")?;
+        let lookup_data_download_period = match config.lookup_data_download_period {
+            Some(v) => v,
+            None => {
```
conradgrobler, line 222: Should we enforce this? I can imagine cases where the data is stable enough that refreshes might not be required outside of the normal restarts of the runtime.

oak_functions/loader/src/main.rs
```diff
@@ -190,12 +215,18 @@ async fn main() -> anyhow::Result<()> {
     let wasm_module_bytes = fs::read(&opt.wasm_path)
         .with_context(|| format!("Couldn't read Wasm file {}", &opt.wasm_path))?;
 
-    let lookup_data = LookupData::new_empty(&config.lookup_data_url);
+    let lookup_data = Arc::new(LookupData::new_empty(&config.lookup_data_url));
     if !config.lookup_data_url.is_empty() {
-        lookup_data
```
conradgrobler, line 195: I think that this should either remain here, or the periodic refresh code should first do the refresh before waiting in the loop. At the moment it seems like the lookup data will remain empty for the first period if I am reading the logic correctly.

oak_functions/loader/src/main.rs
tiziano88: Good point, made more lenient now.

oak_functions/loader/src/main.rs
tiziano88: Done, I think it makes more sense now, thanks!

conradgrobler (approved):

PR #1979, opened 2021-04-15
Add runner commands for oak-functions examples

### LINKED ISSUE #1964: Extend runner with commands for running oak-functions examples
This is particularly needed to be able to run examples as part of the CI builds. 
Ref: https://github.com/project-oak/oak/pull/1958#pullrequestreview-634991559

rbehjati:
- Adds a new command (run-functions-examples) to the `runner`
- Adds `example.toml` file for the oak-functions `hello_world` example.

Fixes #1964


conradgrobler (approved):

runner/src/main.rs
```diff
@@ -600,10 +656,69 @@ struct Example {
...
+        self.backends
+            .iter()
+            .map(|(name, backend)| Step::Single {
+                name: name.to_string(),
+                command: build(&backend.target, build_client),
+            })
+            .collect()
+    }
+
+    // Recursively constructs backend steps.
```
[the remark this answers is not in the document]
rbehjati: Thanks. Fixed.

runner/src/main.rs
```diff
@@ -600,10 +656,69 @@ struct Example {
...
 #[serde(deny_unknown_fields)]
 struct Application {
+    // In case of classic Oak applications, this is the path to a manifest file.
+    // In case of Oak functions applications, this is an empty string.
     manifest: String,
+    // In case of classic Oak applications, this is the path to the `.oak` application file.
+    // In case of Oak functions applications, this is the path to the Wasm module.
     out: String,
     modules: HashMap<String, Target>,
 }
```
tiziano88, line 724: Have you considered specializing this in ApplicationClassic and ApplicationFunction, instead of using a string at the top level to determine how to interpret the fields in this struct?

runner/src/main.rs
```diff
@@ -180,7 +182,30 @@ fn run_examples(opt: &RunExamples) -> Step {
...
+            .filter(|example| match &opt.example_name {
+                Some(example_name) => &example.name == example_name,
+                None => true,
+            })
+            .map(|example| run_functions_example(opt, example))
+            .collect(),
+    }
+}
+
+fn build_wasm_module(name: &str, target: &Target, out_dir: &str) -> Step {
```
tiziano88, line 208: I'm not sure how to interpret `out_dir` here, especially since in #1981 the logic is more tightly bound to how modules actually get compiled, based on the Cargo.toml and workspace configuration. In fact it seems that `out_dir` is ignored for Rust targets now anyways?

oak_functions/loader/src/main.rs
```diff
@@ -212,29 +214,61 @@ async fn background_refresh_lookup_data(lookup_data: &LookupData, period: Durati
...
-    let lookup_data = Arc::new(LookupData::new_empty(
-        &config.lookup_data_url,
-        logger.clone(),
-    ));
+    let address = SocketAddr::from((Ipv6Addr::UNSPECIFIED, opt.http_listen_port));
+    let server_handle = tokio::spawn(async move {
+        create_and_start_server(&address, &wasm_module_bytes, notify_receiver, logger)
+            .await
+            .context("error while waiting for the server to terminate")
+    });
```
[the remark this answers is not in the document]
tiziano88: Ah my bad, I misread the `done.load` below, I thought that was doing the lookup data refresh now, all good! As you say it should still work, even after #1990 (which I will rebase on top of this).

oak_functions/loader/src/main.rs
```diff
@@ -212,29 +214,61 @@ async fn background_refresh_lookup_data(lookup_data: &LookupData, period: Durati
...
+
+    // The server is started in its own thread, so just block the current thread until a signal
+    // arrives. This is needed for getting the correct status code when running with `runner`.
+    while !done.load(Ordering::Relaxed) {
+        // There are few synchronization mechanisms that are allowed to be used in a signal
+        // handler context, so use a primitive sleep loop to watch for the termination
+        // notification (rather than something more accurate like `std::sync::Condvar`).
+        // See e.g.: http://man7.org/linux/man-pages/man7/signal-safety.7.html
+        std::thread::sleep(std::time::Duration::from_millis(100));
+    }
```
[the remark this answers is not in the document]
tiziano88: LGTM, I think tokio::signal (similar to what you have implemented in tests) would work here, but let's look into it separately.

runner/src/main.rs
```diff
@@ -595,19 +653,267 @@ struct Example {
...
+                }
+            })
+    }
+}
+
+/// A construct representing either an Oak Classic or an Oak Functions application.  
+///
+/// The condition that only one of `classic` or `functions` should be non-empty is
+/// checked in each operation of this struct. If neither or both are empty, the
+/// operation panics with an error message.
```
tiziano88, line 709: I guess the YAML serialization makes it difficult to represent this condition idiomatically as an enum? If that works, it may remove the manual checks and panics below (I think).

runner/src/main.rs
```diff
@@ -169,17 +174,38 @@ fn run_examples(opt: &RunExamples) -> Step {
...
             })
             .filter(|example| {
-                example.applications.is_empty()
-                    || example
-                        .applications
-                        .get(opt.application_variant.as_str())
-                        .is_some()
+                example
+                    .applications
+                    .empty_or_has_application_variant(opt.application_variant.as_ref())
```
tiziano88, line 179: (optional) Personally I preferred the previous explicit check for empty or the specific application variant, now this function is just doing two things instead of one.

runner/src/main.rs
```diff
@@ -600,10 +656,69 @@ struct Example {
...
+    fn get_application(&self, application_variant: &str) -> &Application {
+        self.applications
+            .get(application_variant)
+            .unwrap_or_else(|| panic!(
+                "Unsupported application variant: {} (supported variants include: all, rust, cpp, go, nodejs, none)",
+                application_variant)
+            )
+    }
+
+    fn construct_backend_build_steps(&self, build_client: &BuildClient) -> Vec<Step> {
```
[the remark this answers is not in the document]
tiziano88: I thought it was being used like the other methods that return iterator-like things (that get additionally filtered or mapped), not necessarily Vec; either way not important.

runner/src/main.rs
```diff
@@ -595,19 +653,267 @@ struct Example {
...
+        run_clients: Step,
+    ) -> Step {
+        match (&self.classic, &self.functions) {
+            (Some(applications), None) => {
+                if applications.is_empty() {
+                    run_clients
+                } else {
+                    let opt = opt
+                        .left()
+                        .expect("Oak Classic application run with RunExamples options.");
```
tiziano88, line 777: I cannot think of an actionable suggestion for this PR, but it feels there may be a type safe way of expressing these constraints instead of manually checking all the combinations. If / when we come up with one perhaps we can refactor this.

runner/src/main.rs
```diff
@@ -595,19 +653,267 @@ struct Example {
...
+                foreground: Box::new(run_clients),
+            }
+        }
+    }
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+struct ApplicationsList<T> {
+    applications: HashMap<String, T>,
```
tiziano88, line 895: I guess this now results in an extra `applications` field everywhere, and having to forward methods like `is_empty` and `get`, let's think (in a separate PR) if it can be removed perhaps.

tiziano88 (approved):
Thanks!

runner/src/main.rs
rbehjati: Good point. Done.

runner/src/main.rs
rbehjati: Yes. There is no default (de)serializer for that kind of enum. But I think we can make it work if we define our own Serializer and Deserializer. I think this can be done in a separate PR, together with some other refactoring.

runner/src/main.rs
rbehjati: I agree! I have been playing with a few ideas, without much luck. But I'll continue in a separate PR.

runner/src/main.rs
rbehjati: I agree. Removed this class, but that resulted in some repetitions in the code. I'll try to fix that in a separate PR, together with other potential refactoring.

PR #1980, opened 2021-04-15
Enable HTTPS fetching of lookup data

### LINKED ISSUE #1973: Cannot load lookup data over HTTPS from Oak Functions loader
Only HTTP is currently supported.

```
1: error trying to connect: invalid URL, scheme is not http
```

anghelcovici:
# Checklist

      construction.

Fixes #1973

### DIFF
--- a/oak_functions/examples/Cargo.lock
+++ b/oak_functions/examples/Cargo.lock
@@ -33,22 +33,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
+name = "base64"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
+
+[[package]]
 name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
+name = "bumpalo"
+version = "3.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63396b8a4b9de3f4fdfb320ab6080762242f66a8ef174c49d8e19b674db4cdbe"
+
+[[package]]
 name = "byteorder"
 version = "1.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


tiziano88 (approved):
Cool, does this use the CA store installed on the system / docker image?

anghelcovici:
@tiziano88 - I assumed so, but tbh I haven't check to much

oak_functions/loader/src/main.rs
```diff
@@ -107,7 +108,8 @@ impl LookupData {
     /// at a future time.
     async fn refresh(&self) -> anyhow::Result<()> {
         // TODO(#1930): Avoid loading the entire file in memory for parsing.
-        let client = hyper::Client::new();
+        let https = HttpsConnector::new();
```
rbehjati, line 111: In the rest of the code we use `hyper_rustls::HttpsConnector`, which allows using custom CA store:
https://github.com/project-oak/oak/blob/57d965254fdf4d22d87b7bf0c2863caa109cd851/oak_runtime/src/node/http/client.rs#L309-L316

tiziano88: cc @conradgrobler

anghelcovici: Do we need a custom CA store? @tiziano88 @rbehjati

tiziano88: I don't know if @conradgrobler has an opinion, having implemented the CA store for the oak_loader, anyways I don't think for the functions loader it matters that much, since we are not making promises about the confidentiality of the data in terms of TLS certs, we only use it for fetching the lookup data, so I think your version is fine.

tiziano88: That said, it may be nice to use hyper_rustls for consistency with the existing code, even if we would not manually add a CA store. WDYT?

conradgrobler: We needed support for custom CA certificates to run the other examples because we used certificates that we signed ourselves. For this download I assume the production data will usually be hosted somewhere with a proper certificate, so this should not be required here. For test cases it can just use HTTP.

tiziano88: AFAICT

- hyper-rustls uses https://github.com/ctz/rustls
- hyper-tls uses https://github.com/sfackler/rust-native-tls

I don't know if we have an opinion already on which to prefer, but presumably we should stick to one everywhere if possible.

Perhaps we should talk to the security team about this too, @anghelcovici could you get in touch with them?

tiziano88: Also see https://github.com/project-oak/oak/issues/1254 (it seems the conclusion was to use rustls)

anghelcovici: Ok, I've switched to hyper_rustls for consistency. Using native roots for now, we can change this if we need.

tiziano88 (approved):

Issue #1965, opened 2021-04-14
Move common functionality in oak and oak-functions to separate crates


rbehjati:
Refactor common non-oak specific functionality into a separate crate that both oak and oak-functions crates can use.
Ref: Last point in https://github.com/project-oak/oak/pull/1958#pullrequestreview-634991559 
See also:  https://github.com/project-oak/oak/pull/1958#r612534679

PR #1981, opened 2021-04-15
Unify wasm compilation flags

tiziano88:
This removes some of the logic around compiling Wasm modules to a
specific out dir, but I think it makes more sense to keep the default
location and modify the app manifest files to point to it rather than
convincing cargo to put compiled artifacts in non-standard folders, even
if the manifests end up being slightly longer.

- Fix the path of some of the internal dependencies, since some tests
started randomly failing, presumably because of some interference with
externally published crates with the same names (namely `prost`).

### DIFF
--- a/examples/abitest/oak_app_manifest.toml
+++ b/examples/abitest/oak_app_manifest.toml
@@ -1,9 +1,9 @@
 name = "abitest"
 
 [modules]
-frontend_module = { path = "examples/abitest/bin/abitest_0_frontend.wasm" }
-backend_module = { path = "examples/abitest/bin/abitest_1_backend.wasm" }
-linear_handles_module = { path = "examples/abitest/bin/abitest_linear_handles.wasm" }
+frontend_module = { path = "examples/bin/abitest_0_frontend.wasm" }
+backend_module = { path = "examples/bin/abitest_1_backend.wasm" }
+linear_handles_module = { path = "examples/bin/abitest_linear_handles.wasm" }
 
 [initial_node_configuration]
 wasm_module_name = "frontend_module"
--- a/sdk/rust/oak_app_build/src/main.rs
+++ b/sdk/rust/oak_app_build/src/main.rs
@@ -126,7 +126,7 @@ fn get_module_cache_path(manifest_dir: &Path, module_sha256_sum: &str) -> PathBu
 }
 
 /// Get path for the output application file in [`OUTPUT_DIRECTORY`].
-fn get_output_path(manifest_dir: &Path, app_name: &str) -> PathBuf {
+fn get_output_file_path(manifest_dir: &Path, app_name: &str) -> PathBuf {
     manifest_dir
         .join(OUTPUT_DIRECTORY)
         .join(format!("{}.oak", app_name))
@@ -297,10 +297,14 @@ async fn main() -> anyhow::Result<()> {
         module_signatures: signatures,
     };
 
-    let output_file = get_output_path(&manifest_dir, &manifest.name);
+    let output_file_path = get_output_file_path(&manifest_dir, &manifest.name);
+
+    std::fs::create_dir_all(output_file_path.parent().unwrap())
+        .context("Couldn't create output dir")?;
 
     // Serialize application.
-    write_config_to_file(&app_config, &output_file).context("Couldn't write serialized config")?;
+    write_config_to_file(&app_config, &output_file_path)
+        .context("Couldn't write serialized config")?;
 
     Ok(())
 }
--- a/sdk/rust/oak_tests/src/lib.rs
+++ b/sdk/rust/oak_tests/src/lib.rs
@@ -28,7 +28,7 @@ use oak_abi::{
 use oak_client::interceptors::{
     self, auth::AuthInterceptor, label::LabelInterceptor, CombinedInterceptor,
 };
-use std::{collections::HashMap, path::PathBuf, process::Command, sync::Arc};
+use std::{collections::HashMap, process::Command, sync::Arc};
 use tonic::transport::{Certificate, Channel, ClientTlsConfig, Identity};
 
 pub enum Profile {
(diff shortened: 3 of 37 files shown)


sdk/rust/oak_tests/src/lib.rs
```diff
@@ -100,61 +106,43 @@ const RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(800
 /// Convenience helper to build and run a single-Node Application with the given module name, using
```
rbehjati, line 106: The comment still mentions the module name. Can you update it?

sdk/rust/oak_app_build/src/main.rs
```diff
@@ -299,6 +299,9 @@ async fn main() -> anyhow::Result<()> {
 
     let output_file = get_output_path(&manifest_dir, &manifest.name);
 
+    std::fs::create_dir_all(manifest_dir.join(OUTPUT_DIRECTORY))
```
rbehjati, line 302: Could this be moved into `get_output_path`? Since `get_output_path` is only called here.

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -38,33 +38,42 @@ pub enum Profile {
...
+        package_name
+    ));
+    module_path
+        .to_str()
+        .expect("could not convert PathBuf to String")
+        .to_string()
+}
 
+/// Uses cargo to compile a Rust manifest to Wasm bytes.
+pub fn compile_rust_wasm(manifest_path: &str, profile: Profile) -> anyhow::Result<Vec<u8>> {
```
rbehjati, line 65: Does the function with the same name in `oak_functions` need to be updated too? Perhaps in a separate PR, or as part of #1965.

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -100,61 +106,43 @@ const RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(800
...
-    run_single_module_with_config(
-        module_wasm_file_name,
-        entrypoint_name,
-        ConfigMap::default(),
-        permissions,
-    )
+    run_single_module_with_config(entrypoint_name, ConfigMap::default(), permissions)
 }
 
 /// Convenience helper to build and run a single-Node application with the given Wasm module file
```
rbehjati, line 123: Update comment?

sdk/rust/oak_tests/src/lib.rs
```diff
@@ -100,61 +106,43 @@ const RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(800
...
-        entrypoint_name,
-        config_map,
-        permissions,
-    );
+    let combined_config = runtime_config(entrypoint_name, config_map, permissions);
     oak_runtime::configure_and_run(combined_config)
 }
 
 /// Build the configuration needed to launch a test Runtime instance that runs a single-Node
 /// application with the given Wasm module file name and entrypoint.
```
rbehjati, line 135: Update comment?

examples/abitest/oak_app_manifest.toml
```diff
@@ -1,9 +1,9 @@
 name = "abitest"
 
 [modules]
-frontend_module = { path = "examples/abitest/bin/abitest_0_frontend.wasm" }
-backend_module = { path = "examples/abitest/bin/abitest_1_backend.wasm" }
-linear_handles_module = { path = "examples/abitest/bin/abitest_linear_handles.wasm" }
+frontend_module = { path = "examples/target/wasm32-unknown-unknown/release/abitest_0_frontend.wasm" }
+backend_module = { path = "examples/target/wasm32-unknown-unknown/release/abitest_1_backend.wasm" }
+linear_handles_module = { path = "examples/target/wasm32-unknown-unknown/release/abitest_linear_handles.wasm" }
```
anghelcovici, line 6: This doesn't look nice at all. I clearly prefer the previous style and I am unsure why this change is needed here.

anghelcovici:
`presumably because of some interference with externally published crates with the same names `
What does presumably mean here? Was this the reason that the tests were failing?

examples/abitest/oak_app_manifest.toml
tiziano88: This is the canonical location where the artifacts are built. The previous version was compiling everything twice , one for tests and one for the examples, without even reusing the intermediate cache. Basically test code and runner were doing the same job twice. The previous version was also not reusing cache between examples, because it ignored the workspace structure in which the examples are defined. Let me clarify in the commit message and in code comments.

anghelcovici: How does the output path of the binary influences that? This feels strange

tiziano88: It's the target dir that affects it, maybe we can get the best of both by making the bin dir relative to the workspace root, let me try that

sdk/rust/oak_tests/src/lib.rs
tiziano88: Done.

sdk/rust/oak_tests/src/lib.rs
tiziano88: Done.

sdk/rust/oak_tests/src/lib.rs
tiziano88: Done.

sdk/rust/oak_tests/src/lib.rs
tiziano88: Yes, I think all of them should be unified in #1965.

sdk/rust/oak_app_build/src/main.rs
tiziano88: Done.

examples/abitest/oak_app_manifest.toml
tiziano88: PTAL now, I think now it is a compromise between short paths and actually sharing caches in a meaningful way (though personally I still prefer the longer more explicit paths -- but let's have a separate discussion about that perhaps).

tiziano88:
> `presumably because of some interference with externally published crates with the same names `
> What does presumably mean here? Was this the reason that the tests were failing?

Basically things started failing with weird errors about prost and oak_abi versions when I changed some dependencies, I didn't look into it closely, but in general we should always refer to local crates by paths rather than by version, so I did that in a few places. I will do a more thorough cleanup in a separate PR.

anghelcovici (approved):

PR #1991, opened 2021-04-20
Implement internal Wasm allocator API

### LINKED ISSUE #754: Rely on memory allocation in Wasm ABI
Currently, most Wasm ABI functions that expect to read data from other nodes or from the runtime need to pre-allocate a buffer, and pass a pointer (really a Wasm linear memory offset) and size as part of the function call, which is not very ergonomic and potentially also wasteful (since in general the caller does not know the exact size of the data to be returned upfront).

We could mitigate this by requiring an Oak Wasm module to expose a Wasm function (e.g. `oak_alloc(len: usize)` or similar) that allocates the specified number of bytes into Wasm linear memory **using the Rust memory allocator**, and returns its offset in the Wasm linear memory. This would be used by the Oak runtime itself when fulfilling calls that require returning owned allocated objects.

For instance, [channel_read](https://github.com/project-oak/oak/blob/master/docs/abi.md#channel_read) could return two allocated buffers (one for bytes, one for handles), instead of requiring the caller to pass them in. This may get simpler with the addition of [multiple return values](https://hacks.mozilla.org/2019/11/multi-value-all-the-wasm/) to Wasm.

This is just a placeholder issue for now, in case anyone has any additional information that we should consider.

tiziano88:
- Use consistent terminology in docs and code
  - use valid Wasm types to describe the ABI (`i32`)
  - give meaningful names to parameters in ABI doc
  - use "address" (instead of "pointer" or "reference") to refer to a Wasm
  linear memory offset
- make `read_request` and `storage_get_item` rely on the new alloc-based API
  - the implementation is now much simpler and readable
- use hello world wasm module in the loader test
  - this is somewhat superflous since it's very similar to the hello
  world example itself, anyways we can dedup in a future PR
  - remove the testdata directory, since an empty wasm module is not
  actually runnable as it does not have a valid `alloc` function
- copy paste wasm module compilation logic from the other 2 instances that
  already exist in other crates (we really need to unify those)
- remove `ERR_BUFFER_TOO_SMALL` OakStatus value as it's not needed any
  more

Fix #754

# Checklist

      construction.


conradgrobler (approved):

oak_functions/loader/src/server.rs
```diff
@@ -176,12 +175,47 @@ impl WasmInterface {
...
+    pub fn alloc(&mut self, len: u32) -> AbiPointer {
+        let result = self.instance.as_ref().unwrap().invoke_export(
+            ALLOC_FUNCTION_NAME,
+            &[wasmi::RuntimeValue::I32(len as i32)],
+            // When calling back into `alloc` we don't need to expose any of the rest of the ABI
+            // methods.
+            &mut wasmi::NopExternals,
+        );
+        let result_value = result
+            .expect("`alloc` call failed")
```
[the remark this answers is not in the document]
tiziano88: Great point, created https://github.com/project-oak/oak/issues/1992 to track.

oak_functions/loader/src/tests.rs
```diff
@@ -30,9 +29,7 @@ use std::{
 extern crate test;
 use test::Bencher;
 
-const TEST_WASM_MODULE_PATH: &str = "testdata/non-oak-minimal.wasm";
 const MANIFEST_PATH: &str = "examples/hello_world/module/Cargo.toml";
```
ipetr0v, line 32: Is it fine that `oak_functions` crate relies on `examples` crate (which it doesn't include) for testing?

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -27,45 +27,34 @@ use std::{
...
 /// Uses cargo to compile a Rust manifest to Wasm bytes.
-pub fn compile_rust_wasm(
-    manifest_path: &str,
-    module_wasm_file_name: &str,
-) -> anyhow::Result<Vec<u8>> {
-    let mut module_path = PathBuf::from(manifest_path);
-    module_path.pop();
-    module_path.push("bin");
-
+pub fn compile_rust_wasm(manifest_path: &str) -> anyhow::Result<Vec<u8>> {
```
ipetr0v, line 43: Why is it copied from `oak_tests` here? To remove the dependency?

ipetr0v (approved):

oak_functions/loader/src/tests.rs
tiziano88: I think it's fine, otherwise we need yet another example anyways.

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: To make sure that it is using the same compilation logic, and caches are reused.

oak_functions/loader/src/server.rs
```diff
@@ -249,6 +280,22 @@ impl WasmState {
...
         .assert_no_start();
 
+        let alloc = instance
+            .export_by_name(ALLOC_FUNCTION_NAME)
+            .context("could not find Wasm `alloc` export")?
+            .as_func()
+            .cloned()
+            .context("could not interpret Wasm `alloc` export as function")?;
+        let expected_signature = wasmi::Signature::new(&[ValueType::I32][..], Some(ValueType::I32));
+        if alloc.signature() != &expected_signature {
```
rbehjati, line 290: Do we have a similar check for the `main` function too? If not, do we need that?

oak_functions/loader/src/server.rs
```diff
@@ -249,6 +280,22 @@ impl WasmState {
...
+            anyhow::bail!(
+                "invalid signature for `alloc` export: {:?}, expected: {:?}",
+                alloc.signature(),
+                expected_signature
+            );
+        }
+
+        abi.instance = Some(instance.clone());
```
rbehjati, line 298: Since now both `WasmState` and `WasmInterface` have `instance` as a field, maybe they could be merged into one struct like you had at the beginning?

oak_functions/loader/src/server.rs
```diff
@@ -440,14 +485,14 @@ fn oak_functions_resolve_func(
         }
     };
 
-    if &sig != signature {
+    if signature != &expected_signature {
         return Err(wasmi::Error::Instantiation(format!(
-            "Export `{}` doesn't match expected type {:?}",
-            field_name, signature
+            "Export `{}` doesn't match expected type {:?}, expected: {:?}",
```
rbehjati, line 490:
```suggestion
            "Export `{}` doesn't match expected signature: got {:?}, expected: {:?}",
```
nit!

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -183,3 +92,21 @@ pub fn result_from_status<T>(status: i32, val: T) -> Result<T, OakStatus> {
         None => Err(OakStatus::Unspecified),
     }
 }
+
+#[no_mangle]
+pub extern "C" fn alloc(len: u32) -> *mut u8 {
```
rbehjati, line 97: I guess this means that the examples don't have to implement their own `alloc` function. If so, maybe explain in the docs that we have an implementation of `alloc` in our Rust SDK? Or add a doc comment here?

rbehjati (commented):
This is nice! Just a few after-the-fact comments!

oak_functions/loader/src/server.rs
tiziano88: Added in https://github.com/project-oak/oak/pull/2010

oak_functions/loader/src/server.rs
tiziano88: Done in https://github.com/project-oak/oak/pull/2010

oak_functions/sdk/oak_functions/src/lib.rs
tiziano88: Done in https://github.com/project-oak/oak/pull/2010

oak_functions/loader/src/server.rs
tiziano88: I tried but it was still a bit messy, but I agree we should look into it more now.

PR #1993, opened 2021-04-21
Runner refactor

rbehjati:
- Splits `runner/main.rs` into two additional modules for `files` and `examples`.
- Changes `Applications` from a `struct` to be an `enum`, with a specialized `serde::Deserializer`. 
- Separates between internal representation of `OakExample`s for running and the external representation suitable for deserializing `toml` files.

The first commit is just splitting `runner/main.rs` without any changes. The actual change is in the second commit. 

Follow up of #1979


runner/src/examples.rs
```diff
@@ -0,0 +1,970 @@
...
+/// A construct representing either an Oak Classic or an Oak Functions application.
+///
+/// The condition that only one of `classic` or `functions` should be non-empty is
+/// checked in each operation of this struct. If neither or both are empty, the
+/// operation panics with an error message.
+#[derive(Debug)]
+enum Applications {
+    Classic(HashMap<String, ApplicationClassic>),
+    Functions(HashMap<String, ApplicationFunctions>),
+}
```
tiziano88, line 55: I think https://serde.rs/enum-representations.html#internally-tagged may help here, have you looked into it?

Also, I think we can simplify this by pushing the enum one level deep:

```rust
#[serde(tag = "type")]
enum Application {
  Classic { ... }
  Functions { ... }
}

... applications: HashMap<String, Application>
```

I don't think we need to check that applications are all of the same type anyways, I mean it will be pretty obvious if they are mixed and don't work well together, and if and when we do have the use case for applications of different types in the same example, we would just remove that check anyways as the first thing.

WDYT?

rbehjati: Thanks. I pushed an update, but I am not entirely sure what should be done in the following situations: 
- in case of an example that has an empty applications map (such as the `tls_attestation` example). I think my current implementation excludes it from both `run-examples` and `run-functions-examples`
- in case of an example that has a mix of both application types. We don't currently have such an example, but the current implementation could run both commands for it, in each case running only the applications of the relevant type.

rbehjati: Changed it, so now if an example has a mix of both types of applications, it won't run.

rbehjati:
@tiziano88 Just a friendly reminder that this PR is awaiting a review.

runner/src/examples.rs
```diff
@@ -0,0 +1,938 @@
...
+                    name: name.to_string(),
+                    // Each `backend` is included as background step.
+                    background: run(&backend, &self.get_build_client(), Vec::new()),
+                    foreground: Box::new(backend_steps),
+                }
+            })
+    }
+}
+pub struct ClassicExample<'a> {
+    example: &'a Example,
```
tiziano88, line 241: Not sure why all the lifetimes and Box are needed, is it not possible to use objects owned by this struct?

runner/src/examples.rs
```diff
@@ -0,0 +1,938 @@
...
+        } => {
+            let metadata = cargo_metadata::MetadataCommand::new()
+                .manifest_path(cargo_manifest)
+                .exec()
+                .unwrap();
+            Step::Single {
+                name: format!("wasm:{}:{}", name, cargo_manifest.to_string()),
+                command: Cmd::new(
+                    "cargo",
+                    // Keep this in sync with `/oak_functions/sdk/test/utils/src/lib.rs`.
```
tiziano88, line 575: I think a line is missing here compared to where this was branched from.

tiziano88 (approved):
Thanks, the manifests look quite elegant now I think! Although I am not 100% convinced about the trait and different implementation in Runner, it feels like there is some duplication, but maybe it's unavoidable. We can keep iterating if / when we come up with anything more concise.

runner/src/examples.rs
rbehjati: Thanks. Fixed.

runner/src/examples.rs
rbehjati: I tried. But the main problem was that I could not find a nice way to move the example objects from the vector in `run_examples` and pass them to the constructor of this class.

rbehjati (commented):
I also don't like the repeated code. Hopefully, we can find a way to improve that in future iterations.

PR #1995, opened 2021-04-21
Automatically update oak_functions Cargo.lock

ipetr0v:


rbehjati:
What is `version = 3`? We don't seem to have it anywhere else.

ipetr0v:
I'm not sure

tiziano88:
Ah I think this is my fault, I probably run a newer version of Rust myself outside docker just before checking my PR in, but I'm not sure why it was not detected by the git_check_diff step?

tiziano88 (approved):

PR #2001, opened 2021-04-22
Fix Bazel in Android Docker

ipetr0v:


anghelcovici:
How did this work before? :)

tiziano88:
Magic

Issue #2003, opened 2021-04-23
Change default `GITHUB_TOKEN` permissions to read-only


conradgrobler:
This requires that build_rust_docs.yaml and reproducibility.yaml are updated to explicitly grand write permission

conradgrobler:
Reference: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

PR #2004, opened 2021-04-23
Add explicit write permissions to actions

conradgrobler:
This is required before we can change the default `GITHUB_TOKEN` permissions to read-only.

See #2003 and https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

# Checklist

      construction.

### DIFF
--- a/.github/workflows/reproducibility.yaml
+++ b/.github/workflows/reproducibility.yaml
@@ -10,6 +10,9 @@ jobs:
   build_reproducibility_index:
     runs-on: ubuntu-20.04
 
+    permissions:
+      contents: write
+
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


tiziano88 (approved):

.github/workflows/reproducibility.yaml
```diff
@@ -10,6 +10,9 @@ jobs:
   build_reproducibility_index:
     runs-on: ubuntu-20.04
 
+    permissions:
+      contents: write
```
rbehjati, line 14: Does this really work? Publishing reproducibility index on a PR after merge is failing with `status: 403`. I wonder if it is happening because of the changes related to the `codecov` issue.

conradgrobler: Good catch. This is failing because the steps also adds a comment to the PR by posting an issue comment (https://docs.github.com/en/rest/reference/issues#create-an-issue-comment). This requires it to also have the issues:write permission.

PR #2011, opened 2021-04-26
Add policy to oak_functions_loader

### LINKED ISSUE #1987: Simple policy for Oak Functions
Implement support for policies that avoid side-channel leaks from the response. The policies can be specified using protobuf as suggested by @tiziano88:

```protobuf
message Policy {
  // A fixed size for responses returned by the trusted runtime. 
  // If the response computed by the feature code is smaller than this amount, 
  // it is padded with additional data before encryption in order to make the
  // payload size exactly this size. If the response is larger than this amount, 
  // the trusted runtime discards the response and instead sends a message of  
  // exactly this size to the client, containing an error code indicating the 
  // failure.
  int32 constant_response_size_bytes = 1;

  // Similar to the previous one, but controls the amount of time the function is 
  // allowed to run for. If the function finishes before this time, the response is 
  // not sent back until the time is elapsed. If the function does not finish within
  // this deadline, the trusted runtime sends a message to the client containing 
  // an error code indicating the failure, of the size specified by the previous 
  // parameter.
  int32 constant_process_time_millis = 2;
}
```
More properties, for example related to differential privacy, can be added later.

For Oak Functions v0, we'll implement this as part of the server configuration. But in future it will be changed to a client-provided policy.

rbehjati:
Fixes #1987


oak_functions/examples/hello_world/config.toml
```diff
@@ -1,3 +1,5 @@
 # TODO(#1973): Change schema to HTTPS when supported.
 lookup_data_url = "http://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
 lookup_data_download_period = "10m"
+policy.constant_response_size_bytes = 90
+policy.constant_process_time_millis = 20
```
tiziano88, line 5: The time should be a general duration, like the `lookup_data_download_period` above. I.e. it should be expressed as `20ms` (string) here.

tiziano88, line 5: `constant_processing_time_millis` ? I think `process` is a bit ambiguous (is it a CPU process?), plus here it is really the processing time that is being measured.

oak_functions/loader/src/server.rs
```diff
@@ -353,6 +443,98 @@ fn check_export_function_signature(
     }
 }
 
+pub async fn apply_policy<F, S>(
```
tiziano88, line 446: Doc.

oak_functions/loader/src/server.rs
```diff
@@ -353,6 +443,98 @@ fn check_export_function_signature(
...
+            let task = tokio::spawn(async move { function().await });
+
+            // Run the function and the timer tasks in parallel, and measure the time that it
+            // takes for the first task to complete.
+            let start = Instant::now();
+            let response = tokio::select! {
+                _ = timer => Err(anyhow::anyhow!("timeout!")),
+                rsp = task => rsp.unwrap()
+            };
+            let elapsed = start.elapsed().as_millis();
```
tiziano88, line 472: I was expecting a simpler logic here, something like:

```rust
let task = tokio::spawn(function());
tokio::time::sleep(Duration::from_millis(duration)).await;
match task.now_or_never() {
  Some(v) => ..
  None => ..
}
```

would that work?

oak_functions/loader/src/server.rs
```diff
@@ -353,6 +443,98 @@ fn check_export_function_signature(
...
+}
+
+/// Estimates the size of the given response.
+///
+/// Measures the size of the response when formatted according to
+/// [RFC-2616](https://tools.ietf.org/html/rfc2616#section-6).
+/// The only relevant parts of the response are the status code and the response body. We also add
+/// the `oak-policy-padding` header to all responses. This header name is therefore also included
+/// when estimating the size.
+fn estimate_size(response: &hyper::Response<Body>) -> anyhow::Result<i64> {
```
tiziano88, line 504: This seems like it is trying to undo what the inner call has already packaged in an HTTP request.  How about making the Wasm call return the body of the request, and wrap it in an HTTP respose afterwards, in a straightforward way?

rbehjati (commented):
Thanks for the comments.

oak_functions/examples/hello_world/config.toml
```diff
@@ -1,3 +1,5 @@
 # TODO(#1973): Change schema to HTTPS when supported.
 lookup_data_url = "http://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
 lookup_data_download_period = "10m"
+policy.constant_response_size_bytes = 90
+policy.constant_processing_time = 20ms
```
tiziano88, line 5: I think this needs to be quoted? Even github seems to think this is not syntactically valid (perhaps the Rust parser is more lenient).

oak_functions/loader/src/server.rs
```diff
@@ -43,6 +47,90 @@ type AbiPointerOffset = u32;
...
+
+impl Policy {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        match self.constant_response_size_bytes >= MIN_RESPONSE_SIZE {
+            true => Ok(()),
+            false => anyhow::bail!("Response size is too small"),
+        }
+    }
+
+    fn get_error_response(&self, reason: &str) -> anyhow::Result<Response<Body>> {
```
tiziano88, line 89: AFAICT this is actually creating the error response, right (not really getting it from the policy), so maybe rename to make_ or create_ ? Also add some documentation?

oak_functions/loader/src/server.rs
tiziano88: Exactly, I think we can just return `Option<Vec<u8>>` for now? Since the only status is whether the wasm function actually wrote a response or not, right? Although I'm not sure there are other error conditons. But yes, something along those lines. This has the added bonus that the Wasm invocation logic is agnostic of HTTP, so it could be used for gRPC or any other wrapper protocol, which I think we may want to do at some point (and seems a good way of designing this layer of the API). WDYT?

oak_functions/loader/src/server.rs
```diff
@@ -43,6 +47,90 @@ type AbiPointerOffset = u32;
...
+            let padding_bytes: Vec<u8> = repeat(b'0').take(required_padding as usize).collect();
+            http::response::Builder::new()
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                .header(POLICY_PADDING_HEADER, std::str::from_utf8(&padding_bytes)?)
+                .body(body.into())
+                .context("couldn't create response")
+        }
+    }
+
+    /// Pads the response by adding a number of zeros, to make the total size of the response equal
```
tiziano88, line 110: I would either leave the description of exactly what the padding is unspecified ("adding a number of unspecified characters"), or actually specify the character (though I think this is unnecessarily constraining the contract of this function, the specific character seems an impl detail and we should be able to change it without having to update this doc string.

oak_functions/loader/src/server.rs
```diff
@@ -43,6 +47,90 @@ type AbiPointerOffset = u32;
...
+    /// to `self.constant_response_size_bytes`
+    ///
+    /// The zeros are added as the header value for header `oak-policy-padding`.
+    /// This function will return an error if the size of the input `response` is larger than or
```
tiziano88, line 114:
```suggestion
    /// Returns an error if the size of the input `response` is larger than or
```

oak_functions/loader/src/server.rs
```diff
@@ -43,6 +50,92 @@ type AbiPointerOffset = u32;
...
+        let header_body_content = format!("{}{}", POLICY_PADDING_HEADER, body);
+        let required_padding =
+            self.constant_response_size_bytes as i32 - header_body_content.len() as i32;
+
+        if required_padding < 0 {
+            Err(anyhow::anyhow!(
+                "constant_response_size_bytes specified by the policy is too small."
+            ))
+        } else {
+            let padding_bytes: Vec<u8> = (0..required_padding).map(|_| b'0').collect();
```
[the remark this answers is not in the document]
tiziano88: Personally I think it is, since conceptually it's not really mapping anything, just repeating a character a number of times.

Actually I also just realised that `vec![b'0'; required_padding]` (note the semicolon) also works, and perhaps it is even more idiomatic?

oak_functions/loader/src/server.rs
tiziano88: Yes I think it's fine to postpone anyways, maybe we need >2 threads for this to work? Have you tried with like 4?

oak_functions/loader/src/server.rs
```diff
@@ -43,6 +47,110 @@ type AbiPointerOffset = u32;
...
+    #[serde(with = "humantime_serde")]
+    pub constant_processing_time: Duration,
+}
+
+impl Policy {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        match self.constant_response_size_bytes >= MIN_RESPONSE_SIZE {
+            true => Ok(()),
+            false => anyhow::bail!("Response size is too small"),
+        }
```
tiziano88, line 86: Possibly useful: https://docs.rs/anyhow/1.0.40/anyhow/macro.ensure.html

oak_functions/loader/src/server.rs
```diff
@@ -335,6 +443,95 @@ fn check_export_function_signature(
...
+}
+
+/// Estimates the size of the given response.
+///
+/// Measures the size of the response when formatted according to
+/// [RFC-2616](https://tools.ietf.org/html/rfc2616#section-6).
+/// The only relevant parts of the response are the status code and the response body. We also add
+/// the `oak-policy-padding` header to all responses. This header name is therefore also included
+/// when estimating the size.
+fn estimate_size(response: &FunctionsResponse) -> anyhow::Result<i64> {
```
tiziano88, line 504: I'm not sure why the size of the HTTP response needs to be estimated at all, and also why it is estimated in this particular way, and what does it mean for it to be an estimate; also the logic seems quite brittle, e.g. it would stop working for HTTP2, right?

I expect the logic to be simply that the sum of the body and the padding header is constant; no other details should be necessary, in particular no estimation is needed. Or am I missing some case in which that does not work?

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
rbehjati: This is much nicer. Thanks.

oak_functions/loader/src/server.rs
rbehjati: Yes. This is the error: 
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError(())', /workspaces/oak/oak_functions/sdk/test_utils/src/lib.rs:191:19
```

tiziano88: That seems to happen because the termination signal is not sent correctly; maybe the sender is dropped before the signal is sent? Note it needs to be explicitly sent in the test case.

oak_functions/loader/src/server.rs
rbehjati: Different status codes have different lengths. If we apply the size only to the body and the padding header, then the total size of the transmitted response could be different for different status codes. I think in general, for the same request we can still respond with 500 (INTERNAL_SERVER_ERROR) in some cases instead of 200. So, I think the status code should be included in the calculation of the size. 

Of course, if we want to support HTTP/2 this functionality has to be extended. But it is implementing the standard. I would not call it brittle.

It is estimated, because the final response has additional headers like `content-length`, which are not included in the calculation of the size here. I think that is what hyper adds to the response. And I think the size of those additional headers is constant.

oak_functions/loader/src/server.rs
rbehjati: Done.

oak_functions/loader/src/server.rs
tiziano88: Ah thanks, I didn't realise that the status code description actually gets sent back too! I thought it was just the numeric code.

rbehjati: I did think of that, but then we have to publish the schema for the request and the response body, and the client has to be able to deserialize the body, to be able to get the actual response and fail/success status. But I guess that should be fine. Created #2032.

tiziano88: We are effectively using an implicit schema by referring to header names and status codes anyways, so we may as well have an explicit one instead (e.g. either protobuf or JSON). Thanks for creating the issue, let's follow up there.

conradgrobler (approved):

oak_functions/loader/src/server.rs
rbehjati: There are two problems:
1. With multiple threads the client sends the request before the server is up. I could fix this by adding a `sleep`. 
2. I modified https://github.com/project-oak/oak/blob/29a6a2a1300579a292db6e66340491e5f95d2fcc/oak_functions/sdk/test_utils/src/lib.rs#L191-L192 to the following: 
    ```Rust
                    match v {
                        Err(_) => Poll::Pending,
                        Ok(_) => Poll::Ready(()),
                    }
    ```
This fixes the panics, but then again the test fails as before. The policy on processing time is not applied. I have a todo to get back to this when we fix #2026.

PS: I have not pushed this change as it does not really fix the test.

tiziano88 (approved):

Issue #2008, opened 2021-04-26
Investigate CI flakiness


conradgrobler:
There seems to be various race conditions that cause the CI builds to be flaky:

- oak_runtime integration tests: `test_metrics_gives_the_correct_number_of_nodes` is inconsistent. Sometimes the Wasm node still exists and the tests pass, but at on some occasions it has already exited (seeing that the main function does nothing) and then the test fails.
- proxy_attestaton example: the backend attestation service sometimes takes longer than the runtime to start up. The test then fails becuase the server cannot open a connection to the backend service.
- abittest integration tests: `test_leaks` checks the number of channels at a point where one channel on another thread is somtimes still open, but the channel closes later (within a second or so). This is counted as a leak, even though it resolves itself later.

tiziano88:
We should pay more attention at how background tasks / servers are created, terminated, and waited for.

For reference, in https://github.com/project-oak/oak/pull/2010/files I did something similar, in case anyone would like to implement a similar pattern and with similar names in other parts of the codebase.

PR #2019, opened 2021-04-27
Add logic to retry counts in case of possible leak

conradgrobler:
Race conditions in `test_leaks` sometime cause false positives. This happens if a back-end node is a bit slower than expected to close its orphaned channels. It a potential leak is detected, the code will now wait for 5 seconds and then retry the count to make sure it is a genuine leak.

See #2008 

# Checklist

      construction.

### DIFF
--- a/examples/abitest/module_0/rust/tests/integration_test.rs
+++ b/examples/abitest/module_0/rust/tests/integration_test.rs
@@ -22,6 +22,7 @@ use maplit::hashmap;
 use oak_abi::proto::oak::application::ConfigMap;
 use serial_test::serial;
 use std::{collections::HashMap, sync::Arc};
+use tokio::time::{sleep, Duration};
 
 // Constants for Node config names that should match those in the textproto
 // config held in ../../../config.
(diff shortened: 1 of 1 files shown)


conradgrobler:
I am unsure whether there is a more idiomatic way to retry the counts in Rust.

rbehjati (approved):
LGTM!

examples/abitest/module_0/rust/tests/integration_test.rs
```diff
@@ -161,6 +156,23 @@ async fn test_leaks() {
...
+                sleep(Duration::from_secs(5)).await;
+                runtime.object_counts()
+            } else {
+                (after_nodes, after_channels)
+            };
+
+        info!(
+            "Counts change from test: Nodes={} => {}, Channels={} => {}",
+            before_nodes, after_nodes, before_channels, after_channels
+        );
```
tiziano88, line 174: I guess if this become a more common pattern then we may want a higher level fn like `retry<T>(f: Fn() -> T) -> Result<T, RetryErr>`.

tiziano88 (approved):

ipetr0v (approved):

PR #2025, opened 2021-04-28
Add loop to oak_main in oak_runtime integration test

conradgrobler:
To address CI flakiness (see #2008). `test_metrics_gives_the_correct_number_of_nodes` sometimes fails because the Wasm node has already exited when getting the metrics. This makes sure that it is around for a short while by looping 100 000 times in the exported main function.

The 100 000 iteration count is a guess, but it seems sufficient to remove the flakiness (at least locally based on running the test 50 times) without slowing down the test too much.

# Checklist

      construction.

### DIFF
--- a/oak_runtime/tests/integration_test.rs
+++ b/oak_runtime/tests/integration_test.rs
@@ -37,10 +37,24 @@ mod common {
     use wat::parse_str;
 
     pub fn start_runtime() -> Result<Arc<Runtime>, OakError> {
+        // Loop 100 000 times in the main function to make sure the Wasm node is alive for a while
+        // before exiting.
         let wat = r#"
         (module
             (type (;0;) (func (param i64)))
-            (func $oak_main (type 0))
+            (func $oak_main (type 0)
+                (local i64)
+                i64.const 100000
+                local.set 1
+                (block
+                    (loop
+                        local.get 1
+                        i64.const -1
+                        i64.add
+                        local.tee 1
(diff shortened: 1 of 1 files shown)


tiziano88:
Alternatively, could this be an infinite loop, and then we figure out how to interrupt it after the test is done? I think at some we will need to figure out how to interrupt a running Wasm module anyways

cc @rbehjati

tiziano88 (approved):

conradgrobler:
> Alternatively, could this be an infinite loop, and then we figure out how to interrupt it after the test is done? I think at some we will need to figure out how to interrupt a running Wasm module anyways

I agree that would be better. Created #2026 and added a TODO.

oak_runtime/tests/integration_test.rs
```diff
@@ -37,10 +37,26 @@ mod common {
     use wat::parse_str;
 
     pub fn start_runtime() -> Result<Arc<Runtime>, OakError> {
+        // Loop 100 000 times in the main function to make sure the Wasm node is alive for a while
+        // before exiting.
```
rbehjati, line 41: Perhaps also explain that this is needed to avoid race conditions with metrics checks.

rbehjati (approved):

oak_runtime/tests/integration_test.rs
conradgrobler: Thanks, done.

PR #2036, opened 2021-04-28
Deduplicate runner structs

tiziano88:
# Checklist

      construction.

### DIFF
--- a/runner/src/examples.rs
+++ b/runner/src/examples.rs
@@ -124,225 +124,6 @@ struct Executable {
     additional_args: Vec<String>,
 }
 
-impl ApplicationClassic {
-    fn construct_application_build_steps(&self, example_name: &str) -> Vec<Step> {
-        vec![
-            Step::Multiple {
-                name: "build wasm modules".to_string(),
-                steps: self
-                    .modules
-                    .iter()
-                    .map(|(name, target)| build_wasm_module(name, target, example_name))
-                    .collect(),
-            },
-            Step::Single {
-                name: "build application".to_string(),
-                command: build_application(&self),
-            },
-        ]
-    }
-
-    fn construct_example_server_run_step(
-        &self,
-        example: &ClassicExample,
-        run_clients: Step,
-    ) -> Step {
-        let opt = &example.options;
-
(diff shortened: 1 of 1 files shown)


tiziano88:
It turns out that the hierarchy nested structs can be flattened out to just two function calls to `run_example` and `run_function_example` with the appropriate argument types (which makes sense since there is only a single dispatch point conceptually: either run a classic example or a functions example, so we should expect a single dispatch based on this bit of information). I think there is still a corner case for examples with no applications (e.g. the tls_attestation one), but that may be sorted out later.

runner/src/examples.rs
```diff
@@ -478,16 +256,73 @@ fn run_example(example: &ClassicExample) -> Step {
     // clients in the foreground.
     #[allow(clippy::collapsible_if)]
     let run_backend_server_clients: Step = if opt.run_server.unwrap_or(true) {
-        let run_server_clients = example.construct_example_server_run_step(run_clients);
-        example.construct_backend_run_steps(run_server_clients)
+        let run_server_clients = match example.applications.get(app_variant) {
```
rbehjati, line 259: `run_example` and `run_functions_example` functions are now too long, too complex, and have too many responsibilities. This reduces readability and makes future modifications to the code very difficult, especially for someone who is new to this part of the code. I suggest refactoring the inlined code into the following functions:
`construct_example_server_run_step`, `construct_backend_run_steps`, `construct_application_build_steps` and similar functions for the `functions` version. 

This would show the overall functionality of `run_example` and `run_functions_example` much better, and would make it clear that these two functions are doing the same thing.

rbehjati (approved):

runner/src/examples.rs
tiziano88: You are right, let me try to move things around some more. Maybe I can go back to roughly the same functions that you had, but just make them separate top-level functions with explicit arguments, instead of methods defined on traits and structs (which resembles too much OOP to me and I find a bit hard to follow in the context of the rest of the Rust code which looks more functional). Will ping you if / when I get anywhere with that :)

PR #2038, opened 2021-04-29
Rename hello world example to weather lookup

### LINKED ISSUE #1897: Weather service with Private Information Retrieval

tiziano88:
Manually create tokio runtime and allow configuring the number of
threads in the config file. I noticed that when running on Cloud Run,
even fetching the lookup data file takes a long time (20s), and I
thought it may be because the runtime is actually running
single-threaded; by specifying the number of threads now the download
and parsing time seems to be down to <1s, which agrees with my theory

Fix #1897

# Checklist

      construction.


rbehjati (approved):
LGTM. Thanks!

oak_functions/examples/weather_lookup/README.md
```diff
@@ -0,0 +1,41 @@
...
+       --manifest-path=./oak_functions/examples/weather_lookup/module/Cargo.toml \
+       --out-dir=./oak_functions/examples/weather_lookup/bin
+   ```
+
+1. Build `oak_functions_loader` (with debugging enabled if needed):
+
+   ```shell
+   cargo build --manifest-path=./oak_functions/loader/Cargo.toml \
+       --release \
+       --features=oak-unsafe
```
ipetr0v, line 26: Why do we need `unsafe`?
Is it to work with Wasm modules?

tiziano88: Just to have some logging I guess, it was already there.

rbehjati: Yes. It is for logging. We have to removed it for the prod version, I think.

PR #2039, opened 2021-04-29
Add `write_log_message` ABI function

conradgrobler:
A follow-up PR will add support to the SDK to allow examples to use the `log` create to log.

# Checklist

- [x] Pull request affects core Oak Functions functionality (e.g. runtime, SDK, ABI)
      construction.


conradgrobler:
The proto only has 2 fields, and the level is an i32. I am undecided on whether to add the level as a parameter to the ABI function and remove the proto so that the bytes can be interpreted as a utf8 string.

The current implementation is in line with the original Oak log pseudo node in terms of using a proto to wrap the log message, but the simpler structure of the log message makes it feel like it is overkill to use a proto to represent it.

Any thoughts @tiziano88, @rbehjati, @ipetr0v?

tiziano88:
I would just let the caller log the message as a UTF8 string, not even support log levels. I think for the kind of functions we have, the logs are going to be pretty predictable anyways, and especially if we send them back to the caller, I think having extra logic to filter by level is overkill.

rbehjati:
Is the idea to send to the caller all the logs or only the ones generated by the Wasm module?

tiziano88:
Just the Wasm ones.

conradgrobler:
> Is the idea to send to the caller all the logs or only the ones generated by the Wasm module?

The idea is to collect the logs sent by the Wasm module and send it back to the caller if the caller adds a special header to the request. That will be part of a future PR though. For now it will just log to stderr.

docs/oak_functions_abi.md
```diff
@@ -97,6 +97,25 @@ client.
...
+### `write_log_message`
+
+Writes a log message. This message will be considered sensitive, so will only be
+logged if the `oak_unsafe` feature is enabled.
+
+The low-level operation involves reading the response from the WebAssembly
+module's memory. The system will attempt to interpret the bytes as a UTF-8
+encoded string. If the decoding is successful, the string will be logged as a
+debug message. If the bytes are not a valid UTF-8 string a warning message
+containing the UTF-8 decoding error and the raw bytes will be logged.
```
tiziano88, line 112: (nit) perhaps stick to present tense throughout instead of switching between present and future.

oak_functions/loader/src/server.rs
```diff
@@ -248,6 +250,39 @@ impl WasmState {
...
+        buf_len: AbiPointerOffset,
+    ) -> Result<(), OakStatus> {
+        let raw_log = self
+            .get_memory()
+            .get(buf_ptr, buf_len as usize)
+            .map_err(|err| {
+                self.logger.log_sensitive(
+                    Level::Error,
+                    &format!(
+                        "write_log_message(): Unable to read name from guest memory: {:?}",
```
tiziano88, line 266:
```suggestion
                        "write_log_message(): Unable to read message from guest memory: {:?}",
```

tiziano88 (approved):

docs/oak_functions_abi.md
conradgrobler: Done.

oak_functions/loader/src/server.rs
conradgrobler: Done.

Issue #1992, opened 2021-04-21
Handle panics in Oak Functions loader


tiziano88:
We should ensure that no matter what happens while executing the Wasm module, there is no way to make the server (in production mode) panic, or produce any other macroscopically observable effect.

Perhaps we should also set up fuzzing, though I am not sure whether fuzzing the Wasm bytecode would be a productive way of doing this. We could also fuzz user requests, given a fixed Wasm module.

cc @alastairreid 

_Originally posted by @conradgrobler in https://github.com/project-oak/oak/pull/1991#discussion_r617271100_

alastairreid:
One of my takeaways from the Rust-for-Linux RFC discussion on LKML is that trading liveness for safety is not ideal for a server/kernel. That is, panicking when something bad happens (eg array bounds check fails) prevents something bad happening but it may also cause loss of data, loss of logging (eg an attacker might quickly perform an attack and then crash the kernel before logging info can be saved), denial of service, and similar which are maybe just as bad as the original problem.

So, yes, really important to try to eliminate panics from the entire Rust codebase. A combination of fuzzing and our formal verification tools is the answer.

wrt Bytecode...
The thing not to do is generate random files of bytes and call it good. You will find lots of ways to create malformed bytecode but you will only be getting coverage over the bytecode loader.
You need multiple layers of fuzzing - each layer bypassing the issues that the previous layers have already checked.

1. Random byte files - to check for errors in the bytecode loader
2. Correctly structured wasm files - with type errors, overly large stack offsets as immediate fields, illegal opcodes, etc.
3. Well formed wasm that probes memory boundaries, etc.
4. Random sequences of commands. eg if the Oak API provides functions F(_,_), G(_) and H(), then you generate random sequences like [("F", [42, 87]), ("H", []), ("F", [234231,-5]), ("H", [])] and supply that to a Wasm program that loops over the list applying the named function to the list of arguments.   (Variations on this approach exist - but that's the basic idea)

(Did a variation of (4) on a microcontroller OS before - found some issues.)

PR #2043, opened 2021-04-29
Use a custom response in Oak Functions

### LINKED ISSUE #2032: Use custom Request and Response body for Oak Functions

_Originally posted by @tiziano88 in https://github.com/project-oak/oak/pull/2011#discussion_r622233572_

rbehjati:
Ref #2032


oak_functions/proto/response.proto
```diff
@@ -0,0 +1,42 @@
...
+
+// Status codes for the response.
+enum FunctionsStatusCode {
+  UNSPECIFIED = 0;
+  // Indicates success of the operation. Similar to HTTP 200 status code.
+  SUCCESS = 1;
+  // Indicates a problem with the request. Similar to HTTP 400 status code.
+  BAD_REQUEST = 2;
+  // Indicates violation of the security policy.
+  POLICY_VIOLATION = 3;
```
tiziano88, line 39: I think we want more precise errors for policy, right? Surely if we are querying a server and it fails because of policy, we would want to know if it's because of time or size.

oak_functions/proto/response.proto
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak.functions.response;
+
+// Represents the response from an Oak Functions application. This is included in the body of the
+// HTTP response sent to the client.
+message FunctionsResponse {
```
tiziano88, line 23: These objects have names that repeat parts of the package name. Also I think we would want to also have a corresponding Request message at some point (e.g. to combine body and policy). So how about fully qualified name of `oak.functions.invocation.{Response,StatusCode}` ?

oak_functions/examples/weather_lookup/module/src/tests.rs
```diff
@@ -71,25 +73,31 @@ async fn test_server() {
     {
         // Lookup match.
         let response_fut = make_request(server_port, br#"{"lat":52,"lon":0}"#).await;
```
tiziano88, line 75: BTW I don't think this is an actual future, it's already resolved by the await.

oak_functions/loader/src/server.rs
```diff
@@ -87,69 +86,58 @@ impl Policy {
...
+    /// violation is returned. If `self.constant_response_size_bytes` is too small even for the
+    /// policy violation response, or the response already indicates a policy violation, an error is
+    /// returned.
+    fn create_http_response(&self, response: FunctionsResponse) -> anyhow::Result<Response<Body>> {
+        // deserialize response and get its size.
+        let mut encoded = vec![];
+        response
+            .encode(&mut encoded)
+            .context("couldn't encode response")?;
+        let padding_length = self.constant_response_size_bytes as i64 - encoded.len() as i64;
```
tiziano88, line 106: I am a bit worried about overflow and underflow here, would it be possible to use checked conversions and subtraction?

oak_functions/loader/src/server.rs
```diff
@@ -87,69 +86,58 @@ impl Policy {
...
+        if padding_length < 0 {
+            match FunctionsStatusCode::from_i32(response.status) {
+                None => anyhow::bail!("invalid status code"),
+                Some(FunctionsStatusCode::PolicyViolation) => anyhow::bail!(
+                    "constant_response_size_bytes specified by the policy is too small."
+                ),
+                _ => self.create_http_response(FunctionsResponse {
```
tiziano88, line 113: Why is this even necessary? If the response is too large (i.e. padding < 0) then just always return the policy violation message, no?

edit: I just realised that this may be because of the recursive invocation? If so, it should definitely have a clear comment explaining the recursion, this logic, and convincing a reader that it does terminate. Alternatively, I think it would be better to avoid the recursion altogether if possible.

oak_functions/loader/src/server.rs
```diff
@@ -87,69 +86,58 @@ impl Policy {
...
+    /// To conform to the security policy, the response is padded by adding a number of unspecified
+    /// characters, to make the total size of the response equal to
+    /// `self.constant_response_size_bytes`. The padding is added as the value of the
+    /// `oak-policy-padding` header. If the size of the input `response` is larger than
+    /// `self.constant_response_size_bytes` and the response is not already indicating policy
+    /// violation, then the `response` is discarded and an HTTP response indicating the policy
+    /// violation is returned. If `self.constant_response_size_bytes` is too small even for the
+    /// policy violation response, or the response already indicates a policy violation, an error is
+    /// returned.
+    fn create_http_response(&self, response: FunctionsResponse) -> anyhow::Result<Response<Body>> {
```
tiziano88, line 100: Why does this return a Result? I think it should return an actual response in all cases, since even not returning a response would be a policy violation. If there is any error, it should be returned in the response.

As a side note, this function would be a great candidate for fuzzing: pass different responses as input, and check that it does not panic, and that it always returns a response of the same exact size.

#1992

oak_functions/loader/src/server.rs
```diff
@@ -87,69 +86,58 @@ impl Policy {
...
+                    body: "Reason: the response is too long".as_bytes().to_vec(),
+                }),
+            }
         } else {
-            let padding_bytes = vec![b'0'; required_padding as usize];
+            let padding_bytes = vec![b'0'; padding_length as usize];
             http::response::Builder::new()
-                .status(response.status_code)
+                .status(StatusCode::OK)
                 .header(POLICY_PADDING_HEADER, std::str::from_utf8(&padding_bytes)?)
```
tiziano88, line 122: BTW does this work correctly in case of zero bytes? I'm worried about whether there is an extra whitespace between the name and the value of the header that only gets added if the value is non-empty. Could you double check? (And also write it in a comment for future reference).

oak_functions/loader/src/server.rs
tiziano88, line 100: Minor point and somewhat personal preference, but I am not a big fan of defining methods with self receiver on things that are essentially plain data objects, such as here (basically things that get parsed from JSON / protobuf, or similar). Personally I would expect a top-level `fn create_http_response(response: FunctionsResponse, policy: Policy)` (more in line with functional idioms than OOP; i.e. it's not like the policy is creating the response). Feel free to ignore anyways.

oak_functions/loader/src/server.rs
```diff
@@ -468,72 +456,30 @@ where
...
-                                Ordering::Equal => Response::<Body>::try_from(rsp)
-                                    .context("couldn't get the response"),
-                            }
-                        }
+                        // Error generated by `function`
+                        Err(err) => policy.create_http_response(FunctionsResponse {
+                            status: FunctionsStatusCode::InternalServerError as i32,
+                            body: format!("{}", err).as_bytes().to_vec(),
+                        }),
+                        Ok(rsp) => policy.create_http_response(rsp),
```
tiziano88, line 475: Why is `create_http_response` applied within each branch of this match? It seems it should just applied once around it.

rbehjati: Fixed.

oak_functions/proto/response.proto
rbehjati: I am not sure about `invocation`. To me, it sounds like a very oak-specific concept based on the concept of channels. I've removed `Functions` from the names.

oak_functions/proto/response.proto
rbehjati: Done.

rbehjati (commented):
Thanks for the review @tiziano88.

oak_functions/proto/response.proto
tiziano88: Oh, I just meant it as "function invocation" anyways, that's why the endpoint is called `/invoke` already. This is fine anyways.

oak_functions/loader/src/server.rs
```diff
@@ -88,70 +86,91 @@ impl Policy {
...
-    type Error = http::Error;
-
-    fn try_from(rsp: FunctionsResponse) -> Result<Self, Self::Error> {
-        Response::builder()
-            .status(rsp.status_code)
-            .body(rsp.body.map(Body::from).unwrap_or_else(Body::empty))
-    }
+/// Creates an HTTP response, with status `OK` and the given `response` as the response body.
+///
+/// Encodes the given `response` as protobuf, and sets that as the body of the response.
```
tiziano88, line 164:
```suggestion
/// Encodes the given `response` as binary protobuf, and sets that as the body of the response.
```

oak_functions/loader/src/server.rs
tiziano88: Nice, thanks. BTW this may be a small unit test that proves in code what you just wrote here, so that it keeps being checked.

tiziano88 (approved):

oak_functions/loader/src/server.rs
```diff
@@ -88,70 +86,91 @@ impl Policy {
...
+        let padding_length = usize::try_from(padding_length).unwrap();
+        let padding_bytes = vec![b'0'; padding_length];
+
+        // If `padding_length` is zero, the `oak-policy-padding` header is still added, but with
+        // an empty string as the header value. The final encoding of the header in this case is
+        // `oak-policy-padding: ` including the whitespace at the end. This guarantees that the
+        // total size of the transmitted response remains constant.
+        http::response::Builder::new()
+            .status(hyper::StatusCode::OK)
+            .header(POLICY_PADDING_HEADER, std::str::from_utf8(&padding_bytes)?)
```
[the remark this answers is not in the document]
conradgrobler: Ah, I misread the code. I think a less ambiguous character would be good. :)

oak_functions/loader/src/server.rs
```diff
@@ -88,70 +86,91 @@ impl Policy {
...
+            _ => Response {
+                status: StatusCode::PolicySizeViolation as i32,
+                body: "Reason: the response is too long".as_bytes().to_vec(),
+            },
+        };
+        // The match arms above guarantee that a termination condition will be met after a finite
+        // number of recursive calls.
+        create_response_and_apply_policy(rsp, policy)
+    } else {
+        let padding_length = usize::try_from(padding_length).unwrap();
```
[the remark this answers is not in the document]
rbehjati: Changed it to return `anyhow::Result`. Also, changed the type of `padding_length` to `i128`. It should always be possible to safely convert a non-negative `i128` to `usize`, I think.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
rbehjati: Done.

oak_functions/loader/src/server.rs
```diff
@@ -490,88 +509,50 @@ fn check_export_function_signature(
...
 
-            let response = task.now_or_never();
+            let function_response = task.now_or_never();
 
-            match response {
-                None => policy.create_error_response("response not available"),
+            let response = match function_response {
+                None => Response {
+                    status: StatusCode::PolicyTimeViolation as i32,
+                    body: "Reason: response not available".as_bytes().to_vec(),
```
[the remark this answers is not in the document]
rbehjati: Ideally we should be able to tell these two cases apart, if we manage to get tokio parallelism to work! If the function terminated without a response, but still terminated before the policy timeout, then `function_response` would not be `None`. It would contain an error, which is handled in the other match arm.

oak_functions/loader/src/server.rs
```diff
@@ -490,88 +509,50 @@ fn check_export_function_signature(
...
+                },
                 Some(response) => match response {
-                    Err(err) => policy.create_error_response(format!("{}", err).as_str()),
+                    // Error when getting the response from the tokio task
+                    Err(err) => Response {
+                        status: StatusCode::InternalServerError as i32,
+                        body: format!("{}", err).as_bytes().to_vec(),
```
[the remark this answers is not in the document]
rbehjati: Yes, it is `tokio::task::JoinError`. Changed it to a fixed generic response. But I am also not sure what is a good error message here.

rbehjati (commented):
Thanks for the comments.

oak_functions/loader/src/server.rs
rbehjati: `http::response` and `hyper` do not provide a utility for serializing the response into the final transmitted byte array. Moreover, the none of the client libraries seem to provide the raw response received as bytes. So, the test has to start a server and use curl as the client. Can you think of an easier way of testing this?

Actually, we could have done this more easily as part of a bash client for the hello-world example, since the size of the response can be controlled by the request. In general, it would have been easier to test policies with such an example. I wonder if I should add a simple `policy` example. WDYT?

tiziano88: Maybe we should move away from using the HTTP headers as padding (in another PR), and instead have a type that we serialize and pad at the same time before handing it over to HTTP or gRPC or anything else that may handle it. The "contract" may be that the response is of a fixed size, but the expectation is that the receiver will encrypt it before sending it out. Maybe we can do something like defining a trait for serializing and padding (at the same time) a Request object, something like 

```rust
trait FixedSizeSerializer {
  fn serialize(v: Value, size: usize) -> Vec<u8>
}
```

oak_functions/loader/src/server.rs
```diff
@@ -88,70 +86,91 @@ impl Policy {
...
+    // check that the policy is valid
+    policy.validate()?;
+
+    // Serialize the response and get its size.
+    let mut encoded = vec![];
+    response
+        .encode(&mut encoded)
+        .context("couldn't encode response")?;
+    let padding_length =
+        i64::from(policy.constant_response_size_bytes) - i64::try_from(encoded.len()).unwrap();
```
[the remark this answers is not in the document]
tiziano88: I think it gets encoded by hyper with HTTP1 because it only supports text, while HTTP2 supports binary. In any case it should not matter, the point is that it has a constant size, not that the size is exactly that specified in the policy. As long as it does not change based on the content, it is fine if the final size is a deterministic function of the "fixed size" value. Though this should probably be clarified.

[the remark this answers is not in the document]
conradgrobler: Just to clarify: are you saying that the existing implementation will provide a fixed size with HTTP1?

I don't believe it will. If the body is base64 encoded and the header is not, adding 1 byte to the header will not correctly offset  1 byte less in the body. Even if both are base64 encoded I don't think it will always be exact due to the 4 byte chunks that base64 encoding grow in.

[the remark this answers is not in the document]
tiziano88: @conradgrobler you are right, I stand corrected. The encoding only applies to the body, and base64 encoding is not a linear operation, e.g. individually base64 encoding two separate bytes and concatenating the output is not the same as base64 encoding the two concatenated bytes. Thanks for pointing out.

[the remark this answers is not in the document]
conradgrobler: I think the simplest solution is to apply the size limit policy only on the size of the body bytes. It is easy to implement and easy to explain in documentation.

Now that we are always sending an OK status code to the client, the headers should not depend on the Wasm code at all. So this should result in a constant response size for all HTTP1.1 responses and a different constant response size for all the HTTP2 responses.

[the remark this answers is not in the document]
conradgrobler: To be clear, I mean the limit should be applied on the raw body bytes, not the base64-encoded bytes in the final HTTP 1.1 response.

oak_functions/loader/src/server.rs
rbehjati: Is the output byte vector the binary protobuf encoding of the padded value? If it is not a valid protobuf message, then we have to define a protocol for serializing and deserializing. I think it is easier to just rely on protobuf.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
rbehjati: Thanks. I'll fix this in a separate PR.

oak_functions/loader/src/server.rs
conradgrobler: I think that trailing 0 bytes in a byte array would not be a problem for decoding protos. So, I think it could be done by just initialising a fixed length 0-initialised byte array and passing it the to `prost::Message::encode`. It should encode the proto at the start and leave the rest as 0s. This entire fixed length byte array could then be sent in the body, and the client should be able to decode it correctly.

conradgrobler (approved):

oak_functions/loader/src/server.rs
rbehjati: Thanks. I'll try that.

rbehjati: I get the following error when decoding protobuf with trailing 0s:
```
DecodeError { description: "invalid tag value: 0", stack: [] }
```

rbehjati: I think using `encode_length_delimited` and `decode_length_delimited` can fix this.

conradgrobler: That is a pity. I don't think length-delimited encoding is standardised yet, or compatible with all protobuf libraries. See e.g. https://seb-nyberg.medium.com/length-delimited-protobuf-streams-a39ebc4a4565

rbehjati: We are already using length-delimited encoding for the lookup data. But that is not client-facing. 
The other alternative is to include a padding field in the response body. So, we'll have:
```proto
message Response {
  StatusCode status = 1;
  bytes body = 2;
  bytes padding = 3;
}
```
WDYT?

PR #2044, opened 2021-04-29
Update run-cargo-tests to run benchmarks too

### LINKED ISSUE #1933: Improve benchmarking for oak functions loader

rbehjati:
Fixes #1933

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,6 +28,7 @@ jobs:
           - run-cargo-tests --cleanup
           - run-bazel-tests
           - run-tests-tsan
+          - run-bench --cleanup
           - run-examples --application-variant=rust
           - run-examples --application-variant=cpp
           - run-examples --application-variant=rust --example-name=abitest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,10 +25,9 @@ jobs:
           - build-server --server-variant=experimental
           - build-functions-server --server-variant=unsafe
           - build-functions-server --server-variant=base
-          - run-cargo-tests --cleanup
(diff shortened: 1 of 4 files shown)


tiziano88 (commented):
Wouldn't it be much faster to just run benchmarks after (or together with) tests? Since basically everything would be already compiled in that case.

tiziano88:
BTW it may be enough to call `cargo test --benches`

rbehjati:
> BTW it may be enough to call `cargo test --benches`

Thanks. Changed `run-cargo-tests` to use this.

tiziano88 (approved):

rbehjati:
@tiziano88 The name of the `run-cargo-tests` is changed again! Could you please make `runner (run-cargo-tests --cleanup)` an unrequired job, and later add `runner (run-cargo-tests --cleanup --benches)`?
Is there a way to exclude options from the name of the job? Or maybe use a regex or something for specifying the required jobs? (I have no idea how the specification of the required jobs is done)

tiziano88:
Done, it's literally just a list of strings unfortunately, but that's because we are using the command itself as the step names, in principle we could do something fancier in our CI config, not sure it's worth it though

rbehjati:
> Done, it's literally just a list of strings unfortunately, but that's because we are using the command itself as the step names, in principle we could do something fancier in our CI config, not sure it's worth it though

I see. Thanks. Maybe we could make it fancier in the future :)

PR #2049, opened 2021-04-30
Add `write` permission on issues to reproducibility index workflow

conradgrobler:
This is to allow the workflow to add a comment to the current PR discussion.

See https://github.com/project-oak/oak/pull/2004#discussion_r623840132

# Checklist

      construction.

### DIFF
--- a/.github/workflows/reproducibility.yaml
+++ b/.github/workflows/reproducibility.yaml
@@ -12,6 +12,7 @@ jobs:
 
     permissions:
       contents: write
+      issues: write
 
     steps:
       - name: Checkout branch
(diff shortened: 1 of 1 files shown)


.github/workflows/reproducibility.yaml
```diff
@@ -12,6 +12,7 @@ jobs:
 
     permissions:
       contents: write
+      issues: write
```
tiziano88, line 15: Maybe worth a comment?

tiziano88 (approved):

rbehjati (approved):
Thanks.

.github/workflows/reproducibility.yaml
conradgrobler: Done.

conradgrobler:
It seems that this was not sufficient to resolve the issue. I will look into it further. Perhaps it also requires write permissions on the PR or something like that.

rbehjati:
Yes. Perhaps we need `pull-requests: write`.

PR #2057, opened 2021-05-04
Include policy-padding in the response body

rbehjati:
Follow-up of #2043

* The constant size specified by the policy now applies to the body of the Oak Functions response (i.e., `oak_functions_abi::proto::Response`). This also guarantees that the binary protobuf-encoded message has a fixed size, and the HTTP response sent to the client has a fixed size too.
* The Oak Functions response body is padded with trailing 0s to match the size specified by the policy.
* This change also makes policy non-optional throughout.


oak_functions/loader/src/server.rs
```diff
@@ -88,97 +88,85 @@ impl Policy {
     }
 }
 
+/// Enum indicating a serialization error
+enum SerializationError {
+    ByteArrayTooLarge,
+    FailedEncoding(prost::EncodeError),
```
conradgrobler, line 94: I don't think this will actually ever happen. The error indicates that the buffer was too small, but using a mutable Vec as the buffer means that it will be resized as needed. https://docs.rs/prost/0.7.0/prost/struct.EncodeError.html

conradgrobler (approved):

oak_functions/loader/src/server.rs
rbehjati: Thanks. Removed this enum.

Issue #2047, opened 2021-04-30
Fuzz testing for oak_functions


rbehjati:
Add fuzz testing for oak_functions.

See also:
- #1992 
- #1163
- https://github.com/project-oak/oak/pull/2043#discussion_r623281377

Re-enable OSS-Fuzz for project-oak. See:
- https://github.com/google/oss-fuzz/pull/4068
- https://google.github.io/oss-fuzz/getting-started/new-project-guide/rust-lang/

PR #2061, opened 2021-05-06
Add a simple fuzzing target for oak_functions loader

rbehjati:
Adds a simple fuzz target for `oak_functions/loader`. Allows re-enabling oss-fuzz for project oak.

Ref #2047


rbehjati:
Also created https://github.com/google/oss-fuzz/pull/5719

oak_functions/loader/fuzz/fuzz_targets/create_response_and_apply_policy.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        let response = Response {
+            status,
+            body,
+            length,
+        };
+
+        let policy = Policy {
+            constant_response_size_bytes: raw.int_in_range(50..=5000)?,
+            constant_processing_time: std::time::Duration::from_millis(10),
```
tiziano88, line 45: Maybe also randomize this at some point?

tiziano88 (approved):
Awesome! How to actually run this? Or does it get run by runner already? Maybe add a short README somewhere?

rbehjati:
> Awesome! How to actually run this? Or does it get run by runner already? Maybe add a short README somewhere?

I am trying to re-enable oss-fuzz (google/oss-fuzz#5719). That should be the main way to run fuzzing. 
I think it would be nice to have a command in the runner, and perhaps even include it in the CI with a limited number of runs (using the `-runs` option). 
But my current problem is that running `cargo fuzz coverage` fails in the docker. I think there is an LLVM problem. I am not sure how to upgrade to a newer version of LLVM. I tried with a newer debian image, but it did not fix the problem. But perhaps for the runner command we wouldn't need to run coverage. 

I am also looking into ways to make fuzzing faster. For example, adding a seed or an initial corpus could help.

It is a good idea to have README. I'll add one to the `docs` folder in a separate PR.

PR #2065, opened 2021-05-10
Add new runner command for running fuzz targets

rbehjati:
Ref #2047

### DIFF
--- a/runner/src/internal.rs
+++ b/runner/src/internal.rs
@@ -20,6 +20,7 @@ use colored::*;
 use nix::sys::signal::Signal;
 use std::{
     collections::{HashMap, HashSet},
+    path::PathBuf,
     time::Instant,
 };
 use structopt::StructOpt;
@@ -51,6 +52,7 @@ pub enum Command {
     RunCargoTests(RunTestsOpt),
     RunBazelTests,
     RunTestsTsan,
+    RunFuzzTargets(RunFuzzTargets),
     RunCargoDeny,
     RunCargoUdeps,
     RunCi,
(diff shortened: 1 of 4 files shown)


runner/src/internal.rs
```diff
@@ -293,6 +295,41 @@ impl RustBinaryOptions for BuildServer {
...
+    #[serde(default)]
+    pub bin: Vec<CargoBinary>,
+}
+
+/// Partial information about a Cargo binary, as included in a Cargo manifest.
+#[derive(serde::Deserialize, Debug)]
+pub struct CargoBinary {
+    #[serde(default)]
+    pub name: String,
+}
```
tiziano88, line 331: Would https://crates.io/crates/cargo_metadata work here? (we already use it elsewhere: https://github.com/project-oak/oak/blob/9f0ac739e47c6b713b3fd2838ad0ce4738192761/sdk/rust/oak_tests/src/lib.rs#L50-L53)

runner/src/main.rs
```diff
@@ -176,6 +177,67 @@ fn run_tests_tsan() -> Step {
...
+    }
+}
+
+pub fn run_fuzz_targets_in_crate(
+    path: &PathBuf,
+    cargo_manifest: &CargoManifest,
+    opt: &RunFuzzTargets,
+) -> Step {
+    // `cargo-fuzz` can only run in the crate that contains the `fuzz` crate. So we need to use
+    // `Cmd::new_in_dir` to execute the command inside the crate's directory.
```
tiziano88, line 215: But why is the code popping two elements from the path?

tiziano88 (approved):

runner/src/internal.rs
rbehjati: This looks interesting, but I could not find any structs in `cargo_metadata` that matches the structure of a Cargo toml file. I particularly tried `Metadata`, and `Package`.

runner/src/main.rs
rbehjati: Added an explanation.

PR #2070, opened 2021-05-11
Add remote attestation via gRPC

ipetr0v:
This change adds an experimental version of gRPC remote attestation scheme.
This attestation scheme uses:
- [`ring::agreement`](https://briansmith.org/rustdoc/ring/agreement/index.html) to exchange X25519 Diffie-Hellman keys
- [`ring::aead`](https://briansmith.org/rustdoc/ring/aead/index.html) to encrypt and authenticate messages with AES-GCM

### DIFF
--- a/examples/grpc_attestation/example.toml
+++ b/examples/grpc_attestation/example.toml
@@ -1,5 +1,7 @@
 name = "grpc_attestation"
 
+# Currently backend sends responses to the client, so there is no need for Oak application.
+# TODO(#1932): Use gRPC Attestation for Weather Example in Oak Functions.
 [applications]
 
 [backends]
--- a/examples/grpc_attestation/example.toml
+++ b/examples/grpc_attestation/example.toml
@@ -1,0 +1,14 @@
+name = "grpc_attestation"
+
+[applications]
+
+[backends]
+backend = { Cargo = { cargo_manifest = "experimental/grpc_attestation/Cargo.toml" }, additional_args = [
+  "--grpc-listen-address=[::]:8080",
+  "--tee-certificate=./examples/certs/local/ca.pem",
+] }
+
+[clients]
+rust = { Cargo = { cargo_manifest = "examples/grpc_attestation/client/Cargo.toml" }, additional_args = [
+  "--uri=https://localhost:8080"
+] }
--- a/experimental/attestation_common/src/crypto.rs
+++ b/experimental/attestation_common/src/crypto.rs
@@ -1,0 +1,155 @@
+//
+// Copyright 2021 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use anyhow::anyhow;
+use ring::{
+    aead::{self, BoundKey},
+    agreement, rand,
+};
+
+// `ring::aead` uses 96-bit (12-byte) nonces.
+// https://briansmith.org/rustdoc/ring/aead/constant.NONCE_LEN.html
+//
+// TODO(#2087): Use a unique nonce for each message.
+const NONCE: [u8; aead::NONCE_LEN] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+// Algorithm used for encrypting/decrypting messages.
+// https://datatracker.ietf.org/doc/html/rfc5288
+static AEAD_ALGORITHM: &aead::Algorithm = &aead::AES_256_GCM;
+// Algorithm used for negotiating a session key.
+// https://datatracker.ietf.org/doc/html/rfc7748
+static KEY_AGREEMENT_ALGORITHM: &agreement::Algorithm = &agreement::X25519;
+
+/// Nonce implementation used by [`AeadInterface`].
+/// It returns a single nonce once and then only returns errors.
+struct OneNonceSequence(Option<aead::Nonce>);
+
+impl OneNonceSequence {
+    fn new() -> Self {
+        let nonce = aead::Nonce::assume_unique_for_key(NONCE);
+        Self(Some(nonce))
+    }
+}
+
+impl aead::NonceSequence for OneNonceSequence {
+    fn advance(&mut self) -> Result<aead::Nonce, ring::error::Unspecified> {
+        self.0.take().ok_or(ring::error::Unspecified)
+    }
+}
+
+/// Implementation of Authenticated Encryption with Associated Data (AEAD).
+/// https://datatracker.ietf.org/doc/html/rfc5116
+pub struct AeadEncryptor {
+    // Key used for encrypting data.
+    sealing_key: aead::SealingKey<OneNonceSequence>,
+    // Key used for decrypting data.
+    opening_key: aead::OpeningKey<OneNonceSequence>,
+}
+
+impl AeadEncryptor {
+    pub fn new(key: &[u8]) -> Self {
+        // AES-GCM algorithm uses the same key is used for both encryption and decryption.
+        // https://datatracker.ietf.org/doc/html/rfc5288
+        let unbound_sealing_key = aead::UnboundKey::new(AEAD_ALGORITHM, &key).unwrap();
+        let sealing_key = ring::aead::SealingKey::new(unbound_sealing_key, OneNonceSequence::new());
+
+        let unbound_opening_key = aead::UnboundKey::new(AEAD_ALGORITHM, &key).unwrap();
+        let opening_key = ring::aead::OpeningKey::new(unbound_opening_key, OneNonceSequence::new());
+
+        Self {
+            sealing_key,
+            opening_key,
+        }
+    }
+
+    /// Encrypts `data` using [`AeadEncryptor::sealing_key`].
+    ///
+    /// Additional authenticated data is not required for the remotely attested channel,
+    /// since after session key is established client and server exchange messages with a
+    /// single encrypted field.
+    /// And the nonce is authenticated by the AEAD algorithm itself.
+    /// https://datatracker.ietf.org/doc/html/rfc5116#section-2.1
+    pub fn encrypt(&mut self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let mut encrypted_data = data.to_vec();
+        self.sealing_key
+            .seal_in_place_append_tag(
+                aead::Aad::from(vec![]),
+                &mut encrypted_data,
+            )
+            .map_err(|error| anyhow!("Couldn't encrypt data: {:?}", error))?;
+        Ok(encrypted_data)
+    }
+
+    /// Decrypts and authenticates `data` using [`AeadEncryptor::opening_key`].
+    ///
+    /// Additional authenticated data is not required for the remotely attested channel,
+    /// since after session key is established client and server exchange messages with a
+    /// single encrypted field.
+    /// And the nonce is authenticated by the AEAD algorithm itself.
+    /// https://datatracker.ietf.org/doc/html/rfc5116#section-2.1
+    pub fn decrypt(&mut self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let mut decrypted_data = data.to_vec();
+        let decrypted_data = self
+            .opening_key
+            .open_in_place(
+                aead::Aad::from(vec![]),
+                &mut decrypted_data,
+            )
+            .map_err(|error| anyhow!("Couldn't decrypt data: {:?}", error))?;
+        Ok(decrypted_data.to_vec())
+    }
+}
+
+pub struct KeyNegotiator {
+    private_key: agreement::EphemeralPrivateKey,
+}
+
+impl KeyNegotiator {
+    pub fn create() -> anyhow::Result<Self> {
+        let rng = rand::SystemRandom::new();
+        let private_key = agreement::EphemeralPrivateKey::generate(KEY_AGREEMENT_ALGORITHM, &rng)
+            .map_err(|error| anyhow!("Couldn't generate private key: {:?}", error))?;
+        Ok(Self { private_key })
+    }
+
+    pub fn public_key(&self) -> anyhow::Result<Vec<u8>> {
+        let public_key = self
+            .private_key
+            .compute_public_key()
+            .map_err(|error| anyhow!("Couldn't get public key: {:?}", error))?;
+        Ok(public_key.as_ref().to_vec())
+    }
+
+    pub fn derive_session_key(self, peer_public_key: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let peer_public_key =
+            agreement::UnparsedPublicKey::new(KEY_AGREEMENT_ALGORITHM, peer_public_key);
+        let session_key = agreement::agree_ephemeral(
+            self.private_key,
+            &peer_public_key,
+            ring::error::Unspecified,
+            Self::kdf,
+        )
+        .map_err(|error| anyhow!("Couldn't derive session key: {:?}", error))?;
+        Ok(session_key)
+    }
+
+    /// Derives a session key from `key_material`.
+    fn kdf(key_material: &[u8]) -> Result<Vec<u8>, ring::error::Unspecified> {
+        // TODO(#2100): Derive a session key not just from `key_material` but also from server's and
+        // client's public keys.
+        // https://datatracker.ietf.org/doc/html/rfc7748#section-6.1
+        Ok(key_material.to_vec())
+    }
+}
--- a/experimental/attestation_common/src/crypto.rs
+++ b/experimental/attestation_common/src/crypto.rs
@@ -32,7 +32,7 @@ static AEAD_ALGORITHM: &aead::Algorithm = &aead::AES_256_GCM;
 // https://datatracker.ietf.org/doc/html/rfc7748
 static KEY_AGREEMENT_ALGORITHM: &agreement::Algorithm = &agreement::X25519;
 
-/// Nonce implementation used by [`AeadInterface`].
+/// Nonce implementation used by [`AeadEncryptor`].
 /// It returns a single nonce once and then only returns errors.
 struct OneNonceSequence(Option<aead::Nonce>);
 
@@ -84,10 +84,7 @@ impl AeadEncryptor {
     pub fn encrypt(&mut self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
         let mut encrypted_data = data.to_vec();
         self.sealing_key
-            .seal_in_place_append_tag(
-                aead::Aad::from(vec![]),
-                &mut encrypted_data,
-            )
+            .seal_in_place_append_tag(aead::Aad::from(vec![]), &mut encrypted_data)
             .map_err(|error| anyhow!("Couldn't encrypt data: {:?}", error))?;
         Ok(encrypted_data)
     }
@@ -103,10 +100,7 @@ impl AeadEncryptor {
         let mut decrypted_data = data.to_vec();
         let decrypted_data = self
             .opening_key
-            .open_in_place(
-                aead::Aad::from(vec![]),
-                &mut decrypted_data,
-            )
+            .open_in_place(aead::Aad::from(vec![]), &mut decrypted_data)
             .map_err(|error| anyhow!("Couldn't decrypt data: {:?}", error))?;
         Ok(decrypted_data.to_vec())
     }
--- a/experimental/attestation_common/src/crypto.rs
+++ b/experimental/attestation_common/src/crypto.rs
@@ -60,7 +60,7 @@ pub struct AeadEncryptor {
 
 impl AeadEncryptor {
     pub fn new(key: &[u8]) -> Self {
-        // AES-GCM algorithm uses the same key is used for both encryption and decryption.
+        // AES-GCM algorithm uses the same key for both encryption and decryption.
         // https://datatracker.ietf.org/doc/html/rfc5288
         let unbound_sealing_key = aead::UnboundKey::new(AEAD_ALGORITHM, &key).unwrap();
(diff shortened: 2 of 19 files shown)


experimental/grpc_attestation/src/attestation.rs
```diff
@@ -0,0 +1,210 @@
...
+            .context("Stream stopped preemptively")?;
+
+        // Receive client's attestation request containing a public key.
+        let request_type = request.request_type.context("Couldn't read request type")?;
+        let attestation_request =
+            if let RequestType::AttestationRequest(attestation_request) = request_type {
+                Ok(attestation_request)
+            } else {
+                Err(anyhow!("Received incorrect message type"))
+            }?;
```
tiziano88, line 139: It would be nice to model the state transitions more cleanly with Rust types at some point. For instance, have a trait for an object that receives messages from a `Receiver`, and returns a new instance of an object that impl that trait, or an error; then we can have one object for each state in the state transition diagram, that consumes itself by value and produces the next step (or an error). I think https://hoverbear.org/blog/rust-state-machine-pattern/ has a good overview, but it has been a while since I last read it, perhaps leave the current code as it is and look into it separately, tracking in a GitHub issue? Please do read that article carefully first, it has a few options, it's not just the first one that it is mentioned.

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -0,0 +1,210 @@
...
+        info!(
+            "Server received attestation request: {:?}",
+            attestation_request
+        );
+
+        // Generate session key.
+        let key_negotiator =
+            DiffieHellmanNegotiator::create().expect("Couldn't create key negotiator");
```
tiziano88, line 147: Return error instead of panic?

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -0,0 +1,210 @@
...
+        Ok(Self {
+            tee_certificate,
+            request_handler,
+        })
+    }
+
+    /// Attest a single gRPC streaming request. Client messages are provided via `receiver`.
+    async fn attest(
+        receiver: &mut Receiver,
+        tee_certificate: Vec<u8>,
```
tiziano88, line 124: Why is this method not defined on &self directly, and use the cert from there?

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -0,0 +1,210 @@
...
+            yield attestation_response;
+
+            let mut handler = RequestHandler::new(encryptor, request_handler);
+            while let Some(response) = handler.handle_request(&mut receiver).await.expect("Couldn't process stream") {
+                yield response
+            }
+        };
+
+        Ok(Response::new(Box::pin(response_stream)))
+    }
```
tiziano88, line 209: Is it possible to encapsulate only the stream-related logic into something that makes it look essentially like implementing a function with the following type:

```rust
async fn handle(request_stream: Stream<...>, response_stream: Sink<...>) { ... }
```

(using https://docs.rs/futures/0.3.15/futures/stream/trait.StreamExt.html)


or even something with  https://docs.rs/tokio/1.6.0/tokio/sync/mpsc/index.html (though probably not worth the complexity with the buffering) :

```rust
async fn handle(request_stream: Receiver<...>, response_stream: Sender<...>) { ... }
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//!
+//! gRPC Attestation Service performs remote attestation for each incoming streaming gRPC request.
+//!
+//! It is assumed that prior to the attestation workflow Service remotely attests its TEE platform
+//! key with the TEE firmware Provider. Provider sends back a signature rooted in some well-known
+//! root key (e.g. AMD root key), thus confirming the authenticity of the TEE platform key.
+//!
+//! Remote attestation is done as follows:
+//! - Client generates a Diffie-Hellman private/public key pair
+//! - Client sends a streaming gRPC request with a its public key to the Server
```
tiziano88, line 27:
```suggestion
//! - Client sends a streaming gRPC request with its public key to the Server
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//! gRPC Attestation Service implementation.
+//!
+//! gRPC Attestation Service performs remote attestation for each incoming streaming gRPC request.
+//!
+//! It is assumed that prior to the attestation workflow Service remotely attests its TEE platform
+//! key with the TEE firmware Provider. Provider sends back a signature rooted in some well-known
+//! root key (e.g. AMD root key), thus confirming the authenticity of the TEE platform key.
+//!
+//! Remote attestation is done as follows:
+//! - Client generates a Diffie-Hellman private/public key pair
```
tiziano88, line 26:
```suggestion
//! - Client generates an ephemeral (single-use) Diffie-Hellman private/public key pair
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//! - Client sends a streaming gRPC request with a its public key to the Server
+//! - Server generates a Diffie-Hellman private/public key pair
+//! - Server uses its and Client's public keys to generate a session key
```
tiziano88, line 29:
```suggestion
//! - Server uses its and Client's public keys to generate a symmetric session key, which is shared between client and server
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//! - Server puts the hash of its public key into the TEE report request
+//!   - It means that when a TEE report is signed, the public key will also be signed
+//! - Service requests a TEE report from the TEE firmware
```
tiziano88, line 32:
```suggestion
//! - Server requests a TEE report from the TEE firmware
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//!     key presented in the server response
+//!   - It also extracts the TEE measurement from the TEE report and compares it to the expected one
+//!   - If the TEE report is not valid, then the Client closes the connection and aborts the
+//!     protocol
+//! - Client uses its and Server's public keys to generate a session key
+//! - Client uses the session key to encrypt further requests to the Server
+//! - Server uses the session key to decrypt further requests from the Client and encrypt
+//!   corresponding responses
+//!
+//! It's important to note that each new gRPC stream both Client and Server generate new pairs of
```
tiziano88, line 46:
```suggestion
//! It's important to note that for each new gRPC stream both Client and Server generate new pairs of
```

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+use crate::attestation::AttestationServer;
+use anyhow::Context;
+use futures::FutureExt;
+use log::info;
+use oak_grpc_attestation::proto::grpc_attestation_server::GrpcAttestationServer;
+use structopt::StructOpt;
+use tonic::transport::Server;
+
+#[derive(StructOpt, Clone)]
+#[structopt(about = "HTTPS Attestation")]
```
tiziano88, line 64: gRPC?

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+        .parse()
+        .context("Couldn't parse address")?;
+    let tee_certificate =
+        std::fs::read(&opt.tee_certificate).context("Couldn't load certificate")?;
+
+    let request_handler = |request: &[u8]| {
+        let parsed_request = String::from_utf8(request.to_vec()).expect("Couldn't parse request");
+        info!("Server received data: {:?}", parsed_request);
+        EXAMPLE_MESSAGE.as_bytes().to_vec()
+    };
```
tiziano88, line 99: I think this closure only works because it is actually static, but it would be more idiomatic to define it as a top-level fn.

experimental/grpc_attestation_client/rust/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+            .context("Couldn't connect via gRPC channel")?;
+        let mut client = GrpcAttestationClient::new(channel);
+
+        let (sender, mut receiver) = tokio::sync::mpsc::channel(MESSAGE_BUFFER_SIZE);
+
+        let request_stream = async_stream::stream! {
+            while let Some(message) = receiver.recv().await {
+                yield message;
+            }
+        };
```
tiziano88, line 53: Is `async_stream` actually needed here? It seems it's just passing the messages through, there must be a way that only uses very basic types? In any case, probably worth encapsulating into a separate struct / fn. In fact, isn't this similar to the one that is on the server, but implemented in a completely different way?

experimental/grpc_attestation_client/rust/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+            } else {
+                Err(anyhow!("Received incorrect message type"))
+            }?;
+
+        // Verify server's attestation info.
+        AttestationClient::verify_attestation(
+            attestation_response.attestation_info,
+            expected_tee_measurement,
+        )
+        .context("Couldn't verify server attestation")?;
```
tiziano88, line 139: I think this could be encapsulated in an `AssertionVerifier` or `ClaimVerifier` trait and corresponding struct, of which the TEE measurement one is just a special case.

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,123 @@
+//
```
tiziano88, line 1: @sidtelang I think your review would be especially useful for this file :)

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,123 @@
...
+            sealing_key,
+            opening_key,
+        }
+    }
+
+    pub fn encrypt(&mut self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let mut encrypted_data = data.to_vec();
+        self.sealing_key
+            .seal_in_place_append_tag(
+                // Additional authenticated data is not required.
```
conradgrobler, line 69: Perhaps also include why it is not required?

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//! Remote attestation is done as follows:
+//! - Client generates a Diffie-Hellman private/public key pair
+//! - Client sends a streaming gRPC request with a its public key to the Server
+//! - Server generates a Diffie-Hellman private/public key pair
+//! - Server uses its and Client's public keys to generate a session key
+//! - Server puts the hash of its public key into the TEE report request
+//!   - It means that when a TEE report is signed, the public key will also be signed
+//! - Service requests a TEE report from the TEE firmware
+//! - Server sends its public key, a TEE report and Provider's certificate to the Client
+//! - Client checks the validity of the TEE report
```
rbehjati, line 34: Where do all these checks happen?

rbehjati (commented):
Thanks. I think merging it with changes in #2071 should be straightforward.

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//!   - It checks that the TEE report is signed by TEE Provider’s root key
+//!   - It checks that the public key hash from the TEE report is equal to the hash of the public
+//!     key presented in the server response
+//!   - It also extracts the TEE measurement from the TEE report and compares it to the expected one
+//!   - If the TEE report is not valid, then the Client closes the connection and aborts the
+//!     protocol
+//! - Client uses its and Server's public keys to generate a session key
```
sidtelang, line 41: IIUC this handshake protocol does not bind the session key to the instance of the handshake protocol that creates it, opening up replay attacks.
For example: normally creating ephemeral keys for each session independently should mean that if one ephemeral key is compromised then that does not affect other sessions.
However here: compromising one server ephemeral key compromises all sessions in the future: simply replay the server response that attests the compromised ephemeral key.

The way handshake protocols generally deal with this: the server also signs a random challenge generated by the client. You can TAL at go/sc-key-provisioning#heading=h.wryz4qz7sacm for more details: tl;dr binding the ephemeral key to the transcript of the handshake so far which should include a challenge supplied by the peer you are establishing a session with.

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,123 @@
...
+            .private_key
+            .compute_public_key()
+            .map_err(|error| anyhow!("Couldn't get public key: {:?}", error))?;
+        Ok(public_key.as_ref().to_vec())
+    }
+
+    pub fn derive_session_key(self, peer_public_key: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let peer_public_key =
+            agreement::UnparsedPublicKey::new(&agreement::X25519, peer_public_key);
+        let session_key = agreement::agree_ephemeral(
```
sidtelang, line 114: Crypto best practice is to never use keys for more than one purpose: which in handshake protocols generally translates to: deriving single-purpose keys from the ephemeral shared secret.
Furthermore,  these keys can/should be bound to the handshake run by using the transcript as HKDF info (as is done in TLS, ALTS/LOAS, EKEP and our simplification of it too).

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -0,0 +1,210 @@
...
+            let request_type = request.request_type.context("Couldn't read request type")?;
+            let data = if let RequestType::Data(data) = request_type {
+                Ok(data)
+            } else {
+                Err(anyhow!("Received incorrect message type"))
+            }?;
+
+            let decrypted_data = self
+                .encryptor
+                .decrypt(&data)
```
sidtelang, line 83: does the record format for data specify which direction the data is traveling in? for e.g. is it possible for the untrusted network to take data encrypted by the server and feed it back to itself which is then interpreted as data encrypted by the client?

sidtelang (commented):
Left some comments, let me know if you want to chat in more detail about the handshake protocol. To mitigate replay attacks some kind of challenge-response is necessary: either in the handshake or in the application itself
go/sc-key-provisioning#heading=h.wryz4qz7sacm has details on how we incorporate the challenge-response in the handshake.

experimental/grpc_attestation_client/rust/src/lib.rs
ipetr0v: But the method in this struct will still need to have `expected_tee_measurement` as a parameter.
Because making it a `T` parameter may not be enough for functions that require more parameters.

experimental/grpc_attestation/src/attestation.rs
ipetr0v: Because it is used inside a stream implementation, which this class implements:
```rust
        let response_stream = async_stream::try_stream! {
            let mut receiver = Receiver { request_stream };
            let (attestation_response, encryptor) = Self::attest(&mut receiver, tee_certificate)
                .await
                .map_err(Status::internal("Couldn't attest to the client"))?;
```

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,123 @@
...
+            opening_key,
+        }
+    }
+
+    pub fn encrypt(&mut self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let mut encrypted_data = data.to_vec();
+        self.sealing_key
+            .seal_in_place_append_tag(
+                // Additional authenticated data is not required.
+                aead::Aad::from(vec![]),
```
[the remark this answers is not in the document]
ipetr0v: Nonce is a parameter for an AEAD algorithm, though I haven't seen any mentions that this implementation authenticates nonces.

experimental/grpc_attestation/src/main.rs
ipetr0v: Thanks for pointing out!
Created an issue to implement a challenge-response stage of the protocol: https://github.com/project-oak/oak/issues/2105

experimental/grpc_attestation/proto/grpc_attestation.proto
```diff
@@ -0,0 +1,48 @@
...
+message GetAttestedDataResponse {
+  oneof response_type {
+    AttestationResponse attestation_response = 1;
+    bytes data = 2;
+  }
+}
+
+message AttestationResponse {
+  bytes public_key = 1;
+  bytes attestation_info = 2;
```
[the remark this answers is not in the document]
ipetr0v: Created an issue: https://github.com/project-oak/oak/issues/2106

experimental/attestation_common/src/crypto.rs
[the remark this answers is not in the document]
ipetr0v: [RFC-5116](https://datatracker.ietf.org/doc/html/rfc5116#section-2.1) says that:
```
The nonce is authenticated internally to the algorithm, and it is not necessary to include it in the AD input.
```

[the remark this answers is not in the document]
tiziano88: Cool, thanks for the reference, could you add it in a comment?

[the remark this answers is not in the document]
ipetr0v: Done

experimental/attestation_common/src/crypto.rs
ipetr0v: Done

experimental/grpc_attestation_client/rust/src/lib.rs
ipetr0v: Will be refactored as part of the https://github.com/project-oak/oak/issues/2106

experimental/grpc_attestation/proto/grpc_attestation.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+
+service GrpcAttestation {
+  rpc GetAttestedData(stream GetAttestedDataRequest) returns (stream GetAttestedDataResponse);
```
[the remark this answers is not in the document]
rbehjati: Thanks. I did not know about those naming conventions. To better follow the naming convention, perhaps we could drop `Attested` from the method name and the request and response types, since the service name already includes `Attestation`. I'd suggest the following but I don't insist on it, and I am not entirely happy with it either :) 
`rpc SendInvocation(stream InvocationRequest) returns (stream InvocationResponse);`

experimental/grpc_attestation/src/attestation.rs
ipetr0v: There are separate Protobuf messages for client requests and server responses, but their headers are neither encrypted not authenticated with AEAD. Only the data inside them is authenticated and encrypted.

experimental/attestation_common/src/crypto.rs
ipetr0v: Does this mean that we need to generate a session key not just from `KDF(key_material, server_key, client_key)`, but also from some purpose string?

Created an issue: https://github.com/project-oak/oak/issues/2100

experimental/grpc_attestation/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+//! - Client uses the session key to encrypt further requests to the Server
```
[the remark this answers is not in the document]
ipetr0v: Can this counter be a nonce in AES-GCM ?

experimental/attestation_common/src/crypto.rs
ipetr0v: Or `KDF(key_material, server_key, client_key)` is enough, because we generate new ephemeral keys for every streaming gRPC request?

rbehjati (approved):
Apart from the handshake protocol that I cannot really comment on, the rest looks good to me.

examples/grpc_attestation/example.toml
```diff
@@ -0,0 +1,14 @@
+name = "grpc_attestation"
+
+[applications]
+
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct AeadEncryptor {
+    // Key used for encrypting data.
+    sealing_key: aead::SealingKey<OneNonceSequence>,
+    // Key used for decrypting data.
+    opening_key: aead::OpeningKey<OneNonceSequence>,
+}
+
+impl AeadEncryptor {
+    pub fn new(key: &[u8]) -> Self {
+        // AES-GCM algorithm uses the same key is used for both encryption and decryption.
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        Self {
+            sealing_key,
+            opening_key,
+        }
+    }
+
+    /// Encrypts `data` using [`AeadEncryptor::sealing_key`].
+    ///
+    /// Additional authenticated data is not required for the remotely attested channel,
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/attestation_common/src/crypto.rs
```diff
@@ -0,0 +1,155 @@
...
+                aead::Aad::from(vec![]),
+                &mut encrypted_data,
+            )
+            .map_err(|error| anyhow!("Couldn't encrypt data: {:?}", error))?;
+        Ok(encrypted_data)
+    }
+
+    /// Decrypts and authenticates `data` using [`AeadEncryptor::opening_key`].
+    ///
+    /// Additional authenticated data is not required for the remotely attested channel,
```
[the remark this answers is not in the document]
ipetr0v: Done

tiziano88 (approved):

experimental/grpc_attestation/src/main.rs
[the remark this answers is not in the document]
ipetr0v: Added a comment link to the issue: https://github.com/project-oak/oak/issues/2087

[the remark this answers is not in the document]
sidtelang: Re: Can this counter be a nonce in AES-GCM?

Yup, that's indeed what TLS 1.3 does and what our handshaker does too

experimental/attestation_common/src/crypto.rs
sidtelang: Ideally KDF(key_material, info=handshake_transcript) where handshake_transcript will include the server and client ephemeral keys.
We actually maintain separate client and server encryption keys (both client and server have both keys, the client key is used by the client to encrypt and server to decrypt, and vice versa). We append the type of the encryption key to the transcript_hash too. This is to bind the keys to the direction of data flow (i.e. as mentioned in another comment, to disallow the untrusted network to loopback responses as requests).

PR #2071, opened 2021-05-11
Add support for gRPC to Oak Functions loader 

### LINKED ISSUE #2032: Use custom Request and Response body for Oak Functions

### LINKED ISSUE #2069: Add gRPC support to Oak Functions
gRPC is needed to be able to use API proxy on Cloud Run.

rbehjati:
* Adds support for starting an Oak Functions application as a gRPC server.
* Refactors the core Oak Functions server functionality into an internal protocol-agnostic server that can be used by both HTTP/2 and gRPC servers.
* Adds a custom Request type for Oak Functions.
* Adds an option to Oak Functions config to chose between HTTP/2 and gRPC. 
* Adds a rust client to the `weather_lookup` example.

Fixes #2069
Fixes #2032


rbehjati:
@conradgrobler This just provides a very simple gRPC server with an `invoke` method. Is this enough for creating an Endpoint on Cloud Run?

oak_functions/examples/weather_lookup/client/client
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 set -o xtrace
 
+# Send the request and print only the status code and the content type.
 curl \
+  -s -o /dev/null -w "%{http_code}\n%{content_type}\n"\
```
tiziano88, line 6: --silent
--output
--write-out

oak_functions/examples/weather_lookup/client/rust/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[structopt(
+        long,
+        help = "URI of the Oak Functions application to connect to",
+        default_value = "http://localhost:8080"
+    )]
+    uri: String,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
tiziano88, line 46: This is fine, but I think we should have a generic Oak Functions client, to be used similar to `curl`, to which we pass the payload data as stdin and prints out the response to stdout.

oak_functions/loader/Cargo.toml
```diff
@@ -43,10 +43,14 @@ tokio = { version = "*", features = [
   "rt-multi-thread"
 ] }
 toml = "*"
+tonic = { version = "*", features = ["tls"] }
```
tiziano88, line 46: I don't expect we need TLS support for anything.

oak_functions/loader/src/main.rs
```diff
@@ -141,19 +172,34 @@ async fn async_main(opt: Opt, config: Config, logger: Logger) -> anyhow::Result<
...
-            config.policy.unwrap(),
-            async { notify_receiver.await.unwrap() },
-            logger,
-        )
-        .await
-        .context("error while waiting for the server to terminate")
-    });
+
+    // Start server.
+    let server_handle = match opt.protocol {
```
tiziano88, line 177: If anything, I think this switch should be a config / feature. But, I think it's probably fine to just move over to gRPC completely already.

oak_functions/loader/src/server.rs
```diff
@@ -88,75 +82,38 @@ impl Policy {
...
 
-/// Trait with a single function for serializing objects into byte arrays of a fixed size.
-trait FixedSizeSerializer {
-    /// Serializes this instance into a vector of a fixed size, based on the input `size_hint`.
-    fn serialize(&self, size_hint: usize) -> anyhow::Result<Vec<u8>>;
+/// Trait with a single function for padding objects so that they could be serialized into byte
+/// arrays of a fixed size.
+trait FixedSizePadder {
+    /// Add padding to this instance based on the input `size_hint`.
+    fn pad(&self, size_hint: usize) -> anyhow::Result<Self>
```
tiziano88, line 89:
```suggestion
    fn pad(&self, body_size: usize) -> anyhow::Result<Self>
```

oak_functions/proto/invocation.proto
```diff
@@ -44,3 +44,10 @@ enum StatusCode {
   // Indicates other internal errors at the server. Similar to HTTP 500 status code.
   INTERNAL_SERVER_ERROR = 5;
 }
+
+// Represents a request sent to an Oak Functions application. It wraps a byte array that the Wasm
+// module can interpret as a request it can handle.
+message Request {
+  // The content of the request.
+  bytes content = 1;
```
tiziano88, line 52: `body`? This is symmetric to `Response` above. Also perhaps move to the top of the file, usually readers expect to see requests before responses.

oak_functions/proto/server.proto
```diff
@@ -0,0 +1,25 @@
...
+//
+
+syntax = "proto3";
+
+package oak.functions.server;
+
+import "invocation.proto";
+
+service GrpcHandler {
+  rpc Invoke(oak.functions.invocation.Request) returns (oak.functions.invocation.Response);
```
tiziano88, line 24: eventually we would combine this with the bidirectional streaming key exchange that @ipetr0v implemented in #2070

oak_functions/loader/src/server.rs
```diff
@@ -486,22 +443,26 @@ fn check_export_function_signature(
...
+/// response may be padded by a number of trailing 0s before encoding the response as a binary
+/// protobuf message. In this case, the `length` in the response will contain the effective length
+/// of the `body`. For a valid security policy, this response is guaranteed to comply with the
+/// policy's size restriction. If the policy is not valid, the function returns early with an error.
+pub async fn apply_policy<F, S>(policy: Policy, function: F) -> anyhow::Result<Response>
 where
     F: std::marker::Send + 'static + FnOnce() -> S,
     S: std::future::Future<Output = anyhow::Result<Response>> + std::marker::Send,
 {
+    policy.validate()?;
```
tiziano88, line 464: Another option is to have separate `Policy` and `ValidatedPolicy` types; then make `ValidatedPolicy` impl `TryFrom<Policy>` (and this should be the only way of constructing an instance, of course).

tiziano88 (approved):

oak_functions/examples/weather_lookup/client/client
rbehjati: I was looking for something like this. Thanks :)

ipetr0v (approved):

conradgrobler (approved):

conradgrobler:
> @conradgrobler This just provides a very simple gRPC server with an `invoke` method. Is this enough for creating an Endpoint on Cloud Run?

Yes, this should be sufficient.

PR #2086, opened 2021-05-14
Add support for authenticated downloads of lookup data

### LINKED ISSUE #2078: Add support for authenticated downloads of lookup data to Oak Functions
At the moment the function loader can only download lookup data that is publicly available. It would be useful to add support to the loader to authenticate as part of downloading data.

Suggested design:

- Add support for a token fetching URL by adding an additional 
  - In the generic case this can use network-based security to ensure only the loader can download the token
  - In the Google Cloud case the metadata service can be used to get an auth token for the Cloud Run instance service account (https://cloud.google.com/run/docs/securing/service-identity)
- Once the token is fetched, it can be used to authenticate the lookup data download request
  - For now we can just assume it uses OAuth2

conradgrobler:
Fixes #2078 

# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/src/lookup.rs
+++ b/oak_functions/loader/src/lookup.rs
@@ -146,11 +146,11 @@ impl LookupData {
     async fn build_download_request(&self) -> anyhow::Result<Request<Body>> {
         let url: &str = &self.lookup_data_url;
         let mut builder = Request::builder().method("GET").uri(url);
-        let access_token = get_access_token()
+        if let Some(access_token) = get_access_token()
             .await
-            .context("could not get access token")?;
-        if access_token.is_some() {
-            builder = builder.header("Authorization", format!("Bearer {}", access_token.unwrap()));
+            .context("could not get access token")?
+        {
+            builder = builder.header("Authorization", format!("Bearer {}", access_token));
         }
         builder
             .body(Body::empty())
--- a/oak_functions/loader/src/lookup.rs
+++ b/oak_functions/loader/src/lookup.rs
@@ -20,7 +20,20 @@ use hyper::{body::Bytes, client::connect::Connect, Body, Client, Request};
 use hyper_rustls::HttpsConnector;
 use log::Level;
 use prost::Message;
+use serde_derive::Deserialize;
 use std::{collections::HashMap, sync::RwLock, time::Instant};
+
+#[derive(Copy, Clone, Deserialize, Debug)]
+pub enum LookupDataAuth {
+    None,
+    GcpMetadataToken,
+}
+
+impl LookupDataAuth {
+    pub fn default() -> Self {
+        LookupDataAuth::None
+    }
+}
 
 /// An in-memory lookup store instance that can refresh its internal entries from the provided data
 /// file URL.
@@ -29,7 +42,7 @@ use std::{collections::HashMap, sync::RwLock, time::Instant};
 /// protobuf messages according to the definition in `/oak_functions/proto/lookup_data.proto`.
 pub struct LookupData {
     lookup_data_url: String,
-    use_gcp_metadata_auth: bool,
+    lookup_data_auth: LookupDataAuth,
     entries: RwLock<HashMap<Vec<u8>, Vec<u8>>>,
     logger: Logger,
 }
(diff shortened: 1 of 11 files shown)


oak_functions/loader/src/lookup.rs
```diff
@@ -163,3 +178,36 @@ pub fn parse_lookup_entries<B: prost::bytes::Buf>(
     }
     Ok(entries)
 }
+
+async fn request_bytes<C>(client: &Client<C, Body>, request: Request<Body>) -> anyhow::Result<Bytes>
```
rbehjati, line 182: How about `response_bytes` or `send_request`?

oak_functions/loader/src/main.rs
```diff
@@ -48,6 +48,10 @@ struct Config {
     /// How often to refresh the lookup data. If not provided, data is only loaded once at startup.
     #[serde(with = "humantime_serde")]
     lookup_data_download_period: Option<Duration>,
+    /// Whether to use the GCP metadata service to obtain an authentication token for downloading
+    /// the lookup data.
+    #[serde(default)]
+    use_gcp_metadata_auth: bool,
```
rbehjati, line 54: Can we configure one of the examples to use GCP authentication?

rbehjati (approved):
LGTM.

oak_functions/loader/src/main.rs
conradgrobler: The metadata service is only available to workloads running on GCP, so this won't work with local examples or on github CI.

oak_functions/loader/src/lookup.rs
conradgrobler: Renamed to `send_request`.

conradgrobler (commented):
Thanks

oak_functions/loader/src/lookup.rs
```diff
@@ -163,3 +178,36 @@ pub fn parse_lookup_entries<B: prost::bytes::Buf>(
...
+        .await
+        .context("could not read response body")
+}
+
+/// Gets a service account access token from the GCP metadata service.
+async fn get_access_token() -> anyhow::Result<String> {
+    let client = Client::new();
+    let request = Request::builder()
+        .method("GET")
+        .uri("http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token")
```
tiziano88, line 200: Could you add a reference to where this is taken?

oak_functions/loader/src/lookup.rs
```diff
@@ -163,3 +178,36 @@ pub fn parse_lookup_entries<B: prost::bytes::Buf>(
...
+    hyper::body::to_bytes(response.into_body())
+        .await
+        .context("could not read response body")
+}
+
+/// Gets a service account access token from the GCP metadata service.
+async fn get_access_token() -> anyhow::Result<String> {
+    let client = Client::new();
+    let request = Request::builder()
+        .method("GET")
```
tiziano88, line 199: https://docs.rs/http/0.2.4/http/method/struct.Method.html#associatedconstant.GET

oak_functions/loader/src/lookup.rs
```diff
@@ -38,9 +40,14 @@ impl LookupData {
     ///
     /// The returned instance is empty, and must be populated by calling the [`LookupData::refresh`]
     /// method at least once.
-    pub fn new_empty(lookup_data_url: &str, logger: Logger) -> LookupData {
+    pub fn new_empty(
+        lookup_data_url: &str,
+        use_gcp_metadata_auth: bool,
```
tiziano88, line 45: I find the bool not very readable when invoking this fn; another option would be to actually pass in a fn to fetch the token here, and we can either pass a reference to `get_access_token`, or to antoher fn `empty_token` instead that always returns None or empty string. Up to you anyways.

tiziano88 (approved):

oak_functions/loader/src/lookup.rs
conradgrobler: Done.

oak_functions/loader/src/lookup.rs
conradgrobler: I agree that the bool is not very clear, but storing async function references was not great either. So I decided to swtich to using an enum, which hopefully makes it more readable.

oak_functions/loader/src/lookup.rs
conradgrobler: Done.

oak_functions/loader/src/lookup.rs
tiziano88: LGTM; anyways the fact that the fn is async is irrelevant in terms of ref I think, it's just a normal fn that returns `Future<Out>`, right? Or does it not work because of lifetimes and traits? (just curious, no need to spend more time)

conradgrobler: Yes, lifetimes and traits made it more complex. I could make it work with using `BoxFuture`, a closure and pinning, but it made the code look much more complicated.

Issue #2094, opened 2021-05-17
Oak Functions CLI client


tiziano88:
We should have a command line client to invoke Oak Functions instances, which reads the request payload from stdin and prints the response body to stdout (similar to how `curl` works for HTTP).

tiziano88:
Implemented in https://github.com/project-oak/oak/tree/main/oak_functions_client

PR #2095, opened 2021-05-17
Key / value lookup example and generic client

### LINKED ISSUE #2093: key / value lookup example for Oak Functions
since the current weather example may evolve to be more and more weather-specific, it would be useful (also as a reusable building block for other projects, and as a benchmarking baseline) to have an even simpler example that shows how to perform a straight key / value lookup via Oak Functions.

cc @anghelcovici

tiziano88:
The intention is to keep this example around as a building block for
other discussions, and also have something that does not necessarily
need to evolve over time based on other requirements.

The generic client may be used for demos and debugging instances of Oak
Functions running either remotely or locally.

Fix #2093
Ref #2094

# Checklist

      construction.


oak_functions/examples/key_value_lookup/README.md
```diff
@@ -0,0 +1,14 @@
+# Oak Functions `key_value_lookup` example
+
+This example shows the simplest nontrivial application of Oak Functions that
+uses the key / value lookup functionality.
+
+For each incoming client request, it performs an individual lookup using the
+in-memory key / value store, and returns the item, if found, back to the client.
+If the item with the provided key is not found, it returns an empty response.
+
+Both request and response are opaque byte arrays.
```
conradgrobler, line 10: Might be useful to add a paragraph here about reading the lookup from stdin and perhaps even a code snippet on how to use the client. Or, alternatively a readme in the client directory to cover that.

oak_functions/examples/key_value_lookup/example.toml
```diff
@@ -0,0 +1,18 @@
...
+wasm_path = "oak_functions/examples/bin/key_value_lookup.wasm"
+target = { Cargo = { cargo_manifest = "oak_functions/examples/key_value_lookup/module/Cargo.toml" } }
+
+[server]
+additional_args = [
+  "--config-path=./oak_functions/examples/key_value_lookup/config.toml"
+]
+
+[clients]
+rust = { Cargo = { cargo_manifest = "oak_functions/client/Cargo.toml" }, additional_args = [
```
conradgrobler, line 16: I assume this sends a request with an empty body. It might be useful for testing to add support for piping a value into the client via stdin in future.

conradgrobler (approved):

oak_functions/examples/key_value_lookup/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+It is not possible for a client to distinguish between a successful lookup of a
+key / value entry in which the value is an empty byte array, and a failed lookup
+for which the key was not found at all in the in-memory store.
```
ipetr0v, line 14: Should application indicate that there was an error somehow?

oak_functions/examples/key_value_lookup/module/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+//! Oak Functions key / value lookup example.
+
+#[cfg(test)]
+mod tests;
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn main() {
+    let request = oak_functions::read_request().expect("Couldn't read request body.");
+    let response = oak_functions::storage_get_item(&request)
+        .expect("Couldn't look up entry")
```
ipetr0v, line 26: If the application has panicked, should gRPC server return a gRPC error?

ipetr0v (approved):

oak_functions/examples/key_value_lookup/config.toml
```diff
@@ -0,0 +1,8 @@
+protocol = "Grpc"
+lookup_data_url = "https://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
```
rbehjati, line 2: Is the use of weather lookup data intentional?

oak_functions/examples/key_value_lookup/module/src/tests.rs
```diff
@@ -0,0 +1,128 @@
...
+    ));
+    lookup_data.refresh().await.unwrap();
+
+    let policy = Policy {
+        constant_response_size_bytes: 100,
+        constant_processing_time: Duration::from_millis(200),
+    };
+
+    let server_background = test_utils::background(|term| async move {
+        create_and_start_http_server(
```
rbehjati, line 72: I suggest using the gRPC server instead, to be more consistent with the rest of the changes in this PR.

rbehjati (approved):

oak_functions/examples/key_value_lookup/README.md
tiziano88: I don't want to add more arbitrary logic unless it is clear that it is needed yet.

oak_functions/examples/key_value_lookup/config.toml
tiziano88: Sort of, this example is not used for anything yet, so any data source will do. I will probably change the deployment scripts in the future and generate some more predictable data to use in a demo.

oak_functions/examples/key_value_lookup/module/src/lib.rs
tiziano88: Definitely not, that would leak data at each invocation otherwise, right?

oak_functions/examples/key_value_lookup/module/src/tests.rs
tiziano88: For now I just copied the logic from the other example, but I think when we move from HTTP to gRPC we should update both examples at once.

oak_functions/client/src/main.rs
```diff
@@ -0,0 +1,67 @@
...
+    let opt = Opt::from_args();
+    let uri: Uri = opt.uri.parse().context("Error parsing URI")?;
+
+    // Create client.
+    let channel = Channel::builder(uri.clone()).connect().await?;
+    let mut client = GrpcHandlerClient::new(channel);
+
+    // Create and send request.
+    let mut request_body = Vec::new();
+    std::io::stdin()
```
rbehjati, line 56: Since it only reads once from stdin, could the `request_body` be instead passed as an argument to main?

PR #2099, opened 2021-05-18
Update git_check_diff to check the presence of any newly generated files

rbehjati:

### DIFF
--- a/scripts/git_check_diff
+++ b/scripts/git_check_diff
@@ -5,4 +5,7 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 source "$SCRIPTS_DIR/common"
 
 # Check that any generated files match those that are checked in.
-git diff --exit-code -- . ':!*.bazelrc'
+if [[ $(git status --short) ]]; then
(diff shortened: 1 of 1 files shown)


scripts/git_check_diff
```diff
@@ -5,4 +5,7 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 source "$SCRIPTS_DIR/common"
 
 # Check that any generated files match those that are checked in.
-git diff --exit-code -- . ':!*.bazelrc'
+if [[ $(git status --short | wc --line) -gt 0 ]]; then
+    # One or more files are either modified or newly generated, exit with an error code
```
tiziano88, line 9: Could you make sure this actually prints the diff though? Otherwise if it fails we need to spend even more time trying to reproduce locally by rebuilding everything.

tiziano88 (approved):

scripts/git_check_diff
rbehjati: Changed it to print the complete list of modified or added files.

PR #2104, opened 2021-05-18
Deploy GCP API Gateway for the Oak Functions weather lookup example

conradgrobler:
This change also updates the Oak Functions Loader example script to use the generic client to test connectivity after deployment.

# Checklist

      construction.


oak_functions/gcp/weather_lookup_api_config.yaml
```diff
@@ -0,0 +1,28 @@
...
+usage:
+  rules:
+    # Invoke method can be called without an API Key for now.
+    - selector: oak.functions.server.GrpcHandler.Invoke
+      allow_unregistered_calls: true
+
+backend:
+  rules:
+    - selector: '*'
+      address: grpcs://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app
```
tiziano88, line 28: Where do we get this from?

oak_functions/gcp/weather_lookup_api_config.yaml
```diff
@@ -0,0 +1,28 @@
...
+# Name of the service configuration.
+#
+name: '*.apigateway.endpoints.oak-ci.cloud.goog'
+
+#
+# API title to appear in the user interface (Google Cloud Console).
+#
+title: Oak Functions gRPC API Gateway
+apis:
+  - name: oak.functions.server.GrpcHandler
```
tiziano88, line 14: Is this just the protobuf package in the gRPC definition?

scripts/deploy_oak_functions_api
```diff
@@ -0,0 +1,41 @@
...
+# Deploy the API Configuration
+readonly CONFIG_ID='weather-lookup-config'
+readonly API_ID='weather-lookup-api'
+
+gcloud api-gateway api-configs create "${CONFIG_ID}" \
+  --api="${API_ID}" --project="${GCP_PROJECT_ID}" \
+  --grpc-files=./oak_functions/gcp/server.pb,./oak_functions/gcp/weather_lookup_api_config.yaml
+
+# Deploy the API Gateway
+readonly GATEWAY_ID='weather-lookup-grpc'
```
tiziano88, line 31: What's the difference between config, api and gateway id? Is there some documentation to link here that summarizes this?

scripts/deploy_oak_functions_api
```diff
@@ -0,0 +1,41 @@
...
+
+gcloud api-gateway gateways create "${GATEWAY_ID}" \
+  --api="${API_ID}" --api-config="${CONFIG_ID}" \
+  --location="${GCP_REGION}" --project="${GCP_PROJECT_ID}"
+
+# Get the Managed Service name
+MANAGED_SERVICE=$(gcloud api-gateway apis describe "${API_ID}" --format='value(managedService)')
```
tiziano88, line 38: readonly, and quote RHS

tiziano88 (approved):

scripts/deploy_oak_functions_api
```diff
@@ -0,0 +1,41 @@
...
+gcloud auth activate-service-account \
+  --project="${GCP_PROJECT_ID}" \
+  --key-file="${GCP_ACCOUNT_FILE}"
+
+# Rebuild the proto descriptor file for the API configuration.
+protoc \
+  --proto_path=./oak_functions/proto \
+  --include_imports \
+  --include_source_info \
+  --descriptor_set_out=./oak_functions/gcp/server.pb \
```
rbehjati, line 19: Why is `oak_functions/gcp/server.pb` checked-in even though it is generated here?

oak_functions/gcp/weather_lookup_api_config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+#
+# Name of the service configuration.
+#
+name: '*.apigateway.endpoints.oak-ci.cloud.goog'
+
+#
+# API title to appear in the user interface (Google Cloud Console).
+#
+title: Oak Functions gRPC API Gateway
```
rbehjati, line 12: Is this a generic config that could be used with all Oak Functions applications or is it specific to the `weather_lookup` app? If it is the former, perhaps the file name should be changed?

oak_functions/gcp/weather_lookup_api_config.yaml
rbehjati: Could this be made parametric, since this is the only part of this file that is specific to the weather_lookup app?

oak_functions/gcp/weather_lookup_api_config.yaml
conradgrobler: It is the fully qualified service name for the service in the proto file, so the package name and service name concatenated.

oak_functions/gcp/weather_lookup_api_config.yaml
conradgrobler: Added a comment on where this comes from. gcloud does not support parameterising these config files.

scripts/deploy_oak_functions_api
conradgrobler: Added a link. The API is a named container for configs and gateways. A config defines a single gRPC service. A gateway is an Envoy proxy instance that manages traffic based on a config.

rbehjati (approved):

scripts/deploy_oak_functions_api
conradgrobler: Done.

oak_functions/gcp/weather_lookup_api_config.yaml
conradgrobler: This is specific to the instance being deployed.

scripts/deploy_oak_functions_api
conradgrobler: It is reusable for manual deployments as well, so I thought it would be a useful reference point to check it in.

scripts/deploy_oak_functions_api
```diff
@@ -0,0 +1,69 @@
...
+readonly API_TITLE='Oak Functions Weather Lookup Service'
+# Fetch the URL or the Cloud Run backend service instance.
+readonly CLOUD_RUN_URL="$(gcloud beta run services describe "${FUNCTIONS_INSTANCE_NAME}" --project="${GCP_PROJECT_ID}" --region=europe-west2 --platform=managed --format='value(status.url)')" 
+readonly CLOUD_RUN_HOST="$(echo "${CLOUD_RUN_URL}" | cut -c 9-)"
+
+sed \
+  --expression="s/###TITLE###/${API_TITLE}/" \
+  --expression="s/###HOST_NAME###/${CLOUD_RUN_HOST}/" \
+  < ./oak_functions/gcp/api_config.yaml.template \
+  > ./oak_functions/bin/weather_lookup_api_config.yaml \
```
tiziano88, line 50: Use OUT_DIR here too?

tiziano88 (approved):
LGTM, though personally I much prefer the version without templates; I think hardcoded strings are fine, just with links or commands to double check them if / when we redeploy something fundamentally. Currently none of the actual URLs are checked in, so it's hard to figure out whether the current status of the cloud config is in sync with the checked-in version. BTW did you look into using Terraform for this too (here on GitHub for the dev project)? Is it worth considering?

tiziano88:
Also the parametrization is going to stop working as soon as we want to deploy two versions, one with API key access and one without; then we end up parametrizing shell scripts, which IMO is decidedly worse than having two distinct YAML files checked in.

rbehjati:
+1 for considering Terraform.

scripts/deploy_oak_functions_api
conradgrobler: Done.

conradgrobler:
> LGTM, though personally I much prefer the version without templates; I think hardcoded strings are fine, just with links or commands to double check them if / when we redeploy something fundamentally. Currently none of the actual URLs are checked in, so it's hard to figure out whether the current status of the cloud config is in sync with the checked-in version. BTW did you look into using Terraform for this too (here on GitHub for the dev project)? Is it worth considering?

I will leave as is for now, but have created an issue (#2107) to look into using TerraForm for deploying the Cloud Run instance and an associated Cloud Endpoint.

PR #2110, opened 2021-05-20
Run test_long_response_time with multiple threads

rbehjati:


rbehjati:
@tiziano88 Could you please take a look?

oak_functions/loader/src/tests.rs
```diff
@@ -82,13 +81,18 @@ async fn test_long_response_time() {
     // So we expect the request to fail, with `response not available error`.
     let scenario = |server_port: u16| async move {
         let result = make_request(server_port, br#"{"lat":52,"lon":0}"#).await;
-        // TODO(#1987): Check elapsed time
+        // Check the elapsed time, allowing a margin of 10ms.
+        assert!(
+            result.elapsed < Duration::from_millis(allowed_processing_time + 10),
```
tiziano88, line 86: This should be adding Durations, not numbers. Something like

```rust
let margin = Duration::from_millis(10);
let constant_processing_time = Duration::from_millis(10);
...
assert!(result.elapsed < constant_processing_time + margin);
```

oak_functions/loader/src/server.rs
```diff
@@ -493,13 +493,13 @@ where
         // The `function` did not terminate within the policy timeout
         None => Response::create(
             StatusCode::PolicyTimeViolation,
-            "Reason: response not available".as_bytes().to_vec(),
+            "Reason: response not available.\n".as_bytes().to_vec(),
```
tiziano88, line 496: Personally I would not add newlines to these strings, or are they needed for something? Adding newlines makes it harder to include them as part of other messages.

rbehjati: This is the response body, it is not being included in any other messages. Is it? 
Adding the newline gives more readability when the message is printed in stdout. Although, now we only support gRPC and the response is wrapped in a tonic::Response. So, I think it does not matter.

oak_functions/loader/src/tests.rs
rbehjati: Thanks. Done.

oak_functions/loader/src/tests.rs
```diff
@@ -153,6 +157,7 @@ where
     // Yield to the Tokio runtime’s scheduler, to allow the server thread to make progress before
     // starting the client. This is needed for a more accurate measurement of the processing time.
     tokio::task::yield_now().await;
+    std::thread::sleep(Duration::from_secs(1));
```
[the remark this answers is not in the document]
rbehjati: Done.

tiziano88 (approved):

PR #2111, opened 2021-05-20
Add flag to Oak Functions client to pass request

### LINKED ISSUE #2101: Add support to oak_functions_client to pass request body as command line argument
Original comment:

I assume this sends a request with an empty body. It might be useful for testing to add support for piping a value into the client via stdin in future.

_Originally posted by @conradgrobler in https://github.com/project-oak/oak/pull/2095#r634146074_

@conradgrobler I think it's actually simpler to allow the client to accept the request payload as an additional argument, instead of just via stdin.

tiziano88:
Fix #2101

# Checklist

      construction.


oak_functions/client/src/main.rs
```diff
@@ -39,6 +38,8 @@ pub struct Opt {
         default_value = "http://localhost:8080"
     )]
     uri: String,
+    #[structopt(long, help = "request payload")]
+    request: String,
```
conradgrobler, line 42: Seeing that this is required, I assume the flag must be passed in the example as well.

conradgrobler (approved):

tiziano88:
@conradgrobler I removed the weather lookup client, in favour of using the general purpose client. The business logic of the weather lookup is already tested by the integration tests, running the client is only to catch any obvious incompatibility in the gRPC protocol (especially going forward).

tiziano88:
@conradgrobler PTAL

conradgrobler (approved):

PR #2112, opened 2021-05-20
Add Java client for gRPC Attestation

### LINKED ISSUE #2113: Implement Java client for gRPC Attestation
In order to support Android clients for gRPC attestation (https://github.com/project-oak/oak/pull/2070), we first need to implement a Java version of such client.

ipetr0v:
This commit:
- Adds a Java client for gRPC Attestation
- Updates Bazel `WORKSPACE`
- Adds a patch that fixes Tink Java library compilation

Fixes #2113

### DIFF
--- a/experimental/grpc_attestation_client/java/com/google/oak/grpc_attestation_client/AeadEncryptor.java
+++ b/experimental/grpc_attestation_client/java/com/google/oak/grpc_attestation_client/AeadEncryptor.java
@@ -3,6 +3,10 @@ package com.google.oak.grpc_attestation_client;
 import com.google.crypto.tink.subtle.AesGcmJce;
 import java.security.GeneralSecurityException;
 
+/**
+ * Implementation of Authenticated Encryption with Associated Data (AEAD).
+ * https://datatracker.ietf.org/doc/html/rfc5116
+ */
 public class AeadEncryptor {
     private AesGcmJce encryptor;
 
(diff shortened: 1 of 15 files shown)


ipetr0v:
There is `openjdk` library in our Android Docker image.
Not sure if it's necessary to add it to the main one.

conradgrobler (approved):

experimental/grpc_attestation_client/java/com/google/oak/grpc_attestation_client/AttestationClient.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import oak.examples.grpc_attestation.AttestedInvokeResponse;
+import oak.examples.grpc_attestation.AttestedInvokeRequest;
+import oak.examples.grpc_attestation.ClientIdentity;
+import oak.examples.grpc_attestation.ServerIdentity;
+import oak.examples.grpc_attestation.GrpcAttestationGrpc;
+import oak.examples.grpc_attestation.GrpcAttestationGrpc.GrpcAttestationStub;
+
+public class AttestationClient {
```
tiziano88, line 22: This works fine for the gRPC case, but it is too tied to the underlying transport. I think we should still have a separate state machine that accepts messages or Java objects to move from one state to the next, and that may be used in any context (e.g. it may be used on a server, since the protocol is symmetric). Then of course a wrapper like this that is tied to gRPC would be trivial to implement on top of that.

Edit: perhaps KeyNegotiation is similar to what I'm describing already, but it should itself hand out the encryptor / decryptor in a way that makes it impossible to make mistakes. Anyways, we can do this separately.

experimental/grpc_attestation_client/java/com/google/oak/grpc_attestation_client/AeadEncryptor.java
```diff
@@ -0,0 +1,38 @@
+package com.google.oak.grpc_attestation_client;
+
+import com.google.crypto.tink.subtle.AesGcmJce;
```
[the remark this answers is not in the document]
tiziano88: Is there a wrapper interface for the AES and X25519 functionality in tink that is not in `subtle`?

tiziano88 (approved):

experimental/grpc_attestation_client/java/com/google/oak/grpc_attestation_client/AttestationClient.java
ipetr0v: Moved encryptor creation to `KeyNegotiator`.
Also created an issue for implementing a protocol independent state machine: https://github.com/project-oak/oak/issues/2121

PR #2114, opened 2021-05-20
Add fuzzer for WasmHandler

### LINKED ISSUE #2047: Fuzz testing for oak_functions

rbehjati:
Fixes #2047 (more fuzz targets can be added later)


rbehjati:
@alastairreid I thought you might have some insights regarding this PR.

oak_functions/examples/fuzzable/module/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[cfg(test)]
+mod tests;
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn main() {
+    let request = oak_functions::read_request().expect("Couldn't read request body.");
+    let request = AbiFunctions::decode(&*request).expect("Couldn't decode request.");
+
+    // Run all the functions given in the request
```
tiziano88, line 35: Maybe rename functions to instructions? I was a bit confused at first. Also in the future we may add more interesting building blocks, e.g. iterating on some entries, or performing some computation, etc. , it may not be only ABI function calls. (in fact, panic is already not an ABI function call)

oak_functions/examples/fuzzable/module/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    for function in request.functions {
+        match function.function {
+            Some(Function::Panic(_)) => panic!("panic"),
```
tiziano88, line 38: I would specify the ignored messages as patterns, so that if we add fields in the future, the compile will catch it.

oak_functions/examples/fuzzable/module/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+                    .expect("Couldn't find key in the storage")
+                    .unwrap_or_default();
+            }
+            Some(Function::WriteLogMessage(WriteLogMessage { message })) => {
+                oak_functions::write_log_message(
+                    std::str::from_utf8(&message).expect("Couldn't convert bytes to string"),
+                )
+                .expect("Couldn't write log message.")
+            }
+            _ => (),
```
tiziano88, line 56: I would also make this exhaustive for the same reason.

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -0,0 +1,115 @@
...
+use libfuzzer_sys::fuzz_target;
+use oak_functions_abi::proto::Request;
+use oak_functions_loader::{
+    grpc::handle_request, logger::Logger, lookup::LookupData, server::WasmHandler,
+};
+use prost::Message;
+use std::sync::Arc;
+
+#[derive(Arbitrary, Debug)]
+enum AbiFunction {
```
tiziano88, line 33: Could you impl To / From this to the proto version, instead of doing the translation inlined below?

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -0,0 +1,115 @@
...
+    StorageGetItem {
+        // The key to lookup.
+        key: Bytes,
+    },
+    WriteLogMessage {
+        message: Bytes,
+    },
+}
+
+/// Wrapper for a byte vector, to provide a custom `Arbitrary` implementation.
```
tiziano88, line 49: I'm surprised there is no `Arbitrary` impl for `Vec<u8>`?

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const MANIFEST_PATH: &str = "../examples/fuzzable/module/Cargo.toml";
+lazy_static::lazy_static! {
+    static ref WASM_MODULE_BYTES: Vec<u8> =
+        test_utils::compile_rust_wasm(MANIFEST_PATH).expect("Couldn't read Wasm module");
+}
+
+// Randomly generate a list of `AbiFunctions` and send them to the Wasm module to run them.
+fuzz_target!(|methods: Vec<AbiFunction>| {
```
tiziano88, line 69: Not sure why functions are called methods here.

tiziano88 (approved):
This is great! Mostly just a few style comments.

alastairreid:
May be too late for review but...

I tried to figure out how you controlled generation of keys in the fuzzer.
Just picking random Vec<u8>s might make it unlikely to get sequences that use the same key often enough that you test anything other than error reporting code.

And, on a related note: does the fuzzer library give any way of indicating code that you want to have covered?
eg you might be really interested in having code that successfully performs some operation but, to get there, it first has to execute several commands successfully with the same key. A report that says that you are hitting your target piece of code 3% (say) of the time might guide you as to whether the fuzzer is working well enough for your needs.

rbehjati:
Thanks @alastairreid. 

That is a good point about the keys. I think we would want keys of different lengths that exist in the lookup table, and map to values of different lengths. But I am not sure how filling the lookup table for each test case would affect the performance. 

I don't think there is a way to specify the code that we want to cover, but we get coverage reports on [ClusterFuzz](https://oss-fuzz.com/fuzzer-stats?project=oak&fuzzer=libFuzzer&job=libfuzzer_asan_oak&group_by=by-fuzzer). (There is not a coverage report for the fuzz-target in this PR yet. I believe it will be available in 1-2 days.) There is also a performance report, that we could use together with the coverage report to improve the fuzzing.

PR #2122, opened 2021-05-24
Add gRPC attestation support to Oak Functions

### LINKED ISSUE #1932: Remote attestation for Oak functions
We need basic Remote Attestation support for Oak functions. 

Update:
- Integrate support for RA via gRPC (#2070) into the Oak Functions loader.
- Update the Rust Client

rbehjati:
* Changes the type of `handler` in `RequestHandler` and `AttestationServer`, in `grpc_attestation`.
* Removes `oak_functions/proto/server.proto` and `GrpcHandler` in favour of gRPC attestion.
* Updates Cloud Run deployment scripts accordingly. 
* Uses `oak_functions_abi::proto::Request` as the request type accepted by the server. In the future we'll add more fields to the request.
* Adds `tee_certificate_path` to Oak Functions server config. 

Fixes #1932.


experimental/grpc_attestation/src/attestation.rs
```diff
@@ -101,18 +105,21 @@ impl RequestHandler {
...
-    pub fn create(
-        tee_certificate: Vec<u8>,
-        request_handler: fn(&[u8]) -> Vec<u8>,
-    ) -> anyhow::Result<Self> {
+impl<F, S> AttestationServer<F>
+where
+    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
+    S: std::future::Future<Output = anyhow::Result<Vec<u8>>>
+        + std::marker::Send
+        + std::marker::Sync,
```
ipetr0v, line 120: Previously in `impl<F, S> RequestHandler<F>` you didn't need `std::marker::`

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -174,7 +181,11 @@ impl AttestationServer {
 }
 
 #[tonic::async_trait]
-impl RemoteAttestation for AttestationServer {
+impl<F, S> RemoteAttestation for AttestationServer<F>
+where
+    F: 'static + Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
+    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync + 'static,
```
ipetr0v, line 187: In `impl<F, S> AttestationServer<F>` you didn't have `'static`.
Should this template specification be the same everywhere?

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -183,7 +194,7 @@ impl RemoteAttestation for AttestationServer {
         request_stream: Request<tonic::Streaming<AttestedInvokeRequest>>,
     ) -> Result<Response<Self::AttestedInvokeStream>, Status> {
         let tee_certificate = self.tee_certificate.clone();
-        let request_handler = self.request_handler;
+        let request_handler = self.request_handler.clone();
```
ipetr0v, line 197: Why do you need `clone()` if `F` is `static`?

experimental/grpc_attestation/src/main.rs
```diff
@@ -82,10 +81,10 @@ pub struct Opt {
 // Example message sent to the client.
 const EXAMPLE_MESSAGE: &str = "Example message";
 
-fn request_handler(request: &[u8]) -> Vec<u8> {
+async fn request_handler(request: Vec<u8>) -> anyhow::Result<Vec<u8>> {
     let parsed_request = String::from_utf8(request.to_vec()).expect("Couldn't parse request");
```
ipetr0v, line 85: Since it returns `anyhow::Result` now, should `expect` be changed to `context` ?

oak_functions/client/src/lib.rs
```diff
@@ -13,35 +13,32 @@
...
-use crate::proto::grpc_handler_client::GrpcHandlerClient;
 use anyhow::Context;
-use http::uri::Uri;
-use oak_functions_abi::proto::{Request, Response};
-use tonic::transport::Channel;
+use oak_functions_abi::proto::Response;
+use oak_grpc_attestation_client::AttestationClient;
+use prost::Message;
+
+const TEE_MEASUREMENT: &[u8] = br"Test TEE measurement";
```
ipetr0v, line 21: Could you please add a:
```rust
// TODO(#1867): Add remote attestation support.
```

oak_functions/client/src/main.rs
```diff
@@ -33,24 +34,37 @@ pub struct Opt {
     uri: String,
     #[structopt(long, help = "request payload")]
     request: String,
+    #[structopt(long, help = "expected response body")]
+    expected_response: Option<String>,
```
ipetr0v, line 38: Could you add a comment that it's needed for testing?

oak_functions/examples/key_value_lookup/module/src/tests.rs
```diff
@@ -70,6 +70,7 @@ async fn test_server() {
     let server_background = test_utils::background(|term| async move {
         create_and_start_grpc_server(
             &address,
+            vec![],
```
ipetr0v, line 73: Could you add a comment, what this argument refers to?

oak_functions/loader/src/grpc.rs
```diff
@@ -81,13 +69,16 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
...
     // A `Service` is needed for every connection. Here we create a service using the
     // `wasm_handler`.
     tonic::transport::Server::builder()
-        .add_service(GrpcHandlerServer::new(FunctionsServer {
-            wasm_handler,
-            policy,
-        }))
+        .add_service(RemoteAttestationServer::new(
+            AttestationServer::create(tee_certificate, request_handler)
+                .context("Couldn't create proxy")?,
```
ipetr0v, line 80: Should it be called a proxy?
Maybe `Couldn't create remote attestation server` ?

ipetr0v (approved):

experimental/grpc_attestation/src/attestation.rs
rbehjati: Thanks. Fixed.

experimental/grpc_attestation/src/attestation.rs
rbehjati: `'static` is needed because of the lifetime bounds on `AttestationServer`. This is a requirement from `tonic`, because here we are implementing `RemoteAttestation` for `AttestationServer`. But `RequestHandler` does not have this lifetime bound. 
I think it is okey that the template specifications are different.

experimental/grpc_attestation/src/attestation.rs
rbehjati: `F` has a `'static` lifetime, but the moving and borrowing rules still apply. This is the error without `clone`:
```
cannot move out of `_self.request_handler` which is behind a shared reference
```
I could borrow instead of cloning, but that requires specifying lifetimes everywhere, even if `F` is `'static` in `RequestHandler`.

experimental/grpc_attestation/src/main.rs
rbehjati: Done.

oak_functions/client/src/lib.rs
rbehjati: Done!

oak_functions/client/src/main.rs
rbehjati: Done.

oak_functions/examples/key_value_lookup/module/src/tests.rs
rbehjati: Declared a parameter instead. BTW, is tee_certificate that is used for creating `AttestationServer` ever used?

oak_functions/loader/src/grpc.rs
rbehjati: Done.

rbehjati (commented):
Thanks for the review @ipetr0v.

experimental/grpc_attestation/src/main.rs
```diff
@@ -82,10 +81,10 @@ pub struct Opt {
 // Example message sent to the client.
 const EXAMPLE_MESSAGE: &str = "Example message";
 
-fn request_handler(request: &[u8]) -> Vec<u8> {
-    let parsed_request = String::from_utf8(request.to_vec()).expect("Couldn't parse request");
+async fn request_handler(request: Vec<u8>) -> anyhow::Result<Vec<u8>> {
+    let parsed_request = String::from_utf8(request.to_vec()).context("Couldn't parse request")?;
```
conradgrobler, line 85: Is copying the request into a new vector required here?

oak_functions/loader/src/grpc.rs
```diff
@@ -81,13 +69,16 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
         ),
     );
 
+    let request_handler =
+        async move |request: Vec<u8>| handle_request(wasm_handler, policy, request.to_vec()).await;
```
conradgrobler, line 73: Must the request be copied here as well?

experimental/grpc_attestation/src/main.rs
```diff
@@ -82,10 +81,10 @@ pub struct Opt {
 // Example message sent to the client.
 const EXAMPLE_MESSAGE: &str = "Example message";
 
-fn request_handler(request: &[u8]) -> Vec<u8> {
-    let parsed_request = String::from_utf8(request.to_vec()).expect("Couldn't parse request");
+async fn request_handler(request: Vec<u8>) -> anyhow::Result<Vec<u8>> {
```
conradgrobler, line 84: Is there a reason for switching from `&[u8]` to `Vec<u8>`?

conradgrobler (approved):

experimental/grpc_attestation/src/main.rs
rbehjati: This is because I changed the type of `handler` in `RequestHandler`. The signature of the closure now requires `Vec<u8>`. I could not make it work with `&[u8]`. There is either an issue with lifetimes, or unsatisfying the `Clone` bound.

experimental/grpc_attestation/src/main.rs
rbehjati: No. Fixed.

tiziano88:
I'm going to take a look shortly

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -48,15 +48,19 @@ impl Receiver {
...
     /// Handler function that processes data from client requests and creates responses.
-    handler: fn(&[u8]) -> Vec<u8>,
+    handler: F,
 }
 
-impl RequestHandler {
-    pub fn new(encryptor: AeadEncryptor, handler: fn(&[u8]) -> Vec<u8>) -> Self {
+impl<F, S> RequestHandler<F>
+where
+    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
```
tiziano88, line 60: Would it make sense to make this a Fn (not FnOnce) and not Clone?

oak_functions/loader/src/main.rs
```diff
@@ -66,6 +60,10 @@ struct Config {
     ///
     /// See https://docs.rs/tokio/1.5.0/tokio/runtime/struct.Builder.html#method.worker_threads.
     worker_threads: Option<usize>,
+    /// PEM encoded X.509 certificate that signs TEE firmware key. This field must be specified.
```
tiziano88, line 63: This is the path to the certificate file, not the literal certificate, right? Could you clarify in the name (e.g. _file or _path)

scripts/deploy_oak_functions_api
```diff
@@ -20,7 +20,7 @@ readonly OUT_DIR='./oak_functions/bin'
 mkdir --parents "${OUT_DIR}"
 
 cp \
-  ./oak_functions/proto/server.proto \
+  ./remote_attestation/proto/remote_attestation.proto \
```
tiziano88, line 23: I am expecting that the content of the remote attestation proto would move into oak_functions/proto/server.proto. Effectively that is the new interface to Oak Functions from now on.

oak_functions/examples/key_value_lookup/config.toml
```diff
@@ -1,6 +1,7 @@
 lookup_data_url = "https://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
 lookup_data_download_period = "10m"
 worker_threads = 4
+tee_certificate = "./examples/certs/local/ca.pem"
```
tiziano88, line 4: Does this need to be added to some docker file, or am I missing how it is going to work?

tiziano88, line 4: I don't think this certificate is the TEE certificate, is it? AFAICT it's just the root cert that we use for testing TLS connections.

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -48,15 +48,19 @@ impl Receiver {
     }
 }
 
-struct RequestHandler {
+struct RequestHandler<F> {
```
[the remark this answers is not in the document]
rbehjati: Done.

experimental/grpc_attestation/src/attestation.rs
rbehjati: It won't work. `Fn` means the closure must be `async` without `move`. Currently, such async functions cannot have parameters. But we need the closure to also take a parameter.

experimental/grpc_attestation/src/attestation.rs
```diff
@@ -48,15 +48,19 @@ impl Receiver {
...
-    handler: fn(&[u8]) -> Vec<u8>,
+    handler: F,
 }
 
-impl RequestHandler {
-    pub fn new(encryptor: AeadEncryptor, handler: fn(&[u8]) -> Vec<u8>) -> Self {
+impl<F, S> RequestHandler<F>
+where
+    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
+    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync,
```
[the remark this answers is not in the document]
rbehjati: I am not sure how to make it work. The type alias must be `dyn`, but with that I get the `doesn't have a size known at compile-time` error. How can that be fixed? I've tried various combinations of `Box`, `Pin`, and even `?Sized` without luck!

oak_functions/examples/key_value_lookup/config.toml
rbehjati: Good point, I've created #2141 to fix the Cloud Run deployment. 
I think we need something similar to the following: https://github.com/project-oak/oak/blob/c6c6e4a7974228bf05e5ebb4f5151fde41b1d7f1/examples/aggregator/scripts/deploy#L41-L44

rbehjati: I have copied this parameter from `experimental/grpc_attestation/src/main.rs`. 
`tee_certificate` is needed for creating an instance of `AttestationServer`, but I don't quite understand what is does. @ipetr0v Could you please clarify what it is for?

oak_functions/loader/src/main.rs
rbehjati: Done.

scripts/deploy_oak_functions_api
rbehjati: Created #2141, but I am not sure why would `remote_attestation` have a dependency to `oak_functions`. Is remote attestation specific to Oak Functions? My understanding, from the discussion on Slack, was that `remote_attestation` is going to be the more general handshake protocol.

oak_functions/examples/key_value_lookup/config.toml
ipetr0v: It's just currently used as a test certificate for TEE purposes. When we will have remote attestation, we will be able to use authentic certificates instead of this.

PR #2137, opened 2021-05-26
Update .gitignore file

ipetr0v:
This commit adds `.cache` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,6 @@
 # Generated vscode files.
 **/.vscode/
 
-# History files.
-/.bash_history
-
 # emacs stuff
 *~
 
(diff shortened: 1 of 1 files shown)


tiziano88 (approved):
Not sure why bash history would be there

ipetr0v:
If it's not in `.gitignore`, it will always appear in `Untracked files`

tiziano88:
But how did it end up there?

ipetr0v:
If wasn't specified by `.gitignore`
Do you mean that Git should automatically ignore this file?

ipetr0v:
Ah, you mean the file?
hm, it looks like bash in the Android Docker creates it

ipetr0v:
Removing `.bash_history`

tiziano88 (approved):

Issue #2136, opened 2021-05-26
Fix logging in oak_functions/client


rbehjati:
The string formatting in the following line does not work: https://github.com/project-oak/oak/blob/c34a0abccc80b91c949abd53e801c2bb20530839/oak_functions/client/java/src/AttestationClient.java#L53

I think, this formatting only works with [SLF4J](http://www.slf4j.org/manual.html) or some manual formatter libraries.

rbehjati:
@ipetr0v Could you please take a look?

ipetr0v:
Since this library has only 2 calls to the `Logger` - removing it completely and use `System.out.printf`

rbehjati:
Alternatively, since the logs are simple, you could keep the `logger` and build the string using the `+` operator,  instead of using `System.out.printf`:
```java
logger.log(Level.WARNING, "Couldn't receive response: " + status); 
```

PR #2138, opened 2021-05-26
Fix Java logging in gRPC Attestation

ipetr0v:
Fixes https://github.com/project-oak/oak/issues/2136

### DIFF
--- a/oak_functions/client/java/src/AttestationClient.java
+++ b/oak_functions/client/java/src/AttestationClient.java
@@ -10,8 +10,6 @@ import io.grpc.stub.StreamObserver;
 import java.lang.RuntimeException;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import oak.remote_attestation.AttestedInvokeResponse;
 import oak.remote_attestation.AttestedInvokeRequest;
 import oak.remote_attestation.ClientIdentity;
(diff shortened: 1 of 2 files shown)


oak_functions/client/java/src/AttestationClient.java
```diff
@@ -43,14 +40,14 @@ public void onNext(AttestedInvokeResponse response) {
                 try {
                     messageQueue.put(response);
                 } catch (Exception e) {
-                    logger.log(Level.WARNING, "Couldn't send server response to the message queue: {0}", e);
+                    System.out.printf("Couldn't send server response to the message queue: %s\n", e.toString());
```
rbehjati, line 43: Is the call to `toString` really needed?

rbehjati (approved):

oak_functions/client/java/src/AttestationClient.java
ipetr0v: Yes, in order to print it via `%s`.
IIUC Java cannot print raw Objects using `System.out.printf` (but please correct me if I'm wrong)

rbehjati: I could answer this question a year ago. My java is very rusty now :D

tiziano88:
I don't think we should remove proper logging support, especially since this library is going to be used in Android and other places which expect canonical logging formats. 

Definitely it should not pritn to stdout anyways, at most to stderr.

ipetr0v:
Updated and returned `Logger` back.

tiziano88:
Why did the hashes for oak_loader and oak_functions_loader change?

conradgrobler:
> Why did the hashes for oak_loader and oak_functions_loader change?

There were two earlier dependabot updates where the workflow to update these failed. This was the first successful run on `main` since then: https://github.com/project-oak/oak/actions/workflows/reproducibility.yaml

PR #2139, opened 2021-05-26
Change deploy script for Oak Functions to run outside docker

rbehjati:
I could not run this script without docker, because of missing Rust crates, and GCP commands don't work when running inside docker.

This change list also removes (no longer applicable) mentions of #1943, which was merged a long time ago. But we still need to build the loader binary in the script.

### DIFF
--- a/scripts/deploy_oak_functions_loader
+++ b/scripts/deploy_oak_functions_loader
@@ -7,8 +7,7 @@ source "$SCRIPTS_DIR/common"
 # shellcheck source=scripts/gcp_common
 source "$SCRIPTS_DIR/gcp_common"
 
-# TODO(#1943): Remove when #1943 is submitted.
-cargo build --target=x86_64-unknown-linux-musl --manifest-path=./oak_functions/loader/Cargo.toml --release
+./scripts/docker_run cargo build --target=x86_64-unknown-linux-musl --manifest-path=./oak_functions/loader/Cargo.toml --release
 
 # Build and push base Oak Functions Loader image.
 
@@ -22,7 +21,7 @@ docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
 # Build and push the `weather_lookup` example application built on Oak Functions.
 
-cargo -Zunstable-options build --release \
+./scripts/docker_run cargo -Zunstable-options build --release \
   --target=wasm32-unknown-unknown \
   --manifest-path=./oak_functions/examples/weather_lookup/module/Cargo.toml \
   --out-dir=./oak_functions/examples/weather_lookup/bin
@@ -62,7 +61,7 @@ gcloud beta run deploy "${FUNCTIONS_INSTANCE_NAME}" \
 readonly HOST_NAME="$(gcloud api-gateway gateways describe "${GATEWAY_ID}" --location="${GCP_REGION}" --format='value(defaultHostname)')"
 
 #Build the client to test the connection.
-cargo build --manifest-path=./oak_functions/client/Cargo.toml
+./scripts/docker_run cargo build --manifest-path=./oak_functions/client/Cargo.toml
 
 # Attempt to communicate with the newly deployed application.
 ./oak_functions/client/target/debug/oak_functions_client \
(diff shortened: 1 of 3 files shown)


scripts/deploy_oak_functions_loader
```diff
@@ -7,8 +7,7 @@ source "$SCRIPTS_DIR/common"
 # shellcheck source=scripts/gcp_common
 source "$SCRIPTS_DIR/gcp_common"
 
-# TODO(#1943): Remove when #1943 is submitted.
```
tiziano88, line 10: The point of the TODO is to actually use a runner command when implemented, instead of cargo build (not just to remove the TODO) :slightly_smiling_face: (also below)

scripts/deploy_oak_functions_loader
```diff
@@ -22,7 +21,7 @@ docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
 # Build and push the `weather_lookup` example application built on Oak Functions.
 
-cargo -Zunstable-options build --release \
+./scripts/docker_run cargo -Zunstable-options build --release \
```
tiziano88, line 24: This is an improvement, but it's still not ideal, maybe we should split the script into two, one that builds all the things (and is expected to be run inside docker), and another just to build the docker images and deploy on GCP. Anyways we can try it out separately.

tiziano88 (approved):

scripts/deploy_oak_functions_loader
rbehjati: Ah! Fixed.

scripts/deploy_oak_functions_loader
rbehjati: I thought about splitting it into two, but I think it is good to have a single script that does all this. On the other hand, when all the artefacts are already built, it is easier to have a script that only builds the images and deploys to GCP. Let's fix that later.

scripts/deploy_oak_functions_loader
```diff
@@ -22,7 +22,7 @@ docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
...
   --target=wasm32-unknown-unknown \
   --manifest-path=./oak_functions/examples/weather_lookup/module/Cargo.toml \
   --out-dir=./oak_functions/examples/weather_lookup/bin
```
tiziano88, line 28: Isn't there a runner command for this too?

scripts/deploy_oak_functions_loader
```diff
@@ -62,7 +62,7 @@ gcloud beta run deploy "${FUNCTIONS_INSTANCE_NAME}" \
 readonly HOST_NAME="$(gcloud api-gateway gateways describe "${GATEWAY_ID}" --location="${GCP_REGION}" --format='value(defaultHostname)')"
 
 #Build the client to test the connection.
-cargo build --manifest-path=./oak_functions/client/Cargo.toml
+./scripts/docker_run cargo build --manifest-path=./oak_functions/client/Cargo.toml
```
tiziano88, line 65: And for this?

tiziano88 (approved):

scripts/deploy_oak_functions_loader
rbehjati: I don't think we have `runner` commands for building `.wasm` files. It is being done as part of the `run-functions-examples`. Should I just call the `run-functions-examples`, and even skip creating the docker image for the loader?

scripts/deploy_oak_functions_loader
rbehjati: We don't have a separate command for running clients. 
I could use `./scripts/runner --logs run-functions-examples --example-name=weather_lookup --run-server=false`. But I am not sure if it is possible to correctly pass `HOST_NAME` to the client. Also, do we want to get all the logs from the runner in the output?

tiziano88: This is just the build step, can we run it with --run-server=false and --run-client=false, or something like that? Just to get the client binary built.

PR #2132, opened 2021-05-25
Move Java gRPC Attestation client to Oak Functions

ipetr0v:
This change moves gRPC Attestation client from `experimental/grpc_attestation_client/java` to `oak_functions/client/java`.

PR #2143, opened 2021-05-27
Reorganize Oak Functions protos and clients

rbehjati:
Follow up of #2122 
* Moves `remote_attestation` protos to `oak_functions` (ref https://github.com/project-oak/oak/pull/2122#discussion_r639860414)
* Updates `oak_functions` Java client
* Adds a Java client for the `weather_lookup` example
* Removes `examples/grpc_attestation` (ref https://github.com/project-oak/oak/pull/2132#discussion_r638910712)
* Moves the Oak Functions Rust client to `rust` folder


oak_functions/proto/server.proto
```diff
@@ -16,17 +16,23 @@
...
+package oak.functions.server;
 
 option java_multiple_files = true;
-option java_package = "oak.remote_attestation";
+option java_package = "oak.functions.server";
+
+// Wrapper around the encrypted payload.
+message SecureRequest {
+  // Payload encrypted using the session key.
+  bytes encrypted_payload = 1;
```
tiziano88, line 27: @ipetr0v FYI in your PR I meant basically adding the separate IV as a bytes field to this message, instead of concatenating everything in a single byte field.

ipetr0v: Got it
Will update once as part of synchronization with the handshaker

oak_functions/client/java/src/AttestationClient.java
```diff
@@ -99,17 +102,22 @@ protected void finalize() throws Throwable {
     /**
      * Encrypts and sends `message` via an attested gRPC channel to the server.
      */
-    public byte[] Send(byte[] message) throws Exception {
-        byte[] encryptedMessage = encryptor.encrypt(message);
-        AttestedInvokeRequest request = AttestedInvokeRequest
+    public Response Send(Request request) throws Exception {
```
ipetr0v, line 105: Why can't `Request` object be created inside `Send` ?
I think it's much simpler to have an API that accepts `byte[]`

rbehjati: I think, the plan is to extend `Request`, and `SecureRequest` with additional fields, for example a Policy. In that case it is easier to work with `Request` instead of a byte array. I could send the binary encoded protobuf message, but then `send` would have to decode that to `Request` to construct `SecureRequest` from it.

This is because, these additional fields are going to be public (unencrypted). Otherwise, we could just work with the binary encoded protobuf.

oak_functions/proto/server.proto
```diff
@@ -16,17 +16,25 @@
...
+package oak.functions.server;
 
 option java_multiple_files = true;
-option java_package = "oak.remote_attestation";
+option java_package = "oak.functions.server";
+
+// Wrapper around the encrypted payload.
+// TODO(#2144): Extend this with additional potentially public fields, for instance for specifying
+// policies.
+message SecureRequest {
```
tiziano88, line 27: I don't really like the "secure" name in this context, feel free to submit but let's keep thinking about this.

cc @project-oak/core in case anyone has any suggestion

tiziano88 (approved):

oak_functions/proto/server.proto
rbehjati: `RequestWithEncryptedPayload` is more accurate, but that is very long name.

oak_functions/client/java/src/AttestationClient.java
rbehjati: @ipetr0v I ended up using `Request` in more places, particularly in `experimental/grpc_attestation`. I believe when we have the generic state machine, all that will be moved into `oak_functions`.

oak_functions/proto/server.proto
tiziano88: I think we should try to rename this to just `Request` eventually, and find some other name for `AttestedInvokeRequest`, e.g. something with `wrapper` or `frame` or `handshake` or something else.

ipetr0v (approved):

oak_functions/proto/server.proto
rbehjati: Done.

ipetr0v:
Since `examples/grpc_attestation` is now removed. Is there another example that runs Java Attestation Client for the CI ?

ipetr0v:
Ah, it was moved to Weather Lookup

PR #2155, opened 2021-05-31
Minor Java Attestation Client refactoring

ipetr0v:
This change:
- Moves Java Attestation client source from `src` to `com/google/oak/functions/client/`
- Removes `Util` class
- Updates thrown exceptions


oak_functions/client/java/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -38,13 +40,16 @@ public AttestationClient(String uri) throws Exception {
         RemoteAttestationStub stub = RemoteAttestationGrpc.newStub(channel);
 
         // Create server response handler.
-        messageQueue = new ArrayBlockingQueue<AttestedInvokeResponse>(1);
+        messageQueue = new ArrayBlockingQueue<>(1);
         StreamObserver<AttestedInvokeResponse> responseObserver = new StreamObserver<AttestedInvokeResponse>() {
```
rbehjati, line 44:
```suggestion
        StreamObserver<AttestedInvokeResponse> responseObserver = new StreamObserver<>() {
```
Could `AttestedInvokeResponse` be removed here too?

oak_functions/client/java/BUILD
```diff
@@ -24,10 +24,9 @@ package(
 java_library(
     name = "client",
     srcs = [
-        "src/AeadEncryptor.java",
-        "src/AttestationClient.java",
-        "src/KeyNegotiator.java",
-        "src/Util.java",
+        "com/google/oak/functions/client/AeadEncryptor.java",
```
rbehjati, line 27: Don't java projects normally have a `src` folder as well? (or would that violate some g3 file structure requirement?) I mean could not all this still be in a `src` folder, e.g., "src/com/google/oak/functions/client/AeadEncryptor.java"?

rbehjati (approved):
LGTM!

oak_functions/client/java/com/google/oak/functions/client/AttestationClient.java
ipetr0v: I think no.
The type should be specified at least once. Otherwise it would become a raw Java type.

oak_functions/client/java/BUILD
ipetr0v: Done

oak_functions/client/java/com/google/oak/functions/client/AttestationClient.java
rbehjati: I think since the type is specified in the declaration of the variable, it is not necessary in the instantiation statement.

ipetr0v: Done

rbehjati: Ah! I did not realize this is an anonymous type implementation. Both types are needed. I have created a PR to fix this: #2159

Issue #2103, opened 2021-05-18
Refactor gRPC Attestation protocol state machine


ipetr0v:
Currently https://github.com/project-oak/oak/pull/2070 implements gRPC Attestation protocol with a state machine that is represented as simple if statements.
It should be refactored in a more Rust idiomatic way.

ipetr0v:
Resources for clean state machines in Rust: https://hoverbear.org/blog/rust-state-machine-pattern/

PR #2156, opened 2021-06-01
Remote attestation refactoring

ipetr0v:
This change:
- Moves gRPC attestation from experimental
- Creates a protocol independent state-machine for attestation in Rust
- Adds AEAD nonces to protobuf messages

Fixes https://github.com/project-oak/oak/issues/2103


remote_attestation/java/src/com/google/oak/remote_attestation/crypto/AeadEncryptor.java
```diff
@@ -50,14 +51,14 @@ public AeadEncryptor(byte[] key) {
 
     /**
      * Encrypts `data` using AES-GCM.
-     * The resulting encrypted data is prefixed with a random 12 byte initialization vector.
+     * The resulting encrypted data contans a random 12 byte nonce.
```
conradgrobler, line 54: s/contans/contains/

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/AeadEncryptor.java
```diff
@@ -67,40 +68,40 @@ public AeadEncryptor(byte[] key) {
         // single encrypted field.
         byte[] encryptedData = encryptor.doFinal(data);
 
-        // Add `initializationVector` as a prefix to the `encryptedData`.
-        return Bytes.concat(initializationVector, encryptedData);
+        // Add `nonce` as a prefix to the `encryptedData`.
```
conradgrobler, line 71: nit: This comment is no longer accurate, as it is not a prefix but part of the object.

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/AeadEncryptor.java
```diff
@@ -67,40 +68,40 @@ public AeadEncryptor(byte[] key) {
...
-     * `data` must contain an encrypted message prefixed with a 12 byte initialization vector.
      */
-    public byte[] decrypt(byte[] data) throws GeneralSecurityException {
+    public byte[] decrypt(EncryptedData data) throws GeneralSecurityException {
         Cipher decryptor = Cipher.getInstance(AEAD_ALGORITHM);
 
-        // Extract initialization vector from `data`.
-        byte[] initializationVector = Arrays.copyOf(data, INITIALIZATION_VECTOR_LENGTH_BYTES);
-        GCMParameterSpec gcmParameterSpecification = new GCMParameterSpec(TAG_LENGTH_BITS, initializationVector);
+        // Extract nonce from `data`.
```
conradgrobler, line 84: I don't think this comment is needed as extracting the nonce is no longer a complicated step.

conradgrobler (approved):

oak_functions/client/rust/src/attestation.rs
```diff
@@ -82,34 +81,42 @@ impl Client {
...
 impl AttestationClient {
     pub async fn create(uri: &str, expected_tee_measurement: &[u8]) -> anyhow::Result<Self> {
         let mut client = Client::create(uri)
             .await
             .context("Couldn't create gRPC client")?;
-        let encryptor = AttestationClient::attest(&mut client, expected_tee_measurement)
-            .await
-            .context("Couldn't attest server")?;
+        let attested_state_machine =
+            AttestationClient::attest(&mut client, expected_tee_measurement)
```
rbehjati, line 93: Could this be `Self::attest`? I think that reads better. And again, IIUC, `create_session` fits better.

oak_functions/client/rust/src/attestation.rs
```diff
@@ -82,34 +81,42 @@ impl Client {
 /// gRPC Attestation Service client implementation.
 pub struct AttestationClient {
     client: Client,
-    encryptor: AeadEncryptor,
+    attested_state_machine: AttestationStateMachine<AttestedState>,
```
rbehjati, line 84: Why does this struct keep an instance of the `AttestationStateMachine`? `AttestedState` is the final state of the state machine, so the `encrypt` and `decrypt` methods don't need to be defined on the state machine, they could be directly defined on `AttestedState`; and `AttestationClient` could simply keep an instance of `AttestedState`. 

Please also see my other comments on `AttestationStateMachine`.

oak_functions/loader/src/attestation.rs
```diff
@@ -54,7 +51,7 @@ where
     S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync,
 {
     /// Utility object for decrypting and encrypting messages using a Diffie-Hellman session key.
-    encryptor: AeadEncryptor,
+    attested_state_machine: AttestationStateMachine<AttestedState>,
```
rbehjati, line 54: Again, I think just having an instance of `AttestedState` should work.

remote_attestation/rust/src/attestation.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+pub struct AttestedState {
+    /// Utility object for decrypting and encrypting messages using a Diffie-Hellman session key.
+    encryptor: AeadEncryptor,
+}
+
+impl AttestationStateMachine<AttestedState> {
+    pub fn encrypt(&mut self, data: &[u8]) -> anyhow::Result<EncryptedData> {
```
rbehjati, line 159: Since the return types here and in `decrypt` are not `AttestationStateMachine`, I don't think these functions need to be defined on `AttestationStateMachine`.

rbehjati (commented):
Thanks @ipetr0v. I have some general comments on the design of the state machine, and naming. But I could be missing some details.

oak_functions/sdk/Cargo.lock
```diff
@@ -692,18 +639,23 @@ dependencies = [
 ]
 
 [[package]]
-name = "oak_attestation_common"
+name = "oak_abi"
```
rbehjati, line 642: I am very skeptical of this dependency. Could you please double check that `oak_abi` is needed?

oak_functions/sdk/Cargo.lock
```diff
@@ -692,18 +639,23 @@ dependencies = [
...
+ "prost-build 0.7.0",
+ "prost-types 0.7.0",
  "serde",
- "serde_json",
- "sha2",
- "x509-parser",
+]
+
+[[package]]
+name = "oak_derive"
```
rbehjati, line 653: Also this one.

remote_attestation/rust/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+authors = ["Ivan Petrov <ivanpetrov@google.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[dependencies]
+anyhow = "*"
+env_logger = "*"
+log = "*"
+oak_io = { path = "../../oak_io" }
```
rbehjati, line 12: Is this dependency really needed? It is bringing `oak_abi`, `oak_derive` and `oak_io` dependencies into Oak Functions which does not sound quite right to me.

oak_functions/client/rust/src/attestation.rs
ipetr0v: `AttestedState` was changes to `AttestedSession`, and now it's not an attestation state.

oak_functions/loader/src/attestation.rs
ipetr0v: Now it's an `AttestedSession`

remote_attestation/rust/src/attestation.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// PEM encoded X.509 certificate that signs TEE firmware key.
+    tee_certificate: Option<Vec<u8>>,
+    /// Expected value of the peer's TEE measurement.
+    expected_tee_measurement: Option<Vec<u8>>,
+    /// Implementation of the X25519 Elliptic Curve Diffie-Hellman (ECDH) key negotiation.
+    key_negotiator: KeyNegotiator,
+}
+
+impl AttestationStateMachine<UnattestedState> {
+    pub fn create_self_attestation(tee_certificate: &[u8]) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
ipetr0v: Done
Added comments

remote_attestation/rust/src/attestation.rs
```diff
@@ -0,0 +1,176 @@
...
+            .context("Couldn't find report in peer attestation info")?;
+        if expected_tee_measurement == peer_report.measurement {
+            Ok(())
+        } else {
+            Err(anyhow!("Incorrect TEE measurement"))
+        }
+    }
+}
+
+pub struct AttestedState {
```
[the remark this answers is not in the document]
ipetr0v: Renamed to `AttestedSession`

oak_functions/loader/src/attestation.rs
```diff
@@ -134,8 +138,11 @@ where
     /// Attest a single gRPC streaming request. Client messages are provided via `receiver`.
     async fn attest(
```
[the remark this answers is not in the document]
ipetr0v: This function may be called in multiple threads.
Moved outside of `impl` block.

remote_attestation/rust/Cargo.toml
conradgrobler: This is due to using oak_utils to compile the proto files. You should compile the protos using tonic-build directly, as using oak_utils it adds a bunch of Oak dependencies related to handles that we don't need in Oak functions.

remote_attestation/rust/build.rs
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    oak_utils::compile_protos(&["../proto/remote_attestation.proto"], &["../proto"]);
```
rbehjati, line 18: This adds derive `HandleVisit` annotation for protobuf messages in `remote_attestation.proto`. We don't need that. Could you replace this with `prost_build::compile_proto(...)` or `oak_utils::compile_protos_with_option` and passing a `ProtoOptions` with `derive_handle_visit` set to false, and try again without adding a dependency to `oak_io`?

remote_attestation/rust/Cargo.toml
ipetr0v: Done

remote_attestation/rust/build.rs
ipetr0v: Thanks!
Done

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,151 @@
...
+        None
+    }
+
+    fn verify_attestation(self, peer_identity: &AttestationIdentity) -> anyhow::Result<()> {
+        // Verify peer attestation info.
+        let peer_attestation_info = peer_identity
+            .attestation_info
+            .as_ref()
+            .context("Peer identity doesn't contain attestation info")?;
+        verify_attestation(&peer_attestation_info, &self.expected_tee_measurement)
```
rbehjati, line 99: nit: if possible, perhaps pass `peer_identity` to `verify_attestation`, and let it extract `peer_attestation_info` from `peer_identity`.

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,151 @@
...
+
+impl UnattestedPeer {
+    pub fn new(expected_tee_measurement: &[u8]) -> Self {
+        Self {
+            expected_tee_measurement: expected_tee_measurement.to_vec(),
+        }
+    }
+}
+
+/// Represents an attestation process, where current machine attests and a remote peer remotely
```
rbehjati, line 50:
```suggestion
/// Represents an attestation process, where current machine and a remote peer remotely
```

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,151 @@
...
+/// attest each other.
+pub struct MutuallyUnattested {
```
rbehjati, line 52: This is still not used anywhere. When is it going to be used?

rbehjati (approved):
Looks good to me. Thanks. You might still want to wait for @tiziano88 to take a look too.

remote_attestation/rust/src/attestation/unattested_session.rs
ipetr0v: It will be used once we support bidirectional attestation.

rbehjati: I see. Thanks.

remote_attestation/rust/src/attestation/unattested_session.rs
ipetr0v: Done

remote_attestation/rust/src/attestation/unattested_session.rs
ipetr0v: Done

oak_functions/loader/src/attestation.rs
```diff
@@ -226,3 +177,39 @@ where
...
+    let request = receiver
+        .receive()
+        .await
+        .context("Couldn't receive attestation request")?
+        .context("Stream stopped preemptively")?;
+
+    // Receive client's attestation request containing a public key.
+    let request_type = request.request_type.context("Couldn't read request type")?;
+    let client_identity = if let RequestType::ClientIdentity(request) = request_type {
+        Ok(request)
```
tiziano88, line 195: I don't think this should be wrapped in an Ok just to be unwrapped immediately. I would just make this branch return `request`, and use `anyhow::bail` or an explicit return in the other branch. Up to you.

remote_attestation/java/src/com/google/oak/remote_attestation/AttestationStateMachine.java
```diff
@@ -19,40 +19,65 @@
...
+                    .build();
+            AttestationInfo attestationInfo = AttestationInfo.newBuilder()
+                    .setReport(report)
+                    .build();
+
+            AttestationIdentity identity = AttestationIdentity.newBuilder()
+                    .setPublicKey(ByteString.copyFrom(keyNegotiator.getPublicKey()))
+                    .setAttestationInfo(attestationInfo)
+                    .build();
+            return identity;
```
tiziano88, line 59: I'm confused by the function comment; in what way is this identity a public key?

remote_attestation/proto/remote_attestation.proto
```diff
@@ -0,0 +1,68 @@
...
+
+package oak.remote_attestation;
+
+option java_multiple_files = true;
+option java_package = "oak.remote_attestation";
+
+/// Placeholder implementation of TEE report for remote attestation.
+/// https://www.amd.com/system/files/TechDocs/56860.pdf#page=39
+///
+/// TODO(#1867): Add remote attestation support and use real TEE reports.
```
tiziano88, line 27: Note that in practice a claim / assertion is probably just going to be an opaque byte blob as returned by the workload attestation, and it will have various layers of attestations. Anyways, we will see when we get there.

remote_attestation/rust/src/attestation/attested_session.rs
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+//
+
+use crate::{crypto::AeadEncryptor, proto::EncryptedData};
+use anyhow::Context;
+
+pub struct AttestedSession {
+    /// Utility object for decrypting and encrypting messages using a Diffie-Hellman session key.
+    encryptor: AeadEncryptor,
+}
```
tiziano88, line 23: Not sure I understand the point of this wrapper vs just using AeadEncryptor, could you add a comment? Is it just for error handling?

remote_attestation/rust/src/attestation/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+            MutuallyUnattested, UnattestedPeer, UnattestedSelf, UnattestedSession,
+        },
+    },
+    crypto::{get_sha256, AeadEncryptor, KeyNegotiator},
+    proto::AttestationIdentity,
+};
+use anyhow::Context;
+
+/// Remote attestation protocol implementation.
+pub struct AttestationEngine<S> {
```
tiziano88, line 33: Could you add a comment about S?

remote_attestation/rust/src/attestation/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+    /// Type of the remote attestation process.
+    attestation_type: S,
+    /// Implementation of the X25519 Elliptic Curve Diffie-Hellman (ECDH) key negotiation.
+    key_negotiator: KeyNegotiator,
+}
+
+impl<S> AttestationEngine<S>
+where
+    S: UnattestedSession,
```
tiziano88, line 42: Should this type bound be on the struct definition too, for clarity?

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,146 @@
...
+        // Generate attestation info with a TEE report.
+        // TEE report contains a hash of the public key.
+        Some(AttestationInfo {
+            report: Some(AttestationReport::new(&public_key_hash)),
+            certificate: self.tee_certificate.clone(),
+        })
+    }
+
+    fn verify_attestation(self, _peer_identity: &AttestationIdentity) -> anyhow::Result<()> {
+        Ok(())
```
tiziano88, line 84: I'm not sure whether this is the final version or just a placeholder, or anyways what this method is supposed to return. Could you add more comments on the trait definition to explain what the methods do and how they are supposed to be implemented and by whom?

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn new(tee_certificate: &[u8], expected_tee_measurement: &[u8]) -> Self {
+        Self {
+            tee_certificate: tee_certificate.to_vec(),
+            expected_tee_measurement: expected_tee_measurement.to_vec(),
+        }
+    }
+}
+
+pub trait UnattestedSession {
+    fn get_attestation_info(&self, public_key_hash: &[u8]) -> Option<AttestationInfo>;
```
tiziano88, line 69: Does "get" mean "generate" here? If so, could you rename it? I think we used to use the terminology assertion generator / verifier, is this what this is supposed to be too? And is this method dual of the method below? If so, could you clarify why they use AttestationInfo and AttestationIdentity respectively?

remote_attestation/rust/src/attestation/unattested_session.rs
```diff
@@ -0,0 +1,146 @@
...
+) -> anyhow::Result<()> {
+    // Verify peer attestation info.
+    // TODO(#1867): Add remote attestation support, use real TEE reports and check that
+    // `AttestationInfo::certificate` is signed by one of the root certificates.
+    let peer_attestation_info = peer_identity
+        .attestation_info
+        .as_ref()
+        .context("Peer identity doesn't contain attestation info")?;
+    peer_attestation_info
+        .verify()
```
tiziano88, line 133: I'm not sure an attestation info (whatever it ends up being) can "verify itself". Should it not be verified by the caller by checking its signature? I know there is a TODO, but this verify method here seems confusing to me, or maybe I'm missing something.

tiziano88 (approved):

remote_attestation/rust/src/attestation/unattested_session.rs
ipetr0v: Attestation info has both report and corresponding provider's certificate in it.
So it can check the validity of the report without additional arguments.

tiziano88: What's the point of that? The peer can just make up a key and use that to sign whatever it wants, but that's meaningless. We need to check that it was ultimately signed by a cert / key that the client trusts, not just whatever the peer sends over, right?

PR #2160, opened 2021-06-02
Format AttestationCleint.java

rbehjati:
More formatting and linting to make g3 happy.

### DIFF
--- a/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
+++ b/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
@@ -1,15 +1,12 @@
 package com.google.oak.functions.client;
 
 import com.google.common.base.VerifyException;
-import com.google.oak.functions.client.AeadEncryptor;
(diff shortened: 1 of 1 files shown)


ipetr0v (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -1,15 +1,12 @@
 package com.google.oak.functions.client;
 
 import com.google.common.base.VerifyException;
-import com.google.oak.functions.client.AeadEncryptor;
-import com.google.oak.functions.client.KeyNegotiator;
```
ipetr0v, line 5: Why are these lines removed? Can Java see classes from the same package by default?

ipetr0v (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
rbehjati: Yes. So these imports are redundant.

Issue #2121, opened 2021-05-24
Implement a protocol independent state machine in Java gRPC Attestation


ipetr0v:
Currently Java gRPC Attestation client is tied to gRPC. Though it should also support other protocols, such as HTTP.
So we need to wrap key negotiation logic into a separate state machine, which just processes responses and creates requests.

Original coment: https://github.com/project-oak/oak/pull/2112/files#r637876481

PR #2161, opened 2021-06-02
Refactor Java attestation client

ipetr0v:
This change:
- Creates a protocol independent attestation implementation in Java
- Adds an option to use Tink as a crypto library

Fixes https://github.com/project-oak/oak/issues/2121


tiziano88:
Is the plan to keep maintaining both the tink version and the javax.crypto version? I think we should only have one, and if tink seems to be the one we need in any case, then we should only have that. WDYT?

ipetr0v:
I think it's better to keep two versions until we a 100% what we are going to use internally.
And after we test that it works, we can choose the main version.

tiziano88:
Ok got it, makes sense, thanks for the info

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -20,7 +22,6 @@
 import oak.functions.server.RemoteAttestationGrpc;
 import oak.functions.server.RemoteAttestationGrpc.RemoteAttestationStub;
 
-// TODO(#2121): Implement a protocol independent state machine.
```
rbehjati, line 23: I am not sure this PR actually implements a protocol-agnostic state machine. What are the states and the transitions? I thought the state machine was going to be for the entire attestated invocation process, with states such as `ready to exchange keys` -> ... ->  `ready to accept a request` -> `encrypted request` -> ... -> `Done (response sent)`.

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/KeyNegotiator.java
```diff
@@ -1,7 +1,6 @@
-package com.google.oak.functions.client;
+package com.google.oak.remote_attestation.crypto;
```
rbehjati, line 1: pre-existing: this class has a field named `keyPair`. I think it can be `private final`.

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/tink/AeadEncryptor.java
```diff
@@ -0,0 +1,39 @@
+package com.google.oak.remote_attestation.crypto;
```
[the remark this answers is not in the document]
ipetr0v: Specifically here it is temporary, because since we currently have 2 different implementations, we need to do conditional compilation (that Java doesn't support, but it can be done via Bazel).

oak_functions/client/java/BUILD
```diff
@@ -37,6 +35,31 @@ java_library(
...
+        "@com_google_protobuf//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java_util",
+        "@io_grpc_grpc_java//api",
+        "@io_grpc_grpc_java//protobuf",
+        "@io_grpc_grpc_java//stub",
+    ],
+)
+
+java_library(
+    name = "client_tink",
```
rbehjati, line 49: It is not currently used anywhere. Is it possible to use it in the weather_lookup example?

ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/KeyNegotiator.java
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/crypto/tink/AeadEncryptor.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Implementation of Authenticated Encryption with Associated Data (AEAD).
+ * https://datatracker.ietf.org/doc/html/rfc5116
+ */
+public class AeadEncryptor {
+    private final AesGcmJce encryptor;
+
+    public AeadEncryptor(byte[] key) throws GeneralSecurityException {
+        // TODO(#2120): Use a wrapper AEAD interface and don't use `tink.subtle`.
```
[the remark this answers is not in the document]
ipetr0v: Removed

remote_attestation/java/src/com/google/oak/remote_attestation/Encryptor.java
```diff
@@ -0,0 +1,21 @@
+package com.google.oak.remote_attestation;
```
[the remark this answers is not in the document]
ipetr0v: Done

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Updated.
Now I have `UnattestedPeer` with the `attest` method, that returns `AttestedPeer` with `encrypt`/`decrypt` methods.
Similar to what is being done in https://github.com/project-oak/oak/pull/2156

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
```diff
@@ -0,0 +1,37 @@
...
+        // Generate client private/public key pair.
+        keyNegotiator = new KeyNegotiator();
+    }
+
+    /** Returns a Diffie-Hellman public key corresponding to the `keyNegotiator`. */
+    public byte[] getPublicKey() throws GeneralSecurityException {
+        return keyNegotiator.getPublicKey();
+    }
+
+    public Encryptor attest(byte[] peerPublicKey, byte[] peerAttestationInfo) throws GeneralSecurityException {
```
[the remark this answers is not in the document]
ipetr0v: Thanks! Done
Also, the `attest` will accept `ServerIdentity` once https://github.com/project-oak/oak/pull/2156 is done.`

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.base.VerifyException;
+import com.google.oak.remote_attestation.crypto.AeadEncryptor;
+import com.google.oak.remote_attestation.crypto.KeyNegotiator;
+import java.security.GeneralSecurityException;
+
+public class Attestor {
+    /**
+     * Performs remote attestation and key negotiation.
+     */
+    static public class UnattestedPeer {
```
rbehjati, line 28:
```suggestion
    public static class UnattestedPeer {
```
This is the more conventional ordering of the modifiers.

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
```diff
@@ -0,0 +1,76 @@
...
+        private Boolean verifyAttestation(byte[] attestationInfo) {
+            // TODO(#1867): Add remote attestation support.
+            return true;
+        }
+    }
+
+    /**
+     * Performs data encryption/decryption based on the negotiated key.
+     */
+    static public class AttestedPeer {
```
rbehjati, line 61:
```suggestion
    public static class AttestedPeer {
```

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private Boolean verifyAttestation(byte[] attestationInfo) {
+            // TODO(#1867): Add remote attestation support.
+            return true;
+        }
+    }
+
+    /**
+     * Performs data encryption/decryption based on the negotiated key.
```
rbehjati, line 59: Is it really conceptually a peer that does the encryption and decryption?

rbehjati (approved):
I am still unsure about the naming in `Attestor.java`. Looks like `UnattestedPeer` represents the client, and the `peer` in `peerPublicKey` and `peerAttestationInfo` passed to `attest` represents the server. I think using server and client instead of peer could better clarify the process. But if most of this PR is going to be replaced by #2156 then the naming could be fixed there.

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
ipetr0v: In the rust version I use `AttestationStateMachine<Attested>`.
I think here it could be called `AttestationStateMachine.Attested` (inner class)

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Attestor.java
ipetr0v: Done

rbehjati (approved):

PR #2165, opened 2021-06-03
Update deploy scripts

### LINKED ISSUE #2141: Fix the Cloud Run deployment scripts to use gRPC attestation
After #2122 is merged, cloud run scripts must be updated to correctly use gRPC attestation. 

See https://github.com/project-oak/oak/pull/2122#discussion_r639861741 and https://github.com/project-oak/oak/pull/2122#discussion_r639860414.

rbehjati:
Fixes #2141

### DIFF
--- a/oak_functions/gcp/endpoint_api_config.yaml.template
+++ b/oak_functions/gcp/endpoint_api_config.yaml.template
@@ -1,0 +1,29 @@
+type: google.api.Service
+config_version: 3
+
+#
+# Name of the service configuration.
+#
+name: ###SERVICE_HOSTNAME###
+
+#
+# API title to appear in the user interface (Google Cloud Console).
+#
+title: ###TITLE###
+apis:
+  - name: oak.functions.server.RemoteAttestation
+
+#
+# API usage restrictions.
+#
+usage:
+  rules:
+    # Invoke method can be called without an API Key for now.
+    - selector: oak.functions.server.RemoteAttestation.AttestedInvoke
+      allow_unregistered_calls: true
+
+backend:
+  rules:
+    - selector: '*'
+      # The address of the Cloud Run backend instance
+      address: grpcs://###BACKEND_HOSTNAME###
--- a/runner/src/examples.rs
+++ b/runner/src/examples.rs
@@ -558,7 +558,7 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
     }
 }
 
-pub fn build_functions_example(opt: &BuildFunctionsExample) -> Step {
+pub fn build_functions_example(opt: &RunFunctionsExamples) -> Step {
     let example: Example = example_toml_files()
         .map(|path| {
             toml::from_str(&read_file(&path)).unwrap_or_else(|err| {
@@ -588,7 +588,7 @@ pub fn build_functions_example(opt: &BuildFunctionsExample) -> Step {
             .collect(),
     };
 
-    let functions_example = FunctionsExample::new(&example, opt.into());
+    let functions_example = FunctionsExample::new(&example, opt.clone());
 
     Step::Multiple {
         name: example.name.to_string(),
--- a/runner/src/examples.rs
+++ b/runner/src/examples.rs
@@ -515,13 +515,14 @@ fn run_example(example: &ClassicExample) -> Step {
 fn run_functions_example(example: &FunctionsExample) -> Step {
     let opt = &example.options;
 
-    // build any backend server
+    // Build steps for running clients
     let run_clients = run_clients(
         &example.example,
         &opt.build_client,
         opt.client_additional_args.clone(),
     );
 
+    // Build any backend server
     #[allow(clippy::collapsible_if)]
     let run_backend_server_clients: Step = if opt.run_server.unwrap_or(true) {
         let run_server_clients = example.construct_example_server_run_step(run_clients);
@@ -559,19 +560,23 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
 }
 
 pub fn build_functions_example(opt: &RunFunctionsExamples) -> Step {
+    let example_name = &opt
+        .example_name
+        .as_ref()
+        .expect("--example-name must be specified")
+        .clone();
+
     let example: Example = example_toml_files()
         .map(|path| {
             toml::from_str(&read_file(&path)).unwrap_or_else(|err| {
                 panic!("could not parse example manifest file {:?}: {}", path, err)
             })
         })
-        .find(|example: &Example| match &opt.example_name {
-            Some(example_name) => &example.name == example_name,
-            None => true,
-        })
-        .expect("could not find the specified example");
+        .find(|example: &Example| &example.name == example_name)
+        .filter(|example| example.has_functions_app())
+        .expect("could not find the specified functions example");
 
-    // build any backend server
+    // Build steps for building clients
     let build_client = Step::Multiple {
         name: "build clients".to_string(),
         steps: example
--- a/runner/src/examples.rs
+++ b/runner/src/examples.rs
@@ -558,6 +558,59 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
     }
 }
 
+pub fn build_functions_example(opt: &BuildFunctionsExample) -> Step {
+    let example: Example = example_toml_files()
+        .map(|path| {
+            toml::from_str(&read_file(&path)).unwrap_or_else(|err| {
+                panic!("could not parse example manifest file {:?}: {}", path, err)
+            })
+        })
+        .find(|example: &Example| match &opt.example_name {
+            Some(example_name) => &example.name == example_name,
+            None => true,
+        })
+        .expect("could not find the specified example");
+
+    // build any backend server
+    let build_client = Step::Multiple {
+        name: "build clients".to_string(),
+        steps: example
+            .clients
+            .iter()
+            .filter(|(name, _)| match opt.build_client.client_variant.as_str() {
+                ALL_CLIENTS => true,
+                client => *name == client,
+            })
+            .map(|(name, client)| Step::Single {
+                name: format!("build{}", name),
+                command: build(&client.target, &opt.build_client),
+            })
+            .collect(),
+    };
+
+    let functions_example = FunctionsExample::new(&example, opt.into());
+
+    Step::Multiple {
+        name: example.name.to_string(),
+        steps: vec![
+            functions_example.construct_application_build_steps(),
+            // Build the server first so that when running it in the next step it will start up
+            // faster.
+            vec![build_functions_server(&opt.build_server)],
+            if opt.build_docker {
+                vec![build_docker(&example)]
+            } else {
+                vec![]
+            },
+            functions_example.construct_backend_build_steps(),
+            vec![build_client],
+        ]
+        .into_iter()
+        .flatten()
+        .collect::<Vec<_>>(),
+    }
+}
+
 pub fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Step {
     match target {
         Target::Cargo {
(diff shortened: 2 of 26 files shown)


scripts/deploy_oak_functions_loader
```diff
@@ -27,6 +26,9 @@ docker push "${FUNCTIONS_DOCKER_IMAGE_NAME}:latest"
 
 readonly FUNCTIONS_EXAMPLE_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-functions-weather-lookup'
 
+# Copy the certificate file to the example directory.
+cp ./examples/certs/local/ca.pem ./oak_functions/examples/weather_lookup
```
tiziano88, line 30: What is this cert for?

rbehjati: The tee_certificate, required by the config file: https://github.com/project-oak/oak/blob/86e1ef35564f1e3f431810121fa3bbb2fa3fdc37/oak_functions/examples/weather_lookup/config.toml#L4

tiziano88: Ah, right, I keep forgetting that, it still does not seem very natural to me. It's just a binary blob that gets sent by the server at some point in the remote attestation protocol. In the same way, we will have to also have an inclusion proof in a verifiable log of the hash, and other similar things. I think it should be renamed to something more generic, and also not pass another unrelated certificate. Cc @ipetr0v

rbehjati: What should be passed instead of this certificate? At the moment it looks like that the attestation protocol ignores it. So, I could remove this and use an empty vector. But if we have to redo this later, I'll keep it like this. 
On the other hand, I think we should make `tee_certificate_path` a command line option, instead of including it in the config. It would make it easier to use different files in different paths for CI builds, and cloud run deployments. WDYT?

tiziano88: I think it's fine to leave it in for now, but I do think we need to clarify the story, the current structure is not particularly realistic. Even if we wanted to provide the AMD cert there, why would the server send it to the client? What would the client verify about it? What stops the server from sending a self-signed cert? Or is the intention that the client has the AMD root cert already, and this is just one or more intermediate certs that are machine-specific? And even if they are machine-specific, surely they would not appear as files in a Docker image.

ipetr0v: IIUC provider's certificate should be signed by a certificate that client can check (i.e. it's rooted in a well known certificate).

tiziano88: But where does it come from? And who puts it there?

ipetr0v: It should be available by the remote attestation implementation alongside the TEE report.

tiziano88: So it's unlikely it will be loaded from a file path, right? By the way, instead, something like a log inclusion proof would actually be passed as a file eventually, so I think this makes more sense for something like that

rbehjati: Just to be clear, this is the `tee_certificate` in `AttestationServer` that we are talking about: https://github.com/project-oak/oak/blob/b8a986aefca80d26604d227e2283ad80a7e96006/experimental/grpc_attestation/src/attestation.rs#L115-L120. Is that right?

tiziano88: Yes that is my understanding.

rbehjati: Removed it for now.

runner/src/examples.rs
```diff
@@ -558,6 +558,59 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
...
+                panic!("could not parse example manifest file {:?}: {}", path, err)
+            })
+        })
+        .find(|example: &Example| match &opt.example_name {
+            Some(example_name) => &example.name == example_name,
+            None => true,
+        })
+        .expect("could not find the specified example");
+
+    // build any backend server
```
conradgrobler, line 574: This comment mentions "backend server", but the code is all about clients. Also, the comment should be a sentence that starta with a capital and enda with a full stop.

runner/src/examples.rs
```diff
@@ -558,6 +558,59 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
     }
 }
 
+pub fn build_functions_example(opt: &RunFunctionsExamples) -> Step {
+    let example: Example = example_toml_files()
```
conradgrobler, line 562: It is not clear to me how this ensures that only functions examples are built here.

scripts/deploy_oak_functions_api
```diff
@@ -53,6 +53,11 @@ sed \
 readonly CONFIG_ID='weather-lookup-config'
 readonly API_ID='weather-lookup-api'
 
+# First delete the existing api-config and gateway
+# gcloud api-gateway gateways delete "${GATEWAY_ID}" --location="${GCP_REGION}"
+# gcloud api-gateway api-configs delete "${CONFIG_ID}" --api="${API_ID}" --project="${GCP_PROJECT_ID}"
+
```
conradgrobler, line 59: I think an extra step is needed to delete the API as well with `gcloud api-gateways apis delete ...` - the API is automatically created when creating the initial config, but not automatically deleted.

Also, you probably need to disable the API before being able to delete is (I think `gcloud services disable "${MANAGED_SERVICE}"` should do it.

scripts/deploy_oak_functions_api
```diff
@@ -53,6 +53,11 @@ sed \
 readonly CONFIG_ID='weather-lookup-config'
 readonly API_ID='weather-lookup-api'
 
+# First delete the existing api-config and gateway
```
conradgrobler, line 56: It is perhaps a good idea to add this code in an if statemet that runs it is the API already exists, then there is no need to keep it commented out.

conradgrobler (approved):

rbehjati:
@conradgrobler @ipetr0v Following up on our debugging session today, I added some logs. The client gets stuck at the very beginning right at: https://github.com/project-oak/oak/blob/7f2a453ae58f79030c36e31e9814cef29f83599d/experimental/grpc_attestation_client/rust/src/lib.rs#L55-L58

So the request never reaches the application (i.e., the call to `attested_invoke`): https://github.com/project-oak/oak/blob/b8a986aefca80d26604d227e2283ad80a7e96006/experimental/grpc_attestation/src/attestation.rs#L197-L200

Any ideas why this happens?

conradgrobler:
> Any ideas why this happens?

My best guess is that the API gateway is incorrectly configured. Does it work when you change the script to connect directly to the Cloud Run instance on https://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app?

rbehjati:
> Is it happening locally or on Cloud Run?

Cloud Run. 

> My best guess is that the API gateway is incorrectly configured. Does it work when you change the script to connect directly to the Cloud Run instance on https://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app?

Very good guess! What is `https://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app`? I get `{"temperature_degrees":-10}` when I use this one :) 

Is the API gateway config different from what we have in the scripts and the `api_config.yaml.template` file?

conradgrobler:
> Is the API gateway config different from what we have in the scripts and the `api_config.yaml.template` file?

It is not. I think I now see the problem. The only gRPC call that bypasses the need for an API key is `oak.functions.server.GrpcHandler.Invoke`. Seeing that the call have changed, the API usage section of the yaml template should be updated.

rbehjati:
I had updated that before: https://github.com/project-oak/oak/blob/7f2a453ae58f79030c36e31e9814cef29f83599d/oak_functions/gcp/api_config.yaml.template#L19-L23

And I believe the API we deployed yesterday was using this version.

conradgrobler:
Ah, I guess I looked at an older version. Does the generated yaml file in oak_functions/bin have the right URL for the Cloud Run instance?

rbehjati:
> Ah, I guess I looked at an older version. Does the generated yaml file in oak_functions/bin have the right URL for the Cloud Run instance?

Yes, on my machine I have `grpcs://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app` in `oak_functions/bin/weather_lookup_api_config.yaml`.

conradgrobler:
The only other current theory I have is that there might be an additional setting required to support streaming, but I have not been able to find anything like that so far.

rbehjati:
What does the API gateway URL have in addition to `https://oak-functions-weather-lookup-62sa4xcfia-nw.a.run.app` apart from TLS?

scripts/deploy_oak_functions_api
rbehjati: Thanks. Fixed.

runner/src/examples.rs
rbehjati: Fixed.

runner/src/examples.rs
rbehjati: Thanks. Fixed the condition in `find`, also added a `filter` to makes sure the specified example is a "functions" example.

rbehjati (commented):
Thanks for the review @conradgrobler

ipetr0v:
If this PR adds TLS to all Oak Functions clients. How will they work in CI, since local Oak Functions don't have TLS servers?

rbehjati:
> If this PR adds TLS to all Oak Functions clients. How will they work in CI, since local Oak Functions don't have TLS servers?

The example uses `http` instead of `https`. I guessing because of this TLS config is ignored.

rbehjati:
I have added a new script for deploying Cloud Endpoints, instead of API Gateways. And it works. Could you please take another look.

scripts/deploy_oak_functions_endpoints
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+
+readonly SCRIPTS_DIR="$(dirname "$0")"
+# shellcheck source=scripts/common
+source "$SCRIPTS_DIR/common"
+
+# shellcheck source=scripts/gcp_common
+source "$SCRIPTS_DIR/gcp_common"
+
+gcloud auth activate-service-account \
```
rbehjati, line 10: I was not able to run this with the Cloud Run service account that we currently have, I had to use my personal account. 
Here is the first error message:
```
+ gcloud builds submit --tag gcr.io/oak-ci/endpoints-runtime-serverless:2.27.0-weather-lookup-endpoint-62sa4xcfia-nw.a.run.app-2021-06-11r7 . --project=oak-ci
ERROR: (gcloud.builds.submit) HTTPError 403: cloud-run@oak-ci.iam.gserviceaccount.com does not have storage.buckets.list access to the Google Cloud project.
```
We need a service account with additional permissions.

ipetr0v (approved):

oak_functions/gcp/endpoint_api_config.yaml.template
```diff
@@ -0,0 +1,29 @@
+type: google.api.Service
+config_version: 3
+
+#
+# Name of the service configuration.
+#
+name: ###SERVICE_HOSTNAME###
```
tiziano88, line 7: For the record, I still find this a bit over-engineered and less readable than having, say, two separate plain YAML configs. I wonder if we could go back to that at some point @conradgrobler ? Is the main issue the unpredictability of the Cloud Run URL?

tiziano88 (approved):

oak_functions/gcp/endpoint_api_config.yaml.template
conradgrobler: Yes, the main issue is the changing hash in the Cloud Run URL if the instance is recreated. The other issue is that the file will be exactly the same apart from these two values for all deployments. It should be simple to go back to hardcoded config files. We must just then remember to update the files if we recreate the Cloud Run instance.

rbehjati: I think hard-coding `SERVICE_HOSTNAME` and `BACKEND_HOSTNAME`, and having to update them manually could be error-prone.

scripts/deploy_oak_functions_endpoints
rbehjati: This is fixed now.

PR #2166, opened 2021-06-03
Prove most of PNI with partially-secret labels

bMacSwigg:
Completes all of Possibilistic non-interference for partially-secret
labels with the exception of one critical theorem, which relies on a
fact that is no longer true with partially secret labels. The erroneous
fact is commented out, and the critical theorem is left Admitted.

PNI should still be provable, but more work will be required to find a
workaround for that fact and complete the proof.

Also cleans up some unused lemmas.

### DIFF
--- a/experimental/ni_coq/vfiles/PossibilisticNI_PS.v
+++ b/experimental/ni_coq/vfiles/PossibilisticNI_PS.v
@@ -4,16 +4,16 @@ From OakIFC Require Import
     Lattice
     Parameters
     GenericMap
-    RuntimeModel
+    RuntimeModelPS
     ModelSemUtils
-    EvAugSemantics
+    EvAugSemanticsPS
     State
     Events
-    LowEquivalences
+    LowEqPS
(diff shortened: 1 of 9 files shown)


experimental/ni_coq/vfiles/PossibilisticNI_PS.v
```diff
@@ -0,0 +1,410 @@
...
+
+Theorem low_proj_steps_implies_leq_step: forall ell s s1' e1,
+    (step_system_ev (state_low_proj ell s) s1' e1) ->
+    (exists s2' e2,
+        (step_system_ev s s2' e2) /\
+        (state_low_eq ell s1' s2') /\
+        (event_low_eq ell e1 e2)).
+Proof.
+    (*
+     * Relies on state_proj_perserves_chan_lbl (see non-PS version).
```
aferr, line 237: Thanks for the note. I appreciate this!

aferr (approved):
LGTM

PR #2168, opened 2021-06-04
Run runner cargo commands only for modified crates

rbehjati:
Allows running the following commands only for the recently modified files and crates affected by them: 
- run-cargo-tests
- run-examples
- run-functions-examples
- format
- check-format

To use this feature, the `--commits=N` should be passed to the command, where `N>0` is the number of past commits to be considered for finding the modified files (`git diff HEAD~N` is used for finding the modified files). For this to work for new files, the file should be added to the git index.

For now this is intended to be used when running commands locally.


tiziano88:
Nice addition! But could you make sure it is opt-in, since I think this logic may not be accurate in some cases (for instance for my own workflow I think it may not necessarily help much). Also we would not want it in CI either I think. Perhaps a flag like `--git-diff` or similar? Another approach could be to actually hash all the source files at each invocation and store them in a file, and then compute the diff against that. That may actually be simpler and more effective than the git diff (and the flag could be called `--incremental` , similar to compilers).

(For the record, I personally think all of these are overkill, we should try and make each individual step as idempotent and fast as possible, instead of second guessing what it is going to do based on diffs.)

rbehjati:
> Nice addition! But could you make sure it is opt-in, since I think this logic may not be accurate in some cases (for instance for my own workflow I think it may not necessarily help much). Also we would not want it in CI either I think. Perhaps a flag like `--git-diff` or similar? 

Yes. I'll add a flag, exactly because indicating all the steps that need to be executed after each change is not very straightforward. 

> Another approach could be to actually hash all the source files at each invocation and store them in a file, and then compute the diff against that. That may actually be simpler and more effective than the git diff (and the flag could be called `--incremental` , similar to compilers).

Not sure I understand exactly how it would work, but yes, `git diff` does not seem to work in all situations. 

> (For the record, I personally think all of these are overkill, we should try and make each individual step as idempotent and fast as possible, instead of second guessing what it is going to do based on diffs.)

Currently, even with a single change in Oak Functions, we run all the examples and all the tests in all other crates and workspaces, while they are all a completely irrelevant to change. I am not sure if making the steps faster will help that much, as there is so many of them.

runner/src/diffs.rs
```diff
@@ -0,0 +1,298 @@
...
+// to `affected_crates_toml_path`.
+fn add_affected_crates(crate_toml_path: PathBuf, affected_crates_toml_path: &mut HashSet<String>) {
+    if affected_crates_toml_path.contains(&to_string(crate_toml_path.clone())) {
+        return;
+    }
+    let deps = get_local_dependencies(&crate_toml_path);
+    for dep in deps {
+        if !affected_crates_toml_path.contains(&dep) {
+            let dep_path = PathBuf::from(dep.clone());
+            add_affected_crates(dep_path.clone(), affected_crates_toml_path)
```
conradgrobler, line 255: Seeing that this checks all dependencies, including dev and build dependencies, it seems that it is possible that it could pick up circular dependencies if it does not keep track of all visited crates (not just affected crates).

runner/src/internal.rs
```diff
@@ -327,6 +345,14 @@ pub struct RunCargoFuzz {
 pub struct CargoManifest {
     #[serde(default)]
     pub bin: Vec<CargoBinary>,
+    #[serde(default)]
```
conradgrobler, line 348: Is there not an existing, generally used crate to support this?

conradgrobler (approved):

tiziano88 (approved):
Personally I am still unconvinced about this approach, it is a lot of logic to essentially try and second guess what the compiler already does.

My suggestion would still be to actually make sure that each step (at least those involving cargo) is as fast as possible, and in particular making sure that we do not rebuild or rerun things unnecessarily (which I think is still happening currently in some cases).

Also in addition, it would be nice for runner to emit a summary of the steps that failed, and at the same time being able to rerun only those steps.

for instance, I could run `runner run-ci`, and 4 steps would fail, then runner would emit something like:

```
to rerun the failing tests only, use the following command:
runner run-tests --filter=foo,bar run-examples --example-name=weather_lookup
```

(obviously this does not work right now, because each step is identified by a subcommand plus parameters, but it may be simple enough to instead use some sort of fully qualified name for each step, e.g. `tests.foo` `tests.bar` `functions-example.weather_lookup`, and we could pass multiple ones if needed; BTW this is not too different to how things work in bazel)

anyways if you prefer this approach and conrad is happy with it too, feel free to go ahead

rbehjati (commented):
Thanks for the reviews.

rbehjati:
> Personally I am still unconvinced about this approach, it is a lot of logic to essentially try and second guess what the compiler already does.
> 
> My suggestion would still be to actually make sure that each step (at least those involving cargo) is as fast as possible, and in particular making sure that we do not rebuild or rerun things unnecessarily (which I think is still happening currently in some cases).
> 
> Also in addition, it would be nice for runner to emit a summary of the steps that failed, and at the same time being able to rerun only those steps.
> 
> for instance, I could run `runner run-ci`, and 4 steps would fail, then runner would emit something like:
> 
> ```
> to rerun the failing tests only, use the following command:
> runner run-tests --filter=foo,bar run-examples --example-name=weather_lookup
> ```
> 
> (obviously this does not work right now, because each step is identified by a subcommand plus parameters, but it may be simple enough to instead use some sort of fully qualified name for each step, e.g. `tests.foo` `tests.bar` `functions-example.weather_lookup`, and we could pass multiple ones if needed; BTW this is not too different to how things work in bazel)
> 
> anyways if you prefer this approach and conrad is happy with it too, feel free to go ahead

The feature you are describing here is useful, but the use case I have in mind is slightly different. My problem is that just running the very first `run-ci` command takes a long time. Actually, I have not been using it for a very long time. I mostly use `run-cargo-tests` and it takes 10-25 minutes to run. Similarly `format` takes at least 4 minutes. This causes a lot of context switching for me. In my experiments, running these commands only for the modified crates easily reduces the time to less than 30 seconds. 
I should also add that ideally the the IDE should have seamless support for such simple tasks, but Vscode is not sufficiently reliable in highlighting compile errors or formatting.

PR #2174, opened 2021-06-09
Use repo's root as the `proto_path` in calls to `compile_proto` 

### LINKED ISSUE #2167: Use the root repo as the proto_path in all invocations of prost
Ref discussion on Slack: https://project-oak.slack.com/archives/CHE9E13C3/p1622730588061800

rbehjati:
Fixes #2167

The recommended approach for using `protoc` is to use the root of the source tree as the proto path, instead of including multiple paths. All paths must be specified relative to the root of the source tree. This change list updates all invocations of `compile_protos` and `generate_grpc_code` according to this advice. 

I have updated the documentation of `compile_protos` and `generate_grpc_code`, but I don't think there is a way to actually enforce this.

### DIFF
--- a/docs/programming-oak.md
+++ b/docs/programming-oak.md
@@ -245,10 +245,7 @@ to autogenerate Rust code from a
 fn main() {
     // Generate the Oak-specific server and client code for the gRPC service,
     // along with the Rust types corresponding to the message definitions.
-    oak_utils::compile_protos(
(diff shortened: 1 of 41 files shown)


conradgrobler (approved):
Nice cleanup.

rbehjati:
I'll merge this after #2156 is merged.

ipetr0v (approved):

PR #2175, opened 2021-06-09
Add cache to ignored paths in runner

ipetr0v:


rbehjati (approved):
LGTM, but I don't have a `cache` or `.cache` on my local repo. Where does it come from?

ipetr0v:
I comes from `bazel build`

Issue #2087, opened 2021-05-14
Implement unique nonces for gRPC Attestation Service


ipetr0v:
Currently https://github.com/project-oak/oak/pull/2070 implements AES-GCM encryption using a single constant nonce.
We need to implement a counter (shared between client and server) such that each message will have a unique nonce.

ipetr0v:
Comment related to unique counters in messages that are necessary to protect clients against replay attacks: https://github.com/project-oak/oak/pull/2070#discussion_r634462255

PR #2182, opened 2021-06-11
Use random nonce in Remote Attestation

ipetr0v:
Fixes https://github.com/project-oak/oak/issues/2087

### DIFF
--- a/remote_attestation/rust/src/crypto.rs
+++ b/remote_attestation/rust/src/crypto.rs
@@ -16,10 +16,10 @@
 
 use crate::proto::EncryptedData;
 use anyhow::anyhow;
-use rand::Rng;
 use ring::{
     aead::{self, BoundKey},
     agreement,
+    rand::{SecureRandom, SystemRandom},
 };
 use sha2::{digest::Digest, Sha256};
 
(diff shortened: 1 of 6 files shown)


remote_attestation/rust/src/crypto.rs
```diff
@@ -114,6 +113,19 @@ impl AeadEncryptor {
             .map_err(|error| anyhow!("Couldn't decrypt data: {:?}", error))?;
         Ok(decrypted_data.to_vec())
     }
+
+    /// Generate a random nonce.
+    /// `ring::aead` uses 96-bit (12-byte) nonces.
+    /// https://briansmith.org/rustdoc/ring/aead/constant.NONCE_LEN.html
+    fn generate_nonce() -> [u8; aead::NONCE_LEN] {
+        let mut rng = rand::thread_rng();
```
tiziano88, line 121: Is it intentional to use `rand::thread_rng()` here and `ring::rand::SystemRandom` below?

tiziano88 (approved):

remote_attestation/rust/src/crypto.rs
ipetr0v: Changed everything to `ring::rand::SystemRandom`

Issue #2181, opened 2021-06-11
Use separate keys for server and client encryption in Remote Attestation


ipetr0v:
After finishing https://github.com/project-oak/oak/issues/2100 we may need to implement separate keys for server and client encryption.
It means that both client and server will derive a pair of session keys:
- One for encrypting/decrypting client requests
- One for encrypting/decrypting server responses

This scheme will prevent the **Loopback Attack**: e.g. malicious network takes an outgoing packet from server and feeds it back to it as an incoming packet. Though it might not be strictly necessary if we will use different sequence numbers for all packets.

ipetr0v:
cc @tiziano88

PR #2185, opened 2021-06-12
Use separate session keys in Remote Attestation

ipetr0v:
This PR adds separate keys for encrypting data and decrypting peer encrypted data.
It is necessary to prevent the Loopback Attack, where malicious network takes an outgoing packet and feeds it back as an incoming packet.

Fixes https://github.com/project-oak/oak/issues/2181


remote_attestation/rust/src/crypto.rs
```diff
@@ -55,17 +57,25 @@ impl aead::NonceSequence for OneNonceSequence {
...
+/// It is necessary to prevent the Loopback Attack, where malicious network takes an outgoing packet
+/// and feeds it back as an incoming packet.
 pub struct AeadEncryptor {
-    /// Key used for encrypting and decrypting data.
-    ///
-    /// AES-GCM algorithm uses the same key for both encryption and decryption.
-    /// https://datatracker.ietf.org/doc/html/rfc5288
-    key: Vec<u8>,
+    /// Key used for encrypting data.
+    encryption_key: Vec<u8>,
```
tiziano88, line 68: Is it possible to avoid storing these keys here, and just keep the sealing key? In fact, the current code looks like it is re-creating the sealing key every time encrypt is called, which does not seem ideal.

tiziano88 (approved):

remote_attestation/java/src/com/google/oak/remote_attestation/KeyNegotiator.java
```diff
@@ -32,13 +32,25 @@
...
+    private static final String CLIENT_KEY_PURPOSE = "Remote Attestation Protocol Client Session Key";
     // AES-256 with GCM key size.
     // https://datatracker.ietf.org/doc/html/rfc5288
     private static final int KEY_SIZE_BYTES = 32;
     private final byte[] privateKey;
 
+    // Defines the type of encryptor created by `KeyNegotiator.createEncryptor`.
+    public enum EncryptorType {
+        // Defines a server encryptor, which uses server session key for encryption and client
+        // session key for encryption.
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/KeyNegotiator.java
```diff
@@ -48,38 +60,75 @@ public KeyNegotiator() {
...
      * https://datatracker.ietf.org/doc/html/rfc7748#section-6.1
-     * 
-     * TODO(#2181): Use separate keys for server and client encryption.
      */
-    public byte[] deriveSessionKey(byte[] peerPublicKey) throws GeneralSecurityException {
+    byte[] keyDerivationFunction(
+        byte[] keyMaterial,
+        String keyPurpose,
+        byte[] firstPublicKey,
+        byte[] secondPublicKey
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/rust/src/crypto.rs
ipetr0v: First we create an `let unbound_sealing_key = aead::UnboundKey::new`.
But then this key is consumed by `let mut sealing_key = ring::aead::SealingKey::new(unbound_sealing_key,`

The problem is that the sealing key needs a Nonce (which is taken from the message every time, for decryption).
And for encryption, if we provide nonce to the key, I'm not sure we can extract each individual nonce after, in order to put it into Protobuf.

tiziano88: Why not use a simple nonce sequence then, since that seems the expected underlying API? I thought a counter is sufficient. If not, then I guess we would need a nonce sequence to which we can feed the next nonce, extracted from the incoming message, just before calling encrypt or decrypt (though it may look a bit weird because of the extra synchronization). What does TLS use for nonce?

remote_attestation/java/src/com/google/oak/remote_attestation/KeyNegotiator.java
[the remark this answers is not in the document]
tiziano88: But why not rename the parameters to `serverPublicKey` and `clientPublicKey`?

tiziano88 (approved):

remote_attestation/rust/src/crypto.rs
ipetr0v: Currently we are using random nonces, because this is done in Tink (with no way of changing it).

TLS (at least 1.3) is much more complicated in this sense:
```
It uses instead a combination of a fixed per-connection IV (derived at the same time as the key) and the sequence number, XORed - not concatenated.
```

tiziano88: Note that the TLS approach is actually simpler to implement than the one you have here, since it can be encapsulated in a struct that implements the https://briansmith.org/rustdoc/ring/aead/trait.NonceSequence.html trait quite nicely (it just needs to have the IV, and then keep track of the counter internally). Anyways, let's leave that for later.

PR #2191, opened 2021-06-14
Use non-empty lookup data for wasm_invoke fuzz target

rbehjati:
Ref https://github.com/project-oak/oak/pull/2114#issuecomment-846882117

### DIFF
--- a/oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
+++ b/oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
@@ -24,9 +24,9 @@ use arbitrary::Arbitrary;
 use libfuzzer_sys::fuzz_target;
 use maplit::hashmap;
 use oak_functions_abi::proto::Request;
-use oak_functions_loader::{logger::Logger, lookup::LookupDataForTest, server::WasmHandler};
+use oak_functions_loader::{logger::Logger, lookup::LookupData, server::WasmHandler};
 use prost::Message;
-use std::sync::Arc;
+use std::{convert::Into, sync::Arc};
 
 #[derive(Arbitrary, Debug, Clone, PartialEq)]
 enum ArbitraryInstruction {
@@ -45,10 +45,24 @@ enum ArbitraryInstruction {
     },
 }
 
+/// Enum to allow simulating both hit and miss lookup scenarios.
 #[derive(Arbitrary, Debug, Clone, PartialEq)]
 enum LookupKey {
-    Key,
+    /// Uses `FIXED_KEY` for lookup. Covers the scenario where the lookup matches a key in
+    /// LookupData.
+    FixedKey,
+    /// Uses a random key for lookup. Covers the scenario where the key does not match anything in
+    /// LookupData.
     RandomKey { key: Vec<u8> },
+}
+
+impl From<&LookupKey> for Vec<u8> {
+    fn from(key: &LookupKey) -> Self {
+        match key {
+            LookupKey::FixedKey => FIXED_KEY.to_vec(),
+            LookupKey::RandomKey { key } => key.clone(),
+        }
+    }
 }
 
 const FIXED_KEY: &[u8] = b"key";
@@ -96,7 +110,7 @@ fuzz_target!(|instruction_list: Vec<ArbitraryInstruction>| {
 
     let wasm_handler = WasmHandler::create(
         &WASM_MODULE_BYTES,
-        Arc::new(LookupDataForTest::new(entries).lookup_data),
+        Arc::new(LookupData::for_test(entries)),
         Logger::for_test(),
     )
     .expect("Could instantiate WasmHandler");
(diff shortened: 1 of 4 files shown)


oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -37,13 +38,21 @@ enum ArbitraryInstruction {
...
+        key: LookupKey,
     },
     WriteLogMessage {
         message: Vec<u8>,
     },
 }
 
+#[derive(Arbitrary, Debug, Clone, PartialEq)]
+enum LookupKey {
+    Key,
```
tiziano88, line 50: I find it a bit hard to follow this code, basically this enum variant means that the FIXED_KEY value is used, right? Is it just to make sure that there is a balance between the random keys and some key that we think is more useful to find interesting bugs? Or is there some other reason? Maybe also rename `Key` to `DefaultKey` or `FixedKey` and add a comment.

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -106,11 +119,16 @@ impl From<&ArbitraryInstruction> for crate::proto::Instruction {
                     response: response.clone(),
                 }),
             ),
-            ArbitraryInstruction::StorageGetItem { key } => {
-                Some(InstructionVariant::StorageGetItem(
+            ArbitraryInstruction::StorageGetItem { key } => match key {
+                LookupKey::Key => Some(InstructionVariant::StorageGetItem(
+                    crate::proto::StorageGetItem {
+                        key: FIXED_KEY.to_vec(),
```
tiziano88, line 125: It looks like the only thing that needs to change between the two match branches is the actual key, but there is no need to repeat all the rest of the expression, right?

oak_functions/loader/src/lookup.rs
```diff
@@ -153,3 +153,24 @@ pub fn parse_lookup_entries<B: prost::bytes::Buf>(
...
+
+/// Wrapper around a LookupData instance for testing.
+#[allow(dead_code)]
+pub struct LookupDataForTest {
+    pub lookup_data: LookupData,
+}
+
+impl LookupDataForTest {
+    /// Creates an instance of LookupDataForTest, containing an instance of LookupData populated
+    /// with the given entires.
```
tiziano88, line 165:
```suggestion
    /// with the given entries.
```

oak_functions/loader/src/lookup.rs
```diff
@@ -153,3 +153,24 @@ pub fn parse_lookup_entries<B: prost::bytes::Buf>(
     }
     Ok(entries)
 }
+
+/// Wrapper around a LookupData instance for testing.
+#[allow(dead_code)]
+pub struct LookupDataForTest {
+    pub lookup_data: LookupData,
+}
```
tiziano88, line 161: AFAICT this wrapper is never actually used, only the inner LookupData instance? If so, why not have a function to create a LookupData instance for test directly?

tiziano88 (approved):

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
rbehjati: I am not so much concerned about the balance, but just to make sure that at least once in a while the value `key` is used for lookup, so that we get coverage for the path where lookup hits something in the data. Is there a better way to do this? 

Added clarifying comments and renamed to `FixedKey`.

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
rbehjati: Fixed.

oak_functions/loader/src/lookup.rs
rbehjati: True. I was not sure if it is a good idea to add a method for testing to `LookupData`. Fixed.

rbehjati (commented):
Thanks :)

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -107,8 +125,12 @@ impl From<&ArbitraryInstruction> for crate::proto::Instruction {
                 }),
             ),
             ArbitraryInstruction::StorageGetItem { key } => {
+                let key = match key {
+                    LookupKey::FixedKey => FIXED_KEY.to_vec(),
+                    LookupKey::RandomKey { key } => key.clone(),
+                };
```
tiziano88, line 131: BTW this may also be a method `LookupKey::value()` directly, if you prefer (it seems nicely self contained). This is fine too though.

tiziano88 (approved):

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
rbehjati: This is nicer. Added a `From` implementation.

Issue #2172, opened 2021-06-09
Connect to API Gateway from Oak Functions Android client


ipetr0v:
Currently in https://github.com/project-oak/oak/pull/2145 Android client for Oak Functions connects to `localhost`. We need to make it connect to API Gateway in Cloud instead.

Original comment: https://github.com/project-oak/oak/pull/2145/files#r647599007

ipetr0v:
https://github.com/project-oak/oak/pull/2190 adds the ability to use TLS in Java gRPC. It works in Java, but Android shows a Runtime error:
```
java.lang.IllegalStateException: Could not find TLS ALPN provider; no working netty-tcnative, Conscrypt, or Jetty NPN/ALPN available
```

Corresponding Slack thread: https://project-oak.slack.com/archives/CHE9E13C3/p1623678258107400

ipetr0v:
There is a possibility to use [Conscrypt](https://github.com/grpc/grpc-java/blob/master/SECURITY.md#tls-on-android). And this requires to do the following:
1) Add The Runtime TLS provider:
```java
import java.security.Security;
import org.conscrypt.Conscrypt;

// Create TLS provider.
Security.insertProviderAt(Conscrypt.newProvider(), 1);
```
2) Add the following code to the `WORKSPACE`:
```skylark
load("@bazel_tools//tools/build_defs/repo:java.bzl", "java_import_external")

http_archive(
    name = "rules_jvm_external",
    sha256 = "887ce49cb2bd2faa82ab8d09283095901a5e2b48e9e9e375715db70b6a789d76",
    strip_prefix = "rules_jvm_external-5fe826e1757815699a8fdb302512dbb385d7752b",
    urls = [
        # Head commit on 2021-05-21.
        "https://github.com/bazelbuild/rules_jvm_external/archive/5fe826e1757815699a8fdb302512dbb385d7752b.zip",
    ],
)

# TLS provider for Android.
java_import_external(
    name = "conscrypt",
    licenses = ["notice"],
    jar_sha256 = "eaf537d98e033d0f0451cd1b8cc74e02d7b55ec882da63c88060d806ba89c348",
    jar_urls = [
        "https://repo1.maven.org/maven2/org/conscrypt/conscrypt-openjdk-uber/2.5.2/conscrypt-openjdk-uber-2.5.2.jar",
    ],
)

java_import_external(
    name = "conscrypt-android",
    licenses = ["notice"],
[...]

PR #2195, opened 2021-06-15
Use Conscrypt as TLS provider in Android

ipetr0v:
Ref https://github.com/project-oak/oak/issues/2172

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -103,6 +103,16 @@ java_import_external(
     ]
 )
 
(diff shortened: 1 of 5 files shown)


conradgrobler:
There is no longer a need to manually add a TLS provider. The current version of the Oak Functions Android client (using OkHttp as the gRPC transport provider) works fine with TLS as long as it is used on a device/emulator with Google Play Services installed.

tiziano88:
@ipetr0v can this be closed?

PR #2203, opened 2021-06-17
Align Oak Functions Android Client with Java and Android best practices

conradgrobler:
- Follow style guide for all variable casing
- Be explicit about character set when converting strings to bytes
- Update formatting to match style guide
- Descriptions for XML resource strings
- Explicit values for attributes in Android Manifest
- Use OkHttp as a transport provider for gRPC on Android
- Use lite versions for protobuf and gRPC on Android

# Checklist

      construction.


oak_functions/client/android/com/google/oak/functions/android/client/MainActivity.java
```diff
@@ -30,24 +29,12 @@
...
-import java.util.ArrayList;
-import java.util.Optional;
-import java.util.stream.Collectors;
 import oak.functions.invocation.Request;
 import oak.functions.invocation.Response;
 import oak.functions.invocation.StatusCode;
 
 /** Main class for the Oak Functions Client application. */
 public class MainActivity extends Activity {
-  private HandlerThread backgroundHandlerThread;
```
rbehjati, line 47: So these fields were here but never used?!

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -46,106 +46,101 @@
...
-                        .setClientIdentity(identity)
-                        .build();
-        requestObserver.onNext(request);
-        AttestedInvokeResponse response = messageQueue.take();
-
-        // Remotely attest peer.
-        AttestationIdentity peerIdentity = response.getServerIdentity();
-        encryptor = attestationEngine.createClientEncryptor(peerIdentity);
-    }
+  private static final Logger logger = Logger.getLogger(AttestationClient.class.getName());
```
rbehjati, line 49: Is this really the correct formatting? No indentation? If so, we might be able to enable clang-formatting for java files. I experimented with it before, and I found the indentation weird, so I did not update the runner.

remote_attestation/java/src/com/google/oak/remote_attestation/AttestationEngine.java
```diff
@@ -25,74 +25,66 @@
 import oak.remote_attestation.AttestationIdentity;
 import oak.remote_attestation.AttestationInfo;
 import oak.remote_attestation.AttestationReport;
```
rbehjati, line 27: Pre-existing: does it makes sense to move the protos to `com.google.oak` package too? or perhaps to `com.google.oak.protos`?

rbehjati (approved):
LGTM. Thanks!

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
conradgrobler: According to https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation it should be 2 spaces at each level, rather than the 4 that is used for e.g. Rust.

oak_functions/client/android/com/google/oak/functions/android/client/MainActivity.java
conradgrobler: Yes, I think it was a copy-paste from the Hello World Android client, which does the service call in the background. This one does not yet do it. I have created #2204 to track moving this logic to a background thread.

remote_attestation/java/src/com/google/oak/remote_attestation/AttestationEngine.java
conradgrobler: Thanks. Created #2205 to track this.

Issue #2216, opened 2021-06-24
Instrumentation for Oak Functions 


rbehjati:
Add instrumentation support for Oak Functions. One use case is to understand the context switch overhead for each ABI call. 
This is also useful for debugging failures from the fuzzing. 

The instrumentation should only be available in the unsafe mode.

rbehjati:
@conradgrobler found the following that we could potentially use for instrumentation or to implement our own more easily:  
- https://github.com/magnet/metered-rs 
- https://docs.rs/aspect/0.3.0/aspect/#modules

rbehjati:
Closing this, since Oak functions is no longer the focus.

PR #2217, opened 2021-06-24
Fix the timeout problem in wasm_invoke fuzz target

rbehjati:
This change should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35506

The timeout happens when there are too many `read_request` calls. In case of the `fuzzable` example that is used for fuzzing, too many calls also means that the request itself is likely to be large. It seems that it is the combination of these two that causes the timeout. This change list fixes the problem by limiting the number instructions executed by the example, but we may require a more generic solution, either in the `WasmHandler`, or as part of the policy. Instrumentation of the ABI functions (#2216) should help better understanding the problem.


oak_functions/examples/fuzzable/module/src/lib.rs
```diff
@@ -25,38 +25,48 @@ use crate::proto::{
...
+        oak_functions::write_response(br"Request is too large.")
+            .expect("Couldn't write response body.");
+    } else {
+        let request = Instructions::decode(&*request).expect("Couldn't decode request.");
+
+        // Run all the instructions given in the request
+        for instruction in request.instructions {
+            match instruction.instruction_variant {
+                Some(InstructionVariant::Panic(Panic {})) => panic!("panic"),
+                Some(InstructionVariant::ReadRequest(ReadRequest {})) => {
```
conradgrobler, line 49: Based on reading the PR decription, I would have expected a counter and a check here to also limit the number of read requests overall. Based on eading the code, it seems that the max number of read requests is limited only by the limit on the size of the input (unless I am missing something). Perhaps clarify this in the PR description?

conradgrobler (approved):

oak_functions/examples/fuzzable/module/src/lib.rs
rbehjati: Good catch. Fixed the PR description and the example to limit the total number of instructions. I could limit only the number of `read_request` calls, but I think (without actually having done any profiling) that limiting the total number of instructions is a more reliable solution.

PR #2220, opened 2021-07-02
Add release scripts

tiziano88:
Note that currently the hashes do not actually match those in the checked in index, but there seems to be some other non-determinism involved. Anyways, that should probably be sorted out separately in a future PR.

### DIFF
--- a/scripts/release_snapshot
+++ b/scripts/release_snapshot
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# This script prepares the repository for a release process. The goal is to ensure that all the
+# dependencies are vendored inside the working tree, so that a hermetic offline build can be kicked
+# off. In principle this build should produce the same exact artifacts as the usual online builds
+# (e.g. from GitHub actions, or locally), but during the release process we want extra guarantees,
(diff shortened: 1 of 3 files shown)


scripts/release_snapshot
```diff
@@ -0,0 +1,48 @@
...
+    --manifest-path=./oak_functions/loader/Cargo.toml \
+    --sync=./runner/Cargo.toml \
+    ./third_party/vendor
+
+  # Also prevent Cargo from accessing the network for any subsequent operation, so that it is
+  # guaranteed to only rely on the local source files and vendored crates.
+  echo ''
+  echo '[net]'
+  echo 'offline = true'
+} >> .cargo/config
```
conradgrobler, line 48: `.cargo/config` here is inconsistent with the comment in line 35 (`.cargo/config.toml`). Also, what happens if this is run multiple times? Will appending this to the file multiple times break anything, or is it up to whoever is running it to roll back changes before running it again?

conradgrobler (approved):

scripts/release_snapshot
tiziano88: Renamed to config.toml. Also clarified that the script is not idempotent. The future playbook for releases will have more details about it.

PR #2237, opened 2021-07-14
Add release workflow to CI

tiziano88:
This follows the same steps that we would need to execute for an actual
release, and provides us a way to double check that the release process
is reproducible, even though we will rely on GCB for the actual
release.

# Checklist

      construction.


conradgrobler (approved):
Nice

PR #2239, opened 2021-07-14
Change container identifier to an ARG

conradgrobler:
This change it to support buildling the weather lookup container from a specific version of the base Oak Functions container image using build-args.

# Checklist

      construction.

### DIFF
--- a/oak_functions/examples/weather_lookup/Dockerfile
+++ b/oak_functions/examples/weather_lookup/Dockerfile
@@ -1,6 +1,8 @@
-# Allow deploying the latest version for development.
-# hadolint ignore=DL3007
-FROM gcr.io/oak-ci/oak-functions:latest
+# Use an argument to specify container identifier.
+# Default to using the latest version for development.
+ARG container_identifier=:latest
+# hadolint ignore=DL3006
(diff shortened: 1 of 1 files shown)


oak_functions/examples/weather_lookup/Dockerfile
```diff
@@ -1,6 +1,8 @@
-# Allow deploying the latest version for development.
-# hadolint ignore=DL3007
-FROM gcr.io/oak-ci/oak-functions:latest
+# Use an argument to specify container identifier.
+# Default to using the latest version for development.
+ARG container_identifier=:latest
```
tiziano88, line 3: How about

```suggestion
ARG base_image_tag=latest
```

(consistent with https://docs.docker.com/engine/reference/commandline/tag/)

or just make the entire image name an arg.

tiziano88 (approved):

oak_functions/examples/weather_lookup/Dockerfile
conradgrobler: Using just a tag will not allow us to specify a specific hash. I will change to using the arg to cover the entire name.

Issue #2236, opened 2021-07-13
Support API for native inference via TensorFlow models


tiziano88:
https://github.com/sonos/tract seems a good library for this

Thanks to @dominic-mulligan-arm for providing https://github.com/veracruz-project/veracruz/pull/159 as reference

rbehjati:
We can do this by adding an ABI function (with SDK support in Rust) that:
- Assumes a TensorFlow model is loaded by the Oak Functions Loader. A path/URI to the model should be specified in a `config.toml` file similar to the config for the lookup data.
- Has two parameters: a byte array, and a vector specifying the shape of the Tensor. The shape is used to reshape the byte array into a Tensor that can be used for inference by the TensorFlow model. With this design, the WebAssembly developer must ensure that the Tensor shape matches the loaded model's expectation. Alternatively, the Tensor shape could also be specified in `config.toml`. But even in this case, the WebAssembly module must ensure the byte array can be correctly reshaped.

We can later extend this ABI by allowing multiple models, where each model has a unique identified specified in the `config.toml` file. The ABI function should then be extended to accept another parameter specifying the model.

conradgrobler:
This has been superceded by the TFLM implementation.

PR #2241, opened 2021-07-26
Native inference via TensorFlow models

rbehjati:
Ref #2236


rbehjati:
Even though the TensorFlow model is pre-loaded by the Oak Functions loader, it takes about 4 seconds to run the model. But I have not benchmarked the example yet.

oak_functions/abi/src/lib.rs
```diff
@@ -72,4 +72,15 @@ extern "C" {
         value_ptr_ptr: *mut *mut u8,
         value_len_ptr: *mut usize,
     ) -> u32;
+
+    /// See [`tf_model_get_shape`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#tf_model_get_shape).
+    pub fn tf_model_get_shape(shape_ptr_ptr: *mut *mut u8, shape_len_ptr: *mut usize) -> u32;
```
conradgrobler, line 77: I don't think this is needed. The shape of the input tensor should be known beforehand and fixed for a model. If the shape of the input tensor changes it is likely that the data sent by the client and the transformation logic in the Wasm code would also need to change anyway.

oak_functions/examples/mobilenet/config.toml
```diff
@@ -0,0 +1,11 @@
+lookup_data_url = ""
+lookup_data_download_period = "10m"
+worker_threads = 4
+
+[tf_model]
+url = "https://storage.googleapis.com/oak-tensorflow-models/mobilenet_v2_1.4_224_frozen"
+shape = [1, 224, 224, 3]
+
+[policy]
+constant_response_size_bytes = 4096
```
conradgrobler, line 10: Do we normally use 4K as a response size? It seems quite large for an example that only sends back the best prediction.

oak_functions/examples/mobilenet/module/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    let request_bytes = oak_functions::read_request().expect("could not read request body");
+    let image =
+        proto::MobilenetImage::decode(&*request_bytes).expect("could not decode MobilenetImage");
+
+    // Resize the image
+    let shape =
+        oak_functions::tf_model_get_shape().expect("could not get shape of the TensorFlow model");
+    let recreated_image = image::RgbImage::from_raw(image.width, image.height, image.image)
+        .ok_or_else(|| anyhow::anyhow!("could not recreate image"))?;
+    let resized = image::imageops::resize(
```
conradgrobler, line 63: I suspect the image processing inside interpreted Wasm is what is making this slow, not the ML inference. Have you tested the performance if this is done in the client and the bytes of the tensor is passed to the Wasm?

oak_functions/loader/src/tf.rs
```diff
@@ -0,0 +1,104 @@
...
+            .into_runnable()?;
+
+        Ok(TensorFlowModel { model, shape })
+    }
+
+    /// Run the model for the given input and return the resulting inference vector.
+    pub fn get_inference(&self, tensor_bytes: &[u8], tensor_shape: &[usize]) -> Inference {
+        // Reshape the input
+        let tensor: Tensor =
+            unsafe { Tensor::from_raw::<f32>(tensor_shape, tensor_bytes).unwrap() };
```
conradgrobler, line 59: What are the potential consequenses of a malformed Tensor? Are there any potential security risks if the Wasm code sends bad data? Is there some validation that can be done before running unsafe code that parses untrusted data?

oak_functions/loader/src/tf.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        // Get the inference
+        let inference = self.model.run(tvec!(tensor)).unwrap().into_vec();
```
conradgrobler, line 62: This should probably not unwrap (which might panic) but rather return a result.

oak_functions/loader/src/tf.rs
```diff
@@ -0,0 +1,104 @@
...
+            .map(|u| u.into())
+            .collect::<Vec<i32>>();
+        let mut reader = bytes.reader();
+
+        // Creates the model based on the example in https://github.com/sonos/tract/tree/main/examples/tensorflow-mobilenet-v2
+        let model = tract_tensorflow::tensorflow()
+            // load the model
+            .model_for_read(&mut reader)?
+            // specify input type and shape
+            .with_input_fact(0, InferenceFact::dt_shape(f32::datum_type(), dim))?
```
conradgrobler, line 46: This is surprising. I would have expected the stored model to include the input type and shape. Do you know whether there is an option to include that in the stored model so that this is not needed?

docs/oak_functions_abi.md
```diff
@@ -139,3 +139,44 @@ the value in that buffer, and returns it as part of `value_ptr_ptr` and
 
 - `result[0]: i32`: Status of operation as
   [`OakStatus`](https://github.com/project-oak/oak/blob/main/oak_functions/proto/abi.proto)
+
+### `tf_model_get_shape`
```
tiziano88, line 143: I'm not sure why this is needed, isn't the shape part of the model? The caller should know what the shape is anyways.

oak_functions/examples/mobilenet/config.toml
```diff
@@ -0,0 +1,11 @@
+lookup_data_url = ""
+lookup_data_download_period = "10m"
+worker_threads = 4
+
+[tf_model]
+url = "https://storage.googleapis.com/oak-tensorflow-models/mobilenet_v2_1.4_224_frozen"
```
tiziano88, line 6: We should not rely on external resources, could you copy this in third_party and refer to it as a file? Also even in production, we would not want the loader to go and fetch additional files (apart from the key / value entries, by design), they should all be bundled in the Docker image.

oak_functions/loader/src/main.rs
```diff
@@ -67,6 +68,20 @@ struct Config {
...
+    #[serde(default)]
+    tf_model: Option<TensorFlowModelConfig>,
+}
+
+#[derive(Deserialize, Debug, Default)]
+#[serde(deny_unknown_fields)]
+struct TensorFlowModelConfig {
+    /// URL to the TensorFlow model to GET over HTTP.
+    #[serde(default)]
+    url: String,
```
tiziano88, line 81: Following up from my other comment, this should be a local path only.

oak_functions/loader/src/tf.rs
```diff
@@ -0,0 +1,104 @@
...
+use tract_tensorflow::prelude::*;
+type Model = RunnableModel<TypedFact, Box<dyn TypedOp>, Graph<TypedFact, Box<dyn TypedOp>>>;
+
+pub struct TensorFlowModel {
+    model: Model,
+    pub shape: Vec<u8>,
+}
+
+impl TensorFlowModel {
+    /// Creates an instance of TensorFlowModel, by loading the model in the specified path, and
```
tiziano88, line 31: There is no path, I think just the bytes are passed in to this method.

rbehjati (commented):
Thanks for the reviews :)

oak_functions/loader/src/tf.rs
```diff
@@ -0,0 +1,108 @@
...
+            .model
+            .run(tvec!(tensor))
+            .context("could not get inference")?
+            .into_vec();
+
+        let mut shape = vec![inference.len() as u64];
+        shape.append(&mut inference[0].shape().iter().map(|u| *u as u64).collect());
+        let inference_vec = inference
+            .iter()
+            .flat_map(|item| item.to_array_view::<f32>().unwrap())
```
conradgrobler, line 89: Perhaps rather use `.context('...')?` instead of `.unwrap()`? I think it is better to no panic inside a library.

conradgrobler (approved):

oak_functions/loader/src/tf.rs
conradgrobler: I had a look at their code, and I think it is ok. The raw bytes are only used to update the `data` field on the Tensor that is created from the shape. I was worried that the code might doing more complex unsafe casting.

oak_functions/loader/src/tf.rs
rbehjati: Done. The `?` operator cannot be used in the closure, so `flat_map` cannot be used directly on the closure. The type annotation looks a bit ugly but it works.

PR #2242, opened 2021-07-26
Make `permissions` flag optional

rbehjati:
[This `cfg_attr`](https://github.com/project-oak/oak/blob/facc62bff8ad7feed92ffd29c6a59e03838c5e1f/oak_loader/src/options.rs#L48-L51) in the Oak Loader, only applies to the name of the `permissions` flag. When compiling Oak Loader with `oak-unsafe` feature, a permissions parameter is still required: 
```
oak_loader [FLAGS] [OPTIONS] <permissions> --application <application> [--] [ARGS]
```
I tried to fix this by making the `permissions` field in the struct compile conditionally too (using `#[cfg(not(feature = "oak-unsafe"))]`), but it did not solve the problem. So, I changed the `permissions` to be an `Option<String>`.

### DIFF
--- a/examples/aggregator/README.md
+++ b/examples/aggregator/README.md
@@ -58,7 +58,7 @@ localhost. Use the following command to connect the client to an Aggregator
 hosted at a specific domain name; in this case `aggregator.oakexamples.dev`:
 
 ```bash
-./bazel-client-bin/examples/aggregator/client/cpp/client \
+./bazel-bin/examples/aggregator/client/cpp/client \
   --address=aggregator.oakexamples.dev \
   --ca_cert_path=./examples/certs/gcp/ca.pem \
   --bucket=test \
--- a/examples/aggregator/gcp/deployment.yaml
+++ b/examples/aggregator/gcp/deployment.yaml
@@ -28,7 +28,6 @@ spec:
               readOnly: true
           args:
             - --application=aggregator.oak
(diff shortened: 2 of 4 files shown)


rbehjati:
I ended up here while trying to run the following command (from `examples/aggregator/README.md`):
```bash
./bazel-bin/examples/aggregator/client/cpp/client \
  --address=aggregator.oakexamples.dev \
  --ca_cert_path=./examples/certs/gcp/ca.pem \
  --bucket=test \
  --data=1:10,2:20,3:30
```

But even after a successful redeployment of the aggregator I get the following error: 
```
F0726 16:24:27.260879 119912 aggregator.cc:61] Error submitting sample: Unavailable: failed to connect to all addresses
```

Any ideas what the problem might be?

tiziano88 (approved):

conradgrobler (approved):

conradgrobler:
> Any ideas what the problem might be?

If this is still a problem, it could be `--ca_cert_path=./examples/certs/gcp/ca.pem`, which might cause it to ignore the system trusted certs.

rbehjati:
> > Any ideas what the problem might be?
> 
> If this is still a problem, it could be `--ca_cert_path=./examples/certs/gcp/ca.pem`, which might cause it to ignore the system trusted certs.

This did not help. Still getting the same error. 
I'll merge this change for now.

Issue #2244, opened 2021-07-27
Export Differential Privacy based values from Oak Functions


tiziano88:
We should experiment with exporting values during the execution of an Oak Functions module, subject to Differential Privacy (DP) constraint (e.g. epsilon-delta).

For now we can start with allowing the module to export a single boolean value, having the runtime collect those values from various executions over a time window, aggregate them and add sufficient noise to meet the DP constraints for the instance, and finally export them (for now we can log them or export them as metrics).

The time window and the DP parameters should be specified in the configuration file, and they should also be provided by the runtime to the client as part of the remote attestation step (#2243), so that a client is able to detect whether this API is enabled (and with what parameters) *before* sending any data to it.

conradgrobler:
I think we could start with just using Laplacian noise, which would provide epsilon-differential privacy and is stronger than epsilon-delta. Gaussian noise and epsilon-delta-differential privacy could become useful if we distribute the privacy budget over a larger number of labels (Laplacian noise requires scaling epsilon by 1/k per label for k labels, Gaussian noise only requires scaling by 1/sqrt(k)).

My current plan for an initial version is:

- Add an ABI function to report an event with a label (e.g. "error")
- Each Wasm instance can report an event for each label at most once (additional reports are ignored)
- Add configuration option to specify allowed labels and an epsilon privacy budget that will be equally shared over all of them
- Any event reported for a label that is not explicitly allowed is just ignored
- Add configuration option for the reporting batch size
- Once enough Wasm instances have run to match the batch size: calculate the count of event reports for each label, add Laplacian noise based on the label's share of epsilon and log the result

Possible future enhancements
- Option to use Gaussian noise and epsilon-delta parameters
- Allow choosing between batch size or time window to define batched releases
- Allow specifying sub-budgets for labels (e.g. allowing "error" labels to user more of the budget than others)
- Export the results as metrics rather than just logging them

conradgrobler:
One additional potential future enhancement is to support labels that can use predefined integer ranges, rather than just 0 or 1. Values outside the range will be clamped to the limits of the range and noise would be scaled by the size of the range allowed for the specific label.

tiziano88:
[...]

PR #2245, opened 2021-07-29
Add initial support for differentially-private metrics

conradgrobler:
Ref #2244


oak_functions/examples/metrics/README.md
```diff
@@ -0,0 +1,46 @@
...
+one, they should not be able to deduce with high certainty what was sent in the
+remaining request. The addition of Laplacian noise to the batch counts ensures
+this property, assuming appropriate settings are chosen.
+
+The allowed list of event labels, the privacy budget and the batch size are set
+in the server configuration file ('config.toml'). If multiple different event
+labels are allowed the budget would be evenly split across all of them.
+
+Using a privacy budget value (epsilon) of 1.0 and only one allowed label means
+that approximately 39% of the time the batch count will be accurate.
```
tiziano88, line 22: Is the formula something that could be mentioned here, so that a reader has some idea of where the numbers are coming from?

oak_functions/loader/src/tests.rs
```diff
@@ -425,3 +428,104 @@ fn test_format_bytes() {
...
+    let mut max_diff = 0.0;
+    for (index, actual) in histogram.iter().enumerate() {
+        let test = expected[index];
+        let diff = (test - *actual as f64).abs() / test;
+        assert!(diff >= 0.0 && diff <= margin);
+        if diff > max_diff {
+            max_diff = diff;
+        }
+    }
+    println!("Maximum required margin: {}", max_diff);
```
rbehjati, line 524: Is there anything that can be checked about `max_diff`?

oak_functions/examples/metrics/config.toml
```diff
@@ -0,0 +1,12 @@
+lookup_data_url = "https://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
+lookup_data_download_period = "10m"
```
tiziano88, line 2: Perhaps this is not necessary for this example?

oak_functions/examples/metrics/config.toml
```diff
@@ -0,0 +1,12 @@
...
+worker_threads = 4
+
+[metrics]
+epsilon = 1.0
+batch_size = 20
+allowed_labels = ["a"]
```
tiziano88, line 8: I kind of feel that this should be under policy, since it affects the privacy guarantees of the server, and it would be nice if the client only needs to look at the policy section in order to figure out what the guarantees are, rather than looking at fields spread out in the file. Anyways maybe we can refactor things once something like https://github.com/project-oak/oak/pull/2247#issuecomment-891227572 is implemented.

oak_functions/examples/metrics/module/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+//
+
+//! Oak Functions differentially private metrics example.
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn main() {
+    let request = oak_functions::read_request().expect("Couldn't read request body.");
+    if request == b"a" {
+        oak_functions::report_event("a").expect("Couldn't report event.");
+    }
```
tiziano88, line 24: May as well always report whatever the client sent? Also that may make for a nice general purpose and reusable Wasm logic, that we could literally reuse for some use cases, just by changing the list of labels and the epsilon in the config file. But it would already work for some interesting use cases I think.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    sync::{Arc, Mutex},
+};
+
+/// Configuration for differentially-private metrics reporting.
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct PrivateMetricsConfig {
+    pub epsilon: f64,
```
tiziano88, line 30: Add links to explain the meaning of the params if possible.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub batch_size: usize,
+    pub allowed_labels: Vec<String>,
+}
+
+/// Aggregator for count-based differentially private metrics.
```
tiziano88, line 35: Could you mention what happens when the threshold is reached? It looks like it logs it out using the standard logger, we may also want some other mechanism in the future. In fact I think having the logger inside this struct is unnecessarily strict, and also makes testing harder. It should be possible to just get the aggregated result (as values ideally instead of a string), and for the caller to log it in whatever way it wants, or, in tests, to make assertions on it. And in general, it is preferable to keep side effects separate from pure computation.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+                self.events.insert(label.to_owned(), 1);
+            }
+        }
+
+        if self.count == self.batch_size {
+            self.export_events();
+        }
+    }
+
+    /// Logs the current counts after adding appropriate noise, resets the batch count to 0 and
```
tiziano88, line 75: I think the batch count stays the same, right? I mean, what is called `batch_size`. Or did you mean `count` vs `size`? Naming is hard :) Maybe use `event_count` vs `batch_size` in the struct?

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn export_events(&mut self) {
+        let counts: Vec<String> = self
+            .allowed_labels
+            .iter()
+            .map(|label| {
+                let value = self.add_laplace_noise(*self.events.get(label).unwrap_or(&0));
+                format!("{}={}", label, value)
+            })
+            .collect();
+        self.logger.log_public(
```
tiziano88, line 86: Add a comment explaining why this is fine to log publicly.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+                counts.join(";"),
+            ),
+        );
+        self.count = 0;
+        self.events.clear();
+    }
+
+    /// Exports the batch count and bucket counts without adding any noise for use in testing. Also
+    /// resets the batch count to 0 and clears the aggregated data.
+    pub fn export_events_for_test(&mut self) -> (usize, Vec<(String, usize)>) {
```
tiziano88, line 100: Both these methods are called `export_*`, but one logs, while the other returns the things to log. Maybe rename them, or make them more consistent? e.g. this method may log non-publicly, which would work only in the unsafe version anyways.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Aggregator for count-based differentially private metrics.
+pub struct PrivateMetricsAggregator {
+    count: usize,
+    epsilon: f64,
+    batch_size: usize,
+    allowed_labels: HashSet<String>,
+    label_count: f64,
+    events: HashMap<String, usize>,
```
tiziano88, line 42: An alternative would be to just populate the hashmap with one entry per label, initialized to zero; then when incrementing, only use entries that are already present (possibly to zero, but still present). Basically, there is no need for an additional Set. Up to you anyways if you prefer the current approach. Also label_count seems redundant.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+        if p <= 0.0 {
+            return f64::NEG_INFINITY;
+        }
+        let u = p - 0.5;
+        -beta * u.signum() * (1.0 - 2.0 * u.abs()).ln()
+    }
+}
+
+/// Proxy for use by request handler instances to push metrics to the `MerticsAggregator`.
```
tiziano88, line 142:
```suggestion
/// Proxy for use by request handler instances to push metrics to the `MetricsAggregator`.
```

rbehjati (approved):
Nice!

oak_functions/loader/src/tests.rs
```diff
@@ -425,3 +428,104 @@ fn test_format_bytes() {
...
+fn test_private_metrics_aggregator() {
+    let config = new_metrics_config(1.0, vec!["a", "b", "c", "d"]);
+    let aggregator = Arc::new(Mutex::new(PrivateMetricsAggregator::new(
+        &config,
+        Logger::for_test(),
+    )));
+
+    let mut proxy1 = PrivateMetricsProxy::new(aggregator.clone());
+    proxy1.report_event("a");
+    proxy1.publish();
```
tiziano88, line 442: What's happening to the published metrics? Does the test logger allow reading what comes out of it? Also see my other comment re: not using a logger in the first place.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+        (count, buckets)
+    }
+
+    /// Adds Laplacian noise to a count. The Laplacian noise is sampled by sampling from a uniform
+    /// distribution and calculating the inverse of the Laplace cummulative distribution function on
+    /// the sampled value. Rounding of the noise is allowed as acceptable post-processing.
+    pub fn add_laplace_noise(&self, count: usize) -> i64 {
+        // If epsilon is 0 (or smaller), always return a fixed value so we don't leak any
+        // information.
+        if self.epsilon <= 0.0 {
```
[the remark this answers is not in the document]
conradgrobler: Removed here and added a check on `PrivateMetricsAggregator::new`.

oak_functions/loader/src/metrics.rs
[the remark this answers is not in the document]
conradgrobler: Good point. Done. Turns out this does not work with intergation tests, so I moved the metrics tests into the metrics module.

oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs
```diff
@@ -100,6 +103,7 @@ fuzz_target!(|instruction_list: Vec<ArbitraryInstruction>| {
         Arc::new(LookupData::for_test(entries)),
         Arc::new(None),
         Logger::for_test(),
+        None,
```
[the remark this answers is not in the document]
conradgrobler: Create an issue to track this and added a TODO.

conradgrobler (commented):
Thanks for the reviews. PTAL

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,293 @@
...
+            &format!(
+                "metrics export: batch size: {}; counts: {}",
+                self.count,
+                counts.join(";"),
+            ),
+        );
+        self.reset();
+    }
+
+    /// Resets the request count all the bucket counts to 0 and the batch.
```
rbehjati, line 113:
```suggestion
    /// Resets the request count and all the bucket counts to 0.
```

rbehjati (approved):
LGTM. Thanks :)

oak_functions/loader/src/metrics.rs
tiziano88: I think the API should be more "synchronous". When the `report_events` method is called, the caller should get back either None, or the aggregated values (and internally reset the state). Then the caller decides what to do with it. Also in this way there is no need for a separate testing method, I think.

oak_functions/loader/src/metrics.rs
conradgrobler: Thanks, done.

oak_functions/loader/src/metrics.rs
conradgrobler: Done. I don't really like all the conditional compilation in there, but could not think of a better way to do it.

oak_functions/loader/src/main.rs
```diff
@@ -168,6 +171,13 @@ async fn async_main(opt: Opt, config: Config, logger: Logger) -> anyhow::Result<
     let address = SocketAddr::from((Ipv6Addr::UNSPECIFIED, opt.http_listen_port));
     let tee_certificate = vec![];
 
+    let aggregator = match &config.metrics {
+        Some(metrics_config) => Some(Arc::new(Mutex::new(PrivateMetricsAggregator::new(
+            metrics_config,
+        )?))),
+        None => None,
+    };
```
tiziano88, line 179: (optional) could use `config.metrics.map(|c| ... )`

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// bucket counts are reset to 0. By default Laplacian noise is added to each of the aggregated
+    /// bucket counts that are returned. The metrics are returned as a tuple containing the batch
+    /// size and a vector of tuples cotaining the label and count for each bucket.
+    ///
+    /// If the `oak-unsafe` feature is enabled, or this function is called from a unit test, no
+    /// noise is added.
+    pub fn report_events(
+        &mut self,
+        events: HashSet<String>,
+    ) -> Option<(usize, Vec<(String, i64)>)> {
```
tiziano88, line 80: (optional) alternatively you could return `HashMap<String, i64>` here too. Or did you consider and decide against it?

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+                .map(|(label, count)| (label.to_string(), self.add_noise(*count)))
+                .collect();
+            self.reset();
+            Some((self.batch_size, metrics))
+        } else {
+            None
+        }
+    }
+
+    /// Adds no noise when the `oak-unsafe` feature is enabled or running as part of a unit test.
```
tiziano88, line 101: Uhmm I would avoid this if possible, ideally both in tests and in oak-unsafe we would like the behaviour to be closest to the production one, right? I don't think there is a use case in which when running with oak-unsafe we would expect to see the non-noisy aggregation. In fact that may make debugging even harder.

If anything, maybe make the RNG configurable so that in tests a deterministic PRNG with a fixed seed is used, and results can be verified more easily.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn inverse_laplace(beta: f64, p: f64) -> f64 {
+        if p >= 1.0 {
+            return f64::INFINITY;
+        }
+        if p <= 0.0 {
+            return f64::NEG_INFINITY;
+        }
+        let u = p - 0.5;
+        -beta * u.signum() * (1.0 - 2.0 * u.abs()).ln()
+    }
```
tiziano88, line 143: (optional) Perhaps these should be extracted as stand-alone functions (not defined on the struct), since they are quite "pure", at least conceptually (apart from reading the number of events and epsilon).

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Adds Laplacian noise to a count. The Laplacian noise is sampled by sampling from a uniform
+    /// distribution and calculating the inverse of the Laplace cummulative distribution function on
+    /// the sampled value. Rounding of the noise is allowed as acceptable post-processing.
+    pub fn add_laplace_noise(&self, count: usize) -> i64 {
+        // Split the budget evenly over all of the labeled buckets.
+        let beta = self.events.len() as f64 / self.epsilon;
+        let p: f64 = thread_rng().sample(Open01);
+        count as i64 + Self::inverse_laplace(beta, p).round() as i64
```
tiziano88, line 128: IIUC this means that if the count is zero, only half of the noise gets added because of the post-processing rounding, right? Is that fine?

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+        if p <= 0.0 {
+            return f64::NEG_INFINITY;
+        }
+        let u = p - 0.5;
+        -beta * u.signum() * (1.0 - 2.0 * u.abs()).ln()
+    }
+}
+
+/// Proxy for use by request handler instances to push metrics to the `PrivateMetricsAggregator`.
```
tiziano88, line 146: I'm not entirely sure why this proxy is needed, and why it holds both an aggregator and also a list of non-aggregated events?

tiziano88 (approved):

oak_functions/loader/src/main.rs
conradgrobler: I previously used that before `PrivateMetricsAggregator::new` returned `Result<...>`, but the `?` means that it can no longer be used in a closure.

oak_functions/loader/src/metrics.rs
conradgrobler: Done.

oak_functions/loader/src/metrics.rs
conradgrobler: Done.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,307 @@
...
+                .allowed_labels
+                .iter()
+                .map(|label| (label.clone(), 0))
+                .collect(),
+            rng: StdRng::from_rng(thread_rng()).context("Couldn't create rng")?,
+        })
+    }
+
+    #[cfg(test)]
+    pub fn new_for_test(config: &PrivateMetricsConfig, rng: StdRng) -> anyhow::Result<Self> {
```
tiziano88, line 73: It would be nice to share this fn with the regular constructor, since there is some logic going on here; maybe rename to `new_with_rng` or similar, leave it private, then use it from `new` and also from `new_for_test` (or just `cfg(test)` and use `new_with_rng` directly from tests). Up to you.

tiziano88 (approved):

oak_functions/loader/src/metrics.rs
conradgrobler: Good idea. Done.

oak_functions/loader/src/metrics.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    /// Reports new events that should be included in the aggregated counts. If the number of
+    /// requests do not yet match the batch size `None` is returned. If the batch threshold
+    /// is reached the aggregated metrics for the batch are returned and the request count and
+    /// bucket counts are reset to 0. By default Laplacian noise is added to each of the aggregated
+    /// bucket counts that are returned. The metrics are returned as a tuple containing the batch
+    /// size and a vector of tuples cotaining the label and count for each bucket.
+    ///
+    /// If the `oak-unsafe` feature is enabled, or this function is called from a unit test, no
+    /// noise is added.
```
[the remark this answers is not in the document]
conradgrobler: Ah, yes. Removed.

conradgrobler (commented):
Thanks.

PR #2247, opened 2021-07-30
Add `oak-tf` feature to enable TensorFlow API

rbehjati:


tiziano88:
I wonder whether we should think about a more modular approach, by which we can define additional APIs as separate building blocks, and mix and match which ones are enabled both at compile time and run time. 

e.g. we may have a trait

```rust
trait OakApiNativeExtension {
  fn new(config: &Config) -> Self;
  // Similar to wasmi::Externals, but with additional optionality.
  fn invoke_index(
      &mut self,
      index: usize,
      args: wasmi::RuntimeArgs,
  ) -> Option<Result<Option<wasmi::RuntimeValue>, wasmi::Trap>>;
  // Similar to wasmi::ModuleImportResolver, but with additional optionality.
  fn resolve_func(
      &self,
      field_name: &str,
      signature: &wasmi::Signature,
  ) -> Option<Result<wasmi::FuncRef, wasmi::Error>>;
}
```

Which essentially gets dispatched to by the following existing functions:

https://github.com/project-oak/oak/blob/d3243879423c303e68746d00e986509c21ccbfc3/oak_functions/loader/src/server.rs#L389-L425

https://github.com/project-oak/oak/blob/d3243879423c303e68746d00e986509c21ccbfc3/oak_functions/loader/src/server.rs#L427-L435

Alternatively we could make each extension provide a list of (name, signature, implementation) entries, and keep the mapping inside the server.

We would just register the list of Native extensions to make available to the Wasm engine, and each would provide its own implementation. 

We could also move the logging and k/v lookup to separate extensions in the same way at some point (we don't need to feature gate them, just to keep the code organized and the side effects clearly documented).

In fact, if you think about it this is really not much different to how we have pseudo-nodes in the original Oak runtime, which is not by chance :)

rbehjati:
Great idea. I think it makes the design a lot nicer.

> We would just register the list of Native extensions to make available to the Wasm engine, and each would provide its own implementation.

Do you have any suggestions for the instantiation of the extensions? We could use a hash map (if that is what you mean by registering the list of extensions), or we could have explicit `Option` fields for each extension in `WasmState`. I think the latter makes the code more readable, and debugging easier, but it would be more verbose compared to the hash map approach.

tiziano88:
I was thinking of just a Vec of extensions, and for each call to `invoke_index` or `resolve_func` we would iterate until we find the first (ideally only) extension that returns `Some` value.

But if we implement the slightly more complex approach with extensions returning a wrapper around the invocation callable, alongside the name and index of the exported fn, then we can indeed use two hash maps (one indexed by fn name, the other by fn index).

PR #2254, opened 2021-08-04
Oak 2243 config in ra

rbehjati:
# Checklist

      construction.


remote_attestation/proto/remote_attestation.proto
```diff
@@ -50,14 +50,30 @@ message AttestationInfo {
...
+// Additional info to include in the AttestationIdentity. The hash of this field must be included
+// and signed in the AttestationReport. AdditionalInfo could include server configuration details,
+// and inclusion proofs.
+message AdditionalInfo {
+  // Binary protobuf-encoded data. Since protobuf encoding is not deterministic, this field cannot
+  // be directly used for computing the hash.
+  bytes data = 1;
+  // Hash of the data. The hash must have been calculated based on a canonical representation of the
+  // data.
+  bytes hash = 2;
```
ipetr0v, line 62: Currently the hash of the public key is included in the attestation report (inside the `AttestationInfo`).
https://github.com/project-oak/oak/blob/f8eccbd08c19af3e596132c472e37b8fd9300fb5/remote_attestation/rust/src/attestation.rs#L198

Do you think it's better to not include the AdditionalInfo::hash as part of the protobuf message, but combine it with the public key, hash it, and put in the attestation report.

Since we will need this hash to be signed by hardware anyway.

rbehjati: > but combine it with the public key, hash it, and put in the attestation report

This PR actually does combine the public key hash and this hash. What goes into the attestation report is `hash(hash(public_key) + hash(additional_info.data))`.

And I agree that `hash` can be removed from here, and be computed by the server and the client.

PR #2263, opened 2021-08-13
Port part of the Google Go differential privacy library to Rust

conradgrobler:
This is a port of a minimal subset of the Google Go differential privacy library at https://github.com/google/differential-privacy/

Only the minimum features needed to support the current differentially private metrics implementation in Oak Functions are included, specifically adding Laplace noise.

Note for reviewers who might not be familiar with the github review UI: large diffs are not displayed by default, so the two largest files (laplace_noise.rs and secure_noise_math.rs) must be manually loaded during the review by clicking on "Load diff".


conradgrobler:
This PR is not relevant at the moment, seeing that we have removed the differentially private metrics module for now.

Issue #2261, opened 2021-08-12
Align remote attestation handshake / protocol with internal SC version


tiziano88:
cc @sidtelang

PR #2271, opened 2021-08-25
Update Remote Attestation Protocol

ipetr0v:
This change updates Rust and Java versions of the Remote Attestation protocol to match SC:
- Protocol uses 3 message handshake: `AttestationInit`, `ServerIdentity`, `ClientIdentity `
- Random values are added to the messages to prevent replay attacks
- Transcripts of the sent/received messages are hashed and signed with a ECDSA-P256 keys for preventing replay attacks
- Remote Attestation relies on the persistent `signing_key` that was signed by the TEE hardware key

Ref https://github.com/project-oak/oak/issues/2261


oak_functions/client/rust/src/attestation.rs
```diff
@@ -130,11 +135,21 @@ impl AttestationClient {
                 Err(anyhow!("Received incorrect message type"))
             }?;
 
-        // Remotely attest server.
-        let encryptor = attestation_engine
-            .create_client_encryptor(&server_identity)
-            .context("Couldn't attest server")?;
+        // Remotely attest the server and create:
+        // - Client attestation identity containing client's ephemeral public key
+        // - Encryptor used for for decrypting/encrypting messages between client and server
```
tiziano88, line 140:
```suggestion
        // - Encryptor used for decrypting/encrypting messages between client and server
```

oak_functions/client/rust/src/attestation.rs
```diff
@@ -81,45 +81,50 @@ impl Client {
...
-            .context("Couldn't get server identity")?;
-
-        // Create attestation request with client's public key.
+        let attestation_engine = AttestationEngine::<attestation::Client, Initializing>::new(
+            AttestationBehavior::create_peer_attestation(expected_tee_measurement)
+                .context("Couldn't create peer attestation behavior")?,
+        );
+
+        // Send attestation initialization message.
+        let (attestation_init, attestation_engine) = attestation_engine
```
tiziano88, line 111: I am not sure I understand the point of AttestationEngine always returning another instance of AttestationEngine. What is the advantage of doing it this way?

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+    P: AttestationParticipant,
 {
-    /// Type of the remote attestation process.
-    attestation_type: S,
-    /// Implementation of the X25519 Elliptic Curve Diffie-Hellman (ECDH) key negotiation.
-    key_negotiator: KeyNegotiator,
+    fn transition(
+        self,
+        key_negotiator: KeyNegotiator,
+    ) -> anyhow::Result<AttestationEngine<P, Attesting>> {
```
tiziano88, line 43: I don't fully see the point of the generic type here; in what case would you ever use an AttestationEngine without caring about the specific state it is in? It seems to me these may as well be completely distinct types, e.g. AttestationEngineInit, AttestationEngineAttesting, etc. (probably with better naming than the ones I am suggesting here), and explicit methods to move from one to the other. In particular, is there any functionality that is common to all the states (apart from the ability to move to the next state)? I'm probably missing something though, so please feel free to correct me :)

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+        self.state.encryptor
     }
 }
 
-impl AttestationEngine<UnattestedSelf> {
-    pub fn create(tee_certificate: &[u8]) -> anyhow::Result<Self> {
-        let key_negotiator = KeyNegotiator::create().context("Couldn't create key negotiator")?;
+pub trait AttestationParticipant {}
+impl AttestationParticipant for Client {}
+impl AttestationParticipant for Server {}
```
tiziano88, line 400: This seems... unusual :) still not sure exactly what it does in terms of type safety; is there anything that this design prevents at compile time instead of run time? I cannot thin of anything obvious

remote_attestation/rust/src/attestation.rs
```diff
@@ -273,3 +580,12 @@ fn verify_attestation(
         Err(anyhow!("Incorrect TEE measurement"))
     }
 }
+
+// TODO(#2273): Use raw bytes for messages since Protobuf is not deterministic.
```
tiziano88, line 584: Not only that, but I do think the protocol should operate on bytes anyways; i.e. it should return the exact bytes to send over the wire to the other side, not a proto message. Since the caller would not do anything with those proto bytes apart from converting them to bytes immediately anyways. Anyways, let's discuss that separately on the dedicated issue.

remote_attestation/rust/src/crypto.rs
```diff
@@ -136,24 +137,28 @@ impl AeadEncryptor {
...
 /// Implementation of the X25519 Elliptic Curve Diffie-Hellman (ECDH) key negotiation.
 /// https://datatracker.ietf.org/doc/html/rfc7748#section-6.1
 pub struct KeyNegotiator {
+    r#type: KeyNegotiatorType,
     private_key: agreement::EphemeralPrivateKey,
 }
 
 impl KeyNegotiator {
-    pub fn create() -> anyhow::Result<Self> {
+    pub fn create(r#type: KeyNegotiatorType) -> anyhow::Result<Self> {
```
tiziano88, line 154: I think it is more idiomatic to just use a non-conflicting name for the parameter, e.g. `type_`

oak_functions/client/rust/src/attestation.rs
ipetr0v: Since each state is a new class, and transition functions take `self`, we need to create this class and return it.

remote_attestation/rust/src/attestation.rs
ipetr0v: Empty classes were added more for readability.
Because if I remove `participant` and `state` classes, then the code will contain:
- `ClientAttestationEngineInit`
- `ClientAttestationEngineAttesting`
- `ClientAttestationEngineAttested`
- `ServerAttestationEngineInit`
- `ServerAttestationEngineAttesting`
- `ServerAttestationEngineAttested`

And having `AttestationEngine<Participant, State>` seems a bit nicer.

Also making `AttestationEngine` a parent class is currently not possible, because each `<Participant, State>` returns a separate Protobuf message. So processing handshake messages is basically a new function everytime (It could be updated, but only after implementing https://github.com/project-oak/oak/issues/2273, when all responses will look like `Vec<u8>`).

remote_attestation/rust/src/attestation.rs
ipetr0v: As I said in the previous comment, these were added more for readability and because each protocol engine state is basically a new class (since it operates on different Protobuf messages).

remote_attestation/rust/src/crypto.rs
ipetr0v: Done

oak_functions/client/rust/src/attestation.rs
ipetr0v: Done

oak_functions/client/rust/src/attestation.rs
ipetr0v: The advantage is that we don't have an internal variable with a state. And all the errors with incorrect state transitions will be caught by a compiler (and not by `if state == Attested`).

remote_attestation/rust/src/attestation.rs
ipetr0v: It also depends on the SC protocol being fleshed out.
Because currently it doesn't specify all the fields exactly (i.e. all fields related to TEE are still vague).

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+const ATTESTATION_PROTOCOL_VERSION: u8 = 1;
+/// Size (in bytes) of a random array sent in messages to prevent replay attacks.
+const REPLAY_PROTECTION_ARRAY_SIZE_BYTES: usize = 32;
+
+/// Convenience trait for transitioning between states in the state machine.
+pub trait StateMachine<T, A> {
+    fn transition(self, argument: A) -> anyhow::Result<T>;
+}
+
+impl<P> StateMachine<AttestationEngine<P, Attesting>, KeyNegotiator>
```
rbehjati, line 35: Do we have any tests for this crate?

remote_attestation/rust/src/crypto.rs
```diff
@@ -258,19 +260,88 @@ impl KeyNegotiator {
...
+            .key_pair
+            .sign(&rng, input)
+            .map_err(|error| anyhow!("Couldn't sign input: {:?}", error))?
+            .as_ref()
+            .to_vec();
+        Ok(signature)
+    }
+}
+
+pub struct SignatureVerifier {
```
rbehjati, line 310: If this should be kept in sync with the java version it might be good to add a comment about it here.

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+                .context("Couldn't append server identity to transcript")?;
+
+            // Add transcript signature to the server identity message.
+            let transcript_signature = signer
+                .sign(&self.transcript.get_sha256())
+                .context("Couldn't create transcript signature")?;
+            server_identity.transcript_signature = transcript_signature;
+            server_identity
+        } else {
+            ServerIdentity {
```
rbehjati, line 224: Why is not this being added to `self.transcript`?

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+    pub fn contains_peer_attestation(&self) -> bool {
+        self.expected_tee_measurement.is_some()
+    }
 
-impl UnattestedSelf {
-    pub fn new(tee_certificate: &[u8]) -> Self {
-        Self {
-            tee_certificate: tee_certificate.to_vec(),
-        }
+    pub fn contains_self_attestation(&self) -> bool {
```
rbehjati, line 453: What does self attestation mean? I don't understand at what stage of the handshake we can say that something contains self attestation. From the code it seems that only in the case of having self attestation the identity is added to the transcript, but I don't understand why.

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+        mut self,
+        attestation_init: &AttestationInit,
+    ) -> anyhow::Result<(ServerIdentity, AttestationEngine<Server, Attesting>)> {
+        // Create server identity message.
+        let key_negotiator = KeyNegotiator::create(KeyNegotiatorType::Server)
+            .context("Couldn't create key negotiator")?;
+        let ephemeral_public_key = key_negotiator
+            .public_key()
+            .context("Couldn't get ephemeral public key")?;
+        let server_identity = if self.behavior.contains_self_attestation() {
```
rbehjati, line 184: Is it possible to have the creation of the `server_identity` as a method in `AttestationBehavior`, and only update `self.transcript` and `server_identity.transcript_signature` in this function?

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+    /// `ClientIdentity` message contains an ephemeral public key for negotiating session keys.
+    /// If self attestation is enabled this message also provides necessary information to perform
+    /// remote attestation.
+    ///
+    /// Transitions [`AttestationEngine`] state from [`Attesting`] to [`Attested`] state.
+    pub fn process_server_identity(
+        mut self,
+        server_identity: &ServerIdentity,
+    ) -> anyhow::Result<(ClientIdentity, AttestationEngine<Client, Attested>)> {
+        if self.behavior.contains_peer_attestation() {
```
rbehjati, line 254: Could this if-block be a `verify_server_identity` method in `AttestationBehavior` to make this method more readable?

remote_attestation/rust/src/attestation.rs
```diff
@@ -15,238 +15,545 @@
...
+                .context("Couldn't verify server attestation info")?;
+        }
+
+        // Create client identity message.
+        let ephemeral_public_key = self
+            .state
+            .key_negotiator
+            .public_key()
+            .context("Couldn't get ephemeral public key")?;
+        let client_identity = if self.behavior.contains_self_attestation() {
```
rbehjati, line 290: Again could there be a method in `AttestationBehavior` for creating `client_identity`?

remote_attestation/rust/src/attestation.rs
rbehjati: I am also no longer entirely sure what benefit we are getting from using the state machine pattern. Is it readability, or correctness or compile-time error checking? I guess it will be clearer after #2273.

rbehjati (approved):

remote_attestation/rust/src/attestation.rs
ipetr0v: `AttestationBehavior` currently is server/client independant.
But the `verify_server_identity` will not be, since `ClientIdentity` and `ServerIdentity` messages are different (i.e. `ServerIdentity` has additional fields).

remote_attestation/rust/src/attestation.rs
ipetr0v: Since it's a refactoring, I decided to add tests as a separate PR. So that all abstractions a "code-reviewed" first.

tiziano88 (approved):

remote_attestation/rust/src/attestation.rs
ipetr0v: `Server` and `Client` have been removed.
Currently there are sepatate `ClientAttestationEngine` and `ClientAttestationEngine` types.

remote_attestation/proto/remote_attestation.proto
```diff
@@ -50,19 +50,46 @@ message AttestationInfo {
...
+  // Random nonce (initialization vector) used for encryption/decryption.
+  bytes nonce = 1;
+  // Data encrypted using the session key.
+  bytes data = 2;
+}
+
 // TODO(#2105): Implement challenge-response in remote attestation.
 // TODO(#2106): Support various claims in remote attestation.
-message AttestationIdentity {
+message AttestationInit {
```
tiziano88, line 62: Is this the same as ClientHello?

remote_attestation/proto/remote_attestation.proto
```diff
@@ -50,19 +50,46 @@ message AttestationInfo {
...
+
+message ServerIdentity {
+  // Remote attestation protocol version.
+  int32 version = 1;
   // Public key needed to establish a session key.
-  bytes public_key = 1;
+  bytes ephemeral_public_key = 2;
+  // Random vector sent in messages for preventing replay attacks.
+  bytes random = 3;
+  // Signed SHA-256 hash of all previously send and received messaged.
```
tiziano88, line 74: AFAICT this is just the signature of the hash, not the signed hash, right? i.e. the hash is not in this field

tiziano88 (approved):

remote_attestation/proto/remote_attestation.proto
ipetr0v: Yes, it's the same

ipetr0v: Renamed to `ClientHello`

remote_attestation/proto/remote_attestation.proto
ipetr0v: Updated the comment

PR #2278, opened 2021-09-03
Adds generic support for native API extensions to Oak Functions

rbehjati:
* Implements a more modular approach for adding optional native APIs to the Wasm engine. Additional APIs must implement the `OakApiNativeExtension` trait and be registered as an extension when creating the `WasmHandler`. 
* Modifies `TensorFlowModel` to an extension implementing `OakApiNativeExtension`.
* Improves feature gating for ML-inference with `TensorFlowModel` using `oak-tf` and `oak-unsafe`.

Ref. #2247


tiziano88:
Can the lookup functionality also be modelled as an extension at some point? Not right now, but worth thinking whether that would work with the current interface.

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
     response_bytes: Vec<u8>,
     lookup_data: Arc<LookupData>,
-    #[allow(dead_code)]
-    tf_model: Arc<Option<TensorFlowModel>>,
     instance: Option<wasmi::ModuleRef>,
     memory: Option<wasmi::MemoryRef>,
     logger: Logger,
     metrics_proxy: Option<PrivateMetricsProxy>,
+    extensions_indexes: Arc<RwLock<HashMap<usize, Arc<BoxedExtension>>>>,
+    extensions_names: Arc<RwLock<HashMap<String, Arc<BoxedExtension>>>>,
```
tiziano88, line 193: Intuitively I would not expect all this synchronization around a hashmap that essentially never changes once created; why is it necessary?

If anything, the synchronization should be around each Extension, if a single instance is assumed to not be invoked concurrently.

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
     request_bytes: Vec<u8>,
     response_bytes: Vec<u8>,
     lookup_data: Arc<LookupData>,
-    #[allow(dead_code)]
-    tf_model: Arc<Option<TensorFlowModel>>,
     instance: Option<wasmi::ModuleRef>,
     memory: Option<wasmi::MemoryRef>,
     logger: Logger,
     metrics_proxy: Option<PrivateMetricsProxy>,
+    extensions_indexes: Arc<RwLock<HashMap<usize, Arc<BoxedExtension>>>>,
```
tiziano88, line 192: (nit) the correct plural of index is indices, not indexes (though I think both are accepted these days, so not a big deal)

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
     }
 }
 
+/// Trait for implementing extensions, to implement new native functionality.
+pub trait OakApiNativeExtension {
+    /// Similar to `invoke_index` in [`wasmi::Externals`], but may return a result to be
+    /// written into the memory of the `WasmState`.
+    fn invoke_index(
+        &self,
+        wasm_state: &WasmState,
```
tiziano88, line 157: Is it possible to pass `&mut WasmState` here, so that the function can just write directly to memory? It seems odd that it can read directly from it, but it needs to return data via a custom mechanism in order for it to be written.

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
+        args: wasmi::RuntimeArgs,
+    ) -> Result<Result<Option<ExtensionResult>, OakStatus>, wasmi::Trap>;
+
+    /// Similar to `resolve_func` in [`wasmi::ModuleImportResolver`], but returns a tuple with a
+    /// function index and a function signature to be converted into [`wasmi::FuncRef`].
+    fn resolve_func(&self) -> anyhow::Result<(usize, wasmi::Signature)>;
```
tiziano88, line 163: IIUC this only works if the extension exposes a single function; for multiple functions, the API would be more complex. In particular, for the current API, the method `invoke_index` is not appropriately named, since the index does not even appear in the function signature.

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
+    /// Similar to `invoke_index` in [`wasmi::Externals`], but may return a result to be
+    /// written into the memory of the `WasmState`.
+    fn invoke_index(
+        &self,
+        wasm_state: &WasmState,
+        args: wasmi::RuntimeArgs,
+    ) -> Result<Result<Option<ExtensionResult>, OakStatus>, wasmi::Trap>;
+
+    /// Similar to `resolve_func` in [`wasmi::ModuleImportResolver`], but returns a tuple with a
+    /// function index and a function signature to be converted into [`wasmi::FuncRef`].
```
tiziano88, line 162: Could you clarify how to pick these indices, what they mean, what properties they should have, do they need to be sequential, etc.?

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
+    fn resolve_func(&self) -> anyhow::Result<(usize, wasmi::Signature)>;
+
+    /// Registration info, including a function index and a function name, for registering this
+    /// extension in the [`WasmState`].
+    fn registration_info(&self) -> (usize, String);
```
tiziano88, line 167: Does this index need to be the same as the one from `resolve_func`? If so, why the coupling?

oak_functions/loader/src/server.rs
```diff
@@ -514,7 +488,34 @@ impl wasmi::ModuleImportResolver for WasmState {
         field_name: &str,
```
tiziano88, line 488: (pre-existing) Why is this called `field_name`? Isn't it the  function name?

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,54 @@ impl FixedSizeBodyPadder for Response {
...
+        &self,
+        wasm_state: &WasmState,
+        args: wasmi::RuntimeArgs,
+    ) -> Result<Result<Option<ExtensionResult>, OakStatus>, wasmi::Trap>;
+
+    /// Similar to `resolve_func` in [`wasmi::ModuleImportResolver`], but returns a tuple with a
+    /// function index and a function signature to be converted into [`wasmi::FuncRef`].
+    fn resolve_func(&self) -> anyhow::Result<(usize, wasmi::Signature)>;
+
+    /// Registration info, including a function index and a function name, for registering this
```
tiziano88, line 165: Is the function index part of the ABI? i.e. if we change the index in the future, does it invalidate previous Wasm modules? My guess is not, since the index is always looked up by name, but I am actually not sure. Could you clarify either way?

oak_functions/loader/src/tf.rs
```diff
@@ -14,34 +14,43 @@
...
+use prost::Message;
 use std::{fs::File, io::Read};
-#[cfg(feature = "oak-tf")]
 use tract_tensorflow::prelude::{
     tract_ndarray::{ArrayBase, Dim, IxDynImpl, ViewRepr},
     *,
 };
+use wasmi::ValueType;
+
+const TF_MODEL_INFER: usize = 4;
```
tiziano88, line 35: Is this the fn index? Could you clarify?

tiziano88 (commented):
Very nice!

rbehjati:
> Can the lookup functionality also be modelled as an extension at some point? Not right now, but worth thinking whether that would work with the current interface.

I think it should be possible. But I remember earlier we thought it might not work. Was it because of the refresh functionality? I think even with the lookup data as an extension we can keep the load and refresh functionalities as we have now.

conradgrobler (commented):
Nice.

I don't think it is worth exploring for this PR, but one thing that came up when looking at the code is that many of the extension have a generic interface: bytes in, bytes out. The request handler has the same interface. Perhaps in future we can unify the design so that it could be possible to bypass the Wasm logic (via a feature) and directly call the lookup (or TF infer) with the incoming request bytes and directly respond with the results.

oak_functions/loader/src/server.rs
```diff
@@ -523,19 +524,21 @@ impl WasmState {
         module: &wasmi::Module,
         request_bytes: Vec<u8>,
         lookup_data: Arc<LookupData>,
-        tf_model: Arc<Option<TensorFlowModel>>,
         logger: Logger,
         metrics_proxy: Option<PrivateMetricsProxy>,
+        extensions_indexes: Arc<RwLock<HashMap<usize, Arc<BoxedExtension>>>>,
```
ipetr0v, line 529: Could you please add a comment, what is extension index?

ipetr0v (approved):

ipetr0v (approved):

oak_functions/loader/src/main.rs
```diff
@@ -160,7 +162,15 @@ async fn async_main(opt: Opt, config: Config, logger: Logger) -> anyhow::Result<
 
     let lookup_data = load_lookup_data(&config, logger.clone()).await?;
 
-    let tf_model = load_tensorflow_model(&config).await?;
+    let extensions = if cfg!(feature = "oak-unsafe") {
+        if let Some(tf_model) = load_tensorflow_model(&config, logger.clone()).await? {
```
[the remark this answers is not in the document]
rbehjati: I think, the plan is to have a separate feature for each optional/experimental extension, and include the experimental ones in `oak-unsafe`. But we won't have a feature for lookup data if we change that to an extension too.

rbehjati (commented):
Thanks for the reviews.

rbehjati:
> Nice.
> 
> I don't think it is worth exploring for this PR, but one thing that came up when looking at the code is that many of the extension have a generic interface: bytes in, bytes out. The request handler has the same interface. Perhaps in future we can unify the design so that it could be possible to bypass the Wasm logic (via a feature) and directly call the lookup (or TF infer) with the incoming request bytes and directly respond with the results.

What is the benefit of bypassing the Wasm logic? Do you mean for testing?

conradgrobler:
> What is the benefit of bypassing the Wasm logic? Do you mean for testing?

For example: if the lookup is just a straight lookup with no transformation, or the exact label from the client is sent to the metrics collection the Wasm does nothing but copy bytes from one place to another. If we could bypass Wasm for these cases it would reduce the TCB even further by not even having a Wasm engine compiled in.

oak_functions/loader/src/main.rs
```diff
@@ -160,7 +162,15 @@ async fn async_main(opt: Opt, config: Config, logger: Logger) -> anyhow::Result<
 
     let lookup_data = load_lookup_data(&config, logger.clone()).await?;
 
-    let tf_model = load_tensorflow_model(&config).await?;
+    let extensions = if cfg!(feature = "oak-unsafe") {
```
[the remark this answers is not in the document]
conradgrobler: Yes, the linting error is a bit annoying. One option is to just suppress that specific linting error in this one place. The other is to use feature gating with `any()` and `not(any())` to have to different declarations of the vector, but that will probably be a pain to maintain going forward.

conradgrobler (approved):

rbehjati:
> For example: if the lookup is just a straight lookup with no transformation, or the exact label from the client is sent to the metrics collection the Wasm does nothing but copy bytes from one place to another. If we could bypass Wasm for these cases it would reduce the TCB even further by not even having a Wasm engine compiled in.

Do you mean to have a version of Oak Functions with no untrusted code and no sandboxing?

tiziano88:
> > For example: if the lookup is just a straight lookup with no transformation, or the exact label from the client is sent to the metrics collection the Wasm does nothing but copy bytes from one place to another. If we could bypass Wasm for these cases it would reduce the TCB even further by not even having a Wasm engine compiled in.
> 
> Do you mean to have a version of Oak Functions with no untrusted code and no sandboxing?

To me, this way of splitting dependencies is not just a nice refactoring; it is an incremental shift towards distributed runtime, and more specifically, towards isolation between native modules; the next step may be to assume that (at least some) external APIs could in principle be on separate TEE instances, and if we do remote attestation plus enforcing policies (constant time + size) then we can actually allow communication to those native modules that would be running by themselves, be much more limited in their functionality (good for defense in depth), and scale better. Happy to keep chatting about it, if it isn't clear.

tiziano88 (approved):

oak_functions/loader/src/server.rs
```diff
@@ -149,25 +148,63 @@ impl FixedSizeBodyPadder for Response {
...
     response_bytes: Vec<u8>,
     lookup_data: Arc<LookupData>,
-    #[allow(dead_code)]
-    tf_model: Arc<Option<TensorFlowModel>>,
     instance: Option<wasmi::ModuleRef>,
     memory: Option<wasmi::MemoryRef>,
     logger: Logger,
     metrics_proxy: Option<PrivateMetricsProxy>,
+    /// A mapping of internal host functions to the corresponding [`OakApiNativeExtension`].
+    extensions_indices: Arc<HashMap<usize, Arc<BoxedExtension>>>,
```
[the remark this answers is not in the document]
rbehjati: We create a new instance of `WasmState` for each incoming request. `Arc` is needed, because `HashMap` is not clonable. Using `&HashMap` would require specifying lifetimes. I think `Arc` is cleaner and is more consistent with `lookup_data` which is `Arc`.

oak_functions/loader/src/server.rs
rbehjati: I think it would have been possible if extensions were clonable, but I am not sure how to do that without making extensions clonable.

`invoke` is called inside `WasmState::invoke_index`, which borrows `self` mutably. Then somewhere at the end of `invoke_index`, an extension is retrieved from the hashmap: `self.extensions_indices.get(&index)`. When calling `invoke` on the retrieved extension, `self` is still borrowed, so it cannot be borrowed mutably again. 

I think it makes sense, it should not be possible to mutate `self` (by writing into its memory), while it is being borrowed (for accessing one of its extensions).

rbehjati:
> To me, this way of splitting dependencies is not just a nice refactoring; it is an incremental shift towards distributed runtime, and more specifically, towards isolation between native modules; the next step may be to assume that (at least some) external APIs could in principle be on separate TEE instances, and if we do remote attestation plus enforcing policies (constant time + size) then we can actually allow communication to those native modules that would be running by themselves, be much more limited in their functionality (good for defense in depth), and scale better. Happy to keep chatting about it, if it isn't clear.

Sounds good. Thanks for the explanation.

oak_functions/loader/src/server.rs
rbehjati: The Extension is wrapped in `Arc`, so it is actually clonable. Fixed this in #2285.

PR #2281, opened 2021-09-07
Run tests for all features

rbehjati:


tiziano88 (approved):

oak_io/src/lib.rs
```diff
@@ -138,6 +138,7 @@ impl<Init: Decodable, Command: Decodable> Decodable for InitWrapper<Init, Comman
 }
 
 #[cfg(test)]
+#[cfg(not(feature = "linear-handles"))]
```
conradgrobler, line 141: Is there a test that is part of CI that will run without linear handles to ensure that these tests are also executed?

conradgrobler (approved):

oak_io/src/lib.rs
rbehjati: I doubt it. The error message when running with `linear-handle` is in the `drop` method. So I think all the existing tests are affected by it.

Issue #2273, opened 2021-08-26
Use raw bytes instead of Protobuf in Remote Attestation


ipetr0v:
Currently, in order to align with the SC version of remote attestation, we need to use raw serialized bytes instead of Protobuf messages.
The reason is that, we need to sign hashes outgoing messages, but Protobuf is non-deterministic. I.e. several serialized Protobuf messages may represent the same deserialized `struct`.

Crates that could be used: [bitfield](https://docs.rs/bitfield/0.13.2/bitfield) [nom](https://docs.rs/nom/7.0.0/nom/)

Related to https://github.com/project-oak/oak/issues/2261

PR #2282, opened 2021-09-07
Use raw bytes in Remote Attestation messages

ipetr0v:
This change:
- Makes Remote Attestation use raw bytes for messages
- Updates Rust and Java versions of Remote Attestation

Rust version is updated in https://github.com/project-oak/oak/pull/2287

Fixes https://github.com/project-oak/oak/issues/2273
Ref https://github.com/project-oak/oak/issues/2261


remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+
+  public static class ClientIdentity {
+    /** Message header. */
+    private final byte header;
+    /** Public key needed to establish a session key. */
+    public final byte[] ephemeralPublicKey;
+    /**
+     * Signature of the SHA-256 hash of all previously send and received messaged.
```
rbehjati, line 183:
```suggestion
     * Signature of the SHA-256 hash of all previously sent and received messages.
```

oak_functions/proto/server.proto
```diff
@@ -33,19 +38,17 @@ message Request {
...
   oneof request_type {
-    // Initialization message that starts the Remote Attestation protocol.
-    oak.remote_attestation.ClientHello client_hello = 1;
-    // Client part of session key negotiation.
-    oak.remote_attestation.ClientIdentity client_identity = 2;
+    // Remote Attestation handshake message.
+    AttestationMessage attestation_message = 1;
     // Request with payload encrypted using the session key.
-    Request request = 3;
+    Request request = 2;
```
tiziano88, line 44: Why is there still this message here? I was expecting that the service definition becomes effectively bytes in, bytes out; everything else is just about how those bytes are serialized and deserialized. 

Also I thought that the serialization is now done manually (I saw bincode being used), so why are all the types still defined as protobuf too?

remote_attestation/rust/src/message.rs
```diff
@@ -0,0 +1,251 @@
...
+pub struct ServerIdentity {
+    /// Message header.
+    header: u8,
+    /// Remote attestation protocol version.
+    pub version: u8,
+    /// Public key needed to establish a session key.
+    pub ephemeral_public_key: [u8; EPHEMERAL_PUBLIC_KEY_SIZE_BYTES],
+    /// Random vector sent in messages for preventing replay attacks.
+    pub random: [u8; REPLAY_PROTECTION_ARRAY_SIZE_BYTES],
+    /// Signature of the SHA-256 hash of all previously send and received messaged.
```
rbehjati, line 57:
```suggestion
    /// Signature of the SHA-256 hash of all previously sent and received messages.
```

remote_attestation/rust/src/message.rs
```diff
@@ -0,0 +1,251 @@
...
+    pub attestation_info: Vec<u8>,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct ClientIdentity {
+    /// Message header.
+    header: u8,
+    /// Public key needed to establish a session key.
+    pub ephemeral_public_key: [u8; EPHEMERAL_PUBLIC_KEY_SIZE_BYTES],
+    /// Signature of the SHA-256 hash of all previously send and received messaged.
```
rbehjati, line 87:
```suggestion
    /// Signature of the SHA-256 hash of all previously sent and received messages.
```

remote_attestation/rust/src/message.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// certificate. TEE report contains a hash of the `signing_public_key`.
+    ///
+    /// Attestation info must be a serialized `oak.remote_attestation.AttestationInfo` Protobuf
+    /// message.
+    pub attestation_info: Vec<u8>,
+}
+
+pub trait Serializable {
+    fn serialize(&self) -> anyhow::Result<Vec<u8>>;
+    fn deserialize(bytes: &[u8]) -> anyhow::Result<Self> where Self: Sized;
```
rbehjati, line 113: This sounds more like a method in a `Deserializable` trait. Can the trait be split into two?

remote_attestation/rust/src/message.rs
```diff
@@ -0,0 +1,251 @@
...
+            .context("Couldn't serialize client hello message")?;
+        if message.header == CLIENT_HELLO_HEADER {
+            Ok(message)
+        } else {
+            Err(anyhow!("Incorrect client hello message header"))
+        }
+    }
+}
+
+pub struct ServerIdentityBuilder<'a> {
```
rbehjati, line 145: What is the benefit of having a builder? Most of the fields in `ServerIdentity` are public. The same for `ClientIdentity`.

rbehjati (approved):
LGTM.

oak_functions/proto/server.proto
rbehjati: Do we want to have bytes everywhere? I thought we wanted to have bytes in the general remote attestation protocol, but still have types in the Oak Functions server.

ipetr0v: I added it because it's a `oneof` Protobuf field. And having `bytes .. = 1` and `bytes .. = 2` is impossible.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+  public static final int TRANSCRIPT_SIGNATURE_SIZE_BYTES = 64;
+  public static final int SIGNING_PUBLIC_KEY_SIZE_BYTES = 65;
+
+  public static class ClientHello {
+    /** Message header. */
+    private final byte header;
+    /** Random vector sent in messages for preventing replay attacks. */
+    public final byte[] random;
+
+    public ClientHello(byte[] random_) {
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+  public static final int SERIALIZATION_PREFIX_SIZE_BYTES = 8;
+
+  /** Size (in bytes) of a random array sent in messages to prevent replay attacks. */
+  public static final int REPLAY_PROTECTION_ARRAY_SIZE_BYTES = 32;
+  public static final int EPHEMERAL_PUBLIC_KEY_SIZE_BYTES = 32;
+  // TODO(#2277): Use OpenSSL signature format (which is 72 bytes).
+  public static final int TRANSCRIPT_SIGNATURE_SIZE_BYTES = 64;
+  public static final int SIGNING_PUBLIC_KEY_SIZE_BYTES = 65;
+
+  public static class ClientHello {
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+        throw new IllegalArgumentException("Incorrect client hello message header");
+      }
+
+      byte[] random = readFixedSizeArray(inputStream, REPLAY_PROTECTION_ARRAY_SIZE_BYTES, "random value");
+
+      return new Message.ClientHello(random);
+    }
+  }
+
+  public static class ServerIdentity {
```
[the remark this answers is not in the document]
ipetr0v: Done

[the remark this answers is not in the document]
ipetr0v: It's not a top level class. It's an inner class `Message.ServerIdentity`.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: Done

remote_attestation/rust/src/message.rs
ipetr0v: Done

remote_attestation/rust/src/message.rs
ipetr0v: Done

remote_attestation/rust/src/message.rs
ipetr0v: Done

remote_attestation/rust/src/message.rs
ipetr0v: Because these fields are fixed size arrays `[u8; L]`. And creating such parameters will require and additional code each time.
Also, having a new method with `Vec<u8>`'s will be very hard to initialize without errors, because there will be no type checks.
So builder with named fields is just a readability addition.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+      return new ClientIdentity(ephemeralPublicKey, transcriptSignature, signingPublicKey, attestationInfo);
+    }
+  }
+
+  private static byte[] readFixedSizeArray(DataInputStream inputStream, int size, String name) throws IllegalArgumentException {
+    byte[] input = new byte[size];
+    try {
+      inputStream.readFully(input, 0, size);
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Couldn't read " + name + " from input stream");
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,294 @@
...
+  public static final int EPHEMERAL_PUBLIC_KEY_SIZE_BYTES = 32;
+  // TODO(#2277): Use OpenSSL signature format (which is 72 bytes).
+  public static final int TRANSCRIPT_SIGNATURE_SIZE_BYTES = 64;
+  public static final int SIGNING_PUBLIC_KEY_SIZE_BYTES = 65;
+
+  public static class ClientHello {
+    /** Message header. */
+    private final byte header;
+    /** Random vector sent in messages for preventing replay attacks. */
+    public final byte[] random;
```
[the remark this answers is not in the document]
ipetr0v: Done

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -123,21 +123,25 @@ public void onCompleted() {}
...
     // Remotely attest the server and create:
     // - Client attestation identity containing client's ephemeral public key
     // - Encryptor used for decrypting/encrypting messages between client and server
-    ClientIdentity clientIdentity = attestationEngine.processServerIdentity(serverIdentity);
+    byte[] clientIdentity = attestationEngine.processServerIdentity(serverIdentity);
     AttestedInvokeRequest clientIdentityRequest =
-        AttestedInvokeRequest.newBuilder().setClientIdentity(clientIdentity).build();
+        AttestedInvokeRequest.newBuilder().setAttestationMessage(
+            AttestationMessage.newBuilder().setBody(ByteString.copyFrom(clientIdentity)).build()
+        ).build();
```
tiziano88, line 144: possibly for the future, but I don't even think this part of the logic should know about the various messages; it should ask the attestationEngine if there is any data to send, if so, send it (as bytes), and when the server replies, send that data to the attestation engine (as bytes). This class should not know anything about any inner message, they should only exist in the attestationEngine. 

The only output of attestationEngine is whether it expects more data from the other side (in which case it keeps reading it, in a loop), or whether it has succesfully derived the encryptor.

If we change the protocol to require an additional round trip, ideally this class should not change at all, right? If you think it should change instead, could you clarify why?

oak_functions/proto/server.proto
tiziano88: I'm curious why is it impossible? I thought they are just separate fields.

remote_attestation/java/src/com/google/oak/remote_attestation/ClientAttestationEngine.java
```diff
@@ -30,9 +32,6 @@
 import java.util.Random;
 import oak.remote_attestation.AttestationInfo;
 import oak.remote_attestation.AttestationReport;
-import oak.remote_attestation.ClientHello;
-import oak.remote_attestation.ClientIdentity;
-import oak.remote_attestation.ServerIdentity;
 
 enum AttestationState {
```
tiziano88, line 36: I think the state would be better expressed as just the expected id of the next frame.

remote_attestation/java/src/com/google/oak/remote_attestation/ClientAttestationEngine.java
```diff
@@ -104,48 +104,49 @@ public ClientHello createClientHello() throws IllegalStateException {
...
 
     // Create client hello message.
-    ClientHello clientHello =
-        ClientHello.newBuilder().setRandom(ByteString.copyFrom(random)).build();
+    Message.ClientHello clientHello = new Message.ClientHello(random);
 
     // Update current transcript.
-    byte[] serializedClientHello = clientHello.toByteArray();
+    byte[] serializedClientHello = clientHello.serialize();
     transcript = Bytes.concat(transcript, serializedClientHello);
```
tiziano88, line 111: May be useful to have a method to append to the transcript, since it's quite common.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,356 @@
...
+  public static final int SIGNING_PUBLIC_KEY_SIZE_BYTES = 65;
+
+  /** Initial message that starts Remote Attestation handshake. */
+  public static class ClientHello {
+    /** Message header. */
+    private final byte header;
+    /** Random vector sent in messages for preventing replay attacks. */
+    private final byte[] random;
+
+    public ClientHello(byte[] random) {
```
tiziano88, line 58: Doesn't this need to be of a fixed size? It seems here nothing is checking the size.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,356 @@
...
+import java.io.DataOutput;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import static java.lang.Math.toIntExact;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/** Wrapper class for Remote Attestation Protocol messages. */
+public class Message {
+  static final int CLIENT_HELLO_HEADER = 1;
```
tiziano88, line 32: By header do you mean frame type? If so, why not use the same terminology as SC?

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,356 @@
...
+  static final int SERVER_IDENTITY_HEADER = 2;
+  static final int CLIENT_IDENTITY_HEADER = 3;
+
+  /** Remote Attestation protocol version. */
+  public static final int PROTOCOL_VERSION = 1;
+  /**
+   * Size (in bytes) of the prefix that describes the size of a serialized array.
+   * Prefix is encoded with Little-Endian.
+   */
+  public static final int SERIALIZATION_PREFIX_SIZE_BYTES = 8;
```
tiziano88, line 42: I'm not sure the name of this constant reflects what it actually does; serialization prefix of what?

remote_attestation/rust/src/message.rs
tiziano88: Why is additional code required for the arrays? As long as they are correctly typed from the beginning, the types should line up. I think the current code is turning a lot of checks that could be done at compile time into runtime checks.

remote_attestation/rust/src/message.rs
```diff
@@ -0,0 +1,265 @@
...
+impl Serializable for ClientIdentity {
+    fn serialize(&self) -> anyhow::Result<Vec<u8>> {
+        bincode::serialize(&self).context("Couldn't serialize client identity message")
+    }
+}
+
+impl Deserializable for ClientIdentity {
+    fn deserialize(bytes: &[u8]) -> anyhow::Result<Self> {
+        let message: Self = bincode::deserialize(&bytes)
+            .context("Couldn't serialize client identity message")?;
```
tiziano88, line 242:
```suggestion
            .context("Couldn't deserialize client identity message")?;
```

tiziano88 (approved):

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: Header is the terminology used in the SC document.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: It checks the size once it tries to serialize.
I deleted the check in the conscructor, so that it doesn't throw exceptions.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
tiziano88: You are right, even in the implementation they use both frame header and frame type. Maybe we can all pick one at some point.

oak_functions/proto/server.proto
ipetr0v: IIUC it checks the type of a `oneof` variant, not it's name.

ipetr0v: It could be using bytes, but when there will be no `oneof`. Just `bytes` that are passed to the attestation engine.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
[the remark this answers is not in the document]
conradgrobler: Yes, you are right. I missed the indentation.

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -0,0 +1,356 @@
...
+     * certificate. TEE report contains a hash of the `signingPublicKey`.
+     *
+     * Attestation info must be a serialized `oak.remote_attestation.AttestationInfo` Protobuf
+     * message.
+     */
+    private final byte[] attestationInfo;
+
+    public ClientIdentity(
+      byte[] ephemeralPublicKey,
+      byte[] transcriptSignature,
```
conradgrobler, line 253: Will this ever be set using the constructor to anything other than an empty byte array? If not, it is probably better to remove it from the constructor signature. Same appolies to the ServerIdentity.

conradgrobler (approved):

remote_attestation/rust/src/message.rs
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: Renamed to `ARRAY_SERIALIZATION_PREFIX_LENGTH`

Issue #2280, opened 2021-09-06
Enable `timings` flag for Rust compilation


tiziano88:
See 
- https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#timings
- https://matklad.github.io/2021/09/04/fast-rust-builds.html

Perhaps as an option to `runner`.

PR #2283, opened 2021-09-08
Added timings flags in runner

shikharvashistha:
# Checklist

      construction.

Greetings of the day @tiziano88,

Please review the pr and let me know in case of any changes if required.

Thanks & Regards
Shikhar Vashistha

Fixes : #2280
References : [here](https://matklad.github.io/2021/09/04/fast-rust-builds.html) && [here](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#timings)

### DIFF
--- a/runner/src/examples.rs
+++ b/runner/src/examples.rs
@@ -431,7 +431,7 @@ pub fn build_server(opt: &BuildServer, additional_features: Vec<String>) -> Step
                         "CARGO_INCREMENTAL".to_string() => "0".to_string(),
                         "RUSTDOCFLAGS".to_string() => "-Cpanic=abort".to_string(),
                         // grcov instructions suggest also including `-Cpanic=abort` in RUSTFLAGS, but this causes our build.rs scripts to fail.
-                        "RUSTFLAGS".to_string() => "-Z timings -Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic-abort_tests".to_string(),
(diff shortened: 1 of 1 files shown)


tiziano88 (approved):
Thanks!

shikharvashistha:
> Thanks!

Greetings of the day @tiziano88,

I guess the `timings` flag is only available for nightly channel currently !

Thanks & Regards
Shikhar Vashistha

runner/src/examples.rs
```diff
@@ -655,7 +655,7 @@ pub fn build_wasm_module(name: &str, target: &Target, example_name: &str) -> Ste
                         // `--out-dir` is unstable and requires `-Zunstable-options`.
                         "-Zunstable-options".to_string(),
                         "build".to_string(),
-                        "--target=wasm32-unknown-unknown".to_string(),
+                        "-Ztimings --target=wasm32-unknown-unknown".to_string(),
```
tiziano88, line 658: Apparently this ends up being squashed with the --target flag; it needs to be defined separately as its own string, it's not enough to have a space in this string.

shikharvashistha: Greetings of the day @tiziano88,

I've tested the -Z timings locally [here](https://github.com/shikharvashistha/blockchained/blob/807c6078f408e8fc3ab54fdd33e0b2ffdb4b44ac/.github/workflows/rust.yml#L50) && [here](https://github.com/shikharvashistha/blockchained/runs/3573330564?check_suite_focus=true) and came to a conclusion that it is currently only available on nightly channel as stated [here](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#timings) too. So in my opinion if this project uses nightly channel for CI then only it would be beneficial to introduce the -Z timings flag. What do you suggest ?

Thanks & Regards
@shikharvashistha

tiziano88:
Thanks, we ended up using this approach to debug some slow build issues, but it's probably not needed in Runner, so I'll close this for now.

PR #2284, opened 2021-09-08
Allow caller to specify channel for attestation in Java attestation client

conradgrobler:
# Checklist

      construction.

### DIFF
--- a/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
+++ b/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
@@ -45,8 +45,7 @@ import oak.remote_attestation.EncryptedData;
 import oak.remote_attestation.ServerIdentity;
 
 /**
- * Client with remote attestation support for sending requests to an Oak
- * Functions loader application.
+ * Client with remote attestation support for sending requests to an Oak Functions application.
  */
 public class AttestationClient {
   private static final Logger logger = Logger.getLogger(AttestationClient.class.getName());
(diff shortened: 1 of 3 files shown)


oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -45,24 +45,28 @@
...
+  public AttestationClient() {}
+
+  /**
+   * Creates a gRPC channel and creates an attested channel over it.
    *
-   * `url` must contain a protocol used for connection ("https://" or "http://").
+   * @param url must contain a protocol used for the connection ("https://" or "http://")
    */
-  public AttestationClient(String url)
+  public void attest(String url)
```
tiziano88, line 69: (FYI) Once we figure out more of how remote attestation will actually work, I think we will want to change this API; we should probably have an AttestationClientFactory or similar, configured with a policy (not just a hardcoded TEE measurement), and with a method to then create an individual attested clients. In that design, the AttestedClient fields would all be final, and the policy verification logic would only live in the factory, not in the established client.

cc @ipetr0v

tiziano88 (approved):

rbehjati (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
conradgrobler: We should also consider a redesign of the lifecycle, e.g. not overriding finalize. See https://google.github.io/styleguide/javaguide.html#s6.4-finalizers

PR #2285, opened 2021-09-09
Make Extension::invoke API better resemble existing APIs

rbehjati:
Ref https://github.com/project-oak/oak/pull/2278#discussion_r703384995


tiziano88 (approved):
Much cleaner now IMO, thanks!

PR #2288, opened 2021-09-10
First version of Oak Functions readme

rbehjati:


tiziano88 (approved):
Thanks for starting this! Personally I would have made the documentation more stand-alone so that someone with no knowledge of Oak or Oak Functions could just read it and understand it, while I feel the current readme mostly highlights differences from the original Oak, so it assumes that the reader already knows everything about Oak, which is quite unrealistic. Maybe consider switching the point of view in the future?

conradgrobler (approved):
Nice!

rbehjati:
> Thanks for starting this! Personally I would have made the documentation more stand-alone so that someone with no knowledge of Oak or Oak Functions could just read it and understand it, while I feel the current readme mostly highlights differences from the original Oak, so it assumes that the reader already knows everything about Oak, which is quite unrealistic. Maybe consider switching the point of view in the future?

Thanks. Done. There is still a lot more that can be explained, and I think it would be nice to add some code snippets too, for instance for the configuration file. But these can be done in future iterations.

PR #2291, opened 2021-09-16
Allow fetching lookup data from files

### LINKED ISSUE #2219: Crates are rebuilt when they have a build.rs file that emits `cargo:rerun-if` statements
To reproduce (in Docker):

- `cargo build --manifest-path=./oak_functions/abi/Cargo.toml`
  + run the command twice: note that the second time it still compiles the crate
- remove or comment out this line:
  https://github.com/project-oak/oak/blob/1e73c4dc6e63ad6edc6ce3a9fa6197a6f945ac1c/oak_functions/abi/build.rs#L31
- re-run the command twice: this time, the second time the crate is not recompiled, and the command finishes instantly

I don't understand why though, AFAICT those lines are only emitted for the proto files, and those are definitely not changing, so I'm not sure what's triggering the change detection.

### LINKED ISSUE #2253: Oak Functions loader should not require lookup configuration to be set
_Originally posted by @conradgrobler in https://github.com/project-oak/oak/pull/2245#discussion_r682446348_

tiziano88:
Also allow empty lookup data source.

Fix build script to specify the correct paths for
`cargo:rerun-if-changed`. Previously the paths were invalid, which meant
that cargo thought that those files had been deleted and therefore was
always rerunning all the build scripts every time.

Fix #2253
Fix #2219

# Checklist

      construction.

### DIFF
--- a/oak_functions/examples/fuzzable/module/build.rs
+++ b/oak_functions/examples/fuzzable/module/build.rs
@@ -24,6 +24,6 @@ fn main() {
     // https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath
     for proto_path in file_paths.iter() {
         let file_path = std::path::Path::new(proto_path);
-        println!("cargo:rerun-if-changed={}", file_path.display());
+        println!("cargo:rerun-if-changed=../../../../{}", file_path.display());
     }
 }
--- a/oak_functions/loader/src/lookup.rs
+++ b/oak_functions/loader/src/lookup.rs
@@ -35,14 +35,19 @@ impl LookupDataAuth {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum LookupDataSource {
+    Http { url: String, auth: LookupDataAuth },
+    File(std::path::PathBuf),
+}
+
 /// An in-memory lookup store instance that can refresh its internal entries from the provided data
 /// file URL.
 ///
 /// Entries in the data file path must be consecutive binary encoded and length delimited
 /// protobuf messages according to the definition in `/oak_functions/proto/lookup_data.proto`.
 pub struct LookupData {
-    lookup_data_url: String,
-    lookup_data_auth: LookupDataAuth,
+    lookup_data_source: Option<LookupDataSource>,
     entries: RwLock<HashMap<Vec<u8>, Vec<u8>>>,
     logger: Logger,
 }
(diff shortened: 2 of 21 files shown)


conradgrobler (approved):
Nice

oak_functions/loader/src/lookup.rs
```diff
@@ -79,59 +79,81 @@ impl LookupData {
...
-        let start = Instant::now();
-        {
-            *self
-                .entries
-                .write()
-                .expect("could not lock entries for write") = entries;
+        match &self.lookup_data_source {
+            Some(lookup_data_source) => {
+                let start = Instant::now();
+                let lookup_data_buf: Bytes = match lookup_data_source {
```
rbehjati, line 85: This method is doing too many things. I'd suggest moving the block for fetching `lookup_data_buf` into its own method, if possible.

oak_functions/loader/src/main.rs
```diff
@@ -238,12 +238,32 @@ async fn async_main(opt: Opt, config: Config, logger: Logger) -> anyhow::Result<
 }
 
 async fn load_lookup_data(config: &Config, logger: Logger) -> anyhow::Result<Arc<LookupData>> {
+    let lookup_data_source = if config.lookup_data_url.is_empty() {
```
rbehjati, line 241: Can `lookup_data_download_period` be removed from the examples that don't need it. I think previously we had to include them even for the examples that did not use lookup data, like mobilenet and perhaps also metrics.

oak_functions/examples/fuzzable/module/build.rs
```diff
@@ -24,6 +24,6 @@ fn main() {
     // https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath
     for proto_path in file_paths.iter() {
         let file_path = std::path::Path::new(proto_path);
-        println!("cargo:rerun-if-changed={}", file_path.display());
+        println!("cargo:rerun-if-changed=../../../../{}", file_path.display());
```
rbehjati, line 27: There is a similar line in `oak_functions/loader/fuzz/build.rs`. Do you mind updating that too? Also, does https://github.com/project-oak/oak/blob/93bfc0b48d3459387254eca3c405d32527f186f6/oak_utils/src/lib.rs#L321 need to be updated to include `repo_root` in `"cargo:rerun-if-changed={}"`?

oak_functions/loader/src/lookup.rs
```diff
@@ -35,14 +35,19 @@ impl LookupDataAuth {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum LookupDataSource {
```
rbehjati, line 39: I think the doc comment for `lookup_data_url` in https://github.com/project-oak/oak/blob/24fef4d14e1d0153c90d9ca3d7fc52b010d41327/oak_functions/loader/src/main.rs#L55 should be updated.

rbehjati (approved):
LGTM. Thanks.

oak_functions/loader/src/lookup.rs
tiziano88: done.

oak_functions/loader/src/lookup.rs
tiziano88: Done.

tiziano88 (commented):
Thanks

PR #2293, opened 2021-09-18
Refactor attestation messages

ipetr0v:


oak_functions/client/rust/src/attestation.rs
```diff
@@ -101,106 +99,77 @@ impl AttestationClient {
...
-            let response_type = response
-                .response_type
-                .context("Couldn't read response type")?;
-            let encrypted_payload = if let ResponseType::EncryptedPayload(data) = response_type {
-                Ok(data)
-            } else {
-                Err(anyhow!("Received incorrect message type"))
-            }?;
-            let payload = self
+        let response = if let Some(encrypted_response) = encrypted_response {
```
conradgrobler, line 163: This is a bit of an unusual construct. Is there any reason for not using Option::map or a match statement here?

remote_attestation/rust/src/handshaker.rs
```diff
@@ -21,203 +21,153 @@
...
-        serialized_client_hello: &[u8],
-    ) -> anyhow::Result<(Vec<u8>, ServerAttestationEngine<Attesting>)> {
-        let client_hello = ClientHello::deserialize(serialized_client_hello)
-            .context("Couldn't deserialize client hello message")?;
-
-        // Create server identity message.
-        let key_negotiator = KeyNegotiator::create(KeyNegotiatorType::Server)
+    /// Transitions [`ClientHandshaker`] state to [`HandshakerState::ExpectingServerIdentity`]
+    /// state.
+    pub fn create(behavior: AttestationBehavior) -> anyhow::Result<(Self, Vec<u8>)> {
```
conradgrobler, line 133: I know this follows a preexisting pattern, but returning a tuple of the handshaker and the client hello message at the same time does not feel like a very natural API. I would have expected a separate functions for initialising the handshaker and getting the client hello message, or perhaps return the client hello message when calling next_step the first time.

remote_attestation/rust/src/message.rs
```diff
@@ -235,3 +270,62 @@ impl Deserializable for ClientIdentity {
...
+            header: ENCRYPTED_DATA_HEADER,
+            nonce: *nonce,
+            data: data.to_vec(),
+        }
+    }
+}
+
+impl Serializable for EncryptedData {
+    fn serialize(&self) -> anyhow::Result<Vec<u8>> {
+        bincode::serialize(&self).context("Couldn't serialize encrypted data message")
```
conradgrobler, line 286: I am a bit concerned about the blackbox nature of bincode. How does it handle unexpected data such as incorrect message lengths? Does it truncate/pad the resulting arrays, or return an error? Will this behaviour stay the same in future? I think it is worth considering manual serialisation and deserialisation similar to the java implementation.

remote_attestation/rust/src/handshaker.rs
```diff
@@ -21,203 +21,153 @@
...
+            Self::ExpectingClientHello => write!(f, "ExpectingClientHello"),
+            Self::ExpectingServerIdentity(_) => write!(f, "ExpectingServerIdentity"),
+            Self::ExpectingClientIdentity(_) => write!(f, "ExpectingClientIdentity"),
+            Self::Completed(_) => write!(f, "Completed"),
+        }
+    }
+}
+
+pub trait Handshaker {
+    fn next_step(&mut self, message: &[u8]) -> anyhow::Result<Option<Vec<u8>>>;
```
conradgrobler, line 67: I am not sure whether there is enough commonality for this extracting this as a shared trait to be very useful. It feels like it is obfuscating details of the process.

conradgrobler (approved):

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
```diff
@@ -327,6 +329,58 @@ public static ClientIdentity deserialize(byte[] input)
     }
   }
 
+  /** Message containing data encrypted using a session key. */
+  public static class EncryptedData {
+    /** Message header. */
+    private final byte header;
```
tiziano88, line 335: Isn't the point of this header that it is part of the high-level framing? If so, I would expect to see a class that represents that level of abstraction, basically a header byte, and then payload as bytes, and in this case the payload is the nonce + data.

remote_attestation/rust/src/message.rs
tiziano88: +1

remote_attestation/rust/src/message.rs
```diff
@@ -235,3 +270,62 @@ impl Deserializable for ClientIdentity {
...
+    }
+}
+
+/// Deserializes an attestation message from a serialized [`input`] and wraps in a
+/// [`MessageWrapper`].
+pub fn deserialize(input: &[u8]) -> anyhow::Result<MessageWrapper> {
+    if input.is_empty() {
+        return Err(anyhow!("Empty input"));
+    }
+    match input[0] {
```
tiziano88, line 308: This validates my point that there is really an outer structure for the frame and this should match on that outer structure.

tiziano88 (approved):

remote_attestation/rust/src/handshaker.rs
```diff
@@ -21,203 +21,153 @@
...
-    S: AttestationState,
-{
-    /// Behavior of the remote attestation protocol.
-    behavior: AttestationBehavior,
-    /// Current state of the remote attestation protocol.
-    state: S,
-    /// Collection of previously sent and received messages.
-    /// Signed transcript is sent in messages to prevent replay attacks.
-    transcript: Transcript,
+enum HandshakerState {
```
rbehjati, line 39: Perhaps split this into client and server handshaker states. In that case, the client handshaker would better resemble the Java version.

rbehjati (approved):
LGTM.

remote_attestation/rust/src/handshaker.rs
ipetr0v: Done

oak_functions/client/rust/src/attestation.rs
```diff
@@ -101,106 +99,77 @@ impl AttestationClient {
...
-        let (client_hello, attestation_engine) = attestation_engine
-            .create_client_hello()
-            .context("Couldn't create client hello")?;
-        let request = AttestedInvokeRequest {
-            request_type: Some(RequestType::AttestationMessage(AttestationMessage {
-                body: client_hello,
-            })),
-        };
+    ) -> anyhow::Result<Encryptor> {
+        let (mut handshaker, outgoing_message) = ClientHandshaker::create(
```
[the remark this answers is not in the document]
ipetr0v: Done

remote_attestation/rust/src/handshaker.rs
ipetr0v: Done

oak_functions/loader/src/attestation.rs
```diff
@@ -186,64 +181,8 @@ where
...
-
-    let attestation_engine = attestation_engine
-        .process_client_identity(&client_identity)
-        .context("Couldn't attest the client")?;
-    let encryptor = attestation_engine.get_encryptor();
-    Ok(encryptor)
+fn error_to_status<T: std::fmt::Debug>(comment: &str, error: T) -> Status {
+    let message = format!("{}: {:?}", comment, error);
+    warn!("{}", message);
+    Status::internal(message)
```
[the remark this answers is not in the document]
ipetr0v: The higher level function is a `async_stream::try_stream!`, and it's not possible to use `?` there.
So this function is just a code reduction: I use it in `map_err` instead of copy-pasting this code in every `map_err` statement.

oak_functions/client/rust/src/attestation.rs
```diff
@@ -101,106 +99,76 @@ impl AttestationClient {
...
-            None
+        let response = match encrypted_response {
+            Some(encrypted_response) => {
+                let response = self
+                    .encryptor
+                    .decrypt(&encrypted_response.body)
+                    .context("Couldn't decrypt response")?;
+                Some(response)
+            }
+            None => None,
```
tiziano88, line 169: This looks like it should just be a `map`

oak_functions/loader/src/attestation.rs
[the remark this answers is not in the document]
tiziano88: Can you fix the logging though?

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
tiziano88: That's not a usable definition; look at your own code, it is basically already parsing the first byte and then based on that deciding how to parse the rest, so there is definitely a higher level framing already implemented, and I think that structure should be reflected more clearly in the code and documentation (e.g. a Frame struct). As another point, it does not make sense if you have a ClientHello frame to get its header; of course its header is the header for the ClientHello frame. (as usual, it's fine to leave a TODO, it does not need to be implemented now)

remote_attestation/rust/src/handshaker.rs
```diff
@@ -21,203 +21,182 @@
...
+    Initializing,
+    ExpectingServerIdentity(KeyNegotiator),
+    Completed(AeadEncryptor),
+    /// Additional state that represents ongoing message processing.
+    MessageProcessing,
+}
+
+impl Default for ClientHandshakerState {
+    fn default() -> Self {
+        Self::MessageProcessing
```
tiziano88, line 50: Why would the default be this value? Also what does "processing" mean in this context?

remote_attestation/rust/src/handshaker.rs
tiziano88: I also still don't see the value of the additional traits. Encapsulation is done by having pub or private methods, you don't need a trait for that, unless you expect more than one implementation. I guess in this case the two implementation are client and server handshakers, but somehow that still seems backwards to me. Anyway, we can do another pass in the future.

remote_attestation/rust/src/message.rs
```diff
@@ -26,16 +27,36 @@ use serde_big_array::BigArray;
...
 
-/// Remote Attestation protocol version.
+/// Remote attestation protocol version.
 pub const PROTOCOL_VERSION: u8 = 1;
 
 /// Length (in bytes) of the random vector sent in messages for preventing replay attacks.
 pub const REPLAY_PROTECTION_ARRAY_LENGTH: usize = 32;
 
+/// Convenience struct that wraps attestation messages.
+pub enum MessageWrapper {
```
tiziano88, line 39: Isn't this MessageWrapper exactly what I mentioned in the other comment as the high level frame? In fact, I think the parsing and serialization should be done via this type, not the individual message types.

tiziano88 (approved):

oak_functions/client/rust/src/attestation.rs
ipetr0v: `map` doesn't allow to use `?` inside it

remote_attestation/rust/src/handshaker.rs
ipetr0v: I created this state in order to use `match std::mem::take(&mut self.state)`, because next states are created based on previous ones (and they take their internals, i.e. `key_negotiator`).

remote_attestation/java/src/com/google/oak/remote_attestation/Message.java
ipetr0v: Created an issue: https://github.com/project-oak/oak/issues/2295

remote_attestation/rust/src/message.rs
```diff
@@ -235,3 +270,62 @@ impl Deserializable for ClientIdentity {
         }
     }
 }
+
+impl EncryptedData {
```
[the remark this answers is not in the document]
ipetr0v: I did that on purpose, so it's easier to look at all of the protocol messages at once.

oak_functions/loader/src/attestation.rs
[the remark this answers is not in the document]
ipetr0v: Fixed logging

remote_attestation/rust/src/handshaker.rs
ipetr0v: Removed the trait

PR #2299, opened 2021-09-23
Add message tests to remote attestation

ipetr0v:


remote_attestation/rust/src/message.rs
```diff
@@ -23,6 +23,9 @@ use bincode;
 use serde::{Deserialize, Serialize};
 use serde_big_array::BigArray;
 
+/// Maximum message size is set to be equal to 1Kb.
```
tiziano88, line 26: (nit) 1024 is 1KiB, not 1Kb

Ki = 1024
k = 1000

also

b = bit
B = byte

https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+    [Default::default(); L]
+}
+
+/// Converts slices to arrays (expands with zeroes).
+fn to_array<T, const L: usize>(input: &[T]) -> anyhow::Result<[T; L]>
+where
+    T: std::marker::Copy + std::default::Default,
+{
+    if input.len() <= L {
+        let mut result: [T; L] = default_array();
```
tiziano88, line 48: Why is it not sufficient to do `let mut result: [T; L] = Default::default()` here? Is it not implemented for arrays?

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            return TestResult::discard();
+        }
+        let mut server_identity = ServerIdentity::new(
+            to_array(&ephemeral_public_key).unwrap(),
+            to_array(&random).unwrap(),
+            to_array(&signing_public_key).unwrap(),
+            attestation_info,
+        );
+        assert_eq!(server_identity.transcript_signature, default_array(),);
```
tiziano88, line 105:
```suggestion
        assert_eq!(server_identity.transcript_signature, default_array());
```

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+        assert_eq!(
+            server_identity.transcript_signature,
+            transcript_signature_array
+        );
+
+        let result = test_serialize_template(&server_identity);
+        assert!(result.is_ok());
+
+        server_identity.clear_transcript_signature();
+        assert_eq!(server_identity.transcript_signature, default_array(),);
```
tiziano88, line 118:
```suggestion
        assert_eq!(server_identity.transcript_signature, default_array());
```

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn property(random: Vec<u8>) -> TestResult {
+        if random.len() > REPLAY_PROTECTION_ARRAY_LENGTH {
+            return TestResult::discard();
+        }
+        let client_hello = ClientHello::new(to_array(&random).unwrap());
+        let result = test_serialize_template(&client_hello);
+        assert!(result.is_ok());
+        TestResult::from_bool(result.unwrap())
+    }
+    quickcheck(property as fn(Vec<u8>) -> TestResult);
```
tiziano88, line 80: why is the `as` necessary? That seems exactly the same signature

remote_attestation/rust/src/tests/message.rs
ipetr0v: It is only implemented for specific sizes, i.e. 1,2,...,32

tiziano88: That seems worth documenting, right?

remote_attestation/rust/src/handshaker.rs
```diff
@@ -51,7 +51,7 @@ impl Default for ClientHandshakerState {
     }
 }
 
-impl std::fmt::Display for ClientHandshakerState {
+impl std::fmt::Debug for ClientHandshakerState {
```
rbehjati, line 54: Now that this is changed to `Debug`, why not use derive?

remote_attestation/rust/src/tests/message.rs
rbehjati, line 80: Can you add a comment explaining what `quickcheck` does. It is not evident from the name that it tests the function with randomly generated input. 
How long do these tests take to run?

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+#[test]
+fn test_serialize_server_identity() {
+    fn property(
+        ephemeral_public_key: Vec<u8>,
```
rbehjati, line 86: Does the test actually generate random public keys?

remote_attestation/rust/src/tests/message.rs
rbehjati, line 105: What does this actually check? That `transcript_signature` is an empty array of the right size?

remote_attestation/rust/src/message.rs
ipetr0v: Done

remote_attestation/rust/src/tests/message.rs
ipetr0v: Done

remote_attestation/rust/src/tests/message.rs
ipetr0v: Done

remote_attestation/rust/src/tests/message.rs
rbehjati, line 105: What does quickcheck do when these assertions fail?

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+    assert_matches!(deserialized_client_hello, Ok(_));
+    assert_eq!(
+        deserialized_client_hello.unwrap(),
+        MessageWrapper::ClientHello(client_hello)
+    );
+
+    let server_identity =
+        ServerIdentity::new(default_array(), default_array(), default_array(), vec![]);
+    let deserialized_server_identity = deserialize_message(&server_identity.serialize().unwrap());
+    assert_matches!(deserialized_server_identity, Ok(_));
```
rbehjati, line 193: Why not use `is_ok()` here? And everywhere else `assert_matches` is used.

remote_attestation/rust/src/tests/message.rs
ipetr0v: Because `quickcheck` requires `Testable` to be implemented.
And it's implemented as follows:
https://github.com/BurntSushi/quickcheck/blob/defde6fb0ce20b0c8c4e672aa9ae821f7d1f5b38/src/tester.rs#L386-L394

rbehjati (commented):
I am not sure that it is a good idea to have this level of randomness in the unit tests. I think it is better to have simpler tests, with explicit hard-coded values here, but add the logic here as a fuzz target.

remote_attestation/rust/src/tests/message.rs
ipetr0v: Added a comment

remote_attestation/rust/src/tests/message.rs
ipetr0v: Done

remote_attestation/rust/src/tests/message.rs
```diff
@@ -0,0 +1,222 @@
...
+            L,
+            input.len()
+        ))
+    }
+}
+
+fn test_serialize_template<M: Serializable + Deserializable + PartialEq>(
+    message: &M,
+) -> anyhow::Result<bool> {
+    let serialized_message = message.serialize().context("Couldn't serialize message")?;
```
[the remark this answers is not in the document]
ipetr0v: I try to put assertions only to top-level functions marked as `#[test]`, so that it's easier to read errors in the test output (i.e. is nested functions are involved, `anyhow` will show them).

`TestResult` is necessary to allow discarding test values that don't sutisfy some preconditions.

remote_attestation/rust/src/tests/message.rs
ipetr0v: `is_ok()` doesn't print the error message if the test fails.

remote_attestation/rust/src/tests/message.rs
ipetr0v: It shows the test as failed, and tries to minimize the example input that made this test fail.

ipetr0v: It tests the `clear_transcript_signature` function.

remote_attestation/rust/src/tests/message.rs
ipetr0v: Yes, `quickcheck` generates random inputs 100 times (with default settings).

remote_attestation/rust/src/handshaker.rs
ipetr0v: Because `assert_matches` requires `Debug` to be implemented.

remote_attestation/rust/src/tests/message.rs
ipetr0v: Added a comment.
When I'm running them on my machine - they are instantenious.

ipetr0v:
> I am not sure that it is a good idea to have this level of randomness in the unit tests. I think it is better to have simpler tests, with explicit hard-coded values here, but add the logic here as a fuzz target.

These are Property Tests - a different kind of tests compared to Fuzzing.
IIUC Fuzzing works on the program as a whole, i.e. it passes random values as external message.
Property Tests are unit tests that use some very limited fuzzing.

remote_attestation/rust/src/handshaker.rs
rbehjati: Is not it possible to use `#[derive(Debug)]` instead of implementing `Debug` manually?

remote_attestation/rust/src/tests/message.rs
rbehjati: Might be worth adding a comment, so that we don't change it later.

rbehjati (approved):

remote_attestation/rust/src/handshaker.rs
ipetr0v: No, because some of the states contain `KeyNegotiator` and `Encryptor`, and printing them is prohibited.

tiziano88: Can you please add a comment explaining this?

PR #2302, opened 2021-09-24
Add Java tests to Remote Attestation

ipetr0v:


remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -91,88 +75,104 @@ public ClientHandshaker(byte[] expectedTeeMeasurement) {
   /**
    * Initializes the remote attestation handshake by creating an {@code ClientHello} message.
    *
-   * Transitions {@code ClientHandshaker} state from {@code Initializing} to
-   * {@code Attesting} state.
+   * Transitions {@code ClientHandshaker} state from {@code State.Initializing} to
+   * {@code State.ExpectingServerIdentity} state.
    */
   public byte[] createClientHello() throws IllegalStateException, IOException {
```
rbehjati, line 81: Pre-existing: IllegalStateException is a RuntimeException. Does it have to be specified in the signature? 
Also below in `processServerIdentity`.

remote_attestation/java/tests/com/google/oak/remote_attestation/KeyNegotiatorTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Before
+  public void setUp() {
+    clientKeyNegotiator = new KeyNegotiator();
+    serverKeyNegotiator = new KeyNegotiator();
+  }
+
+  @Test
+  public void testCreateKeyNegotiator() throws GeneralSecurityException {
+    clientKeyNegotiator.getPublicKey();
```
rbehjati, line 41: What does this test check? If it is to check that the method does not throw an exception add a comment explaining it. However, something like AssertJ's `doesNotThrowAnyException` is nicer:
```
assertThatCode(() -> {
  // code that should throw an exception
  ...
}).doesNotThrowAnyException();
```

Or perhaps just check that the public key is not empty.

rbehjati (approved):

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -91,88 +75,104 @@ public ClientHandshaker(byte[] expectedTeeMeasurement) {
...
+      if (state != State.ExpectingServerIdentity) {
+        throw new IllegalStateException(
+            "ClientHandshaker is not in the ExpectingServerIdentity state");
+      }
+
+      Message.ServerIdentity serverIdentity =
+          Message.ServerIdentity.deserialize(serializedServerIdentity);
+
+      // Update current transcript.
+      // Transcript doesn't include transcript signature from the server identity message.
```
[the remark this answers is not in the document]
ipetr0v: It's a part of the SC protocol.
It's not possible to include it, because it will be including `ServerIdentity` hash inside `ServerIdentity`

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
ipetr0v: Updated to `throws RuntimeException`

rbehjati: It is an unchecked exception. You don't have to specify it in the signature. I think `public byte[] createClientHello() throws IOException {` should be enough.

remote_attestation/java/tests/com/google/oak/remote_attestation/KeyNegotiatorTest.java
ipetr0v: Done

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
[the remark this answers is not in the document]
rbehjati: Then I'd suggest updating the comment: 
```suggestion
      // Create a copy of serverIdentity but without the signature. Because the transcript must not include transcript signature from the server identity message.
```

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -91,88 +75,104 @@ public ClientHandshaker(byte[] expectedTeeMeasurement) {
...
-    // Verify server attestation info.
-    if (!verifyAttestationInfo(serverIdentity.getAttestationInfo())) {
-      throw new VerifyException("Couldn't verify attestation info");
+    try {
+      if (state != State.ExpectingServerIdentity) {
+        throw new IllegalStateException(
+            "ClientHandshaker is not in the ExpectingServerIdentity state");
+      }
+
+      Message.ServerIdentity serverIdentity =
```
rbehjati, line 122: Could Message be dropped? 
```suggestion
     ServerIdentity serverIdentity =
```

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -91,88 +75,104 @@ public ClientHandshaker(byte[] expectedTeeMeasurement) {
...
+          Message.ServerIdentity.deserialize(serializedServerIdentity);
+
+      // Update current transcript.
+      // Transcript doesn't include transcript signature from the server identity message.
+      ServerIdentity serverIdentityNoSignature = new Message.ServerIdentity(
```
rbehjati, line 127: Either change `ServerIdentity` to `Message. ServerIdentity`, or drop `Message` from the call to the constructor.

rbehjati (approved):

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
ipetr0v: It's a `static` class member, it can't be dropped

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
ipetr0v: Done

ipetr0v:
Added `Message` tests.

runner/src/main.rs
```diff
@@ -981,6 +981,14 @@ fn run_bazel_build() -> Step {
...
     Step::Single {
         name: "bazel test".to_string(),
-        command: Cmd::new("bazel", &["test", "--", "//oak/...:all"]),
+        command: Cmd::new(
+            "bazel",
+            &[
+                "test",
+                "--",
+                "//oak/...:all",
+                "//remote_attestation/java/tests/...:all",
```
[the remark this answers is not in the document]
ipetr0v: We currently don't have Java tests in `oak_functions` since it would require having a Java Remote Attestation server.
We rely only on integration tests in form of examples.

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
rbehjati: You have this import line: `import com.google.oak.remote_attestation.Message.ServerIdentity;`, so I believe `ServerIdentity` works too. But it is better to be consistent. If it is now `Message.ServerIdentity` everywhere, remove the import.

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -91,88 +75,104 @@ public ClientHandshaker(byte[] expectedTeeMeasurement) {
...
    *
-   * Transitions {@code ClientHandshaker} state from {@code Initializing} to
-   * {@code Attesting} state.
+   * Transitions {@code ClientHandshaker} state from {@code State.Initializing} to
+   * {@code State.ExpectingServerIdentity} state.
    */
-  public byte[] createClientHello() throws IllegalStateException, IOException {
-    if (state != HandshakerState.Initializing) {
-      throw new IllegalStateException("ClientHandshaker is not in the Initializing state");
+  public byte[] createClientHello() throws RuntimeException, IOException {
```
rbehjati, line 81:
```suggestion
  public byte[] createClientHello() throws IOException {
```
I am really curious, would not this work?

runner/src/main.rs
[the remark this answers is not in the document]
rbehjati: Thanks for checking :)

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
ipetr0v: Done

ipetr0v: I think this import was added by vscode automatically

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
ipetr0v: Done

PR #2304, opened 2021-09-27
Add Hello World example to Oak Functions

ipetr0v:
Adds a Hello World Oak Functions example, that can be used as a simplified End-to-End test for Remote Attestation.

### DIFF
--- a/oak_functions/examples/Cargo.lock
+++ b/oak_functions/examples/Cargo.lock
@@ -717,6 +717,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "hello_world"
+version = "0.1.0"
+dependencies = [
+ "oak_functions",
+]
+
+[[package]]
+name = "hello_world_client"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "env_logger",
+ "oak_functions_abi",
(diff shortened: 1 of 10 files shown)


tiziano88:
Could we use an existing example, e.g. the key value lookup?

ipetr0v:
IIUC key value lookup requires an external data source:
```
lookup_data_url = "https://storage.googleapis.com/oak_lookup_data/lookup_data_weather"
```

tiziano88:
> IIUC key value lookup requires an external data source:
> https://github.com/project-oak/oak/blob/9db00dd2ece9ed75cf93a4f6dda9fa3a2768b853/oak_functions/examples/key_value_lookup/config.toml#L1-L2
> 
> So it's not really hermetic.

I agree, but that's already there. Over time we should make that more hermetic (e.g. by loading data from a local file or local web server). It should not matter for this test, right? i.e. This would be testing the remote attestation part, not the functional correctness of the lookup logic.

ipetr0v:
> It should not matter for this test, right?

Yes, the only thing that matters for RA is client connecting to the server.
So data loading is more related to specific `oak_functions` tests.

PR #2309, opened 2021-09-29
Change private metrics to be an extension with feature gating support

### LINKED ISSUE #2289: Make metrics support in Oak Functions an extension
Similar to the TensorFlow support. See #2278.
Add feature gating support for it.

rbehjati:
* Adds an `ExtensionFactory` trait with a `create` method, for creating instances of `OakApiNativeExtension`.
* Supports stateful `OakApiNativeExtension` types. 
* Adds feature-gating for private metrics. 

Fixes #2289


oak_functions/loader/src/metrics.rs
```diff
@@ -239,18 +278,112 @@ impl PrivateMetricsProxy {
...
     }
 
-    /// Consumes the proxy and publishes the local request-specific data to the aggregator. If
-    /// publishing the metrics to the aggregator causes the batch threshold to be reached the
-    /// aggregated metrics are returned.
+    /// Publishes the local request-specific data to the aggregator. If publishing the metrics to
+    /// the aggregator causes the batch threshold to be reached the aggregated metrics are returned.
     ///
     /// See [PrivateMetricsAggregator::report_metrics] for more details.
-    pub fn publish(self) -> Option<(usize, Vec<(String, i64)>)> {
```
conradgrobler, line 247: Is it not possible to move `publish_metrics` elsewhere (or let `publish_metrics` also consume self) to keep this pattern of consuming the proxy when sending data to the aggregator? It was specifically done this way as an additional safeguard to ensure there is no way that a proxy can publish the data more than once.

oak_functions/loader/src/metrics.rs
```diff
@@ -239,18 +278,112 @@ impl PrivateMetricsProxy {
...
+    pub fn publish(&self) -> Option<(usize, Vec<(String, i64)>)> {
         if let Ok(mut aggregator) = self.aggregator.lock() {
-            aggregator.report_metrics(self.data)
+            aggregator.report_metrics(self.data.clone())
```
conradgrobler, line 287: Rather than cloning the data, perhaps change it to an `Option<Hashmap>` and use `self.data.take()`. This will consume the data and ensure there is no way to publish the data twice even if the proxy cannot be consumed.

rbehjati: Thanks. Done.

oak_functions/loader/src/metrics.rs
conradgrobler: I see. I think it is ok as long as the data is consumed.

oak_functions/loader/src/metrics.rs
```diff
@@ -218,39 +232,165 @@ fn inverse_laplace(beta: f64, p: f64) -> f64 {
...
+    }
+
+    fn terminate(&mut self) -> anyhow::Result<()> {
+        self.publish_metrics();
+        Ok(())
+    }
+}
+
+/// Corresponds to the host ABI function [`report_metric`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#report_metric).
+pub fn report_metric(
```
conradgrobler, line 359: Does this have to be publid?

conradgrobler (approved):

oak_functions/loader/src/metrics.rs
rbehjati: Fixed.

oak_functions/loader/src/metrics.rs
rbehjati: As you suggested, I added another struct for implementing the extension, and wrapping an instance of the proxy. So here the proxy is consumed.

oak_functions/loader/src/metrics.rs
```diff
@@ -244,13 +290,117 @@ impl PrivateMetricsProxy {
...
-    pub fn publish(self) -> Option<(usize, Vec<(String, i64)>)> {
+    fn publish(self) -> Option<(usize, Vec<(String, i64)>)> {
         if let Ok(mut aggregator) = self.aggregator.lock() {
             aggregator.report_metrics(self.data)
         } else {
             None
         }
     }
+
+    fn publish_metrics(self) {
```
conradgrobler, line 301: nit: perhaps move this function and the logger onto the `PrivateMetricsExtension` so that the proxy is only responsible for reporting metrics, not logging it?

conradgrobler (approved):

oak_functions/loader/src/metrics.rs
rbehjati: Good idea. Done.

PR #2312, opened 2021-10-05
Use wasmi in no_std mode

### LINKED ISSUE #2268: Investigate using Wasmi in no_std mode
Wasmi supports using vectors rather than mmap for memory allocation. This reduces the amount of unsafe code, but might have an adverse performance impact. We should benchmark the performance for some typical workloads to understand the potential performance trade-off.

See https://github.com/paritytech/wasmi/blob/master/Cargo.toml#L42-L54

conradgrobler:
Fixes #2268

# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/Cargo.toml
+++ b/oak_functions/loader/Cargo.toml
@@ -52,6 +52,7 @@ toml = "*"
 tonic = "*"
 tract-tensorflow = { version = "*", optional = true }
 url = "*"
+# Use wasmi in `no_std` mode.
 wasmi = { version = "*", default-features = false, features = ["core"] }
 signal-hook = "*"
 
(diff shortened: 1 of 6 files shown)


oak_functions/loader/Cargo.toml
```diff
@@ -51,7 +51,7 @@ toml = "*"
 tonic = "*"
 tract-tensorflow = { version = "*", optional = true }
 url = "*"
-wasmi = "*"
+wasmi = { version = "*", default-features = false, features = ["core"] }
```
rbehjati, line 54: Could you add a comment explaining that this uses wasmi in no_std mode?

rbehjati (approved):

oak_functions/loader/src/tests.rs
```diff
@@ -254,7 +254,7 @@ fn bench_wasm_handler(bencher: &mut Bencher) {
         // We expect the `mean` time for loading the test Wasm module and running its main function
         // to be less than a fixed threshold.
         assert!(
-            elapsed < Duration::from_millis(3),
+            elapsed < Duration::from_millis(5),
```
tiziano88, line 257: :chart_with_upwards_trend:

tiziano88 (approved):

PR #2370, opened 2021-11-02
Use Criterion for benchmarking

conradgrobler:
This change switches the Oak Functions loader benchmarks to use Criterion.

This change also moves the fixed-data lookup benchmark to the key_value_lookup example to support asserting that the bench takes less than the expected amount of time.

# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/benches/lookup.rs
+++ b/oak_functions/loader/benches/lookup.rs
@@ -19,7 +19,8 @@ pub mod proto {
 }
 
 use criterion::{
-    criterion_group, criterion_main, AxisScale, BenchmarkId, Criterion, PlotConfiguration,
+    criterion_group, criterion_main, measurement::Measurement, AxisScale, BenchmarkGroup,
+    BenchmarkId, Criterion, PlotConfiguration,
 };
 use lookup_data_generator::data::generate_and_serialize_random_entries;
 use oak_functions_abi::proto::{Request, StatusCode};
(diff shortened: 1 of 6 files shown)


oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
 }
 
-#[bench]
-fn large_random_data_multi_lookup(bencher: &mut Bencher) {
-    let test_data = generate_random_test_data_for_bench(1_024, 10_240, 1_000);
-
-    run_lookup_bench(bencher, test_data, 101, Duration::from_millis(25));
-}
+criterion_group!(benches, key_size, value_size, entry_count);
+criterion_main!(benches);
```
rbehjati, line 213: What kind of output do we get from running the benches? In other benchmarks we have an expected bound on the execution time. Can we have a check like that here too, or is that not necessary?

oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
 use proto::{benchmark_request::Action, BenchmarkRequest, LookupTest};
 use rand::SeedableRng;
-use std::{collections::HashMap, sync::Arc, time::Duration};
-use test::Bencher;
+use std::{collections::HashMap, sync::Arc};
+use tokio::runtime::Runtime;
 
 const MANIFEST_PATH: &str = "examples/benchmark/module/Cargo.toml";
+const SINGLE_REQUEST_ITERATIONS: u32 = 1;
+const MULTI_REQUEST_ITERATIONS: u32 = 101;
```
rbehjati, line 39: Why 101? I expected 100.

rbehjati (approved):

oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
-    };
-
-    let test_data = TestData {
-        test_key: br#"key_1"#.to_vec(),
-        expected_value: br#"value_1"#.to_vec(),
-        entries,
-    };
+fn key_size(c: &mut Criterion) {
+    let rt = Runtime::new().unwrap();
+    let wasm_module_bytes = bulid_wasm_module(MANIFEST_PATH);
```
tiziano88, line 43: (pre existing typo) build

oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
 
-    run_lookup_bench(bencher, test_data, 101, Duration::from_millis(10));
-}
+    let mut group = c.benchmark_group("key_size");
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
```
tiziano88, line 46: Is a log scale the most appropriate? I would have thought a linear scale is enough.

oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
+    group.plot_config(plot_config);
+    for size in [16usize, 128usize, 1_024usize].iter() {
```
tiziano88, line 48: If you use `into_iter` then I think you get actual values, not refs. Also, I thought `into_iter` is actually the default if you don't specify anything, but I may be wrong.

oak_functions/loader/benches/lookup.rs
```diff
@@ -33,125 +31,199 @@ use oak_functions_loader::{
...
+    let mut group = c.benchmark_group("entry_count");
+    let plot_config = PlotConfiguration::default().summary_scale(AxisScale::Logarithmic);
+    group.plot_config(plot_config);
+    for count in [100usize, 10_000usize, 1_000_000usize].iter() {
+        let TestData {
+            test_key,
+            expected_value,
+            entries,
+        } = generate_random_test_data_for_bench(64, 256, *count);
+        let lookup_data = Arc::new(LookupData::for_test(entries.clone()));
```
tiziano88, line 168: It seems from here on the logic is the same in all the benches, perhaps extract to a function?

tiziano88 (approved):

oak_functions/loader/benches/lookup.rs
conradgrobler: Running this produces output to stdout/stderr similar to:

```
key_size/multi/16       time:   [7.2419 ms 7.2496 ms 7.2587 ms]                              
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
```

It also generates an HTML report that contains charts of the performance at different points.

We cannot check that the execution time is below some bound, as benchmarking does not return a summary we can programatically inspect. That is why I moved the fixed lookup benchmark to the key_value_lookup example so that we can still check it somewhere.

oak_functions/loader/benches/lookup.rs
conradgrobler: 101 is so that it is easier to divide mentally. Doing 101 iterations and subtracting a single iteration leaves the approximate duration of 100 iterations.

conradgrobler (commented):
Thanks

oak_functions/loader/benches/lookup.rs
conradgrobler: Fixed.

oak_functions/loader/benches/lookup.rs
conradgrobler: I think so. The size values form a geometric sequence, so the linear plot is very skewed.

oak_functions/loader/benches/lookup.rs
conradgrobler: Done.

oak_functions/loader/benches/lookup.rs
conradgrobler: Done.

conradgrobler (commented):
Thanks.

PR #2372, opened 2021-11-03
Update Docker image

mariaschett:
# Checklist

      construction.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -217,12 +217,14 @@ RUN rustup component add \
   rustfmt
 
 # No binary available on Github, have to use cargo install.
-ARG deadlinks_version=0.5.0
+# https://github.com/deadlinks/cargo-deadlinks
+ARG deadlinks_version=0.8.0
 RUN cargo install --version=${deadlinks_version} cargo-deadlinks
 
 # Install cargo-fuzz.
 # To allow local testing of the fuzzing functionality.
-ARG cargo_fuzz_version=0.10.1
+# https://github.com/rust-fuzz/cargo-fuzz
+ARG cargo_fuzz_version=0.10.2
 RUN cargo install --version=${cargo_fuzz_version} cargo-fuzz
 
 # Where to install rust tooling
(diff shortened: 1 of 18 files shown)


Dockerfile
```diff
@@ -121,23 +121,21 @@ RUN mkdir --parents ${GOROOT} \
   && sha256sum --binary ${golang_temp} && echo "${golang_sha256} *${golang_temp}" | sha256sum --check \
   && tar --extract --gzip --file=${golang_temp} --directory=${GOROOT} --strip-components=1 \
   && rm ${golang_temp} \
-  && go version
-
+  && go version \
```
tiziano88, line 124: Was this required by some linter? If not, I think it is useful to keep the layers separate, otherwise we would need to recreate the Go layer even if we just want to update one of these tools.

mariaschett: Yes, hadolint DL3059. It is not [uncontroversial though](https://github.com/hadolint/hadolint/issues/715). 

Two reasons why I kept it: i) may help with performance, and ii) `/scripts/runner run-ci` errors  even if only an info (not an error) is produced by hadolint.

tiziano88: I would be inclined to ignore that lint; choosing how to layer a docker image does not seem something that a linter should be opinionated about, it depends on how the image is meant to be used and built. I think we would have more context to decide when it is appropriate to merge layers. WDYT?

mariaschett: I have no strong feelings either direction, I tend to go with the linter. But happy to change. Do you want me to add DL3059 to be ignred in [hadolint.yaml](https://github.com/project-oak/oak/blob/main/.hadolint.yaml) and revert the change in the Docker?

tiziano88: Yeah I think so. Unless anyone else feels the other way (but I don't think so, from our previous discussions)

mariaschett: Done! I'll leave the merging of layers for another day.

tiziano88 (approved):
ship it!

PR #2375, opened 2021-11-03
Pin Docker image to specific hash

### LINKED ISSUE #2374: Pin Docker image to specific hash for CI and local development

tiziano88:
This ensures that each commit in the repository has a cryptographic
reference to the exact Docker image to use to build any artifacts.

Fixes #2374

### DIFF
--- a/.github/workflows/build_rust_docs.yaml
+++ b/.github/workflows/build_rust_docs.yaml
@@ -39,12 +39,11 @@ jobs:
           docker rmi $(docker image ls --all --quiet)
           df --human-readable
 
-      # Build Docker image, caching from the latest version from the remote repository.
-      - name: Docker build
-        timeout-minutes: 30
+      - name: Docker pull
+        timeout-minutes: 10
         run: |
-          docker pull gcr.io/oak-ci/oak:latest
-          docker build --pull --cache-from=gcr.io/oak-ci/oak:latest --tag=gcr.io/oak-ci/oak:latest .
+          ./scripts/docker_pull
+          df --human-readable
 
       # Remove all files from the "out" folder.
       - name: Clear "out" folder
--- a/scripts/common
+++ b/scripts/common
@@ -11,7 +11,20 @@ set -o pipefail
 export RUST_LOG="${RUST_LOG:-info}"
 
 # See https://pantheon.corp.google.com/gcr/images/oak-ci/GLOBAL/oak
-readonly DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak'
+readonly DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak:latest'
+
+# The difference between Docker image id and image digest is that the image id corresponds to the
+# hash of the contents of the image, while the image digest is a hash of the image and its metadata,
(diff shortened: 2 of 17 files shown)


.github/workflows/build_rust_docs.yaml
```diff
@@ -39,12 +39,11 @@ jobs:
...
-      # Build Docker image, caching from the latest version from the remote repository.
-      - name: Docker build
-        timeout-minutes: 30
+      - name: Docker pull
+        timeout-minutes: 10
         run: |
-          docker pull gcr.io/oak-ci/oak:latest
-          docker build --pull --cache-from=gcr.io/oak-ci/oak:latest --tag=gcr.io/oak-ci/oak:latest .
+          ./scripts/docker_pull
+          df --human-readable
```
mariaschett, line 46: Why do we run this after running ./scripts/docker_pull (also ff)?

scripts/fix_docker_user_and_run
```diff
@@ -0,0 +1,17 @@
...
+
+set -o errexit
+set -o nounset
+set -o xtrace
+set -o pipefail
+
+groupmod --gid="${HOST_GID}" docker
+usermod --uid="${HOST_UID}" --gid="${HOST_GID}" docker
+chown "${HOST_UID}":"${HOST_GID}" "/home/docker" "/home/docker/.cache"
+su docker --session-command="$*"
```
mariaschett, line 17: Why do we need `su` here?

scripts/common
```diff
@@ -11,7 +11,14 @@ set -o pipefail
...
 
 # See https://pantheon.corp.google.com/gcr/images/oak-ci/GLOBAL/oak
-readonly DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak'
+readonly DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak:latest'
+
+# Do not modify manually. This value is automatically updated by ./scripts/docker_build .
+readonly DOCKER_IMAGE_ID='sha256:f71ebe2c5827ea559e4c5fd9bb668c7a23ed8d828a4344d8c3fec0714622188d'
+
+# Do not modify manually. This value is automatically updated by ./scripts/docker_push .
+readonly DOCKER_IMAGE_REPO_DIGEST='gcr.io/oak-ci/oak@sha256:ff9eafaf9f64d6549039fd1e7c5a9163206d60495d0d232971c57fa5a52e7878'
```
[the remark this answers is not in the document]
rbehjati: Thanks for clarifying this. Might be worth explaining this in the comments.

rbehjati (approved):
Thanks :)

.github/workflows/build_rust_docs.yaml
mariaschett: Yes, the `df` command. I see, I was just wondering if the `df` should go into docker_pull script, because they appear together often. But probably not in that case.

scripts/fix_docker_user_and_run
mariaschett: Unfortunately I have no other suggestion.

PR #2382, opened 2021-11-09
Add gRPC API key to Java Oak Functions client

ipetr0v:


oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -70,15 +77,22 @@ public AttestationClient() {}
    *
    * @param url must contain a protocol used for the connection ("https://" or "http://")
    */
-  public void attest(String url, Predicate<ConfigurationInfo> verifier)
+  public void attest(String url, String apiKey, Predicate<ConfigurationInfo> verifier)
```
tiziano88, line 80: Update doc, also clarify what happens if apiKey is empty / null

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -214,4 +227,36 @@ public Response send(Request request)
...
+
+  /**
+   * Intercepts gRPC requests and adds Metadata with an API key.
+   * https://cloud.google.com/endpoints/docs/grpc/restricting-api-access-with-api-keys#java
+   */
+  private static final class Interceptor implements ClientInterceptor {
+    private final String apiKey;
+
+    private static final Metadata.Key<String> API_KEY_HEADER =
+        Metadata.Key.of("x-goog-api-key", Metadata.ASCII_STRING_MARSHALLER);
```
tiziano88, line 239: Where does this string come from? Please add links / documentation, and consider making a constant.

tiziano88 (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Done

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Done

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -178,13 +201,13 @@ boolean verifyServerIdentity(byte[] serializedServerIdentity,
...
       return false;
-    };
+    }
 
     return true;
   }
 
   @Override
-  protected void finalize() throws Throwable {
+  public void finalize() {
```
conradgrobler, line 210: Why has this become public?

conradgrobler (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Needs to be explicitly called so that Oak Functions server doesn't have to wait for Java garbage collector to stop a gRPC session.
Added a comment

conradgrobler: I think the preferred approach is to have a separate public method (e.g. `stop()` or `close()`) to do that and then call stop from the protected `finalize()`.

PR #2384, opened 2021-11-09
Add `./scripts/runner cargo-clean`

mariaschett:
# Checklist

      construction.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -99,6 +99,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             Command::Completion => panic!("should have been handled above"),
             Command::RunCargoDeny => run_cargo_deny(),
             Command::RunCargoUdeps => run_cargo_udeps(),
+            Command::RunCargoClean => run_cargo_clean(),
         };
         // TODO(#396): Add support for running individual commands via command line flags.
         let remaining_steps = steps.len();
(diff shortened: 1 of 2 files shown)


runner/src/main.rs
```diff
@@ -967,6 +968,24 @@ fn run_cargo_udeps() -> Step {
     }
 }
 
+fn run_cargo_clean() -> Step {
+    Step::Multiple {
+        name: "cargo clean".to_string(),
+        steps: crate_manifest_files()
+            .map(|manifest_path| {
+                let manifest_path = manifest_path.to_str().unwrap();
```
rbehjati, line 976: You could first map `manifest_path` to string by calling (`to_string`). It is more idiomatic to do this using another call to map (`map(to_string)`). See https://github.com/project-oak/oak/blob/b599b42578d9af0f51a33dc4eb8f1a7f3c54c184/runner/src/main.rs#L906-L907

runner/src/main.rs
```diff
@@ -967,6 +968,24 @@ fn run_cargo_udeps() -> Step {
...
+                Step::Single {
+                    name: manifest_path.to_string(),
```
rbehjati, line 978: If you first map `manifest_path` to string, then here you can call `clone` here.

rbehjati (approved):
Nice. Thanks :)

runner/src/main.rs
```diff
@@ -972,8 +972,9 @@ fn run_cargo_clean() -> Step {
     Step::Multiple {
         name: "cargo clean".to_string(),
         steps: crate_manifest_files()
+            .map(to_string)
             .map(|manifest_path| {
-                let manifest_path = manifest_path.to_str().unwrap();
+                let manifest_path = manifest_path.clone();
```
rbehjati, line 977: This does not do anything useful :) You don't need this line anymore.

runner/src/main.rs
mariaschett: done!

runner/src/main.rs
mariaschett: done!

runner/src/main.rs
mariaschett: Thank you for catching that!

PR #2388, opened 2021-11-10
Update request coordinates

ipetr0v:


oak_functions/examples/key_value_lookup/example.toml
```diff
@@ -13,7 +13,8 @@ additional_args = [
 ]
 
 [clients]
+# Test request coordinates are defined in `oak_functions/lookup_data_generator/src/data.rs`.
 rust = { Cargo = { cargo_manifest = "oak_functions/client/rust/Cargo.toml" }, additional_args = [
   "--uri=http://localhost:8080",
-  "--request={\"lat\":52,\"lng\":0}"
+  "--request={\"lat\":0,\"lng\":0}"
```
tiziano88, line 19: I don't remember why this k/v lookup example is using weather data?

tiziano88 (approved):

oak_functions/examples/key_value_lookup/example.toml
ipetr0v: Because it's the only example database that we had

PR #2389, opened 2021-11-10
Pass gRPC channel to the Remove Attestation implementation

ipetr0v:


ipetr0v:
@tiziano88 what do you think about splitting TLS and non-TLS attestation into separate functions?
I personally think that just having a `bool tls` argument is very error prone.

tiziano88:
I'm not sure, it seems to replicate the signature and most of the implementation. Could you perhaps take inspiration from how gRPC does that already? In fact, why not just pass the channel builder in the method? Then it can be constructed in whatever way. We also mentioned that clients would probably want to set all sort of other things, e.g. timeouts, additional headers, etc.

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -72,38 +72,26 @@ public VerificationException(String msg) {
...
-      channel = ManagedChannelBuilder.forAddress(parsedUrl.getHost(), parsedUrl.getPort())
-                    .intercept(interceptors)
-                    .build();
-    } else {
-      channel = ManagedChannelBuilder.forAddress(parsedUrl.getHost(), parsedUrl.getPort())
-                    .usePlaintext()
-                    .intercept(interceptors)
-                    .build();
-    }
+    channel = builder.intercept(interceptors).build();
```
tiziano88, line 94: I think this logic should also be done by the caller, perhaps via some other helper. It has nothing to do with attestation. This method should get an already created channel.

conradgrobler: +1

ipetr0v: Done

conradgrobler (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -72,42 +72,32 @@ public VerificationException(String msg) {
...
-   * @param verifier Checks that the ServerIdentity contains the expected attestation info as
-   * described in {@code ServerIdentityVerifier::verifyAttestationInfo}.
+   *
+   * @return an updated instance of a {@code ManagedChannelBuilder}.
    */
-  public void attest(String url, String apiKey, Predicate<ConfigurationInfo> verifier)
-      throws GeneralSecurityException, IOException, InterruptedException, VerificationException {
-    // Create gRPC channel.
-    URL parsedUrl = new URL(url);
+  public static ManagedChannelBuilder addApiKey(ManagedChannelBuilder builder, String apiKey) {
```
tiziano88, line 85: I still think this method has nothing to do with attestation and should not be here, also I'm not sure it's idiomatic to get a channel builder and modify it in place from a method like this. In fact, there is probably some utility that does it already which we probably haven't found yet. It's fine to fix later, but do you agree that this should not be here eventually?

tiziano88 (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Yes, we can create a utility library for creating gRPC channels.
Because otherwise all of this code will have to be in the client.
I.e. we are trying to encapsulate the gRPC channel creation logic and at the same time give clients the ability to fully manage the gRPC channel creation.

Filed an issue: https://github.com/project-oak/oak/issues/2411

PR #2410, opened 2021-11-26
ci: Add cifuzz integration

DavidKorczynski:
Signed-off-by: David Korczynski <david@adalogics.com>

This adds CI integration with CIFuzz. CIFuzz will run the oak OSS-Fuzz fuzzers for 600 seconds for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -4,22 +4,22 @@ jobs:
   Fuzzing:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


rbehjati (approved):
LGTM. Thanks. 
However, the format-check step is failing. We have a runner command that can fix the formatting: `./scripts/docker_run ./scripts/runner format --commits=1`. This command uses `prettier`, which might be easier to run.

Issue #1963, opened 2021-04-14
Add an abitest example for oak-functions


rbehjati:
Similar to the existing `abitest` we need an example to test oak-functions' ABI functions. 
Ref: https://github.com/project-oak/oak/pull/1958#pullrequestreview-634991559 

Some scenarios to include:
- What happens if the abi functions are called more than once

ipetr0v:
> What happens if the abi functions are called more than once

What should happen in this case (not for the `storage_get_item`)?

rbehjati:
We have documented: https://project-oak.github.io/oak/oak_functions/sdk/

For `read_request`: This function is idempotent. Multiple call to this function all return the same value.
For `write_response`: Multiple calls to this function will replace the earlier responses. Only the last response that is written will be kept and returned to the user.

PR #2413, opened 2021-11-26
Refactor example/abitest as integration test of SDK

mariaschett:
# Checklist

      construction.


ipetr0v:
What is `oak_functions/sdk` directory? Is it only used for tests?

ipetr0v (approved):

oak_functions/sdk/oak_functions/tests/module/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+type TestFn = fn(&str) -> ();
+
+struct TestManager<'a> {
+    tests: HashMap<&'a str, TestFn>,
+}
+
+impl TestManager<'static> {
+    fn new() -> Self {
+        Self {
+            tests: hashmap! [
```
tiziano88, line 33: I was thinking about this, I think once we have a way of specifying something like RPCs over channels, we could abstract this away via something like tarpc, with a method for each test, and generate client and server stubs automatically. Would that work?

mariaschett:
> What is `oak_functions/sdk` directory? Is it only used for tests?

I am not sure I get your question. :)

We use the oak_functions/sdk e.g., in [examples](https://github.com/project-oak/oak/blob/3935a27c710cd5ea98a382b69de282cd57b48e4f/oak_functions/examples/key_value_lookup/module/Cargo.toml#L12). But I am not sure that was your question.

Or are you confused by the folder structure within `oak_functions/sdk`?

oak_functions/sdk/oak_functions/tests/module/src/lib.rs
mariaschett: I don't see why not! I'll keep it in mind!

oak_functions/sdk/oak_functions/tests/integration_test.rs
```diff
@@ -0,0 +1,67 @@
...
+use std::sync::Arc;
+use tokio;
+
+lazy_static::lazy_static! {
+    static ref RUNTIME: tokio::runtime::Runtime = tokio::runtime::Runtime::new().unwrap();
+}
+
+#[test]
+fn test_sdk() {
+    let manifest_path = "/workspace/oak_functions/sdk/oak_functions/tests/module/Cargo.toml";
```
rbehjati, line 30: This hard-coded path only works inside the docker image. It is better to programatically get the path to the working directory as we do in other tests:
https://github.com/project-oak/oak/blob/3935a27c710cd5ea98a382b69de282cd57b48e4f/oak_functions/examples/weather_lookup/module/src/tests.rs#L45-L50

oak_functions/sdk/oak_functions/tests/integration_test.rs
```diff
@@ -0,0 +1,67 @@
...
+    for (request_body, expected_response_body) in tests {
+        let request = Request {
+            body: request_body.as_bytes().to_vec(),
+        };
+        let response: Result<Response, _> = RUNTIME.block_on(wasm_handler.handle_invoke(request));
+
+        let actual_response_body: String = {
+            let rb = response.unwrap().body;
+            String::from_utf8(rb).unwrap()
+        };
```
rbehjati, line 64: Is the type annotation (`: String`) necessary?

Also, I think this could be simplified to 
```
 let actual_response_body = std::str::from_utf8(response.unwrap().body).unwrap();
```
(and `actual_response_body` does not have to be `String`, it can be `&str`)

oak_functions/sdk/oak_functions/tests/module/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    // TODO(#2415): Add tests for report_metric.
+
+    // TODO(#2416): Add tests for tf_model_infer.
+}
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn main() {
+    let test_manager = TestManager::new();
+    test_manager.run_test().expect("Couldn't run test");
```
rbehjati, line 108: I am not sure how the wasm engine handles the panics. It might be better to replace all matches and expects in this module with a `Result` that can be returned as a response using `oak_functions::write_response` and be checked in the integration test.

rbehjati (approved):
Thanks :) 
Can #1963 be closed when this PR is merged?

mariaschett:
> Can #1963 be closed when this PR is merged?

I'd say #1963 also depends on #2417, and ideally #2416, #2415.

oak_functions/sdk/oak_functions/tests/integration_test.rs
mariaschett: Done, thank you very much for this!

oak_functions/sdk/oak_functions/tests/module/src/lib.rs
mariaschett: From my understanding, an empty response is returned on a panic, so test_sdk fails. 

About returning a `Response` with a `Result`: I have thought about it, but I hoped for the tests to not rely on the request/response mechanism. But I don't see any way to do that.

Do you think returning a `Result` and checking in the integration test would be much better?

rbehjati: > From my understanding, an empty response is returned on a panic, so test_sdk fails.

The test still fails, but it would be impossible/very difficult to know why it failed. Because we don't get the panic messages logged anywhere. 

> About returning a Response with a Result: I have thought about it, but I hoped for the tests to not rely on the request/response mechanism. But I don't see any way to do that.

Currently `write_response` is the only way for a Wasm module to communicate with the world outside. `write_log_message` is an alternative, but currently we only write log messages to stdout, and so they cannot be used for testing. A different implementation of `write_log_message` that would store the log messages in a vector or something could work. But we don't have that implementation, and from the perspective of the test, it would not be very different from `write_response`.

> Do you think returning a Result and checking in the integration test would be much better?

It would make debugging easier. However, since the Wasm module logic is not very complicated debugging an empty response might not be difficult for now. 

In any case, I'd say this can be done in a separate PR :)

rbehjati:
> I'd say #1963 also depends on #2417, and ideally #2416, #2415.

Ah! Good point!

ipetr0v:
> Or are you confused by the folder structure within oak_functions/sdk?

I think I just wondered why `oak_functions/sdk` had `oak_functions` inside it.
But it could be a separate discussion (for a different PR)

oak_functions/sdk/oak_functions/tests/module/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+authors = ["Maria Schett <mschett@google.com>"]
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dev-dependencies]
+tokio = { version = "*" }
```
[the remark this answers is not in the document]
mariaschett: it's not, I realized. :)

PR #2425, opened 2021-12-06
Inconsistent view of in-memory lookup data while a Wasm function executes

mariaschett:
# Checklist

      construction.

### DIFF
--- a/oak_functions/loader/Cargo.toml
+++ b/oak_functions/loader/Cargo.toml
@@ -34,6 +34,7 @@ humantime-serde = "*"
 log = { version = "*", features = ["max_level_off", "release_max_level_off"] }
 oak_functions_abi = { path = "../abi" }
 oak_remote_attestation = { path = "../../remote_attestation/rust/" }
+once_cell = "*"
 prost = "*"
 rand = "*"
 serde = "*"
--- a/oak_functions/loader/Cargo.toml
+++ b/oak_functions/loader/Cargo.toml
@@ -56,6 +56,7 @@ signal-hook = "*"
 
 [dev-dependencies]
 criterion = "*"
+lazy_static = "*"
 lookup_data_generator = { path = "../lookup_data_generator" }
 maplit = "*"
 tempfile = "*"
--- a/oak_functions/loader/tests/integration_test.rs
+++ b/oak_functions/loader/tests/integration_test.rs
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use futures::executor::block_on;
 use lazy_static::lazy_static;
 use maplit::hashmap;
 use oak_functions_abi::proto::{Request, Response};
@@ -50,7 +51,7 @@ async fn test_consistent_lookup() {
         Logger::for_test(),
     ));
 
-    lookup_data.refresh().await.unwrap();
+    let _ = block_on(lookup_data.refresh());
 
     let wasm_handler = WasmHandler::create(
         &WASM_MODULE_BYTES,
@@ -73,7 +74,7 @@ async fn test_consistent_lookup() {
     let serialized_entries = test_utils::serialize_entries(entries);
     temp_file.as_file().write_all(&serialized_entries).unwrap();
 
-    lookup_data.refresh().await.unwrap();
+    let _ = block_on(lookup_data.refresh());
 
     let request2 = Request {
         body: b"ConsistentLookup".to_vec(),
--- a/oak_functions/loader/tests/integration_test.rs
+++ b/oak_functions/loader/tests/integration_test.rs
@@ -45,15 +45,16 @@ async fn test_consistent_lookup() {
     let temp_file = tempfile::NamedTempFile::new().unwrap();
     temp_file.as_file().write_all(&serialized_entries).unwrap();
 
-    let lookup_data = crate::LookupData::new_empty(
+    let lookup_data = Arc::new(crate::LookupData::new_empty(
         Some(LookupDataSource::File(temp_file.path().to_path_buf())),
         Logger::for_test(),
-    );
+    ));
+
     lookup_data.refresh().await.unwrap();
 
     let wasm_handler = WasmHandler::create(
         &WASM_MODULE_BYTES,
-        Arc::new(lookup_data),
+        lookup_data.clone(),
         vec![],
         Logger::for_test(),
     )
@@ -72,7 +73,7 @@ async fn test_consistent_lookup() {
     let serialized_entries = test_utils::serialize_entries(entries);
     temp_file.as_file().write_all(&serialized_entries).unwrap();
 
-    wasm_handler.refresh_lookup_data().await.unwrap();
+    lookup_data.refresh().await.unwrap();
 
     let request2 = Request {
         body: b"ConsistentLookup".to_vec(),
--- a/oak_functions/loader/tests/integration_test.rs
+++ b/oak_functions/loader/tests/integration_test.rs
@@ -15,7 +15,7 @@
 
 use futures::executor::block_on;
 use maplit::hashmap;
-use oak_functions_abi::proto::{Request, Response};
+use oak_functions_abi::proto::Request;
 use oak_functions_loader::{
     logger::Logger,
     lookup::{LookupData, LookupDataSource},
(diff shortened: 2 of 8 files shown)


oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,85 @@
...
+    let response: Response = wasm_handler.handle_invoke(request).await.unwrap();
+    assert_eq!(response.body().unwrap(), b"Value1");
+
+    let entries = hashmap! {
+       b"ConsistentLookup".to_vec() => b"Value2".to_vec(),
+    };
+    let serialized_entries = test_utils::serialize_entries(entries);
+    temp_file.as_file().write_all(&serialized_entries).unwrap();
+
+    wasm_handler.refresh_lookup_data().await.unwrap();
```
mariaschett, line 75: @project-oak/core Here I need to refresh the lookup data for the wasm_handler after changing the lookupdata in the external file. As `lookupdata` is not public in `WasmHandler`, I implemented `refresh_lookup_data` for WasmHandler. 

I am not particularily happy about it as I implement functionality only for testing. Do you see any other way (e.g., waiting for a automatic refresh/refactor this as a unit test) or do you think it is alright to add `refresh_lookup_data` for `WasmHandler`?

rbehjati: I think you can change `lookup_data` declared in line 48 to be `Arc<LookupData>`. Then pass `lookup_data.clone()` to `WasmHandler::create` and call `lookup_data.refresh()` here instead of having to add a method to `WasmHandler`. Kind of similar to https://github.com/project-oak/oak/blob/6769de3ec5b2e3115708d45f9eb7b19528886c29/oak_functions/loader/src/main.rs#L271-L279

mariaschett: That worked---thank you very much! :)

oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,86 @@
...
+use oak_functions_abi::proto::{Request, Response};
+use oak_functions_loader::{
+    logger::Logger,
+    lookup::{LookupData, LookupDataSource},
+    server::WasmHandler,
+};
+use std::{io::Write, sync::Arc};
+use tokio;
+
+lazy_static! {
```
tiziano88, line 27: I think https://docs.rs/once_cell/latest/once_cell/ is the recommended way of doing static initialisation, these days (see https://news.ycombinator.com/item?id=24672238)

mariaschett:
Well, that all checks have passed is not good: my test should have failed.

rbehjati:
> Well, that all checks have passed is not good: my test should have failed.

I ran it once locally and it failed: 

```
2021-12-06T19:03:03Z DEBUG - lookup data write lock acquisition time: 5µs
thread 'test_consistent_lookup' panicked at 'assertion failed: `(left == right)`
  left: `[86, 97, 108, 117, 101, 50]`,
 right: `[86, 97, 108, 117, 101, 49]`', loader/tests/integration_test.rs:85:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
A delay between the `refresh` and invoking the WASM logic should make the test more deterministic.

oak_functions/loader/tests/integration_test.rs
mariaschett: done, thank you!

oak_functions/loader/Cargo.toml
```diff
@@ -56,6 +57,7 @@ signal-hook = "*"
 
 [dev-dependencies]
 criterion = "*"
+lazy_static = "*"
```
tiziano88, line 60: Remove?

oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,91 @@
...
+    };
+    let serialized_entries = test_utils::serialize_entries(entries);
+    let temp_file = tempfile::NamedTempFile::new().unwrap();
+    temp_file.as_file().write_all(&serialized_entries).unwrap();
+
+    let lookup_data = Arc::new(crate::LookupData::new_empty(
+        Some(LookupDataSource::File(temp_file.path().to_path_buf())),
+        Logger::for_test(),
+    ));
+    let _ = block_on(lookup_data.refresh());
```
tiziano88, line 51: I'm not sure what happens when calling block_on within an async fn. Does .await not work here?

oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,91 @@
...
+    let wasm_handler = WasmHandler::create(
+        &WASM_MODULE_BYTES,
+        lookup_data.clone(),
+        vec![],
+        Logger::for_test(),
+    )
+    .expect("Could not instantiate WasmHandler.");
+
+    let request = Request {
+        body: b"ConsistentLookup".to_vec(),
```
tiziano88, line 62: May be worth creating local constants for the various interesting values

oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,91 @@
...
+    // Now we update the entries and refresh the lookup_data.
+    let entries = hashmap! {
+       b"ConsistentLookup".to_vec() => b"Value2".to_vec(),
+    };
+    let serialized_entries = test_utils::serialize_entries(entries);
+    temp_file.as_file().write_all(&serialized_entries).unwrap();
+    let _ = block_on(lookup_data.refresh());
+
+    // A second Wasm module created at this point would expect the new lookup data.
+    let wasm_handler2 = WasmHandler::create(
```
tiziano88, line 77: The numbering seems a bit inconsistent, it jumps from nothing to 2 :)

oak_functions/loader/tests/integration_test.rs
```diff
@@ -0,0 +1,91 @@
...
+        .expect("Could not read Wasm module")
+});
+
+#[tokio::test]
+async fn test_consistent_lookup() {
+    let entries = hashmap! {
+       b"ConsistentLookup".to_vec() => b"Value1".to_vec(),
+    };
+    let serialized_entries = test_utils::serialize_entries(entries);
+    let temp_file = tempfile::NamedTempFile::new().unwrap();
```
tiziano88, line 44: Some of this logic could probably be encapsulated in a helper fn at some point. I also thought we have a way of setting the data for test without going through a file, but I'm not sure

oak_functions/loader/tests/module/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+//
+
+use anyhow::{anyhow, Context};
+use assert_matches::assert_matches;
+use maplit::hashmap;
+use std::collections::HashMap;
+
+type TestFn = fn(&str) -> ();
+
+struct TestManager<'a> {
```
tiziano88, line 24: What is the purpose of the lifetime here? I would expect all strings to be 'static

oak_functions/loader/tests/module/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+    tests: HashMap<&'a str, TestFn>,
+}
+
+impl TestManager<'static> {
+    fn new() -> Self {
+        Self {
+            tests: hashmap! [
```
tiziano88, line 31: (nit) we usually prefer the { initialization for maps instead of [ (which is for lists)

oak_functions/loader/tests/module/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+                "ConsistentLookup" => Self::test_consistent_lookup as TestFn,
```
tiziano88, line 32: Do you expect there to be more methods in the future? Since this module seems very specific to the test case from which it is used already, I'm not sure we need another layer of indirection within the module itself. But let me know if I'm misunderstanding the purpose of this map.

oak_functions/loader/tests/integration_test.rs
tiziano88, line 62: IIUC the body is the method name to invoke in the module, right? But there is only a single method there, so this is not super obvious, maybe add a comment? Also do you think this may get better once we have an IDL in the unified ABI?

tiziano88 (approved):

mariaschett:
Fixed by #2597 .

PR #2426, opened 2021-12-06
Add definitions of Wasm host functions based on channels

mariaschett:
# Checklist

      construction.

### DIFF
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -86,10 +86,10 @@ extern "C" {
         inference_len_ptr: *mut usize,
     ) -> u32;
 
-    /// Reads a message, i.e., `dest_buf_len_ptr` bytes, from the channel with `channel_handle` into
-    /// the buffer at `dest_buf_ptr_ptr`. After a successful call to `channel_read`
-    /// the buffer at `dest_buf_ptr_ptr` holds bytes encoding a message taken from the specified
-    /// channel.
+    /// Allocates a buffer on the callers memory to write a message from the channel with
+    /// `channel_handle`. After a successful call `dest_buf_ptr_ptr` holds the address of the
+    /// buffer and `dest_buf_len_ptr` holds its length. The caller can then read
+    /// the message from `dest_buf_ptr_ptr`.
     ///
     /// Returns a status code to indicate success. In particular, returns immediately
     /// with an appropriate status code, if no message is available on the channel.
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -108,7 +108,7 @@ extern "C" {
     /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
     /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
     ///
-    /// Returns a status code to indicate success or whether the deadline.
+    /// Returns a status code to indicate success or deadline expiration.
     pub fn channel_wait(
         channel_handle_buf_ptr: *mut i32,
         channel_handle_buf_len: usize,
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -99,7 +99,7 @@ extern "C" {
         dest_buf_len_ptr: *mut usize,
     ) -> u32;
 
-    /// Writes a message, i.e., `src_buf_len` bytes),from `scr_buf_ptr` into the channel with
+    /// Writes a message, i.e., `src_buf_len` bytes, from `scr_buf_ptr` into the channel with
     /// `channel_handle`.
     ///
     /// Returns a status code to indicate success.
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -113,6 +113,8 @@ extern "C" {
     /// `channel_handle_buf_ptr` has a message to read, or
     /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
     /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
+    /// Both, `channel_handle_buf_len` and `ready_channel_handle_buf_len` hold the length of the
+    /// respective buffers in bytes.
     ///
     /// Returns a status code to indicate success or deadline expiration.
     pub fn channel_wait(
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -18,6 +18,7 @@
 //! binary interface (ABI).
 
 type ChannelHandle = u32;
+type StatusCode = u32;
 
 pub mod proto {
     include!(concat!(env!("OUT_DIR"), "/oak.functions.abi.rs"));
@@ -97,7 +98,7 @@ extern "C" {
         channel_handle: ChannelHandle,
         dest_buf_ptr_ptr: *mut *mut u8,
         dest_buf_len_ptr: *mut usize,
-    ) -> u32;
+    ) -> StatusCode;
 
     /// Writes a message, i.e., `src_buf_len` bytes, from `scr_buf_ptr` into the channel with
     /// `channel_handle`.
@@ -107,7 +108,7 @@ extern "C" {
         channel_handle: ChannelHandle,
         src_buf_ptr: *const u8,
         src_buf_len: usize,
-    ) -> u32;
+    ) -> StatusCode;
 
     /// Waits until at least one of the channels from the channel handles in the buffer at
     /// `channel_handle_buf_ptr` has a message to read, or
@@ -123,5 +124,5 @@ extern "C" {
         ready_channel_handle_buf_ptr: *mut *mut i32,
         ready_channel_handle_buf_len: *mut usize,
         deadline_ms: u32,
-    ) -> u32;
+    ) -> StatusCode;
 }
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -104,7 +104,7 @@ extern "C" {
     pub fn channel_write(channel_handle: i32, src_buf_ptr: *mut u8, src_buf_len: usize) -> u32;
 
     /// Waits until at least one of the channels from the channel handles in the buffer at
-    /// `channel_handle_buf_ptr` has a message to read---or
+    /// `channel_handle_buf_ptr` has a message to read, or
     /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
     /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
     ///
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -105,7 +105,7 @@ extern "C" {
     /// Returns a status code to indicate success.
     pub fn channel_write(
         channel_handle: ChannelHandle,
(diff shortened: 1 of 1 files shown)


oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
         inference_ptr_ptr: *mut *mut u8,
         inference_len_ptr: *mut usize,
     ) -> u32;
+
+    /// Reads a message, i.e., `dest_buf_len_ptr` bytes, from the channel with `channel_handle` into
+    /// the buffer at `dest_buf_ptr_ptr`. After a successful call to `channel_read`
```
tiziano88, line 88: Note that there is no buffer at `dest_buf_ptr_ptr` when this method is invoked. Instead, that parameter should point to a variable that will receive the address of a buffer, which is allocated by the runtime by invoking the `alloc` function from the module.

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
         inference_ptr_ptr: *mut *mut u8,
         inference_len_ptr: *mut usize,
     ) -> u32;
+
+    /// Reads a message, i.e., `dest_buf_len_ptr` bytes, from the channel with `channel_handle` into
```
tiziano88, line 87: How about documenting the ABI in the readme file, and referencing it from here? Like the other methods in this file. Otherwise it does not really scale to document the same ABI in each potential new language (e.g. if we add Go / C++ support in the future).

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+    /// the buffer at `dest_buf_ptr_ptr`. After a successful call to `channel_read`
+    /// the buffer at `dest_buf_ptr_ptr` holds bytes encoding a message taken from the specified
+    /// channel.
+    ///
+    /// Returns a status code to indicate success. In particular, returns immediately
+    /// with an appropriate status code, if no message is available on the channel.
+    pub fn channel_read(
+        channel_handle: i32,
```
tiziano88, line 95: I don't remember what was the conclusion of the discussion between i32 and u32 for handles? I would expect that in the runtime we would represent them as u32 (the sign of the handle does not seem to have a meaning, so i32 seems misleading). Granted, both of them would be represented as i32 in Wasm, but as long as we can use a more precise type here (u32) and Rust converts it to the appropriate underlying type (i32), then I think we should do that. Another reason for using u32 on the runtime side is that one possible way of allocating handles to newly created channels would be to just start from 0 and increase, similar to file descriptors; that also suggests that they should be unsigned.

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+        dest_buf_ptr_ptr: *mut *mut u8,
+        dest_buf_len_ptr: *mut usize,
+    ) -> u32;
+
+    /// Writes a message, i.e., `src_buf_len` bytes),from `scr_buf_ptr` into the channel with
+    /// `channel_handle`.
+    ///
+    /// Returns a status code to indicate success.
+    pub fn channel_write(channel_handle: i32, src_buf_ptr: *mut u8, src_buf_len: usize) -> u32;
```
tiziano88, line 104: I was thinking `src_buf_ptr` would be of type `*const u8`?

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+
+    /// Waits until at least one of the channels from the channel handles in the buffer at
+    /// `channel_handle_buf_ptr` has a message to read---or
+    /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
+    /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
+    ///
+    /// Returns a status code to indicate success or whether the deadline.
+    pub fn channel_wait(
+        channel_handle_buf_ptr: *mut i32,
```
tiziano88, line 113: Why mut?

tiziano88: May be worth defining a type alias for handles, to make things slightly more readable.

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+    ) -> u32;
+
+    /// Writes a message, i.e., `src_buf_len` bytes),from `scr_buf_ptr` into the channel with
+    /// `channel_handle`.
+    ///
+    /// Returns a status code to indicate success.
+    pub fn channel_write(channel_handle: i32, src_buf_ptr: *mut u8, src_buf_len: usize) -> u32;
+
+    /// Waits until at least one of the channels from the channel handles in the buffer at
+    /// `channel_handle_buf_ptr` has a message to read---or
```
tiziano88, line 107: (nit) I think `---` is not a valid Markdown syntax?

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+    /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
+    /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
```
tiziano88, line 109: So what happens if, say, one of the handles is invalid?

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +83,37 @@ extern "C" {
...
+    ///
+    /// Returns a status code to indicate success or whether the deadline.
```
tiziano88, line 111: Incomplete sentence?

tiziano88 (commented):
Thanks Maria!

oak_functions/abi/src/lib.rs
mariaschett: Yes, I agree the documentation should move to the ABI readme. However, I was hoping to keep it close to the definitions, while they are still changing and move it once they are stable. 

Do you think it's a problem to keep them here for now?

oak_functions/abi/src/lib.rs
mariaschett: done, thank you!

oak_functions/abi/src/lib.rs
mariaschett: done! thank you!

oak_functions/abi/src/lib.rs
mariaschett: This should not be mut at all, you are right!  

Also added type alias for ChannelHandle. But then: should we also add one for Status? 

A completely different approach would be to use i32 as type for every parameter/return value---because that is, what it boils down to in Wasm. Wdyt?

oak_functions/abi/src/lib.rs
mariaschett: done. I'll try to unlearn using the dashes---I hope. :)

oak_functions/abi/src/lib.rs
mariaschett: yes. should be complete now.

oak_functions/abi/src/lib.rs
mariaschett: I'd expect an invalid handle to be ignored. If all of them are invalid, then eventually the deadline_ms will pass. 

What would you expect?

oak_functions/abi/src/lib.rs
tiziano88: SG!

oak_functions/abi/src/lib.rs
mariaschett: Thank you! Reformulated to make this clearer. Is it?

oak_functions/abi/src/lib.rs
tiziano88: Personally I would expect to get an error immediately if any handle is invalid, there is never a good reason for that to happen and be silently ignored, is it? What do other similar systems do? e.g. what does Fuchsia do? Also, for reference, here is what classic Oak does: https://github.com/project-oak/oak/blob/main/docs/abi.md#wait_on_channels

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +85,41 @@ extern "C" {
...
+
+    /// Waits until at least one of the channels from the channel handles in the buffer at
+    /// `channel_handle_buf_ptr` has a message to read, or
+    /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
+    /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
+    ///
+    /// Returns a status code to indicate success or deadline expiration.
+    pub fn channel_wait(
+        channel_handle_buf_ptr: *const ChannelHandle,
+        channel_handle_buf_len: usize,
```
tiziano88, line 120: Is this the number of handles or the number of bytes in the pointed buffer?

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +85,41 @@ extern "C" {
...
+    ///
+    /// Returns a status code to indicate success. In particular, returns immediately
+    /// with an appropriate status code, if no message is available on the channel.
+    pub fn channel_read(
+        channel_handle: ChannelHandle,
+        dest_buf_ptr_ptr: *mut *mut u8,
+        dest_buf_len_ptr: *mut usize,
+    ) -> u32;
+
+    /// Writes a message, i.e., `src_buf_len` bytes),from `scr_buf_ptr` into the channel with
```
tiziano88, line 102:
```suggestion
    /// Writes a message, i.e., `src_buf_len` bytes), from `scr_buf_ptr` into the channel with
```

oak_functions/abi/src/lib.rs
tiziano88: I don't think it's correct to say that it reads `dest_buf_len_ptr` bytes; first of all, that's a pointer, second, when this function is called, it does not point to anything meaningful, definitely not to the number of bytes to read. It is only used as an output parameter for the runtime to *write* out the number of bytes read. Makes sense?

oak_functions/abi/src/lib.rs
```diff
@@ -83,4 +85,41 @@ extern "C" {
...
+    /// `channel_handle_buf_ptr` has a message to read, or
+    /// until the `deadline_ms` expires. After a successful call to `channel_wait`, the buffer at
+    /// `ready_channel_handle_buf_ptr` holds the channel handles with at least one message to read.
+    ///
+    /// Returns a status code to indicate success or deadline expiration.
+    pub fn channel_wait(
+        channel_handle_buf_ptr: *const ChannelHandle,
+        channel_handle_buf_len: usize,
+        ready_channel_handle_buf_ptr: *mut *mut i32,
+        ready_channel_handle_buf_len: *mut usize,
```
tiziano88, line 122: Same

tiziano88 (approved):

oak_functions/abi/src/lib.rs
mariaschett: Yes, I think I reformulated this already.

oak_functions/abi/src/lib.rs
mariaschett: Fixed to the latter, thank you!

tiziano88 (approved):
I would create tasks to follow up with figuring out what to do about channel_wait perhaps.

oak_functions/abi/src/lib.rs
mariaschett: I will do some research into to this and keep it undefined for now and revisit when we implement the functionality in the runtime.

PR #2433, opened 2021-12-15
Add type for channel between Wasm module and Oak Functions runtime

mariaschett:
# Checklist

      construction.


oak_functions/loader/src/server.rs
```diff
@@ -714,3 +718,31 @@ pub fn format_bytes(v: &[u8]) -> String {
         .map(|s| s.to_string())
         .unwrap_or_else(|_| format!("{:?}", v))
 }
+
+// The EndPoint of a birectional channel. Sender and Receiver are exposed.
+pub struct EndPoint {
```
tiziano88, line 723: I think `Endpoint` is a single word.

oak_functions/loader/src/server.rs
```diff
@@ -714,3 +718,31 @@ pub fn format_bytes(v: &[u8]) -> String {
...
+    pub sender: Sender<AbiMessage>,
+    pub receiver: Receiver<AbiMessage>,
+}
+
+// Create a channel with two symmetrial endpoints.  The AbiMessage sent from one EndPoint are
```
tiziano88, line 728:
```suggestion
// Create a channel with two symmetrial endpoints. The [`AbiMessage`] sent from one [`EndPoint`] are
```

oak_functions/loader/src/server.rs
```diff
@@ -714,3 +718,31 @@ pub fn format_bytes(v: &[u8]) -> String {
...
+// received at the other EndPoint and vice versa by connecting two unidirectional
+// tokio::mpsc channels.
```
tiziano88, line 730: Make this a link too perhaps.

oak_functions/loader/src/server.rs
```diff
@@ -714,3 +718,31 @@ pub fn format_bytes(v: &[u8]) -> String {
...
+//
+// In ASCII art:
+//  sender _____  ____ sender
```
tiziano88, line 733: I would put this in a code block too, so it gets rendered correctly in the documentation.

oak_functions/loader/src/server.rs
```diff
@@ -714,3 +718,31 @@ pub fn format_bytes(v: &[u8]) -> String {
...
+//              \/
+// receiver ____/\____ receiver
+pub fn channel_create() -> (EndPoint, EndPoint) {
+    let (tx0, rx0) = channel::<AbiMessage>(100);
```
tiziano88, line 737: Could you make 100 a const somewhere? Also perhaps add a comment.

oak_functions/loader/src/tests.rs
```diff
@@ -430,3 +431,37 @@ fn test_format_bytes() {
     // Incorrect UTF-8 bytes, as per https://doc.rust-lang.org/std/string/struct.String.html#examples-3.
     assert_eq!("[0, 159, 146, 150]", format_bytes(&[0, 159, 146, 150]));
 }
+
+#[test]
+fn test_start_from_empty_endpoints() {
+    let check_empty = |endpoint: &mut EndPoint| {
```
tiziano88, line 437: I don't think this needs to be a closure, it may be an inner fn (not 100% sure).

oak_functions/proto/abi.proto
```diff
@@ -21,6 +21,20 @@ package oak.functions.abi;
 option java_multiple_files = true;
 option java_package = "oak.functions.abi";
 
+// ChannelStatus returned from a read, write, or wait on a channel.
+enum ChannelStatus {
```
tiziano88, line 25: Isn't this more like an operation status, and the one below a channel status? That is definitely not a channel handle.

oak_functions/proto/abi.proto
```diff
@@ -21,6 +21,20 @@ package oak.functions.abi;
...
+  CHANNEL_UNSPECIFIED = 0;
```
tiziano88, line 26: The enum naming does not seem standard, see http://go/protodosdonts#do-include-an-unspecified-value-in-an-enum

oak_functions/proto/abi.proto
```diff
@@ -21,6 +21,20 @@ package oak.functions.abi;
...
+  // Success.
+  CHANNEL_OK = 1;
+  // Channel is empty.
+  CHANNEL_EMPTY = 2;
+}
+
+enum ChannelHandle {
+  HANDLE_UNSPECIFIED = 0;
```
tiziano88, line 34: "

mariaschett (commented):
Thank you!

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

oak_functions/loader/src/tests.rs
mariaschett: Done! Thank you!

oak_functions/loader/src/tests.rs
```diff
@@ -430,3 +431,37 @@ fn test_format_bytes() {
...
+    };
+
+    let (mut module, mut runtime) = channel_create();
+    check_empty(&mut module);
+    check_empty(&mut runtime);
+}
+
+#[tokio::test]
+async fn test_crossed_write_read() {
+    let check_crossed_write_read = async move |endpoint1: EndPoint, mut endpoint2: EndPoint| {
```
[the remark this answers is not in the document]
mariaschett: Done! Thank you!

oak_functions/proto/abi.proto
mariaschett: Done.

oak_functions/proto/abi.proto
mariaschett: Done.

oak_functions/proto/abi.proto
mariaschett: I agree---ChannelStatus is more like an operation status. Or more precise: the status code after an operation on a channel: How about ChannelStatusCode?

Which one below do you mean: the `ChannelHandle`? Do you think that one is wrongly named?

tiziano88 (approved):
Thanks Maria!

oak_functions/proto/abi.proto
mariaschett: Yes, `ChannelHandle` is the "numbering"/extension of the channel---so the first argument of `channel_read/write/wait()` in the ABI [1]. I'll add some documentation! 

I am worried that `ChannelStatus` -> `Status` looses too much information and I used `ChannelStatus` to distinguish the Status after a (low-level) channel operation from the (high-level SDK) `OakStatus`. I am trying to come up with a better proposal. 

[1]  https://github.com/project-oak/oak/blob/e293c803f6c7d373fb963aac86841934699f7982/oak_functions/abi/src/lib.rs#L108

tiziano88: Will there be a difference between low level and high level status eventually?

mariaschett: I see the difference in: the "low level" gives a status concerning the channel operations and is independent from Oak. The "high level" status is not dependent on Oak (and the Channel Handles)

PR #2435, opened 2021-12-15
scripts/git_check_diff: Check for ignored changes

jul-sh:
Both ci.yaml and couldbuild.yaml specify that this script should check for both tracked and untracked changes. 
https://github.com/project-oak/oak/blob/c33c4e80e589f315c79f2b4c7fd692abab580b2a/.github/workflows/ci.yaml#L76
https://github.com/project-oak/oak/blob/e195f5188a940efeca210921948291323a56e2fc/cloudbuild.yaml#L49

Currently this script only checks tracked files, and historically has done the same. This PR implements its behavior as outlined the aforementioned files. It means checking for all changes, whether tracked or not.

# Checklist

      construction.

### DIFF
--- a/scripts/git_check_diff
+++ b/scripts/git_check_diff
@@ -4,8 +4,15 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 # shellcheck source=scripts/common
 source "$SCRIPTS_DIR/common"
 
-# Check that any generated files match those that are checked in.
-if [[ $(git status --short) ]]; then
-    # One or more files are either modified or newly generated, exit with an error code
+# Changed files that are tracked by git.
+readonly GIT_STATUS=$(git status --short)
+# Changed files that are untracked/ignored by git.
+readonly GIT_CLEAN_STATUS=$(git clean --dry-run -x -d)
+
+if [[ "$GIT_STATUS" || "$GIT_CLEAN_STATUS" ]]; then
+    echo 'Please re-run this script from a clean commit.'
+    "$GIT_STATUS" && echo "The following tracked changes were present: $GIT_STATUS"
+    "$GIT_CLEAN_STATUS" && echo "The following untracked changes were present: $GIT_CLEAN_STATUS"
+
     exit 1
 fi
--- a/scripts/release_snapshot
+++ b/scripts/release_snapshot
@@ -17,8 +17,14 @@ source "$SCRIPTS_DIR/common"
 
 # Ensure that the working tree is clean.
 # Copied from /scripts/git_check_diff.
-if [[ $(git status --short) ]]; then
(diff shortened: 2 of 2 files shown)


tiziano88:
Not sure I understand the goal here, these files are already git-ignored, right? https://github.com/project-oak/oak/blob/e24bc764cc6a95c5f000ad3d411348a28f94d75d/.gitignore#L13 I think it is expected that they can change and still be ignored

jul-sh:
@tiziano88 Not based on the comments in the CI files linked above, which state:

```
      # Check whether any of the previous steps resulted in file diffs that were not checked in or
      # ignored by git.
```

tiziano88:
Ah got it, perhaps that should be clarified :) but I would expect we would want to detect changes to tracked files, as well as new untracked files (please feel free to reword the github action file to reflect that). Ignored files can still be ignored, unless there is some other use case for which detecting that would be useful.

tiziano88:
(the git clean output is nice BTW, maybe we can keep that for the diff we care about?)

jul-sh:
Makes sense, thanks for clarifying. It seems like for the CI usecases we do want precisely what git status does. :) Imo we can probably rm the comment to future avoid conflicts between the comments description of the code's behavior and the actual behavior. 

> Displays paths that have differences between the index file and the current HEAD commit, paths that have differences between the working tree and the index file, and paths in the working tree that are not tracked by Git (and are not ignored by gitignore[5])
Per https://git-scm.com/docs/git-status

This seems sensible to me. We just want to make sure that running the CI does not modify our source-code and that all files that should be ignored are ignored.

jul-sh:
In regards to the `scripts/release_snapshot` script: As far as I understand the intention there is different. The concern is reproducibility not, so we want to make sure the repo is as clean as possible, equivalent to a fresh clone. 

Hence, could we keep the more thorough check for that file?

rbehjati:
> CI is failing, which is expected behavior given that our CI at the moment results in unaccounted file diffs, which are now caught. We'll need to decide how to move forward here. Either we must rethink/re-articulate the goal of the git_diff script to accept the presence of untracked changes, or we delete those changes.
> 
> Note that the first option comes at the likely cost of making reproducibility harder, the second comes at the cost of CI speed (it'll require deleting caches).

Only for oak_functions, the target files that are currently ignored take 1.5GB of space. Checking them in to the repo makes the repo unnecessarily large, and will cause a lot of merge and rebase headache every time we make a change. For reproducibility, I think it is better to start with experimenting with `cargo vendor`.

jul-sh:
Thanks @rbehjati & @tiziano88 for giving more context, closing this in favor of #2450 and #2449

PR #2436, opened 2021-12-15
Cache Rust target dir

tiziano88:
In test runs, it reduced the main command execution time of the
build-functions-server step from 7m to 1m.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,6 +65,34 @@ jobs:
           ./scripts/docker_pull
           df --human-readable
 
+      # Store various common Rust folders to speed up future runs.
+      #
+      # The main cache key includes the combined hash of all Cargo.lock files in the repository, but
+      # falls back on a more generic prefix if an exact match is not found, so that there is at
+      # least some chance that some of the artifacts will be found there.
+      #
+      # We specify the `./cargo-cache` folder (as per `/scripts/docker_run`), as well as various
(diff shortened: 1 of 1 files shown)


jul-sh (approved):
LGTM, but it is worth noting that this comes with potential implication for PR CI reproducibility as changes might accumulate in the caches over the course of multiple runs. 

I'm somewhat confused wrt to our approach to reproducibility as it stands, since we appear to both run reproducibility checks on every change/PR (CI) and also have a separate script for making releases. Do we want every PR to be reproducible (as in a CI/CD model), or preserve this for batched releases? Both? Separate standards of "how reproducible" for each?

This might impact reproducibility of the former approach, whereas the releases approach would be fine.

tiziano88:
Thanks Juliette, good point. In fact I think there are two aspects.

# figure out how to periodically clean up the cache so it does not grow indefinitely

Maybe we could run the CI steps after submit without using the cache, and cache that for future pre-submit CI runs? This way we get fast CI runs when people are usually impatient waiting for it, and "from scratch" longer CI runs after it has been submitted (which usually has no surprises).

# ensure that reproducibility results are not invalidated by the extra cache

For instance, perhaps a build may appear to be reproducible in CI, but that's only because some non-determinism was already cached and carried forward. Perhaps the same build would not be reproducible if done without cache.

---

I think for both these, the answer is to figure out how to not read from cache in post-merge CI runs. I will open a separate issue to track that, since having some level of caching seems more useful in the short term.

PR #2437, opened 2021-12-16
Extract gRPC attestation client and server implementations as a reusable crate

### LINKED ISSUE #2418: Refactor Oak Functions loader attestation logic and extract reusable parts
This should be moved to a shared library where it can be reused, e.g. by the experimental TensorFlow proxy.

Based on https://github.com/project-oak/oak/pull/2408#discussion_r759225327

conradgrobler:
There is significant code duplication in the streaming gRPC implementations of the Remote Attestation protocol for the Oak Functions Loader and the experimental Tensorflow Serving Proxy. This PR extracts the common part of the bidirectional streaming gRPC implemenation of the Remote Attestation protocol into a reusable crate to remove this duplication.

Fixes #2418


experimental/tf_proxy/server/src/grpc.rs
```diff
@@ -54,3 +60,42 @@ impl BackendConnection {
...
+    );
+    Ok(wrapped_response.encode_to_vec())
+}
+
+pin_project! {
+    /// Utility wrapper to implement `Sync` for `Future`s by wrapping the inner Future in a
+    /// `SyncWrapper`.
+    struct FutureSyncWrapper<F> {
+        #[pin]
+        inner: SyncWrapper<F>,
```
tiziano88, line 91: Why are two levels of wrapping needed? Isn't SyncWrapper enough?

experimental/tf_proxy/server/src/main.rs
```diff
@@ -74,9 +76,28 @@ async fn main() -> anyhow::Result<()> {
     let address = SocketAddr::from((Ipv6Addr::UNSPECIFIED, opt.port));
     wait_for_socket(SOCKET).await?;
     let backend = grpc::BackendConnection::connect(SOCKET).await;
-    let handler = AttestationServer::create(Vec::new(), backend)?;
+    let client = backend.create_client();
+
+    // Create fake configuration info for now, as it cannot be empty for the attestation
+    // handshake.
+    // TODO(#2420): Remove once Java client can work without the configuration info.
```
tiziano88, line 83: What would we send instead of the configuration info? Some things are still necessary even from this other binary, e.g. some signature, inclusion proof, etc. cc @rbehjati

grpc_attestation/src/client.rs
```diff
@@ -0,0 +1,181 @@
...
+            .context("Couldn't connect via gRPC channel")?;
+        let mut client = StreamingSessionClient::new(channel);
+
+        let (sender, mut receiver) = tokio::sync::mpsc::channel(MESSAGE_BUFFER_SIZE);
+
+        let request_stream = async_stream::stream! {
+            while let Some(message) = receiver.recv().await {
+                yield message;
+            }
+        };
```
tiziano88, line 50: Isn't there some more functional way of getting a stream from a channel? see https://docs.rs/tokio-stream/0.1.8/tokio_stream/wrappers/struct.ReceiverStream.html

grpc_attestation/src/server.rs
```diff
@@ -0,0 +1,203 @@
...
+use tonic::{Request, Response, Status, Streaming};
+
+/// Handler for subsequent encrypted requests from the stream after the handshake is completed.
+struct EncryptedRequestHandler<F, S>
+where
+    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
+    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync,
+{
+    /// The stream of incoming requests.
+    request_stream: Streaming<StreamingRequest>,
```
tiziano88, line 36: Should this be a generic stream? (not tonic specific). And could the struct also implement stream interface? i.e. could this be expressed as a transformation between streams? e.g. with https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.map

tiziano88 (approved):
Great! Please wait for @ipetr0v 's review too

conradgrobler (commented):
Thanks

experimental/tf_proxy/server/src/grpc.rs
```diff
@@ -54,3 +60,42 @@ impl BackendConnection {
...
+
+pub async fn handle_request(
+    mut client: PredictionServiceClient<Channel>,
+    cleartext_request: Vec<u8>,
+) -> anyhow::Result<Vec<u8>> {
+    let predict_request =
+        PredictRequest::decode(&*cleartext_request).context("couldn't parse decrypted request")?;
+    // Wrap the gRPC client call to make it `Sync`.
+    let sync_predict = FutureSyncWrapper {
+        inner: SyncWrapper::new(client.predict(predict_request)),
```
ipetr0v, line 72: Could you please add more comments on what it tries to predict?

ipetr0v (approved):

experimental/tf_proxy/server/src/grpc.rs
conradgrobler: Done.

experimental/tf_proxy/server/src/main.rs
rbehjati: In future, the server will as well send a LogEntry (e.g., from Rekor) to the client. I think it is easier to have it as part of `ConfigurationInfo`, but it can also be a separate object. 

I am not sure if we want to skip sending configuration info. Perhaps, the Java client is fine but the Rust client must be updated to fail when no Configuration info is provided.

PR #2438, opened 2021-12-16
draft pr debugging build reproducibility

jul-sh:
rerunning the CI in GitHub actions on an old commit to test reproducibility


tiziano88:
> kindly ignore this draft pr

done :heavy_check_mark:

jul-sh:
I opened this to test in the same environment (GitHub actions) whether the hash the built binaries stays consistent over time. 

The source branch of this PR is equivalent to this commit https://github.com/project-oak/oak/commit/a9ad36dede15386a6a9fa98d46aeede3205e2b29 (with an added commit to force CI). When it was added to the main branch on 2021-11-03 CI ran and produced this hash:

`2cfe82f653de08fa4f25faf09b124bfe57a615a06eb5ab853bf5816bd36a7af5  ./oak_functions/loader/bin/oak_functions_loader`

When GitHub CI reran on the same code today, it produced the following hash (different):

`77af89a059b738a56093c698fa0ffa7d91ca61183bd492f7ae7761426d7e1835  ./oak_functions/loader/bin/oak_functions_loader`

When I cloned the branch locally today and ran `./scripts/docker_run ./scripts/runner build-functions-server && sha256sum ./oak_functions/loader/bin/oak_functions_loader` it yielded `b55eb3cd5e0ad07c14ebd46d193406f75246554bdc4faae0a219b0552189dc2d`.

This suggests that hashes are not stable over time. **However**, the code run here does not included https://github.com/project-oak/oak/commit/d8c0c34b27cb1172c0be01b39642aeab749e7c87 (committed 2021-11-08 to main), which pinned the docker image to a specific hash. The built was probably run with a different docker image, which could explain this.

We should probably retest this experiment with the aforementioned commit (https://github.com/project-oak/oak/commit/d8c0c34b27cb1172c0be01b39642aeab749e7c87) to see if this issue persists then.

jul-sh:
Results are in for building https://github.com/project-oak/oak/commit/d8c0c34b27cb1172c0be01b39642aeab749e7c87!

### On 2021-11-08:

GitHub CI built it and returned the hash `2cfe82f653de08fa4f25faf09b124bfe57a615a06eb5ab853bf5816bd36a7af5` for `./oak_functions/loader/bin/oak_functions_loader`. See the relevant run: https://github.com/project-oak/oak/runs/4142402056?check_suite_focus=true

### Today (2021-12-16):

I reran it by force pushing the commit to the source branch of this PR (with one new commit to force CI https://github.com/project-oak/oak/pull/2438/commits/918163deef90ad8a637e3d3751e3a775e9eae714). GitHub built it and returned the hash: `77af89a059b738a56093c698fa0ffa7d91ca61183bd492f7ae7761426d7e1835 ` for `./oak_functions/loader/bin/oak_functions_loader`.

Kokoro built the source branch as well, and yielded `b55eb3cd5e0ad07c14ebd46d193406f75246554bdc4faae0a219b0552189dc2d`  for `./oak_functions/loader/bin/oak_functions_loader`.

Also I cloned the branch locally and ran `./scripts/docker_pull && ./scripts/docker_run ./scripts/runner build-functions-server && sha256sum ./oak_functions/loader/bin/oak_functions_loader`. It printed the hash: `b55eb3cd5e0ad07c14ebd46d193406f75246554bdc4faae0a219b0552189dc2d`.

jul-sh:
Not sure what is happening yet, but I am fairly suspicious of the hash that GitHub actions are printing. 

Specifically, the hash `77af89a059b738a56093c698fa0ffa7d91ca61183bd492f7ae7761426d7e1835` is produced both for this PR (source branch reflects the state of main on 2021-11-08) and also for the last commit in main (today). This seems rather odd. 

See the run of the latest commit in main: https://github.com/project-oak/oak/runs/4551336210?check_suite_focus=true#step:14:16

jul-sh:
Wrt to the spooky hash described above, I found out why this is the case. Unlike what we previously thought, GitHub actions does not run against the source branch. Instead it runs against a hypothetical merge of the source branch onto main. This explains the why the hash of main, and of this PR are identical in said context.

See https://github.com/project-oak/oak/runs/4551336347?check_suite_focus=true#step:2:40
This is explained further here: https://github.community/t/questions-about-pr-workflows-and-actions-checkout-v2/122347/2

Note that this is different from kokoro, which runs against the source branch.

jul-sh:
Reran GitHub Actions on just d8c0c34b27cb1172c0be01b39642aeab749e7c87 in the main branch in the main branch on my fork: https://github.com/jul-sh/oak/runs/4583799163?check_suite_focus=true#step:13:46

This replicates the same run as

> On 2021-11-08:
> GitHub CI built it and returned the hash 2cfe82f653de08fa4f25faf09b124bfe57a615a06eb5ab853bf5816bd36a7af5 for ./oak_functions/loader/bin/oak_functions_loader. See the relevant run: https://github.com/project-oak/oak/runs/4142402056?check_suite_focus=true

GitHub CI produced the hash `724dceef5d1b22349e0e0bcf914d122675ce2e24008ecf1a6cbc103e0303e627`, see the run here: https://github.com/jul-sh/oak/runs/4583799163?check_suite_focus=true#step:13:46

This is still a different hash from the original run, and also different from kororo and local runs, which consistently give `b55eb3cd5e0ad07c14ebd46d193406f75246554bdc4faae0a219b0552189dc2d`.

I have also traced down the source of why GitHub Action runs are different: they build the server with the `--server-variant=unsafe` flag, whereas kororo and my local builds used no flags. 

I then build the binary locally with the `--server-variant=unsafe` flag, giving me the hash `64882610cbcf65888d55e24f23579f83264e162b1ac30132a227077c6bf33df7`.

I'm fairly certain that `--server-variant=unsafe` is the condition that triggers a variable hash that is hard to reproduce. 

Next Steps:

1. Consider whether the use of `--server-variant=unsafe` unsafe is a good idea for the reproducibility index. That the base variant is the version we care about, as that is what would be used in prod. 
2. Trace down what specifically about the unsafe variant causes a variable hash.

rbehjati:
Thanks Juliette. Good catch! I think the base image is the one we want to track in the reproducibility index, and `--server-variant=unsafe` should be removed.

> I'm fairly certain that --server-variant=unsafe is the condition that triggers a variable hash that is hard to reproduce. So far I have got three different hashes when building the same commit (d8c0c34).

Two of the hashes use `--server-variant=unsafe`, one on the CI and one when built locally. Did I understand that correctly?

jul-sh:
> Two of the hashes use --server-variant=unsafe, one on the CI and one when built locally. Did I understand that correctly?

@rbehjati roughly yes, though that's just my experiments. The take-away is that the CI always uses that flag as of right now.

jul-sh:
I also did some more digging by comparing the build artifacts of two builds with `--server-variant=unsafe`: https://github.com/jul-sh/oak/pull/47. Both were built with the docker image, but on local different machines. They produced different binaries with different hashes.

Based on a comparison of the build artifacts, it looks as if `tract-tensorflow` (tensorflow) is likely responsible for introducing the variance. 

I noticed we guard tensorflow behind the "oak-unsafe" flag in functions (see #2247). I'm a bit confused by this, as the comment in Cargo.toml describes the  "oak-unsafe" flag as being about logging and debugging only.

https://github.com/project-oak/oak/blob/a74343333a105dba25c273ee045f89a7374b2d77/oak_functions/loader/Cargo.toml#L9
https://github.com/project-oak/oak/blob/a74343333a105dba25c273ee045f89a7374b2d77/oak_functions/loader/Cargo.toml#L10

rbehjati:
> I noticed we guard tensorflow behind the "oak-unsafe" flag in functions (see #2247). I'm a bit confused by this, as the comment in Cargo.toml describes the "oak-unsafe" flag as being about logging and debugging only.

The comments are outdated. Created #2446 to update the comments.

jul-sh:
Thanks, that's a good idea. 

Out of curiosity, I tried vendoring deps with `./scripts/release_snapshot` to see if it would make builds with `--server-variant=unsafe` reproducible. It does not. I still get different hashes on different machines.

rbehjati:
> Out of curiosity, I tried vendoring deps with `./scripts/release_snapshot` to see if it would make builds with `--server-variant=unsafe` reproducible. It does not. I still get different hashes on different machines.

I don't know how exactly `cargo vendor` works and how it affects `cargo build`, but in `./scripts/release_snapshot`, `cargo vendor` is run without the `--features` flag. So when running `cargo build` with the `--features` flag, I'd expect something to be missing from the vendored dependencies. In particular, I'd expect `./scripts/release_snapshot` not to vendore `tract-tensorflow` dependencies, if that is the source of non-reproducibility.

jul-sh:
In my testing it did vendor it. :) However it looks like the inconsistency at the build step, not when downloading deps from the network.

rbehjati:
> In my testing it did vendor it. :) However it looks like the inconsistency occurs at the build step, not when downloading deps from the network.

Thanks. This is very helpful. Just to clarify, by the build step do you mean when running some of the `build.rs` files?

jul-sh:
> Thanks. This is very helpful. Just to clarify, by the build step do you mean when running some of the build.rs files?
Yup!

jul-sh:
I identified and patched the root cause of the inconsistent builds described above, see https://github.com/sonos/tract/pull/601 for details. Hence closing this, I'll update the repo with the new tract version later, and ideally do a small write up of the debugging process for future reference. People might be helpful when debugging similar inconsistent builds again in the future.

PR #2440, opened 2021-12-16
Re-enable clippy warnings

tiziano88:
Also fix all of the remaining ones.

I think this was accidentally disabled in 183407a71b30a670f645525fd698ba1d2306a5f1.


mariaschett (approved):
Yes, I seem to have accidentially deleted `--deny=warnings`. Thank you for catching that---and fixing my lint errors!

PR #2442, opened 2021-12-16
Cancel concurrent jobs via concurrency setting

tiziano88:
See
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run.
This seems to be a better approach than the current
https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks.


mariaschett (approved):
That's great! Thank you!

PR #2443, opened 2021-12-16
Remove PR template checklist

tiziano88:
I think most people ignore it these days, so if we want it to work we
should probably rethink the approach in the future.


mariaschett (approved):
This is a really good idea---the checklist was confusing me. Thank you!

Issue #2050, opened 2021-04-30
Inconsistent view of in-memory lookup data while a Wasm function executes


tiziano88:
Because of the way the Oak Functions loader refreshes data in the background, it might happen that if a Wasm module performs multiple lookups, a new version of the lookup data might become available between lookups from the same invocation, which may result in inconsistent behaviour of the module.

We could solve this by acquiring a Read lock on the data at invocation time (or the first time the module invokes the lookup ABI function), and hold it until the invocation is terminated (instead of acquiring it and releasing it as part of each lookup ABI call).

When the loader needs to update the data from the background process, it would still acquire a write lock for it, which would be guaranteed to be exclusive. This is fine because this write lock would be held for a very short time (just to swap the reference behind it), so there should not be too much contention.

I am not sure what priority the read and write portion of the RWLock would get though, and whether it is possible that the background process would starve if there is a constant stream of requests that keep requesting the read lock, or if the write lock has higher priority (and with what characteristics).

conradgrobler:
Another option might be to give each Wasm state instance a new cloned Arc pointing to the hashtable for its lookups. When the data is refreshed, it can create a new Arc to a new hashtable which would be cloned for new incoming requests. The existing requests would still reference the previous hashtable, but once all of them have completed the previous hash table will be dropped. This might mean that there are two distinct hash tables in memory for a short time, but this should be acceptable for most scenarios.

tiziano88:
@anghelcovici interested in taking a look? assigning to you for the time being so we don't duplicate work, feel free to un-assign if you don't have time to look into it

anghelcovici:
Sure, thanks for assigning.

mariaschett:
@conradgrobler @tiziano88 I added a (failing) test in https://github.com/project-oak/oak/pull/2425/files. 

Do you think the test has the behaviour we are looking for, i.e., during the lifetime of a WasmModule the lookup data remains unchanged?

conradgrobler:
Yes, I think so. It is not 100% representative of how the runtime works in practice, but I think it replicates the behaviour well enough to be used as a test to validate that a fix provides the expected result.

mariaschett:
Clarification: do we want the lookup data to be consistent during (a) the lifetime of a WasmHandler (as in the test atm), or (b) during a call to `handle_invoke()`? 

[...]

PR #2444, opened 2021-12-17
Make lookup an extension

mariaschett:
# Checklist

      construction.


oak_functions/loader/src/lookup.rs
```diff
@@ -14,247 +14,140 @@
...
-    }
-
-    #[allow(dead_code)]
-    pub fn len(&self) -> usize {
-        self.entries
-            .read()
-            .expect("could not lock entries for read")
-            .len()
+    /// Create and return a lookup factory.
+    pub async fn create(
```
rbehjati, line 51: This function does not need to be async. Does it?

mariaschett: You are right, it does not. 

I made it async because I wanted it to have the same signature as the corresponding functions in tf / metric, and those need to be async. 

Do you think I should change it?

rbehjati: Are you referring to `create_tensorflow_factory` and `create_metrics_proxy_factory`? 

I think in general we should avoid unnecessary asyncs. Because that complicates the code as it results in unnecessary awaits and blocking calls on tokio runtime.

Also I think `create_metrics_proxy_factory` does not have to be async. Perhaps fix that too?

mariaschett: Yes, I meant `create_tensorflow_factory` and `create_metrics_proxy_factory`. 

Perfect, so I will remove the async from this `Lookup::create` and in a seperate PR remove the async from `create_metrics_proxy_factory` too.

SG?

rbehjati (approved):
Very nice. Thanks :)

oak_functions/loader/src/lookup.rs
rbehjati: Yes. I agree that `create_metrics_proxy_factory` can be fixed in a separate PR.

oak_functions/loader/src/lookup_data.rs
```diff
@@ -0,0 +1,260 @@
...
+///
+/// Entries in the data file path must be consecutive binary encoded and length delimited
+/// protobuf messages according to the definition in `/oak_functions/proto/lookup_data.proto`.
+pub struct LookupData {
+    lookup_data_source: Option<LookupDataSource>,
+    entries: RwLock<HashMap<Vec<u8>, Vec<u8>>>,
+    logger: Logger,
+}
+
+impl LookupData {
```
tiziano88, line 55: Isn't this a bit backwards now? Or at least overlapping between the factory and this data. The initial idea of this data struct was that it would refresh itself from upstream; but now, I think it is the factory that should refresh its internal state of the data (perhaps behind an Arc), and provide that to the instances it creates. I think currently there are too many layers :)

tiziano88 (approved):

oak_functions/loader/src/lookup.rs
```diff
@@ -14,247 +14,140 @@
...
-    }
-
-    #[allow(dead_code)]
-    pub fn len(&self) -> usize {
-        self.entries
-            .read()
-            .expect("could not lock entries for read")
-            .len()
+    /// Create and return a lookup factory.
+    pub fn create(
```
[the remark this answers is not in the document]
mariaschett: I had the same thought! 

I'll refactor that in the PR where I refactor this together wit `create_tensorflow_factory` and `create_proxy_metrics_factory`.

oak_functions/loader/src/lookup_data.rs
mariaschett: I think it is still the case that the lookup data will refresh itself. 

However, we will have to think that through, as the lookup data should stay consistent throughout a lookup---especially, when we start thinking about iterators, cf.  #2050.

PR #2446, opened 2021-12-20
Update the comments about oak-unsafe in oak_functions

rbehjati:


jul-sh:
Thanks! Let's also update the relevant parts in `runner/src/internal.rs`

rbehjati:
> Thanks! Let's also update the relevant parts in `runner/src/internal.rs`

Thanks. Done.

oak_functions/loader/Cargo.toml
```diff
@@ -6,8 +6,8 @@ edition = "2018"
 license = "Apache-2.0"
 
 [features]
-# Feature that allows logging of potentially sensitive content.
-# Should only be used for debugging purposes.
+# Feature that allows logging of potentially sensitive content, or using experimental features.
+# Should only be used for debugging purposes or enabling experimental features.
```
jul-sh, line 10: imo we can probs remove at that line at this point :)

jul-sh (approved):

oak_functions/loader/Cargo.toml
rbehjati: Do you mean that the entire comment about this feature can be removed?

PR #2448, opened 2021-12-22
Fix gRPC UNAVAILABLE due to channel shutdownNow invoked

ipetr0v:
This change fixes the following warning received after a gRPC channel is closed:
```
WARNING: Couldn't receive response: Status{code=UNAVAILABLE, description=Channel shutdownNow invoked, cause=null}
```

### DIFF
--- a/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
+++ b/oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
@@ -203,7 +203,6 @@ public class AttestationClient {
   @Override
   public void finalize() {
     requestObserver.onCompleted();
-    channel.shutdownNow();
   }
 
   /**


oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -203,7 +203,6 @@ boolean verifyServerIdentity(byte[] serializedServerIdentity,
   @Override
   public void finalize() {
     requestObserver.onCompleted();
-    channel.shutdownNow();
```
rbehjati, line 206: Does the error occur because the channel is already closed at this point?

rbehjati (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
ipetr0v: Yes

PR #2449, opened 2021-12-23
Release snapshot: Ensure the working tree is free of ignored files

jul-sh:
See #2435 for context. We want to the working tree to reflect the committed changes, hence we want to make sure there aren't ignored changes present, as they can influence the build. (Eg a modified `/target/` dir.)


jul-sh:
Closing this for now, we can iterate on this further once the usage of this script is more clearly defined.

PR #2453, opened 2021-12-29
Add clippy --all-features flag to check non-default crates

mariaschett:


oak_client/src/attestation.rs
```diff
@@ -84,7 +84,7 @@ pub fn create_tls_attestation_config(
     config.set_protocols(&[b"http/1.1".to_vec(), b"h2".to_vec()]);
 
     // Add root TLS certificate.
-    let mut cc_reader = std::io::BufReader::new(&root_tls_certificate[..]);
+    let mut cc_reader = std::io::BufReader::new(root_tls_certificate);
```
rbehjati, line 87: Out of curiosity, what was the linting error for this?

oak_loader/src/attestation.rs
```diff
@@ -50,7 +50,7 @@ pub async fn get_tls_identity_from_proxy(
         "Sending certificate signing request to Proxy Attestation Service: {:?}",
         uri
     );
-    let mut client = create_proxy_client(&uri, &root_tls_certificate)
+    let mut client = create_proxy_client(uri, root_tls_certificate)
```
rbehjati, line 53: What was the error in this case? (if you remember :))

rbehjati (approved):
Nice. Thanks :)

oak_client/src/attestation.rs
mariaschett: The error was:

use the original value instead: `root_tls_certificate` 
= note: `-D clippy::redundant-slicing` implied by `-D warnings`

oak_loader/src/attestation.rs
mariaschett: The error was: "this expression borrows a reference (`&http::Uri` / `&[u8]`) that is immediately dereferenced by the compiler"

Issue #2455, opened 2021-12-29
Consider running CI on PRs only for modified files


rbehjati:
Extends #2168 
See also #2403

To have faster CI on PRs we'd want to run the CI steps only for the modified files and the files affected by those modifications. We will still run the full CI on main, after merging a PR, to prevent untouched parts of the code base from rotting over time.

rbehjati:
The main problem is that currently we checkout the repo with `--depth=1`, so it is not possible to get the list of modified files on GitHub. [Fetching the full history](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches) with the following should fix this: 
```
- uses: actions/checkout@v2
  with:
    fetch-depth: 0
```

With this, it should be possible to use `git diff --name-only main` to get the list of all modified files in the branch.

mariaschett:
To notify the team if the CI failed when merging on main, I see two options:

1. Use the github notifiaction mechanism and [add a notification on failure](https://stackoverflow.com/questions/62304258/github-actions-notifications-on-workflow-failure)

2. [Send an email notifiaction](https://ravgeetdhillon.medium.com/send-an-email-notification-when-github-actions-fails-ea83cbeabbe0).

Do we already have an idea, how we seperate the runs for merging on main/merging PRs?

rbehjati:
> Do we already have an idea, how we seperate the runs for merging on main/merging PRs?

We now have separate conditional steps for `pull-request` and `push` events: https://github.com/project-oak/oak/blob/b29c48a479d92c433e81a7c440174f771f8c65c7/.github/workflows/ci.yaml#L105-L115

rbehjati:
For now we have decided that we don't need email notification since if the CI fails on main, the author and the subscribers will be notified. Closing it now, but we can get back to the email notification later, if we feel the need for it.

PR #2456, opened 2021-12-29
Fetch full repo history for running the CI on GitHub

rbehjati:
Ref #2455

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,8 @@ jobs:
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       # The runner comes with all this software pre-installed: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
       # so we delete some large packages to make sure we have more space available.


tiziano88 (commented):
Isn't depth 1 sufficient?

rbehjati:
> Isn't depth 1 sufficient?

If we want to implement #2455, depth 1 is not enough. Or I cannot think of a way to get the modified files in a branch if we have only depth 1. 

I was worried fetching the entire history would take very long, but in the CI steps for this PR it is 8-17 seconds, which I think is not a big overhead.

tiziano88:
AFAICT depth 1 is exactly what we need, i.e. the latest commit? Or maybe depth 2, perhaps I'm off by one :)

rbehjati:
How is it possible to get the list of all modified files from only the latest commit? 
What if there are more than one commit in a pull request?

tiziano88:
Ah got it, I was assuming 1PR == 1commit, and I thought your point was that with depth=1 even the latest commit was not accessible (hence my suggestion to try with depth=2). But you are right of course, in general that's not the case. I also looked at https://docs.github.com/en/actions/learn-github-actions/contexts#github-context to see if the number of commits is available in some variable, but it does not seem to be the case. Anyways, perhaps setting depth=50 or something like that should at least limit the amount of commits to pull in especially going forward, if the history keeps growing and if e.g. we decide to vendor more crates, in which case it will grow much faster.

rbehjati:
> Anyways, perhaps setting depth=50 or something like that should at least limit the amount of commits to pull in especially going forward, if the history keeps growing and if e.g. we decide to vendor more crates, in which case it will grow much faster.

Currently, for local use we are using the `--commits=N` flag, where we manually specify the number of commits (N). If we want to use the same flag on GitHub and set it to a constant, for instance `--commits=50` then the builds become very non-deterministic. In most cases the 50 commits would include all the changes in 40+ earlier PR merges, which would make this whole work pointless, because there will be a high chance of including a commit that changes a `.lock` file in classic oak (for instance in the PRs from dependabot), so all the steps will have to run for most crates. I think we need a way to be able to compare the changes in the branch with `main`. 

I understand your concern with the repo growing, but I don't think fetching a specific depth alone would solve the problem. I wonder if it is possible to add a step to manually fetch main, with a specified depth, so that we could run `git diff --name-only --merge-base main`. 

Anyway, based on the speedup in [this experiment](https://github.com/project-oak/oak/runs/4668299195?check_suite_focus=true), I think it is worth starting with fetching the entire repo, and later we can optimize this part.

rbehjati:
Subsumed in #2469.

PR #2458, opened 2021-12-30
Seperate steps in `run_cargo_tests`

mariaschett:
Add seperate command `run_cargo_clippy`. Now we run clippy only on directly modified crates (as opposed to all affected crates).


rbehjati (approved):
Nice. Thanks :)

PR #2461, opened 2021-12-31
Add test to check whether `.runner_bash_completion` is up-to-date

mariaschett:


runner/src/main.rs
```diff
@@ -390,6 +382,19 @@ fn check_format(commits: &Commits) -> Step {
     }
 }
 
+fn run_completion(completion: &Completion) -> Step {
+    let mut out_dir = completion.out_dir.clone();
+    out_dir.push(".runner_bash_completion");
```
tiziano88, line 387: (pre-existing) I don't remember why this command takes an output dir and then creates a file with a fixed name within it, instead of taking the full expected file name? It seems that currently, half the knowledge of where the file will end up is in this binary, and half is in the caller script? Though maybe that would make the CI step more complicated?

tiziano88 (approved):
Thanks, the git check diff step is what would actually report any errors, right?

mariaschett:
> Thanks, the git check diff step is what would actually report any errors, right?

Yes, I even did a trial run with a commit to `Manually change .runner_bash_completion so CI fails` -- it failed :)

runner/src/main.rs
mariaschett: Good point---I don't remember there being a reason for this split, so I merged. Thank you!

PR #2462, opened 2021-12-31
Build the base server variant for the reproducibility index

jul-sh:
This change was suggested in my debugging reproducibility thread: https://github.com/project-oak/oak/pull/2438#issuecomment-998160165. It also seems sensible, given that the unsafe flag is intended for experimental purposes


tiziano88 (approved):

rbehjati (approved):
Thanks :)

PR #2466, opened 2022-01-07
Create channel in WasmState

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -680,3 +704,62 @@ pub fn channel_create() -> (Endpoint, Endpoint) {
     };
     (endpoint0, endpoint1)
 }
+
+pub struct ChannelSwitchboard(HashMap<ChannelHandle, Arc<Endpoint>>);
+
+impl ChannelSwitchboard {
+    pub fn create(handles: Vec<ChannelHandle>) -> (Self, Self) {
```
tiziano88, line 711: I don't think the runtime should ever even refer to handles at all. It can just access the Endpoint objects directly. Handles are just an artifact of Wasm not having more precise types to refer to the endpoints, so it must use ints for that, and we have to map them to the actual Endpoint, but the runtime has the full expressivity of the Rust type system, so we should take advantage of that. The switchboard would only be needed to translate from handle to the corresponding endpoint, when the wasm module invokes one of channel_write, channel_read or channel_wait. WDYT?

mariaschett: But how would the runtime know into which endpoint to write? 

So let's say the runtime wants to write the request to the "Client" channel for the Wasm module the runtime has to know which endpoint to write to---no? 

Or, how else would we store the different corresponding endpoints in the runtime?

tiziano88: Perhaps a Rust struct with appropriate types and field names:

```rust
struct RuntimeEndpoints {
  client: Endpoint,
  lookup: Endpoint,
  tf: Endpoint,
}
```

in practice I don't even think that this is necessary long term, as the extensions may just own their own side of the endpoint directly, but we'll see when we get there

tiziano88: Instead of doing `switchboard.get(CLIENT)`, the runtime can just do `switchboard.client`. Since the goal of the type system is to make invalid states unrepresentable, the struct has the advantage that you don't need to deal with e.g. `switchboard.get(999999)`, which is an invalid state. Also `.get` would return an `Option`, but the runtime should statically know all the endpoints it needs, hence there is another inconsistency there. And it should always statically know where the endpoint is in memory, it should have no need to figure it out from a hash map.

mariaschett: I absolutely agree, we want to eventually move the endpoints of the runtime to the extensions.

But I am not very sure how much the above struct gives me at the moment.  For this step I am happy with having two (intertwined) switchboards: one for the runtime and one for the module. The hashmap also [aligns with what we have for extension atm](https://github.com/project-oak/oak/blob/e651aac60f15b24311adcbbb8d2a6a33b0b6a8de/oak_functions/loader/src/server.rs#L154). 

I am sure I will want to change the representation of switchboard for the runtime, but I planned to do this, when I implement `channel_read/write` for the runtime. WDYT?

tiziano88: The struct represents the fact that this information is statically known to the Runtime, which should never have to deal with handles for its own internals (only when processing a call from a Wasm module). The hashmap represents the fact that this informations is *not* statically known, which is the case for Wasm modules, and the runtime needs to perform a lookup from handle to endpoint every time, because of this. Which is represented in the algebraic representation of the endpoints: in the map, the outcome of a lookup is Option<Endpoint>, but from the runtime, it should be just Endpoint. And, relatedly, I don't think there has to be a runtime-specific `channel_read/write`. The runtime should just be able to use the Endpoint objects directly, and asynchronously read and write from the underlying mpsc channels, by invoking the relevant methods on those. Or are you thinking of a different approach?

tiziano88 (approved):

oak_functions/loader/src/server.rs
mariaschett: Great, so I added the proposed struct for the RuntimeEndpoints! It'll likely change when we rework how Extensions and Channels work together. (I believe we'd want at ChannelHandle & Endpoint attached to an Extension.)

And, yes, there doesn't have to be a runtime-specific `channel_read/write`---except maybe thin wrappers.

oak_functions/loader/src/server.rs
```diff
@@ -680,3 +695,67 @@ pub fn channel_create() -> (Endpoint, Endpoint) {
     };
     (endpoint0, endpoint1)
 }
+
+#[allow(dead_code)]
+struct RuntimeEndpoints {
+    lookup: Endpoint,
+}
+struct ChannelSwitchboard(HashMap<ChannelHandle, Arc<Endpoint>>);
```
tiziano88, line 703: BTW does the endpoint really need to be in an Arc? It seems that the hashmap can just own the endpoint outright. Anyways we can iterate on it, but I'm curious if there is already a use case that requires Arcs. Maybe channel_wait?

oak_functions/loader/src/server.rs
```diff
@@ -564,14 +570,23 @@ impl WasmHandler {
             extensions_metadata.insert(name, (ind + EXTENSION_INDEX_OFFSET, signature));
         }
 
-        let request_bytes = request.body;
-        let mut wasm_state = WasmState::new(
+        // keep _endpoints for runtime, give the correspondng switchboard to the WasmHandler
```
tiziano88, line 573: (nit) We usually start comments with capital letter and end with full stop.

tiziano88 (approved):
Thanks!

oak_functions/loader/src/server.rs
mariaschett: Done, thanks. :)

oak_functions/loader/src/server.rs
mariaschett: That is right, `Arc` is a relic here. I removed it.

PR #2467, opened 2022-01-07
Update tract-tensorflow to finalize reproducibility fix

jul-sh:
tract-tensorflow had another issue that we patched upstream via https://github.com/sonos/tract/pull/607

The unsafe variant of oak_functions now produces a binary of the same hash across machines, thereby solving the issue reported in b/209639154.

```
> git clean -d --force --force -x && ./scripts/docker_run ./scripts/runner build-functions-server --server-variant=unsafe && sha256sum ./oak_functions/loader/bin/oak_functions_loader

a296d50bdb301d8b7a2421c0d91fabb100bce4ec647becd93a31f4d23b8959e7  ./oak_functions/loader/bin/oak_functions_loader
```

### DIFF
--- a/oak_functions/Cargo.lock
+++ b/oak_functions/Cargo.lock
@@ -2832,7 +2832,7 @@ dependencies = [
 [[package]]
 name = "tract-core"
 version = "0.15.9-pre"
-source = "git+https://github.com/sonos/tract.git?rev=95afd8d834c1e5ea23bd8b5853476f46d1a219ca#95afd8d834c1e5ea23bd8b5853476f46d1a219ca"
+source = "git+https://github.com/sonos/tract.git?rev=124dc4c8d31f01ae73a2ec3cc02c0cf3db7a13d4#124dc4c8d31f01ae73a2ec3cc02c0cf3db7a13d4"
 dependencies = [
  "anyhow",
  "bit-set",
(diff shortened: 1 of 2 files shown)


tiziano88 (approved):
Wow, amazing! Thanks for fixing it upstream!

rbehjati (approved):
Nice. Thanks :)

rbehjati:
Just a side note: I am not sure this fixes the problem in b/209639154, because the build script used in the example reported in that issue does not use `--server-variant=unsafe`, and therefore the binary should not include `tract`. However, the Git commit used in that  example is before pinning the docker image to a specific hash (#2375). I think it might still be interesting to try and replicate that example with a newer git commit, at some point.

jul-sh:
@rbehjati You're right there's serval aspects to that original issue, which I have investigated and fixed separately. I'm unable to produce an issue now with a newer commit.

The original bug named three hashes, one with a local build (without --server-variant=unsafe), another for a local build without (--server-variant=unsafe) but with vendoring, and third hash for the CI build (that one did use --server-variant=unsafe).

The two local builds are different from each other because that's expected with vendoring. 
The CI build produced a different hash from the local builds as it built (`--server-variant=unsafe`) a different target. 

Does this make sense? :)

jul-sh:
On a higher level, before b/209639154 there was uncertainty and disagreement whether our builds were reproducible or not. In it's resolution we were able to falsify the statement of "builds are reproducible" (see https://github.com/project-oak/oak/pull/2438). 

We were also able to remedy that issue (this PR), such that now we are no longer able to falsify the claim of reproducibility.

rbehjati:
> I'm unable to produce an issue now with a newer commit.

That is great. I am still a bit worried about the dimension of time. Because my original concern was that something was changing in the dependencies over time, causing the non-determinism. But this is a separate issue that we can come back to if it shows up again.

jul-sh:
> > I'm unable to produce an issue now with a newer commit.
> 
> That is great. I am still a bit worried about the dimension of time. Because my original concern was that something was changing in the dependencies over time, causing the non-determinism. But this is a separate issue that we can come back to if it shows up again.

That's a very good point and that might show up. Would you be interested in keeping an eye on that? I would be happy to help debug once/if we identify a build that stops being deterministic with time.

rbehjati:
> That's a very good point and that might show up. Would you be interested in keeping an eye on that? I would be happy to help debug once/if we identify a build that stops being deterministic with time.

Sure. I think with the tooling that we are planning to add internally, it will be easier to detect if there is a violation of reproducibility.

jul-sh:
Agreed :)

PR #2468, opened 2022-01-07
Add default-target in Config.toml

mariaschett:


oak_functions/examples/metrics/module/Cargo.toml
```diff
@@ -1,9 +1,12 @@
+cargo-features = ["per-package-target"]
+
 [package]
 name = "metrics"
 version = "0.1.0"
 authors = ["Conrad Grobler <grobler@google.com>"]
 edition = "2018"
 license = "Apache-2.0"
+default-target = "wasm32-unknown-unknown"
```
tiziano88, line 9: This should probably be a forced-target?

oak_functions/examples/metrics/module/.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[build]
+target-dir = "/workspace/oak_functions/target/wasm"
+out-dir = "/workspace/oak_functions/bin"
```
tiziano88, line 3: Maybe those end up being the default?

mariaschett: So I was not sure why we set `target-dir` for the examples and not simply write in the default directory. Do you know? 

I am also not 100% sure why we need `out-dir`.

oak_functions/examples/metrics/module/Cargo.toml
mariaschett: I would keep default-target. Or do do you see any reason to prohibit this to be overwritten by `cargo --target`?

tiziano88: I think even in the cargo documentation this is the canonical example of when to use forced-target. A Wasm target is really meant to always be compiled to wasm. While other targets may be compiled e.g. for x86 or arm or whatever the host needs (hence it's expected they can be overriden).

oak_functions/examples/metrics/module/.cargo/config.toml
tiziano88: I think it was an attempt to know upfront where the target files end up, since otherwise that depends on whether the crate is in a workspace, etc. But I think now we can actually parse the actual target folder via cargo metadata, so perhaps both target dir and out dir are not necessary any more? Basically, if things work without them (i.e. all tests and examples pass), then it's fine to remove them, they serve no other purpose (that I remember).

oak_functions/examples/metrics/module/Cargo.toml
tiziano88: Sure, I would not worry too much about that, I don't think anyone is compiling natively for mac at the moment.

oak_functions/examples/metrics/module/.cargo/config.toml
tiziano88: Definitely!

mariaschett:
Open #2472 to investigate use of `outdir` and `target-dir`.

oak_functions/examples/metrics/module/.cargo/config.toml
mariaschett: #2472

oak_functions/examples/metrics/module/Cargo.toml
mariaschett: So far I observed the following: 

If we force the `wasm32-unknown-unknown` target for e.g., key value lookup in its [`Cargo.toml`](https://github.com/project-oak/oak/blob/ccad8048c4cd6ebfc6bea04a6aa46dacc5abd16a/oak_functions/examples/key_value_lookup/module/Cargo.toml), for the corresponding (integration) [test](https://github.com/project-oak/oak/blob/ccad8048c4cd6ebfc6bea04a6aa46dacc5abd16a/oak_functions/examples/key_value_lookup/module/src/tests.rs) the call to `cargo test` will fail (as we try to compile the test with the `wasm` target).

I can think of two mitigations:
1. Move the test, e.g., as integration test to loader, so it gets a different `Cargo.toml`.
2. Don't force the target, but only set a `default-target` and with `cargo test` we can overwrite the target to the host target.

Can you see a different option?

mariaschett:
Setting `forced-target = "wasm32-unknown-unknown"` in the `Cargo.toml` is problematic when running `cargo test`. We can think of revisiting setting the target after fixing  #2598 .

PR #2469, opened 2022-01-10
Run the Rust-related CI steps only for the crates affected by the changes in the PR

### LINKED ISSUE #2490: Find a more efficient alternative to cloning the entire repo in the CI workflow
See #2456 for more context.

rbehjati:
Adds a top-level `scope` flag to all commands. It takes three values: `--scope=all`, `--scope=diff_to_main`, `--scope=commits:<N>`. The last one is the same as `--commits=<N>` that some commands had previously.

The default is `--scope=diff_to_main`, which finds all the files that were modified in this branch compared to the main branch, and uses that to find all the crates that are affected by these modifications. The commands are only run for the affected crates. 

Updates the `.github/workflows/ci.yaml` to find the number of commits in the PR. In the `checkout branch` step fetches all those commits plus an extra commit to be able to find the diff (`diff-to-main` does not seem to work on the CI). For `pull_request` events, the CI steps are run with the `--scope=commits:<N>` flag and for the `push` events, they are run with the `--scope=all` flag.  

Ref #2455
Fixes #2490


.github/workflows/ci.yaml
```diff
@@ -47,6 +45,8 @@ jobs:
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
tiziano88, line 49: Could you add a comment on what this means and why we need it?

.github/workflows/cimain.yaml
```diff
@@ -0,0 +1,189 @@
+name: Continuous Integration
```
tiziano88, line 1: I think maintaining two almost-identical files will not very practical. Is it not possible to have a different list of commands based on the branch or some other variable?

.github/workflows/cimain.yaml
```diff
@@ -0,0 +1,189 @@
...
+    runs-on: ubuntu-20.04
+    timeout-minutes: 120
+    # We use the same job template, but parametrize the actual command to be passed to the runner
+    # binary using the matrix strategy, so that we get the commands running in parallel.
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          # Do not add `run-ci` to these commands. We parse them for `runner run-ci`.
+          - check-format --scope=all
```
tiziano88, line 23: Isn't scope=all the default now?

runner/src/diffs.rs
```diff
@@ -44,17 +44,25 @@ impl ModifiedContent {
...
+/// Get all the files that have been modified in the given `scope`. If the scope is `Scope::All`
+/// returns an empty list of files in `ModifiedContent`. For `Scope::Commits` and
+/// `Scope::DiffToMain` the returned files include tracked files with unstaged changes in, but do
+/// not include new files that are not tracked (i.e., not added to git yet). For
+/// `Scope::Commits(N)`, a positive number of commits must be given. In case of zero or
+/// a negative number, only the last commit will be considered for finding the modified files.
+pub fn modified_files(scope: &Scope) -> ModifiedContent {
+    let args = match scope {
+        Scope::All => None,
+        Scope::Commits(commits) => Some(vec![format!("HEAD~{}", std::cmp::max(1, *commits))]),
```
tiziano88, line 56: I think these can just be flattened to Vec, and return empty Vec for Scope::All. The extra optionality does not seem to represent any actual state.

runner/src/internal.rs
```diff
@@ -47,16 +47,16 @@ pub enum Command {
...
-    CheckFormat(Commits),
+    Format(ScopeOpt),
+    CheckFormat(ScopeOpt),
     RunTests,
-    RunCargoClippy(Commits),
+    RunCargoClippy(ScopeOpt),
     RunCargoTests(RunTestsOpt),
     RunBazelTests,
-    RunTestsTsan,
+    RunTestsTsan(ScopeOpt),
```
tiziano88, line 56: Alternatively, you could just move the scope arg to top-level (even if it's not actually used by all the commands), and then the ci.yaml file becomes trivial (either add the scope or not to any command, directly in the step)

runner/src/internal.rs
```diff
@@ -139,17 +136,44 @@ pub struct RunFunctionsExamples {
     pub server_additional_args: Vec<String>,
     #[structopt(long, help = "build a Docker image for the examples")]
     pub build_docker: bool,
-    #[structopt(
-        flatten,
-        help = "run the command only for files affected by the changes in the specified commits, including unstaged changes in tracked files"
-    )]
-    pub commits: Commits,
+    #[structopt(flatten)]
+    pub scope: ScopeOpt,
```
tiziano88, line 140: This should have a default of All (everywhere)

tiziano88 (approved):

.github/workflows/cimain.yaml
rbehjati: Currently, the default is `--scope=diff-to-main`. I think when running locally we'd want to run the faster version by default. But I can change the default.

runner/src/diffs.rs
rbehjati: You are right.  But in the end `files` is `Option`. So, having `args` as option gives a nicer mapping.

I've removed `vec` for now. It will be needed if we manage to update the git version and use the additional `--merge-base` flag.

runner/src/internal.rs
rbehjati: Very good idea. Thanks. Done!

rbehjati (commented):
Thanks for the review.

tiziano88 (approved):
Personally I would leave --scope=all the default, but if you think most people would prefer --scope=diff_to_main and it's accurate enough, I'm happy with it! We can give it a try for a few days and see if we notice any inconsistencies.

.github/workflows/ci.yaml
```diff
@@ -101,10 +105,17 @@ jobs:
             cargo-cache-
 
       - name: Run command
+        if: ${{ github.ref != 'refs/heads/main' }}
         run: |
           ./scripts/docker_run ./scripts/runner --commands ${{ matrix.cmd }}
```
tiziano88, line 110: I would add an explicit scope here, otherwise this will change meaning if / when the default is changed for that flag, right?

.github/workflows/ci.yaml
```diff
@@ -101,10 +105,17 @@ jobs:
...
           df --human-readable
 
+      - name: Run Command - Post-Merge
+        if: ${{ github.ref == 'refs/heads/main' }}
```
tiziano88, line 114: In reproducibility.yaml we use `if: github.event_name == 'push'` for this, could you make sure they are consistent (not sure if one is preferable or they are the same)

tiziano88 (approved):

.github/workflows/ci.yaml
```diff
@@ -101,10 +105,17 @@ jobs:
             cargo-cache-
 
       - name: Run command
+        if: ${{ github.ref != 'refs/heads/main' }}
```
tiziano88, line 108: (nit) you don't need the ${{ here (in fact I run into some issues with that at some point, I don't remember why)

rbehjati:
> Personally I would leave --scope=all the default, but if you think most people would prefer --scope=diff_to_main and it's accurate enough, I'm happy with it! We can give it a try for a few days and see if we notice any inconsistencies.

I agree. Let's give it a try for a few days, and if we are not happy with it, we can change the default, or better make the calculation of diffs more accurate. I'll test a few scenarios before merging.

.github/workflows/ci.yaml
rbehjati: Interesting! I copied this from the [github reference](https://docs.github.com/en/actions/learn-github-actions/contexts#determining-when-to-use-contexts)!

.github/workflows/ci.yaml
rbehjati: Good point. Done.

.github/workflows/ci.yaml
rbehjati: Thanks. I'll try this one.

.github/workflows/ci.yaml
```diff
@@ -45,8 +45,20 @@ jobs:
...
     steps:
+      - name: Get commits count
+        id: commits_count
+        run: |
+          echo "::set-output name=commits_count::`curl --silent ${{ github.event.pull_request.commits_url }} | jq length`"
+      - name: Compute commits fetch count
+        id: fetch_count
+        # Fetch an extra commit to make sure that `git diff` with the given number of commits works.
+        run: |
+          echo "::set-output name=commits_fetch_count::$((${{ steps.commits_count.outputs.commits_count }}+1))"
```
tiziano88, line 56: Maybe these should be combined and produce a single --scope= flag value, which gets used below once and for all (instead of duplicating that step changing only that flag)? i.e. the logic to decide whether we are in pull_request or push would also be moved here

rbehjati: I don't think that these can be combined. We still need two numbers, one to fetch the commit, the other to use in the scope flag. But the value of scope flag could be created here. I might leave that for another PR though.

rbehjati: Actually looks like the way the checkout action fetches the repository adds a single commit that contains all the changes. So, fetching with depth 2, and setting `--scope=commits:1` gives the list of all modified files in the PR.

PR #2471, opened 2022-01-10
Bump Rust edition to 2021

### LINKED ISSUE #2470: Bump Rust edition to 2021
I am hoping that the improved resolver (see https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions ) would allow caching to work more efficiently, especially in CI.

I am using the following command to automatically fix any potentially incompatible crate:

```console
$ find -name Cargo.toml -not -path './cargo-cache/*' -not -path './third_party/*' -not -path './bazel-cache/*' -print0 | xargs -I {} -0 cargo fix --edition --manifest-path={}
```

tiziano88:
- fix warnings / errors
- update some other minor deps

The main effect of this should be to use a new dependency resolver,
which may or may not improve caching. See https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions

Fixes #2470


jul-sh (approved):
lgtm, sweet :)

Issue #2014, opened 2021-04-27
Reenable sccache


conradgrobler:
Sccache is causing partial rebuilds when using `cargo build` followed by `cargo run`. This in turn is causing backend services in CI to start up too slowly and breaking the build.

We should reenable sccache once the runner is updated to run the built binaries for backend services directly rather that running them via `cargo run`

PR #2473, opened 2022-01-10
Cache for Cargo Clippy

tiziano88:


Dockerfile
```diff
@@ -302,12 +302,9 @@ RUN mkdir --parents ${sccache_dir} \
...
   && chmod +x ${sccache_dir}/sccache
 
-ENV SCCACHE_GCS_BUCKET sccache-1
-ENV SCCACHE_GCS_KEY_PATH /workspace/.oak_remote_cache_key.json
-ENV SCCACHE_GCS_RW_MODE READ_WRITE
+# By default, sccache uses `~/.cache/sccache` locally: https://github.com/mozilla/sccache#local.
 
-# TODO(#2014): Re-enable sccache once runner runs backend binaries directly rather than using `cargo run`.
-# ENV RUSTC_WRAPPER sccache
+ENV RUSTC_WRAPPER sccache
```
mariaschett, line 307: Do we want to link #2014 ?

.github/workflows/ci.yaml
```diff
@@ -84,16 +84,33 @@ jobs:
       # See https://doc.rust-lang.org/nightly/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci
       - name: Cache Rust artifacts
         uses: actions/cache@v2
+        if: |
+          ${{ matrix.cmd == 'run-cargo-clippy' || matrix.cmd == 'run-functions-examples --application-variant=rust' }}
```
mariaschett, line 88: Did caching work? Because if I understand the last CI run correctly, run-cargo-clippy took still 36 min (https://github.com/project-oak/oak/runs/4769340244?check_suite_focus=true). We'd expecting this to be less, no? But it seems we just kicked off a run again!

mariaschett (commented):
Thanks! Very much looking forward to caching!

Dockerfile
tiziano88: Done.

.github/workflows/ci.yaml
tiziano88: I got a run with cargo clippy running in 4 mins instead of 40 mins (with no changes to the code), but now it's hard to reproduce because the cache keeps being evicted by other jobs. I think we'll have to submit this and see how it goes to have more useful data.

Dockerfile
tiziano88: Done.

.github/workflows/ci.yaml
mariaschett: Sounds good to submit and observe! 

But should we maybe note somewhere (e.g., the title), that the sccache is only working for run-cargo-clippy and Rust run-function-examples? We still may want to look into run-cargo-tests, if this works well, no?

mariaschett (commented):
Thanks!

.github/workflows/ci.yaml
tiziano88: SG, I think the next thing to explore would be the single workspace though, there is just not enough space available to enable it for another step, otherwise I would do that already.

mariaschett (approved):
Thanks

PR #2477, opened 2022-01-11
Use cargo patch for tract-tensorflow as it is more idiomatic and selects the newest version

jul-sh:
We previously inlined the source as patching caused other breaking dependency updates, but this is no longer the case. Hence let's use patch.


tiziano88 (commented):
I'm not sure whether this solves any problems in particular, but in terms of readability this seems inferior to me. I would rather have the explicit version pinned in two places, and when we decide to remove the pin, we modify those. In my experience `patch` is a last resort in case you need to pin indirect deps, which cannot be specified directly.

PR #2480, opened 2022-01-11
Reenable cancelling of parallel jobs for all github workflows

rbehjati:


tiziano88 (approved):

.github/workflows/build_rust_docs.yaml
```diff
@@ -6,6 +6,11 @@ on:
   pull_request:
     branches: [main]
 
+# See https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run
+concurrency:
+  group: build-rust-docs-${{ github.head_ref || github.ref }}
```
tiziano88, line 11: Is there a way of parametrizing this with the actual workflow name, so we can copy paste it everywhere?

tiziano88, line 11: Could you also rename the existing one to follow the same pattern?

rbehjati: Good idea, I think `env.GITHUB_WORKFLOW` might work. I'll try.

PR #2481, opened 2022-01-11
Unify Rust workspaces

### LINKED ISSUE #2475: Unify Rust workspaces
I think https://github.com/project-oak/oak/pull/2473 would be much more effective and efficient if all the crates were under a single workspace.

IIRC the only reason for having separate workspaces was to make it easier to keep deps separate, but I think we agreed that we may not care about that, as long as, for instance, we can list the exact versions of the deps used by the oak_functions_loader binary, even if it belongs to that workspace.

Or am I forgetting something else? @project-oak/core

tiziano88:
Unify Rust workspaces

- remove old target dirs from caching in CI
- move all crates under a top-level workspace
- fix deps pointing to internal crates to specify paths explicitly
- remove old Cargo.lock and deny.toml files
- fix paths of Wasm modules in examples
- comment out workspace-specific lines in third_party/prost/Cargo.toml,
since nested workspaces are not well supported
- remove unused deps from third party crates

Fix #2475


tiziano88:
@mariaschett @rbehjati PTAL, I think I solved all the remaining issues now. There are a couple of crates that are still outside of the workspace (and they seem to work fine there, while others were not working outside of the workspace); if needed we can move those later. Thanks!

.github/workflows/ci.yaml
```diff
@@ -85,32 +85,15 @@ jobs:
       - name: Cache Rust artifacts
         uses: actions/cache@v2
         if: |
-          matrix.cmd == 'run-cargo-clippy' || matrix.cmd == 'run-functions-examples --application-variant=rust'
+          matrix.cmd == 'run-cargo-clippy' || matrix.cmd == 'run-examples --application-variant=rust' || matrix.cmd == 'run-examples --application-variant=cpp' || matrix.cmd == 'run-functions-examples --application-variant=rust'
```
rbehjati, line 88: nit: Could this be split into multiple shorter lines?

mariaschett, line 88: Just for clarification: are we successively adding commands, i.e., `run-examples for cpp and rust` this time , which use the cache?

third_party/prost/Cargo.toml
```diff
@@ -14,25 +14,24 @@ keywords = ["protobuf", "serialization"]
...
-]
-exclude = [
-  # The fuzz crate can't be compiled or tested without the 'cargo fuzz' command,
-  # so exclude it from normal builds.
-  "fuzz",
-  # Same counts for the afl fuzz targets
-  "afl",
-]
-
+# [workspace]
```
rbehjati, line 17: Should this be removed?

third_party/prost/Cargo.toml
```diff
@@ -48,14 +47,11 @@ prost-derive = { version = "0.8.0", path = "prost-derive", optional = true }
...
 criterion = "0.3"
-env_logger = { version = "0.8", default-features = false }
-log = "0.4"
 proptest = "1"
 rand = "0.8"
 
-[profile.bench]
-debug = true
-
+# [profile.bench]
```
rbehjati, line 53: Remove?

.github/workflows/ci.yaml
```diff
@@ -85,32 +85,15 @@ jobs:
...
         with:
           path: |
             ./cargo-cache/bin
             ./cargo-cache/registry/index
             ./cargo-cache/registry/cache
             ./cargo-cache/git/db
             ./sccache-cache
-            ./examples/abitest/module_0/rust/target
```
mariaschett, line 96: IIUC correctly `abitest/module_0&1` have their own workspace and do not build into `./target`. As we don't keep them here, this means they won't be cached, but that is okay. Is this correct?

runner/Cargo.toml
```diff
@@ -1,5 +1,3 @@
-[workspace]
```
rbehjati, line 1: I wonder if the `runner` could be kept as a separate workspace. If we want to publish the workspace on crate.io, it probably does not make sense to include the runner in what we publish. Since the runner is our development tool.

oak_loader/.dockerignore
```diff
@@ -2,4 +2,4 @@
 **
 
 # Allow only the binaries we care about
-!/target/x86_64-unknown-linux-musl/release/**
+!/bin/**
```
mariaschett, line 5: Out of curiousity: I am not sure what we allow here, but I am a bit confused by the path. I assume we want the to allow `workspace/bin/**` here because nearly every binary is copied there now. But what was allowed before---did we have a `workspace/target/...` folder?

third_party/prost/Cargo.toml
mariaschett, line 53: Why did we comment this out?

mariaschett (approved):
Thank you, lovely! Very much looking forward to sharing more dependencies/builds!

After merging, it may be nice if you could you ping everyone to git clean their workspace.

Also: I am not very sure about the prost-third party part. Maybe you could walk me through or @rbehjati could check this better than me.

rbehjati (approved):
Nice. Thanks :) 

A few things might still need to be updated (some of these are scripts that we might not want to maintain):
`examples/bin` -> `bin`:
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/scripts/deploy_oak_functions_loader#L29 

`*/target/x86_64-unknown-linux-musl/release/...` -> `bin/...`: 
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/BUILD#L41
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/experimental/gcp_api_gateway/scripts/deploy_server#L21
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/experimental/oak_functions_with_envoy/scripts/build.sh#L16
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/oak_functions/.dockerignore#L2
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/oak_functions/Dockerfile#L5
- https://github.com/project-oak/oak/blob/main/docs/programming-oak.md
- https://github.com/project-oak/oak/blob/main/oak_functions/examples/weather_lookup/README.md

`./oak_functions/bin` -> `bin` (these might be false positives):
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/scripts/deploy_oak_functions_endpoints#L14
- https://github.com/project-oak/oak/blob/0f2189703c57845e09d8ab89164a4041c0af0a62/scripts/deploy_oak_functions_api#L17

.github/workflows/ci.yaml
tiziano88: Yes, I may look into those separately anyways.

.github/workflows/ci.yaml
tiziano88: I tried before and got it wrong and it would silently ignore parts of the string, so I'll keep it like this for now, I will remove it at some point though when the cache can contain everything

third_party/prost/Cargo.toml
tiziano88: Done.

third_party/prost/Cargo.toml
tiziano88: Done.

tiziano88: Because it's a workspace-specific value, removed entirely now.

runner/Cargo.toml
tiziano88: The point of the workspace is not for publishing, just for sharing cache, and making it easier for the editor to understand the code. Also when publishing, each crate is completely indepenent of the others, there is no concept of publishing a workspace (that I know of).

oak_functions/loader/fuzz/Cargo.toml
```diff
@@ -23,10 +23,6 @@ tonic = "*"
 [build-dependencies]
 prost-build = "0.8.0"
 
-# Prevent this from interfering with workspaces
```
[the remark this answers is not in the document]
tiziano88: I put this back (it seems to work fine as a separate workspace)

oak_loader/.dockerignore
tiziano88: Previously, this used to be invoked from the context of this package, and target referred to the local target folder. Now, it is invoked at the workspace root, and there we also have the bin folder, so I decided to use that. If we remove --out-dir, we may have to change this to target again, though still relative to the workspace root.

tiziano88 (commented):
Thanks both, I think I fixed most remaining places now, but if I missed something feel free to let me know and I'll create another PR.

runner/Cargo.toml
rbehjati: Great. Thanks for clarifying :)

PR #2488, opened 2022-01-13
Add implementation for hosted ABI call `channel_read`

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -49,6 +51,8 @@ const ABI_CHANNEL_BOUND: usize = 100;
 // of Wasm would use different types.
 pub type AbiPointer = u32;
 pub type AbiPointerOffset = u32;
+// Type alias for the ChannelHandle type, which has to be casted into a ChannelHandle.
```
tiziano88, line 54:
```suggestion
// Type alias for the ChannelHandle type, which has to be cast into a ChannelHandle.
```

oak_functions/loader/src/server.rs
```diff
@@ -248,6 +253,43 @@ impl WasmState {
...
+            .channel_switchboard
+            .get_mut(&channel_handle)
+            .ok_or(ChannelStatus::ChannelHandleInvalid)?;
+        let receiver = &mut endpoint.receiver;
+        let message = receiver.try_recv().map_err(|e| match e {
+            TryRecvError::Empty => ChannelStatus::ChannelEmpty,
+            TryRecvError::Disconnected => ChannelStatus::ChannelEndpointDisconnected,
+        })?;
+
+        // Write message to memory of wasm module.
```
tiziano88, line 275:
```suggestion
        // Write message to memory of Wasm module.
```

oak_functions/loader/src/server.rs
```diff
@@ -248,6 +253,43 @@ impl WasmState {
...
+        let dest_ptr = self.alloc(message.len() as u32);
+
+        let from_oak_status = |e| match e {
+            OakStatus::ErrInvalidArgs => ChannelStatus::ChannelInvalidArgs,
+            _ => ChannelStatus::Unspecified,
+        };
```
tiziano88, line 281: This may be a nested function, but also it would be even nicer to define `From<OakStatus> for ChannelStatus` or something like that, so that the `?` operator does this conversion automatically.

oak_functions/loader/src/server.rs
```diff
@@ -559,9 +603,15 @@ impl WasmHandler {
...
     }
 
-    fn init_wasm_state(&self, request_bytes: Vec<u8>) -> anyhow::Result<WasmState> {
+    fn init(&self, request_bytes: Vec<u8>) -> anyhow::Result<WasmState> {
         let mut extensions_indices = HashMap::new();
         let mut extensions_metadata = HashMap::new();
+        let mut extensions_endpoints = HashMap::new();
+
+        let mut channel_switchboard = ChannelSwitchboard::new();
+        // Register only lookup channel for now
```
tiziano88, line 612:
```suggestion
        // Register only lookup channel for now.
```

oak_functions/loader/src/server.rs
```diff
@@ -559,9 +603,15 @@ impl WasmHandler {
...
+        let lookup_endpoint = channel_switchboard.register(ChannelHandle::LookupData);
+        extensions_endpoints.insert(LOOKUP_ABI_FUNCTION_NAME.to_string(), lookup_endpoint);
```
tiziano88, line 614: So you are using the pre-existing function name to index in the map? That's fine, but it seems we need to be quite careful, since it was not meant for this (at least, I find it a bit unexpected).

oak_functions/loader/src/server.rs
```diff
@@ -740,21 +793,113 @@ mod tests {
...
+        // Guess some memory addresses in linear Wasm memory.
+        let dest_ptr_ptr: AbiPointer = 100;
+        let dest_len_ptr: AbiPointer = 150;
+
+        let result = wasm_state.channel_read(channel_handle, dest_ptr_ptr, dest_len_ptr);
+        assert!(result.is_ok());
+
+        // Get dest_len from dest_len_ptr.
+        let dest_len: u32 =
+            LittleEndian::read_u32(&wasm_state.get_memory().get(dest_len_ptr, 4).unwrap());
```
tiziano88, line 887: I thought we have some helper functions for this sort of read / conversion on WasmState? If not, is it worth adding them?

oak_functions/loader/src/server.rs
```diff
@@ -740,21 +793,113 @@ mod tests {
...
+
+        // Get dest_ptr from dest_ptr_ptr.
+        let dest_ptr: u32 =
+            LittleEndian::read_u32(&wasm_state.get_memory().get(dest_ptr_ptr, 4).unwrap());
+
+        // Assert that dest_ptr holds message.
+        assert_eq!(
+            wasm_state
+                .get_memory()
+                .get(dest_ptr, message.len())
```
tiziano88, line 900: I would use dest_len here for clarity.

tiziano88 (approved):
Very neat PR!

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs
mariaschett: Yes, I added a comment. It is a workaround until I can integrate channels better into extensions. I have added a comment on `let mut extension_endpoints`.

oak_functions/loader/src/server.rs
mariaschett: I only found some for `write_u32/buffer_to_wasm_memory` but no read. 

Makes sense though, because I am not sure where we would need reading from Wasm memory apart from this test, so I would not add them at the moment.

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

mariaschett: Neat! Thank you! :)

tiziano88 (approved):

PR #2493, opened 2022-01-18
Refactor `Endpoint` functionality to be private

mariaschett:
Moved tests to `server.rs` so the functionality related to `Endpoint` can be private. We might need export `Endpoint` later, but we should do that on demand.

### DIFF
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -719,6 +719,8 @@ pub fn format_bytes(v: &[u8]) -> String {
 }
 
 // The Endpoint of a bidirectional channel. Sender and Receiver are exposed.
+#[allow(dead_code)]
+
 struct Endpoint {
     sender: Sender<AbiMessage>,
     receiver: Receiver<AbiMessage>,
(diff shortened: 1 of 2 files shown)


oak_functions/loader/src/server.rs
```diff
@@ -777,24 +779,36 @@ mod tests {
...
+        let (mut module, mut runtime) = channel_create();
+        check_empty(&mut module);
+        check_empty(&mut runtime);
+    }
 
-        let wasm_module_bytes = test_utils::create_some_wasm_module_bytes();
+    #[tokio::test]
+    async fn test_crossed_write_read() {
+        async fn check_crossed_write_read(endpoint1: &mut Endpoint, endpoint2: &mut Endpoint) {
+            let message = String::from("Message").into_bytes();
```
tiziano88, line 796: What's the `from` for? Does `"Message".as_bytes()` work? Also messages are really binary, so maybe just make up a binary Vec with some numbers?

oak_functions/loader/src/server.rs
```diff
@@ -777,24 +779,36 @@ mod tests {
...
 
-    fn create_test_wasm_state() -> WasmState {
-        let wasm_handler = create_test_wasm_handler();
-        wasm_handler
-            .init(b"".to_vec())
-            .expect("could not create wasm_state")
+            assert_eq!(message, received_message);
+        }
+
+        let (mut module, mut runtime) = channel_create();
```
tiziano88, line 807: I would not even mention module and runtime here, just endpoint_0 and endpoint_1 perhaps, or something else more abstract?

tiziano88 (approved):

oak_functions/loader/src/server.rs

mariaschett (commented):
Thank you!

PR #2494, opened 2022-01-18
Use generic gRPC encapsulation rather than explicit protos in Remote Attestation Proxy

conradgrobler:
The Remote Attestation Proxy instance currently has an explicit dependency on the service/proto definitions of the backend (TensorFlow Serving). To make it more generic and reusable this change moves it to a model where it receives a generic gRPC encapsulation request and just passes the bytes through to the backend.

### DIFF
--- a/experimental/tf_proxy/client/src/main.rs
+++ b/experimental/tf_proxy/client/src/main.rs
@@ -24,7 +24,10 @@ use maplit::hashmap;
 use oak_functions_abi::proto::{ConfigurationInfo, Request};
 use oak_functions_client::Client;
 use prost::Message;
-use proto::serving::{ModelSpec, PredictRequest, PredictResponse};
+use proto::{
+    oak::encap::GrpcRequest,
+    serving::{ModelSpec, PredictRequest, PredictResponse},
+};
 use structopt::StructOpt;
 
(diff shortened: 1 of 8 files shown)


tiziano88 (approved):
Very nice!

experimental/tf_proxy/client/src/main.rs
```diff
@@ -24,7 +24,10 @@ use maplit::hashmap;
 use oak_functions_abi::proto::{ConfigurationInfo, Request};
```
rbehjati, line 24: pre-existing: is it possible to move the imported parts from `oak_functions_abi` to `grpc_attestation`, and use that both here and in `oak_functions`?

conradgrobler: I created #2495 to track this. I think there are other parts in the crate (and even in `abi.proto`) that is unrelated to the Wasm ABI.

I will do some of the cleanup as part of the PR that will be updating the Oak Functions loader to use the remote attestation proxy.

rbehjati: Sounds good. I think it make sense to do some refactoring and even splitting the `abi.proto` file.

PR #2496, opened 2022-01-19
Add implementation for hosted ABI call `channel_write`

mariaschett:


oak_functions/proto/abi.proto
```diff
@@ -31,10 +31,14 @@ enum ChannelStatus {
...
   CHANNEL_HANDLE_INVALID = 3;
-  // The endpoint of a channel is disconnected.
+  // The receiving endpoint of a channel is disconnected.
   CHANNEL_ENDPOINT_DISCONNECTED = 4;
   // Writing message from channel to Wasm memory failed.
   CHANNEL_INVALID_ARGS = 5;
+  // Channel is full.
+  CHANNEL_FULL = 6;
+  // The sending endpoint of a channel is closed.
+  CHANNEL_ENDPOINT_CLOSED = 7;
```
conradgrobler, line 41: Will this actually be used in our impelementation? What will cause the channel to be closed?

conradgrobler (approved):
Nice

oak_functions/proto/abi.proto
mariaschett: If the (sender) endpoint goes out of scope then the channel will be closed for the corresponding (receiving) endpoint. I am not yet sure if we will actively use it.

PR #2501, opened 2022-01-21
Add UwabiExtension Trait and Testing extension

mariaschett:
- Define trait for `Uwabi` extensions
- Change `Extensions` to be either a `Native` or `Uwabi` extension
- Define new Testing Extension which implements `Uwabi` extension
- Use Testing Extension instead of Lookup Extension for tests

### DIFF
--- a/oak_functions/loader/src/metrics.rs
+++ b/oak_functions/loader/src/metrics.rs
@@ -17,7 +17,7 @@
 use crate::{
     logger::Logger,
     server::{
-        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, Extension::Native,
+        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, Extension,
         ExtensionFactory, OakApiNativeExtension, WasmState, ABI_USIZE,
     },
 };
@@ -259,7 +259,7 @@ impl ExtensionFactory for PrivateMetricsProxyFactory {
             metrics_proxy: Some(metrics_proxy),
             logger: self.logger.clone(),
         };
-        Ok(Native(Box::new(extension)))
+        Ok(Extension::Native(Box::new(extension)))
     }
 }
 
--- a/oak_functions/loader/src/metrics.rs
+++ b/oak_functions/loader/src/metrics.rs
@@ -17,8 +17,8 @@
 use crate::{
     logger::Logger,
     server::{
-        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, Extension,
-        ExtensionFactory, OakApiNativeExtension, WasmState, ABI_USIZE,
+        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, ExtensionFactory,
+        OakApiNativeExtension, WasmState, ABI_USIZE,
     },
 };
 use oak_functions_abi::proto::OakStatus;
@@ -259,7 +259,7 @@ impl ExtensionFactory for PrivateMetricsProxyFactory {
             metrics_proxy: Some(metrics_proxy),
             logger: self.logger.clone(),
         };
-        Ok(Extension::Native(Box::new(extension)))
+        Ok(BoxedExtension::Native(Box::new(extension)))
     }
 }
 
--- a/oak_functions/loader/src/metrics.rs
+++ b/oak_functions/loader/src/metrics.rs
@@ -17,8 +17,8 @@
 use crate::{
     logger::Logger,
     server::{
-        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, ExtensionFactory,
-        OakApiNativeExtension, WasmState, ABI_USIZE,
+        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, Extension::Native,
+        ExtensionFactory, OakApiNativeExtension, WasmState, ABI_USIZE,
     },
 };
 use oak_functions_abi::proto::OakStatus;
@@ -259,7 +259,7 @@ impl ExtensionFactory for PrivateMetricsProxyFactory {
             metrics_proxy: Some(metrics_proxy),
             logger: self.logger.clone(),
         };
-        Ok(Box::new(extension))
+        Ok(Native(Box::new(extension)))
     }
 }
 
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -157,7 +157,7 @@ pub enum Extension {
 pub type BoxedExtension = Extension;
 pub type BoxedExtensionFactory = Box<dyn ExtensionFactory + Send + Sync>;
 
-/// Trait for implementing an extension which relies on uwabi.
+/// Trait for implementing an extension which relies on UWABI.
 pub trait UwabiExtension {
     /// Get the channel handle to address this extension.
     fn get_channel_handle(&self) -> ChannelHandle;
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -177,7 +177,10 @@ pub struct WasmState {
     extensions_indices: Option<HashMap<usize, BoxedExtension>>,
     /// A mapping of host function names to metadata required for resolving the function.
     extensions_metadata: HashMap<String, (usize, wasmi::Signature)>,
+    /// A mapping from channel handles to the hosted endpoints of channels.
     channel_switchboard: ChannelSwitchboard,
+    /// A mapping from channel handles to the endpoints for the Oak functions runtime.
+    // TODO(#2502) Add extension endpoints to the corresponding endpoint.
     extensions_endpoints: HashMap<ChannelHandle, Endpoint>,
 }
 
@@ -201,6 +204,7 @@ impl WasmState {
         }
     }
 
+    /// Reads the buffer starting at address `buf_ptr` with length `buf_len` from the Wasm memory.
     pub fn read_buffer_from_wasm_memory(
         &self,
         buf_ptr: AbiPointer,
@@ -217,6 +221,8 @@ impl WasmState {
             })
     }
 
+    /// Writes the buffer `source` at the address `dest` of the Wasm memory, if `source` fits in the
+    /// allocated memory.
     pub fn write_buffer_to_wasm_memory(
         &self,
         source: &[u8],
@@ -232,6 +238,7 @@ impl WasmState {
         })
     }
 
+    ///  Writes the u32 `value` at the `address` of the Wasm memory.
     pub fn write_u32_to_wasm_memory(
         &self,
         value: u32,
@@ -248,8 +255,8 @@ impl WasmState {
         })
     }
 
-    // Writes the given `buffer` by allocating `buffer.len()` Wasm memory and writing the address of
-    // the allocated memory to `dest_ptr_ptr` and the length to `dest_len_ptr`.
+    /// Writes the given `buffer` by allocating `buffer.len()` Wasm memory and writing the address
+    /// of the allocated memory to `dest_ptr_ptr` and the length to `dest_len_ptr`.
     pub fn alloc_and_write_buffer_to_wasm_memory(
         &mut self,
         buffer: Vec<u8>,
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -691,7 +691,6 @@ impl WasmHandler {
                 }
                 Extension::Uwabi(_uwabi_extension) => {
                     // TODO(mschett) Add functionality for creating endpoints
-                    panic!("Uwabi Extension in init of WasmState not defined.")
                 }
             }
         }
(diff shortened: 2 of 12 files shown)


oak_functions/loader/src/server.rs
```diff
@@ -145,13 +144,25 @@ pub trait OakApiNativeExtension {
...
-pub type BoxedExtension = Box<dyn OakApiNativeExtension + Send + Sync>;
+pub enum Extension {
+    Native(Box<dyn OakApiNativeExtension + Send + Sync>),
+    Uwabi(Box<dyn UwabiExtension + Send + Sync>),
+}
+
+pub type BoxedExtension = Extension;
 pub type BoxedExtensionFactory = Box<dyn ExtensionFactory + Send + Sync>;
 
+/// Trait for implementing an extension which relies on uwabi.
```
tiziano88, line 160: Should it be

```suggestion
/// Trait for implementing an extension which relies on UWABI.
```

?

oak_functions/loader/src/server.rs
```diff
@@ -145,13 +144,25 @@ pub trait OakApiNativeExtension {
...
+pub trait UwabiExtension {
+    /// Get the channel handle to address this extension.
+    fn get_channel_handle(&self) -> ChannelHandle;
```
tiziano88, line 163: Why not an associated const? Or failing, that, an associated method? AFAICT there is nothing that needs `&self` in the implementation of this fn.

oak_functions/loader/src/server.rs
```diff
@@ -680,13 +709,20 @@ impl WasmHandler {
...
             .expect("no extensions_indices is set in wasm_state")
             .into_values()
         {
-            extension.terminate()?;
+            match extension {
+                Extension::Native(mut native_extension) => {
+                    native_extension.terminate()?;
+                }
+                Extension::Uwabi(_uwabi_extension) => {
+                    panic!("Uwabi Extension inserted in extension_indicies.")
```
tiziano88, line 723: So why is it allowed? Is it possible to use the indexing only for Native extensions, and not Uwabi, using the type system?

oak_functions/proto/abi.proto
```diff
@@ -45,6 +45,7 @@ enum ChannelStatus {
 // ChannelHandle to read, write, or wait on the corresponding channel.
 enum ChannelHandle {
   CHANNEL_HANDLE_UNSPECIFIED = 0;
+  TESTING = 1;
```
tiziano88, line 48: Perhaps add a short comment here?

oak_functions/loader/src/server.rs
```diff
@@ -905,22 +969,20 @@ mod tests {
 
     #[tokio::test]
     async fn test_hosted_channel_read_no_message() {
-        let channel_handle = ChannelHandle::LookupData as i32;
+        let channel_handle = ChannelHandle::Testing as i32;
```
tiziano88, line 972: Why does it matter what handle is used here? Is the test wasm state expected to write to that channel (it didn't seem the case to me, maybe I missed it)?

mariaschett: Because I changed the keys in `extension_endpoints` in `WasmState` to `ChannelHandle` (instead of the `ABI_FUNCTION_NAME`, which UwabiExtensions don't have). 

In the next PR I want to give the endpoints to the extensions themselves, then this will change.

oak_functions/proto/abi.proto
mariaschett: Done.

oak_functions/loader/src/server.rs
rbehjati: I think this part is only about termination logic. If UwabiExtensions don't have termination logic, perhaps an if-statement is more suitable here, than match and panic.

mariaschett (commented):
Thank you!

oak_functions/loader/src/server.rs
```diff
@@ -238,6 +249,21 @@ impl WasmState {
         })
     }
 
+    // Writes the given `buffer` by allocating `buffer.len()` Wasm memory and writing the address of
```
rbehjati, line 252: Use doc comment with `///` instead of `//` :)

oak_functions/loader/src/server.rs
```diff
@@ -191,7 +202,7 @@ impl WasmState {
         }
     }
 
-    fn read_buffer_from_wasm_memory(
+    pub fn read_buffer_from_wasm_memory(
```
rbehjati, line 205: Perhaps add a short doc comment, since it is public.

oak_functions/loader/src/server.rs
```diff
@@ -411,8 +434,8 @@ impl wasmi::Externals for WasmState {
                     .take()
                     .expect("no extensions_indices is set");
                 let extension = match extensions_indices.get_mut(&index) {
-                    Some(extension) => Box::new(extension),
-                    None => panic!("Unimplemented function at {}", index),
+                    Some(Extension::Native(extension)) => Box::new(extension),
+                    _ => panic!("Unimplemented function at {}", index),
```
rbehjati, line 438: I'd suggest distinguishing between `None` and `Some(Extension::Uwabi(...))`, perhaps with a different panic message for UWABI. It is a good practice to keep the match arms exhaustive.

oak_functions/loader/src/server.rs
```diff
@@ -168,7 +179,7 @@ pub struct WasmState {
     /// A mapping of host function names to metadata required for resolving the function.
     extensions_metadata: HashMap<String, (usize, wasmi::Signature)>,
     channel_switchboard: ChannelSwitchboard,
-    extensions_endpoints: HashMap<String, Endpoint>,
+    extensions_endpoints: HashMap<ChannelHandle, Endpoint>,
```
rbehjati, line 182: pre-existing: Can you add doc-comments for this and channel_switchboard?

oak_functions/loader/src/server.rs
```diff
@@ -649,19 +672,25 @@ impl WasmHandler {
...
+            match extension {
+                Extension::Native(ref native_extension) => {
+                    let (name, signature) = native_extension.get_metadata();
+                    extensions_indices.insert(ind + EXTENSION_INDEX_OFFSET, extension);
+                    extensions_metadata.insert(name, (ind + EXTENSION_INDEX_OFFSET, signature));
+                }
+                Extension::Uwabi(uwabi_extension) => {
+                    let channel_handle = uwabi_extension.get_channel_handle();
+                    let endpoint = channel_switchboard.register(channel_handle);
+                    extensions_endpoints.insert(channel_handle, endpoint);
```
rbehjati, line 691: Looks like <`extensions_indices`, `extensions_metadata`> and <`channel_switchboard`, `extensions_endpoints`> are somewhat mutually exclusive now. Is there a way to unify these constructs? Or perhaps rename them to clearly reflect the connection between them? How is `extensions_endpoints` related to `extensions_indices` and `extensions_metadata`?

oak_functions/loader/src/metrics.rs
```diff
@@ -17,8 +17,8 @@
 use crate::{
     logger::Logger,
     server::{
-        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, ExtensionFactory,
-        OakApiNativeExtension, WasmState, ABI_USIZE,
+        AbiPointer, AbiPointerOffset, BoxedExtension, BoxedExtensionFactory, Extension::Native,
```
rbehjati, line 20: nit: I think it is better not to import enum variants like this. I think it is clearer to refer to the enum variant as `Extension::Native` where it is used in the code. (Feel free to ignore this comment :))

oak_functions/loader/src/server.rs
rbehjati: I think it is better not to have test-specific variants in the proto spec. I'd suggest adding a TODO to the proto about removing `TESTING`.

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -257,12 +257,11 @@ pub fn assert_response_body(response: Response, expected: &str) {
     )
 }
 
-// Create some valid wasm bytecode
-pub fn create_some_wasm_module_bytes() -> Vec<u8> {
+/// Create some valid Wasm bytecode only calling Abi functions `read_request` and `write_request`.
+pub fn create_echo_wasm_module_bytes() -> Vec<u8> {
```
rbehjati, line 261: Is not it better to have this function in the test module itself? Or do you think that this function will be used in many tests? 

I feel like this module should contain generic test functionality, and this function suddenly feels very specific.

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs
mariaschett: Yes, you are right---they are mutually exclusive. The `extension_indices` and `extension_metadata` is related to the native extensions and the `channel_switchboard` and `extension_endpoints` to the Uwabi extensions.

I actually want to remove the `extension_endpoints` in the next step (#2502), so I would revisit that then. 

WDYT?

oak_functions/loader/src/server.rs
mariaschett: Changed to `if let` statement so it is less confusing.

oak_functions/loader/src/server.rs
mariaschett: I see your point of not having a test handle in the proto definition. 

But, I would love to have a testing extension to be able to use it in integration tests with the Wasm module for e.g., [here](https://github.com/project-oak/oak/blob/main/oak_functions/sdk/oak_functions/tests/integration_test.rs).

And I see no way except for providing a Test Handle in the proto. 

I think it is justifiable, as we will have some handles which are not used in every application (e.g., Metric/TF).

Can you think of any other way?

oak_functions/sdk/test_utils/src/lib.rs
mariaschett: Hmmm... I wanted it to be visible for other tests.

If you just need some valid Wasm bytecode (e.g., for creating a WasmHandler, but not executing the code), I think it could be useful.

I was looking for this function in the test utils before, so I put it here. But happy to move it, if you are not convinced.

mariaschett (commented):
Thank you! :)

oak_functions/sdk/test_utils/src/lib.rs
rbehjati: We have similar code to this repeated in several places in `oak_functions`. It might be better to change this to a more generic function like `create_wasm_module_bytes(manifest_path: &str, release: bool)` and use that everywhere where we use `compile_rust_wasm`. For instance: https://github.com/project-oak/oak/blob/201977e66a923a6a74cb5217bff6422423042bc2/oak_functions/loader/src/tests.rs#L114-L119

WDYT?

(This can be done in a separate PR.)

oak_functions/loader/src/server.rs
```diff
@@ -145,13 +144,25 @@ pub trait OakApiNativeExtension {
...
 }
+
 pub trait ExtensionFactory {
     fn create(&self) -> anyhow::Result<BoxedExtension>;
 }
 
-pub type BoxedExtension = Box<dyn OakApiNativeExtension + Send + Sync>;
+pub enum Extension {
+    Native(Box<dyn OakApiNativeExtension + Send + Sync>),
+    Uwabi(Box<dyn UwabiExtension + Send + Sync>),
```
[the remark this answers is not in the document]
rbehjati: I think it would be useful to have this explanation as a doc comment on `BoxedExtension`.

I am not sure about keeping two copies of the same logic. Does it mean that we'd have to maintain both and keep them in sync? Anyways, I think we can decide about this later :)

rbehjati (approved):
LGTM. Thanks :)

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
mariaschett: Added a comment on BoxedExtension.

I see your concern about synching. The main difference between the two is whether the extensions is invoked through a dedicated ABI function or through a message on the channel. The actual logic behind the extension should remain the same.

I agree we'll see later if we think that is a good idea :)

oak_functions/loader/src/server.rs
mariaschett: Good idea: #2505 .

oak_functions/loader/src/server.rs
mariaschett: I think it would be very convenient to have an extension for testing (you could think of it as a debug extension, too). 

The way I imagine this to work is that the Wasm module can send an aribitrary string, e.g., "Refresh the Lookup Data" to trigger some behaviour we need for testing and which we implement in the testing extension.

I believe, if we want to use that testing extension from a Wasm Module, there is no way without adding the Testing (or Debug) ChannelHandle to the proto. 

I suggest I try to implement what I have in mind with the Testing Extension for now and we re-evaluate (also you are right, it is nice to have it anyway as a Uwabi Extension)

oak_functions/sdk/test_utils/src/lib.rs
mariaschett: Ah, but I think the main benefit of `create_echo_wasm_bytes` is that we _fix_ the manifest path to the `Echo` module. 

So if---in future tests---we want to just get some valid Wasm code, we can use that. But mayb it is to confusing. 
Do you think it would be better to remove it from utils?

Do you think we gain much from `create_wasm_module_bytes`? I am not too sure, but happy to add an issue :)

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
rbehjati: If the actual logic is to be implemented only in one place, then I think it is good to have both :)

oak_functions/loader/src/server.rs
rbehjati: Sounds good to me if the idea is to allow WASM modules to use it too, but I think we have to make sure that it is only available in the `unsafe` version.

oak_functions/sdk/test_utils/src/lib.rs
rbehjati: I don't see a big difference between `create_echo_wasm_bytes()` and `create_wasm_module_bytes("examples/echo/module/Cargo.toml", false)`, but if we have `create_wasm_module_bytes` we can use it in other places too. 

Anyways, we can perhaps add `create_wasm_module_bytes` in a separate PR, and decide if we want to replace `create_echo_wasm_bytes` too.

oak_functions/loader/src/server.rs
tiziano88: Ah never mind, at the moment you need to get that back from the instance, since it's not known statically.

tiziano88 (approved):

PR #2503, opened 2022-01-24
Send post-merge failure notification

rbehjati:


rbehjati:
Unsuccessful attempt at sending email notifications when the CI fails post-merge (ref #2455). Closing it since it does not seem to be important or necessary to send these notifications, at least at the moment.

Issue #2502, opened 2022-01-24
Add extension endpoints to the corresponding Uwabi Extension


mariaschett:
- remove `extensions_endpoints` from `WasmState` 
- store the endpoint with the corresponding `UwabiExtension`

PR #2506, opened 2022-01-25
Add endpoints to UWABI extensions

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -153,15 +153,26 @@ pub trait ExtensionFactory {
...
+
 pub type BoxedExtensionFactory = Box<dyn ExtensionFactory + Send + Sync>;
 
 /// Trait for implementing an extension which relies on UWABI.
 pub trait UwabiExtension {
     /// Get the channel handle to address this extension.
     fn get_channel_handle(&self) -> ChannelHandle;
+
+    /// Get the endpoint.
+    fn get_endpoint(&mut self) -> Option<&mut Endpoint>;
```
tiziano88, line 169: Why would anything other than the extension itself need access to the endpoint? It seems to break encapsulation.

oak_functions/loader/src/server.rs
```diff
@@ -153,15 +153,26 @@ pub trait ExtensionFactory {
...
+
+    /// Set the endpoint if it has not been set before.
+    // Note: we cannot set the endpoint when we `create` the extension, as this would require
+    // to change the `BoxedExtensionFactory` trait. This helps to keep the changes to the
+    // (existing) Native extensions minimal.
+    fn set_endpoint(&mut self, endpoint: Endpoint);
```
tiziano88, line 175: This is going to be cleaned up when all the extensions are migrated, right?

oak_functions/loader/src/server.rs
```diff
@@ -153,15 +153,26 @@ pub trait ExtensionFactory {
...
 /// Trait for implementing an extension which relies on UWABI.
 pub trait UwabiExtension {
     /// Get the channel handle to address this extension.
     fn get_channel_handle(&self) -> ChannelHandle;
+
+    /// Get the endpoint.
+    fn get_endpoint(&mut self) -> Option<&mut Endpoint>;
+
+    /// Set the endpoint if it has not been set before.
+    // Note: we cannot set the endpoint when we `create` the extension, as this would require
```
tiziano88, line 172: It seems that the existing factory trait is already being stretched by returning two completely different types, and now it seems that it should also have different inputs. Have you considered just having a different factory trait, instead of trying to share the previous one? That may also simplify other things that currently cannot be enforced because of the ambiguity.

oak_functions/loader/src/server.rs
```diff
@@ -903,12 +912,26 @@ mod tests {
...
     impl UwabiExtension for TestingExtension {
         fn get_channel_handle(&self) -> oak_functions_abi::proto::ChannelHandle {
             ChannelHandle::Testing
         }
+
+        fn get_endpoint(&mut self) -> Option<&mut Endpoint> {
+            match &mut self.endpoint {
+                Some(endpoint) => Some(endpoint),
+                None => None,
+            }
```
tiziano88, line 927: https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref_mut

oak_functions/loader/src/server.rs
tiziano88, line 169: Maybe rename to `get_endpoint_mut`, usually that's explicit when returning mut refs.

mariaschett: True, that is not very pretty. I need the access to the endpoint only for tests right now.

It may be a much better idea to add `channel_read` and `channel_write` to the `UwabiExension` trait to be able to read/write to an endpoint. WDYT?

oak_functions/loader/src/server.rs
mariaschett: Adding a different factory trait would require me to change the [`WasmHandler`](https://github.com/project-oak/oak/blob/f1e9783cb872e9df58cce5a6d74ec8e872eb3822/oak_functions/loader/src/server.rs#L662), which I am trying to avoid. 

I think, having this two extension---`Uwabi` and `Native` is the simplest solution for now---leaving most of the current functionality untouched.

oak_functions/loader/src/server.rs
mariaschett: Yes.

oak_functions/loader/src/server.rs
mariaschett: Thank you!

oak_functions/loader/src/server.rs
mariaschett: I added a TODO to remove `get_endpoint`, because this requires some more design decisions #2508 .

oak_functions/loader/src/server.rs
mariaschett: Okay, so although I think using `as_deref_mut` should work, it doesn't. But given that we want to remove this code anyway, I would leave it like that.

tiziano88 (approved):

oak_functions/loader/src/server.rs
```diff
@@ -985,12 +1008,13 @@ mod tests {
...
+        let channel_handle = ChannelHandle::Testing;
         let mut wasm_state = create_test_wasm_state();
-        let extension_endpoints = &mut wasm_state.extensions_endpoints;
-        // Remove the endpoint of the runtime closes one endpoint of the channel.
-        extension_endpoints.remove(&ChannelHandle::Testing);
-        let result = wasm_state.channel_read(channel_handle, 0, 0);
+        // Remove the extension to close one endpoint of the channel.
+        wasm_state
+            .uwabi_extensions
+            .retain(|uwabi_extension| uwabi_extension.get_channel_handle() != channel_handle);
```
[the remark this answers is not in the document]

oak_functions/loader/src/server.rs
mariaschett: #2510

mariaschett:
Fixes #2502

PR #2507, opened 2022-01-25
Fix docker parent image by its digest

jul-sh:
Previously we just used a tag. Tags are mutable and don't describe the actual image. The image described by a tag could be swapped out on the registry. Using the digest means we use CaS and ask for the specific image we expect.


rbehjati (approved):
Oh! Nice :) 
Is a docker-push/docker-pull needed now?

jul-sh:
@rbehjati I don't think so, as it's just a different way of referring to the same image. :)

PR #2509, opened 2022-01-26
Add GitHub action to generate provenance files

jul-sh:
Creates provenance files (oak_loader, oak_functions_loader) for each commit added to the main branch, as outlined in: https://docs.google.com/document/d/1jWloPhrr-IykJBXtWuDdirqn9MH_FCUShBr_w0F3j3Y/edit?usp=sharing&resourcekey=0-q20m0jqEhzMyYn-kAbq_wA

This initial implementation generates provenance files as a TOML file in the shape currently used by https://github.com/project-oak/transparent-release. This will be changed to a SLSA provenance file, once our SLSA [build type](https://github.com/project-oak/transparent-release/pull/3) is fully defined, and transparent-release has been updated to parse that.

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -67,7 +67,7 @@ jobs:
           git -C ./provenance-branch status
           git -C ./provenance-branch diff --staged
           git -C ./provenance-branch commit --allow-empty --message="Provenance files for ${GITHUB_SHA}"
-          git -C ./provenance-branch tag -a -m "record source commit" provenance-${GITHUB_SHA}
+          git -C ./provenance-branch tag --annotate --message "record source commit" provenance-${GITHUB_SHA}
 
       - name: Push provenance branch
         run: git -C ./provenance-branch push --follow-tags
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -1,0 +1,73 @@
+name: Build Provenance
+
+# See https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: [main]
+jobs:
+  build_provenance:
+    runs-on: ubuntu-20.04
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Mount main branch
+        uses: actions/checkout@v2
+
+      - name: Mount provenance branch
+        uses: actions/checkout@v2
+        with:
+          ref: provenance
+          path: provenance-branch
+
+      - name: Git setup
+        run: |
+          git config --global user.email "actions@github.com"
+          git config --global user.name "GitHub Actions"
+
+      # Copied from https://github.com/jens-maus/RaspberryMatic/blob/ea6b8ce0dd2d53ea88b2766ba8d7f8e1d667281f/.github/workflows/ci.yml#L34-L40
+      - name: free disk space
+        run: |
+          df --human-readable
+          sudo swapoff --all
+          sudo rm --force /swapfile
+          sudo apt clean
+          docker rmi $(docker image ls --all --quiet)
+          df --human-readable
+
+      - name: Docker pull
+        timeout-minutes: 10
+        run: |
+          ./scripts/docker_pull
+          df --human-readable
+
+      - name: Build Server Provenance
+        run: |
+          ./scripts/build_provenance \
+          ./scripts/runner\ build-server \
+          ./oak_loader/bin/oak_loader
+
+      - name: Build Functions Provenance
+        run: |
+          ./scripts/build_provenance \
+          ./scripts/runner\ build-functions-server \
+          ./oak_functions/loader/bin/oak_functions_loader
+
+      - name: Move new provenance files to the provenance branch
+        run: cp -r ./provenance/. ./provenance-branch/
+
+      - name: Commit new provenance files to the provenance branch
+        run: |
+          git -C ./provenance-branch add .
+          git -C ./provenance-branch status
+          git -C ./provenance-branch diff --staged
+          git -C ./provenance-branch commit --allow-empty --message="Provenance files for ${GITHUB_SHA}"
+          git -C ./provenance-branch tag -a -m "record source commit" provenance-${GITHUB_SHA}
+
+      - name: Push provenance branch
+        run: git -C ./provenance-branch push --follow-tags
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -16,7 +16,7 @@ readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
 mkdir -p "./provenance"
 mkdir -p "./provenance/${BINARY_SHA_256_HASH}"
 
-readonly PROVENANCE=$(cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
+cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
 repo = "https://github.com/project-oak/oak"
 commit_hash = "$(git rev-parse --verify HEAD)"
 builder_image = "${DOCKER_IMAGE_REPO_DIGEST}"
@@ -24,5 +24,4 @@ command = ["${BUILD_COMMAND}"]
 output_path = "${OUTPUT_PATH}"
 expected_binary_sha256_hash = "${BINARY_SHA_256_HASH}"
 EOF
-)
 
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -16,6 +16,11 @@ readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
 mkdir -p "./provenance"
 mkdir -p "./provenance/${BINARY_SHA_256_HASH}"
 
+# Note that the command currently is a single string wrapped in an array. The
+# reason is that our transparent-release (which parses this file) expects an
+# array as the command value at its current version. There isn't much value
+# in using an array though, and in upcoming versions it will move towards
+# an unwrapped string.
 cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
 repo = "https://github.com/project-oak/oak"
 commit_hash = "$(COMMIT_HASH)"
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -22,7 +22,7 @@ mkdir -p "./provenance/${BINARY_SHA_256_HASH}"
 # an unwrapped string.
 cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
 repo = "https://github.com/project-oak/oak"
-commit_hash = "$(COMMIT_HASH)"
+commit_hash = "${COMMIT_HASH}"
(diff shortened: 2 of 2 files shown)


jul-sh:
Presently this is a simple shell script (a decision I kind of regret, debugging differences between my local bash version and the one in GitHub actions was pretty painful). 

We probably want to move to something else, esp once we generate actual json provenance files. Options include

- Rust? It could be integrated into the runner, but semantically that felt like an odd place for it to live. It would also mean running the code inside docker container whose outputs it describes.
- Go? It's what transparent release uses, but we don't really use it in Oak

Or move it into a separate GitHub action (potentially based in the transparent release repo) that Oak happens to use?

Curious about opinionz 🍿 cc @rbehjati @tiziano88

tiziano88:
Thanks Juliette, I agree that bash is not an ideal language, but I see how it ended up being convenient in this case. I think there is still something to figure out, because for now we are doing the actual build, and then separately generating the provenance file, hoping that it will match the actual build step, but in an ideal world we would have some mechanism that executes a build and also produces the actual provenance file, and that would be applicable to arbitrary repositories too (e.g. how would we do the same for TF, or some other binary?). I don't expect we will have a generalized solution that works in 100% of cases, but perhaps it should work for projects that use a Docker image and a git repository, like ours, even though they may install different tools in the Docker image (e.g. Bazel). I don't have a concrete idea of how this would work / be used though, I'm curious if you have thought about that already?

scripts/build_provenance
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+readonly SCRIPTS_DIR="$(dirname "$0")"
+# shellcheck source=scripts/common
+source "$SCRIPTS_DIR/common"
+
+readonly BUILD_SCRIPT=$1
+readonly OUTPUT_PATH=$2
+
+readonly BUILD_COMMAND="./scripts/docker_run ${BUILD_SCRIPT}"
```
tiziano88, line 10: This seems to tie the provenance to docker_run specifically, i.e. it would only be reproducible if the target repo actually has this script, right? Maybe it's more a question for the transparent release reproduce-build step, but I feel I'm still missing something in how these are related. Perhaps it may be easier to reason about once we generate the SLSA provenance instead of the TOML file?

scripts/build_provenance
```diff
@@ -0,0 +1,28 @@
...
+readonly BINARY_SHA_256_HASH=$(sha256sum "${OUTPUT_PATH}" | cut -d " " -f 1)
+
+readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
+
+mkdir -p "./provenance"
+mkdir -p "./provenance/${BINARY_SHA_256_HASH}"
+
+readonly PROVENANCE=$(cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
+repo = "https://github.com/project-oak/oak"
+commit_hash = "$(git rev-parse --verify HEAD)"
```
tiziano88, line 21: COMMIT_HASH

scripts/build_provenance
```diff
@@ -0,0 +1,28 @@
...
+readonly BUILD_COMMAND="./scripts/docker_run ${BUILD_SCRIPT}"
+eval "${BUILD_COMMAND}"
+readonly BINARY_SHA_256_HASH=$(sha256sum "${OUTPUT_PATH}" | cut -d " " -f 1)
+
+readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
+
+mkdir -p "./provenance"
+mkdir -p "./provenance/${BINARY_SHA_256_HASH}"
+
+readonly PROVENANCE=$(cat << EOF >> "./provenance/${BINARY_SHA_256_HASH}/${COMMIT_HASH}.toml"
```
tiziano88, line 19: Why is this being read into PROVENANCE?

.github/workflows/provenance.yaml
```diff
@@ -0,0 +1,73 @@
...
+          ./scripts/build_provenance \
+          ./scripts/runner\ build-functions-server \
+          ./oak_functions/loader/bin/oak_functions_loader
+
+      - name: Move new provenance files to the provenance branch
+        run: cp -r ./provenance/. ./provenance-branch/
+
+      - name: Commit new provenance files to the provenance branch
+        run: |
+          git -C ./provenance-branch add .
```
tiziano88, line 66: Maybe cd to the right dir first, instead of passing it to each command

.github/workflows/provenance.yaml
```diff
@@ -0,0 +1,73 @@
...
+          git -C ./provenance-branch status
+          git -C ./provenance-branch diff --staged
+          git -C ./provenance-branch commit --allow-empty --message="Provenance files for ${GITHUB_SHA}"
+          git -C ./provenance-branch tag -a -m "record source commit" provenance-${GITHUB_SHA}
```
tiziano88, line 70: -a -> --annotate
-m -> --message

etc. 

in general, I suggest always using long flags when writing scripts

scripts/build_provenance
```diff
@@ -0,0 +1,28 @@
...
+repo = "https://github.com/project-oak/oak"
+commit_hash = "$(git rev-parse --verify HEAD)"
+builder_image = "${DOCKER_IMAGE_REPO_DIGEST}"
+command = ["${BUILD_COMMAND}"]
```
tiziano88, line 23: Shouldn't this be split into the actual args? Or am I missing the point of the array?

rbehjati (approved):
Nice. Thanks!

I don't have a strong preference regarding the choice of the programming language. 

I can imagine a version of the workflow that collects all the information that are currently collected in the `build_provenance` script (builder image info might be tricky, but it there), creates a toml file and passes that to the CLI tool in https://github.com/project-oak/transparent-release to build the binary and the provenance file. 

I know that the CLI tool does not currently actually generate the provenance files, and we sort of decided that we will skip that part and the toml format. But if it simplifies this script and makes it look more generic, we could perhaps keep it and use it.

rbehjati:
> I don't expect we will have a generalized solution that works in 100% of cases, but perhaps it should work for projects that use a Docker image and a git repository, like ours, even though they may install different tools in the Docker image (e.g. Bazel). I don't have a concrete idea of how this would work / be used though, I'm curious if you have thought about that already?

@tiziano88  What do you think of the CLI build tool in https://github.com/project-oak/transparent-release/blob/main/cmd/build/main.go? Is that not what you are describing here? What do you think is missing?

scripts/build_provenance
jul-sh: Frankly there isn't much utility in using an area, and we've decided to move away from it. The only reason this uses an array is legacy, the transparent release parser expects one at the (current) version which this script creates provenance file for.

jul-sh: I'll add a comment to make that clearer, agreed the code looks a bit confusing at this point.

.github/workflows/provenance.yaml
jul-sh: I considered, but intentionally choose this way. I like that it makes target operations very clear, which imo is helpful given that we're operating on two git branches at the same time. 

I don't feel strongly about that preference though, happy to swap.

scripts/build_provenance
jul-sh: Good point, let's jettison that variable and write directly to the the file.

jul-sh:
> I don't expect we will have a generalized solution that works in 100% of cases, but perhaps it should work for projects that use a Docker image and a git repository, like ours, even though they may install different tools in the Docker image (e.g. Bazel). I don't have a concrete idea of how this would work / be used though, I'm curious if you have thought about that already?

@tiziano88 Good point, I've thought about this a little bit already. I agree that a generalized solution probably isn't what we want, as at that point we'd arguably be writing what basically amounts to an OSS version of GitHub action yaml files. 

There might be some things we can do in terms of aliasing docker's API, but for the initial step I wanted to keep this scoped to just our repo.

jul-sh:
@rbehjati @tiziano88 

And yes, intent wise I think this comes very close to the handwritten TOML file approach we took initially. Not quite the same (eg we wouldn't want to clone the repo, GitHub actions does that for us).

I'm sort of cautious about jumping to an abstract, generalized solution right away. Imo the cost of each repo authoring a small script that gens provenance files to describe a build isn't that big (see this script, it's <30L). 

Yes we want a nicer way to do it down the line. But a) we'll want to make sure we get the abstraction right so we don't paint ourselves in a corner, and b) some people would still use scripts like this (in cases our abstraction doesn't cover their use-case) so we want to familiarize ourselves with that process anyways.

scripts/build_provenance
jul-sh: How comes? This isn't a bash variable, but a part of the string that makes up the toml file expected by transparent release. Afaik all the keys in it are lowercase, why wouldn't this be? See https://github.com/project-oak/transparent-release/blob/main/testdata/build.toml

cc @rbehjati

jul-sh: Oh nvm, ignore what I wrote, realized I literally have that bash declaration already 😅

.github/workflows/provenance.yaml
```diff
@@ -0,0 +1,73 @@
...
+          ./oak_functions/loader/bin/oak_functions_loader
+
+      - name: Move new provenance files to the provenance branch
+        run: cp -r ./provenance/. ./provenance-branch/
+
+      - name: Commit new provenance files to the provenance branch
+        run: |
+          git -C ./provenance-branch add .
+          git -C ./provenance-branch status
+          git -C ./provenance-branch diff --staged
```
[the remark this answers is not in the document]
jul-sh: Just for logging, that bit is copied from our reproducibility index tooling.

.github/workflows/provenance.yaml
```diff
@@ -0,0 +1,73 @@
...
+      contents: write
+
+    steps:
+      - name: Mount main branch
+        uses: actions/checkout@v2
+
+      - name: Mount provenance branch
+        uses: actions/checkout@v2
+        with:
+          ref: provenance
```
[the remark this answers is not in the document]
jul-sh: Maybe, I see why. I would personally find the use of plural in a git branch name a bit confusing though.

rbehjati:
> (eg we wouldn't want to clone the repo, GitHub actions does that for us)

Cloning the repo is optional. If you pass a directory to the root of the repo that is already cloned, the tool uses that instead of cloning the repo.

jul-sh:
@tiziano88 @rbehjati 
It seems that in-toto layouts, specifically `in-toto-run` (distinct from in-toto attestations) might offer a standard for this. Will read up on them: https://github.com/in-toto/in-toto

rbehjati:
> @tiziano88 @rbehjati It seems that in-toto layouts, specifically `in-toto-run` (distinct from in-toto attestations) might offer a standard for this. Will read up on them: https://github.com/in-toto/in-toto

Also Rekor has support for in-toto as a pluggable type: https://github.com/sigstore/rekor/blob/f0aa6cf32fc06b545b852ad1b46287fdeb74c5d4/pkg/types/intoto/intoto_schema.json
I have not yet looked into how it works, and I am not sure if it is supported on the public instance of Rekor on sigstore.dev. But it would be interesting to look into uploading signed in-toto layouts to Rekor, in future.

.github/workflows/provenance.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Build Server Provenance
+        run: |
+          ./scripts/build_provenance \
+          ./scripts/runner\ build-server \
+          ./oak_loader/bin/oak_loader
+
+      - name: Build Functions Provenance
+        run: |
+          ./scripts/build_provenance \
+          ./scripts/runner\ build-functions-server \
```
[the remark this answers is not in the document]
jul-sh: It is yeah, as this used the first param only to describe the build command. This was patched in #2519.

PR #2511, opened 2022-01-27
Add function to create Wasm module bytes

mariaschett:


oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -257,11 +257,17 @@ pub fn assert_response_body(response: Response, expected: &str) {
     )
 }
 
-/// Create some valid Wasm bytecode only calling Abi functions `read_request` and `write_request`.
-pub fn create_echo_wasm_module_bytes() -> Vec<u8> {
+/// Create Wasm bytecode from manifest path relative from the `oak_functions` directory.
+fn create_wasm_module_bytes(relative_manifest_path: &str, release: bool) -> Vec<u8> {
```
rbehjati, line 261: Can this now be called in tests that use `compile_rust_wasm`? For instance: https://github.com/project-oak/oak/blob/201977e66a923a6a74cb5217bff6422423042bc2/oak_functions/loader/src/tests.rs#L114-L119.

mariaschett:
@tiziano88 @conradgrobler 

I still think that introducing an environment variable is the easiest way to get a directory from within our repo from which we can start a relative path to a Wasm module.

How about we try it for a bit on this code and see if any problems come up?

tiziano88:
An env variable in the Docker file, in the scripts, on in the Cargo config file?

mariaschett:
> An env variable in the Docker file, in the scripts, on in the Cargo config file?

in .cargo/config, i.e., [this one](https://github.com/project-oak/oak/blob/a659bad82e1f722a13eb6fec1187f4af03862605/.cargo/config.toml#L31)

tiziano88:
> > An env variable in the Docker file, in the scripts, on in the Cargo config file?
> 
> in the Cargo config file .cargo/config, i.e., [this one](https://github.com/project-oak/oak/blob/a659bad82e1f722a13eb6fec1187f4af03862605/.cargo/config.toml#L31)

But this only works if you are already running cargo [from within that folder (kind of)](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure). If you run it from outside (e.g. from `/`), then it does not work any more, which I think defeats your point?

mariaschett:
> > > An env variable in the Docker file, in the scripts, on in the Cargo config file?
> > 
> > 
> > in the Cargo config file .cargo/config, i.e., [this one](https://github.com/project-oak/oak/blob/a659bad82e1f722a13eb6fec1187f4af03862605/.cargo/config.toml#L31)
> 
> But this only works if you are already running cargo [from within that folder (kind of)](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure). If you run it from outside (e.g. from `/`), then it does not work any more, which I think defeats your point?

If we run cargo from `/`, I think not even the original solution with `std::env::current_dir()` would work. In fact, I am not sure which solution would actually work there.

My point is to replace `std::env::current_dir()` by some fixed directory from within the repo (here the Oak Functions Workspace Directory, but I don't mind much which). 

I don't think, that is any less fragile then the current solution based on `std::env::current_dir()`---but gives us some liberty in where we call runner from in the repo.

The cargo environment variable was the best I have come up with.

tiziano88:
> > > > An env variable in the Docker file, in the scripts, on in the Cargo config file?
> > > 
> > > 
> > > in the Cargo config file .cargo/config, i.e., [this one](https://github.com/project-oak/oak/blob/a659bad82e1f722a13eb6fec1187f4af03862605/.cargo/config.toml#L31)
> > 
> > 
> > But this only works if you are already running cargo [from within that folder (kind of)](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure). If you run it from outside (e.g. from `/`), then it does not work any more, which I think defeats your point?
> 
> If we run cargo from `/`, I think not even the original solution with `std::env::current_dir()` would work. In fact, I am not sure which solution would actually work there.

I think both the env variable in the Dockerfile and the env variable in the scripts would work? But I haven't thought enough about it.

tiziano88 (approved):

tiziano88:
> > env variable in the scripts
> 
> I am not sure, which env variable one you mean here?

I meant a hypothetical slightly more general version of

tiziano88:
(anyways please go ahead with this change, it is an improvement over the current state of things, even if not 100% guaranteed to work, and I think you are right that we may care less about running from `/` than about running from outside Docker)

.cargo/config.toml
```diff
@@ -26,3 +26,6 @@ rustflags = ["-C", "target-feature=+crt-static", "-C", "link-arg=-lgcc"]
 # See also https://doc.rust-lang.org/cargo/reference/config.html#httpmultiplexing
 [http]
 multiplexing = false
+
+[env]
+OAK_FUNCTIONS_WORKSPACE = { value = "oak_functions", relative = true }
```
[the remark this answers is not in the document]
conradgrobler: +1, the `WORKSPACE` part of the name confused me initially.

conradgrobler (approved):

oak_functions/sdk/test_utils/src/lib.rs
mariaschett: I think we could! 

However, I suggest before we switch here, we try working with the suggested Cargo Environment variable for a bit to see whether any unexpected problems come up.

PR #2512, opened 2022-01-27
Define relation between OakStatus and ChannelStatus

mariaschett:
OakStatus is returned by all current ABI functions (read/write_request, get_storage_item, tf, metrics, logging).

ChannelStatus by UWABI functions channel_read/write.

For operations on the Wasm memory, we now also return a `ChannelStatus`. These operations are used by both, UWABI and current ABI functions, but when the current ABI is replaced by UWABI, then we will remove OakStatus, so we decided to return a ChannelStatus.

### DIFF
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -47,12 +47,12 @@ pub mod proto {
         }
     }
 
-    // Converts a ChannelStatus into an OakStatus.
-    impl From<OakStatus> for ChannelStatus {
-        fn from(oak_status: OakStatus) -> Self {
-            match oak_status {
-                OakStatus::ErrInvalidArgs => ChannelStatus::ChannelInvalidArgs,
-                _ => ChannelStatus::Unspecified,
+    // Converts a OakStatus into an ChannelStatus.
+    impl From<ChannelStatus> for OakStatus {
+        fn from(channel_status: ChannelStatus) -> Self {
+            match channel_status {
+                ChannelStatus::ChannelInvalidArgs => OakStatus::ErrInvalidArgs,
+                _ => OakStatus::Unspecified,
             }
         }
     }
--- a/oak_functions/abi/src/lib.rs
+++ b/oak_functions/abi/src/lib.rs
@@ -47,7 +47,7 @@ pub mod proto {
         }
     }
 
-    // Converts a OakStatus into an ChannelStatus.
+    // Converts an OakStatus into a ChannelStatus.
     impl From<ChannelStatus> for OakStatus {
         fn from(channel_status: ChannelStatus) -> Self {
             match channel_status {
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -440,14 +440,25 @@ impl wasmi::Externals for WasmState {
     ) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
         match index {
             READ_REQUEST => {
-                map_host_errors(self.read_request(args.nth_checked(0)?, args.nth_checked(1)?))
+                oak_to_wasmi(self.read_request(args.nth_checked(0)?, args.nth_checked(1)?))
             }
             WRITE_RESPONSE => {
-                map_host_errors(self.write_response(args.nth_checked(0)?, args.nth_checked(1)?))
+                oak_to_wasmi(self.write_response(args.nth_checked(0)?, args.nth_checked(1)?))
             }
             WRITE_LOG_MESSAGE => {
-                map_host_errors(self.write_log_message(args.nth_checked(0)?, args.nth_checked(1)?))
+                oak_to_wasmi(self.write_log_message(args.nth_checked(0)?, args.nth_checked(1)?))
             }
+            CHANNEL_READ => uwabi_to_wasmi(self.channel_read(
+                args.nth_checked(0)?,
+                args.nth_checked(1)?,
+                args.nth_checked(2)?,
+            )),
+            CHANNEL_WRITE => uwabi_to_wasmi(self.channel_write(
+                args.nth_checked(0)?,
+                args.nth_checked(1)?,
+                args.nth_checked(2)?,
+            )),
+
             _ => {
                 let mut extensions_indices = self
                     .extensions_indices
@@ -460,7 +471,7 @@ impl wasmi::Externals for WasmState {
                     }
                     None => panic!("Unimplemented function at {}", index),
                 };
-                let result = map_host_errors(extension.invoke(self, args)?);
+                let result = oak_to_wasmi(extension.invoke(self, args)?);
                 self.extensions_indices = Some(extensions_indices);
                 result
             }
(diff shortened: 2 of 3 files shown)


oak_functions/abi/src/lib.rs
```diff
@@ -47,12 +47,12 @@ pub mod proto {
         }
     }
 
-    // Converts a ChannelStatus into an OakStatus.
-    impl From<OakStatus> for ChannelStatus {
-        fn from(oak_status: OakStatus) -> Self {
-            match oak_status {
-                OakStatus::ErrInvalidArgs => ChannelStatus::ChannelInvalidArgs,
-                _ => ChannelStatus::Unspecified,
+    // Converts a OakStatus into an ChannelStatus.
```
conradgrobler, line 50: nit: "a" and "an" should also be swapped.

oak_functions/abi/src/lib.rs
```diff
@@ -47,12 +47,12 @@ pub mod proto {
...
+    impl From<ChannelStatus> for OakStatus {
+        fn from(channel_status: ChannelStatus) -> Self {
+            match channel_status {
+                ChannelStatus::ChannelInvalidArgs => OakStatus::ErrInvalidArgs,
+                _ => OakStatus::Unspecified,
```
conradgrobler, line 55: It seems like it might be an issue in future if there is not a channel status related to and internal error if all statusses will be channel statusses in future.

oak_functions/loader/src/server.rs
```diff
@@ -816,15 +831,26 @@ fn oak_functions_resolve_func(field_name: &str) -> Option<(usize, wasmi::Signatu
...
 
+/// A helper function to move between our specific result type `Result<(), ChannelStatus>` and the
+/// `wasmi` specific result type `Result<Option<wasmi::RuntimeValue>, wasmi::Trap>`, mapping:
+/// - `Ok(())` to `Ok(Some(ChannelStatus::Ok))`
+/// - `Err(x)` to `Ok(Some(x))`
+fn uwabi_to_wasmi(
+    result: Result<(), ChannelStatus>,
+) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
+    Ok(Some(wasmi::RuntimeValue::I32(result.map_or_else(
+        |x: ChannelStatus| x as i32,
```
conradgrobler, line 849: Will it cause a problem that the same status (e.g. invalid args) will be represented by different i32 values?

oak_functions/loader/src/tf.rs
```diff
@@ -190,7 +190,8 @@ fn tf_model_infer(
     let buf_ptr = wasm_state.alloc(encoded_inference.len() as u32);
     wasm_state.write_buffer_to_wasm_memory(&encoded_inference, buf_ptr)?;
     wasm_state.write_u32_to_wasm_memory(buf_ptr, inference_ptr_ptr)?;
-    wasm_state.write_u32_to_wasm_memory(encoded_inference.len() as u32, inference_len_ptr)
+    wasm_state.write_u32_to_wasm_memory(encoded_inference.len() as u32, inference_len_ptr)?;
+    Ok(())
```
conradgrobler, line 194: Why is this change needed?

oak_functions/abi/src/lib.rs
```diff
@@ -47,12 +47,12 @@ pub mod proto {
...
     }
 
-    // Converts a ChannelStatus into an OakStatus.
-    impl From<OakStatus> for ChannelStatus {
-        fn from(oak_status: OakStatus) -> Self {
-            match oak_status {
-                OakStatus::ErrInvalidArgs => ChannelStatus::ChannelInvalidArgs,
-                _ => ChannelStatus::Unspecified,
+    // Converts a OakStatus into an ChannelStatus.
+    impl From<ChannelStatus> for OakStatus {
```
tiziano88, line 51: I know this will go away at some point, but it seems backwards to me. Should we not have the functions return ChannelStatus, and convert to OakStatus?

tiziano88 (approved):

oak_functions/loader/src/server.rs
```diff
@@ -816,15 +831,26 @@ fn oak_functions_resolve_func(field_name: &str) -> Option<(usize, wasmi::Signatu
 /// `wasmi` specific result type `Result<Option<wasmi::RuntimeValue>, wasmi::Trap>`, mapping:
 /// - `Ok(())` to `Ok(Some(OakStatus::Ok))`
 /// - `Err(x)` to `Ok(Some(x))`
-fn map_host_errors(
-    result: Result<(), OakStatus>,
-) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
+fn oak_to_wasmi(result: Result<(), OakStatus>) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
```
rbehjati, line 834: Could this be `oak_status_to_wasmi_value`?

oak_functions/loader/src/server.rs
mariaschett: I am not sure, I'll investigate and follow up.

oak_functions/abi/src/lib.rs
mariaschett: Which functions should return a ChannelStatus?

oak_functions/loader/src/server.rs
mariaschett: Are you suggesting to rename the function? 

Happy to, because I am not too convinced of as it is right now, but not too sure how. I think, we are not really mapping an OakStatus but somehow between "results in Oak and Wasmi", no?

rbehjati: I agree that `map_host_errors` is a bit misleading, as it is not mapping only the errors. But also we are not mapping `oak` :) Perhaps `oak_result_to_wasmi_result` or `map_host_result`, or `map_host_result_to_wasmi`.

oak_functions/abi/src/lib.rs
tiziano88: Never mind, I actually read the proposed change the other way round, this makes sense. :)

oak_functions/abi/src/lib.rs
conradgrobler: Good point about error codes being transmitted over the channel, rather than being indicated by the channel status. I guess the current wierd state is because the transition is in-progress.

conradgrobler (approved):

oak_functions/loader/src/server.rs
mariaschett: We will statically know if we expect a ChannelStatus or an OakStatus---so this overlap should be fine!

oak_functions/loader/src/server.rs
```diff
@@ -816,13 +831,25 @@ fn oak_functions_resolve_func(field_name: &str) -> Option<(usize, wasmi::Signatu
...
+    let oak_status_from_result = result.map_or_else(|x: OakStatus| x, |()| OakStatus::Ok);
+    let wasmi_value = wasmi::RuntimeValue::I32(oak_status_from_result as i32);
+    Ok(Some(wasmi_value))
+}
+
+/// A helper function to move between our specific result type `Result<(), ChannelStatus>` and the
+/// `wasmi` specific result type `Result<Option<wasmi::RuntimeValue>, wasmi::Trap>`, mapping:
+/// - `Ok(())` to `Ok(Some(ChannelStatus::Ok))`
+/// - `Err(x)` to `Ok(Some(x))`
+fn from_channel_status_result(
```
mariaschett, line 846: @rbehjati I renamed both to `from_channel_status_result` and `from_oak_status_result` and refactored a bit. WDYT?

rbehjati: I like these names. Thanks :)

PR #2514, opened 2022-02-02
Refactor gRPC channel duration

ipetr0v:


conradgrobler (approved):

oak_functions/client/java/src/com/google/oak/functions/client/AttestationClient.java
```diff
@@ -223,9 +216,12 @@ public void finalize() {
    * decrypts the response.
    *
    * This method can only be used after the {@code attest} method has been called successfully.
+   *
+   * @param request contains a request to be sent via the attested gRPC channel.
+   * @param connectionTimeout contains an inactivity threshold for gRPC connections.
    */
   @SuppressWarnings("ProtoParseWithRegistry")
-  public Response send(Request request)
+  public Response send(Request request, Duration connectionTimeout)
```
tiziano88, line 224: Why is the connection timeout being provided for each request? Also, shouldn't all this be handled at the channel level anyways, which is already provided to the attest method?

ipetr0v: Removed

PR #2519, opened 2022-02-04
Accept an array of arguments when building the provenance

jul-sh:
This aligns the invocation of the provenance script with the provenance file, which stores build arguments as an array.

Previously the build_provenance script accepted exactly two arguments 1) exactly one build arg, 2) the output path.

Now it accepts 1) the output path, followed by one or multiple build args.

### DIFF
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -10,7 +10,7 @@ readonly OUTPUT_PATH=$1
 # build arguments.
 shift
 
-eval ./scripts/docker_run "${@}"
+./scripts/docker_run "${@}"
 readonly BINARY_SHA_256_HASH=$(sha256sum "${OUTPUT_PATH}" | cut -d " " -f 1)
 
 readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -10,7 +10,7 @@ readonly OUTPUT_PATH=$1
 # build arguments.
 shift
 
-eval "./scripts/docker_run ${@}"
+eval ./scripts/docker_run "${@}"
 readonly BINARY_SHA_256_HASH=$(sha256sum "${OUTPUT_PATH}" | cut -d " " -f 1)
 
 readonly COMMIT_HASH=$(git rev-parse --verify HEAD)
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -4,13 +4,24 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 # shellcheck source=scripts/common
 source "$SCRIPTS_DIR/common"
 
-readonly BUILD_COMMAND=$1
-readonly OUTPUT_PATH=$2
+readonly OUTPUT_PATH=$1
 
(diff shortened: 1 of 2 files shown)


scripts/build_provenance
```diff
@@ -4,14 +4,25 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
...
-readonly BUILD_COMMAND=$1
-readonly OUTPUT_PATH=$2
+readonly OUTPUT_PATH=$1
 
-eval "./scripts/docker_run ${BUILD_COMMAND}"
+# Shift arguments to omit the output path. The remaining arguments are the
+# build arguments.
+shift
+
+eval "./scripts/docker_run ${@}"
```
tiziano88, line 13: I'm surprised this works with the the quotes, does eval do its own splitting / parsing? Or maybe docker_run does

tiziano88 (approved):
LG but is there any way of testing it with the transparent release repo? Or do we postpone that anyways?

scripts/build_provenance
jul-sh: Bash arg handling is very confusing, but: In `docker_run` we use `$*` when invoking the docker CLI. Afaik `$*` effectively means args a delineated by whitespace. Hence passing as args as a joined string doesn't make difference.

jul-sh: An upside of the code I wrote in this for formatting bash args into a toml string is that it doubles as a handy tool for investigating bash arg handling :p

jul-sh: docker_run concats args into a single string by default. That string is passed to `fix_docker_user_and_run`, which executes that string using `su --session-command`. I think thats where the splitting happens?

jul-sh:
@tiziano88 RE linking it with the repo, it's planned, not postponed. Here's my planned timeline:

1) Update transparent release to use Amber SLSA files (it makes not a lot of sense to do the next steps before this, as they depend on it)
2) Update this GitHub action to produce an Amber SLSA file
3) Update our Kokoro CI action to consume Amber SLSA file that are pushed to the provenance branch, and verify them with the transparent release tooling. 

Step 3) does two things which are pretty nice: It verifies that our SLSA files are valid, B) It also adds a continuous check that our builds stay reproducible, as the Kokoro CI action will fail if the build yields a different hash inside of GitHub actions and inside of Kokoro. (Not quite "same hash on three systems in a row", but close). 😇

cc @rbehjati on this too.

rbehjati:
> Update our Kokoro CI action to consume Amber SLSA files that are pushed to the provenance branch, and verify them with the transparent release tooling.

Just to clarify: Do you mean that the Kokoro CI action could verify all the provenances? This is what I understand from your point B on "It adds a continuous check...".

jul-sh:
@rbehjati It would be possible, yeah. Note that provenances are only generated for commits in main, so basically it would run for each commit in main to verify the provenance. 

We can tweak that if it's too much of a load, but it might not be much heavier than the Actions we already run.

jul-sh:
I'd be inclined towards running it right away, and changing that approach in case of problems.  It uses a service, so I think that there wouldn't be an impact on the speed / availability of GitHub runners executing our PR CI actions.

PR #2521, opened 2022-02-09
Replace BoxedUwabiExtension using trait bounds

mariaschett:
This will allow us to add a function `f(mut self: Box<Self>)` to the `UwabiExtension` trait.


tiziano88 (approved):
Good to know!

PR #2522, opened 2022-02-09
Add comment on clearing of Channel Switchboard

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -908,6 +920,16 @@ impl ChannelSwitchboard {
     fn get_mut(&mut self, channel_handle: &ChannelHandle) -> Option<&mut Endpoint> {
         self.0.get_mut(channel_handle)
     }
+
+    // By removing all endpoints from the ChannelSwitchboard their connected endpoints will know
+    // that they will not receive any more messages and get an error when sending messages.
+    fn clear(&mut self) {
```
tiziano88, line 926: Maybe call this `close_all` or similar?

tiziano88 (approved):

oak_functions/loader/src/server.rs
mariaschett: Because `close` is also used to close the Receiver in tokio::mpsc channel, I looked for something different to avoid confusion. 

How about we `unregister_all`---analogous to `register`ing a handle?

mariaschett: But we do two things: 
(1) we close the receiver of every endpoint, and we also 
(2) drop all endpoints by draining the hash table. 

(1) might not strictly be necessary, but seems to be good practice. 

I think that (2) might be lost when we call it close, so I'd go with `unregister_all`, if you don't object.

oak_functions/loader/src/server.rs
```diff
@@ -908,6 +920,16 @@ impl ChannelSwitchboard {
...
+        // Closing the receivers of the endpoint is good practice.
+        for endpoint in self.0.values_mut() {
```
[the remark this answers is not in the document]
tiziano88: Yes, as per https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.drain .

oak_functions/loader/src/server.rs
tiziano88: I'm a bit confused though, why do 1 if it's not necessary? The canonical way of closing a mpsc channel is for it to go out of scope, not to call close on it, right? In fact, close seems to be there for the specific use case of wanting to close the channel, but still go through the buffered messages; is that what we need here? If not, I suggest just dropping the receivers by naturally letting them go out of scope. Which in fact probably does not need any code?

mariaschett: For a [clean shutdown](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#clean-shutdown) we should close the _receiver_ in the endpoint. But you are right, we are not consuming the messages. For the sender in the endpoint, we can just let it got out of scope/drop it. 

I'd prefer to explicitly have a function to drop the Wasm module endpoints to control when that happens---especially as this affects the Runtime endpoints.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
mariaschett: Yes, I am using `drain` to drain the hash map already. But I can't easily manipulate the values of the hashmap and the hashmap at the same time. What am I missing?

oak_functions/loader/src/server.rs
tiziano88: We don't need clean shutdown, because we don't process messages after closing a channel (I think). So yes, perhaps just a function to [clear](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.clear) the switchboard is enough?

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
tiziano88: I think you are using drain to actually clear the hashmap at the moment (which you could do with https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.clear ). If you wanted to iterate on the elements, does this work?

```rust
for endpoint in self.0.drain() {
  ....
}
```

That said, as per my other message, I don't think we need anything more than just clearing the map anyways, so I don't think we need the for loop at all, right?

oak_functions/loader/src/server.rs
```diff
@@ -908,6 +920,16 @@ impl ChannelSwitchboard {
     fn get_mut(&mut self, channel_handle: &ChannelHandle) -> Option<&mut Endpoint> {
         self.0.get_mut(channel_handle)
     }
+
+    // By removing all endpoints from the ChannelSwitchboard their connected endpoints will know
+    // that they will not receive any more messages and get an error when sending messages.
+    fn unregister_all(&mut self) {
```
mariaschett, line 926: @tiziano88 WDYT?

oak_functions/loader/src/server.rs
tiziano88: I do think this is misleading, in that if you remove this line, nothing observable would change (i.e. you would not even be able to write a meaningful test for it), because the endpoints are actually closed elsewhere. I would at the very least add a comment explaining this perhaps? I think the clean shutdown would make sense if we decide to keep consuming those messages, but that's not the current design, so I think it is better to drop it.

oak_functions/loader/src/server.rs
tiziano88: I still think this is unnecessary, but perhaps it is not too bad to keep this method that just clears the switchboard, but again with comments saying that it is not actually necessary, etc.?

oak_functions/loader/src/server.rs
mariaschett: You convinced me!

I tried to find a use case for a clean shutdown, but I couldn't find one. After the Wasm module finishes, I see no reason to handle its unread messages. 

So I replaced `unregister_all` by a comment. PTAL.

PS: I could not even find a reason to log the unread messages (cf. previous commit). It may be perfectly fine for the Wasm module to not read messages, e.g., if it sends the same request to two extensions and receives one first.

oak_functions/loader/src/server.rs
```diff
@@ -1024,6 +1030,20 @@ mod tests {
...
+    async fn test_drop_channel_switchboard_stops_recv() {
+        let mut channel_switchboard = ChannelSwitchboard::new();
+        let mut endpoint_2 = channel_switchboard.register(ChannelHandle::Testing);
+
+        let stopped_receiving = tokio::spawn(async move {
+            endpoint_2.receiver.recv().await;
+            true
+        });
+
+        std::mem::drop(channel_switchboard);
```
tiziano88, line 1043: Could this be in an inner scope and just be dropped naturally? Just so it's closer to what actually happens, but feel free to ignore if it complicatees things.

tiziano88 (approved):

PR #2523, opened 2022-02-09
Add Remote Attestation description

ipetr0v:
This change adds a description of the Remote Attestation protocol used by Oak.
Current version only describes a unidirectional attestation.


docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+   - Client application that connects to the Server
+1. TEE Platform
+   - Firmware that provides a TEE support (i.e. Intel SGX or AMD-SEV-SNP capable
+     CPU)
+   - Possesses a firmware key
+1. TEE Provider
+   - TEE platform provider that can confirm authenticity of the TEE Platform
+     firmware keys using its root key
+   - Client must have the TEE Provider's root public key
+   - It’s important to note that the Provider is an external server (i.e.
```
conradgrobler, line 75: It is not always required to connect to an external server to validate the attestation report. Only one of the SGX attestation models requires this as far as I remember.

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+1. **Client** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
+1. **Client** sends a `ClientHello` message to the **Server**
+   - Which includes a random string
+1. **Server** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
+1. **Server** sends a request for the `AttestationReport` to the **TEE
+   Platform**
+   - This request contains
+     [SHA-256](https://datatracker.ietf.org/doc/html/rfc6234) hash of the
```
conradgrobler, line 95: This request should probably also contain the client random string so the client can validate that it is was generated for this specific request.

conradgrobler (approved):
Nice!

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
+# Remote Attestation
+
+A crucial advantage of a Trusted Execution Environment (TEE) is the ability to
+perform **Remote Attestation**. It allows software to remotely check the TEE
+platform authenticity and also provides them with the information about the code
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+running inside the TEE. And if the code of the application is **reproducibly
+buildable**, software can check that the TEE platform is running the application
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+One of the main components used in the remote attestation process is an
+**Attestation report**, which is a data structure signed by the TEE platform and
+containing information identifying the code that is running inside the TEE. This
+report can then be checked and signed by the **TEE Provider** (e.g. AMD or
+Intel), which results in evidence that the code is running on a genuine TEE
+platform.
+
+## Overview
+
+Remote Attestation protocol is implemented as a simplification of the
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+     pair
+   - **TEE Provider** creates a signature of the **TEE Platform**'s firmware key
+   - **Client** is provided with a **TEE Provider**'s _Root_ key
+1. **Client** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
+1. **Client** sends a `ClientHello` message to the **Server**
+   - Which includes a random string
+1. **Server** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
+1. **Server** sends a request for the `AttestationReport` to the **TEE
```
rbehjati, line 92: For my own understanding... what type of request is it? Is it some sort of IPC or RPC?

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+     [AES-256-GCM](https://datatracker.ietf.org/doc/html/rfc5288) _Session_
+     keys:
+     - Server Session Key
+     - Client Session Key
+   - **Client** and **Server** use Authenticated Encryption/Decryption
+     ([AEAD](https://en.wikipedia.org/wiki/Authenticated_encryption)) for
+     communication
+1. **Client** and **Server** use generated _Session_ keys for communication
+   between them
+   - _Note_: for each new request **Client** performs the Remote Attestation
```
rbehjati, line 142: nit: So I think the end goal of the remote attestation protocol is for the client and the server to validate each other's identities, and create session keys, that they can use for encrypting the requests and response. If so, I think it is better to have this note as a separate paragraph, not as an indented item in the list.

rbehjati (approved):
Lovely! Thanks. I was looking for this :)

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+
+A crucial advantage of a Trusted Execution Environment (TEE) is the ability to
+perform **Remote Attestation**. It allows software to remotely check the TEE
+platform authenticity and also provides the remote software with the information
+about the code running inside the TEE. And if the code of the application is
+**reproducibly buildable**, the software can check that the TEE platform is
+running the application that is expected to run.
+
+One of the main components used in the remote attestation process is an
+**Attestation report**, which is a data structure signed by the TEE platform and
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+that are used during the _Data Exchange_ stage:
+
+- Server Session Key
+  - Which is used to encrypt messages sent by the Server
+- Client Session Key
+  - Which is used to encrypt messages sent by the Client
+
+## Workflow
+
+The workflow of the Remote Attestation protocol involves 3 interacting entities:
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+   - Client application that connects to the Server
+1. TEE Platform
+   - Firmware that provides a TEE support (i.e. Intel SGX or AMD-SEV-SNP capable
+     CPU)
+   - Possesses a firmware key
+1. TEE Provider
+   - TEE platform provider that can confirm authenticity of the TEE Platform
+     firmware keys using its root key
+   - Client must have the TEE Provider's root public key
+   - It’s important to note that the Provider is an external server (i.e.
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
ipetr0v: I think it depends on the implementation. But this request is sent to a `/dev/${tee_name}` file.

docs/remote-attestation.md
```diff
@@ -0,0 +1,147 @@
...
+     belongs to Intel or AMD)
+
+The complete workflow of the Remote Attestation protocol looks as follows:
+
+1. Pre-handshake steps that are performed at startup:
+   - Both **Client** and **Server** generate an individual
+     [ECDSA-P256](https://datatracker.ietf.org/doc/html/rfc6979) _Signing_ key
+     pair
+   - **TEE Provider** creates a signature of the **TEE Platform**'s firmware key
+   - **Client** is provided with a **TEE Provider**'s _Root_ key
```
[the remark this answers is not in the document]
ipetr0v: It's baked into the client. Moved out of the list.

docs/remote-attestation.md
ipetr0v: Moved to the previous paragraph

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+     firmware keys using its root key
+   - Client must have the TEE Provider's root public key
+   - It’s important to note that the TEE Provider is an external server (i.e.
+     belongs to Intel or AMD)
+
+The complete workflow of the Remote Attestation protocol looks as follows:
+
+1. Pre-handshake steps that are performed at startup:
+   - Both **Client** and **Server** generate an individual
+     [ECDSA-P256](https://datatracker.ietf.org/doc/html/rfc6979) _Signing_ key
```
tiziano88, line 86: I didn't realise that the signing key is P256. Then I guess it is a no-brainer to also use P256 for DH key agreement later on, in fact it would strictly reduce the number of curves that need to be trusted.

@jul-sh @conradgrobler WDYT?

docs/remote-attestation.md
tiziano88: But I agree with @conradgrobler , if we don't pass the random string to the TEE, then the client cannot verify the freshness of the TEE signature, only that of the server, right?

tiziano88 (approved):
Excellent! I wonder if it would be possible to somehow make the Rust implementation refer to this, or the other way round, or something in between. Does anyone have any idea how to do something like that?

ipetr0v:
> I wonder if it would be possible to somehow make the Rust implementation refer to this

I have a smaller `README.md` (in this PR) file in the `remote_attestation/` that has a link to this documentation.

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+[ECDSA-P256](https://datatracker.ietf.org/doc/html/rfc6979) keys called
+**Signing Keys**. These keys are persistent between connections (but regenerated
+at startup) and are used to sign **Transcripts**: hashes of previous handshake
+messages that are sent in each new handshake message to prevent
+[Replay Attacks](https://en.wikipedia.org/wiki/Replay_attack). Each handshake
+message also includes a random string so that transcripts always contain new
+random values.
+
+Also for each individual connection both Client and Server generate a pair of
+[X25519](https://datatracker.ietf.org/doc/html/rfc7748) Diffie-Hellman keys
```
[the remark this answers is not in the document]
ipetr0v: Done

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+[AES-256-GCM](https://datatracker.ietf.org/doc/html/rfc5288) **Session Keys**
+that are used during the _Data Exchange_ stage:
+
+- Server Session Key
+  - Which is used to encrypt messages sent by the Server
+- Client Session Key
+  - Which is used to encrypt messages sent by the Client
+
+It's important to node that for each new request **Client** performs the Remote
+Attestation from the start in order to create a new pair of Session keys.
```
jul-sh, line 59: Aren't these just like the same key? The client and server both have a copy, but of the same key.

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+   - It’s important to note that the TEE Provider is an external server (i.e.
+     belongs to Intel or AMD)
+
+The complete workflow of the Remote Attestation protocol looks as follows:
+
+1. Pre-handshake steps that are performed at startup:
+   - Both **Client** and **Server** generate an individual
+     [ECDSA-P256](https://datatracker.ietf.org/doc/html/rfc6979) _Signing_ key
+     pair
+   - **TEE Provider** creates a signature of the **TEE Platform**'s firmware key
```
jul-sh, line 88: firmware _public_ key?

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+1. **Client** and **Server** use both _Ephemeral_ public keys to create an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Shared Secret_
+1. **Client** and **Server** derive _Session_ keys from the _Shared Secret_ and
+   both _Ephemeral_ public keys using a _Key Derivation Function_
+   ([HKDF](https://datatracker.ietf.org/doc/html/rfc5869))
+   - Each side generates 2
+     [AES-256-GCM](https://datatracker.ietf.org/doc/html/rfc5288) _Session_
+     keys:
+     - Server Session Key
+     - Client Session Key
```
jul-sh, line 139: Each side has both of them? I'm a bit confused by this, isn't AES 256 symmetric? Do we use two symmetric keys?

jul-sh (commented):
Nice :)

docs/remote-attestation.md
ipetr0v: These are different keys `K_s` and `K_c`.
But Client and server have both of them.

`K_s` is used by the Server to encrypt outgoing messages, and by the Client to decrypt incoming messages.
`K_c` is used by the Client to encrypt outgoing messages, and by the Server to decrypt incoming messages.

This is necessary to prevent the attacker from sending the message back to the sender.

docs/remote-attestation.md
```diff
@@ -0,0 +1,148 @@
...
+   - Both **Client** and **Server** generate an individual
+     [ECDSA-P256](https://datatracker.ietf.org/doc/html/rfc6979) _Signing_ key
+     pair
+   - **TEE Provider** creates a signature of the **TEE Platform**'s firmware key
+1. **Client** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
+1. **Client** sends a `ClientHello` message to the **Server**
+   - Which includes a random string
+1. **Server** generates an
+   [X25519](https://datatracker.ietf.org/doc/html/rfc7748) _Ephemeral_ key pair
```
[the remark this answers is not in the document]
ipetr0v: Ephemeral keys are used to create a shared Diffie-Hellman secret, and Signing keys basically work as TEE proxy keys for signing protocol transcripts, i.e. TEE only sings a signing key - and it's then used to sign message transcripts.

docs/remote-attestation.md
ipetr0v: These are 2 different symmetric keys (see a previous response).

docs/remote-attestation.md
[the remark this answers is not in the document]
jul-sh: Thanks for the context, makes sense

PR #2524, opened 2022-02-09
Generate Amber SLSA provenance files

jul-sh:
Updates the generated provenance files to use the Amber SLSA build type we defined for transparent release

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -49,12 +49,14 @@ jobs:
       - name: Build Server Provenance
         run: |
           ./scripts/build_provenance \
+          "oak_loader" \
           ./oak_loader/bin/oak_loader \
           ./scripts/runner build-server
 
       - name: Build Functions Provenance
         run: |
           ./scripts/build_provenance \
+          "oak_functions" \
           ./oak_functions/loader/bin/oak_functions_loader \
           ./scripts/runner build-functions-server
 
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -4,13 +4,19 @@ readonly SCRIPTS_DIR="$(dirname "$0")"
 # shellcheck source=scripts/common
 source "$SCRIPTS_DIR/common"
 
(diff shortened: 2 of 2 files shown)


jul-sh:
Questions worth discussing:

- Am I in trouble for continuing to do this in bash? 😅
- Should we delete or keep the previously generated files on the provenance branch?
- Do we want to validate the schema shape at creation time, or is it enough to do that when verifying it later in CI, as proposed in https://github.com/project-oak/oak/pull/2519#issuecomment-1030234644 ?

@tiziano88 @rbehjati

tiziano88:
> * Am I in trouble for continuing to do this in bash? sweat_smile

My suggestion is still to switch to a real programming language sooner rather than later, but I'll leave that to you to decide, as long as the bash code does not get much more complex than it currently is (see below).



> * Should we delete or keep the previously generated files on the provenance branch?

We can leave them there for now, and delete them manually at some later point, perhaps? I would not worry too much about it. If this was a general purpose CAS, it would be fine to keep adding stuff to it, no need to go back and delete entries, even if they are not relevant. But I agree it would be nice to clean things up at some point. Maybe create another low-priority task to pick up on a slow day :)

> * Do we want to validate the schema shape at creation time, or is it enough to do that when verifying it later in CI, as proposed in [Accept an array of arguments when building the provenance #2519 (comment)](https://github.com/project-oak/oak/pull/2519#issuecomment-1030234644) ?

I think as long as it's checked in CI at some point before merging that's fine. I would definitely not try to implement this in Bash, maybe when we have a Go / Rust tool, it would be easier to generate and validate at the same time (or perhaps the added type safety would make the validation not really necessary? not sure)

rbehjati (approved):
LGTM.

> Do we want to validate the schema shape at creation time, or is it enough to do that when verifying it later in CI, as proposed in https://github.com/project-oak/oak/pull/2519#issuecomment-1030234644 ? 

I agree that it can be done later in the CI. Should be straightforward when we have 
https://github.com/project-oak/transparent-release/issues/19.

PR #2527, opened 2022-02-11
Remove UWABI extensions from WasmState

mariaschett:
When we start tasks for UWABI extensions we need to move the UWABI extension to the task, so we don't want to hold them in the WasmState.


tiziano88 (approved):
I'm not entirely sure what "own themselves" means, but as long as you think it helps with other changes, this seems fine? Except that it's usually not super idiomatic to return anonymous tuples.

mariaschett:
Updated description to remove "own themselves".

I see your point with tuples!

It is similar to returning a [tuple of endpoints when creating a channel](https://github.com/project-oak/oak/blob/8b1eda180e83565a0ff516131d23166f7704778e/oak_functions/loader/src/server.rs#L878): the two---channel switchboard & UWABI extensions) are linked together.

Let me know if you think it shoud change.

PR #2532, opened 2022-02-14
Run Uwabi Extensions

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -965,6 +968,19 @@ mod tests {
                 self.endpoint = Some(endpoint);
             }
         }
+
+        fn run(mut self: Box<Self>) -> tokio::task::JoinHandle<()> {
+            tokio::spawn(async move {
```
tiziano88, line 973: Pass-by comment: I suggest to not have the extension do the spawning, instead just return a Future and let the caller spawn it in whatever executor it uses.

mariaschett: Thank you---yes, that makes sense!

oak_functions/loader/src/server.rs
```diff
@@ -1183,45 +1227,57 @@ mod tests {
...
+        uwabi_extensions.remove(pos_of_extension)
+    }
 
-        // Get endpoint from the extension.
-        extension
-            .get_endpoint_mut()
-            .expect("No endpoint set for extension.")
+    // Helper function for testing to make an UWABI message human readable in code.
+    fn from_uwabi_message(uwabi_message: &UwabiMessage) -> String {
+        String::from_utf8(uwabi_message.clone())
```
tiziano88, line 1275: The UTF8 parsing / conversion seems unnecessary / misleading, given messages are just uninterpreted byte strings, but up to you if you think it adds any value.

tiziano88 (approved):

oak_functions/loader/src/server.rs
mariaschett: Ah, I see how it can be misleading. Maybe `from_uwabi_message/to_uwabi_message` should be added to the `UwabiExtension` trait? Because I do, in fact, interpret the byte as strings in the `run` method of the testing extension.

If you don't mind much, I would leave it for this PR but revisit it as soon as we get to the IDL. WDYT?

tiziano88: I would not even interpret the byte as strings, that's also not what an IDL would do. Perhaps look at the first byte and switch on that value, then decide what to do with the rest of the message? Then it's obvious how that would be replaced by e.g. some code generated from the IDL, in principle.

PR #2536, opened 2022-02-14
Remove most code related to "classic" Oak

tiziano88:
Not used any more, and just taking up space and mental overhead, plus
additional CI time and compatibility burden when upgrading deps.


tiziano88:
Note for reviewers:

I would like to get approval from all of you before merging.

I will remove the now-unused "required" checks from CI just before submitting.

There are a lot of files removed, and the GitHub UI struggles a bit; I suggest waiting a few seconds until all the diffs are loaded, whlie in files changed view, then look at the few files that got their contents changed, and then read through the names of the deleted files. If you think anything should not be deleted, please let me know, probably best if in a top-level comment as part of the review reply.

Thanks!

ipetr0v:
Can we just move this into a separate repo like `project-oak/oak-classic` or `project-oak/oak-archive` ?
Because we might still use some of it at some point.

Just too sad to completely remove so much work.

jul-sh (approved):
LGTM. I would think of it less as "removing it forever" and more as in "setting it to rest in the version history, waiting for the right time to be resurrected"

tiziano88:
> Did we already close the issues related to this code only?

Not yet, I think there are still a fair number of issues that we can close after merging this.

tiziano88:
> Can we also simplify our [Dockerfile](https://github.com/project-oak/oak/blob/main/Dockerfile)?

Yes but I would prefer doing that in a separate PR.

mariaschett (approved):
Amazing! -84k LoC.

oak_functions/README.md
```diff
@@ -137,39 +137,6 @@ requirements, before publishing the results by logging to stdout.
 The Remote Attestation protocol implemented in Oak is currently integrated in
 Oak Functions.
 
-## Comparison with Oak
```
rbehjati: I think it would be useful to add some details from the deleted documentation here. For instance how to run an application or create configuration files. There is also some technical background that could be relevant, for instance about the use of WebAssembly. I'd suggest creating an issue to do this in a separate PR.

runner/src/internal.rs
```diff
@@ -55,18 +55,15 @@ pub struct Opt {
 
 #[derive(StructOpt, Clone, Debug)]
 pub enum Command {
-    RunExamples(RunExamples),
```
rbehjati: This files also has a server variant related to the introspection-client that caused some complication, particularly related to running tests for all features. That, and the code related to it, can be removed too.

rbehjati (approved):
Great. Thanks. I am very happy to see that prost in third_party is gone. We should celebrate this 🎉 

I think in the top-level README we have some links to some of the deleted documentation. Perhaps create an issue to fix the documentation after this is merged.

scripts/build_gh_pages
```diff
@@ -34,51 +34,21 @@ readonly TARGET_ABS_PATH="$(realpath "${TARGET_DIR}")"
 
 # All the dirs that we want to generate docs for.
 declare -ar SOURCE_PATHS=(
-  'oak_abi'
-  'oak_client'
-  'oak_derive'
   'oak_functions/loader'
   'oak_functions/sdk'
```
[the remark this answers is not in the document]
tiziano88: Fixed doc generation now.

scripts/common
```diff
@@ -30,25 +30,6 @@ readonly SERVER_DOCKER_IMAGE_NAME='gcr.io/oak-ci/oak-server'
 readonly CACHE_DIR='bazel-cache'
 readonly SERVER_BIN_DIR="${PWD}/oak_loader/bin"
 
-# List all workspaces and stand-alone crates.
-declare -ar ALL_CRATES=(
```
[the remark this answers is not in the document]
tiziano88: Not any more.

oak_functions/README.md
tiziano88: Thanks, created https://github.com/project-oak/oak/issues/2541

PR #2539, opened 2022-02-15
Add README for Oak Functions Loader

mariaschett:
To have the loader crate more self contained I added a seperate README and moved content from the Oak Functions README.


oak_functions/loader/README.md
```diff
@@ -0,0 +1,84 @@
+# Oak Functions Loader
+
+The _Oak Functions loader_ is responsible for starting the Oak Functions trusted
+runtime, and loading an application with a single Wasm module as the workload.
+The trusted runtime instantiates the given Wasm module for each incoming user
+request, runs the request through it in order to produce a response, and then
+terminates the Wasm instance; each Wasm instance is short lived and cannot
+persist state outside of the request lifetime.
+
+As part of our shift to a distributed runtime, in the future, we may allow
```
mariaschett, line 10: @conradgrobler Is this paragraph still accurate? Or shall we remove it?

conradgrobler: This is still part of the future plans, but I don't think this will be implemented in the near future. Also, it is possible that such native instance would actually be implemented as different loaders. So, I think it is fine to remove now and add details in future when we implement this type of support.

oak_functions/README.md
```diff
@@ -50,89 +50,7 @@ from the conventional computing model.
 <!-- https://docs.google.com/drawings/d/1ZPeJ93IkyOOJVI8CFSbEeEKn6wVozB-d6E1SekK2QyQ/edit -->
 <img src="images/ComputingModel.png" width="1000">
 
-## Features
```
conradgrobler, line 53: Perhaps add a paragraph here with a link to the loader directory to clarify how this is actually used?

conradgrobler (approved):

oak_functions/README.md
mariaschett: Good point, done, thank you!

PR #2542, opened 2022-02-15
Align the generated provenances with the transparent release CLI

jul-sh:
- updates the buildType key to match https://github.com/project-oak/transparent-release/pull/26
- simplifies the docker configuration to address https://github.com/project-oak/transparent-release/issues/25

Down the road we probably want to change to change how transparent-release/amber-provenances captures the the docker config used, see @rbehjati thoughts in https://github.com/project-oak/transparent-release/pull/26. We may also want to use the transparent release CLI itself to generate provenance statements once it supports doing so (see https://github.com/project-oak/transparent-release/blob/main/common/common.go#L250). 

For now this aligns how provenances are built with how they are verified by the transparent release CLI. Conceptually that means building the binaries with minimal env flags & no caches proxied from the host. The conceptual purpose of `scripts/docker_run` becomes that of a convenience script for developing, offering extras such as using host caches, and host user file ownership.

### DIFF
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -14,7 +14,7 @@ shift 2
 # docker user and does not pass ENV vars. Flags are intentionally a subset of
 # the ones used in `./scripts/docker_run` which proxies host cache dirs and
 # file system permissions for developer convenience.
-docker run --rm --tty --volume=$PWD:/workspace --workdir=/workspace "$DOCKER_IMAGE_ID" "${@}"
+docker run --rm --tty "--volume=${PWD}:/workspace" --workdir=/workspace "${DOCKER_IMAGE_ID}" "${@}"
 
 # Get digest of the docker image, as defined in https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests
 readonly DOCKER_DIGEST=$(docker images --digests --format="{{.Digest}}" "${DOCKER_IMAGE_REPO_DIGEST}")
--- a/scripts/build_provenance
+++ b/scripts/build_provenance
@@ -10,7 +10,11 @@ readonly OUTPUT_PATH=$2
 # Shift arguments so that the remaining arguments are the just build arguments.
 shift 2
 
-./scripts/docker_run "${@}"
+# Run the build in a simplified docker configuration that uses the default
+# docker user and does not pass ENV vars. Flags are intentionally a subset of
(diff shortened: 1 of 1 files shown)


scripts/build_provenance
```diff
@@ -10,7 +10,11 @@ readonly OUTPUT_PATH=$2
 # Shift arguments so that the remaining arguments are the just build arguments.
 shift 2
 
-./scripts/docker_run "${@}"
+# Run the build in a simplified docker configuration that uses the default
```
rbehjati, line 13: What is the default docker user? Are the builds reproducible if we use the default docker user?

rbehjati (approved):

scripts/build_provenance
jul-sh: It's root. :) But also whichever the default docker user is, it'll be the same across here, and the transparent release CLI. 

And yeah, the provenance produced by this is successfully verified by the transparent release CLI, tested by running thie provenance GitHub action (that runs this script) on my Oak fork

jul-sh: Also, once we add kokoro to verify the generated provenances (soon, CL is ready) we'll fail GitHub checks if ever the build stops being reproducible.

scripts/build_provenance
```diff
@@ -10,7 +10,11 @@ readonly OUTPUT_PATH=$2
...
+# docker user and does not pass ENV vars. Flags are intentionally a subset of
+# the ones used in `./scripts/docker_run` which proxies host cache dirs and
+# file system permissions for developer convenience.
```
tiziano88, line 16: Nice description, thanks for taking the time to write down the why (and not just the what).

tiziano88 (approved):

scripts/build_provenance
rbehjati: Is that a post-merge check?

jul-sh: atm it would be, but we can decide to make it pre merge if using CI runner resources for it is fine :) It probs would be

rbehjati: Thanks. I think post-merge is fine for now. I wonder how it shows up on the CI, since it is triggered by the changes in the provenance branch, I think.

jul-sh: Yup atm it would be. It'd show the same way checks show on the main branch, as a checkmark or cross icon next to each commit. Down the road we should increase the visibility of that check failing

Issue #2544, opened 2022-02-15
Update rustls


tiziano88:
Blocked on https://github.com/hyperium/tonic/issues/891

conradgrobler:
We are no longer using TLS for our GRPC tests, so this is no longer relevant.

PR #2545, opened 2022-02-15
Update deps

### LINKED ISSUE #2476: Update to `prost 0.9.0`
Likely involves updating related dependencies such as tonic, which no longer requires (or implements) Sync for bidirectional streaming from v0.5 to 0.6, so also not a straightforward upgrade (which is useful, because it allows to drop the dependency on the sync_wrapper crate and simplify some of our logic)

tiziano88:
Fix dependabot

Fixes #2476


experimental/https_attestation/Cargo.toml
```diff
@@ -36,4 +37,4 @@ tokio = { version = "*", features = [
   "sync",
   "rt-multi-thread"
 ] }
-tokio-rustls = "*"
+tokio-rustls = "0.22"
```
rbehjati, line 40: Does the `# TODO(#2544)` apply here too?

rbehjati (approved):
Nice :)

experimental/https_attestation/src/main.rs
```diff
@@ -33,38 +33,38 @@
...
 use anyhow::{anyhow, Context};
+use clap::Parser;
 use futures::FutureExt;
 use log::info;
-use structopt::StructOpt;
 
-#[derive(StructOpt, Clone)]
-#[structopt(about = "HTTPS Attestation")]
+#[derive(Parser, Clone)]
+#[clap(about = "HTTPS Attestation")]
```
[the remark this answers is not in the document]
tiziano88: Good idea, changed everywhere.

experimental/https_attestation/Cargo.toml
tiziano88: May or may not, I hope it will be more obvious when we do the update.

tiziano88 (commented):
Thanks for the quick review

Issue #2543, opened 2022-02-15
Fix provenance generation in CI


tiziano88:
#2536 broke the provenance generation script in CI: https://github.com/project-oak/oak/runs/5205293325?check_suite_focus=true (sorry about that!)

@jul-sh is it worth considering running this also pre-merge, without actually writing the provenance anywhere (dry run), so that at least it catches such errors?

jul-sh:
Thanks for the heads up, I'll look into it. 

> @jul-sh is it worth considering running this also pre-merge, without actually writing the provenance anywhere (dry run), so that at least it catches such errors?

Might be yeah, assuming that using up the GitHub CI runner resources is fine. One possible way to mitigate that would be to replace the existing reproducibility index script with the provenance logic, since it already runs on each PR. :]

tiziano88:
I think this is probably obsolete, right?

rbehjati:
> I think this is probably obsolete, right?

I think so. Closing it.

PR #2548, opened 2022-02-16
Remove build-server from the provenance generation step

rbehjati:
* Removes the `build-server` step from the provenance generation
* Replaces `build-server` with `build-functions-server` in the docs

Ref #2543

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -46,14 +46,7 @@ jobs:
           ./scripts/docker_pull
           df --human-readable
 
-      - name: Build Server Provenance
-        run: |
-          ./scripts/build_provenance \
-          "oak_loader" \
-          ./oak_loader/bin/oak_loader \
-          ./scripts/runner build-server
-
-      - name: Build Functions Provenance
+      - name: Build Functions Server Provenance
         run: |
           ./scripts/build_provenance \
           "oak_functions" \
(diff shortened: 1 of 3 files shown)


tiziano88 (approved):
Thanks @rbehjati .

Also I hope we can fix https://github.com/project-oak/oak/issues/2543 at some point so we notice such failures before submit.

jul-sh (approved):
thanks!

rbehjati:
Oh! I had not seen #2543. Updated the workflow to run the build step on the PRs too. Could you take another look? :)

tiziano88 (approved):
LG for now, looking forward to @jul-sh approach replacing reproducibility index with this at some point

.github/workflows/provenance.yaml
```diff
@@ -46,24 +49,19 @@ jobs:
...
           "oak_functions" \
           ./oak_functions/loader/bin/oak_functions_loader \
           ./scripts/runner build-functions-server
 
       - name: Move new provenance files to the provenance branch
+        if: github.event_name == 'push'
         run: cp -r ./provenance/. ./provenance-branch/
 
       - name: Commit new provenance files to the provenance branch
+        if: github.event_name == 'push'
```
jul-sh, line 64: nice! Thanks a bunch 💚 for adding that right away

jul-sh (approved):
lgtm

PR #2550, opened 2022-02-16
Split Oak Functions binaries

### LINKED ISSUE #2549: Split Oak Functions binaries according to config features
At the moment we compile the same top-level binary with different set of config flags, to obtain, for instance, the `base` and `unsafe` version, but bot of them end up having the same binary name and overwriting each other in the target dir.

I think we should consider having two distinct top level binaries with different features statically enabled / disabled, and with different resulting names.

tiziano88:
The main change is to turn oak_functions/loader into a library instead
of a combined library + binary. This means that most of the code in
main.rs is now in lib.rs, but the actual main is in the separate bin
crates now.

Fixes #2549


oak_functions/loader/src/lib.rs
```diff
@@ -25,3 +25,314 @@ pub mod metrics;
...
+                Level::Error,
+                &format!("error refreshing lookup data: {}", err),
+            );
+        }
+    }
+}
+
+/// This crate is just a library so this `main` function does not get executed directly by anything,
+/// it needs to be wrapped in the "actual" main from a bin crate.
+pub fn main() -> anyhow::Result<()> {
```
conradgrobler, line 157: nit: to avoid confusion (e.g. by someone ignoring the comments above) perhaps it would be useful to rename to something like `lib_main` or `runtime_main` or something similar.

conradgrobler (approved):
Nice

oak_functions/loader/src/lib.rs
tiziano88: Good idea, done.

PR #2552, opened 2022-02-17
Update Docker deps

tiziano88:
Also switch to BuildKit


tiziano88:
I was also trying to get rid of openssl from our docker image, but it seems it is still used by the experimental attestation crates. Worth thinking whether we want to keep them around or delete them, in the future.

ipetr0v:
We can just have a separate experimental Docket (that inherits the main one).

Dockerfile
```diff
@@ -132,7 +132,7 @@ RUN go get github.com/campoy/embedmd@97c13d6 \
 
 # Install liche (Markdown link checker) (via Go).
 # https://github.com/raviqqe/liche
-RUN go get github.com/raviqqe/liche@3ac05a3 \
+RUN go get github.com/raviqqe/liche@f9ba5f2 \
```
ipetr0v, line 135: Could you please add a Date or Version comment, so it's easier to understand what to update later?

ipetr0v (approved):

Dockerfile
tiziano88: Unfortunately there's no version, so the update here is just to get the latest commit at head. Even a comment with the date does not really help I think, does it?

Dockerfile
```diff
@@ -202,7 +202,7 @@ RUN curl --location https://sh.rustup.rs > /tmp/rustup \
 # We currently need the nightly version in order to be able to compile some of the examples.
 # See https://rust-lang.github.io/rustup-components-history/ for how to pick a version that supports
 # the appropriate set of components.
-ARG rust_version=nightly-2022-01-09
+ARG rust_version=nightly-2022-02-17
```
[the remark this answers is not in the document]
tiziano88: Yes, it seems to work fine locally.

.devcontainer.json
```diff
@@ -4,7 +4,7 @@
 // - https://code.visualstudio.com/docs/remote/devcontainerjson-reference
```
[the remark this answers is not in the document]
tiziano88: No, but this is a sort of special JSON that only VSCode parses I guess :)

tiziano88:
> We can just have a separate experimental Docker image (that inherits the main one).

Not sure it's worth the effort, also which Docker image would you use in VScode? It's already quite clunky with a single docker image, I would not want to image how it would be with two.

Maybe we should resurrect the idea of using nix?

Dockerfile
ipetr0v: I often put a date just to understand how old the version is.

ipetr0v:
It could be. But we'll need to do experimental work in the future anyway (though maybe each experimental would require a separate Docker).

> Maybe we should resurrect the idea of using nix?

Not sure if it will work with vscode

tiziano88:
> It could be. But we'll need to do experimental work in the future anyway (though maybe each experimental would require a separate Docker).

Not sure why each experimental folder would need its own Docker though. The only issue at the moment is that the experiments use openssl instead of rustls, otherwise it's perfectly fine for them to use the same Docker.

PR #2559, opened 2022-02-22
Remove tonic dep from remote_attestation crate by natively including protos

jul-sh:
This makes sense conceptually too, as the remote_attestation crate is not tied to a specific protocol, whereas tonic is. 

As an aside, I like how `remote_attestation` is already a seperate crate from `grpc_attestation`, makes the unary request migration a lot nicer. :)


tiziano88 (approved):
Nice, thanks

conradgrobler (approved):

PR #2560, opened 2022-02-22
Move release logic for vendoring deps to version history

jul-sh:
Do we still need this to run on each PR? It seems like this may have been superseded by transparent release.


jul-sh:
Vendoring deps could make sense still, but if we want that we should probs include that in the transparent release process.

tiziano88 (approved):

scripts/release_build_oak_functions
```diff
@@ -1,14 +0,0 @@
-#!/usr/bin/env bash
```
rbehjati, line 1: This is used in our internal cloudbuild script for release. Is there a plan for updating that as well?

jul-sh: Created cl/430414831 as a minimal update to internal references. There's more details in the CL description, we can discuss it there. :)

rbehjati: Thanks :)

rbehjati (approved):

PR #2567, opened 2022-02-28
Add hosted `channel_wait`

mariaschett:


mariaschett:
Closed because we will change to blocking calls in the ABI.

PR #2570, opened 2022-03-01
Make remote attestation diagram readable on dark mode

jul-sh:


jul-sh:
Screen Shot 2022-03-01 at 6 48 52 pm
atm one cannot really discern the arrows on darkmode

tiziano88 (approved):
LG, how can we do this for other images though, e.g. when we modify and re-export them? Or maybe can we set a background colour in the img element?

jul-sh:
@tiziano88 I don't think we can set it in the image tag in markdown, this PR literally just edits the png. Modifying the image would reserve it. But perhaps we used some tool?

Either way, on an approach level we Imo we should use GitHub's built in diagram support instead of images. This makes it clear how to edit diagrams, makes them readable as source code, and crucially can be read by people that use screen readers. https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

tiziano88:
Yes we (i.e. @ipetr0v ) used an internal tool to generate it, see:

https://github.com/project-oak/oak/blob/ccad8048c4cd6ebfc6bea04a6aa46dacc5abd16a/docs/remote-attestation.md?plain=1#L148-L150

jul-sh:
@tiziano88 thx for the context, I still wouldn't know how to edit the diagram in that internal tool, the link just takes me to a viewer. 

Imo using mermaid (what powers GitHub rendering) would be the best fit. Moreso than fixing rendering, storing the source of this diagram as part of the OSS repo seems like a good improvement.

tiziano88:
Nice, mermaid would indeed be ideal! @ipetr0v could you share the source of the seq diagram, I also can only see the PNG at the moment in the internal tool, not its source.

jul-sh:
@tiziano88 found a way to extract it, it's possible, internal tools just don't make it obvious how 😅

```
participant Client [fillcolor="lightgreen", fontcolor="black"]
participant Server [fillcolor="green", fontcolor="white"]
participant TEE Platform [fillcolor="blue", fontcolor="white"]
participant TEE Provider [fillcolor="grey", fontcolor="white"]

TEE Provider-->Client: Root public key (PKI)
TEE Platform->TEE Provider: Firmware key
TEE Provider->TEE Platform: Firmware key signature
Client-->Client: Generate a Signing key pair
Server-->Server: Generate a Signing key pair

Client-->Client: Generate an Ephemeral key pair
Client->Server: ClientHello
Server-->Server: Generate an Ephemeral key pair

Server-->Server: Put Signing public key hash\nto the AttestationReport request
Server->TEE Platform: AttestationReport request
TEE Platform->Server: Signed AttestationReport

Server->Client: ServerIdentity
Client-->Client: Validate AttestationReport

Client->Server: ClientIdentity

Client-->Client: Derive Session Keys
Server-->Server: Derive Session Keys
Note over Client,Server: Secure channel is created [fillcolor="lightgray"]

Client->Server: Encrypted request
Server->Client: Encrypted response
```

tiziano88:
Ah, I just found out how to see the source by changing the URL from view to edit: https://sequencediagram.googleplex.com/edit/6235412701904896

jul-sh:
It uses the syntax from https://github.com/bramp/js-sequence-diagrams, which is a bit different from Mermaid, but shouldn't be too hard to rewrite it.

PR #2571, opened 2022-03-02
Document why EKEP messages use custom serialization logic

jul-sh:
This wasn't obvious from reading the runtime code, which apart from this piece makes heavy use of protos to define RPC messages.

### DIFF
--- a/remote_attestation/rust/src/message.rs
+++ b/remote_attestation/rust/src/message.rs
@@ -17,8 +17,8 @@
 // Messages used to implement the Enclave Key Exchange Protocol (EKEP).
 //
 // Unlike other RPC messages these are expressed as Rust structs and serialized
-// with custom logic. The primary reason for is maintaining binary-compatibility
-// with other implementations of the Enclave Key Exchange Protocol (EKEP).
+// with custom logic. This is done to maintain binary-compatibility with other
+// implementations of the Enclave Key Exchange Protocol (EKEP).
 
 use crate::crypto::{
     KEY_AGREEMENT_ALGORITHM_KEY_LENGTH, NONCE_LENGTH, SIGNATURE_LENGTH,
--- a/remote_attestation/rust/src/message.rs
+++ b/remote_attestation/rust/src/message.rs
@@ -14,6 +14,12 @@
 // limitations under the License.
 //
 
+// Messages used to implement the Enclave Key Exchange Protocol (EKEP).
+//
+// Unlike other RPC messages these are expressed as Rust structs and serialized
+// with custom logic. The primary reason for is maintaining binary-compatibility
+// with other implementations of the Enclave Key Exchange Protocol (EKEP).
+
 use crate::crypto::{
     KEY_AGREEMENT_ALGORITHM_KEY_LENGTH, NONCE_LENGTH, SIGNATURE_LENGTH,
     SIGNING_ALGORITHM_KEY_LENGTH,
--- a/remote_attestation/rust/src/message.rs
+++ b/remote_attestation/rust/src/message.rs
@@ -14,11 +14,12 @@
 // limitations under the License.
 //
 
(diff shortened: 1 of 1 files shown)


remote_attestation/rust/src/message.rs
```diff
@@ -14,6 +14,12 @@
 // limitations under the License.
 //
 
+// Messages used to implement the Enclave Key Exchange Protocol (EKEP).
```
ipetr0v, line 17: I think we should refer to it as the Remote Attestation Protocol
And the note that it's a EKEP inspired is in the docs

jul-sh: I think it's helpful to delineate between the full Remote Attestation Protocol, and the key exchange. The full protocol includes attestation report, which is defined as a proto (and presumably not intended to be binary compatible).

conradgrobler: This protocol is not EKEP, but a simplified version, so using "EKEP" is not accurate. "Remote attestation handshake protocol" might be more accurate. Also, perhaps add a link to the readme where the protocol is described.

remote_attestation/rust/src/message.rs
```diff
@@ -14,6 +14,12 @@
...
+//
+// Unlike other RPC messages these are expressed as Rust structs and serialized
+// with custom logic. This is done to maintain binary-compatibility with other
+// implementations of the Enclave Key Exchange Protocol (EKEP).
```
conradgrobler, line 21: Also here: not quite EKEP.

remote_attestation/rust/src/message.rs
```diff
@@ -14,6 +14,12 @@
 // limitations under the License.
 //
 
+// Messages used to implement the Enclave Key Exchange Protocol (EKEP).
+//
+// Unlike other RPC messages these are expressed as Rust structs and serialized
```
conradgrobler, line 19: nit: these are not RPC messages, so perhaps rather "Unlike RPC message, ...."

remote_attestation/rust/src/message.rs
conradgrobler: Another point I forgot to mention: this protocol includes the remote attestation report, but does not specify the format (it is just an opaque binary blob from the protocol perspective). The format for the attestation report is platform- and implementation-dependent. We currently use a proto as a placeholder, but once once we have a real implementation using the appropriate real hardware it will likely be a different format.

jul-sh: thx clarified it's simplified EKEP, not EKEP.

remote_attestation/rust/src/message.rs
jul-sh: Eg what makes the Attestation Report an RPC message that these messages do not have

conradgrobler: Just to clarify further: for me the property that makes something RPC is that it is the input parameter to a remote function or the result from the function execution. While you could argue the handshaker steps are also remote functions being executed, I see them as part of the initial setup, not part of the remote calculation that the client is interested in.

conradgrobler (approved):

remote_attestation/rust/src/message.rs
tiziano88: I don't know if it's the actual definition of RPC, but in my mind RPC suggests a mechanism for invoking arbitrary procedures / functions remotely. e.g. in gRPC , the RPC part comes from the fact that a message has the name of the method to invoke, and the payload of the request. in capnproto instead of the name there is a number corresponding to the method, but conecptually similar. Here, it is just a fixed sequence of frames back and forth, there are no "procedures" being invoked.

jul-sh: Fair, thanks for the context. That does mean we probably don't want to delineate format use based on what technically is an RPC message. Replaced RPC messages with "messages exchanged between client & server" here 😅

PR #2574, opened 2022-03-02
Stop cloning additional_info into each ServerHandshaker

jul-sh:
We presently clone `additional_info` for each client that remotely attests the server. This doesn't seem quite ideal, as this value should be consistent everywhere. Cloning it might lead to unexpected state propagation.


remote_attestation/rust/src/handshaker.rs
```diff
@@ -329,40 +329,45 @@ pub struct ServerHandshaker {
...
     /// If [`None`] is returned, then no messages should be sent out to the client.
-    pub fn next_step(&mut self, message: &[u8]) -> anyhow::Result<Option<Vec<u8>>> {
-        self.next_step_util(message).map_err(|error| {
-            self.state = ServerHandshakerState::Aborted;
-            error
-        })
+    pub fn next_step(
+        &mut self,
+        message: &[u8],
+        additional_info: Vec<u8>,
```
tiziano88, line 350: I may be missing something, but what determines what to pass here, and at exactly which step? AFAICT the additional info is only used at one particular step, so we still need to pass something else for all the other steps? Looking at the tests below it seems that we do keep passing the same value, but that is only actually used in one.

Also the idea of this method is that we just pipe in messages from the underlying transport, and it should move along the various steps in the state machine (and eventually produce the encryptor), so adding this argument here seems to complicate the API.

I'll let @conradgrobler and @ipetr0v weigh in too.

Another thing to consider is that, in practice, in most protocols that have some space for additional info, only a fixed amount of additional data can be passed in, usually 256 bits, usually the result of a hashing function (rather than the actual data). So maybe another approach worth considering in order to reduce the memory usage would be to make this a fixed size array, but keep it in the constructor. Would that work?

remote_attestation/rust/src/handshaker.rs
```diff
@@ -448,7 +454,6 @@ impl ServerHandshaker {
                 .as_ref()
                 .context("Couldn't get TEE certificate")?;
 
-            let additional_info = self.additional_info.clone();
             let attestation_info =
                 create_attestation_info(signer, additional_info.as_ref(), tee_certificate)
```
tiziano88, line 458: As a side note, would it work to just take a reference here instead of cloning it? It seems we are cloning and getting a ref immediately anyways, I don't remember if there is a reason for that or it's just an accident.

cc @ipetr0v

remote_attestation/rust/src/handshaker.rs
jul-sh: Yes, only used some for some steps, passing it always is still not optimal.

> what determines what to pass here
afaik the config stays unchanged after the runtime is initiated. Hence it'd always be a clone of the same vec. 

Which leads me to the suggestion I raised in the chat, and which be the optimal approach for memory: We create a single instance of the additional_info vec in the loader, and only pass down references (with a static lifetime).

tiziano88: I think it's safe to assume it does not change, but I'm not sure that implies it can be `'static` -- maybe it can be coerced to a static value at runtime by "leaking" it? but I'm not sure that's idiomatic.

tiziano88: Also see my other comment below, maybe the fix you are looking for is just a one line change there?

remote_attestation/rust/src/handshaker.rs
jul-sh: Yes we ref the clone, but the actual clone is passed to `ServerIdentity` a few lines down.

jul-sh: To be fair we can clone it there, (which means we no longer need to clone for every step only this one). I added a commit for that. It's an improvement but we still need the other changes (stop keeping a clone in the Handshaker struct)

remote_attestation/rust/src/handshaker.rs
conradgrobler: How about using an Arc<Vec<u8>> rather than a static reference? Cloning it should be cheap, and it would remove the need to pass the additional info into every step.

jul-sh: Not entirely convinced we need.want reference counting if this will never be dropped. But either way, wouldn't Rc be better since it's a single thread? https://doc.rust-lang.org/std/rc/struct.Rc.html

tiziano88: I agree with @jul-sh that reference counting (atomic or not) seems superfluous here.

I think there is probably a solution that involves adding explicit lifetimes to the various structs involved, but I'm not sure it's worth optimizing for that yet.

Also just to correct myself on another point: it looks like this extra data is indeed only used as input to a hash function (so it ends up being fixed size in the wire format):

https://github.com/project-oak/oak/blob/1cf966842d26d0c63df9deffc37074852b88576f/remote_attestation/rust/src/handshaker.rs#L713

Anyways, since we know that this gets used only once, maybe using `mem::take` or `Option::take` may be another approach?

conradgrobler: If the Rust compiler can infer that the AttestationServer instance will always outlive the handshaker (which hopefully it should be able to do in spite of the async stream), the additional info passed to the handshaker could just be a reference e.g. `&[u8]`.

The suggestion for using reference counting was based on the possibility that the compiler might not be able to accurately infer the lifetimes due to the async stream.

Regarding the wireformat: the additional info is also included separately in the server identity message (https://github.com/project-oak/oak/blob/main/remote_attestation/rust/src/handshaker.rs#L456-L464) and serialised (https://github.com/project-oak/oak/blob/main/remote_attestation/rust/src/message.rs#L269)

jul-sh: The compiler isn't able to infer that, storing a ref in the handshaker struct requires a lifetime annotation. It is able to infer lifetimes when passing the ref to methods, which this PR implements. 

If we really want the additional_info to be maintained in the handshaker struct (as opposed being passed to methods) it's either ref counting, or the explicit lifetime approach I suggested. 

My opinion is that passing to methods is fine. If we feel strongly otherwise imo we should use the explicit lifetime approach.  

We definitely do not want the status quo of cloning.

tiziano88: Passing to the `next_step` method seems out of question for me, it breaks encapsulation of a method that is supposed to deal with a stream of opaque messages from the outside world. In fact eventually it may even implement `Sink<Vec<u8>>` or similar (at least conceptually).

What would this look like with the explicit lifetime annotation @jul-sh ?

jul-sh: I don't really see it breaking encapsulation. It's just a question of whether this additional data (quite literally the variable name 😅) is passed on each call or initially and cloned as state. Happy to talk about it more though to understand your PoV. 

Anyways, with explicit lifetime annotations we would continue passing the additional info when the `Handshaker` is instantiated by `AttestationServer`. We use them to declare that `AttestationServer` always outlives `Handshaker`. That is the case (and always will be) but because of the async stream the compiler is unable to infer that implicitly.

ipetr0v: > Also the idea of this method is that we just pipe in messages from the underlying transport

I agree, I think having more arguments in the `next_step` function may be inconvenient.

Also, could we just have a shared reference to the `additional_info` instead? If the only problem is it being copied multiple times, shared reference could be fine, I think

jul-sh: If passing it to the method is not idiomatic, we can make it a shared reference (std::sync::Arc) indeed. That would probably be easier than declaring the lifetimes. 

The only caveat is that as of #2556 the `oak_remote_attestation` crate is no_std compatible. This means we cannot use `std::sync::Arc` without breaking that. 

@conradgrobler Do we plan to support shared references in no_std in the future?

conradgrobler: You can use `alloc::sync::Arc`. We will always have an allocator, even in no_std environements.

jul-sh: Thanks for the pointer, will use that. Also realized that for no_std we'll probs be moving this gRPC communication logic out of the trusted runtime anyways.

tiziano88: Will we? I thought all this stuff will remain in the trusted binary, since it drives the remote attestation protocol.

Anyways, I still don't think we should change this method to accept the additional info param. Especially in the non-std, blocking model, this method should look like a `call` or `invoke` function from bytes to bytes (with some extra wrappers), so that we can compose it more nicely with the rest of the building blocks of the runtime. And at some point it should itself also accept another callback that it invokes with the decrypted data from remote attestation, which itself would follow the same paradigm.

grpc_streaming_attestation/src/server.rs
```diff
@@ -114,7 +114,7 @@ where
         Ok(Self {
             tee_certificate,
             request_handler,
-            additional_info,
+            additional_info: Arc::new(additional_info),
```
jul-sh, line 117: Do we prefer a let declaration over instantiating the Arc in the struct instantiation?

jul-sh, line 117: Do we prefer a let declaration over instantiating the Arc in the struct instantiation?

jul-sh:
@tiziano88 @conradgrobler would you mind taking another look? Updated this to use reference counting.

jul-sh:
@tiziano88 @conradgrobler would you mind taking another look? Updated this to use reference counting.

grpc_streaming_attestation/src/server.rs
tiziano88: This looks fine to me.

oak_functions/client/rust/src/attestation.rs
```diff
@@ -30,7 +30,7 @@ pub fn into_server_identity_verifier(
     config_verifier: ConfigurationVerifier,
 ) -> ServerIdentityVerifier {
     let server_verifier = move |server_identity: ServerIdentity| -> anyhow::Result<()> {
-        let config = ConfigurationInfo::decode(server_identity.additional_info.as_ref())?;
+        let config = ConfigurationInfo::decode(&*server_identity.additional_info.as_slice())?;
```
tiziano88, line 33: I think there should be some way of doing this without `&*`, maybe a combination of `as_bytes` , `as_slice`, `as_ref`, but I'm never sure in which order.

tiziano88 (approved):
LG, fewer clones means less risk that state may get out of sync, but I still think that we may refactor this entire code to be non-async because of no_std requirement, and in that case we should be able to use plain `&` refs with explicit lifetimes. Anyways we can look into it later.

@jul-sh just to confirm, is that (i.e. non-async) the plan for unary attestation? Once we separate it from the gRPC logic of course (which can stay async). Actually I think I'm still a bit unclear exactly which crates will go into the trusted vs untrusted binary eventually, could you remind me?

grpc_unary_attestation/src/server.rs
```diff
@@ -68,7 +71,7 @@ impl SessionTracker {
         match self.known_sessions.pop(&session_id) {
             None => match AttestationBehavior::create_self_attestation(&self.tee_certificate) {
                 Ok(behavior) => Ok(SessionState::HandshakeInProgress(Box::new(
-                    ServerHandshaker::new(behavior, self.additional_info.clone()),
+                    ServerHandshaker::new(behavior, Arc::new(self.additional_info.clone())),
```
conradgrobler, line 74: Is it not possible to move the Arc even higher up so that each session tracker also stores a reference to the same fixed additional info to avoid cloning the vector here?

conradgrobler (approved):

jul-sh:
> LG, fewer clones means less risk that state may get out of sync, but I still think that we may refactor this entire code to be non-async because of no_std requirement, and in that case we should be able to use plain & refs with explicit lifetimes. Anyways we can look into it later.

Yup, that'd pretty much what I tried to express in https://github.com/project-oak/oak/pull/2574#discussion_r830992522, sorry if it wasn't clear. Does mean that these code changes will be obsolete soon, but I think it was still a valueable excerise.

> @jul-sh just to confirm, is that (i.e. non-async) the plan for unary attestation? Once we separate it from the gRPC logic of course (which can stay async).

Yup! We'd have to, since the current plan for no_std is to not support threads.

> Actually I think I'm still a bit unclear exactly which crates will go into the trusted vs untrusted binary eventually, could you remind me?

Depends on how we make no_std work, but on a high level as follows:

trusted binary: current runtime minus `grpc_unary_attestation`
untrusted binary: basically just `grpc_unary_attestation`

grpc_unary_attestation/src/server.rs
jul-sh: Possible to move it up higher, but there will only ever be on `SessionTracker`. That's why it was called `SessionsTracker` (plural) in my initial PR. 

What do you think would be a good way to make that clearer? Could add a comment above it

oak_functions/client/rust/src/attestation.rs
jul-sh: yeah, I wasn't sure what the rusty convention is here. The syntax I first used was even more unwieldy: `&**server_identity.additional_info` 😅

jul-sh: in steps it's basically:
deref the Arc,
get additional info of the content
get that as a slice
pass the result as a ref

jul-sh: Changed to `server_identity.additional_info.as_ref().as_slice()` which seems nicer

grpc_unary_attestation/src/server.rs
jul-sh: true, updated it, thx

remote_attestation/rust/src/tests/message.rs
```diff
@@ -131,7 +131,9 @@ fn test_serialize_server_identity() {
         assert!(result.is_ok());
         TestResult::from_bool(result.unwrap())
     }
-    quickcheck(property as fn(Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>) -> TestResult);
+    quickcheck(
+        property as fn(Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>, Vec<u8>, Arc<Vec<u8>>) -> TestResult,
```
ipetr0v, line 135: What does this test do?

PR #2575, opened 2022-03-02
Add a lookup data checker

ipetr0v:
This PR adds an utility that checks the correctness of the weather lookup data file.

### DIFF
--- a/oak_functions/lookup_data_checker/README.md
+++ b/oak_functions/lookup_data_checker/README.md
@@ -1,3 +1,9 @@
 # Lookup Data Checker
 
 The binary in this directory checks the correctness of the lookup data file.
+
+Usage:
+
+```bash
+RUST_LOG=debug cargo run --manifest-path=oak_functions/lookup_data_checker/Cargo.toml -- --file-path=weather_data_file
+```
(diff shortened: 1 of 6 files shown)


oak_functions/lookup_data_checker/README.md
```diff
@@ -0,0 +1,3 @@
+# Lookup Data Checker
+
+The binary in this directory checks the correctness of the lookup data file.
```
conradgrobler, line 3: perhaps add a code example on how to use it?

conradgrobler (approved):
Nice, thanks!

oak_functions/lookup_data_checker/README.md
ipetr0v: Done

tiziano88 (approved):
Nice! ideally this could be also a utility library that runs on the output from the generator, and there could be an integration test between the two. Perhaps create an issue to track that @ipetr0v ? (does not have to be done now, we can just leave it around)

ipetr0v:
Created https://github.com/project-oak/oak/issues/2578

PR #2577, opened 2022-03-03
Make TF inference a separate crate

conradgrobler:
This is the first of a series of changes to move components out of the Oak Functions Loader library into separate crates, each with the minimal possible dependencies. Where possible these crates will be no_std compatible, but it is not possible for the TF Inference as tract-tensorflow is not no_std compatible.


oak_functions/logger/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+//
+
+use log::Level;
+
+pub trait OakLogger {
+    /// Logs the message, which might contain sensitive information, at the specified `Level`.
+    ///
+    /// Only insecure debug-only implementations may provide a non-empty implementation. Production
+    /// implementations must not do anything.
+    #[allow(unused_variables)]
```
tiziano88, line 24: I would avoid the default methods unless necessary, IMO it makes this look too much OOP, while it's not really. I think it should be fine to just always implement them explicitly?

conradgrobler: Removed. I added it only to make clear that it should be an empty implementation by default.

rbehjati (approved):
Nice. Thanks.

Issue #2580, opened 2022-03-04
Convert the `oak_functions_metrics` random number generation to be no_std commpatible


conradgrobler:
At the moment we are using `StdRng` and `thread_rng`, which is not compatible with `no_std`.

conradgrobler:
This code has been deleted.

Issue #2579, opened 2022-03-04
Implement a no_std compatible Mutex and RwLock


conradgrobler:
We could use the `lock_api` crate. For cases where we have an OS, we can use `std::sync::Mutex` or `parking_lot::Mutex` to provide the implementation. For UEFI we can start with spinlocks and later do something more sophisticated.

PR #2582, opened 2022-03-04
Make Metrics a separate crate

conradgrobler:
The `oak_functions_metrics` crate is not yet `no_std` compatible, but will be in future once #2579 and #2580 have been fixed.

### DIFF
--- a/oak_functions/examples/fuzzable/module/src/tests.rs
+++ b/oak_functions/examples/fuzzable/module/src/tests.rs
@@ -16,7 +16,7 @@
 use crate::proto::{
     instruction::InstructionVariant, Instruction, Instructions, Panic, WriteResponse,
 };
-use maplit::hashmap;
+use maplit::btreemap;
 use oak_functions_abi::proto::{ServerPolicy, StatusCode};
 
 use oak_functions_loader::{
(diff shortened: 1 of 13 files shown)


oak_functions/examples/fuzzable/module/src/tests.rs
```diff
@@ -16,18 +16,20 @@
 use crate::proto::{
     instruction::InstructionVariant, Instruction, Instructions, Panic, WriteResponse,
 };
-use maplit::hashmap;
+use maplit::btreemap;
```
rbehjati, line 19: Is this related to making the crate no-std? What is the performance implication? Or does it not matter since we expect these maps to be small?

rbehjati (approved):
LGTM. Thanks :)

oak_functions/examples/fuzzable/module/src/tests.rs
conradgrobler: Yes, there is no hashmap in the alloc crate as the default implementation relies on a std source of randomness. We could switch to a different impelementation (such as hashbrown), but these would only contain the various metrics buckets so I don't expect it to contain more than 5 or 10 items. I think a btreemap is probably more efficient for that number of entries.

Issue #2569, opened 2022-03-01
Document feature flagging approach and refactor existing feature flags to match


jul-sh:
I think our approach to feature flags could be improved a bit. We currently use both runtime and compile time feature flags, and sometimes mix them. (eg tensorflow is a runtime flag but also requires unsafe mode, a config of safe + tensorflow is possible but wouldn't work).

Imo we should default to using compile time configuration as a default. This means only using runtime flags when being able to set a flag at runtime is a functional requirement. (Eg tensorflow would just be a compile time flag). This keeps the resulting binary small, which is not just a general best practice, but especially important for security critical software like Oak and for simplifying the implementation of projects like no_std support. 

Thoughts? cc @tiziano88 @conradgrobler

jul-sh:
The one caveat worth pointing out is that this means we'll end up with more binaries, and we'll need to compile more binaries to run tests. Imo an acceptable tradeoff.

jul-sh:
Thanks for the link to #2551, didn't see it before. That seems very similar indeed. I think if we go with the approach suggested in it (imo we should), it would still make sense to think about how that relates to feature flags. 

> I am in favour of this approach, or even taking it further as suggested in https://github.com/project-oak/oak/issues/2551. Runtime configuration should be used to configure the functionality that is compiled into the binary (e.g. the Wasm business logic to use), not switch functionality on and off.

Or to say it more clearly, we probably do not need any runtime configuration. All configuration could & should be parsed at compile time. 

> I also think that we should prefer setting configuration in a config file rather than using multiple command-line arguments. In some environments command-line arguments will not be feasible (e.g. bare metal/UEFI).

+ 1. In that case, we should decide if we want a) a list of runtimes an app can choose [as the suggestion appears to be in #2551], or b)each app defines which components should be contained in the runtime it uses. To use a culinary analogy, whether apps choose an oak menu, or assembly the runtime a-la-carte. 

Imo ideally it should be b). It allows for combinations that we'll need eventually (eg TF _and_ metrics) and also more easily allows us to move other items like `worker_threads`, `policy`, etc to compile time.

tiziano88:
I agree with the overall sentiment (as https://github.com/project-oak/oak/issues/2551 suggests and https://github.com/project-oak/oak/pull/2550 already starts implementing), but

> Or to say it more clearly, we probably do not need any runtime configuration. All configuration could & should be parsed at compile time.

[...]

Issue #2551, opened 2022-02-17
Allow assembling a Runtime binary with a selection of statically defined extensions


tiziano88:
Currently we have two flavours of the Oak Functions runtime binary (base and unsafe). They both depend on the Oak Functions library, and indirectly decide what extensions are included in that by means of specifying different features:

https://github.com/project-oak/oak/blob/7bdc4440f43cf839fa6a615aeffec781701176a5/oak_functions/oak_functions_loader_base/Cargo.toml#L9

https://github.com/project-oak/oak/blob/7bdc4440f43cf839fa6a615aeffec781701176a5/oak_functions/oak_functions_loader_unsafe/Cargo.toml#L9

And the main entry point is identical for both:

https://github.com/project-oak/oak/blob/7bdc4440f43cf839fa6a615aeffec781701176a5/oak_functions/oak_functions_loader_unsafe/src/main.rs#L21-L23

Instead, ideally we would have a static config in the runtime entry point:

```rust
pub fn main() -> anyhow::Result<()> { 
  oak_functions_loader::lib_main(hashmap!{
    LOOKUP_DATA_HANDLE => LookupDataExtension::new(),
    TF_DATA_HANDLE => TFExtension::new(),
  })
} 
```

This definitely requires more thinking, we can use this issue to discuss.

tiziano88:
Also, for now I don't think we need to treat Wasm as an extension, otherwise we need to start worrying about extensions of extensions. And I think in practice we are treating Wasm as our "userspace" where processes runs, using the OS terminology. Figuring out how to remove that entirely from a trusted runtime can be left for even further future work.

@mariaschett @conradgrobler WDYT?

conradgrobler:
I think this is a good way forward for now, and agree that we can worry about the possibility of making Wasm an extension later.

For future reference when we start looking at it, here are some of my thoughts on a possible approach:

Every piece of self-contained functionality becomes a service (in a separate crate)
- IO (e.g. the current gRPC server listening for user requests)
- Wasm host
  - Creates a Wasm sandbox instance for every request
  - Creates copies of the same channels it has for the sandbox
  - Proxies comms from the sanbox instance to the corresponding service (e.g. lookup, IO)
- Lookup data provider
- Lookup data refresher

Each service would implemment a trait that would support adding a channel endpoint. We can either automatically assign channel handles based on the order in which the endpoint is added, or allow specifying a handle when adding the channel (for the following examples I will assume we automatically assign channels handles).

[...]

PR #2584, opened 2022-03-04
Add runtime flag to specify gRPC request model

jul-sh:
Implements the feature flag as a runtime flag as that's what we do atm for lookupdata, tensorflow etc. Worth mentioning alternatively we could make this a build time flag (ref our previous discussion in https://github.com/project-oak/oak/issues/2569). Doing so would introduce additional complexity however since we currently only have two binary variants. 

Opening this so we can decide whether we want a build time or runtime flag for this. Imo build time would be preferable in principle, but runtime is the pragmatic choice right now: It's our current model and changing it for this particular flag probs isn't worth it. This flag is transient and has limited security implications.

### DIFF
--- a/oak_functions/examples/fuzzable/module/src/tests.rs
+++ b/oak_functions/examples/fuzzable/module/src/tests.rs
@@ -20,7 +20,7 @@ use maplit::btreemap;
 use oak_functions_abi::proto::{ServerPolicy, StatusCode};
 
 use oak_functions_loader::{
-    grpc::{create_and_start_grpc_server, create_wasm_handler},
+    grpc::{create_and_start_grpc_server, create_wasm_handler, RequestModel},
     logger::Logger,
     lookup::LookupFactory,
     lookup_data::LookupData,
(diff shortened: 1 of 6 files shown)


tiziano88 (approved):
LG as a pragmatic approach, especially since it's temporary.
Long term, even more than a compile time flag, it could even be a binary assembled with a different set of building blocks (see https://github.com/project-oak/oak/issues/2551), and e.g. the streaming and unary parts would be different crates implementing the same interface (but haven't thought enough about it, not sure it'd actually work)

jul-sh:
@tiziano88 
> Long term, even more than a compile time flag, it could even be a binary assembled with a different set of building blocks (see https://github.com/project-oak/oak/issues/2551), and e.g. the streaming and unary parts would be different crates implementing the same interface (but haven't thought enough about it, not sure it'd actually work)

+1, long term that seems like it would be ideal setup for feature flags.

conradgrobler (approved):

rbehjati (approved):
What do you mean by runtime flag in the discussions here and in #2569? Do you mean the options that are included in the config, e.g., `TensorFlowModelConfig`?

jul-sh:
@rbehjati yes I mean those. :) Perhaps runtime config would be the more accurate term.

PR #2585, opened 2022-03-06
Add comment explaining config behevaior and rationale

jul-sh:
This wasn't clear to me when first reading the code. Without a comment it likely wouldn't be for other newcomers and/or external reviewers. [Ref the related discussions in #2569 and #2551].

### DIFF
--- a/oak_functions/loader/src/lib.rs
+++ b/oak_functions/loader/src/lib.rs
@@ -61,6 +61,16 @@ use std::{
 #[cfg(test)]
 mod tests;
 
+/// Runtime Configuration of Runtime.
+///
+/// This struct serves as a schema for a static TOML config file provided by
+/// application developers. In deployment, this static config file is typically
+/// bundled with the Oak Runtime binary. The config file is only ever read at
+/// startup. Config is not expected to ever change for a running Oak instance.
+/// Conceptually, these config values can be thought of as akin to compile-time
+/// constants. The only reason application config is not literally provided at
+/// compile-time is so that different applications can use the same
+/// trusted/endorsed runtime binary.
 #[derive(Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 struct Config {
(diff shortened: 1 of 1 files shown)


oak_functions/loader/src/lib.rs
```diff
@@ -61,6 +61,16 @@ use std::{
 #[cfg(test)]
 mod tests;
 
+/// Runtime Configuration of Runtime.
+///
+/// This struct serves as a schema for a static TOML config file provided by
+/// application developers. In deployment, this static config file is typically
+/// bundled with the Oak Runtime binary. The config file is only ever read at
+/// startup. Config is not expected to ever change for a running Oak instance.
+/// Conceptually, these config values can be thought of as akin to compile-time
```
tiziano88, line 70: The original intention was for this struct and the config file to be a type safe version of regular command line flags (as used by most other programs), so maybe clarifying that may be useful too (and maybe sufficient?). I.e. conceptually I would say that these are akin to command line flags, rather than constants. Then sure, command line flags are themselves like constants, but set at runtime and read only once etc., but that's perhaps more familiar to developers and readers already.

jul-sh: Thanks for the background, updated.

 I saw that we also have actual CLI flags alongside these, the distinction wasn't quite clear to me (eg lookupdata path goes into the config, WASM mode path remains an actual CLI flag). Probably worth thinking about moving all flags into the config file then. :)

conradgrobler (approved):

ipetr0v (approved):

oak_functions/loader/src/lib.rs
tiziano88: Yeah I suspect we will end up with something similar to how in the original Oak we had to "bundle" configuration and code together in an "application". Or we can keep at least the code separate, and let the untrusted launcher read config and code and provide them to the trusted runtime.

tiziano88 (approved):
Thanks!

PR #2586, opened 2022-03-06
Add missing flag to development docs

jul-sh:
Without the logs flag the expected output mentioned in the docs is never shown.


tiziano88 (approved):
Thanks for the fix

PR #2588, opened 2022-03-07
Add toml files for echo example

### LINKED ISSUE #2587: Echo example exits without waiting for requests
Expected behavior: Running `./scripts/docker_run ./scripts/xtask --scope=all --logs run-functions-examples --example-name=echo --client-variant=none` (that's the [example command from the docs](https://github.com/project-oak/oak/blob/main/docs/development.md#run-oak-functions-server) with example-name set to echo) starts the server, it listens to requests.

Actual behavior: Running that command exits without listening for requests.

conradgrobler:
Fixes #2587

### DIFF
--- a/oak_functions/examples/echo/config.toml
+++ b/oak_functions/examples/echo/config.toml
@@ -1,0 +1,5 @@
+worker_threads = 4
+
+[policy]
+constant_response_size_bytes = 90
+constant_processing_time = "20ms"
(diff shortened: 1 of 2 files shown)


mariaschett:
I added this example a while ago to have some minimal compiling bytecode [for tests](https://github.com/project-oak/oak/blob/cf884d0677091588ee4e1c5b7fbe1f1bd1ccd899/oak_functions/sdk/test_utils/src/lib.rs#L275)---but I never really run it! Thanks @conradgrobler for fixing. 

I can think of a different way to get the bytecode, if you prefer not to have the example around.

jul-sh (approved):
Thanks! :))

PR #2596, opened 2022-03-08
Implement Server Side Unary Remote Attestation

jul-sh:
Implements the server side component of unary remote atttestion


grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,89 @@
...
 use tonic;
 
-/// gRPC Attestation Service implementation.
-pub struct AttestationServer<F, L: LogError> {
+type SessionId = u64;
+enum SessionState {
+    // Boxed due to large size difference, ref: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
+    HandshakeInProgress(Box<ServerHandshaker>),
+    EncryptedMessageExchange(Encryptor),
+}
```
jul-sh, line 32: This arguably duplicates the `ServerHandshakerState` state key inside of the `ServerHandshaker`, which itself is an enum with a `ServerHandshaker::Completed(AeadEncryptor)` value which corresponds `EncryptedMessageExchange(Encryptor)`.

Without additional refactoring however, `ServerHandshaker` cannot be used directly. It's `get_encryptor` method consumes the wrapping `ServerHandshaker` instance. 

Effectively this comes up again at L77 in `SessionsTracker.pop_session_state`, which checks for handshake completion when retrieving session state, as a completed `ServerHandshaker` and `Encryptor` can functionally represent the same state. (L77 checks for this, always returns the underlying `Encryptor` to ensure consistency).

jul-sh: Imo the ideal architecture would be to merge `ServerHandshaker` & `Encryptor` into a single struct. While doing so it would also be good to remove the duplication of state currently present in `ServerHandshaker`: it keeps state both a state enum and a transcript, duplicate state can lead to unexpected behavior down the road, enabling security vulnerabilities.

jul-sh: I choose not do this refactoring for this PR, as it necessitates significant changes to both the abstract attestation logic, and the grpc streaming implementation. However it is worth considering doing this refactoring after we remove the grpc streaming implementation.

oak_functions/lookup_data_checker/Cargo.toml
```diff
@@ -14,5 +14,4 @@ location_utils = { path = "../location_utils" }
 log = "*"
 oak_functions_abi = { path = "../abi" }
 prost = "*"
-rand = "*"
 serde = { version = "*", features = ["derive"] }
```
jul-sh, line 17: A little bit confused why this PR needs to update this. It appears as that this unused dep is an issue that precedes these changes, but was flagged when running CI locally.

grpc_unary_attestation/src/server.rs
conradgrobler: I agree that this is not optimal, but I think it is OK for an initial implementation. There is probably a need for a joint refactoring of the streaming and unary versions that is too big for this PR. Perhaps add an issue and a TODO.

grpc_unary_attestation/src/server.rs
```diff
@@ -59,19 +130,75 @@ where
...
+        };
 
-        error_logger.log_error(&format!(
-            "Received unary request, but request handling isn't yet implemented: {:?}",
-            request
-        ));
-        return Err(tonic::Status::unimplemented(""));
+        // Note that we only get here if no errors occured during the preceding
+        // steps. If errors do occur the session state as tracked by the server
+        // is effectovely earsed. This allows the client to negotiate a new
```
conradgrobler, line 194: nit: "effectively erased"

oak_functions/lookup_data_checker/Cargo.toml
conradgrobler: cargo-udeps was disabled on the server CI for a while. I am not sure whether it has been re-enabled.

grpc_unary_attestation/src/server.rs
jul-sh: >  but I think it is OK for an initial implementation
agreed, hence why I implemented it this way. Will file an issue where we can aggregate refactoring needs

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+    /// intial state if the session is not known.
+    ///
+    /// Note that getting the remote attestaion state of a session always
+    /// implicitly removes it from the set of tracked sessions. After
+    /// using the state to process a request with this state it must explicitly
+    /// be put back into the SessionsTracker. This an intentional choice meant
+    /// meant to ensure that faulty state that leads to errors when preocessing
+    /// a request is not persistent.
+    pub fn pop_session_state(&mut self, session_id: SessionId) -> Result<SessionState, String> {
+        return match self.known_sessions.pop(&session_id) {
```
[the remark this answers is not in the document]
jul-sh: it's a tiny function that barely wraps the pop operation. cloning additional_info is the most computationally it does. But agreed that there's a possibility of scope creep with future changes that might be worth guarding against. 

If we move the Mutex into `SessionTracker` (while keeping it behind an Arc) we'll no longer be able to get the returned session_state as mutable on L149. What do you think is the best way to solve that?

Two solutions I can think of:

- We could clone the state, but it means we'd have to implement cloning for all the various structs used in the handshaker / encryptor. Not sure we want that.
- We could merge `SessionsTracker` and `AttestationServer`, extracting the normalization logic currently in `pop_session_state` into a static function. But it wouldn't be as clean of a separation of concerns. IMO if we expose the native LRU instance there's a significant the risk of future changes leading to inconsistently applied normalization.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+struct SessionsTracker {
     /// PEM encoded X.509 certificate that signs TEE firmware key.
-    _tee_certificate: Vec<u8>,
-    /// Processes data from client requests and creates responses.
-    _request_handler: F,
+    tee_certificate: Vec<u8>,
     /// Configuration information to provide to the client for the attestation step.
-    _additional_info: Vec<u8>,
+    additional_info: Vec<u8>,
+    known_sessions: LruCache<SessionId, SessionState>,
```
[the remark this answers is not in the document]
jul-sh: Note that this is pretty much already what is happening, with the exception that the Mutex is held while normalizing SessionState (which is v cheap). The mutex isn't locked while processing a request, but locked separately for retrieval & return.

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
conradgrobler: I don't see why it would not be able to be mutable. What is the compiler error? This removes it from the LRU cache, so there should no longer be anything that has a reference to it once it is returned from this function, so you should be able to make it mutable.

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
conradgrobler: I know it is quite cheap, but wouldn't agree that it is very cheap. Cloning the additional info, which could be large, and handling errors by generating strings from the error details (which could also be large) might take some time. It would be preferable not to have such an important mutex locked for that entire time.

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
tiziano88: I also would have thought that it's sufficient to limit the lock here, given the entry is removed (popped) from the LRU cache, so there is no need to hold references or clone. Or am I not understanding what pop actually does?

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+                ))),
+                Err(error) => Err(format!(
+                    "Couldn't create self attestation behavior: {:?}",
+                    error
+                )),
+            },
+            Some(SessionState::HandshakeInProgress(handshaker)) => {
+                // Completed handshakers are functionally just wrap an
+                // encryptor. In that case the underlying handshaker is
+                // returned, ensuring consistent state representation.
```
tiziano88, line 75: Is this actually supposed to happen? I would have thought that when the handshake completes, then the encryptor gets inserted in the cache directly. If it does happen at the moment, would it make sense to "fast forward" the session state to the encryptor *before* putting it back into the cache instead?

jul-sh: Yes, that does happen at the moment. We could indeed move the check for completion to a later point, and directly insert the encryptor here. 

I do not think we want that however, I intentionally placed this check here to increase resilience & maintainability. While this PR can take care to always insert the encryptor directly we may well forget to do so in later code changes. At that point we may conceivably arrive in a situation where depending on the code pathway, sometimes the encryptor is inserted directly, and sometimes it isn't. The resulting inconsistent state representation would be undesirable and may even enable security exploits down the line.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+                }
+            }
+            Some(SessionState::EncryptedMessageExchange(encryptor)) => {
+                Ok(SessionState::EncryptedMessageExchange(encryptor))
+            }
+        };
+    }
+
+    pub fn put_session_state(&mut self, session_id: SessionId, session_state: SessionState) {
+        let _ = &self.known_sessions.put(session_id, session_state);
```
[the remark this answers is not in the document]
jul-sh: thx, yeah we don't need to take the reference in which case the assignment isn't needed

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
jul-sh: See the compiler error, given on L149 & L199 with the inner mutex:

> "cannot borrow data in an `Arc` as mutable
> trait `DerefMut` is required to modify through a dereference, but it is not implemented for `std::sync::Arc<SessionsTracker>`rustc[E0596](https://doc.rust-lang.org/error-index.html#E0596)" 

For convince of debugging, I've also pushed the resulting code onto the following branch which you can clone from my fork: `unary-handler-inner-mutex`

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+    additional_info: Vec<u8>,
+    known_sessions: LruCache<SessionId, SessionState>,
+}
+
+impl SessionsTracker {
+    pub fn create(tee_certificate: Vec<u8>, additional_info: Vec<u8>) -> Self {
+        Self {
+            tee_certificate,
+            additional_info,
+            known_sessions: LruCache::new(10000),
```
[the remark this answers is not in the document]
jul-sh: To clarify, do you mean a let assignment earlier in this function's body, or a compile time const? I'm not sure it's possible to make this a compile time const.

jul-sh:
@tiziano88 thx for the review, implemented relevant changes, left comments where clarification was needed

grpc_unary_attestation/src/server.rs
tiziano88: So is the "contract" that other parts of the code may decide whether to insert the encryptor or the session state? If so it may be worth clarifying (maybe on put_session_state, and / or on this method)

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
tiziano88: I haven't tried that code, but the problem may be that the Mutex and Arc are "too far" from each other. The usual idiom is to have `Arc<Mutex<Thing>>` (like you have here). Also in that code I'm not sure I understand why the entire SessionTracker is inside an Arc -- shouldn't there be a single instance of it ever? Who / what creates other references to it? Anyways, if that's needed, then maybe an additional Arc may be needed to wrap the Mutex anyways? Not sure though

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,92 @@
...
+    pub fn pop_session_state(
+        &mut self,
+        session_id: SessionId,
+    ) -> anyhow::Result<SessionState, String> {
+        return match self.known_sessions.pop(&session_id) {
+            None => match AttestationBehavior::create_self_attestation(&self.tee_certificate) {
+                Ok(behavior) => Ok(SessionState::HandshakeInProgress(Box::new(
+                    ServerHandshaker::new(behavior, self.additional_info.clone()),
+                ))),
+                Err(error) => Err(format!(
```
tiziano88, line 71: Consider https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context (if it works here), and / or below.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+}
+
+/// gRPC Attestation Service implementation.
+pub struct AttestationServer<F, L: LogError> {
+    /// Business logic processor, accepts decrypted request and returns responses.
+    request_handler: F,
     /// Error logging function that is required for logging attestation protocol errors.
     /// Errors are only logged on server side and are not sent to clients.
     error_logger: L,
+    sessions_tracker: Arc<Mutex<SessionsTracker>>,
```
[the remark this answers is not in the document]
tiziano88: Is the Arc here even necessary, in that case?

grpc_unary_attestation/src/server.rs
```diff
@@ -59,19 +134,75 @@ where
...
+        let request_inner = request.into_inner();
+
+        let mut session_state = {
+            self.sessions_tracker
+                .lock()
+                .expect("Couldn't lock session_state mutex")
+                .pop_session_state(request_inner.session_id)
+                .map_err(|error| {
+                    error_logger.log_error(&error);
+                    tonic::Status::internal("")
```
tiziano88, line 154: Are these error logged by tonic, or are they returned to the client? Would it help to just return the appropriate error and letting it be logged by the caller code, instead of returning empty errors here? Or would it leak information somehow?

tiziano88 (approved):
LGTM, but please wait for approval from @ipetr0v too

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
tiziano88: I meant just the magic number.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,88 @@
...
+    /// Consumes remote attestation state of an existing session. Creates
+    /// intial state if the session is not known.
+    ///
+    /// Note that getting the remote attestaion state of a session always
+    /// implicitly removes it from the set of tracked sessions. After
+    /// using the state to process a request with this state it must explicitly
+    /// be put back into the SessionsTracker. This an intentional choice meant
+    /// meant to ensure that faulty state that leads to errors when preocessing
+    /// a request is not persistent.
+    pub fn pop_session_state(&mut self, session_id: SessionId) -> Result<SessionState, String> {
```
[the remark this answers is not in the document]
tiziano88: I think we always use anyhow for "unstructured" errors (i.e. the equivalent of returning either the thing, or a message to log). For things that we expect the caller to handle differently based on the error type, then a custom Error enum may be necessary instead.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,92 @@
...
+    tee_certificate: Vec<u8>,
     /// Configuration information to provide to the client for the attestation step.
-    _additional_info: Vec<u8>,
+    additional_info: Vec<u8>,
+    known_sessions: LruCache<SessionId, SessionState>,
+}
+
+impl SessionsTracker {
+    pub fn create(tee_certificate: Vec<u8>, additional_info: Vec<u8>) -> Self {
+        let known_sessions = LruCache::new(10000);
```
ipetr0v, line 45: Could you please add this `10000` as a constant

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,92 @@
...
+        Self {
+            tee_certificate,
+            additional_info,
+            known_sessions,
+        }
+    }
+
+    /// Consumes remote attestation state of an existing session. Creates
+    /// intial state if the session is not known.
```
ipetr0v, line 54:
```suggestion
    /// initial state if the session is not known.
```

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
jul-sh: Ah, yes makes sense. Good point, thanks for clarifying.

grpc_unary_attestation/src/server.rs
```diff
@@ -18,20 +18,92 @@
...
-    _request_handler: F,
+    tee_certificate: Vec<u8>,
     /// Configuration information to provide to the client for the attestation step.
-    _additional_info: Vec<u8>,
+    additional_info: Vec<u8>,
+    known_sessions: LruCache<SessionId, SessionState>,
+}
+
+impl SessionsTracker {
+    pub fn create(tee_certificate: Vec<u8>, additional_info: Vec<u8>) -> Self {
```
ipetr0v, line 44: Why does SessionTracker need `tee_certificate` and `additional_info`?

ipetr0v: ah, I see
So it creates a Handshaker

grpc_unary_attestation/src/server.rs
ipetr0v: I think we never return any information to the client

ipetr0v (approved):

grpc_unary_attestation/src/server.rs
jul-sh: Will clarify in the comment. :) It's not necessarily an intentional "contract". It's more meant as resilience because future code changes may happen to do that, whether we think it's a good idea or not.  

Basically it's just a thing I learned from working on large codebases: it's almost impossible to be prescriptive about how a function/struct may be used in the future, so it's best to write it in a way that's resilient to it being used in all kinds of ways. 

In practical terms the alternative would be to add a comment above the `put_session_state` method saying something like: "Please only remember to always insert the raw encryptor". That probably wouldn't be very effective in practice. :)

grpc_unary_attestation/src/server.rs
jul-sh: The `error_logger` logs server side, tonic just sends the empty status (mirrors the streaming client).

Without knowing the reasoning behind the initial implementation, I think not sending a specific error message is probably fine (and keeps us on the safe side security wise). In any case, after an error the client would attempt to restart the handshake.

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
jul-sh: Investigated this a little. I think what makes this behavior a bit hard to reason about for me is the `#[tonic::async_trait]` (which is [this crate](https://github.com/dtolnay/async-trait) under the hood), which seems a little magical (or at least I'm not sure what it does precisely). Do you know more about it?

There is indeed only ever one instance of the `SessionTracker` per `AttestationServer`, and you're correct that we indeed do not need an Arc. 

Invoking the async `message` function in my understanding does start a new thread, hence why I assumed an Arc was needed. It isn't though, I assume `#[tonic::async_trait]` already provides each thread with a shared reference to the `self` object. 

Adding an Arc inside `SessionTracker` does not fix it. It also seems like it wouldn't be needed, there's only ever one session_tracker.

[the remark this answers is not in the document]
jul-sh: I would be inclined to move forward with this PR with the outer mutex for now. However I'd be keen to debug this further later, would like to understand what exactly is needed.

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
jul-sh: nope! Removed it. :) See discussion in: https://github.com/project-oak/oak/pull/2596#discussion_r823977849

grpc_unary_attestation/src/server.rs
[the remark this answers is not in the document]
tiziano88: SG, thanks

grpc_unary_attestation/src/server.rs
tiziano88: I know you didn't mean it as a literal contract, but I think we should explicit on what's allowed / expected, instead of assuming that callers will mostly do one thing, but also taking into account other options. Instead if we force ourselves to actually treat it as a contract ("you can pass it A or B") then there is less risk that in some future refactoring this defense-in-depth protection will be accidentally removed, because the person forgot that there are indeed cases that may happen. Also, if we had tests for this function (which I hope we will have soon), we will want to test all the "supported" options.

grpc_unary_attestation/src/server.rs
```diff
@@ -59,19 +131,75 @@ where
...
     F: 'static + Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
-    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync + 'static,
+    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync,
     L: Send + Sync + Clone + LogError + 'static,
 {
     async fn message(
         &self,
         request: tonic::Request<UnaryRequest>,
-    ) -> Result<tonic::Response<UnaryResponse>, tonic::Status> {
+    ) -> anyhow::Result<tonic::Response<UnaryResponse>, tonic::Status> {
```
tiziano88, line 140: This particular result is just a plain old Result, there is already structure in tonic::Status, so I don't think anyhow::Result helps here, does it?

grpc_unary_attestation/src/server.rs
```diff
@@ -59,19 +131,75 @@ where
...
-            request
-        ));
-        return Err(tonic::Status::unimplemented(""));
+        // Note that we only get here if no errors occured during the preceding
+        // steps. If errors do occur the session state as tracked by the server
+        // is effectively erased. This allows the client to negotiate a new
+        // handshake.
+        self.sessions_tracker
+            .lock()
+            .expect("Couldn't lock session_state mutex")
```
tiziano88, line 199: `session_tracker` ?

tiziano88 (approved):

grpc_unary_attestation/src/server.rs
jul-sh: Happy to remove it. :) Just used it because my impression was were using anyhow everywhere

grpc_unary_attestation/src/server.rs
jul-sh: good catch, thx!

grpc_unary_attestation/src/server.rs
jul-sh: It is, they map over over to gRPC status codes: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

jul-sh:
@tiziano88 @ipetr0v thanks for the reviews. :)

PR #2601, opened 2022-03-09
Read i32 list from Wasm memory

mariaschett:


oak_functions/loader/src/server.rs
```diff
@@ -1291,4 +1386,17 @@ mod tests {
...
+    // Helper function for testing to write a buffer of i32s in Wasm memory.
+    fn write_i32_buffer_to_wasm_memory(
+        wasm_state: &mut WasmState,
+        i32_buffer: &Vec<i32>,
+        dest: AbiPointer,
+    ) -> Result<(), ChannelStatus> {
+        let mut bytes_buffer: Vec<u8> = Vec::new();
+        for i in i32_buffer {
+            bytes_buffer.extend_from_slice(&i.to_le_bytes());
+        }
```
tiziano88, line 1399: If you want, you could also extract the core functionality of those two functions (converting bytes to i32, and vice versa), and then have a property based test to show that they are one the inverse of the other (as long as the buffer is correctly sized, of course).

mariaschett:
We don't have an immediate use for this helper function, so we are closing it.

mariaschett:
No immediate use for this helper.

PR #2603, opened 2022-03-10
Propose approach for a modular implementation of the Oak Functions runtime

conradgrobler:
This is not inteded to be an actual implementation, but just a proposal for a possible way we could approach implementing #2551.


experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,59 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{Service, ServiceType};
+use std::{collections::BTreeMap, sync::Arc};
+
+/// A stream demultiplexer.
+pub struct Demux {
```
rbehjati, line 21: Currently `Demux` is only used in `IoListener`. Is the plan to use it anywhere else in future?

experimental/modular_oak_functions/src/session.rs
```diff
@@ -0,0 +1,67 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{Service, ServiceProxy};
+use std::sync::Arc;
+
+/// Session state service.
+pub struct SessionService {
+    // Per-session state will also be stored here, likely in a LRU cache.
```
rbehjati, line 22: Why is a `Session` required? Is it related to the support for `unary` requests? Are `per-session` and `per-request` synonymous now?

experimental/modular_oak_functions/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    };
+    let wasm: Arc<Box<dyn Service>> = Arc::new(Box::new(wasmi::WasmiService::new(services)));
+
+    // Create the policy enforcment service that forwards data to the Wasm sandbox service.
+    let session: Arc<Box<dyn Service>> =
+        Arc::new(Box::new(session::SessionService::new(wasm.clone())));
+
+    // Create the policy enforcment service that forwards data to the Wasm sandbox service.
+    let policy: Arc<Box<dyn Service>> =
+        Arc::new(Box::new(policy::PolicyService::new(session.clone())));
```
rbehjati, line 79: My understanding is that with this setup the policy will be applied at the session-level, instead of being applied to each request. Is that what we want? 
Also, is it possible, or does it make sense, to have an application that creates a `SessionService` with a `PolicyService` as the `next`, and the `WasmiService` as the `next` for `PolicyService`?

experimental/modular_oak_functions/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+pub mod lookup;
+pub mod policy;
+pub mod session;
+pub mod wasmi;
+
+use maplit::btreemap;
+use std::{collections::BTreeMap, sync::Arc};
+
+/// Manages the shared state for a service.
+pub trait Service {
```
rbehjati, line 29: I feel like `SessionService` and `PolicyService` are a very specific types of `Service`. In particular, I guess, they'd never be passed to `WasmiService`, and will never be associated with an ABI function. So, I wonder if it is correct, conceptually, to have them as a `Service`.

Also, can we have an application that does not use `SessionService` or `PolicyService`?

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+We have previously discussed using a more general term, such as "Service" rather
+than "Extension". My suggestion is to use `Service` as a replacement for
+`ExtensionFactory` and `ServiceProxy` as a replacement for `Extension`.
+
+The shared state used by an extension is generally owned and managed by the
+extension factory, and each extension is specific to a single `WasmState`
+instance that manages the Wasm execution for a single request. I also considered
+`ServiceFactory` and `Service` as respective replacements, but seeing that the
+`ServiceFactory` would also manage shared state it seems somewhat inaccurate.
```
jul-sh, line 18: Happy with extension as the name, as long as we can answer "What are we extending?". Is it the `WasmState`?

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Design Approach
+
+The `Service` implementation for a feature would manage shared state and create
```
jul-sh, line 22: shared between whom?

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Each service implementation would decide whether it must be able to call other
+services, and if so provide the ability to inject these services into a
+constructor. Each service will also be able to receive configuration information
+through a standardised interface.
+
+This single approach would support most of the existing components in the Oak
+Functions runtime, but we will need some additional services for things that do
+not operate at the same per-request level. One example is a stream
+demultiplexer. It will receive one multiplexed stream and split it into
```
jul-sh, line 35: > one example is a stream demultiplexer

is the implementation of this planned already? What would we use this for? 
Could it be implemented as part of an untrusted launcher instead? (the untrusted launcher as proposed in [this architecture diagram](https://docs.google.com/drawings/d/1X_H279lIdp91-je2r8Ar1Z6JNlRdXGo7mhhPdLDAuZ0/edit?usp=sharing&resourcekey=0-2ZrJ_rcJ2iDfhyOMEu35Sw))

experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,59 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{Service, ServiceType};
+use std::{collections::BTreeMap, sync::Arc};
+
+/// A stream demultiplexer.
```
jul-sh, line 20: Slightly confused by this, I read the earlier section on the design approach as indicating that a stream demultiplexer wouldn't be possible with the proposed design. 

> `ServiceProxy` instances that can be used to perform per-request service-related
> actions, and interact with the shared state owned by its parent service where
> needed.
> 
> Each service implementation would decide whether it must be able to call other
> services, and if so provide the ability to inject these services into a
> constructor. Each service will also be able to receive configuration information
> through a standardised interface.
> 
> This single approach would support most of the existing components in the Oak
> Functions runtime, but we will need some additional services for things that do
> not operate at the same per-request level. One example is a stream
> demultiplexer.

experimental/modular_oak_functions/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Creates a new service proxy
+    fn create_proxy(&self) -> Box<dyn ServiceProxy>;
+
+    /// Configures the servcice with the given data.
+    ///
+    /// Each service is responsible for interpreting the bytes approriately.
+    ///
+    /// Most services should only react to receiving configuration data once, except for the lookup
+    /// service which will periodically receive lookup data refreshes through this interface.
+    fn configure(&self, data: &[u8]) -> anyhow::Result<()>;
```
jul-sh, line 39: When is a service expected to be configured? Once, at startup time? Repeatedly? Where would this method be invoked?

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Naming
+
+We have previously discussed using a more general term, such as "Service" rather
+than "Extension". My suggestion is to use `Service` as a replacement for
+`ExtensionFactory` and `ServiceProxy` as a replacement for `Extension`.
+
+The shared state used by an extension is generally owned and managed by the
+extension factory, and each extension is specific to a single `WasmState`
+instance that manages the Wasm execution for a single request. I also considered
```
jul-sh, line 16: To clarify, after this we would no longer use extensions? This code would replace it?

experimental/modular_oak_functions/src/io.rs
```diff
@@ -0,0 +1,42 @@
...
+//
+
+use crate::demux::Demux;
+
+pub struct IoListener {
+    demux: Demux,
+}
+
+impl IoListener {
+    pub fn new(demux: Demux) -> Self {
```
ipetr0v, line 24: Will it take a socket file as input?

experimental/modular_oak_functions/src/log.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl LogProxy {
+    fn new() -> Self {
+        Self {}
+    }
+}
+
+impl ServiceProxy for LogProxy {
+    fn call(&self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
```
ipetr0v, line 59: How will log communicate with the outside world? Using the same multiplexed IO ?

ipetr0v (approved):

experimental/modular_oak_functions/README.md
jul-sh: makes sense, thanks for explaining

jul-sh (approved):

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+instances. This document proposes a generalisation of extensions and the ideas
+mentioned in issue [#2551](https://github.com/project-oak/oak/issues/2551).
+
+## Naming
+
+We have previously discussed using a more general term, such as "Service" rather
+than "Extension". My suggestion is to use `Service` as a replacement for
+`ExtensionFactory` and `ServiceProxy` as a replacement for `Extension`.
+
+The shared state used by an extension is generally owned and managed by the
```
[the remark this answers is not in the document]
conradgrobler: All these services will be defined at compile time, so will have an effect on the has of the binary. So the remote attestation report will indirectly depend on them through the binary hash.

experimental/modular_oak_functions/README.md
conradgrobler: I don't think so, as we won't have any gRPC in the TCB.

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+The `Service` implementation for a feature would manage shared state and create
+`ServiceProxy` instances that can be used to perform per-request service-related
+actions, and interact with the shared state owned by its parent service where
+needed.
+
+Each service implementation would decide whether it must be able to call other
+services, and if so provide the ability to inject these services into a
+constructor. Each service will also be able to receive configuration information
+through a standardised interface.
```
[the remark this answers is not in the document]
conradgrobler: It could be arbitrary, but in our case it is mostly linear. It is only the Wasm service that allows the business logic to choose different services to send data to.

experimental/modular_oak_functions/README.md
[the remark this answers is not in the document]
conradgrobler: We will only have one communication channel inside the TCB, so we can't create a new connection for every client connection.

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,75 @@
...
+approach can easily be extended to support the experimental services such as
+differentially private metrics and ML inference.
+
+All services are currently implemented to do passthrough or echo just to show
+how data might flow through the system.
+
+In a real implementation each module will be implemented in a separate crate,
+and, where possible, will be `no_std`-compatible.
+
+The implementation starts with a configuration flow:
```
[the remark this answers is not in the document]
conradgrobler: Updated the text a bit to try to clarify.

experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct Demux {
+    config_services: BTreeMap<ServiceType, Arc<Box<dyn Service>>>,
+    next: Arc<Box<dyn Service>>,
```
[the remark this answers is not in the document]
conradgrobler: Added documentation.

experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,59 @@
...
+            config_services,
+            next,
+        }
+    }
+
+    /// Handles a single admin frame from a multiplexed stream.
+    ///
+    /// This will likely be part of `handle_frame` in a real implementation, just with different
+    /// frametypes, but it is split out here for clarity.
+    pub fn handle_admin_frame(&self, data: &[u8]) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
conradgrobler: Updated the terminology.

experimental/modular_oak_functions/src/log.rs
conradgrobler: This will likely use the existing stdio devices to send logs.

experimental/modular_oak_functions/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    let lookup: Arc<Box<dyn Service>> = Arc::new(Box::new(lookup::LookupService::new()));
+
+    // Create Wasm sandbox service with references to the services it can use.
+    let services: BTreeMap<ServiceType, Arc<Box<dyn Service>>> = btreemap! {
+        ServiceType::Log => log.clone(),
+        ServiceType::Lookup => lookup.clone(),
+    };
+    let wasm: Arc<Box<dyn Service>> = Arc::new(Box::new(wasmi::WasmiService::new(services)));
+
+    // Create the policy enforcment service that forwards data to the Wasm sandbox service.
```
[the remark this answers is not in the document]
conradgrobler: Fixed.

experimental/modular_oak_functions/src/main.rs
conradgrobler: Fixed the order.

experimental/modular_oak_functions/src/io.rs
conradgrobler: Updated the comment above. In different environments we will have different implementation of the IoListener: e.g. vsock or UDS listener for Linux-based implementations running in a VM, serial listener (or virtio-pci) for the UEFI implementation, etc.

experimental/modular_oak_functions/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+pub trait ServiceProxy {
+    /// Processes the provided data, potentially calling other service proxies and returns the
+    /// result.
+    fn call(&self, data: &[u8]) -> anyhow::Result<Vec<u8>>;
+}
+
+/// The types of services we can support.
+#[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Ord)]
+pub enum ServiceType {
+    Log,
```
mariaschett, line 52: Does it make sense to distinguish between services serving Wasm ABI calls (e.g., lookup, metrics, tf, logging) and Wasm/Policy/Log?

experimental/modular_oak_functions/src/wasmi.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl ServiceProxy for WasmiProxy {
+    fn call(&self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        eprintln!("wasm engine invoked");
+        let mut result = data.to_vec();
+        // The real implemenation will execute the Wasm business logic and call the appropriate
+        // service proxies based on ABI calls. For now, just call each proxy and replace the
+        // previous result every time.
+        for (_, proxy) in self.proxies.iter() {
```
mariaschett, line 66: I am not sure what you mean by `For now just call each proxy...`---why?

experimental/modular_oak_functions/src/log.rs
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{Service, ServiceProxy};
+
+/// Service to provide loggin functionality.
```
mariaschett, line 19: Will we have one logger or many? Who will be using the logger?

mariaschett (commented):
Nice, thank you Conrad!

A high-level question: Do we have a plan how to get from the current state to the proposed state? 

I think it would be good to have some intermediate steps and not do all at once. WDYT?

experimental/modular_oak_functions/README.md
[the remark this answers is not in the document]
rbehjati: > the untrusted loader will combine all the requests into a single stream

I think it would be nice to include this in the explanation, if it is not already there. It was not clear to me.

rbehjati (approved):
Nice :)

experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Handles a single control frame from a multiplexed stream.
+    ///
+    /// This will likely be part of `handle_frame` in a real implementation, just with different
+    /// frame types, but it is split out here for clarity.
+    pub fn handle_control_frame(&self, data: &[u8]) -> anyhow::Result<()> {
+        eprintln!("control frame decapsulated");
+        for (_, service) in self.config_services.iter() {
+            service.configure(data)?;
+        }
```
tiziano88, line 47: Not sure what's happening here -- is the same data being sent to all services? I would expect something in the frame to specify who the data is for?

experimental/modular_oak_functions/src/demux.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /// Handles a single control frame from a multiplexed stream.
+    ///
+    /// This will likely be part of `handle_frame` in a real implementation, just with different
+    /// frame types, but it is split out here for clarity.
+    pub fn handle_control_frame(&self, data: &[u8]) -> anyhow::Result<()> {
+        eprintln!("control frame decapsulated");
+        for (_, service) in self.config_services.iter() {
```
tiziano88, line 45: Consider https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

I think it may read something like

```rust
self.config_services.iter().map(|service| service.configure(data)).collect()?
```

(but I haven't actually tried)

experimental/modular_oak_functions/src/log.rs
```diff
@@ -0,0 +1,67 @@
...
+        Self {}
+    }
+}
+
+impl ServiceProxy for LogProxy {
+    fn call(&self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
+        eprintln!("log called");
+        // The real implementation will use the logger. For now we just interpret the data as a UTF8
+        // string, write to stderr and echo the data base.
+        eprintln!("logged: {}", std::str::from_utf8(data)?);
```
tiziano88, line 64: Conceptually, I think this is doing too much stuff for a proxy. I would expect that all the interesting functionality is actually implemented in the Service itself, and the proxy just has an Arc or some other references to that, and it can invoke that functionality (this is also consistent with my other suggestion of using Client / Service terminology, but even with the current terminology it still applies IMO).

In this case for instance, the implementation of this method may be something like

```
fn call(&self, data: &[u8]) -> anyhow::Result<Vec<u8>> {
  self.service.log(data)
}
```

In fact, maybe the concept of Proxy / Client can just be abstracted completely, so that a Proxy / Client is literally just an Arc for a Service (i.e. we only need to manually implement Services, no Proxies / Clients))

experimental/modular_oak_functions/src/demux.rs
conradgrobler: Yes, this is just a placeholder to ensure we get logging responses from all the config sections. The actual implementation would route control frames to the right service based on an identifier in the control frame.

experimental/modular_oak_functions/src/log.rs
conradgrobler: Having thought about this more, it might collect the public per-request logs and send them through the same multiplexed IO to the external untrusted loader so that it can log appropriately.

experimental/modular_oak_functions/src/log.rs
conradgrobler: There will be one logging service implementation per runtime. This logging service will be used primarily by the Wasm workload to do debug logging.

experimental/modular_oak_functions/src/log.rs
conradgrobler: I agree that this is doing more than a proxy should as mentioned in earlier comments on the naming. In an actual implementation it is likely that the log service client/instance will collect the per-request logs and these would be potentially included in the response to the client (if the client requested debug information).

experimental/modular_oak_functions/src/wasmi.rs
conradgrobler: This is just a placeholder implementation that makes one call to each of the proxies just to show a potential data flow.

experimental/modular_oak_functions/README.md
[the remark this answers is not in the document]
conradgrobler: Added additional clarification to the README.

experimental/modular_oak_functions/README.md
```diff
@@ -0,0 +1,80 @@
...
+We currently have ExtensionFactories that generate Extensions. Extensions
+provide the functions available to the per-request Wasm business logic
+instances. This document proposes a generalisation of extensions and the ideas
+mentioned in issue [#2551](https://github.com/project-oak/oak/issues/2551).
+
+## Naming
+
+We have previously discussed using a more general term, such as "Service" rather
+than "Extension". My suggestion is to use `Service` as a replacement for
+`ExtensionFactory` and `ServiceProxy` as a replacement for `Extension`.
```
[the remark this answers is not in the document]
tiziano88: I still think that the client should do almost nothing apart from invoking the server. The server can keep per-client (which is also per-request) state, perhaps by passing an Arc to additional data back and forth with each call. Or, we can just have the client have the logic, if you prefer. Even in the metrics case, that would be similar to a gRPC client that does some aggregation before sending the data to a server. I think client / server is still an appropriate and useful terminology.

deny.toml
```diff
@@ -16,6 +16,8 @@ ignore = [
   "RUSTSEC-2020-0071",
   # TODO(#2421): Remove when no longer required.
   "RUSTSEC-2021-0127",
+  # TODO(#2652): Remove when Wizer is updated.
+  "RUSTSEC-2022-0016",
```
tiziano88, line 20: I suggest splitting this out in a separate PR so it can be submitted ASAP

conradgrobler: Done in #2653

conradgrobler:
Converting this PR to a draft for now as it is on hold for the foreseeable future.

conradgrobler:
With the move to the restricted kernel we have gone in a significantly different direction, so I am closing this for now.

We can revisit the modular construction in future in a way that fits better with the restricted kernel.

PR #2605, opened 2022-03-10
Create a skeleton UEFI app

andrisaar:
This adds a skeleton UEFI "hello world" The whole goal is to have something that we can start integrating into our build/CI systems.

Case in point, for now `experimental/uefi` needs to be excluded in the root `Cargo.toml` because commands that execute from the root directory (eg `cargo udeps`) fail. This is because Cargo will only read the `.cargo/config.toml` file from the current durectory, but right now the only way to set all of it up properly is to use a `.cargo/config.toml` under `experimental/uefi` as we need some experimental features. This means that if you run, say, `cargo doc` from `experimental/uefi` it will succeed, but running it from the root directory will fail.

There is hope that the `per-package-target` Cargo feature will (eventually) help us, but we'll need to wait for https://github.com/rust-lang/cargo/issues/9451 (or something equivalent) to be submitted before we can start using it.


experimental/uefi/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+#![no_std]
+#![feature(abi_efiapi)]
+#![feature(custom_test_frameworks)]
+#![test_runner(crate::test_runner)]
+#![reexport_test_harness_main = "test_main"]
+
+use uefi::{prelude::*, table::runtime::ResetType, ResultExt};
+
+#[entry]
+fn _start(_handle: Handle, mut system_table: SystemTable<Boot>) -> Status {
```
tiziano88, line 27: Does this need to be named `_start` or is it just a convention? Either way, could you link to some reference material?

experimental/uefi/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    let status = if cfg!(test) {
+        #[cfg(test)]
+        test_main();
+        Status::SUCCESS
+    } else {
+        main(_handle, &mut system_table)
+    };
+
+    system_table
```
tiziano88, line 38: Could you add some comment here to explain what's happening (and why)?

experimental/uefi/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#![no_main]
+#![no_std]
+#![feature(abi_efiapi)]
+#![feature(custom_test_frameworks)]
+#![test_runner(crate::test_runner)]
+#![reexport_test_harness_main = "test_main"]
```
tiziano88, line 22: Where does the `test_main` string come from? Is it important that it stays like that, or can it be changed? Please add some details

experimental/uefi/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+        .runtime_services()
+        .reset(ResetType::Shutdown, status, None);
+}
+
+fn main(_handle: Handle, system_table: &mut SystemTable<Boot>) -> Status {
+    use core::fmt::Write;
+
+    let status = writeln!(system_table.stdout(), "Hello World!");
+
+    match status {
```
tiziano88, line 48: Can this be expressed more "functionally"? e.g. `status.map(|_| SUCCESS).or(DEVICE_ERROR)` or something like that

experimental/uefi/src/main.rs
tiziano88, line 27: I imagine some of this will end up being part of some generic harness library, while some other parts will be Oak-specific. Not in this PR, but it may be worth start thinking about that split going forward.

xtask/src/main.rs
```diff
@@ -681,37 +681,41 @@ fn run_cargo_test(opt: &RunTestsOpt, all_affected_crates: &ModifiedContent) -> S
...
             // Exclude `fuzz` crates, as there are no tests and binaries should not be executed.
             .filter(|path| !is_fuzzing_toml_file(path))
-            .map(to_string)
-            .filter(|path| all_affected_crates.contains(path))
+            .filter(|path| all_affected_crates.contains_path(path))
             .map(|entry| {
                 let test_run_step = |name| Step::Single {
                     name,
-                    command: Cmd::new(
+                    command: Cmd::new_in_dir(
```
tiziano88, line 688: Could you add a comment explaining why this is necessary?

tiziano88 (commented):
Thanks Andri, cool stuff!

experimental/uefi/src/main.rs
andrisaar: I assume it can be changed; I've added a link to the blog post this code draws upon heavily for the testing infrastructure.

experimental/uefi/src/main.rs
andrisaar: Just a convention. I've added some comments.

experimental/uefi/src/main.rs
andrisaar: I could do `status.map_or_else(|_| DEVICE_ERROR, |_| SUCCESS)` but quite frankly I think that's more opaque and hard to read than that simple match.

tiziano88: I agree, `map_or_else` is quite unreadable unless you know exactly which way around the args are.

I think the version of what I suggested that *actually* works would be `status.map(|_| SUCCESS).unwrap_or(ERROR)` (which I personally think reads like idiomatic Rust).

andrisaar: Yeah, that's much better. Done.

experimental/uefi/src/main.rs
tiziano88: Great, thanks! This stuff is pretty new for everyone, including people familiar with "vanilla" Rust, so expect to have to over communicate / document things :)

tiziano88 (approved):

conradgrobler (approved):
Nice, thanks.

jul-sh:
nice, was interesting read :)

PR #2606, opened 2022-03-10
Switch to a newer Buster snapshot and rate-limit apt-get

andrisaar:
Right now rebuilding the Oak docker image often fails when running `apt-get`; this PR attempts to alleviate that.

First, switch the base to debian/snapshot:buster-20220228. The original idea was that maybe our previous snapshot was too old, but that doesn't seem to be the case.
Second, add a rate-limit to `apt-get`. That seems to make things better.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,10 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # > /etc/apt/sources.list
 
 # Getting curl and certificates dependecies.
+# We're rate-limiting HTTP requests to 500 kB/s as otherwise we may get timeout errors
+# when downloading from snapshot.debian.org.
 RUN apt-get --yes update \
-  && apt-get install --no-install-recommends --yes \
+  && apt-get install --no-install-recommends --yes --option Acquire::http::Dl-Limit=500 \
   apt-transport-https \
   build-essential \
   ca-certificates \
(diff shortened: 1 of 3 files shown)


tiziano88 (approved):

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Use fixed snapshot of Debian to create a deterministic environment.
 # Snapshot tags can be found at https://hub.docker.com/r/debian/snapshot/tags
-ARG debian_snapshot=sha256:eeed67d1ae0846429668170fdab5e8f7ed884234db1b2c8075471bd8365cf3a7
+ARG debian_snapshot=sha256:89f4b36d2e15a87382699808a2ca96c6da5f1db928eb0a5cd8bfa38e4edcc189
```
tiziano88, line 3: You still need to call `./scripts/docker_build` and `./scripts/docker_push` though, otherwise this is not really getting used.

Dockerfile
```diff
@@ -16,7 +16,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 # Getting curl and certificates dependecies.
 RUN apt-get --yes update \
-  && apt-get install --no-install-recommends --yes \
+  && apt-get install --no-install-recommends --yes -o Acquire::http::Dl-Limit=500 \
```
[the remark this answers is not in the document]
andrisaar: Done.

Dockerfile
tiziano88: FYI still need to fix this before submitting.

andrisaar: Done; I've added the updated files in a separate commit.

tiziano88 (approved):
Thanks! Feel free to merge, make sure the commits are squashed together and that the final message makes sense (GitHub otherwise just concatenates the individual commit messages)

PR #2607, opened 2022-03-14
Implement unary request client

jul-sh:


grpc_unary_attestation/src/client.rs
```diff
@@ -0,0 +1,118 @@
...
+                response = client
+                    .message(UnaryRequest {
+                        body: request,
+                        session_id,
+                    })
+                    .await
+                    .context("Couldn't send client hello message")?
+                    .into_inner();
+            }
+        }
```
jul-sh, line 79: This loop matches the streaming client pretty closely, but I'm not 100% happy with it. The reason is it expects an implicit correlation between `handshaker.is_completed()` being false (the loop condition), and the handshaker returning `Some` request from `handshaker.next_step(&response.body)`. If that's ever not the case there will be an infinite loop. Tracked this issue for later refactoring in #2608.

grpc_unary_attestation/src/client.rs
```diff
@@ -0,0 +1,118 @@
...
+        let mut handshaker = ClientHandshaker::new(
+            AttestationBehavior::create_peer_attestation(expected_tee_measurement),
+            server_verifier,
+        );
+        let client_hello = handshaker
+            .create_client_hello()
+            .context("Couldn't create client hello message")?;
+
+        let mut response = client
+            .message(UnaryRequest {
```
tiziano88, line 56: It may be nice to at some point abstract the underlying client here, in something like a trait that allows sending and receiving messages for a session.

Something like

```rust
trait Transport {
  async fn call(session_id: SessionId, req: &[u8]) -> Result<Vec<u8>>;
}
```

or perhaps even better:

```rust
trait Transport {
  type S: Session;
  async fn create_session() -> S;
}

trait Session {
  async fn call(req: &[u8]) -> Result<Vec<u8>>;
}
```

(or some variation on it)

Then we can implement different remote attestation variants on top of this transport.

Also it may even be possible to define the entire remote attestation protocol in such a way that it is agnostic of whether the underlying protocol is unary or streaming. The "interesting" part would be written only once in a struct similar to the one proposed in this PR, but it would be parametrized with a type implementing Transport, of which we will have a streaming version and a unary version. Not sure if it would actually work though, would like to hear your opinion @jul-sh @ipetr0v .

tiziano88, line 56: It may be nice to at some point abstract the underlying client here, in something like a trait that allows sending and receiving messages for a session.

Something like

```rust
trait Transport {
  async fn call(session_id: SessionId, req: &[u8]) -> Result<Vec<u8>>;
}
```

or perhaps even better:

```rust
trait Transport {
  type S: Session;
  async fn create_session() -> S;
}

trait Session {
  async fn call(req: &[u8]) -> Result<Vec<u8>>;
}
```

(or some variation on it)

Then we can implement different remote attestation variants on top of this transport.

Also it may even be possible to define the entire remote attestation protocol in such a way that it is agnostic of whether the underlying protocol is unary or streaming. The "interesting" part would be written only once in a struct similar to the one proposed in this PR, but it would be parametrized with a type implementing Transport, of which we will have a streaming version and a unary version. Not sure if it would actually work though, would like to hear your opinion @jul-sh @ipetr0v .

tiziano88: Additionally, then this entire struct can itself implement the same trait in turn, so that callers can just treat it as a Transport. That may open the door to more interesting use cases such as IDLs and sharing types / logic across domains (e.g. native vs Wasm)

grpc_unary_attestation/src/client.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use crate::proto::{unary_session_client::UnarySessionClient, UnaryRequest};
+use anyhow::Context;
+use oak_functions_abi;
+use oak_remote_attestation::handshaker::{
+    AttestationBehavior, ClientHandshaker, Encryptor, ServerIdentityVerifier,
+};
+use tonic::transport::Channel;
+
+type SessionId = u64;
```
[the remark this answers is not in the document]
ipetr0v: I agree
Having a number as a session ID may result in the HTTP string representation always having a different length.

grpc_unary_attestation/src/client.rs
jul-sh: Agreed, I did actually play around with this locally. Doing this for the client is relatively straightforward, was planning on refactoring both clients to implement the same trait in a later PR.

Doing this for the server is a bit tricker, the method type signatures must be different since they implement different protos.

jul-sh: I think for the purposes of this PR it's fine to not have an abstraction, but I think we can & probs want to do what you proposed in a follow up PR.

ipetr0v (approved):

grpc_unary_attestation/src/client.rs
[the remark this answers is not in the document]
jul-sh: Agreed I like the idea. I would be inclined to keep u64 as effectively a mock for the initial implementation however, it matches the other PR. The specific type doesn't matter much for runtime logic and atm the SessionId is pretty much just a mock anyways.,

PR #2610, opened 2022-03-15
Run docker build

jul-sh:
Ran docker build (and push), it yielded these hashes. Seems like the present ones represent a previous snapshot, this updates them.


mariaschett (approved):
Great, thanks!

PR #2615, opened 2022-03-16
Use unary requests in the echo example

jul-sh:


oak_functions/client/rust/src/lib.rs
```diff
@@ -48,14 +63,35 @@ impl Client {
...
-            .and_then(|rsp| Response::decode(rsp.as_ref()).context("Could not decode the response"))
+}
+
+#[async_trait]
+impl Session for Client {
+    async fn invoke(&mut self, request: Request) -> anyhow::Result<Response> {
+        handle_response(self.inner.send(request).await)
+    }
+}
+
```
jul-sh, line 74: This doesn't define the `new` method (despite the shared type signature), as the compiler warns that not all futures involved implement the `Send` trait (which the async-trait macro requires by default. 

async-trait does offer a [non-thread safe option](https://crates.io/crates/async-trait), but not sure if it's safe to use. Probably fine to keep it like though.

tiziano88: Yeah the current version seems fine, also I believe there is never a case in which the new method needs to be invoked from a dyn trait object, it's always called on the concrete type, so lifting it into this trait would probably not be appropriate anyways.

tiziano88 (approved):

FORCE_CI
```diff
@@ -1 +1 @@
-8
+9
```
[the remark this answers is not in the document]
jul-sh: image

i'm just the messenger 😅

oak_functions/examples/echo/example.toml
```diff
@@ -15,5 +15,6 @@ additional_args = ["--config-path=./oak_functions/examples/echo/config.toml"]
 rust = { Cargo = { cargo_manifest = "oak_functions/client/rust/Cargo.toml" }, additional_args = [
   "--uri=http://localhost:8080",
   "--request=test",
-  "--expected-response-pattern=test"
+  "--expected-response-pattern=test",
+  "--unary-request-model"
```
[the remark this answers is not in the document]
jul-sh: idk, ideally compile time flag would be nicest, but will do for now. :)

oak_functions/client/rust/src/lib.rs
jul-sh: Would potentially make the main.rs slightly nicer, but agree not strictly needed

jul-sh:
@tiziano88 thx for the review!

PR #2619, opened 2022-03-17
Run CI for --scope=all

rbehjati:


.github/workflows/ci.yaml
```diff
@@ -80,7 +80,7 @@ jobs:
       - name: Run command
         if: github.event_name == 'pull_request'
         run: |
-          ./scripts/docker_run ./scripts/xtask --commands --scope=commits:1 ${{ matrix.cmd }}
+          ./scripts/docker_run ./scripts/xtask --commands --scope=all ${{ matrix.cmd }}
```
rbehjati, line 83: I could merge the conditional steps into one, now that they are the same.

tiziano88: Yeah it may make sense to do that; if we need to rollback we can revert the entire commit anyways.

tiziano88 (approved):

PR #2621, opened 2022-03-17
Use byte array as session_id

jul-sh:
Addressees https://github.com/project-oak/oak/pull/2607#discussion_r826410844

### DIFF
--- a/grpc_unary_attestation/src/server.rs
+++ b/grpc_unary_attestation/src/server.rs
@@ -142,17 +142,10 @@ where
     ) -> anyhow::Result<tonic::Response<UnaryResponse>, tonic::Status> {
         let error_logger = self.error_logger.clone();
         let request_inner = request.into_inner();
-        let session_id: SessionId = {
-            let boxed_slice: Box<SessionId> = request_inner
-                .session_id
-                .into_boxed_slice()
-                .try_into()
-                .map_err(|error| {
-                    error_logger.log_error(&format!("Received malformed session_id: {:?}", error));
-                    tonic::Status::invalid_argument("")
-                })?;
-            *boxed_slice
-        };
+        let session_id: SessionId = request_inner.session_id.try_into().map_err(|error| {
+            error_logger.log_error(&format!("Received malformed session_id: {:?}", error));
+            tonic::Status::invalid_argument("")
+        })?;
 
         let mut session_state = {
             self.session_tracker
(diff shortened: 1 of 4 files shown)


grpc_unary_attestation/src/server.rs
```diff
@@ -140,12 +142,23 @@ where
...
+            let boxed_slice: Box<SessionId> = request_inner
+                .session_id
+                .into_boxed_slice()
+                .try_into()
+                .map_err(|error| {
+                    error_logger.log_error(&format!("Received malformed session_id: {:?}", error));
+                    tonic::Status::invalid_argument("")
+                })?;
+            *boxed_slice
+        };
```
jul-sh, line 156: Not super happy with this, I wish we could declare the array length in the proto instead, but protobuf does not support this. :/ Thoughts?

grpc_unary_attestation/src/server.rs
```diff
@@ -140,12 +142,23 @@ where
...
         let error_logger = self.error_logger.clone();
         let request_inner = request.into_inner();
+        let session_id: SessionId = {
+            let boxed_slice: Box<SessionId> = request_inner
+                .session_id
+                .into_boxed_slice()
+                .try_into()
+                .map_err(|error| {
+                    error_logger.log_error(&format!("Received malformed session_id: {:?}", error));
+                    tonic::Status::invalid_argument("")
```
tiziano88, line 152: Could you create a bug to revisit these errors (throughout the existing code) at some point? It would be useful to return **something** to the client if any error happens, as long as it does not leak any unwanted internal state or private data.

grpc_unary_attestation/src/server.rs
tiziano88: I think this is as good as we can get with protos, but see my other comment re: returning an actionable error message to the client in case the precondition on the session id length is not met.

grpc_unary_attestation/src/server.rs
```diff
@@ -140,12 +142,23 @@ where
     ) -> anyhow::Result<tonic::Response<UnaryResponse>, tonic::Status> {
         let error_logger = self.error_logger.clone();
         let request_inner = request.into_inner();
+        let session_id: SessionId = {
+            let boxed_slice: Box<SessionId> = request_inner
```
tiziano88, line 146: Is there no direct conversion available without going through the boxed slice?

tiziano88 (approved):

grpc_unary_attestation/src/server.rs
jul-sh: done,  thx

jul-sh:
@tiziano88 thx for the review

PR #2623, opened 2022-03-18
Migrate Metrics, Mobilenet examples to unary communication

jul-sh:


oak_functions/examples/metrics/client/rust/src/main.rs
```diff
@@ -19,7 +19,7 @@
 use anyhow::Context;
 use clap::Parser;
 use oak_functions_abi::proto::Request;
-use oak_functions_client::{Client, Session};
+use oak_functions_client::{Session, UnaryClient as Client};
```
tiziano88, line 22: Would it be possible to avoid renaming the import? It's rather unusual (in my experience, at least), unless it's actually necessary to avoid a collision.

tiziano88 (approved):

oak_functions/examples/metrics/client/rust/src/main.rs
jul-sh: Not strictly needed, removed

PR #2624, opened 2022-03-21
Extract read/write to Wasm memory in extensions

mariaschett:


oak_functions/loader/src/metrics.rs
```diff
@@ -100,15 +100,13 @@ impl OakApiNativeExtension for PrivateMetricsExtension<Logger> {
     }
 }
 
-/// Corresponds to the host ABI function [`report_metric`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#report_metric).
-fn report_metric(
-    wasm_state: &mut WasmState,
+pub fn read_args(
     extension: &mut PrivateMetricsExtension<Logger>,
```
conradgrobler, line 104: It seems like the only reason that extension is actdually passed here is for logging. I think it would be cleaner to not have the logging in here, but return a result with a more detailed error. The error can then be logged by the caller, which will have the extension reference already.

conradgrobler (approved):

oak_functions/loader/src/metrics.rs
conradgrobler: It does not matter which logger does the logging. Any logger that is accessible at the point can be used. The only reason for having multiple loggers is so that we don't have to rely on some shared singleton logger that would require synchronisation. I think we should revisit the logger design and see whether we could simplify it.

mariaschett: Perfect, this allows me to remove the logic of reading arguments/writing results from the core functionality of each extension---getting closer to an `Vec<u8>` to `Vec<u8>` interface. Could you have another look? Thank you!

conradgrobler (approved):
Nice, thanks.

PR #2625, opened 2022-03-22
Build all variants of Functions Server by default

mariaschett:
- remove --server-variant flag from CI by always building all variants
- print manifest path in step to build rust binary flag

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -54,7 +54,7 @@ jobs:
           ./scripts/build_provenance \
           "oak_functions" \
           ./target/x86_64-unknown-linux-musl/release/oak_functions_loader_base \
-          ./scripts/xtask build-functions-server
+          ./scripts/xtask build-functions-server-variants
 
       - name: Move new provenance files to the provenance branch
         if: github.event_name == 'push'
--- a/xtask/src/examples.rs
+++ b/xtask/src/examples.rs
@@ -17,6 +17,7 @@
 use crate::{files::*, internal::*};
 use maplit::hashmap;
 use std::collections::HashMap;
+use strum::IntoEnumIterator;
 
 #[cfg(target_os = "macos")]
 const DEFAULT_SERVER_RUST_TARGET: &str = "x86_64-apple-darwin";
@@ -259,15 +260,10 @@ pub fn run_functions_examples(opt: &RunFunctionsExamples, scope: &Scope) -> Step
 
 /// Build every variant of the function server.
 pub fn build_functions_server_variants(opt: &BuildFunctionsServer) -> Step {
-    let variants = vec![
-        "oak_functions/oak_functions_loader_base",
-        "oak_functions/oak_functions_loader_unsafe",
-    ];
     Step::Multiple {
         name: "cargo build all variants of function server".to_string(),
-        steps: variants
-            .iter()
-            .map(|path| build_rust_binary(path, opt, &hashmap! {}))
+        steps: FunctionsServerVariant::iter()
+            .map(|variant| build_rust_binary(&variant.path_to_manifest(), opt, &hashmap! {}))
             .collect(),
     }
 }
--- a/xtask/src/examples.rs
+++ b/xtask/src/examples.rs
@@ -257,6 +257,7 @@ pub fn run_functions_examples(opt: &RunFunctionsExamples, scope: &Scope) -> Step
     }
 }
 
+/// Build every variant of the function server.
 pub fn build_functions_server_variants(opt: &BuildFunctionsServer) -> Step {
     let variants = vec![
         "oak_functions/oak_functions_loader_base",
@@ -295,8 +296,8 @@ fn run_functions_example(example: &FunctionsExample) -> Step {
         steps: vec![
             example.construct_application_build_steps(),
             if opt.run_server.unwrap_or(true) {
-                // Build the server first so that when running it in the next step it will start up
-                // faster.
+                // Build (all variants of) the server first so that when running a variant in the
+                // next step it will start up faster.
                 vec![build_functions_server_variants(&opt.build_server)]
             } else {
                 vec![]
--- a/xtask/src/examples.rs
+++ b/xtask/src/examples.rs
@@ -259,6 +259,8 @@ pub fn run_functions_examples(opt: &RunFunctionsExamples, scope: &Scope) -> Step
 }
 
 /// Build every variant of the function server.
+/// It's easier to always build all variants than to control which variant to build and
+/// the overhead of building all variants is acceptable.
 pub fn build_functions_server_variants(opt: &BuildFunctionsServer) -> Step {
     Step::Multiple {
         name: "cargo build all variants of function server".to_string(),
--- a/xtask/src/examples.rs
+++ b/xtask/src/examples.rs
@@ -257,18 +257,18 @@ pub fn run_functions_examples(opt: &RunFunctionsExamples, scope: &Scope) -> Step
     }
 }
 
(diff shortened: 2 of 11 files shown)


xtask/src/examples.rs
```diff
@@ -363,10 +360,7 @@ pub fn build_functions_example(opt: &RunFunctionsExamples, scope: &Scope) -> Ste
             functions_example.construct_application_build_steps(),
             // Build the server first so that when running it in the next step it will start up
```
tiziano88, line 361: Can you clarify in a comment here or elsewhere that we always build all server variants, even if the particular example being built / run only needs one?

tiziano88 (approved):

xtask/src/examples.rs
mariaschett: Done (on `build_functions_server_variants`).

mariaschett:
@tiziano88 Could you fix on Github so this run passes?

tiziano88:
> @tiziano88 Could you fix on Github so this run passes?

There are still a couple of CI steps running, could you fix those and then let me know and I'll change the config?

tiziano88:
The config is only about which steps are required before merging, it does not change anything about their execution.

It looks like the rest of the workflows need to be fixed in this PR first, e.g. 

https://github.com/project-oak/oak/blob/36c698cbae3376a951a16fca0211910453918132/.github/workflows/reproducibility.yaml#L59-L60

xtask/src/internal.rs
```diff
@@ -196,10 +197,20 @@ impl std::str::FromStr for FunctionsServerVariant {
     }
 }
 
+impl FunctionsServerVariant {
+    // Get path to manifest for the variant.
+    pub fn path_to_manifest(&self) -> String {
```
tiziano88, line 202: (nit) this can return `&'static str`

xtask/src/examples.rs
```diff
@@ -257,18 +258,16 @@ pub fn run_functions_examples(opt: &RunFunctionsExamples, scope: &Scope) -> Step
...
-        opt,
-        &hashmap! {},
-    )
+/// Build every variant of the function server.
+/// It's easier to always build all variants than to control which variant to build and
+/// the overhead of building all variants is acceptable.
+pub fn build_functions_server_variants(opt: &BuildFunctionsServer) -> Step {
+    Step::Multiple {
+        name: "cargo build all variants of function server".to_string(),
+        steps: FunctionsServerVariant::iter()
```
tiziano88, line 267: Nice

tiziano88 (approved):

.github/workflows/provenance.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
           ./scripts/build_provenance \
           "oak_functions" \
           ./target/x86_64-unknown-linux-musl/release/oak_functions_loader_base \
```
rbehjati, line 56: Is this name still valid? How many binaries are generated now? We probably need to update `./scripts/build_provenance` to generate multiple provenances. No?

mariaschett: `build-functions-server-variants` currently builds the two binaries for the two variants `base` and `unsafe`.

If intrpret lines 54-56 correctly, it seems we did not have a provenance for `unsafe` before. Do we want one now?

rbehjati: You are right. That path is still valid. I don't think we need a provenance for the `unsafe` version. At least not now. But if there is going to be a `bare-metal` version in the future, we'd probably need that. But maybe this change is not related to that. (I don't quite have the context for this change 🙂)

tiziano88: @rbehjati  BTW once we add TOML files for each "transparently released" binary, we should consider adding one for the unsafe version too. In fact, since they are both generated by the same command now, that would be a nice use case to show that we can track provenance of multiple artifacts generated from the same process (i.e. perhaps a single TOML file will specify both binaries as output, to match the current logic). We'll see when we get there anyways

tiziano88:
@mariaschett I have updated the repo settings, if all the CI steps pass, you should be able to merge now

.github/workflows/provenance.yaml
rbehjati: I think, from the perspective of generating provenances, it is better to have the possibility of generating binaries separately. 

We can have a single provenance referencing multiple subjects, but in this particular case, the two binaries have very different security properties, and while it is possible to have separate claims for them, having a human user (e.g., an auditor) constantly look for `oak_functions_loader_base` vs `oak_functions_loader_unsafe` is not good usability.

rbehjati: I think it makes sense to have a single command that generates a number of related binaries or libraries. But having a single command for generating a number of **alternative** binaries with opposing security properties, and including them all in a single provenance sounds like a bad design to me, as it can increase the chance of mistakes.

mariaschett: Ah, sorry, merged this already. Shall we open an issue together?

rbehjati: Can we have the previous `build_functions_server`, but with a new `all` variant (perhaps as default) to generate all variants as it now is in `build_functions_server_variants`? Also, I still don't know why we need to generate all variants together :)

mariaschett: Building binaries both simplifies the interface of `xtask`, because the server variant can be set (and conflict) in two places: in (i) `example.toml` and (ii) with a command line flag. 

Now if `build_function_server_variants` always builds both variants we can remove (ii).

But: I was only thinking from the perspective of *calling* the binary (in an example), where I need to specify the variant anyway---and by building both before I know whatever I am picking is up-to-date.

Now I am also not sure if I fully understand the problem with building them in one go for provenance?

rbehjati: > the server variant can be set (and conflict) in two places: in (i) example.toml and (ii) with a command line flag

Do you mean the `run-functions-example` command has a `server-variant` flag? 

> Now I am also not sure if I fully understand the problem with building them in one go for provenance?

I think including both `base` and `unsafe` variants in the subject of a single provenance file is wrong, as I explained above. We could use `build_function_server_variants` in the provenance, and include only the `base` variant in the subject of the provenance. But then the provenance could be considered misleading, because the command is actually generating an `unsafe` version too. The command could potentially be generating other malicious files too, which the provenance does not say anything about. I don't think we have a way of documenting in the provenance that "the command generates additional binaries that have similar functionality, but very different security properties than the subject of the provenance". Moreover, from the perspective of generating provenances, the choice of using a command that produces a bunch of other binaries too sounds very random. I think keeping the previous flags, and adding an `all` variant is a simple solution that can solve both problems with little effort.

tiziano88: A couple of points:

- what I suggested was just a single TOML file for two expected outputs, but each output would still have its own provenance file; I think for our use cases we never expect more than one output in a provenance file; anyways I agree it is probably not even worth doing this with the TOML file only, it may just be confusing
- that said, I think it's fine to have a single command generate multiple binaries. In fact, any command probably generates thousands of files in the process (e.g. intermediate artifacts in the target dir, cache entries, etc.), which is why the TOML file needs to know which one is the actual output file to measure.

rbehjati: Ah, I thought the idea was to use a single TOML file and generate a single provenance containing both variants in the subject.

> any command probably generates thousands of files in the process (e.g. intermediate artifacts in the target dir, cache entries, etc.),

True. I still *feel* like it is better to use a command that generates the fewest number of files, and ideally only the ones that are required for generating the file in the subject.

PR #2636, opened 2022-03-25
Use Wizer to enable Wasm module warmup

conradgrobler:
Wizer runs an initialisation function on a Wasm module, takes a snapshot of the state and updates the module binary so that the state for new instances match the post-warmup state.

This is particularly useful for the weather lookup example. The S2 geometry library we use populates lookup tables during the first S2 cell lookup. This is expensive and we don't want to repeat it for every request. We use Wizer to create a version of the Wasm module where the startup state contains the initislised lookup tables:

```
$ cargo bench --manifest-path=oak_functions/examples/weather_lookup/module/Cargo.toml 
    Finished bench [optimized] target(s) in 0.28s
     Running unittests (target/release/deps/weather_lookup-19f777eae18e99bf)

running 3 tests
test tests::test_server ... ignored
    Finished release [optimized] target(s) in 0.18s
test tests::bench_wasm_handler_no_warmup   ... bench:   1,394,769 ns/iter (+/- 123,624)
    Finished release [optimized] target(s) in 0.17s
test tests::bench_wasm_handler_with_warmup ... bench:     726,146 ns/iter (+/- 44,674)

```

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -4,7 +4,7 @@
 // - https://code.visualstudio.com/docs/remote/devcontainerjson-reference
 {
   // Do not modify manually. This value is automatically updated by ./scripts/docker_build .
-  "image": "sha256:d2ac0f6ad41502fdfe336b8074fbf1e4eff462fdb39b164f1cfba419f5eb3a60",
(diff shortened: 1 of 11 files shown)


jul-sh (approved):
Nice! Pretty excited about this

PR #2637, opened 2022-03-25
Migrate experimental example to unary requests

jul-sh:
Overlooked this one, sorry


conradgrobler (approved):
The PR is still marked as Draft, but I assume that it is actually ready for review.

PR #2640, opened 2022-03-29
Add ABI function `invoke` calling `testing` extension

mariaschett:
Next, we want to generalize the ABI function `invoke_testing` to `invoke` taking an handler as argument.


oak_functions/abi/src/lib.rs
```diff
@@ -81,4 +81,11 @@ extern "C" {
         inference_ptr_ptr: *mut *mut u8,
         inference_len_ptr: *mut usize,
     ) -> u32;
+
+    pub fn invoke(
+        args_ptr: *const u8,
+        args_len: usize,
+        result_ptr_ptr: *mut *mut u8,
+        result_len_ptr: *mut usize,
```
tiziano88, line 89: suggestion: request / response (see other comment)

PR #2643, opened 2022-03-29
Add a second serial port to the system.

andrisaar:
The first serial port we have is used for the UEFI stdio, so let's add a second port that's under our control. Right now it just echos everything it receives back to the sender, and writes the bytes received into the log (which will be written to stdio, so the first serial port).

### DIFF
--- a/experimental/uefi/.cargo/config.toml
+++ b/experimental/uefi/.cargo/config.toml
@@ -7,7 +7,7 @@ runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_C
 
 # Otherwise, (a) the first serial port gets routed to a log, and (b) the second serial gets attached to stdio.
 [target.'cfg(not(test))']
-runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_CODE.fd -serial file:console.log -serial stdio -machine q35 -device isa-debug-exit,iobase=0xf4,iosize=0x04 -kernel"
+runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_CODE.fd -serial file:target/console.log -serial stdio -machine q35 -device isa-debug-exit,iobase=0xf4,iosize=0x04 -kernel"
 
 [unstable]
 build-std = ["core", "alloc"]
--- a/experimental/uefi/.cargo/config.toml
+++ b/experimental/uefi/.cargo/config.toml
@@ -1,9 +1,14 @@
 [build]
 target = "x86_64-unknown-uefi"
 
-[target.x86_64-unknown-uefi]
+# Test runners get only one serial port, routed to stdio.
+[target.'cfg(test)']
 runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_CODE.fd -serial stdio -machine q35 -device isa-debug-exit,iobase=0xf4,iosize=0x04 -kernel"
 
+# Otherwise, (a) the first serial port gets routed to a log, and (b) the second serial gets attached to stdio.
+[target.'cfg(not(test))']
+runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_CODE.fd -serial file:console.log -serial stdio -machine q35 -device isa-debug-exit,iobase=0xf4,iosize=0x04 -kernel"
+
 [unstable]
-build-std = ["core"]
+build-std = ["core", "alloc"]
 build-std-features = ["compiler-builtins-mem"]
--- a/experimental/uefi/src/main.rs
+++ b/experimental/uefi/src/main.rs
@@ -17,13 +17,24 @@
 #![no_main]
 #![no_std]
 #![feature(abi_efiapi)]
+#![feature(never_type)]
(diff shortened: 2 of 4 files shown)


andrisaar:
> Just checking: is this actually how we are expecting things to work? I thought we expect to multiplex all IO onto a single communication channel, which would also help with making a version of the trusted runtime that can be controlled by another binary (e.g. an untrusted loader that exposes an HTTP interface).

I think it's better to have a complete separate (and clean) channel for our communication. Building the untrusted loader is the next step and it'll just mean we get two pipes instead of one; that doesn't sound much more complicated to handle.

experimental/uefi/src/main.rs
```diff
@@ -17,13 +17,24 @@
...
+#![feature(never_type)]
 #![feature(custom_test_frameworks)]
 // As we're in a `no_std` environment, testing requires special handling. This
 // approach was inspired by https://os.phil-opp.com/testing/.
 #![test_runner(crate::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 
-use uefi::{prelude::*, table::runtime::ResetType, ResultExt};
+#[macro_use]
+extern crate log;
```
conradgrobler, line 28: Just a note for future consideration: Using the log crate is fine for now as we are still early in the development. The actual runtime will disable the log crate to ensure that other libraries don't accidentally log sensitive information and use a custom logger.

experimental/uefi/.cargo/config.toml
```diff
@@ -1,9 +1,14 @@
 [build]
 target = "x86_64-unknown-uefi"
 
-[target.x86_64-unknown-uefi]
+# Test runners get only one serial port, routed to stdio.
+[target.'cfg(test)']
 runner = "qemu-system-x86_64 -nodefaults -nographic -bios /usr/share/OVMF/OVMF_CODE.fd -serial stdio -machine q35 -device isa-debug-exit,iobase=0xf4,iosize=0x04 -kernel"
 
+# Otherwise, (a) the first serial port gets routed to a log, and (b) the second serial gets attached to stdio.
```
conradgrobler, line 8: I assume this is temporary, and the second serial port will be routed to a UDS that is used by the untrusted loader for communication.

conradgrobler (approved):

conradgrobler:
> > Just checking: is this actually how we are expecting things to work? I thought we expect to multiplex all IO onto a single communication channel, which would also help with making a version of the trusted runtime that can be controlled by another binary (e.g. an untrusted loader that exposes an HTTP interface).
> 
> I think it's better to have a complete separate (and clean) channel for our communication. Building the untrusted loader is the next step and it'll just mean we get two pipes instead of one; that doesn't sound much more complicated to handle.

I agree that keeping stdio separate from the actual communication channel is fine. QEMU will own the stdio and we can use that for debug output during development. The loader will own and communicate over the other channel.

experimental/uefi/src/main.rs
```diff
@@ -51,14 +62,57 @@ fn _start(_handle: Handle, mut system_table: SystemTable<Boot>) -> Status {
...
+    let serial = bt.open_protocol::<Serial>(
+        OpenProtocolParams {
+            handle: *serial_handle,
+            agent: handle,
+            controller: None,
+        },
+        OpenProtocolAttributes::Exclusive,
+    )?;
+    // Dereference the raw pointer we get to the serial interface.
+    let serial = unsafe { &mut *serial.interface.get() };
```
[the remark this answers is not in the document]
tiziano88: Even if unsafe is necessary, I think it can probably be more confined, something like 
 
```
let serial = &mut unsafe { *serial.interface.get() };
```

Or does that not work?

experimental/uefi/src/main.rs
```diff
@@ -51,14 +62,57 @@ fn _start(_handle: Handle, mut system_table: SystemTable<Boot>) -> Status {
...
+fn echo_loop(serial: &mut uefi::proto::console::serial::Serial) -> Result<!, uefi::Error<()>> {
+    let mut buf: [u8; 1024] = [0; 1024];
+
+    loop {
+        // read() returns Ok if it managed to fill the whole buffer, or the error will contain
+        // the number of bytes read. The only error we're fine with is TIMEOUT, as we can simply
+        // retry that (and we'll keep getting TIMEOUTs when nobody is talking to us). In case of
+        // any other error, bail out.
+        let len = serial.read(&mut buf).map(|_| buf.len()).or_else(|err| {
+            if err.status() == Status::TIMEOUT {
```
tiziano88, line 82: Does it mean that if no one is talking to us, this will be a busy loop? Or is there an actual timeout that it waits for incoming data before hitting this?

tiziano88 (approved):

experimental/uefi/.cargo/config.toml
andrisaar: Yep, the untrusted loader will need to set up sockets for these serial ports.

experimental/uefi/src/main.rs
andrisaar: Yes, this is a busy loop, for now.

By default, UEFI gives us (a) an one-second timeout and (b) a buffer size of one byte. Both of these can be changed (and we definitely need to make the buffer larger to not lose any data in any real scenario); in the future we'd likely need to do some kind of flow control to ensure that we're not sending data while we haven't processed what's in the buffer yet.

Issue #2641, opened 2022-03-29
Generate a list of `Handles` from the configuration of the Oak Functions runtime


mariaschett:
We want to generate a list of handles for the extensions currently activated in the Oak Functions runtime, e.g., currently

- [oak-tf](https://github.com/project-oak/oak/blob/910bf1a7b8c036099213bc6c527c63986d6a8b97/oak_functions/loader/src/lib.rs#L202)
- [oak-metrics](https://github.com/project-oak/oak/blob/910bf1a7b8c036099213bc6c527c63986d6a8b97/oak_functions/loader/src/lib.rs#L214)

This list of handles should also include the lookup extension: 
- [lookup](https://github.com/project-oak/oak/blob/910bf1a7b8c036099213bc6c527c63986d6a8b97/oak_functions/loader/src/lib.rs#L200)

The Wasm module use these handles to address the respective extensions. 

@tiziano88

PR #2644, opened 2022-03-30
Implement basic Oak IDL via Rust macros

tiziano88:


mariaschett:
Related we want to keep the Handles in sync with the Extensions provided by the runtime ( #2641) .

tiziano88:
The current model is that a service is exposed over a channel, but it is made up of one or more methods. Usually methods that are related to each other would be in the same service, similar to how traits or interfaces or RPC services already work. The correspondance between a channel handle and the type of the service exposed over that channel instead is not part of the IDL. This is conceptually similar to how a gRPC service is reachable at a particular IP address / port, which is not statically bound to the service implementation. So, the expected use case for this IDL would be that the caller creates a transport over a channel handle, and uses that to establish a "connection" to a service. For instance, in Oak, the entirety of the Oak ABI may be implemented as a single service, and reachable at a given channel handle. The IDL does not generate the mapping to the channel handle, but an Oak SDK may trivially pair up the channel handle with the client generated by the IDL, and provide an even simpler way of accessing the service (even though that matching logic will have to be manually kept in sync with the server implementation). Additional methods (e.g. testing) may be defined on the same service, or as separate services. Even in case in which testing methods are added to the same service, the runtime may just return errors if they are invoked in a non-testing version of the runtime; in other words, it's not necessary to have a separate service definition (though it is possible to do so if desired).

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+        // The serialization of the outer message is fixed by the transport, since the header needs
+        // to be deserializable by the other end of the transport, but the body of the
+        // request / response is serialized at a higher level, and in potentially different format,
+        // since that's handled by the application.
+        let request_bytes =
+            bincode::serialize(&request_message).map_err(|_| TransportError::InvalidMessage)?;
+        let response_bytes = invoke(self.0, &request_bytes)?;
+        let response_message =
+            bincode::deserialize(&response_bytes).map_err(|_| TransportError::InvalidMessage)?;
+        Ok(response_message)
```
conradgrobler, line 89: Seeing that invoke is transactional, I think it would be a good idea to validate that the transaction id of the response is the same as the one in the request and return an error if not. Possibly the same for the method id.

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+                        transport
+                    }
+                }
+
+                $(
+                    pub fn $method(&self, $request_arg: $request_type) -> Result<$response_type, $crate::ClientError> {
+                        let request_body = bincode::serialize(&$request_arg).map_err(|_| $crate::ClientError::InvalidRequest)?;
+                        let request_message = $crate::Message {
+                            header: $crate::Header {
+                                transaction_id: 0,
```
conradgrobler, line 134: I assume setting this to 0 is temporary. Will this be incremented automatically by the client?

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Unique identifier of a method within a service.
+type MethodId = u32;
+
+/// Header of a transactional message.
+///
+/// See <https://fuchsia.dev/fuchsia-src/reference/fidl/language/wire-format#transactional-messages>
+#[derive(Serialize, Deserialize)]
+pub struct Header {
+    pub transaction_id: u32,
```
conradgrobler, line 53: Can you clarify what the transaction_id does? Is it supposed to be globally unique, unique in combination with the method id?

conradgrobler (approved):
Nice idea.

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+/// itself. For instance, the Message may be serialized via bincode, but the paload may be
+/// serialized via protobuf.
+#[derive(Serialize, Deserialize)]
+pub struct Message {
+    pub header: Header,
+    pub body: Vec<u8>,
+}
+
+/// A message-oriented transport that allows performing invocations.
+pub trait Transport {
```
mariaschett, line 71: high-level question: how much do we want to stay with the mental models of "message", "transport" and "channel" versus invoke actually being a (blocking) function call (which it is at the moment)?

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// A wrapper for a message-oriented channel handle, which implements the [`Transport`] trait.
+struct Handle(u32);
+
+impl Transport for Handle {
+    fn invoke(&self, request_message: &Message) -> Result<Message, TransportError> {
+        // The serialization of the outer message is fixed by the transport, since the header needs
+        // to be deserializable by the other end of the transport, but the body of the
+        // request / response is serialized at a higher level, and in potentially different format,
+        // since that's handled by the application.
```
mariaschett, line 83: application == Wasm module / Extension?

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// This macro generates the following objects (assuming it is invoked with the first argument set
+/// to `Name`):
+///
+/// - a struct named `NameClient`, exposing a method for each method defined in the macro.
+/// - a struct named `NameServer`, which implements the [`Transport`] trait, dispatching each
+///   request to the appropriate method on the underlying service implementation.
+/// - a trait named `Name`, with a method for each method defined in the macro, and an additional
+///   default method named `serve` which returns an instance of `NameServer`; the developer of a
```
mariaschett, line 106: So `serve` would correspond to `invoke`, e.g., in the lookup extension here: https://github.com/project-oak/oak/blob/a978dc57fec21d62bcbe0d4f59e0d97245ef2e96/oak_functions/loader/src/lookup.rs#L83

oak_idl/tests/oak_abi.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use oak_idl::*;
+
+// An example of a (subset of) the Oak ABI expressed as an Oak Service.
+//
+// Note that this service is exposed on a single channel handle, and the numbers correspond to
+// method identifiers within that channel.
+service! {
+    Oak {
+        42 => fn lookup_data(key: Vec<u8>) -> Option<Vec<u8>>;
```
mariaschett, line 24: For this mapping from (channel) handle to extensions/service we need to make sure it is in sync with the runtime too.

I believe the handles would roughly take the place of the `*_ABI_FUNCTION_NAME` (e.g., https://github.com/project-oak/oak/blob/a978dc57fec21d62bcbe0d4f59e0d97245ef2e96/oak_functions/loader/src/lookup.rs#L31) in the OakFunctions runtime.

We'll also need to make sure it is in sync with the (feature-gated) configuration of binary of the runtime, i.e., if we expose `report_metric` as a service the binary needs to be compiled with feature=oak-metrics.

oak_idl/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub enum TransportError {
+    InvalidMessage,
+    InvalidMethodId,
+}
+
+/// Unique identifier of a method within a service.
```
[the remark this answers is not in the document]
tiziano88: It depends, the two extremes are:

- a single service, in which each method corresponds to a different extension
- 1 service per extension, each with a single method

we probably want something in between in practice, but the IDL is not forcing one or the other.

tiziano88 (commented):
Thanks both

oak_idl/tests/oak_abi.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    #[allow(clippy::unused_unit)]
+    fn log(&self, log: String) -> () {
+        eprintln!("log: {log}");
+    }
+}
+
+#[test]
+fn foo() {
```
mariaschett, line 46: nit: `foo` vs. `minimal_lookup_example`

mariaschett (approved):
Thank you! 

Let me know how/when we use it with the ABI. I am not too sure about the UEFI requirements.

PR #2647, opened 2022-03-31
Expose remote attestation endpoint for webclients

jul-sh:
This adds an endpoint for future web clients to communicate with the runtime directly, by exposing a [grpc-web](https://github.com/grpc/grpc-web) endpoint. grpc-web is an adaption of the gRPC protocol that supports wire formats that web clients can use. 

We'll be using the binary format, see the OSS description here: https://github.com/grpc/grpc-web#wire-format-mode. Defacto this means that requests are sent as unary HTTP/1 requests, with the body consisting of binary encoded protobuf messages. 

Imo this is a nice approach to supporting OSS web clients in the linux-binary version of the runtime. The runtime itself will expose compatible endpoints, with no need to spin up a separate translation layer server. 

In prod deployments too secret to discuss in depth here we'd use a translation service. 
In the bare-metal version of the runtime the companion app will expose this endpoint.

### DIFF
--- a/oak_functions/loader/src/grpc.rs
+++ b/oak_functions/loader/src/grpc.rs
@@ -27,6 +27,7 @@ use oak_logger::OakLogger;
 use oak_utils::LogError;
 use prost::Message;
 use std::{future::Future, net::SocketAddr};
+use tonic_web;
 
 async fn handle_request(
     wasm_handler: WasmHandler,
--- a/oak_functions/loader/src/grpc.rs
+++ b/oak_functions/loader/src/grpc.rs
@@ -27,7 +27,6 @@ use oak_logger::OakLogger;
 use oak_utils::LogError;
 use prost::Message;
 use std::{future::Future, net::SocketAddr};
-use tonic_web;
 
 async fn handle_request(
     wasm_handler: WasmHandler,
@@ -98,6 +97,7 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
     // and start is handled entirely witin each respective conditional arm, as
     // the added service alters the type signature of the created server.
     if cfg!(feature = "oak-web") {
+        #[cfg(feature = "oak-web")]
         tonic::transport::Server::builder()
             .accept_http1(true)
             .add_service(grpc_unary_attestation_service.clone())
--- a/oak_functions/loader/src/grpc.rs
+++ b/oak_functions/loader/src/grpc.rs
@@ -82,21 +82,34 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
 
     let additional_info = config_info.encode_to_vec();
 
-    tonic::transport::Server::builder()
-        .add_service(
-            grpc_unary_attestation::proto::unary_session_server::UnarySessionServer::new(
-                grpc_unary_attestation::server::AttestationServer::create(
-                    tee_certificate,
-                    request_handler,
-                    additional_info,
-                    ErrorLogger { logger },
-                )
-                .context("Couldn't create remote attestation server")?,
-            ),
-        )
-        .serve_with_shutdown(*address, terminate)
-        .await
-        .context("Couldn't start server")?;
+    let grpc_unary_attestation_service =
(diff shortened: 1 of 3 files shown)


oak_functions/loader/src/grpc.rs
```diff
@@ -82,21 +82,34 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
...
+            )
+            .context("Couldn't create remote attestation server")?,
+        );
+
+    // Create a server and add the relevant service defintions. Server creation
+    // and start is handled entirely witin each respective conditional arm, as
+    // the added service alters the type signature of the created server.
+    if cfg!(feature = "oak-web") {
+        tonic::transport::Server::builder()
+            .add_service(grpc_unary_attestation_service.clone())
```
conradgrobler, line 102: Is it required to add the service twice? I would have expected `enable` to add support for grpc-web requests without breaking normal gRPC.

oak_functions/loader/src/grpc.rs
```diff
@@ -82,21 +82,34 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
...
+                additional_info,
+                ErrorLogger { logger },
+            )
+            .context("Couldn't create remote attestation server")?,
+        );
+
+    // Create a server and add the relevant service defintions. Server creation
+    // and start is handled entirely witin each respective conditional arm, as
+    // the added service alters the type signature of the created server.
+    if cfg!(feature = "oak-web") {
```
conradgrobler, line 100: I think you need to use `#[cfg(....)]`. The `tonic_web` reference would not be available if the oak-web feature is not enabled.

oak_functions/loader/src/grpc.rs
```diff
@@ -82,21 +82,34 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
...
+        tonic::transport::Server::builder()
```
conradgrobler, line 101: I think this needs `.accept_http1(true)` as there is no TLS, so the browser won't use ALPN to negotiate the HTTP version.

oak_functions/loader/src/grpc.rs
```diff
@@ -27,6 +27,7 @@ use oak_logger::OakLogger;
 use oak_utils::LogError;
 use prost::Message;
 use std::{future::Future, net::SocketAddr};
+use tonic_web;
```
conradgrobler, line 30: I don't think this is needed.

conradgrobler (approved):

oak_functions/loader/src/grpc.rs
jul-sh: good point, overlooked that indeed. added, thanks.

oak_functions/loader/src/grpc.rs
jul-sh: You're right, the documentation on it isn't super clear, but it does pass through non grpc-web requests to the underlying service. 

https://github.com/hyperium/tonic/blob/350ccf95e4ee71f85a2a1af7dc3e1df6cca1a2e3/tonic-web/src/service.rs#L158

I probably would've preferred if the enable method didn't do that (or exposed an option to not do that), but it is what it is. Updated.

jul-sh:
@conradgrobler thanks for the review. :)

oak_functions/loader/src/grpc.rs
```diff
@@ -82,21 +82,35 @@ pub async fn create_and_start_grpc_server<F: Future<Output = ()>>(
...
+                additional_info,
+                ErrorLogger { logger },
+            )
+            .context("Couldn't create remote attestation server")?,
+        );
+
+    // Create a server and add the relevant service defintions. Server creation
+    // and start is handled entirely witin each respective conditional arm, as
+    // the added service alters the type signature of the created server.
+    if cfg!(feature = "oak-web") {
```
conradgrobler, line 99: nit: you can remove the `if` statement if you mark the statement in the else clause with `#[cfg(not(...))]`.

PR #2649, opened 2022-03-31
Fix invalid xtask command in docs & scripts

jul-sh:
This command was renamed in #2612. Our docs and scritps feature older versions that don't work.

### DIFF
--- a/scripts/run_tests_coverage
+++ b/scripts/run_tests_coverage
@@ -6,7 +6,7 @@ source "$SCRIPTS_DIR/common"
 
 # Build the example client and Application in coverage mode and run them.
 export RUST_LOG=trace
-"${SCRIPTS_DIR}/xtask" --scope=all run-examples --example-name=abitest --server-variant=coverage
+"${SCRIPTS_DIR}/xtask" --scope=all run-oak-functions-examples --example-name=abitest --server-variant=coverage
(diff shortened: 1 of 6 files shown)


mariaschett (approved):
Thank you for fixing that, @jul-sh !

scripts/run_tests_coverage
```diff
@@ -6,7 +6,7 @@ source "$SCRIPTS_DIR/common"
 
 # Build the example client and Application in coverage mode and run them.
 export RUST_LOG=trace
-"${SCRIPTS_DIR}/xtask" --scope=all run-examples --example-name=abitest --server-variant=coverage
+"${SCRIPTS_DIR}/xtask" --scope=all run-oak-functions-examples --example-name=abitest --server-variant=coverage
```
rbehjati, line 9: This does not look right. We don't have an oak-fuctions "abitest" example.
Also, for oak-functions we don't have a server-variant for "coverage".

mariaschett: That is true, that seems to be a left-over of Oak classic. 

So we should probably just remove this call? 

(IIUC it does currently simply nothing because xtask will not find an example with this name---correction, just tried: it fails because we don't have the server-variant "coverage".)

PR #2655, opened 2022-04-01
Use new cargo feature resolver to avoid feature flags unexpectedly being set

jul-sh:
The default feature resolver leads to some unexpected behavior regarding features, which are set globally. This has masked errors, by unexpectedly enabling features we thought were not enabled. (See the chat discussion: http://chat/room/AAAAp6czer4/R-8PDylbqF8)

Ref: https://doc.rust-lang.org/cargo/reference/features.html#feature-resolver-version-2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
   "experimental/attestation_common",
   "experimental/https_attestation",
--- a/remote_attestation/rust/src/crypto.rs
+++ b/remote_attestation/rust/src/crypto.rs
@@ -194,11 +194,15 @@ impl KeyNegotiator {
             .map_err(|error| anyhow!("Couldn't get public key: {:?}", error))?
             .as_ref()
             .to_vec();
-        public_key.as_slice().try_into().context(format!(
-            "Incorrect public key length, expected {}, found {}",
-            KEY_AGREEMENT_ALGORITHM_KEY_LENGTH,
-            public_key.len()
-        ))
+        public_key
+            .as_slice()
+            .try_into()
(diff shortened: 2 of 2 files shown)


tiziano88:
https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions

> if the root package specifies edition = "2021" or a newer edition. Otherwise the default is "1".

I assumed we were already using the new resolver because we are using 2021 edition everywhere, or did it not work because it's not in the workspace root?

jul-sh:
> if the root package specifies edition = "2021" or a newer edition. Otherwise the default is "1".
> 
> I assumed we were already using the new resolver because we are using 2021 edition everywhere, or did it not work because it's not in the workspace root?

@tiziano88 looks like we didn't (behavior is indeed different if the key is added). That's part of what I don't love about cargo workspaces, config behaves differently/unexpectedly depending on whether a package is inside a workspace. 

On a closer look the docs do explain this, they state:

> The resolver is a global option that affects the entire workspace. The resolver version in dependencies is ignored, only the value in the top-level package will be used. If using a [virtual workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-manifest), the version should be specified in the [workspace] table, for example:

https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions


We also cannot set the `edition` in our (virtual) workspace file (as it does not specify a package), we have to explicitly set the resolver. 


We could alternatively set the edition in the workspace config, which does also implicitly set the new resolver. Do you think that'd be preferable?

tiziano88 (approved):
Thanks for getting to the bottom of this, I was totally mistaken about the current state of things!

PR #2656, opened 2022-04-01
Create an untrusted loader for the UEFI app

andrisaar:
This is a fairly straightforward implementation of an untrusted loader that wraps `qemu`. The key feature is setting up unix sockets and the plumbing so that we can just pass the file descriptors to `qemu` and not give a name to the sockets.

Next up after this: instead of relying on stdio, use gRPC.


experimental/uefi/loader/src/qemu.rs
```diff
@@ -0,0 +1,122 @@
...
+        })
+    }
+
+    pub async fn kill(mut self) -> Result<std::process::ExitStatus> {
+        println!("Cleaning up and shutting down.");
+        self.console.shutdown().await?;
+        self.comms.shutdown().await?;
+        self.qmp.shutdown().await?;
+        self.instance.kill().await.map_err(anyhow::Error::from)?;
+        self.instance.wait().await.map_err(anyhow::Error::from)
```
conradgrobler, line 120: I don't think this does anything after `kill`.

experimental/uefi/loader/src/qemu.rs
```diff
@@ -0,0 +1,122 @@
...
+        Ok(Qemu {
+            instance: cmd.spawn()?,
+            console: params.console,
+            comms: params.comms,
+            qmp: qmp.0,
+        })
+    }
+
+    pub async fn kill(mut self) -> Result<std::process::ExitStatus> {
+        println!("Cleaning up and shutting down.");
```
conradgrobler, line 115: In the untrusted world I think it would be a good idea to use the `log` crate to log at various levels rather than using println directly. `env_logger` can then be used to configure the logging.

conradgrobler (approved):
Nice.

Please add a README.md for the loader to give a short expalanation of the intended use and example steps on how to get it to run (e.g. I assume the UEFI app needs to be built first)

jul-sh (approved):
Nice! +1 to the readme, would be nice :)

experimental/uefi/loader/src/qemu.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl Qemu {
+    pub fn start(params: QemuParams) -> Result<Qemu> {
+        let mut cmd = tokio::process::Command::new(params.binary);
+
+        // There should not be any communication over stdin/stdout/stderr, but let's inherit
+        // stderr just in case.
+        cmd.stdin(Stdio::null());
+        cmd.stdout(Stdio::null());
+        cmd.stderr(Stdio::inherit());
```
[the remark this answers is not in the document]
andrisaar: Yeah, fair enough, I've set it to inherit as well. In the future what'd probably be good would be to set up a separate channel for both of these and whatever is printed on stdout/stderr gets dumped into the log as a warning, as we're not really expecting any console I/O.

experimental/uefi/loader/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+        Ok(())
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let cli = Args::parse();
+
+    let console = UnixStream::pair()?;
```
tiziano88, line 60: Consider something like

```suggestion
    let (console_tx, console_rx) = UnixStream::pair()?;
```

and below

tiziano88 (approved):

experimental/uefi/loader/src/qemu.rs
```diff
@@ -0,0 +1,122 @@
...
+        // https://www.qemu.org/docs/master/system/invocation.html for all available options.
+
+        // We're going to run qemu as a noninteractive embedded program, so disable any
+        // graphical outputs.
+        cmd.arg("-nographic");
+        // Don't bother with default hardware, such as a VGA adapter, floppy drive, etc.
+        cmd.arg("-nodefaults");
+        // Use the more modern `q35` machine as the basis.
+        // Note that q35 comes with a ton of stuff we don't need (eg a PC speaker). We should
+        // use something simpler (microvm?), if possible.
```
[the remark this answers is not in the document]
andrisaar: Filed https://github.com/project-oak/oak/issues/2679/.

experimental/uefi/loader/src/qemu.rs
andrisaar: I think you're correct, but kill() doesn't give you an `ExitStatus`.

I've changed it to `start_kill()`, which just sends the signal (and doesn't return a future), leaving the waiting for the `wait()` below.
(Either that, or we change the signature to return `()` and drop the `wait` completely.)

experimental/uefi/loader/src/main.rs
andrisaar: It's not really `tx` and `rx`, but fair enough, using that pattern is a bit more readable than accessing `.0` and `.1`.

PR #2657, opened 2022-04-04
Update fuzzable.wasm to fix CI

conradgrobler:
The file needs to be updated to resolve the CI break on the main branch.

See https://github.com/project-oak/oak/runs/5795119943?check_suite_focus=true


jul-sh:
Thanks! Some questions: I have
- What does fuzzable.wasm do?
- How was it updated?
- Do we want this check to be blocking, so we don't accidentally merge PRs to that break this? :)

jul-sh (approved):

conradgrobler:
> Thanks! Some questions: I have
> 
> * What does fuzzable.wasm do?
> * How was it updated?
> * Do we want this check to be blocking, so we don't accidentally merge PRs to that break this? :)

- It is a pre-compiled wasm module that is used during fuzzing
- Either running the CI locally, or `xtask run-cargo-fuzz --build-deps -- -max_total_time=2`
- That might be a good idea. I don't know why it is not blocking. @tiziano88 any thoughts?

PR #2661, opened 2022-04-04
Use vendored version of ring in remote attestation 

jul-sh:
Uses the vendored version of ring (added in #2681) in our remote attestation crate. This necessitated the following changes:

- Patch the build script of ring to allow it to build in our repo, see commit message / comment for details.
- Update our code to accommodate breaking API changes, since the vendored version is newer.

### DIFF
--- a/third_party/ring/src/lib.rs
+++ b/third_party/ring/src/lib.rs
@@ -48,7 +48,10 @@
     missing_debug_implementations,
     non_camel_case_types,
     non_snake_case,
-    unsafe_code
+    unsafe_code,
+    // Oak Note: Do not warn when building deprecated code. Ring includes
+    // code that is, and our CI fails on warnings
+    deprecated
 )]
 // `#[derive(...)]` uses `trivial_numeric_casts` and `unused_qualifications`
 // internally.


deny.toml
```diff
@@ -43,7 +43,6 @@ copyleft = "deny"
 
 [[licenses.clarify]]
 name = "ring"
-version = "*"
```
jul-sh, line 46: The reason this needed removing is somewhat counterintuitive: 

In cargo, `version = "*"` does not in fact mean "any version", but instead "any version that is published on crates.io". We need to remove this (optional) key to truly check any version (in our case our vendored version).

remote_attestation/rust/src/crypto.rs
```diff
@@ -234,11 +234,10 @@ impl KeyNegotiator {
     ) -> anyhow::Result<(EncryptionKey, DecryptionKey)> {
         let type_ = self.type_.clone();
         let self_public_key = self.public_key().context("Couldn't get self public key")?;
-        let (encryption_key, decryption_key) = agreement::agree_ephemeral(
+        agreement::agree_ephemeral(
             self.private_key,
             &agreement::UnparsedPublicKey::new(KEY_AGREEMENT_ALGORITHM, peer_public_key),
-            anyhow!("Couldn't derive session keys"),
-            |key_material| {
+            |key_material| -> anyhow::Result<(EncryptionKey, DecryptionKey)> {
```
jul-sh, line 240: The updated ring version no longer takes an error as a parameter, but instead returns an additional result

conradgrobler (approved):
Nice, thanks.

PR #2663, opened 2022-04-04
Clarify threat model and terminology

tiziano88:


README.md
```diff
@@ -17,12 +17,18 @@ with privileged access to the machine that handles the data) has unrestricted
 access to the machine hardware resources, and can leverage that to bypass any
 security mechanism on the machine itself and extract secret keys and data.
 
-As part of Project Oak, data are end-to-end encrypted between _enclaves_, which
-are isolated computation compartments that can be created on-demand, and provide
-strong confidentiality, integrity, and attestation capabilities via a
-combination of hardware and software functionality. Enclaves protect data and
-code even from the operating system kernel and privileged software, and are
-intended to protect from most hardware attacks.
+[Oak Functions](/oak_functions) is a Trusted Runtime server binary that exposes
```
andrisaar, line 20: Should Trusted Runtime also be defined below?

andrisaar, line 20: Should Trusted Runtime also be defined as a term?

README.md
```diff
@@ -31,46 +37,39 @@ Further information is included in the following documents:
...
 - **untrusted**:
   - most hardware (memory, disk, motherboard, network card, external devices)
   - Operating System (kernel, drivers, libraries, applications)
-  - platform provider (hardware, software, employees)
-  - third-party developers
+  - Platform Provider
+  - Untrusted Business Logic Authors
 - **trusted-but-verifiable**:
-  - Project Oak codebase (and its transitive dependencies)
+  - Oak Trusted Runtime Authors (and its transitive dependencies)
```
andrisaar, line 62: The "transitive dependencies" sounds weird here now. Do you mean our children? :)

README.md
tiziano88: Removed from here for now, but I agree we should think about that. Probably more related to the Transparent Release work anyways, so let's do another PR for that.

README.md
tiziano88: Reworded, this should also get better once we talk about Transparent Release here. cc @rbehjati

tiziano88:
@andrisaar the re-request review button in the UI does not seem to work, PTAL

andrisaar (approved):

Issue #2500, opened 2022-01-20
Design IDL for ABI


tiziano88:
We should be able to statically express the types and wire serialization of the messages that are exchanged via UWABI.

Probably makes sense to use as much as possible existing IDLs, mainly protobuf.

cc @mariaschett

mariaschett:
Some Context:

The messages currently exchanged on the channels are [`pub type AbiMessage = Vec<u8>`](https://github.com/project-oak/oak/blob/d287e2cac6c74c629b40a2d997a279ca2fc2ae76/oak_functions/loader/src/server.rs#L47).

The interpretation of the message depends on which `ChannelHandle` they are sent / the corresponding extension at the runtime. 

Currently we have 3 extensions which we want to gradually want to rewrite to [`UwabiExtensions`](https://github.com/project-oak/oak/blob/d287e2cac6c74c629b40a2d997a279ca2fc2ae76/oak_functions/loader/src/server.rs#L164) exchanging `AbiMessages`:

1. the [lookup extension](https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/metrics.rs)
2. the [metrics extension](https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/metrics.rs)
3. the [tensor flow extension](https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/tf.rs)

For example, for the lookup extension the Wasm module would send a message with the `key : Vec<u8>`. 

The runtime would return a message with a `value : Vec<u8>`, but we also need to indicate if no value was found in the message. (Currently, this currently indicated through the `OakStatus` of [`ERR_STORAGE_ITEM_NOT_FOUND`](https://github.com/project-oak/oak/blob/d287e2cac6c74c629b40a2d997a279ca2fc2ae76/oak_functions/proto/abi.proto#L64) returned by the ABI call [`storage_get_item`](https://github.com/project-oak/oak/blob/d287e2cac6c74c629b40a2d997a279ca2fc2ae76/oak_functions/abi/src/lib.rs#L82), but this will not be possible once we have UWABI.)

Similarily for the other two extensions and [error codes from `OakStatus`](https://github.com/project-oak/oak/blob/d287e2cac6c74c629b40a2d997a279ca2fc2ae76/oak_functions/proto/abi.proto#L55).

mariaschett:
@tiziano88 Can I close this issue? I think it is superseded by other ideas.

mariaschett:
[...]

PR #2676, opened 2022-04-05
Read label and value from a serialzed message in Metric extension

mariaschett:


mariaschett:
Unsure where I want to define [MetricMessage](https://github.com/project-oak/oak/blob/0e9c11932bcc3be0b7127584923be155abe07b9e/oak_functions/sdk/oak_functions/src/lib.rs#L99) in SDK and [MetricMessage](https://github.com/project-oak/oak/blob/0e9c11932bcc3be0b7127584923be155abe07b9e/oak_functions/loader/src/metrics.rs#L144) in extension

Maybe in the [abi](https://github.com/project-oak/oak/blob/0e9c11932bcc3be0b7127584923be155abe07b9e/oak_functions/abi/src/lib.rs) for them both to have access?

@tiziano88

oak_functions/abi/src/lib.rs
```diff
@@ -48,6 +49,15 @@ pub mod proto {
     }
 }
 
+/// Holds the `raw_label` and the `value` to report a metric.
+#[derive(Serialize, Deserialize)]
+pub struct ReportMetricRequest {
+    pub raw_label: Vec<u8>,
```
conradgrobler, line 55: I think this can just be `label: String`, which would let the serialization process validate that the bytes represent a string.

oak_functions/loader/src/metrics.rs
```diff
@@ -120,19 +118,23 @@ impl OakApiNativeExtension for PrivateMetricsExtension<Logger> {
...
     extension: &mut PrivateMetricsExtension<Logger>,
-    raw_label: Vec<u8>,
-    value: i64,
+    request: Vec<u8>,
 ) -> Result<(), OakStatus> {
-    let label = std::str::from_utf8(raw_label.as_slice()).map_err(|err| {
+    let request: ReportMetricRequest =
+        bincode::deserialize(&request).expect("Fail to deserialize report metric request.");
+
+    let label = std::str::from_utf8(request.raw_label.as_slice()).map_err(|err| {
```
conradgrobler, line 126: As mentioned this should no longer be required with the serialisation logic.

oak_functions/abi/src/lib.rs
```diff
@@ -48,6 +49,15 @@ pub mod proto {
     }
 }
 
+/// Holds the `raw_label` and the `value` to report a metric.
+#[derive(Serialize, Deserialize)]
+pub struct ReportMetricRequest {
```
conradgrobler, line 54: I think this should be feature-gated. It is weird to have this struct here in the cases where the metrics features is not enabled.

conradgrobler (approved):
Thanks.

It looks like `fuzzable.wasm` needs to be udpated again from the task failure.

oak_functions/abi/src/lib.rs
mariaschett: It seems that if I feature-gate `ReportMetricRequest` here, then I have in `metric.rs` an "unresolved import `oak_functions_abi::ReportMetricRequest`". 

Feature-gating in `metric.rs` seems wrong to me.

Am I missing something here?

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -107,8 +107,14 @@ pub fn report_event<T: AsRef<str>>(label: T) -> Result<(), OakStatus> {
 /// See [`report_metric`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#report_metric).
 #[cfg(feature = "oak-metrics")]
 pub fn report_metric<T: AsRef<str>>(label: T, value: i64) -> Result<(), OakStatus> {
-    let buf = label.as_ref().as_bytes();
-    let status = unsafe { oak_functions_abi::report_metric(buf.as_ptr(), buf.len(), value) };
+    let raw_label = label.as_ref().as_bytes().to_vec();
```
mariaschett, line 110: And this I can also simply replace by `String::from(label.as_ref()`---or do I miss any problem here?

oak_functions/loader/src/metrics.rs
mariaschett: Done.

oak_functions/abi/src/lib.rs
mariaschett: Beautiful, thank you!

mariaschett (commented):
Thank you, Conrad! 

Could you have a quick look wrt to feature-gating the `ReportMetricRequest` stuct?

oak_functions/abi/src/lib.rs
conradgrobler: The entire metric.rs is already feature-gated in https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/lib.rs#L25, so it should not try to build it unless the feature is enabled.

I think you need to define a feature called `oak-metrics` in the config.toml of the abi, and also update the features in the loader and SDK config.toml files to enable the oak-metrics abi feature if the oak-metrics feature is enabled on the loader, something like:
`oak-metrics = ["oak_functions_abi/oak-metrics"]`

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -107,8 +109,14 @@ pub fn report_event<T: AsRef<str>>(label: T) -> Result<(), OakStatus> {
 /// See [`report_metric`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#report_metric).
 #[cfg(feature = "oak-metrics")]
 pub fn report_metric<T: AsRef<str>>(label: T, value: i64) -> Result<(), OakStatus> {
-    let buf = label.as_ref().as_bytes();
-    let status = unsafe { oak_functions_abi::report_metric(buf.as_ptr(), buf.len(), value) };
+    let label = String::from(label.as_ref());
```
conradgrobler, line 112: I think this is equivalent, but for consistency we usually use `to_owned` to convert `&str` to `String`, e.g. `label.as_ref().to_owned()`.

conradgrobler (approved):

oak_functions/abi/src/lib.rs
tiziano88: Personally I wouldn't bother feature gating a plain struct (which will be optimized away anyways).

oak_functions/loader/src/metrics.rs
```diff
@@ -120,19 +118,16 @@ impl OakApiNativeExtension for PrivateMetricsExtension<Logger> {
...
-        ));
-        OakStatus::ErrInvalidArgs
-    })?;
-    extension.log_debug(&format!("report_metric(): {}", label));
-    extension.report_metric(label, value).map_err(|err| {
-        extension.log_error(&format!("report_metric(): {:?}", err));
-        OakStatus::ErrInternal
-    })
+    let request: ReportMetricRequest =
+        bincode::deserialize(&request).expect("Fail to deserialize report metric request.");
```
ipetr0v, line 124: Why are we using bincode instead of proto serialization?
Or it's just to encode and decode Rust structures?

ipetr0v (approved):

oak_functions/loader/src/metrics.rs
mariaschett: Yes, we just use it to decode Rust. I think we'll revisit this when we integrate with #2500

mariaschett (commented):
Thank you!

Issue #2614, opened 2022-03-15
Trusted Shuffler example


ipetr0v:
We need to have an example implementation of the Trusted Shuffler. It will be represented as a separate binary fro Oak Functions accompanied by a test client and backend.

This example can be represented by the following entities:
- `trusted_shuffler`
  - A library implementation of the Trusted Shuffler logic
  - Specifically made as a separate create, so it could later be used in the real Trusted Shuffler application
  - Shouldn't depend on any protocols: i.e. it receives bytes and sends bytes
- `backend`
  - A simple Echo service that sends back the requests
- `server`
  - A proxy server that receives responses from the client and uses the Trusted Shuffler library to shuffle them before sending them to the backend
- `client`
  - Sends multiple requests to the server
  - Can use multiple threads or async so the requests are sent in parallel

ipetr0v:
Done in #2680, #2790

PR #2680, opened 2022-04-05
Add a Trusted Shuffler initial implementation

ipetr0v:
This PR adds an initial implementation of the Trusted Shuffler struct.

Ref #2614


experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+const TEST_ANONYMITY_VALUE: usize = 10;
+const TEST_TIMEOUT_MILLISECONDS: u64 = 100;
+
+async fn send_request(request: Vec<u8>) -> Vec<u8> {
+    request
+}
+
+#[tokio::test]
+async fn trusted_shuffler_test() {
+    let trusted_shuffler = Arc::new(TrustedShuffler::new(2, send_request));
```
mariaschett, line 31: Maybe: `let anonymity_value = 2;` (and later)

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+    let background_request_clone = background_request.clone();
+    let background_result = tokio::spawn(async move {
+        trusted_shuffler_clone
+            .invoke(background_request_clone)
+            .await
+    });
+
+    let response = trusted_shuffler.invoke(request.clone()).await;
+    assert!(response.is_ok());
+    assert_eq!(request, response.unwrap());
```
mariaschett, line 46: Here we assume that our "test backend" echoes the the request, yes? Maybe we want to comment this here. Or maybe even remove this assertion. We just want to know we got a response, no?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+#[tokio::test]
+async fn trusted_shuffler_test() {
+    let trusted_shuffler = Arc::new(TrustedShuffler::new(2, send_request));
+
+    let request = "Test request".to_string().as_bytes().to_vec();
+    let background_request = "Test request background".to_string().as_bytes().to_vec();
+
+    let trusted_shuffler_clone = trusted_shuffler.clone();
+    let background_request_clone = background_request.clone();
+    let background_result = tokio::spawn(async move {
```
mariaschett, line 38: Could we use `join` in the test too and spawn tasks for both requests? I am not sure why they are treated differently.

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+    let result = timeout(
+        Duration::from_millis(TEST_TIMEOUT_MILLISECONDS),
+        response_future,
+    )
+    .await;
+    assert!(result.is_err());
+}
+
+#[tokio::test]
+async fn trusted_shuffler_background_test() {
```
mariaschett, line 75: Why `background_test`?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+pub struct TrustedShuffler<F, O>
+where
+    F: Send + Sync + 'static + FnOnce(Request) -> O,
+    O: Send + Sync + Future<Output = Response>,
+{
+    // Value k that represents k-anonymity.
+    anonymity_value: usize,
+
+    // Current batch of requests to be shuffled.
+    current_requests: Arc<Mutex<Vec<Message>>>,
```
mariaschett, line 54: Could we document why we need a mutex here?

mariaschett (commented):
That's amazing, Ivan! 

I need to look over it again, but here some inital comments.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+type Response = Vec<u8>;
+
+struct Message {
+    // Arbitrary message data.
+    data: Vec<u8>,
+    // Channel for sending responses back to the client async tasks.
+    response_sender: oneshot::Sender<Response>,
+}
+
+// Trusted Shuffler inmpelentation.
```
tiziano88, line 44:
```suggestion
// Trusted Shuffler implementation.
```

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+
+        Self::shuffle_responses(responses)
+    }
+
+    // Shuffles received backend responses and sends them back to the client async tasks using
+    // corresponding channels defined in the [`Message::response_sender`].
+    fn shuffle_responses(mut responses: Vec<Message>) -> anyhow::Result<()> {
+        responses.shuffle(&mut thread_rng());
+
+        for response in responses.drain(..) {
```
tiziano88, line 138: Does it not work just removing drain then?

```suggestion
        for response in responses {
```

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+        receiver.await.context("Couldn't receive response")
+    }
+
+    // Shuffles requests and sends them to the backend using the
+    // [`TrustedShuffler::request_sender`].
+    async fn shuffle_requests(mut requests: Vec<Message>, request_sender: F) -> anyhow::Result<()> {
+        requests.shuffle(&mut thread_rng());
+
+        let response_futures: Vec<_> = requests
+            .drain(..)
```
tiziano88, line 115: Not sure about this `drain(..)` here. What is the error if removed?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+                let request_sender_clone = request_sender.clone();
+                async move {
+                    let response = (request_sender_clone)(request.data).await;
+                    Message {
+                        data: response,
+                        response_sender: request.response_sender,
+                    }
+                }
+            })
+            .collect();
```
tiziano88, line 127: AFAICT there is no need to collect, they just need to be an iterator.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+            }
+        }
+
+        receiver.await.context("Couldn't receive response")
+    }
+
+    // Shuffles requests and sends them to the backend using the
+    // [`TrustedShuffler::request_sender`].
+    async fn shuffle_requests(mut requests: Vec<Message>, request_sender: F) -> anyhow::Result<()> {
+        requests.shuffle(&mut thread_rng());
```
tiziano88, line 112: Adding randomness seems unnecessary and actually confuses things here IMO; it should be enough to just deterministically sort requests lexicographically, right?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+            .collect();
+        let responses = join_all(response_futures).await;
+
+        Self::shuffle_responses(responses)
+    }
+
+    // Shuffles received backend responses and sends them back to the client async tasks using
+    // corresponding channels defined in the [`Message::response_sender`].
+    fn shuffle_responses(mut responses: Vec<Message>) -> anyhow::Result<()> {
+        responses.shuffle(&mut thread_rng());
```
tiziano88, line 136: At this point the responses are still sorted by the request shuffled order, right? In which case, there should be no need to re-shuffle things again. Either way, worth clarifying (and perhaps testing too)

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+
+        for response in responses.drain(..) {
+            response
+                .response_sender
+                .send(response.data.clone())
+                .map_err(|error| anyhow!("Couldn't send the response: {:?}", error))?;
+        }
+        Ok(())
```
tiziano88, line 144: I think this for loop could be something like this expression (which would be the return value of the method itself)

```rust
responses
  .into_iter()
  .map(|Message { data, response_sender }| response_sender.send(data))
  .collect()
  .map(|_: Vec<()>| ())
```

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+use crate::TrustedShuffler;
+use futures::future::join_all;
+use std::{sync::Arc, time::Duration};
+use tokio::time::timeout;
+
+const TEST_ANONYMITY_VALUE: usize = 10;
+const TEST_TIMEOUT_MILLISECONDS: u64 = 100;
+
+async fn send_request(request: Vec<u8>) -> Vec<u8> {
+    request
```
tiziano88, line 26: Maybe return something slightly different for the sake of the example, e.g. add an element or something. Also add a comment that this is just used as an example but in practice this function would talk to a backend etc.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+    sync::{Arc, Mutex},
+};
+use tokio::sync::oneshot;
+
+type Request = Vec<u8>;
+type Response = Vec<u8>;
+
+struct Message {
+    // Arbitrary message data.
+    data: Vec<u8>,
```
tiziano88, line 39: This data is always the request, right? If so, then perhaps just call it `request`, also for symmetry with the field below.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+    future::Future,
+    marker::{Send, Sync},
+    mem::replace,
+    ops::DerefMut,
+    sync::{Arc, Mutex},
+};
+use tokio::sync::oneshot;
+
+type Request = Vec<u8>;
+type Response = Vec<u8>;
```
tiziano88, line 35: Seems unnecessary IMO, but up to you.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+    }
+
+    // Asynchronously handles an incoming request.
+    pub async fn invoke(&self, request: Request) -> anyhow::Result<Response> {
+        let (sender, receiver) = oneshot::channel();
+
+        let request = Message {
+            data: request,
+            response_sender: sender,
+        };
```
tiziano88, line 80:
```suggestion
        let (response_sender, response_receiver) = oneshot::channel();

        let request = Message {
            data: request,
            response_sender,
        };
```

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+    O: Send + Sync + Future<Output = Response>,
+{
+    // Value k that represents k-anonymity.
+    anonymity_value: usize,
+
+    // Current batch of requests to be shuffled.
+    current_requests: Arc<Mutex<Vec<Message>>>,
+
+    // Async function that sends a request to the backend and returns a response.
+    request_sender: F,
```
tiziano88, line 57: The term `*_sender` is already used for the oneshot sender, so I found this confusing. How about `request_fn` or something else?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+    data: Vec<u8>,
+    // Channel for sending responses back to the client async tasks.
+    response_sender: oneshot::Sender<Response>,
+}
+
+// Trusted Shuffler inmpelentation.
+pub struct TrustedShuffler<F, O>
+where
+    F: Send + Sync + 'static + FnOnce(Request) -> O,
+    O: Send + Sync + Future<Output = Response>,
```
tiziano88, line 48: I would suggest defining your own trait (with a single function) at some point, instead of this nontrivial bound on a FnOnce. You probably will want to encapsulate the sending logic in a struct anyways, so it would feel natural to impl that trait for it.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,131 @@
...
+                });
+            }
+        }
+
+        receiver.await.context("Couldn't receive response")
+    }
+
+    // Shuffles requests and sends them to the backend using the
+    // [`TrustedShuffler::request_sender`].
+    async fn shuffle_requests(mut requests: Vec<Message>, request_sender: F) -> anyhow::Result<()> {
```
tiziano88, line 111: It seems a bit odd that this fn is defined on the struct itself, and, because it does not actually take self as arg, it needs the caller to manually pass the request_sender param which is actually on self. At that point, may as well make this a stand-alone function. I guess this saves some repetition of the trait bounds, but that's even less of an issue if using the trait as I suggested above. WDYT?

tiziano88 (commented):
Great, looks promising! I have a few comments though, I only looked at the non-test code for now.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Async functions in traits are not supported yet: https://doc.rust-lang.org/error-index.html#E0706

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: It will have to look like this:
```rust
responses
    .into_iter()
    .map(|Message { data, response_sender }| response_sender.send(data))
    .collect::<Vec<anyhow::Result<(), Vec<u8>>>>()
    .map(|_: Vec<anyhow::Result<(), Vec<u8>>>| ())
```
And even then it still doesn't want to build the vec:
```
error[E0599]: the method `map` exists for struct `Vec<Result<(), Vec<u8>>>`, but its trait bounds were not satisfied
   --> experimental/trusted_shuffler/trusted_shuffler/src/lib.rs:143:14
    |
143 |             .map(|_: Vec<anyhow::Result<(), Vec<u8>>>| ())
    |              ^^^ method cannot be called on `Vec<Result<(), Vec<u8>>>` due to unsatisfied trait bounds
    |
   ::: /usr/local/cargo/toolchains/nightly-2022-02-17-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:400:1
    |
400 | pub struct Vec<T, #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global> {
    | ------------------------------------------------------------------------------------------------ doesn't satisfy `Vec<Result<(), Vec<u8>>>: Iterator`
```

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
ipetr0v: Changed to `join_test`

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: It would create new duplication, because in the new function I'll need to also add
```rust
where
    F: Send + Sync + Clone + 'static + FnOnce(Request) -> O,
    O: Send + Sync + Future<Output = Response>,
```

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,130 @@
...
+        response_receiver.await.context("Couldn't receive response")
+    }
+
+    // Lexicographically sorts requests and sends them to the backend using the
+    // [`TrustedShuffler::request_sender`].
+    async fn shuffle_requests(
+        mut requests: Vec<Message>,
+        request_handler: F,
+    ) -> anyhow::Result<()> {
+        requests.sort_by(|first, second| first.data.cmp(&second.data));
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
tiziano88: May be better to use https://docs.rs/futures/latest/futures/stream/struct.FuturesOrdered.html explicitly though, and clarify in a comment why this was chosen.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,130 @@
...
+        });
+        let responses = join_all(response_futures).await;
+
+        Self::shuffle_responses(responses)
+    }
+
+    // Lexicographically sorts received backend responses and sends them back to the client async
+    // tasks using corresponding channels defined in the [`Message::response_sender`].
+    fn shuffle_responses(mut responses: Vec<Message>) -> anyhow::Result<()> {
+        responses.sort_by(|first, second| first.data.cmp(&second.data));
```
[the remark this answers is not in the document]
ipetr0v: But if we don't shuffle back, wouldn't the network attacker also deasonymize clients: because the responses will arrive to Oak and depart from Oak with the same order?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Yes, I used `async_trait` too in UWABI. Here the [PR](https://github.com/project-oak/oak/pull/2639/files).

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,123 @@
...
+fn generate_response(request: Vec<u8>) -> Vec<u8> {
+    let parsed_request = String::from_utf8(request).expect("Couldn't parse request");
+    format!("Response for: {}", parsed_request)
+        .to_string()
+        .as_bytes()
+        .to_vec()
+}
+
+// Generates a request and a corresponding response from a string.
+fn generate_request_response(data: &str) -> (Vec<u8>, Vec<u8>) {
```
mariaschett, line 42: Here the respone only echoes the request, no? Is there any reason to have that for the future? Or am I missing something?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,123 @@
...
+    let result = timeout(
+        Duration::from_millis(TEST_TIMEOUT_MILLISECONDS),
+        response_future,
+    )
+    .await;
+    assert!(result.is_err());
+}
+
+#[tokio::test]
+async fn trusted_shuffler_join_test() {
```
mariaschett, line 96: Nice test! Do we need both tests though, the `trusted_shuffler_test` and this one? They seem to be doing the same thing or not?

I was thinking what happens if k==1? Should our Trusted Shuffler reject this?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let background_result = background_result.await;
+    assert!(background_result.is_ok());
+    let background_response = background_result.unwrap();
+    assert!(background_response.is_ok());
+    assert_eq!(expected_background_response, background_response.unwrap());
+}
+
+#[tokio::test]
+async fn trusted_shuffler_waiting_test() {
```
mariaschett, line 77: What am I missing... we are not doing anything with timing yet, are we? What are we testing here?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,123 @@
...
+
+// Generates a request and a corresponding response from a string.
+fn generate_request_response(data: &str) -> (Vec<u8>, Vec<u8>) {
+    let request = format!("Request: {}", data).to_string().as_bytes().to_vec();
+    let response = generate_response(request.clone());
+    (request, response)
+}
+
+#[tokio::test]
+async fn trusted_shuffler_test() {
```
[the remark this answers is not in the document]
mariaschett: Sorry, I wasn't being precise! I was suggesting to rename the test to `trusted_shuffler_test_received_two_responses` or something similar.

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: ah, I didn't realize! :) Can we rename the test or comment that?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Ah, but 1-anonymity does not really make sense, does it? I think we should require that the anonymity value is > 1. Or even better: should we hardcode the anonymity_value in the TrustedShuffler in general?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Ah, I see now! Thanks! 

Does it make sense to call it `generate_test_request_response`? 

And maybe rename `response` to `expected_response`. We are creating the request/expected_response pair here, no?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+
+// Generates a request and a corresponding response from a string.
+fn generate_request_response(data: &str) -> (Vec<u8>, Vec<u8>) {
+    let request = format!("Request: {}", data).to_string().as_bytes().to_vec();
+    let response = generate_response(request.clone());
+    (request, response)
+}
+
+#[tokio::test]
+async fn trusted_shuffler_k2_test() {
```
mariaschett, line 54: Ah, I am not sure people will know `k`. How about `anonymity_value_2_test`?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+    let trusted_shuffler = Arc::new(TrustedShuffler::new(
+        anonymity_value,
+        Arc::new(TestRequestHandler {}),
+    ));
+
+    let (request, expected_response) = generate_request_response("Test");
+    let (background_request, expected_background_response) =
+        generate_request_response("Background test");
+
+    let trusted_shuffler_clone = trusted_shuffler.clone();
```
mariaschett, line 65: Just for my understanding: this would be the `background_trusted_shuffler`? I.e., the trusted shuffler for the background task?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    for (expected_response, result) in expected_responses.iter().zip(results.iter()) {
+        assert!(result.is_ok());
+        let response = result.as_ref().unwrap();
+        assert!(response.is_ok());
+        assert_eq!(expected_response.clone(), *response.as_ref().unwrap());
+    }
+}
+
+// Test that the Trusted Shuffler with k=3 waits for the 3rd client and doesn't
```
mariaschett, line 112: k -> `anonymity_value`

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+    sync::{Arc, Mutex},
+};
+use tokio::sync::oneshot;
+
+type Request = Vec<u8>;
+type Response = Vec<u8>;
+
+struct Message {
+    // Determines the original order in which messages arrived.
+    number: usize,
```
mariaschett, line 38: how about `arrived_as` or `arrival_position`? Maybe we could document what we use this for? Also, is this the order of when the `data` for this message arrived?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+    async fn handle(&self, request: Vec<u8>) -> Vec<u8>;
+}
+
+// Trusted Shuffler implementation.
+pub struct TrustedShuffler {
+    // Value k that represents k-anonymity.
+    anonymity_value: usize,
+
+    // Current batch of requests to be shuffled.
+    // Mutex is used because messages are collected from different async tasks.
```
mariaschett, line 56: clarification question: different asyc tasks == different clients?

mariaschett (approved):
It's getting more and more beautiful!

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+                });
+            }
+        }
+
+        response_receiver.await.context("Couldn't receive response")
+    }
+
+    // Lexicographically sorts requests and sends them to the backend using the
+    // [`TrustedShuffler::request_sender`].
+    async fn shuffle_requests(
```
tiziano88, line 112: It still seems odd to me that this method is not defined on self, and it expects to receive the request_handler (which is already on self) as a separate paramter. Could you add a comment why this choice was made? (also below)

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+        mut requests: Vec<Message>,
+        request_handler: Arc<dyn RequestHandler>,
+    ) -> anyhow::Result<()> {
+        requests.sort_by_key(|v| v.data.clone());
```
tiziano88, line 116: Turns out that to avoid the cloning, the previous version is necessary: https://stackoverflow.com/a/47126516 . I think we should do that, since in general the requests may be large, and we don't want to clone them unnecessarily.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+            async move {
+                let response = request_handler_clone.handle(request.data).await;
+                Message {
+                    number: request.number,
+                    data: response,
+                    response_sender: request.response_sender,
+                }
+            }
+        });
+        let responses = join_all(response_futures).await;
```
tiziano88, line 129: As discussed, could you use https://docs.rs/futures/latest/futures/stream/struct.FuturesOrdered.html to make it explicit that the order is maintained?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+
+        Self::shuffle_responses(responses)
+    }
+
+    // Restores the original order in which messages arrived and sends them back to the client async
+    // tasks using corresponding channels defined in the [`Message::response_sender`].
+    fn shuffle_responses(mut responses: Vec<Message>) -> anyhow::Result<()> {
+        responses.sort_by(|first, second| first.number.cmp(&second.number));
```
tiziano88, line 137: Is there any reason why this uses `sort_by` instead?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,140 @@
...
+            .into_iter()
+            .map(
+                |Message {
+                     number: _,
+                     data,
+                     response_sender,
+                 }| {
+                    response_sender
+                        .send(data)
+                        .map_err(|_| anyhow!("Couldn't send response"))
```
tiziano88, line 149: Why do we just throw away the error here? Doesn't that make it hard to debug?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
tiziano88: +1 it took me a while to understand why a request is being generated twice.

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+        Arc::new(TestRequestHandler {}),
+    ));
+
+    let (request, expected_response) = generate_request_response("Test");
+    let (background_request, expected_background_response) =
+        generate_request_response("Background test");
+
+    let trusted_shuffler_clone = trusted_shuffler.clone();
+    let background_result =
+        tokio::spawn(async move { trusted_shuffler_clone.invoke(background_request).await });
```
tiziano88, line 67: Is it really necessary to put the trusted shuffler in an Arc, then use it in an async closure, and move the Arc into that? I would have thought that it's enough to just get the future out of it and spawn the future itself only.

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let (requests, expected_responses): (Vec<Vec<u8>>, Vec<Vec<u8>>) = (0..anonymity_value)
+        .collect::<Vec<_>>()
+        .iter()
+        .map(|k| generate_request_response(&format!("Test {}", k)))
+        .unzip();
+
+    let mut result_futures = vec![];
+    for request in requests.iter() {
+        let trusted_shuffler_clone = trusted_shuffler.clone();
```
tiziano88, line 96: Again I'm not sure why all the Arcs and the cloning are necessary -- perhaps they are, but could you confirm that it's really the case?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+            tokio::spawn(async move { trusted_shuffler_clone.invoke(request_clone).await });
+        result_futures.push(result_future);
+    }
+    let results = join_all(result_futures).await;
+
+    for (expected_response, result) in expected_responses.iter().zip(results.iter()) {
+        assert!(result.is_ok());
+        let response = result.as_ref().unwrap();
+        assert!(response.is_ok());
+        assert_eq!(expected_response.clone(), *response.as_ref().unwrap());
```
tiziano88, line 108: I don't think cloning should be necessary here, is it?

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+// Non-async version of the `handle_request` function used to create expected responses.
+fn generate_response(request: Vec<u8>) -> Vec<u8> {
+    let parsed_request = String::from_utf8(request).expect("Couldn't parse request");
+    format!("Response for: {}", parsed_request)
+        .to_string()
+        .as_bytes()
+        .to_vec()
```
tiziano88, line 43:
```suggestion
    format!("Response for: {}", parsed_request).into_bytes()
```

to avoid unnecessary clones (and below)

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
tiziano88: Thanks, perhaps something like this may clarify

```suggestion
                        .map_err(|_data| anyhow!("Couldn't send response"))
```

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
tiziano88: Thanks for confirming!

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -16,4 +16,147 @@
...
+                        data: response,
+                        response_sender: request.response_sender,
+                    }
+                }
+            })
+            .collect();
+        let mut responses = vec![];
+        while let Some(response) = response_futures.next().await {
+            responses.push(response);
+        }
```
tiziano88, line 136: I think you can use https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.collect here?

tiziano88 (approved):

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
ipetr0v: Done

PR #2685, opened 2022-04-06
Upgrade rustls and tokio-rustls

conradgrobler:
Some code changes are required to bump the versions of `rustls` and `tokio-rustls` that was attempted by dependabot in #2675 and #2674.


tiziano88 (approved):
Great, thanks!

PR #2686, opened 2022-04-06
Include cache version in github action

### LINKED ISSUE #2683: `collect2: error: ld returned 1 exit status` error in CI
See https://github.com/project-oak/oak/runs/5836846751

Possibly caused by 02728149e8958100fcd1da8c59263952b9750664

tiziano88:
Fixes #2683


jul-sh (approved):

jul-sh:
Thinking this might actually a good opportunity to remove that `FORCE_CI` file 😅. Else we'll have two mechanisms for forcing a CI rerun, with or without clearing caches.

PR #2689, opened 2022-04-06
Remove CODEOWNERS file

tiziano88:
It was added in 36a23872bac39e99e56d78263b879163efc2260d but I was
hoping it would help with auto-assigning reviewers in a round robin
fashion. Instead it seems to just add the team as an additional
reviewer, which does not get converted into an auto assignment.


rbehjati (approved):

tiziano88:
Putting this on hold while we figure out exactly how auto assignment works.

Issue #2654, opened 2022-04-01
Include the UEFI app in the workspace


andrisaar:
Right now the UEFI app under `experimental/uefi` is excluded from workspace in the root `Cargo.toml` file, as we need a custom `.cargo/config.toml` file for the UEFI app.

Ideally, we'd want to get rid of the `.cargo/config.toml` file there, and (a) fold things into the root `.cargo/config.toml` file and/or (b) move things to the `Cargo.toml` file of the UEFI app.

Unfortunately we can't do that yet.
Looking at what we have in the `.cargo/config.toml` file and what we can do about those fields:

- `target` could move to `Cargo.toml` using https://doc.rust-lang.org/cargo/reference/unstable.html#per-package-target
- the `runner` stanzas could likely be moved to the root `.cargo/config.toml` with relative ease
- `build-std` and `build-std-features` needs https://github.com/rust-lang/cargo/issues/9451 before we can do anything about them.

jul-sh:
For `build-std` and `build-std-features`, could we pass these values as command line flags instead? 

Eg `-Zbuild-std-features=compiler-builtins-mem` `-Zbuild-std=core,alloc`. It seems like that should work, though it may conflict with the existing workspace config file.

andrisaar:
That won't help when you're running, say, `cargo doc` in the root directory as `xtask` does. I assume we don't want to rebuild `std` for *every* crate, just the limited UEFI one(s).

jul-sh:
Yeah it would require some (potentially unwieldy) changes to xtask. The upside of using xtask though is that that's feasible. Eg it'd work quite nicely for running examples etc

mariaschett:
Maybe relevant: I tried using the `per-package-target` before (#2468 ), but didn't get it to work.

andrisaar:
> Maybe relevant: I tried using the `per-package-target` before (#2468 ), but didn't get it to work.

It looks like your main issue was that the tests got also compiled for wasm. That's the case here as well (tests get compiled for uefi), but as we specify a loder (`qemu`) `cargo test` will be able to execute the tests just fine.

jul-sh:
tested my earlier idea of adapting cargo (with custom runners) a bit, but don't think it's worth it. Ref: #2733. 
If the workspaces PR doesn't progress to unblock this, we could alternatively consider using cargo vendor to share deps, ala what Mozilla did: https://github.com/rust-lang/cargo/issues/5332#issuecomment-380088273

[...]

PR #2693, opened 2022-04-06
Build remote attestation for UEFI

jul-sh:
- Applies the UEFI compatibility patch for ring from https://github.com/briansmith/ring/pull/1406
- Adds needed build tooling to build ring for a UEFI target
- Includes it as a dep in the UEFI app

### DIFF
--- a/experimental/uefi/app/Cargo.toml
+++ b/experimental/uefi/app/Cargo.toml
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 uefi = { version = "*", features = ["exts"] }
 uefi-services = "*"
 log = { version = "*" }
+oak_remote_attestation = { path = "../../../remote_attestation/rust" }
 
 [dev-dependencies]
 uefi-services = { version = "*", features = ["qemu"] }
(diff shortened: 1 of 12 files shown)


experimental/uefi/app/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 uefi = { version = "*", features = ["exts"] }
 uefi-services = "*"
 log = { version = "*" }
+oak_remote_attestation = { path = "../../../remote_attestation/rust" }
```
conradgrobler, line 12: I am surprised this does not cause cargo udeps to complain, seeing that the code does not currently use anything from the remote attestation crate. Do we ignore the UEFI app when running cargo udeps?

conradgrobler (approved):

experimental/uefi/app/Cargo.toml
jul-sh: Good catch, forgot about us using `cargo udeps`. The reason it doesn't complain is that we run it on the workspace, and the UEFI app is excluded from the workspace bc of https://github.com/project-oak/oak/issues/2654.

Should we keep this as is? This crate should start actually using this dep fairly soon anyways.

conradgrobler: I think it is fine for now, but we should probably update xtask at some point to run cargo udeps for this as well.

PR #2700, opened 2022-04-07
Give request as argument to OakAbiNativeExtension

mariaschett:
First step towards: #2700 .


tiziano88 (approved):
Personally I would have moved both request and response out at the same time, anyways this looks fine.

ipetr0v (approved):

PR #2702, opened 2022-04-08
Return Option<Vec<u8>> in invoke in OakApiNativeExtension

### LINKED ISSUE #2699: Return `Option<Vec<u8>>` in invoke in OakApiNativeExtension

mariaschett:
Fixes #2699.

### DIFF
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -116,11 +116,11 @@ impl FixedSizeBodyPadder for Response {
 
 /// Trait for implementing extensions, to implement new native functionality.
 pub trait OakApiNativeExtension {
-    /// Invokes the extension with the given request and optionally returns a result. If no result
-    /// is expected, it returns None.  An error within the extension is reflected in the
+    /// Invokes the extension with the given request and returns a result. If no result
+    /// is expected, the result is empty.  An error within the extension is reflected in the
     /// `OakStatus`.
     /// TODO(#2701): Stop returning OakStatus for errors internal to extensions.
-    fn invoke(&mut self, request: Vec<u8>) -> Result<Option<Vec<u8>>, OakStatus>;
+    fn invoke(&mut self, request: Vec<u8>) -> Result<Vec<u8>, OakStatus>;
 
     /// Metadata about this Extension, including the exported host function name, the function's
     /// signature, and the corresponding ExtensionHandle.
@@ -375,13 +375,8 @@ impl WasmState {
                 let response_ptr_ptr: AbiPointer = args.nth_checked(3)?;
                 let response_len_ptr: AbiPointer = args.nth_checked(4)?;
                 match response {
-                    Ok(Some(response)) => {
+                    Ok(response) => {
                         self.write_extension_result(response, response_ptr_ptr, response_len_ptr)
-                    }
-                    Ok(None) =>
-                    // No response was expected. We write the empty response.
-                    {
-                        self.write_extension_result(vec![], response_ptr_ptr, response_len_ptr)
                     }
                     Err(err) => Err(err),
                 }
@@ -462,23 +457,21 @@ impl wasmi::Externals for WasmState {
                     Ok(request) => {
                         let response = extension.invoke(request);
                         match response {
-                            Ok(Some(response)) => {
-                                // Careful: We assume every ABI call which returns a response
-                                // provides these arguments (which
-                                // is true). We will remove this, when we call every
-                                // extension through `invoke`.
-                                let response_ptr_ptr: AbiPointer = args.nth_checked(2)?;
-                                let response_len_ptr: AbiPointer = args.nth_checked(3)?;
-                                self.write_extension_result(
-                                    response,
-                                    response_ptr_ptr,
-                                    response_len_ptr,
-                                )
-                            }
-                            Ok(None) =>
-                            // No response was expected.
-                            {
-                                Ok(())
+                            Ok(response) => {
+                                if !response.is_empty() {
+                                    // Careful: We assume every ABI call which returns a non-emtpy
+                                    // response provides these arguments
+                                    // (which is true).
+                                    let response_ptr_ptr: AbiPointer = args.nth_checked(2)?;
+                                    let response_len_ptr: AbiPointer = args.nth_checked(3)?;
+                                    self.write_extension_result(
+                                        response,
+                                        response_ptr_ptr,
+                                        response_len_ptr,
+                                    )
+                                } else {
+                                    Ok(())
+                                }
                             }
                             Err(err) => Err(err),
                         }
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -441,10 +441,13 @@ impl wasmi::Externals for WasmState {
             ),
             INVOKE => self.invoke_extension_with_handle(args),
 
+            // TODO(#2710), TODO(#2711), TODO(#2712): Remove the following code, once all extensions
+            // are called via handles.
             _ => {
                 // Careful: We assume that here for the ABI call the first two arguments are the
                 // request (which is true). We will remove this, when we call every
-                // extension through `invoke`.
+                // extension through `invoke`. Here, args are off-by-one compared to
+                // `invoke_extension_with_handle` because there `args[0]` is the handle.
                 let request_ptr: AbiPointer = args.nth_checked(0)?;
                 let request_len: AbiPointerOffset = args.nth_checked(1)?;
 
--- a/oak_functions/loader/src/server.rs
+++ b/oak_functions/loader/src/server.rs
@@ -336,13 +336,33 @@ impl WasmState {
         Ok(())
     }
 
+    // Finds the extension parsing the given i32 handle. Returns appropriate OakStatus, if
+    // the handle is invalid or no extension with this handle can be found in the extension of
+    // WasmState.
+    fn find_extension(&mut self, handle: i32) -> Result<&mut BoxedExtension, OakStatus> {
+        let handle: ExtensionHandle = ExtensionHandle::from_i32(handle).ok_or({
+            self.log_error(&format!("Fail to parse handle {:?}.", handle));
+            OakStatus::ErrInvalidArgs
+        })?;
+
+        // TODO(#2707): Refactor to return OakStatus::ErrInvalidHandle. Add logging.
+        self.extensions_indices
+            .iter_mut()
+            .find_map(|(_, extension)| {
+                if extension.get_handle() == handle {
+                    Some(extension)
+                } else {
+                    None
+                }
+            })
+            .ok_or(OakStatus::ErrInvalidArgs)
+    }
+
     pub fn invoke_extension_with_handle(
         &mut self,
         args: wasmi::RuntimeArgs,
     ) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
         let handle = args.nth_checked(0)?;
-        let handle: ExtensionHandle =
-            ExtensionHandle::from_i32(handle).expect("Fail to parse handle.");
 
         let request_ptr: AbiPointer = args.nth_checked(1)?;
         let request_len: AbiPointerOffset = args.nth_checked(2)?;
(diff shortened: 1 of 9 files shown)


tiziano88:
I think for the assignments to work correctly, you need to only add core at first, and only after a reviewer gets auto-assigned, you can add another one. Otherwise GitHub seems to keep core around (I think this is a GitHub bug BTW)

mariaschett:
This was actually a combination with a bug from my side, I think.

I wanted to add it as a draft, then it assigned correctly one reviewer (you).

When I converted it as draft, IIRC core got re-added.

oak_functions/loader/src/server.rs
```diff
@@ -116,14 +116,11 @@ impl FixedSizeBodyPadder for Response {
...
-        &mut self,
-        wasm_state: &mut WasmState,
-        args: wasmi::RuntimeArgs,
-        request: Vec<u8>,
-    ) -> Result<Result<(), OakStatus>, wasmi::Trap>;
+    /// Invokes the extension with the given request and optionally returns a result. If no result
+    /// is expected, it returns None.  An error within the extension is reflected in the
+    /// `OakStatus`.
+    /// TODO(#2701): Stop returning OakStatus for errors internal to extensions.
+    fn invoke(&mut self, request: Vec<u8>) -> Result<Option<Vec<u8>>, OakStatus>;
```
tiziano88, line 123: I'm not sure I understand the motivation of the `Result<Option<...>>` nesting. e.g. here it seems it should just return an empty vector, why add another layer of optionality?
I mean, by symmetry then you should make the request also an Option in case there is no request to be passed in, but that's already covered by passing an empty vec as input.

oak_functions/loader/src/server.rs
```diff
@@ -384,14 +356,38 @@ impl WasmState {
...
+            .extensions_indices
+            .iter_mut()
+            .find_map(|(_, extension)| {
+                if extension.get_handle() == handle {
+                    Some(extension)
+                } else {
+                    None
+                }
+            })
+            .expect("Fail to find extension with given handle.");
```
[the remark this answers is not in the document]
tiziano88: I think this is covered by the TODO above, but I also agree this is super important, otherwise a Wasm module can invoke a non-existing extension in order to cause a panic in the entire Wasm execution, which may be externally observable.

oak_functions/loader/src/server.rs
```diff
@@ -384,14 +356,38 @@ impl WasmState {
...
+                let response_ptr_ptr: AbiPointer = args.nth_checked(3)?;
+                let response_len_ptr: AbiPointer = args.nth_checked(4)?;
+                match response {
+                    Ok(Some(response)) => {
+                        self.write_extension_result(response, response_ptr_ptr, response_len_ptr)
+                    }
+                    Ok(None) =>
+                    // No response was expected. We write the empty response.
+                    {
+                        self.write_extension_result(vec![], response_ptr_ptr, response_len_ptr)
```
tiziano88, line 384: This also confirms my point above: why return None if  it's just treated as an empty Vec anyways?

oak_functions/loader/src/server.rs
```diff
@@ -461,19 +447,44 @@ impl wasmi::Externals for WasmState {
...
+                            "Unable to read input from guest memory: {:?}",
                             err
                         ));
                         OakStatus::ErrInvalidArgs
                     });
 
+                let extension = &mut self
+                    .extensions_indices
+                    .get_mut(&index)
+                    .unwrap_or_else(|| panic!("Unimplemented function at {}", index));
```
tiziano88, line 459: Could you confirm it's fine to panic here, i.e. it's not possible for untrusted Wasm code to cause this panic? I would still always worry about this every time I read this code, so perhaps worth leaving a comment to explain this, or, better, avoid the panic altogether.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
mariaschett: Good point, both! I did fix it immediately, but will iterate on the TODO.

oak_functions/loader/src/server.rs
```diff
@@ -384,14 +356,38 @@ impl WasmState {
...
+                    None
+                }
+            })
+            .expect("Fail to find extension with given handle.");
+
         let result = match request {
-            Ok(request) => extension.invoke(self, args, request)?,
+            Ok(request) => {
+                let response = extension.invoke(request);
+                let response_ptr_ptr: AbiPointer = args.nth_checked(3)?;
```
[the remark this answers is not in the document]
mariaschett: Yes, now it is.

oak_functions/loader/src/server.rs

oak_functions/loader/src/server.rs
```diff
@@ -438,19 +434,9 @@ impl wasmi::Externals for WasmState {
             WRITE_LOG_MESSAGE => from_oak_status_result(
                 self.write_log_message(args.nth_checked(0)?, args.nth_checked(1)?),
             ),
-            INVOKE => self.invoke_extension_with_handle(args.nth_checked(0)?, args),
+            INVOKE => self.invoke_extension_with_handle(args),
 
             _ => {
```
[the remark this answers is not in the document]
mariaschett: Yes, this is the part we want to eventually remove. I will gradually refactor every extension to be called via `invoke`, i.e., be handled in the `INVOKE` case in line 441.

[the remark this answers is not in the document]
mariaschett: Yes, this is the part we want to eventually remove. I will gradually refactor every extension to be called via `invoke`, i.e., be handled in the `INVOKE` case in line 441.

oak_functions/loader/src/server.rs
```diff
@@ -461,19 +447,44 @@ impl wasmi::Externals for WasmState {
...
+                    .unwrap_or_else(|| panic!("Unimplemented function at {}", index));
+
                 let result = match request {
-                    Ok(request) => extension.invoke(self, args, request)?,
+                    Ok(request) => {
+                        let response = extension.invoke(request);
+                        match response {
+                            Ok(Some(response)) => {
+                                // Careful: We assume every ABI call which returns a response
+                                // provides these arguments (which
```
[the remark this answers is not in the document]

oak_functions/loader/src/server.rs
```diff
@@ -461,19 +447,44 @@ impl wasmi::Externals for WasmState {
...
+                                // is true). We will remove this, when we call every
+                                // extension through `invoke`.
+                                let response_ptr_ptr: AbiPointer = args.nth_checked(2)?;
```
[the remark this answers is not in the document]
mariaschett: You mean reading `args.nth_checked(2)` for the response? 

Here we cover the ABI calls `get_storage_item`, `report_metric` and `tf_infer`. All of them have two arguments for the request (i.e., `args.nth_checked(0)` and `args.nth_checked(`1`)). 

If the ABI call expects a response, it has two more argument (i.e., `args.nth_checked(2)` and `args.nth_checked(`3`)---which is every call except `report_metric`.

The off-by-one weirdness comes from `invoke_extension_with_handle` having as first argument the handle.

mariaschett (commented):
Thank you both!

oak_functions/loader/src/server.rs
mariaschett: This code will disappear, once we call every extension through invoke. Are you against keeping it as it is? I don't see immediately how this could be triggered from the Wasm code.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
rbehjati: Perhaps add a comment about it.

rbehjati (approved):
Nice. Thanks :)

oak_functions/loader/src/server.rs
tiziano88: SG

oak_functions/loader/src/server.rs
```diff
@@ -461,19 +452,42 @@ impl wasmi::Externals for WasmState {
...
+                                    // (which is true).
+                                    let response_ptr_ptr: AbiPointer = args.nth_checked(2)?;
+                                    let response_len_ptr: AbiPointer = args.nth_checked(3)?;
+                                    self.write_extension_result(
+                                        response,
+                                        response_ptr_ptr,
+                                        response_len_ptr,
+                                    )
+                                } else {
+                                    Ok(())
```
tiziano88, line 483: It looks like here it is not writing anything to the Wasm memory, which seems wrong? It should at least write a zero in the response_len_ptr, right? Otherwise how would the caller know? It should just write an empty vector -- and if there is special logic for empty vectors, that should be handled inside `write_extension_result`, not here.

mariaschett: This code is actually called through the soon-to-be-deprecated ABI functions `get_storage_item`, `tf_infer`, and `report_metric`. 

The latter does not write anything to memory (has no result), hence in does not currently have `args.nth_checked(2-3)`.

We write an empty vector for the [testing extension here](https://github.com/project-oak/oak/blob/9c639b300e7e2a6d2afe63c16bc89de6e2b9ac25/oak_functions/loader/src/testing.rs#L44), but the testing extension is already called via `invoke`, i.e., [`invoke_extension_with_handle`](https://github.com/project-oak/oak/blob/9c639b300e7e2a6d2afe63c16bc89de6e2b9ac25/oak_functions/loader/src/server.rs#L388).

oak_functions/loader/src/server.rs
```diff
@@ -384,14 +376,23 @@ impl WasmState {
                 OakStatus::ErrInvalidArgs
             });
 
-        let result = match request {
-            Ok(request) => extension.invoke(self, args, request)?,
-            Err(err) => Err(err),
+        let extension = self.find_extension(handle);
+
+        let result = match (extension, request) {
```
tiziano88, line 381: If request is an error, why even look up the extension and then check two errors at once? It seems it could be checked in the right order.

oak_functions/loader/src/server.rs
tiziano88: Uhmm ok I'm still a bit confused by what code is temporary and what code is here to stay, but I trust it will become clearer down the line.

tiziano88 (approved):

oak_functions/loader/src/server.rs
```diff
@@ -461,19 +452,42 @@ impl wasmi::Externals for WasmState {
...
+                                    self.write_extension_result(
+                                        response,
+                                        response_ptr_ptr,
+                                        response_len_ptr,
+                                    )
+                                } else {
+                                    Ok(())
+                                }
+                            }
+                            Err(err) => Err(err),
```
tiziano88, line 486: Whenever you are passing erros unmodified, it probably means you could have shortcut the error via the `?` operator first, or at least you could use `Result::map` to only deal with the ok case.

tiziano88 (approved):

oak_functions/loader/src/server.rs
rbehjati: I think this will be easier if everything before `from_oak_status_result(result)` can be refactored into a function returning `Result<???, OakStatus >` to be able to use the `?` operator.

oak_functions/loader/src/server.rs
mariaschett: This is a bit tricky with the [`from_oak_status_result` before return](https://github.com/project-oak/oak/blob/ac0c134a93153e428fb6b4129c766cf28b1b9c02/oak_functions/loader/src/server.rs#L477). There is some magic happening here, which I hope to simplify later on.

oak_functions/loader/src/server.rs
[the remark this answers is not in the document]
mariaschett: Done.

oak_functions/loader/src/server.rs
```diff
@@ -382,17 +377,10 @@ impl WasmState {
...
-            Ok(request) => extension.invoke(self, args, request)?,
-            Err(err) => Err(err),
-        };
-
-        // We put the extension indices back.
-        self.extensions_indices = Some(extensions_indices);
-
-        from_oak_status_result(result)
+            })?;
+        let extension = self.find_extension(handle)?;
```
mariaschett, line 381: Simplified a lot, thank you for your comments!

PR #2704, opened 2022-04-08
Add CBOR framing to the serial channel

andrisaar:
In preparation for proper gRPC request/response flow this change adds some basic framing to the serial data channel: instead of just sending bytes back and forth, we use CBOR as a framing mechanism.

For now, we're still sending strings back and forth, though.

Also, while I was in there, I moved the serial-specific stuff in the UEFI app into its own file.

### DIFF
--- a/experimental/uefi/app/src/main.rs
+++ b/experimental/uefi/app/src/main.rs
@@ -26,15 +26,12 @@
 
 #[macro_use]
 extern crate log;
+extern crate alloc;
 
-use uefi::{
-    prelude::*,
-    proto::console::serial::Serial,
-    table::{
-        boot::{OpenProtocolAttributes, OpenProtocolParams},
-        runtime::ResetType,
-    },
-};
+use ciborium::{de, ser};
+use uefi::{prelude::*, table::runtime::ResetType};
+
+mod serial;
 
 // The main entry point of the UEFI application.
 //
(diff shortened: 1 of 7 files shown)


experimental/uefi/loader/Cargo.toml
```diff
@@ -21,4 +22,6 @@ tokio = { version = "*", features = [
   "process",
   "signal",
 ] }
-tokio-util = { version = "*", features = ["codec"] }
+tokio-serde-cbor = "*"
```
conradgrobler, line 25: This is fine for now, but there are potential incompatibilities between `ciborium` and `serde-cbor` if we start using more complex types. See https://docs.rs/ciborium/latest/ciborium/#compatibility-with-other-implementations

experimental/uefi/app/src/main.rs
```diff
@@ -26,15 +26,12 @@
...
 
-use uefi::{
-    prelude::*,
-    proto::console::serial::Serial,
-    table::{
-        boot::{OpenProtocolAttributes, OpenProtocolParams},
-        runtime::ResetType,
-    },
-};
+use ciborium::{de, ser};
```
conradgrobler, line 31: While this works for now, I don't think CBOR is a suitable framing mechansim for the long term. If we start creating more complex structs to represents frames with headers this will start adding a lot of overhead. AFAICT structs are represented as maps, with each field name as a string key. This will waste bandwith and be slower to serialise and deserialise. (see https://docs.rs/ciborium/latest/ciborium/#packed-encoding)

I think we should implemment the framing manually, as I believe this communication channel is likely to be a performance bottleneck.

Perhaps create and issue and a TODO somewhere to track this.

experimental/uefi/app/src/serial.rs
```diff
@@ -0,0 +1,99 @@
...
+        )?;
+        // Dereference the raw pointer (*mut Serial) we get to the serial interface.
+        // This is safe as according to the UEFI spec for the OpenProtocol call to succeed the
+        // interface must not be null (see Section 7.3 in the UEFI Specification, Version 2.9).
+        let serial = unsafe { &mut *serial.interface.get() };
+
+        // Increase the FIFO size from the default 1 byte to something bigger. Experiments with
+        // qemu suggest 16 is the best we can do.
+        let mut io_mode = *serial.io_mode();
+        io_mode.receive_fifo_depth = 16;
```
conradgrobler, line 52: nit: perhaps make this a constant with a more descriptive name.

experimental/uefi/loader/src/main.rs
```diff
@@ -79,16 +80,29 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     });
 
+    // Unfortunately OVMF writes some garbage (clear screen etc?) + our Hello World to the
+    // other serial port, so let's skip some bytes before we set up the CBOR codec.
+    let mut junk = [0; 71];
```
conradgrobler, line 85: Is there some way that we can tell the junk is finished? Hardcoding a fixed number is brittle. I think it is fine for now, as the serial port is only a temporary mechanism, but perhaps make the 71 a constant with a more descriptive name.

experimental/uefi/app/src/main.rs
```diff
@@ -70,53 +67,31 @@ fn main(handle: Handle, system_table: &mut SystemTable<Boot>) -> Status {
     serial_echo(handle, system_table.boot_services()).unwrap();
 }
 
-fn echo_loop(serial: &mut Serial) -> Result<!, uefi::Error<()>> {
-    let mut buf: [u8; 1024] = [0; 1024];
-
+fn serial_echo(handle: Handle, bt: &BootServices) -> Result<!, uefi::Error<()>> {
+    let mut serial = serial::Serial::get(handle, bt, 1)?;
```
conradgrobler, line 71: nit: perhaps make this a constant with a more descriptive name.

conradgrobler (approved):

experimental/uefi/loader/Cargo.toml
andrisaar: Fair enough. In addition, `serde-cbor` is effectively unmaintained: https://crates.io/crates/serde_cbor

As we'll likely replace CBOR anyway I'll leave it as is for now.

experimental/uefi/loader/src/main.rs
andrisaar: Yeah, this is a massive hack. I'll file a bug and add a proper TODO about this.

A much cleaner way would be to figure out how to configure OVMF not to output that junk to the second serial port (ideally it'd be only the Hello World message that gets sent to the first serial port as well), but I haven't yet found if it's possible to do this.

Interestingly, log messages get only sent to the first serial port, not the second. Maybe it's something that stdout gets sent to all serial ports by default, and stderr gets sent to the first, or something.

experimental/uefi/app/src/main.rs
tiziano88: +1 I think at this level we should implement a simple manual serialization, e.g. something like a standard length delimited framing.

andrisaar: I'll leave it as-is for now but get back to it later. On the loader side we can use https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/index.html; we'd likely need to implement something similar on the UEFI app side.

PR #2706, opened 2022-04-09
Add zxh404.vscode-proto3 to the dev container.

andrisaar:
Just for syntax highlighting of `.proto` files, and said plugin seems to be the most popular one.


jul-sh (approved):
thanks, lgtm

PR #2716, opened 2022-04-13
Add OakStatus for `InvalidExtensionHandle` and `ExtensionNotAvailable`

### LINKED ISSUE #2707: Add `OakStatus::ErrInvalidHandle`

mariaschett:
Fixes #2707 .


oak_functions/proto/abi.proto
```diff
@@ -46,6 +46,10 @@ enum OakStatus {
   ERR_STORAGE_ITEM_NOT_FOUND = 4;
   // Error when running the TensorFlow model, due to bad input tensor.
   ERR_BAD_TENSOR_FLOW_MODEL_INPUT = 5;
+  // Error when extension handle is invalid.
+  ERR_INVALID_EXTENSION_HANDLE = 6;
+  // Error when extension is not available.
+  ERR_EXTENSION_NOT_AVAILABLE = 7;
```
tiziano88, line 52: I'm not sure I get the difference between an invalid extension handle vs an extension not available, could you clarify?

oak_functions/proto/abi.proto
```diff
@@ -46,6 +46,10 @@ enum OakStatus {
   ERR_STORAGE_ITEM_NOT_FOUND = 4;
   // Error when running the TensorFlow model, due to bad input tensor.
   ERR_BAD_TENSOR_FLOW_MODEL_INPUT = 5;
+  // Error when extension handle is invalid.
+  ERR_INVALID_EXTENSION_HANDLE = 6;
```
tiziano88, line 50: At this level I don't think there should be a mention of extensions, this is just a transport for performing invocations over some handle. The fact that there is an extension or a service or another Wasm module or something else should not appear at this level. Also we may want to use this for other communications, e.g. between trusted and untrusted runtime, or between Oak and other systems, etc. I suggest just renaming to ERR_INVALID_HANDLE.

oak_functions/loader/src/server.rs
```diff
@@ -337,28 +344,28 @@ impl WasmState {
...
-            self.log_error(&format!("Fail to parse handle {:?}.", handle));
-            OakStatus::ErrInvalidArgs
-        })?;
-
-        // TODO(#2707): Refactor to return OakStatus::ErrInvalidHandle. Add logging.
-        self.extensions_indices
+        let extension = self
+            .extensions_indices
             .iter_mut()
             .find_map(|(_, extension)| {
```
tiziano88, line 355: What does the `find_map` add here compared to just `find`? It seems to add a layer of Option, which is removed immediately anyways.

mariaschett: `find_map` maps the pair of (_, `extension`) to return just the extension. But, good point I refactored using `values_mut` and inlined.

oak_functions/loader/src/server.rs
```diff
@@ -336,19 +343,8 @@ impl WasmState {
...
-        handle: AbiExtensionHandle,
-    ) -> Result<&mut BoxedExtension, OakStatus> {
-        let handle: ExtensionHandle = ExtensionHandle::from_i32(handle).ok_or({
-            self.log_error(&format!("Fail to parse handle {:?}.", handle));
-            OakStatus::ErrInvalidArgs
-        })?;
-
-        // TODO(#2707): Refactor to return OakStatus::ErrInvalidHandle. Add logging.
+    // Finds the extension for the given handle.
+    fn find_extension(&mut self, handle: ExtensionHandle) -> Option<&mut BoxedExtension> {
```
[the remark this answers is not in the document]
mariaschett: Not sure, but I inlined the function now.

Issue #2717, opened 2022-04-13
Update devcontainer setup to use our docker scripts, instead of replicating their behavior


jul-sh:
The devcontainer presently does not use our docker scripts (scripts/docker_run, and scripts/fix_docker_user_and_run), instead mirroring their behavior in regards to user permissions: https://github.com/project-oak/oak/blob/bd071bad6822ec83d7b27a4abd0c2ee78a4d8915/.devcontainer.json#L23

It would be nice if the docker container would instead use (parts of) these scripts. This would allow it to make use of host caches (it currently does not, the scripts do), and also make any future changes to our config more straightfoward (only the scripts need updating. 

The devcontainer config provides options for running scripts on the host before container creation, and inside the container after creation, hence it seems possible to do this. Ref: https://code.visualstudio.com/docs/remote/devcontainerjson-reference#_lifecycle-scripts

jul-sh:
Looked into this a bit, unfortunately the lifecycle script that runs on the host cannot set env vars that will be forwarded to the host. We need that though, to inform the guest of the hosts GID of the kvm group for the permissions logic we do in our scripts. 

The (messy) workaround is to write those to a local file instead, then from within the guest read and subsequently delete that temp file.

jul-sh:
Closed, see #2995

PR #2718, opened 2022-04-13
Setup docker scripts to allow the image to use KVM

jul-sh:
This only applies when running docker via our scripts, which are not (yet) invoked by the VSCode dev container. Supporting the dev container is likely best done by configuring it to use our docker scripts, added https://github.com/project-oak/oak/issues/2717 to track that.

### DIFF
--- a/scripts/docker_run
+++ b/scripts/docker_run
@@ -15,6 +15,11 @@ else
   readonly HOST_DOCKER_GID="$(getent group docker | cut -d: -f3)"
 fi
 
+# In order to use KVM inside docker we need permissions for the user with the
+# same KVM gid as the host
+readonly HOST_KVM_GID="$(getent group kvm | cut -d: -f3)"
+export HOST_KVM_GID
+
 mkdir -p './bazel-cache'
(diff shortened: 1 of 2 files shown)


scripts/docker_run
```diff
@@ -48,6 +53,9 @@ docker_run_flags=(
   # To do that, we map the socket from the host and add the right group
   '--volume=/var/run/docker.sock:/var/run/docker.sock'
   "--group-add=$HOST_DOCKER_GID"
+  #
+  "--device=/dev/kvm"
+  '--env=HOST_KVM_GID'
```
jul-sh, line 58: TODO (will update PR): only provide these if KVM is available on the host

ipetr0v (approved):

conradgrobler (commented):
Could you provide more info on the docker-kvm submodule? I don't know what it contains, what its purpose is, or where its original location is. I also don't know how to find more info about the submodule using the github review tool.

jul-sh:
@ipetr0v @conradgrobler thx for the review, sorry for the many commits to get the bash implementation right

PR #2719, opened 2022-04-13
Patch how ring gets CPU feature bits when detecting RDRAND support

jul-sh:
**Please review carefully.** 

The original UEFI patch uses OpenSSL to get CPU feature bits, but it does not seem to be working, see the code comment for details. 

One way to patch it would be to use the rust core lib for this. (Another could be to try and debug the OpenSSL solution). 

I've no strong opinion on the best way to approach this, raising the PR to gather input. cc @conradgrobler

### DIFF
--- a/third_party/ring/src/rand.rs
+++ b/third_party/ring/src/rand.rs
@@ -444,15 +444,22 @@ mod uefi {
     #[cfg(any(target_arch = "x86_64"))]
     fn fill_impl(dest: &mut [u8]) -> Result<(), error::Unspecified> {
         fn is_avaiable() -> bool {
-            // TODO(xiaoyuxlu): use cpu::intel::RDRAND.avaiable when cpu.rs updated
-            // https://github.com/briansmith/ring/pull/1406#discussion_r720394928
-            // Current implementation may cause problem on AMD cpu. REF:
-            // https://github.com/nagisa/rust_rdrand/blob/f2fdd528a6103c946a2e9d0961c0592498b36493/src/lib.rs#L161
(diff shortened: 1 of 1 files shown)


third_party/ring/src/rand.rs
```diff
@@ -444,15 +444,22 @@ mod uefi {
...
+            // regardless of the CPU model used. Hence we use the rust core lib
+            // to get the CPU feature bits instead. That implementation is
+            // copied from https://github.com/nagisa/rust_rdrand/blob/f2fdd528a6103c946a2e9d0961c0592498b36493/src/lib.rs#L161.
+            // See the discussion on the UEFI patch's open PR as well. The patch
+            // originally used the rdrand crate to handle feature detection
+            // (among other things) before switching to OPENSSL_ia32cap_P[1] as
+            // a temporary implementation while CPU feature detection in ring is
+            // undergoing wider changes.
+            // Ref: https://github.com/briansmith/ring/pull/1406#discussion_r720394928
+            let cpu_feature_bits = unsafe { core::arch::x86_64::__cpuid(1).ecx };
```
conradgrobler, line 458: Could you add additional comments on why this is safe (e.g. something similar to https://docs.rs/raw-cpuid/latest/src/raw_cpuid/lib.rs.html#91) and explain why leaf number 1 is used for the call?

conradgrobler (approved):
I think this approach is good for the time-being.

third_party/ring/src/rand.rs
jul-sh: done!

PR #2720, opened 2022-04-13
Stop returning OakStatus for `BadTensorFlowModelInput`

mariaschett:


mariaschett (commented):
WDYT?

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -150,10 +152,24 @@ pub fn tf_model_infer(input_vector: &[u8]) -> Result<Inference, OakStatus> {
...
+            let response: TfModelInferResponse =
+                bincode::deserialize(&response).expect("Fail to deserialize tf response.");
+            let result: Result<Vec<u8>, TfModelInferError> = response.result;
+            match result {
+                Ok(inference_bytes) => {
+                    let inference = Inference::decode(&*inference_bytes);
+                    // TODO(mschett): We currently need to return an OakStatus here.
+                    inference.map_err(|_| OakStatus::ErrInvalidArgs)
+                }
+                // TODO(mschett): We currently need to return an OakStatus here.
```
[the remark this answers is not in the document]
mariaschett: Do you mean `TfModelInferResponse` as wrapper type? That we have already.

But do we have any ideas how we pass errors on within the SDK? Atm we just return the `OakStatus`.

Shall we define an error to capture "transport errors" and "extension errors" within the SDK?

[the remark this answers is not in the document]
mariaschett: I think I just pushed what you suggested.

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -150,12 +154,23 @@ pub fn tf_model_infer(input_vector: &[u8]) -> Result<Inference, OakStatus> {
...
+    if let OakStatus::Ok = status {
+        // We read and interpret the response.
+        let response = from_alloc_buffer(inference_ptr, inference_len);
+        let response: TfModelInferResponse =
+            bincode::deserialize(&response).expect("Fail to deserialize tf response.");
+
+        // We decode the inference bytes.
+        let tf_result = response.result.and_then(|inference_bytes| {
+            Inference::decode(&*inference_bytes)
+                .map_err(|_| TfModelInferError::ErrorDecodingInference)
```
mariaschett, line 167: Note, here we have a slight change and interpret a decoding error as an TfError, not an OakStatus. Would you disagree?

oak_functions/loader/src/tf.rs
```diff
@@ -35,18 +38,18 @@ const TF_ABI_FUNCTION_NAME: &str = "tf_model_infer";
...
             ));
-            // TODO(#2701): Remove ErrBadTensorFlowModelInput from OakStatus.
-            OakStatus::ErrBadTensorFlowModelInput
-        })?;
-        Ok(inference.encode_to_vec())
+            TfModelInferError::BadTensorFlowModelInput
+        });
+        let result = inference.map(|inference| inference.encode_to_vec());
+        let response = bincode::serialize(&TfModelInferResponse { result })
+            .expect("Fail to serialize tf response.");
```
tiziano88, line 50: (and elsewhere)

```suggestion
            .expect("Failed to serialize TF response");
```

mariaschett: Done.

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -136,7 +138,9 @@ pub fn write_log_message<T: AsRef<str>>(message: T) -> Result<(), OakStatus> {
 ///
 /// See [`tf_model_infer`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#tf_model_infer).
 #[cfg(feature = "oak-tf")]
-pub fn tf_model_infer(input_vector: &[u8]) -> Result<Inference, OakStatus> {
+pub fn tf_model_infer(
+    input_vector: &[u8],
+) -> Result<Result<Inference, TfModelInferError>, OakStatus> {
```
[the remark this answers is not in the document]
mariaschett: Not sure what the end is exaclty.

The double result is currently motivated by: 
- the outer result is the result of the transport (hence failing to an OakStatus), 
- the inner result is the result of the extension.

I guess when we have a different solution for the transport this could change. Do you think I am missing something or we should have  a different solution here?

oak_functions/sdk/oak_functions/src/lib.rs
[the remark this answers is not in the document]
mariaschett: Done, but this is the first time something is logged in the low level SDK using `log!`. I think it is fine, but please let me know if I am missing something.

oak_functions/abi/src/lib.rs
```diff
@@ -57,6 +57,21 @@ pub struct ReportMetricRequest {
...
+    BadTensorFlowModelInput,
+    // Error when decoding the inference bytes.
+    ErrorDecodingInference,
+}
+
+#[derive(Serialize, Deserialize)]
+#[cfg(feature = "oak-tf")]
+pub struct TfModelInferResponse {
+    pub result: Result<Vec<u8>, TfModelInferError>,
+}
```
tiziano88, line 73: Not sure why this is a struct and not just a type alias, or if we really need a newtype then a single-element tuple.

oak_functions/sdk/oak_functions/tests/tf_module/src/lib.rs
```diff
@@ -17,16 +17,21 @@
...
-use oak_functions_abi::proto::OakStatus;
+use oak_functions_abi::TfModelInferError;
 
 #[cfg_attr(not(test), no_mangle)]
 pub extern "C" fn main() {
     // Read the input_vector from the request.
     let input_vector = oak_functions::read_request().expect("Fail to read request body.");
 
-    let result = oak_functions::tf_model_infer(&input_vector);
+    let transport = oak_functions::tf_model_infer(&input_vector);
```
tiziano88, line 27: `transport` seems misleading here, this is really just a result too, right? Maybe result_(outer|inner) or something like that?

tiziano88 (approved):

oak_functions/abi/src/lib.rs
mariaschett: We can only derive serialize for structs and enums. When we revisit how we serialize we can probably simplify that.

oak_functions/sdk/oak_functions/src/lib.rs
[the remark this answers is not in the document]
tiziano88: Ah yes you are right, this is the SDK, my bad. I guess here we would always expect the double result then.

PR #2733, opened 2022-04-19
Include the UEFI app by in the workspace

jul-sh:
This is done by keeping a list of non standard crates in xtask, and treating them appropriately. 

Ref #2654

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -57,6 +57,14 @@ use check_build_licenses::CheckBuildLicenses;
 
 static PROCESSES: Lazy<Mutex<Vec<i32>>> = Lazy::new(|| Mutex::new(Vec::new()));
 
+// While these packages are part of the workspace, they cannot be built as
+// usual. This is because they target UEFI. To build them, extra flags need to
+// be passed. The flags cannot currently be specified in the workspace file.
+// Ref: https://github.com/rust-lang/cargo/issues/9451
+// In the interim, they are listed here, and so xtask can invoke them
+// appropriately.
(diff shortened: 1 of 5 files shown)


xtask/src/main.rs
```diff
@@ -798,9 +808,11 @@ fn run_cargo_udeps(scope: &Scope) -> Step {
 
     Step::Multiple {
         name: "cargo udeps".to_string(),
-        steps: workspace_manifest_files()
+        steps: crate_manifest_files()
```
tiziano88, line 811: Just checking: I seem to vaguely remember that switching from compiling crates to compiling workspaces was faster; could you check there are no major regressions by undoing that?

xtask/src/main.rs
```diff
@@ -57,6 +57,14 @@ use check_build_licenses::CheckBuildLicenses;
 
 static PROCESSES: Lazy<Mutex<Vec<i32>>> = Lazy::new(|| Mutex::new(Vec::new()));
 
+// While these packages are part of the workspace, they cannot be built as
+// usual. This is because they target UEFI. To build them, extra flags need to
+// be passed. The flags cannot currently be specified in the workspace file.
+// Ref: https://github.com/rust-lang/cargo/issues/9451
+// In the intermim, they are listed here, and so xtask can invoke them
```
tiziano88, line 64:
```suggestion
// In the interim, they are listed here, and so xtask can invoke them
```

xtask/src/main.rs
jul-sh: Fair point, updating this to keep using workspaces via the `--exclude` flag then

PR #2734, opened 2022-04-19
xtask: Always log commands used

jul-sh:
Found it helpful for debugging


tiziano88:
Should we rather make the commands flag true by default? I also tend to always always pass that since I want to see the actual command as output.

https://github.com/project-oak/oak/blob/4a6666ac9f0cb8acf3a4287666af65112b681914/xtask/src/internal.rs#L35

jul-sh:
@tiziano88 Good idea. Updated this PR to remove that flag and perform its behavior by default.

tiziano88 (approved):
Thanks!

PR #2735, opened 2022-04-19
Import rust-vmm/vmm-reference to third_party.

andrisaar:
This is a completely clean import of `vmm-reference` with no local modifications.

The import is from commit 1ced764be4dbba0a545ceaaddb98da3b89787a94 in https://github.com/rust-vmm/vmm-reference.


tiziano88 (approved):
Could you include at least in the commit message the exact URL and commit id from which this was cloned, so that it is preserved in the history of our own repo?

PR #2738, opened 2022-04-20
Add experimental offline attestation echo example

conradgrobler:
This is a simple example that uses an offline attestation approach for encryption and decryption.

### DIFF
--- a/experimental/offline_attestation/README.md
+++ b/experimental/offline_attestation/README.md
@@ -12,14 +12,14 @@ distinct parts:
   enclave that generated the private key. The public key and attestation report
   are then made availble to clients.
 - Encrypted communications: Clients use the public key (once they are satisfied
-  byt the evidence that only appropriate enclaves would have access to the
+  by the evidence that only appropriate enclaves would have access to the
   private key) to encrypt requests to the server. Clients also provide
   additional information (e.g. the client's public key) in the request to allow
   the server to encrypt the response.
 
 Since the same public key is used for encrypting all requests this approach does
 not have the same forward secrecy that is provided by ephemeral keys in the
-interactive handshake. Avoiding the need to a handshake has some advantages
+interactive handshake. Avoiding the need for a handshake has some advantages
 though:
 
 - Reduced bandwith and latency
@@ -37,8 +37,8 @@ version of the server code. For this example we just generate a fake empty
 attestation report that is always accepted by the client.
 
 For more realistic implementations that support load-balancing a
-key-provisioning mechanism is required to ensure all instances of the server
-have access to the private key. This could be done in two ways:
+key-provisioning mechanism is required to ensure that all instances of the
+server have access to the private key. This could be done in two ways:
 
 - A separate key management server (KMS): The KMS would generate the key-pair
   and attestation report. When a server instance starts up it sets up a
@@ -89,7 +89,7 @@ using is private key and processes the cleartext content.
 This example uses Tink's implementation of
 [ECIES hybrid encryption](https://github.com/project-oak/tink-rust/tree/main/hybrid):
 
-- Key agreement: ECDH with the NIST P-256 curve
+- Key agreement: ECDH using the NIST P-256 curve
 - Key derivation: HKDF using SHA256 with an empty salt
 - Authenticated encryption: 128-bit AES_GCM
 
--- a/experimental/offline_attestation/README.md
+++ b/experimental/offline_attestation/README.md
@@ -5,11 +5,11 @@
 Start an instance of the server:
 
 ```bash
-RUST_LOG=info cargo run --manifest-path=experimental/offline_attestation/server/Cargo.toml
+RUST_LOG=debug cargo run --manifest-path=experimental/offline_attestation/server/Cargo.toml
 ```
 
 In another terminal, run the client:
 
 ```bash
-RUST_LOG=info cargo run --manifest-path=experimental/offline_attestation/client/Cargo.toml
+RUST_LOG=debug cargo run --manifest-path=experimental/offline_attestation/client/Cargo.toml
 ```
--- a/experimental/offline_attestation/README.md
+++ b/experimental/offline_attestation/README.md
@@ -1,0 +1,15 @@
+# Offline Attestation
+
+## Running the example
+
+Start an instance of the server:
+
+```bash
+RUST_LOG=info cargo run --manifest-path=experimental/offline_attestation/server/Cargo.toml
+```
+
+In another terminal, run the client:
+
+```bash
+RUST_LOG=info cargo run --manifest-path=experimental/offline_attestation/client/Cargo.toml
+```
--- a/experimental/offline_attestation/README.md
+++ b/experimental/offline_attestation/README.md
@@ -7,7 +7,7 @@ the runtime. For some scenarios it might be preferable to avoid the handshake.
 The offline attestation approach splits the attestation process into two
 distinct parts:
 
-- Key generation and publication: A single asymmetric key-pair is generated. The
+- Key generation and export: A single asymmetric key-pair is generated. The
   public key is bound to an attestation report that provides evidence about the
   enclave that generated the private key. The public key and attestation report
   are then made availble to clients.
@@ -17,13 +17,14 @@ distinct parts:
   additional information (e.g. the client's public key) in the request to allow
   the server to encrypt the response.
 
-Since the same public key is used for encrypting all requests this approach does
-not have the same forward secrecy that is provided by ephemeral keys in the
-interactive handshake. Avoiding the need for a handshake has some advantages
-though:
-
-- Reduced bandwith and latency
-- No need for sticky sessions in a load-balanced environment
+- This approach has some advantages compared to the interactive handshake-based
+  approach:
+  - Reduced bandwith and latency
+  - No need for sticky sessions in a load-balanced environment
+- And some disadvantages:
+  - No forward secrecy since the same public key is used for encrypting all
+    requests
+  - No built-in replay protection
 
 ## Implementation Details
 
@@ -31,7 +32,7 @@ This experimental example provides simplified client and server implementations.
 
 ### Server
 
-The server generates an assymetric key-pair at startup. It then generates an
+The server generates an asymmetric key-pair at startup. It then generates an
 attestation report that binds the public key to the enclave hardware and the
 version of the server code. For this example we just generate a fake empty
 attestation report that is always accepted by the client.
@@ -77,7 +78,7 @@ report should include checks to ensure that the code would treat the decrypted
 data appropriately and not leak the private key from the enclave.
 
 Once the client has the server's public key it encrypts the request. The client
-also generates its own assymetric key-pair and includes the public key in the
+also generates its own asymmetric key-pair and includes the public key in the
 request. This provides a mechansim for the server to encrypt the response back
 to the client.
 
--- a/experimental/offline_attestation/README.md
+++ b/experimental/offline_attestation/README.md
@@ -1,15 +1,110 @@
 # Offline Attestation
 
-## Running the example
+The current Oak Functions Runtime implementation uses an interactive handshake
+for setting up an attested end-to-end encrypted channel between the client and
+the runtime. For some scenarios it might be preferable to avoid the handshake.
+
+The offline attestation approach splits the attestation process into two
+distinct parts:
+
+- Key generation and publication: A single asymmetric key-pair is generated. The
+  public key is bound to an attestation report that provides evidence about the
+  enclave that generated the private key. The public key and attestation report
+  are then made availble to clients.
+- Encrypted communications: Clients use the public key (once they are satisfied
+  byt the evidence that only appropriate enclaves would have access to the
+  private key) to encrypt requests to the server. Clients also provide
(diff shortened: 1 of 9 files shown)


experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,110 @@
+# Offline Attestation
+
+The current Oak Functions Runtime implementation uses an interactive handshake
+for setting up an attested end-to-end encrypted channel between the client and
+the runtime. For some scenarios it might be preferable to avoid the handshake.
+
+The offline attestation approach splits the attestation process into two
+distinct parts:
+
+- Key generation and publication: A single asymmetric key-pair is generated. The
```
tiziano88, line 10: Sometimes "exporting" is used instead of "publication" to describe this concept, up to you if you think that helps.

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,110 @@
...
+  additional information (e.g. the client's public key) in the request to allow
+  the server to encrypt the response.
+
+Since the same public key is used for encrypting all requests this approach does
+not have the same forward secrecy that is provided by ephemeral keys in the
+interactive handshake. Avoiding the need for a handshake has some advantages
+though:
+
+- Reduced bandwith and latency
+- No need for sticky sessions in a load-balanced environment
```
tiziano88, line 26: I would point out the downsides too, namely lack of replay protection and forward secrecy. WDYT?

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Implementation Details
+
+This experimental example provides simplified client and server implementations.
+
+### Server
+
+The server generates an assymetric key-pair at startup. It then generates an
```
tiziano88, line 34:
```suggestion
The server generates an asymetric key-pair at startup. It then generates an
```

experimental/offline_attestation/client/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+        help = "The message to send to the server",
+        default_value = "test"
+    )]
+    message: String,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    env_logger::init();
+    init();
```
tiziano88, line 50: I would leave this qualified with the crate name for clarity.

experimental/offline_attestation/client/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+    let response_public_key_handle = private_key_handle
+        .public()
+        .map_err(|error| anyhow!("Couldn't get public key: {}", error))?;
+    let response_public_key = serialize_public_key(&response_public_key_handle)?;
+
+    // Build the encrypted request to send to the server.
+    let encrypted_request = EncryptedRequest {
+        ciphertext,
+        response_public_key,
+    };
```
tiziano88, line 87: May be worth abstracting this at some point (i.e. generating the response key, and creating a struct that can decrypt the incoming response once it's received, without ever exposing the private key (handle) at all).

experimental/offline_attestation/shared/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+pub struct EncryptedRequest {
+    /// The encrypted payload.
+    ///
+    /// Contains the raw output from Tink hybrid encryption.
+    pub ciphertext: Vec<u8>,
+
+    /// The public key that should be used for encrypting the response.
+    ///
+    /// Contains the binary-serialised form of the Tink keyset representing the public key.
+    pub response_public_key: Vec<u8>,
```
tiziano88, line 38: I'm curious why you went with this approach (which leaves the public key unencrypted) as opposed to encrypting the public key as part of the payload directly. What are the trade offs between the two options?

experimental/offline_attestation/server/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        .map(move || warp::reply::json(public_key_info.clone().as_ref()))
+}
+
+/// Path filter for "POST /encrypted_echo" with a JSON body.
+fn encrypted_echo(
+    private_key_handle: Arc<Handle>,
+) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    warp::post()
+        .and(warp::path(ECHO_PATH))
+        // Limit the maximum request body size to 4kb.
```
tiziano88, line 101:
```suggestion
        // Limit the maximum request body size to 4KiB.
```

(I think)

namely:

- Ki = 1024 vs k = 1000
- B = bytes vs b = bits

experimental/offline_attestation/server/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+            "Sending encrypted respone: {}",
+            serde_json::to_string(&response)?
+        );
+
+        response
+    };
+    match result {
+        Ok(response) => warp::reply::json(&response).into_response(),
+        Err(_) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
+    }
```
tiziano88, line 133: Maybe something like

```
result.map(warp::reply::json).unwrap_or(StatusCode::INTERNAL_SERVER_ERROR).into_response()
```

experimental/offline_attestation/shared/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// attestation.
+    pub attestation: Vec<u8>,
+}
+
+impl PublicKeyInfo {
+    /// Constructs a new `PublicKeyInfo` instance from the public key handle.
+    pub fn new(public_key_handle: &Handle) -> anyhow::Result<Self> {
+        // In a real implementation this would ask the enclave for an attestation report that binds
+        // the private key to the enclave. For now we just use and empty vector.
+        let attestation = Vec::new();
```
tiziano88, line 79: I think it would still be nice to extract the "attestation" function, even if it's just implemented to return an empty vec for now. It would also start being useful when discussing interactive attestation in more detail.

tiziano88 (approved):
Very neat!

experimental/offline_attestation/shared/src/lib.rs
conradgrobler: The idea is that the client would generate a new key pair for every request/response, so the public key does not contain any usable information and should not enable tracking. Encrypting it does not seem to be required.

I considered a version where a symmetric key was included in the encrypted payload, but Tink tries really hard to stop us from getting to the underlying symmetric keys. If we do a future implemenation using lower-level crypto primitives we could consider that option again.

experimental/offline_attestation/shared/src/lib.rs
conradgrobler: I added a placeholder AttestationReport struct with serialisation and deserialisation, and added a functions for generating the placeholder report.

conradgrobler (commented):
Thanks for the review. PTAL.

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+- Key generation and export: A single asymmetric key-pair is generated. The
+  public key is bound to an attestation report that provides evidence about the
+  enclave that generated the private key. The public key and attestation report
+  are then made availble to clients.
+- Encrypted communications: Clients use the public key (once they are satisfied
+  by the evidence that only appropriate enclaves would have access to the
+  private key) to encrypt requests to the server. Clients also provide
+  additional information (e.g. the client's public key) in the request to allow
+  the server to encrypt the response.
```
ipetr0v, line 18: Can a client send a pre-shared secret instead. So that both sides can create a shared key from it?
And we don't need to have bidirectional asymmetric encryption?

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+- This approach has some advantages compared to the interactive handshake-based
+  approach:
+  - Reduced bandwith and latency
+  - No need for sticky sessions in a load-balanced environment
+- And some disadvantages:
+  - No forward secrecy since the same public key is used for encrypting all
+    requests
+  - No built-in replay protection
```
ipetr0v, line 27: If we will send an encrypted pre-shared secret (encrypted with server's public key) as random value used in the shared key generation - can it help with replay protection?

I.e. the key will be generated as a `PRF(server_pub_key, random_pre_shared_secret)`

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,111 @@
...
+  KMS is then responsible for validating that the server is running an
+  appropriate version of the software in an appropriate enclave before sharing
+  the private key. The attestation report will include evidence about the KMS,
+  not the server, so the KMS code would need to be reviewed to ensure that it
+  will do the server validation correctly.
+- Key-sharing between servers: The first server to start up will be instructed
+  to generate a key-pair and attestation reports. When other servers start up,
+  they will be instructed to connect to an existing server to get the private
+  key via a mutually-attested connection. This approach might present some
+  challenges for key rotation or upgrading to new versions of the server code.
```
ipetr0v, line 56: Why can't they connect to the KMS as well?
Is it needed for scalability?

experimental/offline_attestation/README.md
```diff
@@ -0,0 +1,111 @@
...
+key for the client. The server processes the cleartext, encrypts it with the
+client's public key, and returns the encrypted response.
+
+### Client
+
+For illustration purposes the client fetches the public key and associated
+attestation report from the server and validates the report. In a more realistic
+implemenation the retrieval and validation of the public key and attestation
+report would happen out-of-band. The public key would then be provided to the
+client app once the validity of the key and attestation report has been
```
ipetr0v, line 69: Will this key be hardcoded into the app? If so, how would we work with revocations?

experimental/offline_attestation/shared/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub ciphertext: Vec<u8>,
+
+    /// The public key that should be used for encrypting the response.
+    ///
+    /// Contains the binary-serialised form of the Tink keyset representing the public key.
+    pub response_public_key: Vec<u8>,
+}
+
+impl EncryptedRequest {
+    /// Deserialises the `response_publlic_key` as a keyset handle.
```
jul-sh, line 42: (nit) `response_public_key` (with only one l)

experimental/offline_attestation/server/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+    offline_attestation_shared::init();
+    let opt = Opt::parse();
+    let address = SocketAddr::from((Ipv6Addr::UNSPECIFIED, opt.port));
+
+    // Generate a private key for the lifetime of the server at startup.
+    // In a production implementation there would typically be some form of key provisioning
+    // mechanism that will ensure all servers that run the right code in an appropriate enclave
+    // receive the same key.
+    debug!("Generating private key");
+    let private_key_handle = Arc::new(generate_private_key()?);
```
ipetr0v, line 55: Just to confirm, is the same private key shared between multiple backend servers?

ipetr0v (approved):

jul-sh (approved):
Nice :). Was interesting to see the a concrete implement of the concept

experimental/offline_attestation/README.md
conradgrobler: Yes that would work as well, and for an implementation that uses lower-level crypto primitives that would probably be preferable for performance reasons. This implementation uses hybrid encryption both ways because Tink prevents us from exporting a generated AES key to send it to the server as part of the encrypted request.

experimental/offline_attestation/README.md
conradgrobler: Any type of random value (e.g. a nonce or key) could help, but the server must then keep track of all the nonces it has received to check whether it has seen the current one before. In the interactive handshake, each handshake is based on a different ephemeral key from the server. That, combined with the transcript and the state-machine nature of the handshaker means that reply will be detected at the protocol level.

experimental/offline_attestation/README.md
conradgrobler: These problems for this option are cause when there is no KMS and the keys are just kept alive between all the server instance. If a KMS is introduced it should provide a way to work around these.

experimental/offline_attestation/README.md
conradgrobler: It could be hardcoded or done via configuration. Revocations would be tricky. One option would be to deploy new versions of the client with new keys. As this roll-out will likely take time it will be a challenge to ensure that all the old clients still connect to server instances that have the old key and all new clients connect to servers with the new key. This could be done by creating a new versioned URL every time the key is rotated. Another option is to allow servers to have multiple keys and then stop using the old ones when all the clients have been upgraded. Neither of these are great, but I can't think of many better options.

experimental/offline_attestation/server/src/main.rs
conradgrobler: Yes, it will be in the case where the runtime supports multiple servers sharing keys.

conradgrobler (commented):
Thanks for the review

experimental/offline_attestation/shared/src/lib.rs
tiziano88: Makes sense, thanks.

tiziano88 (approved):

PR #2739, opened 2022-04-20
UEFI example: Use binary messages

jul-sh:
Done in preparation for adding remote attestation handshake support, which takes place using messages serialized to binary using our remote attestation protocol.

### DIFF
--- a/experimental/uefi/proto/echo.proto
+++ b/experimental/uefi/proto/echo.proto
@@ -19,11 +19,11 @@ syntax = "proto3";
 package oak.experimental.uefi;
 
 message EchoRequest {
-    string message = 1;
+    bytes message = 1;
 }
 
 message EchoResponse {
-    string message = 1;
+    bytes message = 1;
 }
 
 service Echo {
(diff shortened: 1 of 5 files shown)


experimental/uefi/proto/echo.proto
```diff
@@ -19,11 +19,11 @@ syntax = "proto3";
...
-    string message = 1;
+    bytes message = 1;
 }
 
 message EchoResponse {
-    string message = 1;
+    bytes message = 1;
 }
 
 service Echo {
```
jul-sh, line 29: This is starting to look very close to `oak_functions/proto/unary_server.proto`. We could use the same proto, but leaving this untouched for now as it relates to the open question of which framing protocol we'll use.

conradgrobler (approved):

ipetr0v (approved):

experimental/uefi/proto/echo.proto
andrisaar: Yeah, we need to take a step back here and figure out how the communication is going to work. But that's a separate task.

andrisaar (approved):

PR #2745, opened 2022-04-21
Move request logic into a separate crate

andrisaar:
This moves the trivial echo code from the UEFI app code into a separate crate.

The goals are two-fold:
(a) First, the UEFI app crate should only be kind of hardware abstraction layer, with no cleverness of its own. All logic should be somewhere else.
(b) Second, this allows us to create a bare-metal 'kernel' which offers a similar abstraction layer, and use the common runtime library shared with the UEFI app.

Future work to be done: all of the attestation-related things should move to this crate as well; I've let @jul-sh know she may need to do some refactoring.


experimental/uefi/runtime/src/echo.rs
```diff
@@ -0,0 +1,67 @@
...
+pub enum Error<T> {
+    // An error occured while deserializing.
+    De(ciborium::de::Error<T>),
+
+    // An error occured while serializing.
+    Ser(ciborium::ser::Error<T>),
+}
+
+/// Basic hardware abstraction layer for sending data.
+pub trait Interface {
```
conradgrobler, line 32: nit: "Interface" is quite an overloaded term, so it might be worth considering a more specific name for clarity (perhaps Channel or CommsInterface?). Also, it seems that this trait is not specific to the echo , so I would consider moving it out of the echo module (either into lib.rs or its own module). Both of these could be addressed in future when non-echo functionality is added to the runtime.

conradgrobler (approved):

experimental/uefi/runtime/src/echo.rs
mariaschett: +1 for Interface being an overloaded term.

experimental/uefi/app/src/serial.rs
```diff
@@ -71,24 +71,16 @@ impl<'boot> Serial<'boot> {
...
 
-    fn flush(&mut self) -> Result<(), Self::Error> {
-        Ok(())
-    }
-}
-
-impl<'boot> ciborium_io::Read for Serial<'boot> {
-    type Error = Error<()>;
-
     // Try to fill the buffer, ignoring any timeout errors. Any other errors
```
mariaschett, line 81: pre-existing: why are we ignoring timeout errors?

mariaschett (approved):
LGTM.

experimental/uefi/app/src/serial.rs
andrisaar: We want to block until we've read a full line, but by default read() will time out every 100 ms or so. Yes, this is inefficient busy-waiting and we'll need to replace it with something better.

PR #2750, opened 2022-04-22
Extract Extension trait

### LINKED ISSUE #2576: Move the structs and traits related to extensions to a separate crate
This will allow us to move the implementations of extensions to separate crates without running into circular reference issues.

mariaschett:
Fixes #2576 .

Note, that we still need to remove our dependency on wasmi which we will be able to do after #2752 .


ipetr0v (approved):

conradgrobler (approved):
Great, thanks.

PR #2754, opened 2022-04-22
Use Ent in Dockerfile to pin artifacts

tiziano88:
This was already done via a combination of curl and sha256, this just
makes it slightly more idiomatic. In the future the ent CLI may add
support for verifying signatures and inclusion proofs, which may also be
of interest.


jul-sh (approved):
lgtm

PR #2759, opened 2022-04-22
Share client <-> server communication proto between runtimes

jul-sh:
Follow up to #2758. The nice thing about this is that it means we'll likely be able to share client logic between the two runtimes. Ie the `oak_functions_client` should work across runtimes with little changes


experimental/uefi/client/src/main.rs
```diff
@@ -20,10 +20,12 @@ use clap::Parser;
...
 
 pub mod proto {
-    tonic::include_proto!("oak.experimental.uefi");
+    tonic::include_proto!("oak.session.unary.v1");
 }
 
-use proto::{echo_client::EchoClient, EchoRequest};
+use proto::{unary_session_client::UnarySessionClient, UnaryRequest};
+
+const MOCK_SESSION_ID: [u8; 8] = [0; 8];
```
jul-sh, line 28: Next up is actually sending this over the UEFI <-> loader app channel, at which point we'll be able to use an actually unique session_id. 

Will have to decide on how to serialize it; could just use protobuf, but we probs don't want it in the UEFI app. Maybe rust struct that supports serde, which would work with our CBOR lib.

tiziano88:
I was trying to find out how to force a reviewer re-assignment; it turns out we need to first remove all reviewers first (including core), save that, and then add back *just* core (then it will be pickes up correctly)

experimental/uefi/client/src/main.rs
tiziano88: Why bother with any serialization framework? AFAICT it's always exactly 8 bytes, so it's fine to just treat is as such. Or are you talking about how to serialize it as part of a struct that also includes the actual data? But even in that case, I think it's fine to assume that the struct is length prefixed, and we just always read the first 8 bytes as session id. I don't think we need to go fancier than that for now.

tiziano88 (approved):

experimental/uefi/client/src/main.rs
jul-sh: > Or are you talking about how to serialize it as part of a struct that also includes the actual data

yes, this. 

> But even in that case, I think it's fine to assume that the struct is length prefixed, and we just always read the first 8 bytes as session id. I don't think we need to go fancier than that for now.

That's fair, probably the best solution; means we can stick to CBOR for just length encoding for now.

Issue #2741, opened 2022-04-21
Share remote attestation server logic


jul-sh:
Most of the non-gRPC specific bits in `grpc_unary_attestation/src/server.rs` will be needed for both the linux runtime (that exposes a gRPC endpoint directly) _and_ the bare-metal/UEFI runtime (that performs handshakes but communicates with clients through a loader app). Specifically those bits are

- [X] the LRU cache track of sessions
- [ ] and the logic for iterating the handshaker with incoming messages.

It would be nice to break out those parts from the grpc crate and use them in both runtimes.

conradgrobler:
I agree. I think this will be very useful refactoring.

tiziano88:
@jul-sh what's the state of this? Has it already happened as part of your refactorings?

jul-sh:
Sorry for the late update. This issue consists of two tasks, the more impactful of which is done. 

- [X] the LRU cache track of sessions (done in #2743)
- [ ] and the logic for iterating the handshaker with incoming messages.

It'd be nice to also share the logic of iterating the handshaker, these two methods atm contain a lot of duplicate code: https://github.com/project-oak/oak/blob/228236fefa5c8279b42138d2452b7fd87f7ff9b1/experimental/oak_baremetal_runtime/src/remote_attestation.rs#L64 and https://github.com/project-oak/oak/blob/main/grpc_unary_attestation/src/server.rs#L71.

Right now however, sharing that logic is non-trivial however. 

The linux runtime's logic needs to be async, while bare-metal runtime does not have an async executor. Logic could only be shared by either a) making more of the linux runtime logic blocking (unacceptable runtime perf cost) or b) adding a blocking executor (ala https://github.com/jonay2000/blocker) to the bare-metal runtime (added complexity imo not worth it).

It will be easier to solve this once the bare-metal runtime supports multiple CPUs (some time away). This issue is left open until then.

conradgrobler:
We are no longer using the intereactive remote attestation handshaker, so this is no longer relevant.

PR #2760, opened 2022-04-22
Make WASM invocation synchronous

jul-sh:
Is there an (implicit) reason why this would be unsafe? The underlying function doesn't seem to need to be async:

`oal/oak_functions/loader/fuzz/fuzz_targets/wasm_invoke.rs`
```
  pub async fn handle_invoke(&self, request: Request) -> anyhow::Result<Response> {
        let request_bytes = request.body;
        let mut wasm_state = self.init_wasm_state(request_bytes)?;

        wasm_state.invoke();
        for (_, extension) in wasm_state.extensions_indices.iter_mut() {
            extension.terminate()?;
        }
        Ok(Response::create(
            StatusCode::Success,
            wasm_state.get_response_bytes(),
        ))
    }
```

In my understanding this seems like it would be safe, each user request is handled in a separate thread already. 
`oak/grpc_unary_attestation/src/server.rs`
```
#[tonic::async_trait]
impl<F, S, L> UnarySession for AttestationServer<F, L>
where
    F: 'static + Send + Sync + Clone + FnOnce(Vec<u8>) -> S,
    S: std::future::Future<Output = anyhow::Result<Vec<u8>>> + Send + Sync,
    L: Send + Sync + Clone + LogError + 'static,
{
    async fn message(
        &self,
        request: tonic::Request<UnaryRequest>,
    ) -> anyhow::Result<tonic::Response<UnaryResponse>, tonic::Status> {
              // ...
                let response = (self.request_handler.clone())(decrypted_request)
                    .await
                    .map_err(|error| {
                        error_logger.log_error(&format!("Couldn't handle request: {:?}", error));
                        tonic::Status::aborted("")
                    })?;
               // ...
    }
}
```

--- 

Looking into this as it'd help with #2741: refactoring remote attestation logic to share it across runtimes.


conradgrobler (approved):
Nice, I wanted to do this for quite a while but never got around to doing it.

conradgrobler:
Before merging, please increase the version number to clear the cache so that the run-cargo-tests CI step also gets run.

oak_functions/loader/src/server.rs
```diff
@@ -592,14 +592,13 @@ fn check_export_function_signature(
 /// response may be padded by a number of trailing 0s before encoding the response as a binary
 /// protobuf message. In this case, the `length` in the response will contain the effective length
 /// of the `body`. This response is guaranteed to comply with the policy's size restriction.
-pub async fn apply_policy<F, S>(policy: ServerPolicy, function: F) -> anyhow::Result<Response>
+pub async fn apply_policy<F>(policy: ServerPolicy, function: F) -> anyhow::Result<Response>
```
tiziano88, line 595: Uhmm I'm not super convinced though. For instance this function is async, but it expects the underlying function to be sync, which means that e.g. if there is a sleep or some other blocking IO this async task will be stuck in there. Also ideally even the Wasm engine itself would be async, in which case we would have to undo this change, right? That said, please go ahead with this change if you think it's useful (and while our wasm engine of choice is sync anyways).

jul-sh: > Uhmm I'm not super convinced though. For instance this function is async, but it expects the underlying function to be sync, which means that e.g. if there is a sleep or some other blocking IO this async task will be stuck in there.

I'm not sure I follow, why does the async status of the accepted function make a difference? It's still run inside the async task spawned by this function: `let task = tokio::spawn(async move { function() });`

conradgrobler: I don't think there are any async Wasm engines available. The only place where async things could possibly make sense would be inside an ABI call, but I don't know how a Wasm engine would implement that. Even if it was supported I think we should always try to avoid async in extensions/ABI calls, as these are only useful for the types of things we want to avoid (e.g. disk or network access).

jul-sh: Reading about this a bit more, out of curiosity, how we considered using tokio's `spawn_blocking` instead of the normal `spawn` here? It seems like it's intended for the kind of thing we're doing. https://docs.rs/tokio/0.2.25/tokio/task/fn.spawn_blocking.html

tiziano88: I (for one) was not aware of it, from reading the Rust doc I agree it seems sensible

jul-sh: Will open a PR to use it then. :D

PR #2764, opened 2022-04-25
Separate Oak Invocation Protos from remote attestation logic

jul-sh:
Makes it so the remote attestation client expects to be invoked with binary payloads. This allows us to share it across runtimes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1776,7 +1776,6 @@ name = "grpc_unary_attestation"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "oak_functions_abi",
  "oak_remote_attestation",
  "oak_remote_attestation_sessions",
  "oak_utils",
--- a/grpc_unary_attestation/Cargo.toml
+++ b/grpc_unary_attestation/Cargo.toml
@@ -8,7 +8,6 @@ license = "Apache-2.0"
 [dependencies]
 oak_remote_attestation = { path = "../remote_attestation/rust/" }
 oak_remote_attestation_sessions = { path = "../remote_attestation_sessions/" }
-oak_functions_abi = { path = "../oak_functions/abi/" }
 anyhow = "*"
 prost = "*"
 prost-types = "*"
--- a/grpc_unary_attestation/src/client.rs
+++ b/grpc_unary_attestation/src/client.rs
@@ -16,7 +16,6 @@
 
 use crate::proto::{unary_session_client::UnarySessionClient, UnaryRequest};
 use anyhow::Context;
-use oak_functions_abi;
 use oak_remote_attestation::handshaker::{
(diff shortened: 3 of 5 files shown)


tiziano88:
What will / should happen to the corresponding Response message?

jul-sh:
@conradgrobler Fair point. On the runtime level that asymmetry isn't new, but agreed it's not ideal.

tiziano88:
I agree with @conradgrobler ; my suggestion is to leave things as they are and wait until we have a better story for the IDL, at which point we can replace both request and response together if we decide to do so.

That said, the current dependency of grpc_remote_attestation on oak_functions_abi does seem odd, would it be possible to only get rid of that for now? To your point that is currently used by the client but not in any meaningful way.

tiziano88:
i.e. perhaps the client should keep both the request / response protos, but just convert them to bytes when passing them through the remote attestation layers?

jul-sh:
> That said, the current dependency of grpc_remote_attestation on oak_functions_abi does seem odd, would it be possible to only get rid of that for now? To your point that is currently used by the client but not in any meaningful way.

Agreed with it seeming odd, imo the remote attestation client should have no knowledge-of-, or link to the other runtime logic. It should just be "request bytes in", "response bytes out". 

We can remove "just" the usage of that proto there. That said, it's the _only_ place where said proto is used. 

> i.e. perhaps the client should keep both the request / response protos, but just convert them to bytes when passing them through the remote attestation layers?

Not possible, for requests the runtime expects to receive just bytes, not a serialized message proto. The asymmetry lies on the runtime level and precedes this change. It expects just bytes and a serialized proto. 


The only thing we could do is move the reading of the request proto one layer up, from the  `grpc_remote_attestation` client into the `oak_functions_client`. 

Happy to do that since it allows us to keep it proto. But worth emphasizing that this won't fix the underlying asymmetry in what the runtime expects.

tiziano88:
So, the main asymmetry is due to the fact that the input is a slice of bytes, but the output, after the constant-size policy is applied, is actually two slices (one with the real response, one with zeroes -- or, equivalently, the two slices concatenated plus a constant size number of where the first slice ends). So in a way it makes sense that the output type is "more complicated" than the input one. There is also the extra complication of the status code, but perhaps we can ignore that for now.

So, in this case, perhaps what we would want to do is to define a custom serialization for the constant size output bytes (including the size of the "real" response) that does not use protobuf (it can just be a length delimited framing, perhaps exactly the same that we would use elsewhere anyways -- cc @andrisaar ). Then, the Wasm logic would accept bytes and return bytes, but the policy layer on top of that would actually wrap the response bytes into this new format; on the client, a corresponding layer needs to be added so that, after decoding the result from the remote attestation protocol, the extra padding bytes are removed. Again, in this case it may seem odd for this to be asymmetric between request and response, but I think it actually makes sense.

tiziano88 (approved):

PR #2767, opened 2022-04-25
Call `storage_item_get` through invoke

mariaschett:


oak_functions/sdk/oak_functions/tests/lookup_module/src/lib.rs
```diff
@@ -97,7 +97,7 @@ impl TestManager<'static> {
         oak_functions::write_response(&value.unwrap()).expect("Failed to write response.");
     }
 
-    /// Tests [`oak_functions_abi::storage_get_item`] when the key is not in the lookup data. The
+    /// Tests `storage_get_item` when the key is not in the lookup data. The
```
conradgrobler, line 100: nit: the new comments look narrower (above as well) than the original ones. Do you have a different width setting in the formatter?

conradgrobler (approved):

PR #2768, opened 2022-04-25
Add remote attestation support to the UEFI client

jul-sh:
Completes the process of adding remote attestation to the UEFI app, post merging this both client and server perform a handshake.


experimental/uefi/client/src/main.rs
```diff
@@ -14,18 +14,13 @@
 // limitations under the License.
 //
 
+use anyhow::Context;
+use grpc_unary_attestation::client::AttestationClient;
```
jul-sh, line 18: Because of the prior refactor of https://github.com/project-oak/oak/pull/2764, we can simply drop in the existing handshake client & reuse it's logic.

tiziano88 (approved):

PR #2769, opened 2022-04-25
Support multiple sessions in the UEFI app

jul-sh:
This adds supports for multiple sessions by sending the actual session_id to the UEFI app alongside each request. The UEFI app already supports maintaining handshakes for multiple sessions, but the key was previously mocked.


experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,16 @@ use proto::{
     UnaryRequest, UnaryResponse,
 };
 
+fn serialize_request(request: UnaryRequest) -> Vec<u8> {
+    // The payload is the request's body prepended with the 8 byte session_id.
+    // This takes adavantage of the session_id's fixed size to avoid needing
+    // to use a key/value binary serialization protocol.
```
jul-sh, line 34: As suggested by @tiziano88 in https://github.com/project-oak/oak/pull/2759#discussion_r856435136 :)

experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,17 @@ use proto::{
...
+    // The payload is the request's body prepended with the 8 byte session_id.
+    // This takes adavantage of the session_id's fixed size to avoid needing
+    // to use a key/value pair binary serialization protocol.
+    request
+        .session_id
+        .into_iter()
+        .chain(request.body.into_iter())
+        .collect()
+}
+
```
jul-sh, line 41: Note that the UEFI <-> loader channel does technically use a key/value pair serialization protocol; CBOR added in: #2704. However it's currently only used to delineate messages, that consists purely out of a byte array. 

See #2704 for more discussion on CBOR, we're considering moving away from it.

experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,17 @@ use proto::{
...
```
jul-sh, line 41: Note that the UEFI <-> loader channel does technically use a key/value pair serialization protocol alread: CBOR added in: https://github.com/project-oak/oak/pull/2704.

However it's currently only used to length encode messages, that each consist purely out of a byte array.

See https://github.com/project-oak/oak/pull/2704 for more discussion on CBOR, we're considering moving away from it. There's a wider ongoing effort to define the full framing protocol used for UEFI <-> loader comms.

Hence this PR does not add to the use of CBOR, instead sticking sticking with this simpler serialization of requests.

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -22,12 +22,31 @@
...
 use alloc::vec::Vec;
-use anyhow::Context;
-use oak_remote_attestation_sessions::{SessionId, SessionState, SessionTracker};
+use anyhow::{bail, Context};
+use oak_remote_attestation_sessions::{SessionId, SessionState, SessionTracker, SESSION_ID_LENGTH};
 
 /// Number of sessions that will be kept in memory.
 const SESSIONS_CACHE_SIZE: usize = 10000;
 
+fn deserialize_request(msg: Vec<u8>) -> anyhow::Result<(SessionId, Vec<u8>)> {
```
conradgrobler, line 31: nit: a slice (`&[u8]`) can be passed here.

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -22,12 +22,31 @@
...
+    let mut session_id: SessionId = [0; SESSION_ID_LENGTH];
+    let mut request_body: Vec<u8> = Vec::with_capacity(msg.len() - SESSION_ID_LENGTH);
+
+    if msg.len() < SESSION_ID_LENGTH {
+        bail!("Message too short to contain a SessionId. The length of a SessionId is {} bytes, the message received contained only {} bytes", SESSION_ID_LENGTH, msg.len());
+    }
+
+    for (index, byte) in msg.into_iter().enumerate() {
```
conradgrobler, line 39: nit: I think that splitting the slice (https://doc.rust-lang.org/std/primitive.slice.html#method.split_at) and then using `copy_to_slice` for the message id and `extend_from_slice` for the request body might make this more readable.

experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,17 @@ use proto::{
     UnaryRequest, UnaryResponse,
 };
 
+fn serialize_request(request: UnaryRequest) -> Vec<u8> {
+    // The payload is the request's body prepended with the 8 byte session_id.
+    // This takes adavantage of the session_id's fixed size to avoid needing
+    // to use a key/value pair binary serialization protocol.
+    request
```
conradgrobler, line 35: nit: I think that initialising the vec with the right capacity and then calling `extend_from_slice` twice might be more efficient.

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -22,12 +22,31 @@
...
 
 /// Number of sessions that will be kept in memory.
 const SESSIONS_CACHE_SIZE: usize = 10000;
 
+fn deserialize_request(msg: Vec<u8>) -> anyhow::Result<(SessionId, Vec<u8>)> {
+    let mut session_id: SessionId = [0; SESSION_ID_LENGTH];
+    let mut request_body: Vec<u8> = Vec::with_capacity(msg.len() - SESSION_ID_LENGTH);
+
+    if msg.len() < SESSION_ID_LENGTH {
+        bail!("Message too short to contain a SessionId. The length of a SessionId is {} bytes, the message received contained only {} bytes", SESSION_ID_LENGTH, msg.len());
```
conradgrobler, line 36: nit: this is a very long statement, and the formatting doesn't formact insidde macros. Perhaps break the string manually into multiple lines.

conradgrobler (approved):
thanks

experimental/uefi/runtime/src/remote_attestation.rs
jul-sh: Thanks, I'm curious, why would a slice be preferable?

conradgrobler: A slice is more general, so it would allow calling the function with a slice from an array, a string, a vector etc without having to convert it into a vector first. At the moment I guess this is only used in one place, but in future there might be more uses (e.g. for tests).

experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,17 @@ use proto::{
     UnaryRequest, UnaryResponse,
 };
 
+fn serialize_request(request: UnaryRequest) -> Vec<u8> {
+    // The payload is the request's body prepended with the 8 byte session_id.
```
ipetr0v, line 32: What does session id correspond to?
Is it a grpc or https session, or it's an uefi specific thing?

ipetr0v (approved):

experimental/uefi/loader/src/server.rs
jul-sh: It's sent by the client to identify requests as coming from the same client. In the current implementation sessions are done over gRPC, though that technically does not matter for the loader <-> UEFI link.

experimental/uefi/loader/src/server.rs
jul-sh: Yup, looks like it'd be faster: https://stackoverflow.com/a/63902932

I guess that makes sense, when calling collect rust does not know the size of the final Vec in advance.

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -22,12 +22,31 @@
...
+        bail!("Message too short to contain a session_id.");
+    }
+
+    for (index, byte) in msg.into_iter().enumerate() {
+        if index <= SESSION_ID_LENGTH {
+            session_id[index] = byte;
+        } else {
+            request_body.push(byte);
+        }
+    }
```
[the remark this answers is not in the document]
jul-sh: Updated this as part of improving readability as suggested by @conradgrobler. Not sure if cloning at once is more performant however, even with it effectively just iterates: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice.

It would be nice to split without reallocating, but it seems that rust doesn't let you do that because of how it handles memory: 
https://stackoverflow.com/a/68180453
https://users.rust-lang.org/t/split-owned-vec-t-without-reallocation/6346

andrisaar (approved):

experimental/uefi/loader/src/server.rs
```diff
@@ -28,6 +28,19 @@ use proto::{
...
+    // This takes adavantage of the session_id's fixed size to avoid needing
+    // to use a key/value pair binary serialization protocol.
```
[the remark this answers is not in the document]
tiziano88: It would be good to perhaps list what are the factors that we should consider in the first place, since for instance IIRC we already have a protobuf definition of this struct, but we are not using that directly in order to avoid requiring a hard dependency on protobuf for the trusted UEFI app (@jul-sh please correct me if I'm wrong).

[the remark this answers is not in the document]
jul-sh: It basically goes back to the previous discussion here: https://github.com/project-oak/oak/pull/2759#discussion_r856430066. We already have a protobuf definition of this struct, but adding protobuf to the UEFI app just for that seems overkill.

The important thing imo is that this structure is very simple. All the interesting bits are in the binary-serialized body, the only thing that isn't is the fixed size session_id. It's seems rather unlikely that we'd want to add any other key/value pairs to this struct. 

Imo how individual client requests are serialised also isn't necessarily linked to our framing layer (currently CBOR). The framing layer is about separating different messages of various types that are sent UEFI runtime (lookup data, control frames, user requests, etc pp). This seems like it's a layer underneath that.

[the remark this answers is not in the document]
tiziano88: So, from the client point of view we have gRPC / protobuf, since we know that's already something that most clients will have to deal with anyways. Actually that may be transcoded to JSON + HTTP anyways, so perhaps even clients that don't want to depend on gRPC / protobus may not have to depend on it.

Instead between Untrusted and Trusted runtime, the requirements come mostly from the trusted runtime itself and its dependencies (since they all end up being part of the TCB); in addition, even though it is not a short term goal, I think we should think about other trusted runtimes being written in other languages in the future (C++ possibly for some other potential collaboration). In any case, I think that the reference format of this serialization layer should just be specified in a text file somewhere, similar to how like TCP packets are defined (and the spec should be sufficient to implement de/serializers in any language); e.g. https://en.wikipedia.org/wiki/Transmission_Control_Protocol#TCP_segment_structure . In our case, I think having a "spec" that says that the first 8 bytes are the session id and the rest are the payload seems enough to me, we can naively implement that in any language without any additional dependencies; we don't need backwards compatibility or other features that protobuf and CBOR are usually used for.

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -22,12 +22,31 @@
...
 use alloc::vec::Vec;
-use anyhow::Context;
-use oak_remote_attestation_sessions::{SessionId, SessionState, SessionTracker};
+use anyhow::{bail, Context};
+use oak_remote_attestation_sessions::{SessionId, SessionState, SessionTracker, SESSION_ID_LENGTH};
 
 /// Number of sessions that will be kept in memory.
 const SESSIONS_CACHE_SIZE: usize = 10000;
 
+fn deserialize_msg(msg: Vec<u8>) -> anyhow::Result<(SessionId, Vec<u8>)> {
```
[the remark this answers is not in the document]
jul-sh: @tiziano88 updated the PR with the requested refactor to a struct & traits, could you take another look? :)

tiziano88 (approved):
Nice,I like it! Thanks for implementing it

PR #2770, opened 2022-04-25
Use tokio's spawn_blocking for when invoking the WASM module

jul-sh:
Follow up to the discussion on: https://github.com/project-oak/oak/pull/2760#discussion_r857801686;

The background is that tokio executes futures on a small number of threads, swapping the running task in a thread at await points. Futures that take a long time to yield (like ours) can block a thread. For these tokio recommends the usage of `spawn_blocking` API, which runs those futures on a separate thread pool. 

See https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code for more details. 

Questions: 
- Why does tokio use this approach of using only a handful of threads and swapping tasks? Is it purely a performance improvement or is there another reason we'd like to limit the number of threads used? (Tokio [offers guidance](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) on how to run blocking tasks on a limited number of threads).
- Tokio also has a[n API to set the timeout for threads in the blocking pool](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_keep_alive). Are we fine with the default?


oak_functions/examples/weather_lookup/module/src/tests.rs
```diff
@@ -81,7 +81,12 @@ async fn test_server() {
...
     let policy = ServerPolicy {
         constant_response_size_bytes: 100,
-        constant_processing_time_ms: 200,
+        // Multiply the constant_processing_time_ms since runtime performance
+        // is expected to be slower when running tests. Cargo's test profile
+        // builds the WASM engine without optimizations. Based on test config
+        // the WASM module itself may also not be optimized. Creating lookup
+        // tables for the weather example is computationally expensive.
+        constant_processing_time_ms: 200 * 10,
     };
```
jul-sh, line 90: It appears that prior to this change the `constant_processing_time_ms` was not enforced in tests (eg tests would pass even after setting it to 1ms). Unlike `spawn_blocking`, in the case of `spawn` the executor uses only a handful of threads ([in our case 4](https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/tests.rs#L71)). When those threads become blocked it stops the executor from driving futures forward . By default with `spawn_blocking` tokio uses a much higher upper limit for threads.

See: https://docs.rs/tokio/0.2.25/tokio/task/fn.spawn_blocking.html

tiziano88: Thanks for figuring this out, I'm still not sure exactly what was happening before this PR though; it was using all 4 threads, but why did that result in the 1ms test case passing? Is that because the start of the 1ms timer didn't actually start until after the actual function had finished running already?

tiziano88 (approved):

oak_functions/examples/weather_lookup/module/src/tests.rs
conradgrobler: This test used only a single thread (based on the `#[tokio::test]` attribute). To use 4 threads it would need `#[tokio::test(flavor = "multi_thread", worker_threads = 4)]`. Even with that on `test_long_response_time` it was not consistent. It seemed that tokio sometimes decided to spawn the timer and the Wasm task on the same thread even if there were 4 threads available, causing the test to be flaky. Perhaps it uses some randomised algorithm for deciding which thread to use.

conradgrobler (approved):

PR #2772, opened 2022-04-26
Make workload logging an extension

conradgrobler:
This change removes `write_log_message` from the ABI and uses an extension to provide logging functionality to Wasm business logic instances.


oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -112,12 +112,10 @@ pub fn report_metric<T: AsRef<str>>(
...
 /// These log messages are considered sensitive, so will only be logged by the runtime if the
 /// `oak_unsafe` feature is enabled.
-///
-/// See [`write_log_message`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#write_log_message).
 pub fn write_log_message<T: AsRef<str>>(message: T) -> Result<(), OakStatus> {
     let buf = message.as_ref().as_bytes();
-    let status = unsafe { oak_functions_abi::write_log_message(buf.as_ptr(), buf.len()) };
-    result_from_status(status as i32, ())
+    invoke(oak_functions_abi::ExtensionHandle::LoggingHandle, buf)?;
+    Ok(())
```
andrisaar, line 118: I don't know what the type signature of `invoke()` is, but do we need the explicit `Ok(())` here? If the types match, then just calling `invoke()` without the `?` should work.

andrisaar (approved):

oak_functions/workload_logging/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+            .log_sensitive(Level::Debug, &format!("[Wasm] {}", log_message));
+        Ok(Vec::new())
+    }
+
+    fn get_metadata(&self) -> (String, wasmi::Signature) {
+        let signature = wasmi::Signature::new(
+            &[
+                ABI_USIZE, // buf_ptr
+                ABI_USIZE, // buf_len
+            ][..],
```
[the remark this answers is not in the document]
conradgrobler: The function first argument must implement `Into<Cow<'static, [ValueType]>>>`. This is implemented for a slice, but not a reference to an array.

oak_functions/sdk/oak_functions/src/lib.rs
conradgrobler: The return value for invoke is `Result<Vec<u8>>`.

conradgrobler (commented):
Thanks

PR #2774, opened 2022-04-26
Remove extension calls through extension indices

### LINKED ISSUE #2715: Replace `extension_indices` by mapping from handle to extension

### LINKED ISSUE #2752: Remove `get_metadata` from `OakApiNativeExtension`
Once we call all extensions with `invoke` we don't need the `get_metadata` anymore.

mariaschett:
Fixes #2752. Fixes #2715 .


oak_functions/loader/src/server.rs
```diff
@@ -332,16 +328,15 @@ impl WasmState {
...
-            .extensions_indices
-            .values_mut()
-            .find(|extension| extension.get_handle() == handle)
-            .ok_or({
-                // TODO(#2715): Add logging, which will be easier when we have a lookup to
-                // extension_indices.
-                // self.log_error(&format!("Cannot find extension with handle {:?}.", handle));
-                OakStatus::ErrInvalidHandle
-            })?;
+        // TODO(mschett): Make more idiomatic
```
mariaschett, line 331: I should be able to make this more idiomatic, but I cannot see how. The problem here is that `self.log_error` needs a mutable reference to self.

Can you see a way?

oak_functions/loader/src/server.rs
```diff
@@ -699,7 +635,8 @@ impl WasmHandler {
 
         wasm_state.invoke();
 
-        for (_, extension) in wasm_state.extensions_indices.iter_mut() {
+        // TODO(mschett): Make more idiomatic.
```
mariaschett, line 638: Here I am not sure how I can deal with the `?` if I use something like `.for_each()` to iterate. Is there a better way?

oak_functions/loader/src/server.rs
```diff
@@ -389,65 +384,17 @@ impl wasmi::Externals for WasmState {
...
-                                    Ok(())
-                                }
-                            }
-                            Err(err) => Err(err),
-                        }
-                    }
-                    Err(err) => Err(err),
-                };
-                from_oak_status_result(result)
+                // TODO(mschett): Should we use host error instead?
```
mariaschett, line 395: And if so: have we used `TrapKind::Host` before?

oak_functions/loader/src/server.rs
tiziano88: maybe something like this? (I haven't actually tried it)

```rust
wasm_state.extensions.iter_mut().map(|e| e.terminate()).collect::<Result<Vec<_>, _>>()?;
```

oak_functions/loader/src/server.rs
tiziano88: Can that be made to take &self instead? Either way, what do you mean with more idiomatic? Like

```rust
let extension = self.extensions.get_mut(&handle).or(Err(OakStatus::ErrInvalidHandle))?;
```

but with logging?

oak_functions/loader/src/server.rs
tiziano88: Or maybe this should be `TrapKind::Unreachable`? Since the actual calls should all be statically checked by wasmi at load time, right? But otherwise, `TrapKind::Host` seems sensible too.

mariaschett: `Unreachable` seems good enough too, but I am not sure how to justify that I chose that.

oak_functions/loader/src/server.rs
mariaschett: That worked! Thank you!

oak_functions/loader/src/server.rs
conradgrobler: Unrelated to the borrowing: if anything should happen only when the value is `None` (e.g. logging) it should use `ok_or_else`. With `ok_or` the logging will always happen because the block will first be evaluated and then the result will be passed to `ok_or`.

Regarding the borrowing, I think the only way around it is to get a clone of the logger from self beforehand and then use that logger in the closure to do the logging. I don't think that that would be an improvement over the current code though. One minor thing to make this slightly more idiomatic would be to remove the `return` and `;` in the `None` arm of the match.

oak_functions/loader/src/server.rs
mariaschett: In their example, wasmi seems to panic with `Unimplemented function at {}` in the `_`  case [1]. Should we copy that?

 [1] https://paritytech.github.io/wasmi/wasmi/trait.Externals.html#examples

oak_functions/loader/src/server.rs
mariaschett: Thank you, Conrad! Yes, I understand now!

oak_functions/loader/src/server.rs
```diff
@@ -389,65 +384,16 @@ impl wasmi::Externals for WasmState {
...
-                                }
-                            }
-                            Err(err) => Err(err),
-                        }
-                    }
-                    Err(err) => Err(err),
-                };
-                from_oak_status_result(result)
+                // Following https://paritytech.github.io/wasmi/wasmi/trait.Externals.html#examples.
+                panic!("Unimplemented function at {}.", index)
```
tiziano88, line 396: I don't think we can actually panic here (or anywhere) though, wouldn't this result in the panic being observable, and maybe exploitable by malicious Wasm code to leak information?

oak_functions/loader/src/server.rs
```diff
@@ -672,23 +607,16 @@ impl WasmHandler {
...
 
-        for (ind, factory) in self.extension_factories.iter().enumerate() {
+        // Create an extension from every factory.
+        for factory in self.extension_factories.iter() {
             let extension = factory.create()?;
-            let (name, signature) = extension.get_metadata();
-            extensions_indices.insert(ind + EXTENSION_INDEX_OFFSET, extension);
-            extensions_metadata.insert(name, (ind + EXTENSION_INDEX_OFFSET, signature));
+            extensions.insert(extension.get_handle(), extension);
         }
```
tiziano88, line 616: Maybe this could be something like this, but not sure it's worth it, up to you.

```rust
let extensions = self
  .extension_factories
  .iter()
  .map(|f| f.create())
  .collect::<Result<Vec<_>, >()?
  .into_iter()
  .map(|e| (e.get_handle(), e))
  .collect::<HashMap<_, _>();
```

Or maybe something even nicer with https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_collect

tiziano88 (approved):

oak_functions/loader/src/server.rs
mariaschett: Next time I try :)

oak_functions/loader/src/server.rs
mariaschett: Weak defence, but: it did panic before and it really shouldn't happen.

Do you think returning the Unreachable/a Host Error is better?

tiziano88: I prefer we don't panic at all, otherwise it will be something that we always have to remember why we did an exception for (like seeing unsafe code in our own codebase, for instance).

Issue #2533, opened 2022-02-14
Flatten Rust crates hierarchy


tiziano88:
Currently some of our own Rust crates are under other folders, but crates actually all live in the same namespace, so this level of nesting is usually more confusing than helpful, since e.g. it is not reflected in the name of the crate or anywhere else. 

I think it would be more idiomatic to have all the crates in the workspace at the same level, and use crate name prefixes to group similar crates (e.g. `oak_example_*`).

conradgrobler:
We should probably decide on an overall approach for all the content on the site. If we move everything, including e.g. java code to the top level it might become difficult to find anything an navigate.

Also, it might be useful to differentiate between library crates, which might at some point be published to crates.io and binary crates. This would be especially useful if we lean more strongly into the idea of having a distinct crate for every combination of runtime + plugins + platform.

tiziano88:
Those are all good things, I just don't think they should be expressed by nesting crates under specific paths, which gives the impression of a nested hierarchy of crates, while crates all exist in a flat namespace (and so for instance ideally will probably require finding unique names, even if we decide not to publish them); e.g. for lib vs bin crates, the Cargo.toml file is the source of truth; if we want to keep differentiating experimental from "ready" crates, we should probably do that by documenting it explicitly anyways, etc.

re: Java vs Rust vs other languages, perhaps we should split those at top level? e.g. have `rust` and `java` folders in the root of the repo, instead of, say, nesting the Java remote attestation logic under the Rust remote attestation crate. But I'm also not sure that's ideal.

conradgrobler:
> re: Java vs Rust vs other languages, perhaps we should split those at top level? e.g. have `rust` and `java` folders in the root of the repo, instead of, say, nesting the Java remote attestation logic under the Rust remote attestation crate. But I'm also not sure that's ideal.

Yes, it is not ideal, but I also cannot really think of any better approaches.

tiziano88:
After some more thought, I think we should split things by language first (similar to https://github.com/google/tink ). Though in the short term we can just assume that everything in the root is Rust, and just create top-level folders for non-rust languages (i.e. `java` and `cc`), and move the corresponding code under those.

For the Rust crates, we should flatten them so they are all at the same level in the root, and each of them should be named exactly as the folder that contains it.

Additionally, even though we are not planning to publish most of them to crates.io, it would be good if all our crates had globally unique names, so I suggest we at least prefix all of them with `oak_` if they are not already.

[...]

PR #2777, opened 2022-04-26
Update instructions on how to run UEFI code

andrisaar:


experimental/uefi/README.md
```diff
@@ -1,38 +1,37 @@
...
-This directory contains a bare-bones UEFI app and an "untrusted loader" built
-around `qemu` that communicates with said app. The goal of the experiment is to
-try out different communication mechanisms etc to see how Oak could run in a
-trusted environment as an UEFI app with minimal dependencies, with code that
-doesn't need to be trusted left outside the trusted environment.
+This directory contains the following crates:
 
-To run:
+- `runtime`: common business logic that should run in a VM
+- `app`: UEFI application that wraps `runtime` for execution under UEFI
```
tiziano88, line 6: Note that these are not actually the crate name; that would be this:

https://github.com/project-oak/oak/blob/3268b34f431f9ecf57d16275d980ae30ccbf942c/experimental/uefi/app/Cargo.toml#L2

BTW I think there are inconsistencies with the naming already, may be worth fixing them.

In fact we should actually avoid nesting crates into other folders, since crates actually live in a global namespace, so the directory nesting is actually more confusing than helpful. See https://github.com/project-oak/oak/issues/2533

andrisaar: Fair enough, I've stopped calling them crates. :)

I think right now it makes sense to keep them under experimental, at least until we know which platforms we are targeting.

experimental/uefi/README.md
```diff
@@ -1,38 +1,37 @@
...
-around `qemu` that communicates with said app. The goal of the experiment is to
-try out different communication mechanisms etc to see how Oak could run in a
-trusted environment as an UEFI app with minimal dependencies, with code that
-doesn't need to be trusted left outside the trusted environment.
+This directory contains the following:
 
-To run:
+- `runtime`: common business logic that should run in a VM
+- `app`: UEFI application that wraps `runtime` for execution under UEFI
+- `loader`: wrapper around `qemu` that loads the above and exposes and gRPC
```
conradgrobler, line 7: nit: "exposes a gRPC"

conradgrobler (approved):

jul-sh (approved):
thanks!

PR #2778, opened 2022-04-26
Update QEMU CPU model to lowest Intel CPU that supports RDRAND.

jul-sh:
The previous model of Broadwell also supports other newer features like RDSEED, which we don't actually require need. This leads to issues on older hosts that don't support that feature.

The enforce flag is also added, to trigger an error if hosts don't required support CPU features.


tiziano88 (approved):
Seems fine but I'll let @andrisaar give the final approval.

andrisaar (approved):

PR #2790, opened 2022-04-28
Add Trusted Shuffler server

ipetr0v:
This PR adds Trusted Shuffler to the corresponding server binary.

Ref #2614

### DIFF
--- a/experimental/trusted_shuffler/scripts/run
+++ b/experimental/trusted_shuffler/scripts/run
@@ -2,15 +2,22 @@
 
 readonly SCRIPTS_DIR="$(dirname "$0")"
 
-cargo build --manifest-path=experimental/trusted_shuffler/backend/Cargo.toml
- "${SCRIPTS_DIR}/run_backend" &
-readonly BACKEND_PID=$!
-
 cargo build --manifest-path=experimental/trusted_shuffler/server/Cargo.toml
-"${SCRIPTS_DIR}/run_server" &
+cargo build --manifest-path=experimental/trusted_shuffler/backend/Cargo.toml
+cargo build --manifest-path=experimental/trusted_shuffler/client/Cargo.toml
+
+RUST_LOG=info cargo run --manifest-path=experimental/trusted_shuffler/server/Cargo.toml -- \
+  --k=1 --listen-address="[::]:8080" --backend-url="http://localhost:8888" &
+sleep 1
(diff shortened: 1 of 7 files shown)


experimental/trusted_shuffler/scripts/run
```diff
@@ -2,15 +2,22 @@
...
+cargo build --manifest-path=experimental/trusted_shuffler/server/Cargo.toml
 cargo build --manifest-path=experimental/trusted_shuffler/backend/Cargo.toml
- "${SCRIPTS_DIR}/run_backend" &
-readonly BACKEND_PID=$!
+cargo build --manifest-path=experimental/trusted_shuffler/client/Cargo.toml
 
-cargo build --manifest-path=experimental/trusted_shuffler/server/Cargo.toml
-"${SCRIPTS_DIR}/run_server" &
+RUST_LOG=info cargo run --manifest-path=experimental/trusted_shuffler/server/Cargo.toml -- \
+  --k=1 --listen-address="[::]:8080" --backend-url="http://localhost:8888" &
```
tiziano88, line 10: k=1 ?

tiziano88 (approved):

experimental/trusted_shuffler/scripts/run
ipetr0v: This PR doesn't include the possibility of sending more than 1 request yet

ipetr0v: The next PR will be about sending multiple requests from the test client (and checking the delay as well)

experimental/trusted_shuffler/scripts/run
```diff
@@ -2,15 +2,22 @@
...
+sleep 1
```
mariaschett, line 11: Why do we sleep?

mariaschett (approved):

experimental/trusted_shuffler/scripts/run
ipetr0v: Otherwise the background task may run after the client and thus fail

mariaschett: I assumed something like that. :) Maybe we want to add a comment?

ipetr0v: It's the ultimate Bash synchronization primitive

PR #2793, opened 2022-04-29
Spawn clients to simulate QPS

mariaschett:


experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -36,12 +39,24 @@ pub struct Opt {
     listen_address: String,
 }
 
+lazy_static! {
+    static ref START_TIME: Instant = Instant::now();
```
tiziano88, line 43: I'm curious whether this actually works? I thought it is initialized the first time it is accessed (hence the laziness), not really at "static initialization" time (which I don't think it's really something that exists). But curious to know either way.

tiziano88: I think START_TIME is only initialized when you first call it, which is also when you call elapsed on it, so I imagine you would get almost zero elapsed here the first time you invoke it. But I see your point, since you are only using it for relative offsets, right?

mariaschett: yes, only relative offsets!

experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -16,13 +16,16 @@
 
 //! Backend server for the Trusted Shuffler example.
 
+use std::time::Instant;
+
```
tiziano88, line 20: remove blank line

experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -36,12 +39,27 @@ pub struct Opt {
     listen_address: String,
 }
 
+lazy_static! {
+    // We use START_TIME to time the time elapsed between every request and response.
```
tiziano88, line 43: Is it though? AFAICT only request times are logged? Is the idea that callers can figure out the elapsed time by comparing the time on the logs?

experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -36,12 +39,27 @@ pub struct Opt {
...
+    static ref START_TIME: Instant = Instant::now();
```
tiziano88, line 44: Does https://doc.rust-lang.org/std/lazy/index.html work instead of lazy_static (which we generally try to avoid)?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -18,9 +18,12 @@
 
 use anyhow::Context;
 use clap::Parser;
+use futures::future::join_all;
 use hyper::Method;
-use log::info;
-use std::time::Instant;
+use primes::{PrimeSet, Sieve};
+use std::time::{Duration, Instant};
+
```
tiziano88, line 25: remove blank line

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
         default_value = "http://localhost:8080"
     )]
     server_url: String,
+    #[structopt(long, help = "The QPS we are aiming to simulate", default_value = "10")]
```
tiziano88, line 38:
```suggestion
    #[structopt(long, help = "The QPS to simulate", default_value = "10")]
```

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+    qps: u32,
+    #[structopt(
+        long,
+        help = "How many rounds, i.e., seconds, we sent requests",
```
tiziano88, line 42: Is each round == 1 second? If so, why? Or is it just a convenient starting point for local testing?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
-        .await
-        .context("Couldn't receive response")?;
-    let request_duration = request_start.elapsed();
-
-    let parsed_response = String::from_utf8(response).context("Couldn't decode response body")?;
-    info!(
-        "Received response: {} in {:?}",
-        parsed_response, request_duration
-    );
+    let expected_qps = opt.qps;
```
tiziano88, line 58: In what sense is this QPS "expected"?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+    let rounds = opt.rounds;
+
+    // Start with giving one second iterations.
+    let start_time = Instant::now();
+
+    // Header of table printed by logger.
+    log::info!("phase,id,request_elapsed_in_ms,response_elapsed_in_ms");
```
tiziano88, line 65: I'm not sure what this means, is the expectation to output something like a CSV? If it's meant to be machine readable, how about outputting it to a separate file as an actual CSV (or perhaps as JSON)? But feel free to follow up in a separate PR (if you think it's useful anyways)

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+            let request_start = start_time.elapsed();
+
+            let url = format!("{}/request", server_url);
+            let response = send_request(&url, Method::POST, request.as_bytes())
+                .await
+                .context("Couldn't receive response");
+
+            let request_duration = start_time.elapsed();
+
+            if let Ok(response) = response {
```
tiziano88, line 89: Why are errors ignored here? It seems something we would like to know about

experimental/trusted_shuffler/backend/src/main.rs
conradgrobler: Could you ellaborate on the reason for using a static instance rather than creating a new `Instant` for every request? I think this might actually be slower due to the locking involved with using a single static instance.

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
-
-    let parsed_response = String::from_utf8(response).context("Couldn't decode response body")?;
-    info!(
-        "Received response: {} in {:?}",
-        parsed_response, request_duration
-    );
+    let expected_qps = opt.qps;
+    let rounds = opt.rounds;
+
+    // Start with giving one second iterations.
```
tiziano88, line 61: AFAICT this const has nothing specific about one second.

experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -36,12 +39,27 @@ pub struct Opt {
...
+
+            let request_start = START_TIME.elapsed();
+            // Currently we assume the backend takes no time, i.e., no time elapsed between request
+            // and response. For consistency with client and server we give the time
+            // twice.
+            log::info!(
+                "backend,{:?},{},{}",
+                body,
+                request_start.as_millis(),
+                request_start.as_millis()
```
ipetr0v, line 60: Why does it print `request_start` 2 times?

experimental/trusted_shuffler/backend/src/main.rs
```diff
@@ -53,9 +71,14 @@ async fn handler(request: Request<Body>) -> Result<Response<Body>, hyper::Error>
     }
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "multi_thread", worker_threads = 2)]
```
ipetr0v, line 74: Do you know if it's possible to specify the number of threads in the binary agrument?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+        // If there are no queries still to fit in this round and there is time left in the round,
+        // we sleep until the second has elapsed.
+        let delta = time_still_left_this_round
+            .checked_div(queries_still_to_fit_this_round)
+            .unwrap_or(
+                Duration::checked_sub(Duration::from_secs(1), time_elapsed_this_round)
+                    .unwrap_or(Duration::from_secs(0)),
+            );
+
+        sleep(delta).await;
```
tiziano88, line 130: Would it work to spawn a task every exactly 1s that essentially performs the body of this loop, instead of trying to calculate deltas? Or alternatively start the 1s sleep in the background at the beginning of this loop, and then await for it here?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+        .take((expected_qps * rounds) as usize)
+        .enumerate()
+    {
+        let server_url = opt.server_url.clone();
+
+        // We currently spawn a new client for every new request. Alternatively we could re-use the
+        // client for every round.
+        clients.push(tokio::spawn(async move {
+            let request = format!("{}", p);
+            let request_start = start_time.elapsed();
```
tiziano88, line 80: Is it intentional that we use the time before spawning the task as start time? Why not create a start time inside the spawned task directly?

experimental/trusted_shuffler/scripts/run
```diff
@@ -2,12 +2,16 @@
 
 readonly SCRIPTS_DIR="$(dirname "$0")"
 
+readonly K=2
+readonly QPS=10
+readonly ROUNDS=2
+
```
tiziano88, line 8: If only we could use a real programming language to coordinate the execution of various things... :upside_down_face:

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+            let response = send_request(&url, Method::POST, request.as_bytes())
+                .await
+                .context("Couldn't receive response");
+
+            let request_duration = start_time.elapsed();
+
+            if let Ok(response) = response {
+                // We don't check whether response is exepected response.
+                let _parsed_response =
+                    String::from_utf8(response).context("Couldn't decode response body");
```
conradgrobler, line 92: Will this be used in future? It seems weird to parse the response and then ignore the result.

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+    let expected_qps = opt.qps;
+    let rounds = opt.rounds;
+
+    // Start with giving one second iterations.
+    let start_time = Instant::now();
+
+    // Header of table printed by logger.
+    log::info!("phase,id,request_elapsed_in_ms,response_elapsed_in_ms");
+
+    let mut clients = vec![];
```
tiziano88, line 67: AFAICT these are not really clients, just futures that terminate once each round trip is done, right?

experimental/trusted_shuffler/client/src/main.rs
```diff
@@ -32,27 +35,106 @@ pub struct Opt {
...
+
+            let url = format!("{}/request", server_url);
+            let response = send_request(&url, Method::POST, request.as_bytes())
+                .await
+                .context("Couldn't receive response");
+
+            let request_duration = start_time.elapsed();
+
+            if let Ok(response) = response {
+                // We don't check whether response is exepected response.
```
tiziano88, line 90: Out of curiosity, why not? It seems you already implemented all the fancy logic with primes and stuff, just to ignore the response? :)

experimental/trusted_shuffler/scripts/run
```diff
@@ -2,12 +2,16 @@
 
 readonly SCRIPTS_DIR="$(dirname "$0")"
 
+readonly K=2
+readonly QPS=10
+readonly ROUNDS=2
```
ipetr0v, line 7: What do `ROUNDS` correspond to?
Does it mean, that if we have 1000 QPS, then 2 rounds will mean 2 milliseconds?
Could we have `SECONDS` instead?

ipetr0v (approved):

experimental/trusted_shuffler/scripts/run
ipetr0v: `xtask` won't be able to spawn it on multiple machines (which is required to reach large QPS numbers for testing)

experimental/trusted_shuffler/client/src/main.rs
ipetr0v: This is already a CSV format
It is needed to generate QPS/latency plots

tiziano88: But isn't it mixed with the logs? Which means that we need to modify the logger (now I see why), and also not log anything else. I think it would be much easier to emit it to a separate machine readable file, instead of logging.

experimental/trusted_shuffler/scripts/run
tiziano88: How does this script spawn on multiple machines? Or will it do it in the future?

experimental/trusted_shuffler/client/src/main.rs
ipetr0v: ah, makes sense

experimental/trusted_shuffler/backend/src/main.rs
tiziano88: Why only 2 threads anyways? It seems that the default config should generally be more optimal anyways: https://docs.rs/tokio/latest/tokio/attr.main.html#multi-threaded-runtime

experimental/trusted_shuffler/client/src/main.rs
tiziano88: I would create an mpsc tokio channel somewhere, and let the tasks write onto it, then collect all the values on the other side in a background task, and write them to a file using some CSV library (also, CSV is a surprisingly nontrivial format to implement, I recommend using a library instead of trying to approximate it with format statements)

experimental/trusted_shuffler/client/src/main.rs
tiziano88: I think the functionality is fine, it's just that the name is not very accurate IMO.

experimental/trusted_shuffler/client/src/main.rs
tiziano88: I was just curious why there is logic to generate some request payload, but it then does not get used for anything. It seems that if we assume that the server is just doing echos, the logic here should be able to know exactly what to expect back and check that.

experimental/trusted_shuffler/backend/src/main.rs
tiziano88: Is part of the problem that server and client logs are being collected in the same "file"? It may be useful to define the CSV schema upfront, so that we know whether we are collecting the appropriate data from the right components. In other words: what do we want to plot eventually?

experimental/trusted_shuffler/scripts/run
tiziano88: I think the fact that 1 round == 1 second should be explained / configured somewhere, it does not follow from the fact that we are using QPS for measuring the "speed".

experimental/trusted_shuffler/backend/src/main.rs
tiziano88: But why not get a new instant every time instead of trying to synchronize on a global instant?

experimental/trusted_shuffler/backend/src/main.rs
tiziano88: It seems simpler to just use the actual local instant everywhere, I mean that's what it does already, right? It gives you the number of seconds since epoch.

experimental/trusted_shuffler/scripts/run
tiziano88: For instance, I think the client should be configurable with either a timeout after which it stops (e.g. "run at 10QPS for 20 seconds", or just run forever until interrupted (e.g. "run at 10QPS until I tell you to stop").

experimental/trusted_shuffler/client/src/main.rs
tiziano88: See my other comment, IMO this should really be "maximum_qps", i.e. it will limit the actual QPS to be no more than that (but it cannot make it go any faster than the underlying machine can actually go)

experimental/trusted_shuffler/client/src/main.rs
mariaschett: Thank you, we changed the output of the program for now, but we'll come back that, when we want csv output from the program. You are right about csv and logging!

experimental/trusted_shuffler/client/src/main.rs
mariaschett: I added a comment clarifying that a client sends only one request and terminates.

experimental/trusted_shuffler/client/src/main.rs
mariaschett: True, I added an assert to check the response is correct.

Issue #2794, opened 2022-04-29
Improve error handling in UEFI/bare-metal runtimes


conradgrobler:
At the moment there are multiple errors that will cause panics, some of which could be influenced by user requests etc. The code should only call panic-inducing functions (e.g. `expect`, `unwrap` etc) in cases where the error is unrecoverable and global. Any errors that are caused by user-specific data or per-request processing must not cause panics. These should either be ignored, or logged if it can be shown that logging the error will not leak data. An error response should be sent back to the client if processing was aborted due to an error.

Issue #2783, opened 2022-04-27
Redesign logging approach


conradgrobler:
At the moment we block all logging from the log crate and use a custom logger that is repeatedly cloned. As part of a redesign we should consider the following features:

- Allow the log crate, but provide a logger that will treat everything from the log crate as sensitive
- Implement support for a per-request logger that collects sensitive logs for a specific request so that it could optionally be returned to the client (based on a flag in the client request) for debugging
- Add support for public logging via a reference rather than cloning the logger

PR #2795, opened 2022-04-29
Add intial wasm support to the UEFI and bare-metal runtimes

conradgrobler:
We can't currently send the Wasm module bytes to the runtime from the loader, so the compiled `echo` Wasm module is embedded into the runtime and loaded from the embedded bytes for now.


oak_functions/wasm/src/lib.rs
```diff
@@ -463,7 +463,12 @@ where
     }
 
     pub fn handle_invoke(&self, request: Request) -> anyhow::Result<Response> {
-        let request_bytes = request.body;
+        let response_bytes = self.handle_raw_invoke(request.body)?;
+        Ok(Response::create(StatusCode::Success, response_bytes))
+    }
```
jul-sh, line 468: Do you have any thoughts whether we plan to eventually align the the two runtimes response handling? I think there's good arguments against using the current runtime's proto wrapper in the UEFI runtime, but also eventual consistency would be nice.

experimental/uefi/runtime/src/framing.rs
```diff
@@ -14,26 +14,13 @@
...
-    // An error occured while deserializing.
-    Deserialization(ciborium::de::Error<T>),
-
-    // An error occured while serializing.
-    Serialization(ciborium::ser::Error<T>),
-
-    // An error occured in remote attestation.
-    Attestation(anyhow::Error),
-}
-
```
jul-sh, line 34: One thing I liked about this struct is facilitating differentiation between types of errors that may either be private or could be forwarded to the client. We can properly tackle that as part of #2783 though, for now using anyhow consistently seems nice. :)

jul-sh (approved):
thank you. lgtm!

experimental/uefi/runtime/src/framing.rs
conradgrobler: I agree with the idea about deciding which error types should be reported to the client, but I don't think this enum provided a meaningful distinction. I agree that this should be addressed in a more systematic way as part of #2783 and #2794.

oak_functions/wasm/src/lib.rs
conradgrobler: I think the plan is to eventually align the two runtimes once we have settled on an approach to framing and an IDL.

conradgrobler (commented):
Thanks for the review.

experimental/uefi/app/src/main.rs
```diff
@@ -24,7 +24,7 @@
 #![test_runner(crate::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 
-use runtime::echo;
+use runtime::framing;
```
andrisaar, line 27: I will have to admit that I'd much rather see the (a) renaming of `echo` to `framing` and (b) refactoring the error handling in a separate PR, but it's not like Github makes that easy.

andrisaar (approved):

PR #2802, opened 2022-05-03
Don't expect floats to be exactly the same

andrisaar:
If you look at, say, https://github.com/project-oak/oak/pull/2801 where Dependabot tries to update the version of `tract-tensorflow`, the tests fail because:
```
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ╔════════════════════════
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║ ════╡ stderr ╞════
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║     Finished release [optimized] target(s) in 0.25s
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║      Running `target/x86_64-unknown-linux-gnu/release/mobilenet_client`
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║ thread 'main' panicked at 'assertion failed: `(left == right)`
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║   left: `"Best result: Some((0.1752324, 789))"`,
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║  right: `"Best result: Some((0.17523259, 789))"`', oak_functions/examples/mobilenet/client/rust/src/main.rs:83:5
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ║ note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[14:56:07 ✓:1,✗:0,⠇:4] │││││││ ╚════════════════════════
```

Expecting floats always to be exactly the same goes against best practice; you should allow some variation. Unfortunately we're comparing strings in the test, so... let's try matching against a regex that allows some variance after the 7th significant digit.

### DIFF
--- a/oak_functions/examples/mobilenet/client/rust/src/main.rs
+++ b/oak_functions/examples/mobilenet/client/rust/src/main.rs
@@ -80,6 +80,8 @@ async fn main() -> anyhow::Result<()> {
         .await
         .context("Could not invoke Oak Functions")?;
 
+    // Allow some small variance in the result, as the number can change slightly between different
+    // versions of libraries. We still want the test to fail for any significant changes, though.
     let response_body = std::str::from_utf8(response.body().unwrap()).unwrap();
     let regex = Regex::new(r"^Best result: Some\(\(0.175232\d+, 789\)\)$").unwrap();
     assert!(
--- a/oak_functions/examples/mobilenet/client/rust/src/main.rs
+++ b/oak_functions/examples/mobilenet/client/rust/src/main.rs
@@ -19,6 +19,7 @@
 use anyhow::Context;
 use oak_functions_abi::proto::{ConfigurationInfo, Request};
 use oak_functions_client::Client;
+use regex::Regex;
 use tract_tensorflow::prelude::*;
 
 // Shape of the input tensor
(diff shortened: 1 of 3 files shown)


jul-sh (approved):
imo we should do this. Worth noting though that I proposed this earlier, when the test failed after an dep updated before, and was advised to keep full precision: http://chat/room/AAAAp6czer4/WE36vadYc-c

oak_functions/examples/mobilenet/client/rust/src/main.rs
```diff
@@ -80,7 +81,13 @@ async fn main() -> anyhow::Result<()> {
...
 
     let response_body = std::str::from_utf8(response.body().unwrap()).unwrap();
-    assert_eq!(response_body, "Best result: Some((0.17523259, 789))");
+    let regex = Regex::new(r"^Best result: Some\(\(0.175232\d+, 789\)\)$").unwrap();
+    assert!(
+        regex.is_match(response_body),
+        "Response \"{}\" does not match regex \"{}\"",
+        response_body,
+        regex.as_str()
+    );
```
jul-sh, line 90: Should we add a comment above the regex explaining the approximate comparison for increased readability?

tiziano88:
> Thanks! imo we should do this. Worth noting though that iirc I proposed this earlier and was advised to keep full precision: http://chat/room/AAAAp6czer4/WE36vadYc-c

FWIW I would still not do it personally, and I would stick to a "golden" expected value that we periodically manually update when necessary, since it also confirms that the output is deterministic (the current regex based approach does not). Anyways if both of you think this is a good idea please go ahead.

tiziano88 (approved):

tiziano88:
> That's the thing, it's _not_ deterministic, as evidenced by the fact that we need to keep updating the value :)

Well, at least we know it is deterministic for each version of the code, right? (but yes, I see your point, no action needed)

PR #2803, opened 2022-05-03
Add an end-to-end test for bare metal/UEFI using xtask

andrisaar:
This comes with multiple caveats:

  * KVM is required, so you can't run it in the VS Code container nor any of our current CI environments
  * Bare-metal binary doesn't build, so that branch is untested (but as we know it'd fail anyway due to ring)

### DIFF
--- a/experimental/uefi/client/src/main.rs
+++ b/experimental/uefi/client/src/main.rs
@@ -27,6 +27,24 @@ struct Args {
     /// address of the server
     #[clap(long, default_value = "http://127.0.0.1:8000")]
     server: String,
+
+    /// message to be sent to the server. If not specified (with a response), uses stdio.
+    #[clap(long, requires = "response")]
+    request: Option<String>,
+
+    /// expected response from the server. If not specified (with a request), uses stdio.
+    #[clap(long, requires = "request")]
+    response: Option<String>,
+}
+
+async fn chat(client: &mut AttestationClient, message: String) -> anyhow::Result<String> {
+    let response = client
+        .send(message.as_bytes().to_vec())
+        .await
+        .context("Error invoking Oak Functions instance")?
+        .ok_or_else(|| anyhow::anyhow!("Empty response"))?;
+
+    String::from_utf8(response).map_err(|err| anyhow::anyhow!(err))
 }
 
 #[tokio::main]
@@ -38,13 +56,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .await
             .context("Could not create client")?;
 
-    for line in stdin().lock().lines() {
-        let response = client
-            .send(line.unwrap().as_bytes().to_vec())
-            .await
-            .context("Error invoking Oak Functions instance")?
-            .ok_or_else(|| anyhow::anyhow!("Empty response"))?;
-        println!("Response: {:?}", core::str::from_utf8(&response));
+    if cli.request.is_some() && cli.response.is_some() {
+        let response = chat(&mut client, cli.request.unwrap()).await?;
+        assert_eq!(response, cli.response.unwrap());
+    } else {
+        for line in stdin().lock().lines() {
+            let response = chat(&mut client, line.unwrap()).await?;
+            println!("Response: {:?}", response);
+        }
     }
     Ok(())
 }
--- a/xtask/src/vm.rs
+++ b/xtask/src/vm.rs
@@ -1,0 +1,108 @@
+//
+// Copyright 2020 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


experimental/uefi/client/src/main.rs
```diff
@@ -27,6 +27,24 @@ struct Args {
...
+}
+
+async fn chat(client: &mut AttestationClient, message: String) -> anyhow::Result<String> {
+    let response = client
+        .send(message.as_bytes().to_vec())
+        .await
+        .context("Error invoking Oak Functions instance")?
+        .ok_or_else(|| anyhow::anyhow!("Empty response"))?;
+
+    String::from_utf8(response).map_err(|err| anyhow::anyhow!(err))
```
tiziano88, line 47: maybe

```suggestion
    String::from_utf8(response).map_err(Into)
```

experimental/uefi/client/src/main.rs
```diff
@@ -38,13 +56,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            .ok_or_else(|| anyhow::anyhow!("Empty response"))?;
-        println!("Response: {:?}", core::str::from_utf8(&response));
+    if cli.request.is_some() && cli.response.is_some() {
+        let response = chat(&mut client, cli.request.unwrap()).await?;
+        assert_eq!(response, cli.response.unwrap());
+    } else {
+        for line in stdin().lock().lines() {
+            let response = chat(&mut client, line.unwrap()).await?;
+            println!("Response: {:?}", response);
+        }
```
tiziano88, line 66: Something like

```rust
    match (cli.request, cli.response) {
        (Some(request), Some(expected_response)) => {
          let response = chat(&mut client, cli.request.unwrap()).await?;
          assert_eq!(response, expected_response);
        },
        _ => {
          for line in stdin().lock().lines() {
              let response = chat(&mut client, line.unwrap()).await?;
              println!("Response: {:?}", response);
          }
        }
```

experimental/uefi/client/src/main.rs
```diff
@@ -27,6 +27,24 @@ struct Args {
...
     #[clap(long, default_value = "http://127.0.0.1:8000")]
     server: String,
+
+    /// message to be sent to the server. If not specified (with a response), uses stdio.
+    #[clap(long, requires = "response")]
+    request: Option<String>,
+
+    /// expected response from the server. If not specified (with a request), uses stdio.
+    #[clap(long, requires = "request")]
+    response: Option<String>,
```
tiziano88, line 37:
```suggestion
    expected_response: Option<String>,
```

xtask/src/vm.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[derive(Debug, Display, Clone, PartialEq, EnumIter)]
+enum Variant {
+    Uefi,
+    Baremetal,
+}
+
+impl Variant {
+    pub fn enabled(&self) -> bool {
+        !matches!(self, Variant::Baremetal)
```
tiziano88, line 32: Does this not work?

```suggestion
        self != Variant::Baremetal
```

xtask/src/vm.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::internal::*;
+
+#[derive(Debug, Display, Clone, PartialEq, EnumIter)]
+enum Variant {
+    Uefi,
+    Baremetal,
+}
+
+impl Variant {
+    pub fn enabled(&self) -> bool {
```
tiziano88, line 31: Does this mean we are only running one variant? Could you add a comment?

tiziano88 (approved):

xtask/src/vm.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    pub fn loader_mode(&self) -> &'static str {
+        match self {
+            Variant::Uefi => "uefi",
+            Variant::Baremetal => "bios",
+        }
+    }
+
+    pub fn binary_path(&self) -> &'static str {
```
conradgrobler, line 49: In future we will want to run multiple examples using each variant as part of CI. It would likely require something similar to the current `example.toml` logic used for Oak Funciotns examples. It might be worth creating an issue and adding a TODO to track that.

conradgrobler (approved):

jul-sh (approved):

jul-sh (approved):

experimental/uefi/client/src/main.rs
andrisaar: error[E0423]: expected value, found trait `Into`

xtask/src/vm.rs
andrisaar: That was a bit too obvious a solution...

PR #2809, opened 2022-05-05
Handle Serialization Errors

### LINKED ISSUE #2819: Return Error when Serialization Fails in Extensions

mariaschett:
Fixes #2819.


oak_functions/abi/src/lib.rs
```diff
@@ -73,6 +73,7 @@ pub struct ReportMetricRequest {
 #[cfg(feature = "oak-metrics")]
 pub enum ReportMetricError {
     ProxyAlreadyConsumed,
+    DeOrSerializingFailed,
```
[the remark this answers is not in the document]
tiziano88: Makes sense, but I think the naming is a bit odd. I haven't looked around at how tarpc or grpc or capnproto or other RPC-like systems do, but I think it would be useful to do that to see if there is any best practice we can learn from them.

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -94,18 +94,26 @@ pub fn report_metric<T: AsRef<str>>(
...
-
-    Ok(response.result)
+    let response = match bincode::serialize(&request) {
+        Ok(serialized_request) => invoke(
+            oak_functions_abi::ExtensionHandle::MetricsHandle,
+            &serialized_request,
+        ),
+        Err(err) => {
+            log!("Failed to serialize request: {}", err);
+            return Ok(Err(ReportMetricError::DeOrSerializingFailed));
```
tiziano88, line 104: I find this a bit surprising here that if we cannot serialize the request, we make it look like the request went through and we got this response back. IMO that obfuscates what actually is going on, i.e. that the request was not even sent at all. Perhaps it should return ERR_INVALID_ARGS at the ABI level? Also, thinking how this would work in other extensions: it would not work to figure out which error variant of each extension corresponds more closely to something about serialization; and in a hypothetical IDL, the serialization logic on the client side will have to be auto generated, so we definitely cannot assume that there is a mapping. FWIW the response of the invocation may not even be a Result at all, in general, so we need to convey the serialization error in some other way.

tiziano88:
Perhaps the fundamental mismatch is that you are trying to hide all the transport errors from the SDK level, but actually I don't think that's possible; the fact that doing an invocation may fail needs to be handled at some point, and I think that requires basically exposing callers to the possibility of failure. The same way as using gRPC would require the caller to deal with those errors somehow.

mariaschett:
I think our last discussion was to have a single `OakStatus::ErrSerializing` to indicate to the client that serialization went wrong. WDYT?

tiziano88:
> I think our last discussion was to have a single `OakStatus::ErrSerializing` to indicate to the client that serialization went wrong. WDYT?

yes, and also avoid "hiding" (or converting) that error in the SDK; just pass it on to the callers

oak_functions/sdk/oak_functions/src/lib.rs
```diff
@@ -90,18 +90,23 @@ pub fn report_metric<T: AsRef<str>>(
     let label = label.as_ref().to_owned();
     let request = ReportMetricRequest { label, value };
 
-    let serialized_request =
-        bincode::serialize(&request).expect("Fail to serialize report metric request.");
+    let serialized_request = bincode::serialize(&request).map_err(|err| {
```
mariaschett, line 93: @tiziano88 Is this where you mean I am "hiding" the error?

If I don't map the err to OakStatus (and below in 103), I'd have change the return type of the function `Result<Result<(), ReportMetricError>, OakStatus>` I believe. I am okay with doing so, but I'd rather do that in a seperate PR.

Is that what you meant? Or am I missing something?

tiziano88: Yes, since that's also what an IDL would generate anyways, right? Well, in fact an IDL would probably not even be able to serialize a "nice" Result as the inner return value, so it will probably be something `Result<Response, OakStatus>`, and somehow `Response` should represent either success or failure (e.g. in protobuf or flatbuffer). But alternatively, we could augment the OakStatus error space to match e.g. the gRPC error space, and allow the extension to return one of those error codes directly (very similar to gRPC itself).

oak_functions/proto/abi.proto
```diff
@@ -39,12 +39,14 @@ enum OakStatus {
...
-  // Arguments invalid.
+  // Arguments invalid, e.g., deserializing the request failed.
   ERR_INVALID_ARGS = 2;
-  // Internal error.
+  // Internal error, e.g., when serialzing the response failed.
   ERR_INTERNAL = 3;
   // Error when the given number is not a valid handle.
   ERR_INVALID_HANDLE = 4;
+  // Error when serialization fails.
+  ERR_SERIALIZING = 5;
```
tiziano88, line 49: Now this one looks off, since the other errors that are not named after serialization do mention serialization as an example, but this one not.

tiziano88 (approved):

oak_functions/proto/abi.proto
```diff
@@ -39,12 +39,14 @@ enum OakStatus {
   UNSPECIFIED = 0;
   // Success.
   OK = 1;
-  // Arguments invalid.
+  // Arguments invalid, e.g., deserializing the request failed.
   ERR_INVALID_ARGS = 2;
-  // Internal error.
+  // Internal error, e.g., when serialzing the response failed.
```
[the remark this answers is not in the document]
mariaschett: About the comma, I think there are different opinions, but I left it out for now. Thank you!

oak_functions/proto/abi.proto
mariaschett: Very good point, fixed!

PR #2810, opened 2022-05-05
Make SerializedRequest an explicit type to improve readability

jul-sh:
This will help when we add a Frame type in #2811, which can also be converted to & from bytes

### DIFF
--- a/experimental/uefi/runtime/src/lib.rs
+++ b/experimental/uefi/runtime/src/lib.rs
@@ -38,7 +38,7 @@ pub struct SerializeableRequest {
     pub session_id: SessionId,
     pub body: Vec<u8>,
 }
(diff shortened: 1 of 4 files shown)


experimental/uefi/runtime/src/lib.rs
```diff
@@ -38,9 +38,10 @@ pub struct SerializeableRequest {
     pub session_id: SessionId,
     pub body: Vec<u8>,
 }
+pub type SerializedRequest = Vec<u8>;
```
tiziano88, line 41: IMO this makes things less readable (since we are really only talking about bytes), but I imagine it will get better again with the framing. But in that case, how will this change? Are you thinking of changing this type to something else then?

PR #2812, opened 2022-05-05
Add RustCrypto implementation of remote attestation

conradgrobler:
This allows us to choose which implementation of the crypto primitives to use for remote attestation.

Only the `ring` implementation is currently tested in CI, but I have manually tested the RustCrypto version. The next step after this PR it to use this RustCrypto implementation for the bare-metal binary. The client will still use the `ring` version, so we will also test interoperability between the two implementations.

### DIFF
--- a/remote_attestation/rust/src/crypto/mod.rs
+++ b/remote_attestation/rust/src/crypto/mod.rs
@@ -19,16 +19,25 @@
 // Should be kept in sync with the Java implementation of the remote attestation
 // protocol.
 
-#[cfg(not(feature = "ring-crypto"))]
-compiler_error!("A cryptographic implementation must be specified.");
+#[cfg(not(any(feature = "ring-crypto", feature = "rust-crypto")))]
+compile_error!("A cryptographic implementation must be specified.");
+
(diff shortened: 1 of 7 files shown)


jul-sh (approved):
nice :)

remote_attestation/rust/src/crypto/mod.rs
```diff
@@ -19,16 +19,25 @@
 // Should be kept in sync with the Java implementation of the remote attestation
 // protocol.
 
-#[cfg(not(feature = "ring-crypto"))]
-compiler_error!("A cryptographic implementation must be specified.");
+#[cfg(not(any(feature = "ring-crypto", feature = "rust-crypto")))]
+compile_error!("A cryptographic implementation must be specified.");
+
+// When all both implementations are selected (e.g. when testing with all features) we use the
```
andrisaar, line 25: nit: remove either "all" or "both"

andrisaar (approved):

remote_attestation/rust/src/crypto/mod.rs
conradgrobler: Thanks, done.

PR #2816, opened 2022-05-06
Enable bare-metal e2e test

conradgrobler:
Enabling the tests required a few fixes:

- Patching `curve25519-dalek` and `x25519-dalek` to use a newer version of `rand_core` so that it could be built for `x86_64_unknown_none`
- `sha2` could not be built without SIMD support
- The baremetal binary had to enable SSE and AVX support on the CPU via control registers to allow execution of SIMD code
- Rust assumes that the stack is 16-byte aligned, but the bare metal boot code caused it to be misaligned by 8 bytes because of the way it jumps into the entry point for Rust code
- The serial communication code was sending bytes as if it were text-based rather than raw binary, so changed to `send_raw`

### DIFF
--- a/experimental/uefi/baremetal/src/main.rs
+++ b/experimental/uefi/baremetal/src/main.rs
@@ -61,7 +61,9 @@ fn main(info: &dyn boot::Info) -> ! {
     runtime::framing::handle_frames(serial).unwrap();
 }
 
-/// Enables Advanced Vector Extensions (AVX)
+/// Enables Streaming SIMD Extensions (SEE) and Advanced Vector Extensions (AVX).
+///
+/// See https://wiki.osdev.org/SSE for more information.
 fn enable_avx() {
     unsafe {
         let mut cr0 = Cr0::read();
--- a/experimental/uefi/baremetal/src/main.rs
+++ b/experimental/uefi/baremetal/src/main.rs
@@ -28,14 +28,20 @@ mod logging;
 mod memory;
 mod serial;
 
+extern crate alloc;
 #[macro_use]
 extern crate log;
 
 use core::panic::PanicInfo;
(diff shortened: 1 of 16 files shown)


experimental/uefi/baremetal/src/main.rs
```diff
@@ -55,6 +61,24 @@ fn main(info: &dyn boot::Info) -> ! {
...
+        Cr0::write(cr0);
+        let cr0 = Cr0::read();
+        assert!(cr0 & Cr0Flags::TASK_SWITCHED != Cr0Flags::TASK_SWITCHED);
+        let mut cr4 = Cr4::read();
+        cr4 |= Cr4Flags::OSFXSR | Cr4Flags::OSXMMEXCPT_ENABLE | Cr4Flags::OSXSAVE;
+        Cr4::write(cr4);
+        let mut xcr0 = XCr0::read();
+        xcr0 |= XCr0Flags::X87 | XCr0Flags::SSE | XCr0Flags::AVX;
+        XCr0::write(xcr0);
+    }
```
tiziano88, line 79: Where does this come from? Could you add a reference?

xtask/src/vm.rs
```diff
@@ -29,9 +29,10 @@ enum Variant {
 
 impl Variant {
     pub fn enabled(&self) -> bool {
-        // The bare metal variant is disabled for now, because the crypto library we use right now
-        // doesn't work on bare metal.
-        self != &Variant::Baremetal
+        match self {
+            Variant::Baremetal => true,
+            Variant::Uefi => true,
+        }
```
tiziano88, line 35: Looks like this method can be removed now?

tiziano88 (approved):

conradgrobler (commented):
Thanks.

PR #2821, opened 2022-05-06
Run VM test under Kokoro

andrisaar:
This was easier than expected, but not without its warts: namely, on Kokoro machines `/dev/kvm` was owned by `root:root`. However, that's nothing a quick `sudo chown` can't fix.

You can see the test logs if you go look at the Kokoro output -- unfortunately right now it requires corp credentials.

Steps for the future: (a) make the test run in Kokoro `prod`, not `qa` and (b) look into Resultstore so that we could make the build logs available without requiring Google corp credentials.


jul-sh (approved):
lgtm :]

conradgrobler (approved):

Issue #2732, opened 2022-04-19
Improve or disable fuzzing in CI


tiziano88:
Currently whenever dependencies are updated, the generated Wasm file that is used as part of fuzzing gets outdated, and needs to be manually updated and checked in. This makes automatic dependency management by @dependabot very inconvenient, since most of those auto generated PRs fail because they (rightly) are not aware of the wasm file that needs to be updated.

Ideally we should change the fuzzing logic so that the wasm file is actually compiled on the fly as part of fuzzing, instead of being checked in as a binary artifact.

In the meanwhile, I suggest we remove fuzzing from CI, since I don't think it has been particularly helpful so far. But if anyone has any other suggestion, please let me know here.

rbehjati:
I don't think so. I think we should disable it, or at least try and solve the issue with `fuzzable.wasm`. I can spend some time on it next week and see if I can find a work around for `fuzzable.wasm`.

rbehjati:
Closing this for now.

PR #2829, opened 2022-05-09
Remove wasm_invoke

rbehjati:
Ref #2732

Removes `wasm_invoke` and the `fuzzable` module, and the functionality in `xtask` for building the wasm target. Also, removes the `--build-deps` flag from the `run-cargo-fuzz` command.


rbehjati:
I experimented a bit more trying to fix the path (as suggested in https://github.com/google/oss-fuzz/issues/7690), but could not make the `run_fuzzer` command to build the wasm module at runtime. So this PR now removes `wasm_invoke` and the `fuzzable` module. 

The error message when running `python infra/helper.py run_fuzzer oak wasm_invoke` is:

```
manifest_path: "/out/oak_functions/loader/fuzz/fuzzable/module/Cargo.toml"
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })', /src/oak/oak_functions/sdk/test_utils/src/lib.rs:60:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
AddressSanitizer:DEADLYSIGNAL
```

But `/out/oak_functions/loader/fuzz/fuzzable/module/Cargo.toml` is a valid path inside the container. I don't quite understand where the "No such file or directory" error comes from.

ipetr0v (approved):

jonathanmetzman:
> ```
> 
> ```
> 
> 
>     
> 
>   
> 
> But `/out/oak_functions/loader/fuzz/fuzzable/module/Cargo.toml` is a valid path inside the container. I don't quite understand where the "No such file or directory" error comes from

Ah this failed because of an intentional safeguard I added. Nice! (sorry it made your life harder, we move /out to /tmp/not-out temporarily).
Essentially, we don't want you hardcoding any path to /out because although that path is available in OSS-Fuzz's runner and builder images, it is not on the images run on ClusterFuzz. I think the best way to find a path relative to your binary is via argv[0] which can be gotten using LLVMFuzzerInitialize (https://llvm.org/docs/LibFuzzer.html#startup-initialization) or `readlink /proc/self/exe`

rbehjati:
I tried `readlink /proc/self/exe` (actually `std::fs::read_link("/proc/self/exe")` in the Rust fuzz target), but am still getting "No such file or directory". The fuzz target that I removed in this PR was not very effective actually, and we have thought about writing another one for fuzzing the ABI. With that we would still need to check the wasm module in the Git repository, but it would not be affected by other changes that much.

PR #2836, opened 2022-05-10
Make the baremetal binary use Linux boot protocol

andrisaar:
This makes it compatible with crosvm. Seems to work, but the loader doesn't obviously know anything about crosvm, yet.


conradgrobler:
I fixed a few issues that caused the e2e tests to fail in https://github.com/project-oak/oak/pull/2816. Some of these were in code deleted here, so it might require porting them here.

PR #2843, opened 2022-05-13
Change frame length prefix to match planned framing

jul-sh:
Increases its size. Changes it from denoting just the length of body to the length of the whole frame (including the length prefix itself and any other future header values).

### DIFF
--- a/experimental/uefi/runtime/src/lib.rs
+++ b/experimental/uefi/runtime/src/lib.rs
@@ -34,9 +34,9 @@ pub trait Channel {
     fn recv(&mut self, data: &mut [u8]) -> anyhow::Result<()>;
 }
 
-pub type FrameLength = u16;
-// The frame length is a u16, which is two bytes encoded.
-pub const FRAME_LENGTH_ENCODED_SIZE: usize = 2;
+/// Length of the entire frame, including the header.
+pub type FrameLength = u32;
+const FRAME_BYTES_SIZE: usize = 4;
 
 pub struct Frame {
     pub body: Vec<u8>,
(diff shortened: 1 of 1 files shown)


experimental/uefi/runtime/src/lib.rs
```diff
@@ -52,24 +52,36 @@ impl<T: Channel> Framed<T> {
...
-        encoded_frame.extend(frame.body);
-        self.inner.send(&encoded_frame)?;
+        let length_bytes = {
+            let length = FrameLength::try_from(FRAME_BYTES_SIZE + frame.body.len())
+                .map_err(anyhow::Error::msg)
+                .context("the frame is too large")?;
+            length.to_be_bytes()
+        };
+        self.inner.send(&length_bytes)?;
+        self.inner.send(&frame.body)?;
```
jul-sh, line 84: Calling send in two increments avoids the extra allocation that sometimes occurred in the previous approach using extend

tiziano88:
> Increases its size. Changes it from denoting just the length of body to the length of the whole frame (including the length prefix itself and any other future header values).

AFAICT commonly length delimited framing does not include the length of the prefix itself, just the length of the payload (e.g. see https://docs.rs/tokio-util/latest/tokio_util/codec/length_delimited/index.html#getting-started , https://seb-nyberg.medium.com/length-delimited-protobuf-streams-a39ebc4a4565). The idea is that if someone is reading this from a stream, they can read the prefix (fixed size), interpret it as an integer N, and then read N more bytes, to complete the frame.

tiziano88:
OTOH, I see that IP does include everything in the length field: https://en.wikipedia.org/wiki/IPv4#Total_Length

jul-sh:
As another counter example, in the UDP spec the length describes the length of the entire frame https://www.ietf.org/rfc/rfc768.txt

Don't feel super strongly about this, but in my mind specifying the whole length makes things a little bit easier to reason about when we'll add additional headers

experimental/uefi/runtime/src/lib.rs
```diff
@@ -52,24 +52,36 @@ impl<T: Channel> Framed<T> {
...
-        self.inner.recv(&mut body)?;
+        let length = {
+            let mut length_bytes = [0; FRAME_BYTES_SIZE];
+            self.inner.recv(&mut length_bytes)?;
+            FrameLength::from_le_bytes(length_bytes)
+        };
+
+        let body = {
+            let body_length: u32 = length - FRAME_BYTES_SIZE as u32;
+            let mut body: Vec<u8> = vec![
```
conradgrobler, line 63: Does it have to be initialised to 0, or is `Vec::with_capacity` sufficeiont here?

conradgrobler (approved):
Thanks.

experimental/uefi/runtime/src/lib.rs
jul-sh: Unfortunately it does need to be initialised, since our receive function fills the length of vector.

PR #2846, opened 2022-05-13
Endorsement

rbehjati:


oak_functions/client/rust/src/tests.rs
```diff
@@ -32,17 +32,26 @@ fn test_verify_rekor_log_entry() {
     // `bb05be1bd813f8afb7b77b2d9f7be5ae25b396d111c7a26a04b785c48c277372`.
     let log_entry_path = "testdata/logentry.json";
 
+    // Public key of the product team that signed the endorsement file.
```
mariaschett, line 35: Small clarification: is the product team relevant here? We are just using any test key, or?

oak_functions/client/rust/src/tests.rs
```diff
@@ -32,17 +32,26 @@ fn test_verify_rekor_log_entry() {
...
+
     // Public key of the Rekor instance hosted by sigstore.dev. It is downloaded from https://rekor.sigstore.dev/api/v1/log/publicKey.
     let rekor_pubkey_path = "testdata/rekor_public_key.pem";
 
     let endorsement_bytes = fs::read(endorsement_path).expect("Couldn't read endorsement file.");
     let log_entry_bytes = fs::read(log_entry_path).expect("Couldn't read log entry file.");
     let rekor_pem_bytes =
         fs::read(rekor_pubkey_path).expect("Couldn't read Rekor's public key file.");
+    let pubkey_pem_bytes =
+        fs::read(pubkey_path).expect("Couldn't read product team's public key file.");
```
mariaschett, line 46: Maybe "Couldn't read public key file."?

oak_functions/client/rust/src/rekor.rs
```diff
@@ -20,19 +20,23 @@
 use anyhow::Context;
 use ecdsa::Signature;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use signature::Verifier;
-use std::str::FromStr;
+use std::{cmp::Ordering, str::FromStr};
+
+/// Number of bytes required for storing a SHA256 hash.
+pub const SHA256_HASH_LENGTH: usize = 32;
```
mariaschett, line 28: nit: we only use it once. Do we need a `pub` constant? I think we could also just inline and document where used.

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
     pem_encoded_public_key_bytes: &[u8],
 ) -> anyhow::Result<()> {
     let signature_bundle = rekor_signature_bundle(log_entry_bytes)?;
+
+    verify_signature(
+        &signature_bundle.base64_signature,
+        &signature_bundle.canonicalized,
+        pem_encoded_public_key_bytes,
+    )
+    .context("failed to verify signedEntryTimestamp of the Rekor LogEntry")
```
mariaschett, line 215: nit: failed -> Failed and add . at the end. Also for error messages below, but I am not 100% sure what our convention is.

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+    let endorsement_hash_hex = hex::encode(endorsement_hash);
+    if endorsement_hash_hex != body.spec.data.hash.value {
+        anyhow::bail!(
+            "the hash of the endorsement file ({:?}) does not match the hash of the data in the body of the rekor entry ({:?})",
+            endorsement_hash_hex,
+            body.spec.data.hash.value
+        )
+    }
+
+    // Check that the public key in the body matches the given public key. This in fact checks the
```
mariaschett, line 253: clarification: the "given public key". I see it is the argument of `verify_rekor_body`, but where did we actually get the key from in the first place?

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+    // consistency of the Rekor LogEntry, and we expect these public keys to always be the same.
+    let public_key_bytes = base64::decode(body.spec.signature.public_key.content.as_bytes())
```
mariaschett, line 255: Maybe rename to "public_key_from_rekor" and "public_key_from_?".

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+        .expect("failed to base64-decode the public key bytes in the Rekor LogEntry body");
+    if compare_keys(&public_key_bytes, pem_encoded_public_key_bytes)? != Ordering::Equal {
+        anyhow::bail!(
+            "the given public key is different from the public key in the rekor entry: {:?} vs. {:?}",
```
mariaschett, line 259: nit: would you need `unmarshal_pem_to_p256_public_key` for the output in `{:?}` to be meaningful?

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+        endorsement_bytes,
+        pem_encoded_public_key_bytes,
+    )
+    .context("failed to verify signature over the endorsement file")
+}
+
+/// Verifies the given base64-encoded signature over the given data bytes, using the given
+/// PEM-encoded public key.
+///
+/// Returns `Ok(())` if the verification succeeds, otherwise returns `Err()`.
```
mariaschett, line 276: Just to throw it out there: you could change the return to Result<bool> and return false if the last step (`verify`) did not succeed. (because I kind of expected verify to return a bool, but I understand your choice, too)

mariaschett (approved):
Very cool, thank you!

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+
+/// Verifies the signature in the `body` over the `endorsement_bytes`, using the public key in
+/// `pem_encoded_public_key_bytes`.
+///
+/// Returns `Ok(())` if the verification succeeds, otherwise returns `Err()`.
+pub fn verify_rekor_body(
+    body: &Body,
+    endorsement_bytes: &[u8],
+    pem_encoded_public_key_bytes: &[u8],
+) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
tiziano88: Another thought: perhaps, in the spirit of https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/ , this function could actually return a struct containing the public key (validated against the signature and the content), and the hash of the data, something like `ValidatedRekorEntry` or similar. Though I'm not actually sure that's possible given that it looks like the verification of the signature requires the actual source data (not just the hash).

oak_functions/client/rust/src/tests.rs
rbehjati: I use "product team" to contrast it with the Rekor public key. Updated the comment to be clear it is a test public key.

oak_functions/client/rust/src/tests.rs
rbehjati: Two public keys are used in this test scenario. I want to distinguish between them in the error message.

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+            pem_encoded_public_key_bytes,
+        )
+    }
+
+    verify_signature(
+        body.spec.signature.content.as_str().as_bytes(),
+        endorsement_bytes,
+        pem_encoded_public_key_bytes,
+    )
+    .context("failed to verify signature over the endorsement file")
```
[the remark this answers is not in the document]
rbehjati: I could not find specific logic for this in Go, perhaps because the Go code is meant to be generic and support various signing formats. The closest thing to it is `ecdsa.VerifyASN1`, but without the base64 decoding.

rbehjati (commented):
Thanks for the reviews.

oak_functions/client/rust/src/rekor.rs
tiziano88: FWIW I think we should use consistent format for all errors, regardless of whether they are final or not. But we can discuss that separately.

oak_functions/client/rust/src/rekor.rs
```diff
@@ -159,10 +206,85 @@ pub fn verify_rekor_signature(
...
+}
+
+/// Verifies the given base64-encoded signature over the given data bytes, using the given
+/// PEM-encoded public key.
+///
+/// Returns `Ok(())` if the verification succeeds, otherwise returns `Err()`.
+pub fn verify_signature(
+    base64_signature_bytes: &[u8],
+    content_bytes: &[u8],
```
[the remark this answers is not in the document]
tiziano88: if it's hashedrekord then I don't think we actually need the content bytes here, but I don't remember the details.

tiziano88 (approved):

oak_functions/client/rust/src/rekor.rs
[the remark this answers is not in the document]
rbehjati: Do you mean that is should be the hashed content instead? I am actually a bit confused myself now.

PR #2849, opened 2022-05-13
Move communication channel logic to its own crate

### LINKED ISSUE #2844: Split `Framing` and `Channel` out of the Uefi/bare-metal runtime
At the moment the loader depends on the entire runtime, meaning that it has a dependency on e.g. Wasmi. This should not be needed. The framing-related code should be moved to a separate crate that can be used by the runtime and the loader.

jul-sh:
Addresses & closes #2844 :)


experimental/uefi/channel/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+                .context("the frame is too large")?;
+            length.to_le_bytes()
+        };
+        self.inner.write_all(&length_bytes)?;
+        self.inner.write_all(&frame.body)?;
+        Ok(())
+    }
+}
+
+pub struct SerializeableRequest {
```
conradgrobler, line 82: I don't think this should be moved here, as it means that the channel crate depends on remote attestation and crypto libraries. I think it is fine to leave it in the runtime for now, and later when we add more serialisable objects (or add full-fledged IDL support) move this elsewhere.

A side note: I don't think the definition of the session ID should be in a crate that forces a dependency on crypto libraries, so as part of the future move to an IDL-related crate the dependency should probably be reversed (i.e. SessionId should be defined in the new crate and remote_attestation_session should depend on it)

experimental/uefi/channel/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "channel"
+version = "0.1.0"
+authors = ["Juliette Pretot <julsh@google.com>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[features]
```
conradgrobler, line 8: As mentioned in the other comment, this crate would not require features if the SerializableRequest is not moved here.

conradgrobler (approved):
Thanks :)

experimental/uefi/channel/src/lib.rs
conradgrobler: Thinking about it a bit more, moving this struct to the `remote_attestation_sessions` crate might be useful as an initial step, seeing that this is tightly coupled with attestation sessions. It would hopefully remove the need for the loader to depend on the runtime. The runtime would still depend on the crypto libraries (which it should not need) for now, but that can be resolved later when moving this to an IDL-related crate.

conradgrobler: Clarity of the mental model, which you mentioned is probalby the most important, especially to simplify the review process for external reviewers.

Clearer separation of dependencies and reposnsibilities is usually a desirable property. if the loader/launcher (running outside the enclave) depends on the runtime (running inside the enclave) it would need to be rebuilt even if an unrelated part of the runtime changed. It also inherits a bunch of other dependencies, which are built but unused. Having more indirect dependencies also complicates the review process, as the reviewer needs to figure out which indirect dependencies are actually used.

We also want to start adding claims (in a human readable format for reviewers) to explain which properties are enforced/ensured by each crate, how this contributes to the overall security/privacy, story and how this is achieved in the code. This becomes much more difficult if there are long chains of indirect depenencies or complex crates doing multiple different things. Every time a dependency changes these need to be reviewed to ensure they are still valid.

No_std support could also be affected by some indirect dependencies, e.g. if the dependency chain was the other way around some std-related features enabled for the loader could stop the runtime from building for a bare-metal target. Porting the runtime to no_std is significantly simplified by the earlier work of splitting components into smaller crates with fewer dependencies.

PR #2850, opened 2022-05-13
Update base docker image and install flatc

tiziano88:
The distribution is now bullseye, and we don't need to use backports any
more.

Flatc is the flatbuffer compiler, which may be used to prototype an IDL
for communication between trusted and untrusted components, and possibly
  for Wasm. See https://github.com/google/flatbuffers


mariaschett (approved):
LGTM. (oh, but not to our CI it seems)

PR #2860, opened 2022-05-16
Free disk space in GitHub actions

tiziano88:
Revert bdeaee31a26cb7f0984e5dfe58b5688aabd48c33 , since it seems that
we are again often running out of space.


rbehjati (approved):
Will we still need `CACHE_VERSION` with this change?

PR #2865, opened 2022-05-17
Shut down bare-metal code via i8042 when we panic.

andrisaar:
i8042 is a fun little device: nominally a "keyboard controller" handling interfacing with PS/2 keyboards, it has over time acquired critical features like handling the A20 gate (so a hack controls a hack...) and most interestingly for us, it allows us to reset the system by writing to a magic port.

This device is supported by both qemu and crosvm, and thus will make the shutdown more generic (right now we use the qemu-specific port to force an exit).

The downside is that we won't be able to return a status code; that's a feature only provide by the qemu-specific port. Thus, I've left the test code alone for now; we need to figure out how to handle testing in the crosvm future anyway as crosvm is not compatible with github actions.


conradgrobler (approved):
Nice.

PR #2867, opened 2022-05-17
Create a generic Vmm trait wrapper in loader

andrisaar:
This is in preparation of introducing a second type of VMM: crosvm.

UEFI will still use qemu, but baremetal will use crosvm.

Right now some changes are a bit silly: for example, both the `Uefi` and `Bios` branches are pretty much equivalent; this is intentional, as this means once we enable crosvm it's just a simple change to the `Bios` branch.

### DIFF
--- a/experimental/uefi/loader/src/vmm.rs
+++ b/experimental/uefi/loader/src/vmm.rs
@@ -20,11 +20,19 @@ use std::{os::unix::net::UnixStream, path::PathBuf};
 
 #[derive(Debug)]
 pub struct Params {
+    /// Path to the VMM binary to execute.
     pub binary: PathBuf,
+
+    /// Optional path to the firmware blob to pass to the VMM.
     pub firmware: Option<PathBuf>,
+
+    /// Path to the binary to load into the VM.
     pub app: PathBuf,
 
+    /// Stream to connect to the console of the VM.
     pub console: UnixStream,
+
+    /// Stream to use for communicating with the app inside the VM.
     pub comms: UnixStream,
 }
 
--- a/experimental/uefi/loader/src/vmm.rs
+++ b/experimental/uefi/loader/src/vmm.rs
@@ -1,0 +1,35 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


experimental/uefi/loader/src/vmm.rs
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+//
+
+use anyhow::Result;
+use async_trait::async_trait;
+use std::{os::unix::net::UnixStream, path::Path};
+
+#[derive(Debug)]
+pub struct Params<'a> {
+    pub binary: &'a Path,
```
tiziano88, line 23: May be more simpler to use https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html instead of carrying around a lifetime for what are basically strings.

tiziano88 (commented):
drive-by comment

experimental/uefi/loader/src/vmm.rs
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use anyhow::Result;
+use async_trait::async_trait;
+use std::{os::unix::net::UnixStream, path::PathBuf};
+
+#[derive(Debug)]
+pub struct Params {
```
mariaschett, line 22: Do we want to call them `VmmParams` similar to `QuemuParams` before?

mariaschett, line 22: I think it would be good to have some /// documentation for the fields---as they are public. But that might also be pre-mature?

andrisaar: I think just `Params` is fine, but meh.

andrisaar: Good idea, done.

experimental/uefi/loader/src/vmm.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Stream to connect to the console of the VM.
+    pub console: UnixStream,
+
+    /// Stream to use for communicating with the app inside the VM.
+    pub comms: UnixStream,
+}
+
+#[async_trait]
+pub trait Vmm {
+    async fn wait(&mut self) -> Result<std::process::ExitStatus>;
```
mariaschett, line 41: Maybe I am missing something, but I think you could make `wait` and `kill`, `pub` here (if that is important to you).

andrisaar: IIRC the compiler yelled at me when I tried doing that, stating that `pub` already comes from the fact that it's a `pub trait`.

PR #2872, opened 2022-05-19
Add virtqueue implementation

conradgrobler:
This adds a simple virtqueue implementation for use by a virtio vsock or console device.


experimental/virtio/src/queue.rs
```diff
@@ -0,0 +1,606 @@
...
+    flags: RingFlags,
+    /// The next index that will be used in the ring (modulo `QUEUE_SIZE`).
+    ///
+    /// Starts at 0 and always increments. Must never be decremented.
+    idx: u16,
+    /// The ring-buffer containing the used elements.
+    ring: [UsedElem; QUEUE_SIZE],
+}
+
+impl<const QUEUE_SIZE: usize> Default for UsedRing<QUEUE_SIZE> {
```
andrisaar, line 128: As I understand it this pretty much zero-initializes the data structure (no flags, and all `UsedElem`s are empty). Would `#[derive(Default)]` not work on `UsedRing`?

experimental/virtio/src/queue.rs
```diff
@@ -0,0 +1,606 @@
...
+
+/// The intended use for the queue.
+#[derive(PartialEq, Eq)]
+enum QueueType {
+    DeviceWriteOnly,
+    DriverWriteOnly,
+}
+
+/// A container for a boxed virtqueue and its associated buffers.
+pub struct Queue<const QUEUE_SIZE: usize, const BUFFER_SIZE: usize> {
```
andrisaar, line 172: Given that github is already complaining about the size of this file, maybe split the file in two: one that contains all the virtio data structures from the spec, and one that just implements the `Queue`.

experimental/virtio/src/queue.rs
```diff
@@ -0,0 +1,606 @@
...
+    ///
+    /// If a used buffer is found, this also advances the last used index by one.
+    ///
+    /// Note: we can only read data from a device-write-only queue.
+    pub fn read_next_used_buffer(&mut self) -> Option<Vec<u8>> {
+        assert!(
+            self.queue_type != QueueType::DriverWriteOnly,
+            "Cannot read from a driver-write-only queue."
+        );
+        if let Some(UsedElem { id, len }) = self.get_next_used_element() {
```
andrisaar, line 299: would `let UsedElem{id, len} = self.get.next_used_element()?;` work here?

experimental/virtio/src/queue.rs
conradgrobler: Unfortunatley not, as Default is not implemented for const generically sized arrays, only array with fixed sizes up to 32. So I have to manually implement it for any structs that have arrays with sizes based on const generics.

experimental/virtio/src/queue.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// The indices of free descriptors that can be used by the driver for writing.
+    free_ids: VecDeque<u16>,
+}
+
+impl<const QUEUE_SIZE: usize, const BUFFER_SIZE: usize> Queue<QUEUE_SIZE, BUFFER_SIZE> {
+    /// Creates a new queue where the descriptor buffers are only writable by the driver.
+    ///
+    /// This is typically used for transmission queues.
+    pub fn new_driver_write_only() -> Self {
+        let mut result = Self::new(QueueType::DriverWriteOnly);
```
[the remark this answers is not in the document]
conradgrobler: I added to two wrapper types. It feels a bit clunky because of having to expose the inner queue as public to avoid a lot of code duplication, but at least the asserts are no longer required.

experimental/virtio/src/queue.rs
conradgrobler: Thanks, done.

andrisaar (approved):

PR #2873, opened 2022-05-19
Update frame length implementation to remove extranous constant

jul-sh:
The unsigned integer type primitive we has a stable size exposed by Rust's core::mem::size_of function. We can use that instead a constant, removing a piece of duplicate information.

### DIFF
--- a/experimental/uefi/channel/src/lib.rs
+++ b/experimental/uefi/channel/src/lib.rs
@@ -22,16 +22,16 @@ extern crate alloc;
 use alloc::{vec, vec::Vec};
 use anyhow::Context;
 use ciborium_io::{Read, Write};
+use core::mem::size_of;
 
 /// Length of the entire frame, including the header.
 pub type FrameLength = u32;
-const FRAME_BYTES_SIZE: usize = 4;
 
 pub struct Frame {
     pub body: Vec<u8>,
 }
 
-pub struct Framed<T: Read + Write> {
+pub struct Framed<T: Read<Error = anyhow::Error> + Write<Error = anyhow::Error>> {
     inner: T,
 }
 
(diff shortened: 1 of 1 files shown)


jul-sh:
@tiziano88 Thanks for chiming in. I agree for the most part, in fact documenting the spec separately is the goal in: https://docs.google.com/document/d/1Q4u5hYzOFF74_RfwZzpRWgSSwJwfQ-GXRvNllgCXOO8/edit?usp=sharing. Moving it to a markdown file once finalized seems like a good idea too. :) 

It might still be nice to remove the extra constants in the rust file then though, since the rust code wouldn't be "source of truth" for the spec anyways. I personally find that it would just make the code slightly easier to reason about, but also don't feel precious about it.

tiziano88:
> @tiziano88 Thanks for chiming in. I agree for the most part, in fact documenting the spec separately is the goal in: https://docs.google.com/document/d/1Q4u5hYzOFF74_RfwZzpRWgSSwJwfQ-GXRvNllgCXOO8/edit?usp=sharing. Moving it to a markdown file once finalized seems like a good idea too. :)
> 
> It might still be nice to remove the extra constants in the rust file then though, since the rust code wouldn't be "source of truth" for the spec anyways. I personally find that it would just make the code slightly easier to reason about, but also don't feel precious about it.

The best of both worlds would may be to have both the constant in the Rust code (so that it can be eyeball-matched by reading the spec), and also an assertion (again in Rust) to ensure that the size of the Frame is exactly the same as that (so that e.g. if someone adds a field to the Frame struct, the assertion would fail). Maybe something like https://docs.rs/static_assertions/latest/static_assertions/macro.assert_eq_size.html can help? It may have to be within a test though I imagine.

But in the code (i.e. in the Rust impl) I would still suggest to use the literal constant (coming from the spec, and in fact linking to the exact part of the spec that defines that) so that it's easy for someone who reads the implementation to ensure that it conforms to the spec.

jul-sh:
@tiziano88 Fair point, thanks for explaining. Let's keep the constants then :)

PR #2881, opened 2022-05-20
Oak IDL based on flatbuffers

tiziano88:
The goal of the IDL is to be used both as part of the communication
channel between trusted and untrusted runtime, as well as for the
communication between Wasm modules and the trusted runtime.

Create the following new crates:

- `oak_idl_gen_services`
- `oak_idl_gen_structs`
- `oak_idl_gen_tests`

Align the existing `oak_idl` crate to use flatbuffers via the
aforementioned crates.

Import flatbuffers reflection schema in `third_party`.

Remove unused deps throughout to make cargo udeps and cargo deny happy.

Ref https://github.com/project-oak/oak/issues/2500


oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        .intersperse("\n".to_string())
+        .collect())
+}
+
+/// Generate the Rust objects from the input [`Service`] instance, corresponding to an `rpc_service`
+/// entry.
+fn generate_service(service: &Service) -> anyhow::Result<String> {
+    let mut lines = Vec::new();
+    lines.extend(vec![
+        format!("// Generated file, do not edit."),
```
mariaschett, line 139: I think we could add for which service the file is generated, how it is generated, and a link to how it might be used?

That might help with interpreting.

Related: do generated files need a license?

oak_idl_tests/tests/oak_abi.rs
```diff
@@ -0,0 +1,78 @@
...
+    #![allow(clippy::derivable_impls, clippy::needless_borrow)]
+    #![allow(dead_code, unused_imports)]
+
+    include!(concat!(env!("OUT_DIR"), "/oak_generated.rs"));
+    include!(concat!(env!("OUT_DIR"), "/oak_services.rs"));
+}
+
+struct OakImpl;
+
+/// An implementation of the [`Oak`] service trait for testing.
```
mariaschett, line 29: We are testing the IDL and nothing from Oak Functions, no?

I think it would help me to emphasize that this is a totally independent test. (If that is the case, oc).

oak_functions/sdk/oak_functions/Cargo.toml
```diff
@@ -5,23 +5,21 @@ authors = ["Razieh Behjati <razieh@google.com>"]
 edition = "2021"
 license = "Apache-2.0"
 
+[dependencies]
+bincode = "*"
```
conradgrobler, line 9: Is the use of bincode and prost here only temporary untill everything is converted to flatbuffers, or do we plan on having them around longer term?

oak_idl/src/lib.rs
```diff
@@ -114,7 +118,7 @@ impl Transport for Channel {
         let request_bytes =
             bincode::serialize(&request_message).map_err(|_| TransportError::InvalidMessage)?;
```
conradgrobler, line 119: Is the plan to convert this to flatbuffers later, or will these be deleted later, or maintained in parallel?

mariaschett (commented):
Wow, nice, but I'll definitely have to look through this again. In the mean time some comments!

oak_idl/src/lib.rs
```diff
@@ -125,97 +129,117 @@ impl Transport for Channel {
...
+/// #     }
+/// # }
+/// #
+/// # impl Foo {
+/// #     pub fn create(b: &mut flatbuffers::FlatBufferBuilder) -> flatbuffers::WIPOffset<Foo> {
+/// #         flatbuffers::WIPOffset::new(0)
+/// #     }
+/// # }
+/// #
+/// let mut b = oak_idl::MessageBuilder::<Foo>::default();
```
conradgrobler, line 162: It is perhaps because I have not used flatbuffers, but this example code does not feel like it adds much clarity.

oak_idl/src/lib.rs
```diff
@@ -125,97 +129,117 @@ impl Transport for Channel {
...
+/// let v = b.create_vector::<u8>(&[14, 12]);
+/// let foo = Foo::create(&mut b);
```
conradgrobler, line 164: Is this a limitation of flatbuffers that the offset must be created separately? Is it not possible to have a trait that the type must implement so that it can be automatically called by `finish`.

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    qualified_name
+        .split('.')
+        .last()
+        // If split returned an empty iterator, then the input name did not contain any dots, so we
+        // can return it as-is.
+        .unwrap_or(qualified_name)
+        .to_string()
+}
+
+#[test]
```
conradgrobler, line 361: Having tests mixed inbetween normal functions makes it more difficult to follow for me personally.

conradgrobler (approved):
Perhaps it will be clearer when actually used in the code (or perhaps when I am more familiar with flatbuffers), but so far from looking at this code flatbuffers do not feel very convenient to use.

oak_idl/src/lib.rs
```diff
@@ -125,97 +129,117 @@ impl Transport for Channel {
...
+/// # }
+/// #
+/// # impl Foo {
+/// #     pub fn create(b: &mut flatbuffers::FlatBufferBuilder) -> flatbuffers::WIPOffset<Foo> {
+/// #         flatbuffers::WIPOffset::new(0)
+/// #     }
+/// # }
+/// #
+/// let mut b = oak_idl::MessageBuilder::<Foo>::default();
+/// let v = b.create_vector::<u8>(&[14, 12]);
```
jul-sh, line 163: `v` seems to be unused in the example code?

oak_idl/src/lib.rs
```diff
@@ -125,97 +129,117 @@ impl Transport for Channel {
     }
 }
 
+/// A helper struct to facilitate building a [`Message`].
+///
+/// It delegates most methods to the underlying [`flatbuffers::FlatBufferBuilder`] instance, but it
+/// adds a [`MessageBuilder::finish`] method that returns a completed [`Message`] instance.
```
jul-sh, line 135: It would be useful to document the purpose of this as part of this comment. 

Is this help meant as a generalized improvements to the ergonomics of flatbuffer on rust? If so it might be helpful to move it to a seperate helpers file that's non Oak IDL specific

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+///
+/// rpc_service Name {
+///     Method(Request) : Response (method_id: 42);
+/// }
+/// ```
+///
+/// For a service called `Name`, this method generates the following objects:
+///
+/// - a struct named `NameClient`, exposing a method for each method defined in the service. This
+///   may be used to connect to underlying transport and to invoke methods on the corresponding
```
jul-sh, line 65: Should we clarify this a bit perhaps? It seems we could say that _in tests_ it can be used to (directly) invoke methods on the corresponding `Server` object. In real use cases this will always be used to connect to an underlying transport channel.  :)

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+///   `Server` object.
+/// - a struct named `NameServer`, which implements the `oak_idl::Transport` trait, dispatching each
+///   request to the appropriate method on the underlying service implementation.
+/// - a trait named `Name`, with a method for each method defined in the macro, and an additional
+///   default method named `serve` which returns an instance of `NameServer`; the developer of a
+///   service would usually define a concrete struct and manually implement this trait for it.
+///
+/// For an input flatbuffer file with name `foo.fbs`, the generated Rust file will be located at
```
jul-sh, line 73: Thanks for the writing the comments for documentation, I think it helps with making the code generation feel a little less magical.

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        .into_iter()
+        .intersperse("\n".to_string())
+        .collect())
+}
+
+/// Generate the Rust objects from the input [`Service`] instance, corresponding to an `rpc_service`
+/// entry.
+fn generate_service(service: &Service) -> anyhow::Result<String> {
+    let mut lines = Vec::new();
+    lines.extend(vec![
```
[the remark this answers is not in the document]
jul-sh: I'm surprised we can't use a rust macro, might be helpful to explain the choice in the comment briefly. :)

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            "impl <S: {}> oak_idl::Transport for {}<S> {{",
+            service_name(service),
+            server_name(service)
+        ),
+        format!("    fn invoke(&mut self, request_message: oak_idl::TransportMessage) -> Result<oak_idl::TransportMessage, oak_idl::TransportError> {{"),
+        format!("        let response_header = oak_idl::Header {{"),
+        format!("            transaction_id: request_message.header.transaction_id,"),
+        format!("            method_id: request_message.header.method_id,"),
+        format!("        }};"),
+        format!("        match request_message.header.method_id {{"),
```
jul-sh, line 182: After reading the flatbuffer docs I'm a little surprised that there isn't an invoke based transport abstraction like this incorporated into flatbuffer itself. 

Instead their approach seems to be to generated gRPC code directly (only available in a select few languages). With an invoke based abstraction layer they could implement transport agnostic stubs once for each language and then allowing users to swap in a transport protocol that works for them.

oak_idl_gen_services/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+/// namespace Test;
+///
+/// attribute "method_id";
+///
+/// table Request {}
+/// table Response {}
+///
+/// rpc_service Name {
+///     Method(Request) : Response (method_id: 42);
+/// }
```
jul-sh, line 59: I'm a fan of the method_id personally to identify methods across renames & re-orders. That said, given that flatbuffer (like grpc) does not seem to employ that concept in their RPC interface definition format, I'm not 100% convinced that adding it on is a good idea: https://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html.

Might lead to surprises for people famiiar with flatbuffer/grpc, not expecting to find our own flavor of it. :)

oak_idl_tests/oak_services.rs.txt
```diff
@@ -0,0 +1,79 @@
+// Generated file, do not edit.
+
+pub struct OakClient<T: oak_idl::Transport> {
```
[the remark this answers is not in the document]
jul-sh: I think the naming's fine. :) Though I'm also not generally a fan of snapshot tests (imo it's usually nicer to test the actual functionality) but I think it should be fine here. 

The upside is that having the snapshot checked in may aid in understanding which code is generated. :)

oak_idl_tests/tests/oak_abi.rs
jul-sh: Seems like an independent test. Agreed that making that clearer would be nice (by swapping out "Oak" in the example for something like `MyService`

jul-sh (approved):
Thanks! It was good to see this fleshed out, I quite like it overall. 

Plus one to @conradgrobler's comment's on bincode, surprised it's still included in the code. 

As general feedback I think @mariaschett made a good point about making clearer what is Oak specific. In essence this seems to be basically just a generic invoke style abstraction to generate transport agnostic rust stubs from flatbuffer RPC interface declarations.

In principle it could be used eg for other projects as well (and in fact may be actually very useful for them). 

Making that clearer might be nice, as a someone new to the code might be lead to believe that this is tailored to the Oak Runtime, just by everything being called "oak". :P

oak_idl_tests/tests/oak_abi.rs
tiziano88: Renamed to TestService.

tiziano88 (commented):
Thanks for all the comments, I think I applied all of your suggestions now, looking much better!

tiziano88:
> Thanks! It was good to see this fleshed out, I quite like it overall.
> 
> Plus one to @conradgrobler's comment's on bincode, surprised it's still included in the code.

Going away in the next PR or so, I hope.

> As general feedback I think @mariaschett made a good point about making clearer what is Oak specific. In essence this seems to be basically just a generic invoke style abstraction to generate transport agnostic rust stubs from flatbuffer RPC interface declarations.
> 
> In principle it could be used eg for other projects as well (and in fact may be actually very useful for them).
> 
> Making that clearer might be nice, as a someone new to the code might be led to believe that this is tailored to the Oak Runtime, just by everything being called "oak". :P

I'll leave in the Oak branding for now, but I would be keen to split it out into a separate crate at some point, after we get some practice with it ourselves (and if it turns out not to be a terrible idea).

oak_idl_gen_services/src/lib.rs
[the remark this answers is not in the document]
mariaschett: I am still a bit worried about maintaining/changing the code printed out by `format`. 

I can't come up with a better suggestion, but I think it is worth to investigate soon. With macros, as Juiette said or also, e.g., there seems to be [codegen](https://docs.rs/codegen/0.1.3/codegen/).

I am not sure what you considered already, maybe it is worth adding a bried comment on your decision.

Is the code generated in the CI? Then we will note if there is some problem with the code, e.g., a new linting rule fires, yes? 

In general we could not do identation here, but use rustfmt to get the file formatted (I believe).

oak_idl_tests/tests/generate.rs
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#[test]
+fn generate() {
+    // This test simply copies the output created by the `build.rs` script back into the source
+    // tree, so that it can be tracked in Git. We use the `.txt` extension so that other tools do
```
mariaschett, line 19: Related to my comment above: why do we not want this code to be checked? 

I think we want this code to be checked.

oak_idl/src/lib.rs
```diff
@@ -114,7 +116,7 @@ impl Transport for Channel {
         let request_bytes =
             bincode::serialize(&request_message).map_err(|_| TransportError::InvalidMessage)?;
         let response_bytes = invoke(self.handle, &request_bytes)?;
-        let response_message: Message =
+        let response_message: TransportMessage =
             bincode::deserialize(&response_bytes).map_err(|_| TransportError::InvalidMessage)?;
```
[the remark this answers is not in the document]
tiziano88: Transport only deals with Transport messages anyways. I renamed that struct to TransportMessage to distinguish it from flatbuffer message (or protobuf message), but we may want to find another term for it. I think some of this should become clearer when plugged in the underlying framing protocol by @jul-sh .

In our case, the body of a transport message happens to be a flatbuffer message.

oak_idl/src/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+//! Utilities for improving the ergonomics of FlatBuffers in Rust.
+
+use std::{
+    marker::PhantomData,
+    ops::{Deref, DerefMut},
+};
+
+/// A helper struct to facilitate building a [`Message`].
+///
+/// It delegates most methods to the underlying [`flatbuffers::FlatBufferBuilder`] instance, but it
```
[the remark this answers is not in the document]
tiziano88: Done, and added some clarification, but it's really just a convenience thing, the same thing could be done manually, but we would have to keep track of who owns the buffers etc.

oak_idl_gen_services/src/lib.rs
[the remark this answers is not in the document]
tiziano88: Macros and codegen are not the way to go IMO. This generator will have to generate C++ code in the future, so it really just needs to build a string for the target language, it's just a coincidence that it's written in Rust and also generates Rust.

This is more or less what flatc (and almost any other code generator) does already: 

https://github.com/google/flatbuffers/blob/9aa08a429e340bb8adf9b2e978ce2817cc8007c5/src/idl_gen_rust.cpp#L2038-L2042

The generated code is not covered by CI, like generated protobuf code is not covered by CI. Linting and formatting rules don't apply to generated code, since it's almost never checked in (we only check it in here to have a regression test, and it would be pointless to update the generator to emit code that is correctly formatted for our own CI). Adding a dependency on rustfmt for this generator seems overkill, especially since it will be used for other languages too. But even if it were possible, it would be pointless, since it's only generated in order to be used by the compiler (and the IDE perhaps), not to be read.

oak_idl_tests/tests/generate.rs
tiziano88: See my other comment. We definitely don't want this to be checked, it's just a string from the point of view of the generator. E.g. we will add C++ generation at some point, and we don't want to apply our arbitrary formatting rules to it, since it will be used in other contexts (and not read by humans, in general). It's like generated protobuf code, or any other intermediate representation.

oak_idl_tests/tests/test_schema.rs
```diff
@@ -0,0 +1,88 @@
...
+        eprintln!("log: {}", request.entry().unwrap());
+        let mut b = oak_idl::utils::MessageBuilder::default();
+        let m = test_schema::LogResponse::create(&mut b, &test_schema::LogResponseArgs {});
+        b.finish(m).unwrap()
+    }
+}
+
+#[test]
+fn test_lookup_data() {
+    let s = TestServiceImpl;
```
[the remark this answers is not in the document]
tiziano88: Done.

tiziano88 (commented):
Thanks

oak_idl/src/utils.rs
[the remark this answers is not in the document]
tiziano88: Maybe I'll just remove these utils at some point and just use the plain flatbuffers functions, it seems to cause more confusion than benefit.

oak_idl/src/lib.rs
```diff
@@ -14,14 +14,10 @@
...
 
-//! A Rust-based IDL implemented via `macro_rules`. It is inspired by
-//! <https://github.com/google/tarpc> and
-//! <https://fuchsia.dev/fuchsia-src/reference/fidl/language/wire-format>.
-//!
-//! For each service definition, it generates both a client stub and a service trait, as well as a
-//! method to dispatch invocations to a service implementation.
-
 use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
```
conradgrobler, line 18: This should probable use `core::fmt::Debug` and have a `#[no_std]` attribute in this file for compatibility with the bare-metal runtime.

oak_idl/src/utils.rs
```diff
@@ -0,0 +1,142 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+//! Utilities for improving the ergonomics of FlatBuffers in Rust.
+
+use std::{
```
conradgrobler, line 18: should probably use `core` here as well.

conradgrobler (approved):

oak_idl/src/lib.rs
```diff
@@ -125,97 +129,117 @@ impl Transport for Channel {
...
+/// # }
+/// #
+/// let mut b = oak_idl::MessageBuilder::<Foo>::default();
+/// let v = b.create_vector::<u8>(&[14, 12]);
+/// let foo = Foo::create(&mut b);
+/// let m = b.finish(foo);
+/// ```
+pub struct MessageBuilder<'a, T> {
+    buf: flatbuffers::FlatBufferBuilder<'a>,
+    _phantom: PhantomData<T>,
```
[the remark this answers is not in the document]
jul-sh: Could we possible add a comment saying exactly that? :)

oak_idl_gen_services/src/lib.rs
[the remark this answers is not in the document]
jul-sh: Fair. I personally wouldn't mind using a macro, but I get your point about the generator & generated code only coincidentally sharing the same language. 

I do personally think that code should be generated with the prospect of people reading it in mind (I tend do that quite often when debugging, or even reviewing the inner working of a code generator). Hence it's imo a valid point. But don't feel super strongly about it, over the years I've learned how to read unreadable code :p.

oak_idl/src/lib.rs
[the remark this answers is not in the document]
tiziano88: I'll add it, but this is just exactly what PhantomData is for in general in Rust and other functional languages, it's not something that needs to be explained every time it's used IMO.

oak_idl/src/lib.rs
tiziano88: Done.

oak_idl_gen_services/src/lib.rs
jul-sh: thx! :)

oak_idl_gen_services/src/lib.rs
jul-sh: fair :)

oak_idl/src/lib.rs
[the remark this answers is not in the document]
tiziano88: If you are curious, see also https://doc.rust-lang.org/core/marker/struct.PhantomData.html and https://doc.rust-lang.org/nomicon/phantom-data.html for more details (Haskell has https://wiki.haskell.org/Phantom_type which is similar).

oak_idl_gen_services/src/lib.rs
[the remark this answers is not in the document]
tiziano88: Yeah I see your point, maybe a macro that just helps with the indentation would be useful across languages. I'll look into that separately and see if it makes things more readable (in the generator).

But nicely formatting the generated code is not useful IMO (also because in general the user of the generated code is someone else who may have different formatting preferences). E.g. having the generator try to break lines at 80 characters, or running rustfmt afterwards are unnecessary complications and make the generator much less portable.

P.S.: Thanks for reading the unreadable generated code :upside_down_face:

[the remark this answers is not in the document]
tiziano88: Also to clarify my previous point: I meant that it should not be covered by format / lint checks in CI. Correctness is actually checked in CI by means of the oak_idl_tests crate.

PR #2889, opened 2022-05-24
Refactor to remove `ExtensionConfigInfo`

### LINKED ISSUE #2851: Refactor to remove `ExtensionConfigurationInfo`
We should try to give this information to the Extensions directly on creation.

mariaschett:
Fixes #2851 .


rbehjati (approved):
Nice. Thanks :) 

I am just not entirely sure about renaming `ConfigurationInfo` to `ConfigurationReport`. It may sound like it is a cryptographic content since it is meant to be used in the context of remote attestation.

mariaschett:
Thank you, I agree `ConfigurationReport` is not ideal either. I'd leave it for now, because we should be revisiting how the Oak Functions Runtime produces this information anyway!

PR #2891, opened 2022-05-24
Rustify the memory allocator setup code

andrisaar:
Time to put on the nitpicking hats!

The whole goal of this PR is to make the code more Rust-y by using slices and proper generic data types, so style/readability comments are welcome.

I also generated `hvm_start_info.rs` from scratch using bindgen instead of using the one from Rust Hypervisor. The data structure is exactly the same (unsurprisingly), but this clean-room implementation means we can implement our traits on those data structures more cleanly.


experimental/uefi/baremetal/src/hvm_start_info.rs
```diff
@@ -0,0 +1,79 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/* Generated with:
+ * $ bindgen --no-layout-tests --use-core \
+ *   --ctypes-prefix c_types include/public/arch-x86/hvm/start_info.h
```
tiziano88, line 19: Would it work / make sense to actually do this in a build.rs file? Something like https://github.com/project-oak/oak/blob/fbb62f1303481506474a899ba0bc3d45e2da6a29/oak_idl_gen_services/src/lib.rs#L76

experimental/uefi/kernel/src/boot.rs
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use strum::{Display, FromRepr};
+
+#[derive(Debug, PartialEq, Eq, Display, FromRepr)]
+#[repr(u32)]
+/// E820 address range types according to Chapter 15 of th ACPI Specification, Version 6.3.
```
tiziano88, line 21:
```suggestion
/// E820 address range types according to Chapter 15 of the ACPI Specification, Version 6.3.
```

tiziano88, line 21: Perhaps also add a link to it? e.g. https://uefi.org/specs/ACPI/6.4/15_System_Address_Map_Interfaces/Sys_Address_Map_Interfaces.html (BTW this is v 6.4,)

experimental/uefi/kernel/src/lib.rs
```diff
@@ -39,19 +40,20 @@ mod serial;
...
+use rust_hypervisor_firmware_subset::paging;
 
 /// Main entry point for the kernel, to be called from bootloader.
-pub fn start_kernel(info: &dyn boot::Info) -> ! {
+pub fn start_kernel<E: boot::E820Entry>(info: &dyn boot::BootInfo<E>) -> ! {
     avx::enable_avx();
     logging::init_logging();
     paging::setup();
-    memory::init_allocator(info);
+    // If we don't find memory for heap, it's ok to panic.
```
tiziano88, line 50: uhmmm yes, that may be a problem

experimental/uefi/kernel/src/memory.rs
```diff
@@ -14,8 +14,9 @@
 // limitations under the License.
 //
 
+use crate::boot::{self, E820Entry};
+use anyhow::Result;
 use linked_list_allocator::LockedHeap;
-use rust_hypervisor_firmware_subset::boot;
```
tiziano88, line 20: (nit) remove blank line

tiziano88 (approved):
LITM (Looks Idiomatic To Me)

experimental/uefi/baremetal/src/hvm_start_info.rs
andrisaar: Maybe, I'll have to read up on bindgen a bit (you *can* call it from `build.rs`, that's the recommended way to do it).

However, that implies:
a) we will bundle the C header files from Xen and Linux with our code;
b) the files have some manual modifications (such as implementing the traits and overriding `c_types`), not sure how that works with bindgen.

For now, I'll leave it be, but yes, generating it from authoritative source would be a good idea.

experimental/uefi/kernel/src/boot.rs
andrisaar: Done. I missed the fact there's a newer version available.

experimental/uefi/baremetal/src/hvm_start_info.rs
andrisaar: I'm now generating it using bindgen during build time.

experimental/uefi/baremetal-crosvm/src/bootparam.rs
```diff
@@ -32,26 +29,27 @@ mod c_types {
...
-        "Linux Boot Protocol"
+impl E820Entry for boot_e820_entry {
+    fn entry_type(&self) -> E820EntryType {
+        E820EntryType::from_repr(self.type_).unwrap()
     }
-    fn rsdp_addr(&self) -> u64 {
-        0
+
+    fn addr(&self) -> usize {
+        self.addr.try_into().unwrap()
```
mariaschett, line 38: Out of curiousity: there are quite a bit of `unwrap`s here. What happens if they go wrong? Can they go wrong?

mariaschett (approved):
LGTM, but I am missing a lot of context, so I am relying on the other reviewers here.

Maybe some documentation/logging would be helpful to get more context, e.g., the pub traits `E820Entry`, `BootInfo` or the public functions.

experimental/uefi/baremetal-crosvm/src/bootparam.rs
andrisaar: The `addr` and `size` fields are fixed as `u64` (`addr` is effectively a pointer).
The `addr()` function wants to return `usize`, which depends on the target (it's the "pointer size") -- if you're running 64-bit code, `usize` is equivalent to `u64` and this conversion is safe. If you're running 32-bit code, this will panic, but we're not.

I've added a note stating why this is safe.

PR #2892, opened 2022-05-24
Add low-level vsock device driver

conradgrobler:
The low-level driver is used for sending and receiving raw packets without interpreting their meaning. A connection-oriented socket will be built on top of this.


experimental/virtio/src/vsock/mod.rs
```diff
@@ -14,4 +14,178 @@
...
+    /// The receive queue.
+    rx_queue: DeviceWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// The event queue used by the device to notify the driver that the guest CID has changed. We
+    /// ignore it for now as we don't support live migration, but it still must exist and be
+    /// configured.
+    event_queue: DeviceWriteOnlyQueue<QUEUE_SIZE, EVENT_BUFFER_SIZE>,
+    /// The the CID assigned to this VM.
+    guest_cid: u64,
+    /// Whether the device have been initialized and is ready to use.
+    ready: bool,
```
andrisaar, line 85: Is it ever possible to have a device that is not ready? Looks like `find_and_configure_device()` calls `init()`, so it's impossible to have it not initialized.

andrisaar (approved):

PR #2893, opened 2022-05-24
Start populating the IDT and handle double faults.

andrisaar:


conradgrobler (approved):

experimental/uefi/kernel/src/interrupts.rs
```diff
@@ -0,0 +1,58 @@
...
+
+lazy_static! {
+    static ref IDT: InterruptDescriptorTable = {
+        let mut idt = InterruptDescriptorTable::new();
+        idt.double_fault.set_handler_fn(double_fault_handler);
+        idt
+    };
+}
+
+extern "x86-interrupt" fn double_fault_handler(
```
rbehjati, line 31: What is "x86-interrupt"? Is this a type of linker that the Rust compiler already knows about or is it defined somewhere? Could you add a clarifying comment?

experimental/uefi/kernel/src/interrupts.rs
```diff
@@ -0,0 +1,58 @@
...
+    stack_frame: InterruptStackFrame,
+    _error_code: u64,
+) -> ! {
+    // We're not calling `panic!` here because (a) the panic location would be all wrong (the error
+    // is not in the fault handler itself as the location would suggest), (b) in theory the
+    // panic handler itself could cause a double fault, and (c) we want to be sure that we shut
+    // down the machine after us.
```
rbehjati, line 38: nit: is "shut down the machine after us" an idiom that I don't know of? :)

rbehjati (approved):

experimental/uefi/kernel/src/interrupts.rs
andrisaar: No, it's not about linker -- it changes the calling convention of the function and what code the compiler emits. The particular calling convention is required by the x86-64 ABI.

You can read more about the `x86-interrupt` in particular here: https://github.com/rust-lang/rust/issues/40180

In short, you don't really expect to see that used anywhere else except for interrupt handlers.

rbehjati: I think a comment with a link to that issue would help us understand what is going on later when we look at the code.

experimental/uefi/kernel/src/interrupts.rs
rbehjati: Then I think describing it in those words would be clearer :)

PR #2899, opened 2022-05-27
Remove bincode and serde from IDL

### LINKED ISSUE #2898: Remove dependency on bincode and serde from IDL
We should strip down the IDL layer to a bare minimum, in which it merely generates code that serializes / deserializes request / response messages, and method ids. Anything else related to how an actual transport may work should be implementable outside of the IDL crate(s).

cc @jul-sh this should help with the comms channel implementation.

Part of
- https://github.com/project-oak/oak/issues/2500

tiziano88:
Unify client and transport errors.

Simplify transport, and rename it to handler to avoid name clashes with
underlying transport / channel mechanisms in the future.

Use reference in request message to avoid copying the request buffer at
all.

Improve error handling in generated code (no unwrap).

Fix #2898


jul-sh (approved):
LGTM, thanks :)

oak_idl/src/lib.rs
```diff
@@ -20,120 +20,39 @@ extern crate alloc;
...
-use serde::{Deserialize, Serialize};
 
 pub mod utils;
 
 #[derive(Debug)]
-pub enum ClientError {
+pub enum Error {
+    /// The request message could not be deserialized correctly.
     InvalidRequest,
+    /// The response message could not be serialized correctly.
```
jul-sh, line 30: Should those comments be swapped? Surely the request would get serialised, and the response is deserialized

oak_idl/src/lib.rs
```diff
@@ -20,120 +20,39 @@ extern crate alloc;
...
-    }
-}
-
-#[derive(Debug)]
-pub enum TransportError {
-    InvalidMessage,
-    InvalidTransactionId,
+    /// The method id provided for the invocation was not implemented by the server.
     InvalidMethodId,
 }
```
jul-sh, line 34: Should we keep this? I always thought that the seperation between serialization errors (error occurred on the client) and an error response was useful

oak_idl/src/lib.rs
tiziano88: I think they are the right way round, but let me double check by writing it down.

The generated client takes a `&[u8]` (which must have been serialized by the caller), and tries to deserialize it to make sure it corresponds to the correct type. If that fails, then the client returns `InvalidRequest` to the caller directly.

When the server receives the request, it also tries to deserialize it, and returns `InvalidRequest` to the client in that case.

The generated client also deserializes the response from the server, and if that fails it returns `InvalidResponse`. As I'm typing this, I realise that indeed the comment on `InvalidResponse` was incorrect!

So perhaps actually both comments should say "deserialized", since as a matter of fact the serialization to flatbuffer is done entirely by the caller (never by the generated code). Interesting.

oak_idl/src/lib.rs
tiziano88: I think we would need to many layers of errors in that case, and there is really not much point for the caller to determine what to do differently in each case (apart from logging something). So, I'm trying to get closer to something like gRPC errors (which are the same on client and server, but some may be generated by the client or server -- see https://grpc.github.io/grpc/core/md_doc_statuscodes.html ). But let's see how it goes, also when we introduce the errors at the framing layer.

oak_idl/src/lib.rs
tiziano88: (fixed the comment now, thanks!)

oak_idl/src/lib.rs
```diff
@@ -20,120 +20,39 @@ extern crate alloc;
...
 use alloc::vec::Vec;
 use core::fmt::Debug;
-use serde::{Deserialize, Serialize};
 
 pub mod utils;
 
 #[derive(Debug)]
-pub enum ClientError {
+pub enum Error {
+    /// The request message could not be deserialized correctly.
```
mariaschett, line 28: If we only want to use the errors `InvalidRequest/Response` for deserialization problems, we could also reflect that in the name `UnDeserializeableRequest` or so :)

mariaschett: Also: Do we have no potential errors when serializing?

oak_idl/src/lib.rs
```diff
@@ -20,120 +20,39 @@ extern crate alloc;
...
-    ///
-    /// The client picks a unique transaction id to put in the request header, and when the server
-    /// replies to that message, it also uses the same transaction id in the response message.
-    pub transaction_id: u32,
-
+/// A request message representing an invocation.
+pub struct Request<'a> {
     /// Identifies the method to be invoked, as defined by the IDL.
     pub method_id: MethodId,
+    /// The serialized payload supplied as the request.
```
mariaschett, line 43: Do I understand this correctly: `body` is the serialized request for the method identified by `method_id`? I am not sure I understand this from the comment.

oak_idl/src/lib.rs
```diff
@@ -20,120 +20,39 @@ extern crate alloc;
...
-pub struct Header {
-    /// Identifies the transaction over the underlying transport. Each transaction (a.k.a. call or
-    /// invocation) over the same transport must have a unique transaction id (regardless of the
-    /// method_id value).
-    ///
-    /// The client picks a unique transaction id to put in the request header, and when the server
-    /// replies to that message, it also uses the same transaction id in the response message.
-    pub transaction_id: u32,
-
+/// A request message representing an invocation.
```
mariaschett, line 39: Do I understand correctly: A `Request` to invoke a method identified by `method_id` with the arguments given in `body` ?

oak_idl/src/lib.rs
jul-sh: Thanks for updating, that makes a lot more sense. What's the rationale behind the Client accepting a byte slice (and deserializing it to verify) as opposed to instance of the `request_type`?

oak_idl/src/lib.rs
jul-sh: Alright :). Yh I'm more familiar with HTTP status codes (which are sent by the server only)

oak_idl/src/lib.rs
tiziano88: Yes, clarified.

oak_idl/src/lib.rs
jul-sh: In that case a different (client specific) error is usually thrown. Eg see the `fetch` API, which returns a JavaScript promise (that can resolve to a value or error). In case that a response is received, the promise succeeds, even if the status code is eg 500 (internal server error). 

In case of a network error, the promise throws a JS error: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

jul-sh: Which isn't to say that there aren't some client implementations out there that do construct a "fake" status code, but it's not the typical case.

PR #2900, opened 2022-05-27
Add timeout for responses

### LINKED ISSUE #2903: Add timeout to Trusted Shuffler

mariaschett:
Fixes #2903.

### DIFF
--- a/experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
+++ b/experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
@@ -198,7 +198,6 @@ async fn one_empty_response_test() {
 // If all requests receive a response from the backend within the given timeout, then the Trusted
 // Shuffler responds with the all responses.
 #[tokio::test]
-
 async fn no_empty_response_test() {
     let anonymity_value = 2;
     // This should give the non-empty response enough time to arrive.
@@ -221,7 +220,6 @@ async fn no_empty_response_test() {
 // If no request receives a response from the backend within the given timeout, then the Trusted
 // Shuffler responds with the all empty responses.
 #[tokio::test]
-
 async fn all_empty_responses_test() {
     let anonymity_value = 2;
     let timeout = Duration::from_millis(200);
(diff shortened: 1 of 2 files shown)


experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -28,6 +28,12 @@ impl RequestHandler for TestRequestHandler {
     // In the real use-case, this function should send requests to the backend and
     // return responses.
     async fn handle(&self, request: Vec<u8>) -> anyhow::Result<Vec<u8>> {
+        if drop_request(request.clone()) {
+            // In that case we don't send an answer ever.
+            loop {
+                tokio::task::yield_now().await;
+            }
```
tiziano88, line 35: Maybe too extreme, how about returning it after, say, 1 second? (i.e. above the threshold)

tiziano88 (approved):

conradgrobler (approved):

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Why do you think it is too extreme? 

I was trying to find a way to say to don't anwer ever, and that was the best I could think of. But I didn't find any best practise, sadly.

tiziano88: I think this will use a task forever (i.e. it will never be "released"), which does not seem great (it's only fine because it's in test, otherwise the unused tasks will pile up). Or is there something else that makes it less bad that I'm missing?

mariaschett: I had assumed that the yielding task will only be held while the single test is running, but is released once the test finishes. 

In that case I think it doesn't much matter if we sleep for longer or yield repeatedly (the latter allows me to not think of time at all).

But I am not sure I am right about the task being stopped after the test finishes. Any idea how I can confirm, or should I straight move to sleeping?

Issue #2876, opened 2022-05-19
Improve efficiency and throughput of virtqueue operations


conradgrobler:
The current implementation is very simple, but involves copies of buffers that could be avoided.

We should also experiment with different queue lenghts and buffer sizes to optimise throughput.

conradgrobler:
Another optimisation it to not always flush, but accumulate multiple writes until a buffer is full or flush is called explicitly.

PR #2902, opened 2022-05-27
Use virtio vsock socket as comms channel on crosvm

conradgrobler:
This PR changes the crosvm-compatible bare-metal implementation to use a virtio vsock channel for host-to-guest communication.

The vsock channel can only be connected after the VM has booted when using crosvm. When using a serial port on QEMU for the comms the loader's comms channel for connecting to the serial port (a unix domain socket) must be created before the VM boots. To work around this, a wrapper struct for the VsockStream has been added so that it can be connected later, the first time we try to use it for actual comms. It is now also up to the VMM to create the comms channel, rather than passing it in as part of the VMM startup parameters.

### DIFF
--- a/experimental/uefi/channel/src/lib.rs
+++ b/experimental/uefi/channel/src/lib.rs
@@ -74,6 +74,7 @@ where
         };
         self.inner.write_all(&length_bytes)?;
         self.inner.write_all(&frame.body)?;
+        self.inner.flush()?;
         Ok(())
     }
 }
--- a/experimental/uefi/kernel/src/lib.rs
+++ b/experimental/uefi/kernel/src/lib.rs
@@ -63,24 +63,26 @@ pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
     main(info);
 }
 
-#[cfg(feature = "vsock_channel")]
 fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
     info!("In main! Boot protocol:  {}", info.protocol());
+    let attestation_behavior =
+        AttestationBehavior::create(EmptyAttestationGenerator, EmptyAttestationVerifier);
+    runtime::framing::handle_frames(get_channel(), attestation_behavior).unwrap();
+}
+
+#[cfg(not(feature = "vsock_channel"))]
+fn get_channel() -> serial::Serial {
+    serial::Serial::new()
+}
+
+#[cfg(feature = "vsock_channel")]
+fn get_channel() -> virtio::vsock::socket::Socket {
     let vsock = virtio::vsock::VSock::find_and_configure_device()
         .expect("Couldn't configure PCI virtio vsock device.");
     info!("Socket device status: {}", vsock.get_status());
     let listener = virtio::vsock::socket::SocketListener::new(vsock, VSOCK_PORT);
     let socket = listener.accept().unwrap();
-    runtime::framing::handle_frames(socket).unwrap();
-}
-
-#[cfg(not(feature = "vsock_channel"))]
-fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
-    info!("In main! Boot protocol:  {}", info.protocol());
-    let serial = serial::Serial::new();
-    let attestation_behavior =
-        AttestationBehavior::create(EmptyAttestationGenerator, EmptyAttestationVerifier);
-    runtime::framing::handle_frames(serial, attestation_behavior).unwrap();
+    socket
 }
 
 /// Common panic routine for the kernel. This needs to be wrrapped in a
--- a/experimental/uefi/kernel/src/lib.rs
+++ b/experimental/uefi/kernel/src/lib.rs
@@ -81,8 +81,7 @@ fn get_channel() -> virtio::vsock::socket::Socket {
         .expect("Couldn't configure PCI virtio vsock device.");
     info!("Socket device status: {}", vsock.get_status());
     let listener = virtio::vsock::socket::SocketListener::new(vsock, VSOCK_PORT);
-    let socket = listener.accept().unwrap();
-    socket
+    listener.accept().expect("Couldn't accept connection.")
 }
 
 /// Common panic routine for the kernel. This needs to be wrrapped in a
--- a/experimental/uefi/kernel/src/lib.rs
+++ b/experimental/uefi/kernel/src/lib.rs
@@ -38,6 +38,7 @@ mod interrupts;
 mod libm;
 mod logging;
 mod memory;
+#[cfg(not(feature = "vsock_channel"))]
 mod serial;
 
 use core::panic::PanicInfo;
@@ -46,6 +47,10 @@ use oak_remote_attestation::handshaker::{
     AttestationBehavior, EmptyAttestationGenerator, EmptyAttestationVerifier,
 };
 use rust_hypervisor_firmware_boot::paging;
+
+#[cfg(feature = "vsock_channel")]
+// The virtio vsock port on which to listen.
+const VSOCK_PORT: u32 = 1024;
 
 /// Main entry point for the kernel, to be called from bootloader.
 pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
@@ -58,6 +63,18 @@ pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
     main(info);
 }
 
+#[cfg(feature = "vsock_channel")]
+fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
+    info!("In main! Boot protocol:  {}", info.protocol());
+    let vsock = virtio::vsock::VSock::find_and_configure_device()
+        .expect("Couldn't configure PCI virtio vsock device.");
+    info!("Socket device status: {}", vsock.get_status());
+    let listener = virtio::vsock::socket::SocketListener::new(vsock, VSOCK_PORT);
+    let socket = listener.accept().unwrap();
+    runtime::framing::handle_frames(socket).unwrap();
+}
+
+#[cfg(not(feature = "vsock_channel"))]
 fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
     info!("In main! Boot protocol:  {}", info.protocol());
     let serial = serial::Serial::new();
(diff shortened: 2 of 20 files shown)


experimental/uefi/kernel/src/lib.rs
```diff
@@ -60,10 +65,23 @@ pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
...
 fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
     info!("In main! Boot protocol:  {}", info.protocol());
-    let serial = serial::Serial::new();
     let attestation_behavior =
         AttestationBehavior::create(EmptyAttestationGenerator, EmptyAttestationVerifier);
-    runtime::framing::handle_frames(serial, attestation_behavior).unwrap();
+    runtime::framing::handle_frames(get_channel(), attestation_behavior).unwrap();
+}
+
+#[cfg(not(feature = "vsock_channel"))]
```
mariaschett, line 73: Hu, took me some time to see the difference between the two `get_channel`, i.e., the feature gating with not. Maybe some brief comment with when/why we fall back on that one would be nice.

experimental/uefi/loader/src/crosvm.rs
```diff
@@ -14,32 +14,88 @@
...
-    os::unix::{
-        io::{AsRawFd, OwnedFd},
-        net::UnixStream,
-    },
+    os::unix::{io::AsRawFd, net::UnixStream},
     process::Stdio,
 };
+use vsock::VsockStream;
+
+/// The guest VM context ID for virtio vsock connecitons.
```
mariaschett, line 33: connecitons -> connections. I don't think `VS_GUEST_CID` is public, hence // are enough, or not?

third_party/rust-hypervisor-firmware-virtio/src/pci.rs
```diff
@@ -391,7 +391,7 @@ impl VirtioTransport for VirtioPciTransport {
     }
 
     fn set_used_ring(&self, addr: u64) {
-        // queue_used: 0x28
+        // queue_used: 0x30
```
mariaschett, line 394: For this comment: it doesn't give me much more insight, so I'd think we could either we expand or remove it.

third_party/rust-hypervisor-firmware-virtio/src/device.rs
```diff
@@ -136,6 +136,11 @@ where
         self.transport.read_device_config(offset)
     }
 
+    /// Gets the device status.
```
mariaschett, line 139: Do we have a link to the status codes it may return?

scripts/docker_run
```diff
@@ -60,6 +60,20 @@ if [[ -e "/dev/kvm" ]]; then
   )
 fi
 
+# If the host supports virtio vhost vsock, allow the image to use it
```
mariaschett, line 63: nit: . at end, maybe: image -> docker image (also below).

kokoro/presubmit.sh
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
 
-# On Kokoro instances /dev/kvm is owned by root:root. Let's fix the permissions
-# so that it's owned by group `kvm` as our scripts expect it to.
+# On Kokoro instances /dev/kvm, dev/vhost-vsock and dev/vsock are owned by root:root.
+# Let's fix the permissions so that it's owned by group `kvm` as our scripts expect it to.
```
mariaschett, line 4: Is "Let's fix the permissions" something we still need to do? I don't think so. Maybe re-phrase to: "Set permissions so that /dev/kvm, /dev/vhost-sock, and /dev/vsock (?) are owned by group `kvm` ..."

experimental/virtio/src/vsock/socket.rs
```diff
@@ -344,6 +356,7 @@ impl ciborium_io::Write for Socket {
 
     fn flush(&mut self) -> Result<(), Self::Error> {
         // We always flush on write, so do nothing.
+        // TODO(#2876): Don't always flush on write.
```
mariaschett, line 359: pre-existing: This is a bit confusing. Why do we have the flush method in the first place then, if it does nothing? 

And, just because: would it hurt to actually flush? I mean, we promise to flush...

experimental/uefi/channel/src/lib.rs
```diff
@@ -74,6 +74,7 @@ where
         };
         self.inner.write_all(&length_bytes)?;
         self.inner.write_all(&frame.body)?;
+        self.inner.flush()?;
```
mariaschett, line 77: Should we add #2876 here?

experimental/virtio/src/queue/virtq.rs
```diff
@@ -122,6 +125,8 @@ pub struct UsedRing<const QUEUE_SIZE: usize> {
     pub idx: u16,
     /// The ring-buffer containing the used elements.
     pub ring: [UsedElem; QUEUE_SIZE],
+    /// Event details. Only used if VIRTIO_F_EVENT_IDX has been negotiated.
```
mariaschett, line 128: I am not sure I understand "Only used if ...". Does it mean I have to make sure "VIRTIO_EVENT" has been negotiated before I use the `avail_event`? (Similar in line 99)

experimental/uefi/loader/src/crosvm.rs
```diff
@@ -75,4 +131,8 @@ impl Vmm for Crosvm {
         self.instance.start_kill()?;
         self.wait().await
     }
+
+    fn create_comms_channel(&self) -> Result<Box<dyn ReadWrite>> {
+        Ok(Box::new(VsockStreamChannel { stream: None }))
```
mariaschett, line 136: Why is the stream `None`?

mariaschett: Ah, I think it says in line 47, maybe it is worth to repeat/link here? Or maybe rename `create_comms_channel_without_stream`?

experimental/uefi/loader/src/crosvm.rs
```diff
@@ -14,32 +14,88 @@
...
-    },
+    os::unix::{io::AsRawFd, net::UnixStream},
     process::Stdio,
 };
+use vsock::VsockStream;
+
+/// The guest VM context ID for virtio vsock connecitons.
+const VSOCK_GUEST_CID: u32 = 3;
+
+/// The guest VM virtio vsock port.
```
mariaschett, line 36: I don't think `VS_GUEST_PORT` is public, hence // are enough, or not?

experimental/uefi/loader/src/crosvm.rs
```diff
@@ -14,32 +14,88 @@
...
 pub struct Crosvm {
     console: UnixStream,
     instance: tokio::process::Child,
 }
 
+/// A wrapper for a vsock stream that can be used as a communication channel between the loader and
+/// the guest VM.
+///
+/// The vsock channel can only be created after the VM is booted, so we start without a stream and
+/// only connect later when we try to use it for the first time.
```
mariaschett, line 48: Where does "connect later" actually happen: in `ensure_stream`? Do we check every time we write/read? Is this negligible?

experimental/uefi/loader/src/crosvm.rs
```diff
@@ -14,32 +14,88 @@
...
+pub struct VsockStreamChannel {
+    stream: Option<VsockStream>,
+}
+
+impl VsockStreamChannel {
+    /// Ensures that the inner stream is connected to the VM.
```
mariaschett, line 54: I think here we do more than `ensure_stream`---we also connect the stream if it is not there, no? Maybe we can update the comment to reflect this, i.e., something along the lines "ensures the inner stream is connected ... by connecting/creating the stream if necessary" ?

mariaschett (approved):
Thank you, Conrad!

I am missing context to fully appreciate, but I am happy to have a second look again/together, if you want!

I mainly added comments where something was confusing me on first sight.

experimental/uefi/loader/src/crosvm.rs
conradgrobler: Yes, it happens in ensure_stream. We do check every time, but it is negligible, as it just checks whether the inner stream is `None` or `Some` to decide whetehr it is connected or not.

experimental/uefi/loader/src/crosvm.rs
conradgrobler: Done.

experimental/uefi/loader/src/crosvm.rs
conradgrobler: Done.

experimental/uefi/loader/src/crosvm.rs
conradgrobler: Done.

experimental/uefi/channel/src/lib.rs
conradgrobler: This was added so that we don't need to update this code when implementing #2876. We would still flush for every frame, but not for every invocation of `write_all`.

experimental/uefi/kernel/src/lib.rs
conradgrobler: Added some comments.

experimental/uefi/loader/src/crosvm.rs
conradgrobler: Added a comment. The function is part of a trait, so the other implementation on QEMU would have to be renamed. On QEMU it is connected before booting the VM, so the name change would not make sense there.

experimental/virtio/src/queue/virtq.rs
conradgrobler: The field is not used in our implementation, but is there as padding. Reading or writing it will have no effect. Updated the comment to clarify.

experimental/virtio/src/vsock/socket.rs
conradgrobler: `flush` is a required function on the `Read` trait. The current implementation flushes every time we write some bytes, which is not always needed. This means that we flush twice for a single frame (once for the header, once for the bofy) The better implementation would use a buffer to only flush once the frame header and body has been written.

Updated the comment to try to clarify.

kokoro/presubmit.sh
conradgrobler: Updated the comments.

scripts/docker_run
conradgrobler: Done.

third_party/rust-hypervisor-firmware-virtio/src/device.rs
conradgrobler: Added link to the spec.

third_party/rust-hypervisor-firmware-virtio/src/pci.rs
conradgrobler: This is part of imported third_party code. I just fixed the typo which caused some confusion.

conradgrobler (commented):
Thanks for the review

PR #2904, opened 2022-05-27
Implement InternalError in IDL

jul-sh:
The server should be able to indicate that a method invocation failed. The alteratives would be panicking (the tests do this at the moment) or silently failing by sending an empty / default value. Conceptually equivalent to a 500 status code in HTTP https://http.cat/500

### DIFF
--- a/oak_idl/src/lib.rs
+++ b/oak_idl/src/lib.rs
@@ -32,18 +32,6 @@ pub enum Error {
     /// The method id provided for the invocation was not implemented by the server.
     InvalidMethodId,
     /// An error occured while invoking the method on the server.
-    MethodError(MethodError),
-}
-
-impl From<MethodError> for Error {
-    fn from(e: MethodError) -> Self {
-        Error::MethodError(e)
-    }
-}
-
-#[derive(Debug)]
-pub enum MethodError {
-    /// An error occured when invoking the method.
     InternalError,
 }
 
--- a/oak_idl/src/lib.rs
+++ b/oak_idl/src/lib.rs
@@ -31,6 +31,20 @@ pub enum Error {
     InvalidResponse,
     /// The method id provided for the invocation was not implemented by the server.
     InvalidMethodId,
+    /// An error occured while invoking the method on the server.
+    MethodError(MethodError),
+}
+
+impl From<MethodError> for Error {
+    fn from(e: MethodError) -> Self {
+        Error::MethodError(e)
+    }
+}
+
+#[derive(Debug)]
+pub enum MethodError {
+    /// An error occured when invoking the method.
+    InternalError,
 }
 
 /// Unique identifier of a method within a service.
--- a/oak_idl_gen_services/src/lib.rs
+++ b/oak_idl_gen_services/src/lib.rs
@@ -282,7 +282,7 @@ fn generate_service_method(rpc_call: &RPCCall) -> Vec<String> {
     // implementation of this method, therefore needs to be wrapped in `oak_idl::Message` in order
     // to transfer its ownership to the caller.
     vec![format!(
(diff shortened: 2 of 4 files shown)


oak_idl/src/lib.rs
```diff
@@ -31,6 +31,20 @@ pub enum Error {
     InvalidResponse,
     /// The method id provided for the invocation was not implemented by the server.
     InvalidMethodId,
+    /// An error occured while invoking the method on the server.
+    MethodError(MethodError),
```
tiziano88, line 35: I think the extra layer of error does not really add much (which is why in 457e70e6aa67ce674d95a681c6b2a0f28f603436 I just removed the one I originally added). In the end the caller will probably just log something and move on, so having a complex error hierarchy does not really help much IMO. I would reserve this pattern for when some class of error comes from a different library, and we don't want to just convert to our own error space, but we don't have that problem here (yet).

And I think we should probably just switch to the gRPC error codes at some point, and also allow the method to return any of them (e.g. it may return NOT_FOUND directly if it needs to, which will be useful).

oak_idl_tests/test_schema_services.rs.txt
```diff
@@ -56,8 +56,8 @@ impl <S: TestService> oak_idl::Handler for TestServiceServer<S> {
 }
 
 pub trait TestService: Sized {
-    fn lookup_data(&self, request: &LookupDataRequest) -> oak_idl::utils::Message<LookupDataResponse>;
-    fn log(&self, request: &LogRequest) -> oak_idl::utils::Message<LogResponse>;
+    fn lookup_data(&self, request: &LookupDataRequest) -> Result<oak_idl::utils::Message<LookupDataResponse>, oak_idl::MethodError>;
```
tiziano88, line 59: How is the error going to be serialized at the framing level? Will it be an additional numeric field in the header of the response?

oak_idl/src/lib.rs
jul-sh: Alright, can do. I personally disagree and think that a lot of meaning is lost in gRPC's approach, but can go that way. In that case we should probably indeed switch to gRPC codes. It's worth pointing out though that even in gRPC's case, some codes are never meant to originate in the client. Hence we'd either need to maintain hierarchy like this, or rely on the method implementation to never send those errors.

oak_idl_tests/test_schema_services.rs.txt
jul-sh: The current framing doc includes a status field, which includes a code for internal errors:
https://docs.google.com/document/d/1Q4u5hYzOFF74_RfwZzpRWgSSwJwfQ-GXRvNllgCXOO8/edit

oak_idl_gen_services/src/lib.rs
```diff
@@ -282,7 +282,7 @@ fn generate_service_method(rpc_call: &RPCCall) -> Vec<String> {
     // implementation of this method, therefore needs to be wrapped in `oak_idl::Message` in order
     // to transfer its ownership to the caller.
     vec![format!(
-        "    fn {}(&self, request: &{}) -> oak_idl::utils::Message<{}>;",
+        "    fn {}(&self, request: &{}) -> Result<oak_idl::utils::Message<{}>, oak_idl::Error>;",
```
tiziano88, line 285: This may need to be `core::Result`, right?

tiziano88 (approved):

oak_idl_gen_services/src/lib.rs
jul-sh: Doesn't seem like it from my testing, I think Vec was unique because it's a part of alloc.

Issue #2608, opened 2022-03-14
Remote Attestation Refactoring


jul-sh:
There's some opportunity for refactoring our remote attestation implementation, this issue is meant to track aspects that could be improved. It's worth noting that we currently support remote attestation over both unary and bidi-streaming gRPC. We plan to remove support for attestations over bidi-streaming gRPC in the future. It'll be easiest to tackle these issues once that has happened. 

Probable refactoring opportunities:

- [ ] Overlap between the Handshaker & Encryptor struct, both of which can represent the same state.  See discussion on #2596: ([relevant comment](https://github.com/project-oak/oak/pull/2596#discussion_r823726105))
- [] The [Handshaker](https://github.com/project-oak/oak/blob/ba64c4b2f86e7c668f4dc1daf4b57a7fe91500b3/remote_attestation/rust/src/handshaker.rs) maintains duplicate handshake state in a transcript key and state enum, which are implicitly expected to have the same state. It'd be nice to deduplicate this, to increase resilience to future changes and prevent unexpected behavior. 
- [] Shared configuration state is cloned into each Handshaker. Cloning shared state may lead to issues down the road when state changes may not be in all clones. It'd be nice to use references instead. Ref: #2574
- The handshaker exposes `next_step` method that optionally returns a next step (if not completed) and a `is_completed` method. They're expected to behave the same, if `is_completed` returns false `next_step` should return `None`. However with future changes that may not be the case, which would lead to infinite loops in our client implementation: [link](https://github.com/project-oak/oak/blob/ba64c4b2f86e7c668f4dc1daf4b57a7fe91500b3/grpc_streaming_attestation/src/client.rs). It'd be nice to simplify the API of Handshaker in that regard.

jul-sh:
Issue 1 & 4 could be resolved relatively nicely by refactoring the signature of the `handshaker.next_step`. Currently it returns: `anyhow::Result<Option<Vec<u8>>>`. It uses a nested Result and Option, to effectively an Enum of "in progress here's next message", "done, no message", "error". 

Instead it could return an enum of "in progress, returns next message", "done, returns encryptor (consuming the handshaker)", "error".

jul-sh:
The ideal structure would probably be:

For each client, the only state that is maintained on the server is a transcript of messages (equivalent of the current `Transcript` inside the handshakers, a Vec of bytes). When a new message is received, it is appended to the transcript. That transcript is then passed to a stateless function that derives verifies it and derives the next step. 

[...]

PR #2906, opened 2022-05-30
Refactor: Make remote attestation client-side logic transport protocol agnostic

jul-sh:
By extracting it as a trait into the `remote_attestation_sessions` crate. Clients then implement that trait using a transport layer of their choice. :) 

Pulled this out of #2896 to make it more straightforward to review. 

I recommend reviewing by commit, which makes it clearer which code is newly changed vs what has just been moved around.


remote_attestation_sessions/src/lib.rs
```diff
@@ -16,129 +16,13 @@
...
 
-extern crate alloc;
+pub mod server;
 
-use alloc::{boxed::Box, sync::Arc, vec::Vec};
-use anyhow::bail;
-use lru::LruCache;
-use oak_remote_attestation::handshaker::{AttestationBehavior, Encryptor, ServerHandshaker};
+#[cfg(feature = "client")]
+pub mod client;
```
tiziano88, line 25: Why not have two separate crates then? Especially since IIUC the server part of this crate is expected to be no_std compatible, and it is really meant for a different class of programs.

grpc_unary_attestation/src/client.rs
```diff
@@ -16,98 +16,70 @@
...
 };
-use oak_remote_attestation_sessions::SessionId;
 use tonic::transport::Channel;
 
-/// gRPC Attestation Service client implementation.
-pub struct AttestationClient {
-    session_id: SessionId,
-    encryptor: Encryptor,
-    client: UnarySessionClient<Channel>,
+pub struct GrpcClient {
```
tiziano88, line 27: Could you add a comment to this struct?

grpc_unary_attestation/src/client.rs
```diff
@@ -16,98 +16,70 @@
...
-                    })
-                    .await
-                    .context("Couldn't send client hello message")?
-                    .into_inner();
-            }
-        }
+/// gRPC Attestation Service client implementation.
+pub struct AttestationClient {
+    inner: GenericAttestationClient<GrpcClient>,
+}
```
tiziano88, line 63: This feels like it should be in a separate crate (see my other comment too).

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::SessionId;
+use anyhow::Context;
+use async_trait::async_trait;
+use oak_remote_attestation::handshaker::{
+    AttestationBehavior, ClientHandshaker, Encryptor, ServerIdentityVerifier,
+};
+
+#[async_trait]
+pub trait UnaryClient {
+    async fn message(&mut self, session_id: SessionId, body: Vec<u8>) -> anyhow::Result<Vec<u8>>;
```
tiziano88, line 26: Doc?

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+            .message(session_id, client_hello)
+            .await
+            .context("Couldn't message client hello message")?;
+
+        while !handshaker.is_completed() {
+            let request = handshaker
+                .next_step(&response)
+                .context("Couldn't process handshake message")?;
+
+            if let Some(request) = request {
```
tiziano88, line 62: What if not? Also, is it worth adding some debug logging throughout, or do you think just returning the errors should be enough in practice?

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+        let client_hello = handshaker
+            .create_client_hello()
+            .context("Couldn't create client hello message")?;
+
+        let mut response = client
+            .message(session_id, client_hello)
+            .await
+            .context("Couldn't message client hello message")?;
+
+        while !handshaker.is_completed() {
```
tiziano88, line 57: Probably pre-existent, but it seems that there is some coupling between the `is_completed` method and `get_encryptor`. I wonder if it could be expressed in a more type safe way by making `handshaker.next_step` return an enum with two options, one with `InProgress(Handshaker)` and another one with `Completed(Encryptor)`. But I'm not sure that works out given the recursion in one side (at the very least it would have to take `self` as first param (by value), but even that may not be sufficient). @ipetr0v maybe you looked into it at some point and it didn't work?

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        Ok(Self {
+            session_id,
+            encryptor,
+            client,
+        })
+    }
+
+    /// Sends data encrypted by the [`Encryptor`] to the server and decrypts the server responses.
+    pub async fn message(&mut self, payload: Vec<u8>) -> anyhow::Result<Option<Vec<u8>>> {
```
tiziano88, line 82:
```suggestion
    pub async fn message(&mut self, payload: &[u8]) -> anyhow::Result<Option<Vec<u8>>> {
```

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+            .context("Couldn't get encryptor")?;
+
+        Ok(Self {
+            session_id,
+            encryptor,
+            client,
+        })
+    }
+
+    /// Sends data encrypted by the [`Encryptor`] to the server and decrypts the server responses.
```
tiziano88, line 81: Could you document the return type, given it involves both Result and Option, it may not be obvious when each one is used to signal failure.

remote_attestation_sessions/src/client.rs
```diff
@@ -0,0 +1,101 @@
...
+            .encrypt(&payload)
+            .context("Couldn't encrypt request")?;
+
+        let encrypted_response = self
+            .client
+            .message(self.session_id, encrypted_request)
+            .await
+            .context("Couldn't message encrypted data request")?;
+
+        let encoded_response = self
```
tiziano88, line 94: The structure of this method is nicely symmetric, but the naming seems inconsistent (e.g. payload vs encoded_response). How about request_plaintext_bytes and response_plaintext_bytes or something like that?

remote_attestation_sessions/src/lib.rs
```diff
@@ -16,129 +16,13 @@
 
 //! Logic for performing remote attestation in multiple sessions
 
-#![no_std]
+// The client logic requires std.
+#![cfg_attr(not(feature = "client"), no_std)]
```
conradgrobler, line 20: Although I think it is fine for now, it would be nice if this is split further into client, server and shared crates as part of future refactoring. That refactoring would also hopefully mean that the untrusted loader would not have to have a dependency on the remote attestation and crypto crates.

grpc_unary_attestation/src/client.rs
conradgrobler, line 27: nit: a doc comment to clarify the intended use might be useful.

conradgrobler (approved):

remote_attestation_sessions/src/client.rs
jul-sh: Agreed, it's one if the items I listed in: https://github.com/project-oak/oak/issues/2608

remote_attestation_sessions/src/client.rs
jul-sh: If not the handshake is done. This goes back to your previous comment, there's some implicit coupling between this and `is_completed()`. Back in the day I listed this in the refactoring ticket: https://github.com/project-oak/oak/issues/2608

jul-sh: Also pre-existent, this PR just moves the method but does not really modify it

remote_attestation_sessions/src/lib.rs
jul-sh: Mirroring the other remote attestation related crates that have a client server module structure. Agreed that here two crates would be better though.

remote_attestation_sessions/src/client.rs
jul-sh: Can do, but this is pre-existing. This is L90 from grpc_unary_attestation/src/client.rs.

remote_attestation_sessions/src/client.rs
jul-sh: Can do, but likewise pre-existing

remote_attestation_sessions/src/client.rs
jul-sh: Actually can't do, I'm not sure why it uses an inner Option, it seems that this function only bails on the Result level. @ipetr0v do you happen to have more context? :)

remote_attestation_sessions/src/lib.rs
jul-sh: Agreed, updated

remote_attestation_sessions/src/client.rs
jul-sh: done

grpc_unary_attestation/src/client.rs
jul-sh: Agreed about splitting out the std client from the rest of `remote_attestation_sessions`. But why should this be a separate crate? This piece of code is just the existing interface of the gRPC client, now changed to use the `GenericAttestationClient` under the hood. 

The diff is somewhat misleading by showing this as new code, which for the most part it isn't. :)

remote_attestation_sessions/src/client.rs
jul-sh: Agreed that'd be a nicer name, updated. Should note that this naming wasn't changed, this is just a moved line, L106 from grpc_unary_attestation/src/client.rs

remote_attestation_sessions/src/client.rs
jul-sh: Pretty sure it's not needed, removed the inner Option from here (and a few other players that proxied this function).

remote_attestation_sessions/src/client.rs
jul-sh: Actually ended up reverting to `encoded_response` after spending some more time in the remote attestation code. 

Yes, doesn't seem clear in the isolated context here. but pretty sure it was named this way as the bytes are an encoded proto that is decoded in `oak_functions/client/rust/src/lib.rs` L59.

The serialized proto is a bit counterintuitive, it's part of the items listed in: https://github.com/project-oak/oak/issues/2608

grpc_unary_attestation/src/client.rs
```diff
@@ -16,98 +16,70 @@
...
+use oak_remote_attestation_sessions_client::{GenericAttestationClient, UnaryClient};
 use tonic::transport::Channel;
 
-/// gRPC Attestation Service client implementation.
-pub struct AttestationClient {
-    session_id: SessionId,
-    encryptor: Encryptor,
-    client: UnarySessionClient<Channel>,
+/// gRPC implementation of of the `UnaryClient`. Serves as an inner of the
+/// public `AttestationClient`.
```
tiziano88, line 26:
```suggestion
/// gRPC implementation of of [`UnaryClient`]. Serves as an inner of the
/// public [`AttestationClient`].
```

grpc_unary_attestation/src/client.rs
tiziano88: Diff aside (I'm not trying to figure out whether it's pre-existing or not), my point was whether we should try and keep client and server libraries separate. But maybe we don't care that much about the gRPC specific logic, since even in its server side component it's not expected to be part of the bare-metal binary.

tiziano88 (approved):

remote_attestation_sessions/src/client.rs
tiziano88: I understand it's pre-existing, but it still seems not super intuitive to me; I would either rename, or add some comments around if possible.

grpc_unary_attestation/src/client.rs
jul-sh: Oh, I see, the suggestion was about splitting this crate into two as well. :) 

Imo for the gRPC boys one crate is fine. Like you said, it wouldn’t be part of bare metal either way

remote_attestation_sessions/src/client.rs
jul-sh: Happy to add a comment. :) And agreed it's confusing, imo the underlying issue is the asymmetry in what the runtime expects. It expects a raw byte slice in the request, but responds with an (encoded) proto that contains the reply and a status code. There's a bit of an inherent asymmetry.

jul-sh: Added comments to explain :)

jul-sh:
@conradgrobler @tiziano88 thanks for the review!

remote_attestation_sessions/src/client.rs
tiziano88: Thanks, appreciated!

PR #2913, opened 2022-05-31
Add multiboot1 protocol support

andrisaar:
Highly experimental, but I've added multiboot1 support as that's what... a certain VMM nominally supports natively.

As it stands, it's untestable in our current framework as Qemu refuses to boot ELF64 binaries which have a multiboot header (as multiboot is nominally a 32-bit only protocol). You have to recompile qemu if you want to remove that check; if you do, the code seems to work. But I don't really want to build a patched qemu in our container.

However, as it stands, it _should_ be enough to test whether (claimed) multiboot1 support works or not in that other VMM, so I'm inclined to merge it even with the known large caveats.

I've left it as an optional feature, off by default, as otherwise qemu will try using multiboot (and fail because we're a 64-bit binary) before even trying to boot with PVH.

(Technically, PVH is also a 32-bit protocol, so qemu should reject ELF64 binaries here as well... but let's not ask too many questions here.)

### DIFF
--- a/experimental/uefi/baremetal/build.rs
+++ b/experimental/uefi/baremetal/build.rs
@@ -21,6 +21,7 @@ use std::{env, path::PathBuf};
 fn main() {
     println!("cargo:rerun-if-changed=target.json");
     println!("cargo:rerun-if-changed=layout.ld");
+    println!("cargo:rerun-if-changed=../../../third_party/rust-hypervisor-boot/layout.ld");
 
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
 
@@ -38,5 +39,19 @@ fn main() {
         .unwrap();
     bindings
         .write_to_file(out_path.join("start_info.rs"))
+        .unwrap();
+
+    let bindings = bindgen::Builder::default()
+        .clang_arg("--target=x86_64-pc-linux-gnu")
+        .header("src/multiboot.h")
+        .allowlist_type("multiboot_info")
+        .allowlist_type("multiboot_mmap_entry")
+        .use_core()
+        .ctypes_prefix("::core::ffi")
+        .layout_tests(false)
+        .generate()
+        .unwrap();
+    bindings
+        .write_to_file(out_path.join("multiboot.rs"))
         .unwrap();
 }
--- a/experimental/uefi/baremetal/src/main.rs
+++ b/experimental/uefi/baremetal/src/main.rs
@@ -18,7 +18,7 @@
 #![no_main]
 #![feature(lang_items)]
 #![feature(alloc_error_handler)]
-#![feature(bench_black_box)]
+#![feature(core_ffi_c)]
 #![feature(custom_test_frameworks)]
 // As we're in a `no_std` environment, testing requires special handling. This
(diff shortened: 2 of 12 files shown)


experimental/uefi/baremetal/build.rs
```diff
@@ -21,6 +21,7 @@ use std::{env, path::PathBuf};
 fn main() {
     println!("cargo:rerun-if-changed=target.json");
     println!("cargo:rerun-if-changed=layout.ld");
+    println!("cargo:rerun-if-changed=../../../third_party/rust-hypervisor-subset/layout.ld");
```
conradgrobler, line 24: nit: I think this should be "boot" instead of "subset"

experimental/uefi/baremetal/src/main.rs
```diff
@@ -18,34 +18,38 @@
...
 // approach was inspired by https://os.phil-opp.com/testing/.
 #![test_runner(crate::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 
 use core::panic::PanicInfo;
-use rust_hypervisor_firmware_boot::pvh;
+
+mod asm;
 mod hvm_start_info;
+mod multiboot;
```
conradgrobler, line 32: Could this be feature-gated as well?

experimental/uefi/baremetal/src/main.rs
```diff
@@ -18,34 +18,38 @@
...
 #![feature(custom_test_frameworks)]
 // As we're in a `no_std` environment, testing requires special handling. This
 // approach was inspired by https://os.phil-opp.com/testing/.
 #![test_runner(crate::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 
 use core::panic::PanicInfo;
-use rust_hypervisor_firmware_boot::pvh;
+
+mod asm;
```
conradgrobler, line 30: asm seems to contain only multi-boot code. could the feature-gating not happen at this level?

experimental/uefi/baremetal/src/main.rs
```diff
@@ -18,34 +18,38 @@
...
 
 #[no_mangle]
 #[cfg(not(test))]
-pub extern "C" fn rust64_start(rdi: &hvm_start_info::hvm_start_info) -> ! {
-    // Ensure that `PVH_NOTE` is not optimized away, otherwise we won't be able to boot our
-    // kernel when testing.
-    core::hint::black_box(&pvh::PVH_NOTE);
-    kernel::start_kernel(rdi);
+pub extern "C" fn rust64_start(start_info: usize, magic: u64) -> ! {
+    if magic == 0x2BADB002 {
```
conradgrobler, line 44: Is there a reference where this is defined (like a spec)? If so, perhaps add a link in a comment.

conradgrobler (approved):

experimental/uefi/baremetal/src/main.rs
andrisaar: Done. I feature-gated the whole multiboot main entry point as well, as that let me get rid of the ugly unsafe block, below.

Issue #2914, opened 2022-05-31
Support additional attestation info in `ClientIdentity`


tiziano88:
`ServerIdentity` contains a field for storing additional attestation data:

https://github.com/project-oak/oak/blob/b960c7821c775f4176159e9d773d5d3153ad403c/remote_attestation/rust/src/message.rs#L122-L127

but `ClientIdentity` does not seem to, even though the two should be symmetric. 

https://github.com/project-oak/oak/blob/b960c7821c775f4176159e9d773d5d3153ad403c/remote_attestation/rust/src/message.rs#L130-L160

conradgrobler:
We are no longer using the interactive remote attestation, so this is no longer relevant.

Issue #2911, opened 2022-05-30
Identify how to transfer additional data for remote attestation


tiziano88:
We should clarify how we expect the remote attestation protocol to convey additional data, for instance inclusion proofs or intermediate certificates necessary to validate the remote attestation report.

Currently it seems that the field below is meant to be used for this purpose, but I am not sure we actually want this data to originate from inside the TEE, since it seems in most cases it will actually originate from the untrusted part of the runtime. Then again, in some cases there may not be a separate untrusted part of the application, so maybe it makes sense to keep it where it is? But in that case we should have a mechanism for provisioning this data at initialization time.

https://github.com/project-oak/oak/blob/37b17805e43511c446140ee5f0fc0cae3ffdb66c/remote_attestation/rust/src/message.rs#L122-L127

@rbehjati WDYT? How do you expect this to work with respect to Transparent Release Process?

cc @ipetr0v @jul-sh @conradgrobler 

The outcome of this issue should be either code changes (if we decide to go that way), or some additional documentation on how this is expected to work.

rbehjati:
## What to do with additional_info?

We have to clarify what `additional_info` is for. Is this additional info that the server sends to the client to verify, or does it contain additional info that should be included in the attestation report? 

In any case, inclusion proofs (more accurately the endorsement statement and the proof of its inclusion in Rekor (I call these together endorsement evidence)) do not need to be verified as part of the remote attestation handshake, and do not need to be included in the attestation report. Endorsement evidence still needs to be verified prior to sending requests to the server, but not as part of the remote attestation handshake or key exchange protocol. This is also what we agreed on earlier (see #2535). So we should at least update the doc comment to be clear about this. 

For configuration info, we have to discuss the process. I believe it only needs to be included as the data that is passed to the `AttestationGenerator`, and should not be included in `additional_info`, but I am not sure. @ipetr0v What do you think? Does the protocol for some other reason require the `additional_info` field, or can we remove it?

## How to transfer additional data for remote attestation?

Here is a bit more about my understanding of the attestation and verification behaviour, and some design ideas that will lead to code changes. 

We have two flavours of remote attestation. One is [offline attestation](https://github.com/project-oak/oak/tree/main/experimental/offline_attestation), the other is [interactive attestation](https://github.com/project-oak/oak/tree/main/remote_attestation).

[...]

Issue #2842, opened 2022-05-13
Add support for generating a remote attestation report from hardware


conradgrobler:
The TEE measurement and related information should be generated by the TEE hardware.

conradgrobler:
The first target will be AMD SEV-SNP. See section 7.3 in https://www.amd.com/system/files/TechDocs/56860.pdf for more information.

conradgrobler:
We can now successfully generate an AMD SEV-SNP atestation report.

PR #2915, opened 2022-05-31
Encapsulate Remote Attestation generation and verification

### LINKED ISSUE #2909: Encapsulate Remote Attestation generation and verification as Rust traits
It may be useful at some point to abstract the remote attestation logic more nicely, perhaps with a trait that represents the underlying platform, and its ability to generate a remote attestation quote.

Something like 

```rust
trait RemoteAttestationGenerator {
  fn generate_remote_attestation(data: &[u8]) -> Result<Vec<u8>>;
}
```

Then specific implementation of this (which would delegate to the underlying platform, e.g. AMD-SEV-SNP) would be plugged in.

For instance, it would replace this logic with something nicer:

https://github.com/project-oak/oak/blob/066b1d7a8c9ba0829a7a52ae3d86a35d7b0e2122/remote_attestation_sessions/src/lib.rs#L39-L40

https://github.com/project-oak/oak/blob/066b1d7a8c9ba0829a7a52ae3d86a35d7b0e2122/remote_attestation_sessions/src/lib.rs#L70-L75

(and probably correspondingly on the client side).

### LINKED ISSUE #2910: Deduplicate Remote Attestation Report data structures
https://github.com/project-oak/oak/blob/37b17805e43511c446140ee5f0fc0cae3ffdb66c/experimental/attestation_common/src/report.rs#L31-L50

https://github.com/project-oak/oak/blob/37b17805e43511c446140ee5f0fc0cae3ffdb66c/remote_attestation/proto/remote_attestation.proto#L24-L41

@ipetr0v or is there any reason why they are distinct?

### LINKED ISSUE #2919: ServerHandshaker does not verify transcript signature unless configured with peer attestation behavior
https://github.com/project-oak/oak/blob/b43decc6069103a256c368d9c7cef85e6b9252fe/remote_attestation/rust/src/handshaker.rs#L508-L533

This seems wrong, since the transcript should always be checked.

See also #2918.

tiziano88:
- various tech debt repayments throughout the codebase
- create separate crate with placeholder implementation of AMD-specific remote attestation functionality
- remove explicit TEE certificate argument that was previously passed everywhere; this functionality (and corresponding root cert) is now expected to be implemented in the remote attestation verification trait (the generation will be done in hardware by the generator trait); e.g. an Intel verifier will have the Intel root cert baked in, while an AMD verifier will have the AMD root cert baked in.
- clarify terminology `attestation_info` vs `attestation_report`
- remove ConfigurationInfo verifier; this was not doing anything anyways, and will need to be designed correctly as part of #2911 
- fix documentation throughout (mainly links and formatting)
- use generic types to determine the remote attestation generation and verification (so that they can be reified and optimized at compile time)
- correctly account for ephemeral public key, signing key, and additional data in attestation report (previously this was inconsistent and possibly incorrect)
- use a proper way of hashing multiple values of different lenghts (by hashing individual values first)
- make client and server attestation more symmetric (though there are still asymmetries, see #2914)
- Fix #2910
- Fix #2909
- Fix #2919


tiziano88:
Reviewers: I realise this is quite a large PR, I had to touch quite a few layers in order for the refactoring to be effective (and I managed to discover / fix a few issues along the way).

As I will be OOO next week, I would appreciate it if you could prioritize reviewing this PR, with the aim for me to merge it on 2022-06-01 ; if anything looks unclear (or wrong), please comment liberally, and I'll try to clarify things by adding more comments.

Thanks in advance!

experimental/attestation_common/src/certificate.rs
```diff
@@ -53,9 +51,7 @@ const DEFAULT_DNS_NAME: &str = "localhost";
 /// Indicates whether to add a custom TEE extension to a certificate.
 #[derive(PartialEq)]
 pub enum AddTeeExtension {
-    /// Enum value contains a PEM encoded TEE Provider's X.509 certificate that signs TEE firmware
-    /// key.
-    Yes(Vec<u8>),
```
tiziano88, line 58: This is not necessary any more, since the TEE cert should be handled by the Verifier and Generator trait impls.

experimental/attestation_common/src/keying_material.rs
```diff
@@ -30,41 +33,31 @@ const ASSERTION_TYPE: &str = "amd_sev_snp_report";
 /// Assertion used for remote attestation.
 #[derive(Serialize, Deserialize)]
 pub struct Assertion {
-    attestation_info: AttestationInfo,
```
tiziano88, line 33: Renamed to attestation_report, since that's what it is normally called. We still need to figure out what to do with the additional data, but that's for another PR.

experimental/attestation_common/src/report.rs
```diff
@@ -25,75 +24,11 @@ use x509_parser::der_parser::{self, oid::Oid};
 //
 // <https://www.alvestrand.no/objectid/2.16.840.1.113730.1.13.html>
 pub const TEE_EXTENSION_OID: Oid<'static> = der_parser::oid!(2.16.840 .1 .113730 .1 .13);
-// TODO(#1867): Add remote attestation support.
-const TEST_TEE_MEASUREMENT: &str = "Test TEE measurement";
 
-/// Placeholder implementation of TEE report for remote attestation.
```
tiziano88, line 31: Moved to the AMD specific crate.

experimental/tf_proxy/client/src/main.rs
```diff
@@ -73,10 +73,7 @@ async fn main() -> anyhow::Result<()> {
     env_logger::init();
     let opt = Opt::parse();
 
-    // Config is not relevant for the proxy for now.
-    let config_verifier = |_: ConfigurationInfo| Ok(());
```
tiziano88, line 77: Most of these were not even very realistic placeholders.

experimental/tls_attestation/src/server.rs
```diff
@@ -43,13 +43,9 @@ async fn proxy(
 
 /// Runs HTTPS proxy server that provides self-signed TLS certificates with TEE attestation
 /// credentials inside a custom X.509 extension.
-pub async fn run_server(
-    address: &str,
-    app_uri: Uri,
-    tee_certificate: Vec<u8>,
```
tiziano88, line 49: This is subsumed in the attestation generator trait now.

experimental/uefi/client/src/main.rs
```diff
@@ -50,10 +47,9 @@ async fn chat(client: &mut AttestationClient, message: String) -> anyhow::Result
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let cli = Args::parse();
 
-    let mut client =
-        AttestationClient::create(&cli.server, TEE_MEASUREMENT, Box::new(|_config| Ok(())))
```
tiziano88, line 54: More placeholders that were not particularly meaningful, now this logic is centralized in the new trait (and probably we will have to add something to allow passing additional predicates at some point).

oak_functions/client/java/src/com/google/oak/functions/client/ServerConfigurationVerifier.java
```diff
@@ -52,11 +52,11 @@ public ServerConfigurationVerifier(
    * @return the result of calling test {@code configurationVerifier} as described above.
    */
   public boolean verify() {
-    byte[] configBytes = serverIdentity.getAdditionalInfo();
```
tiziano88, line 55: Renamed these fields for consistency with the Rust structs.

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
```diff
@@ -188,36 +187,50 @@ public AeadEncryptor getEncryptor() throws RuntimeException {
...
+    byte[] additionalAttestationData = serverIdentity.getAdditionalAttestationData();
+
+    // Ensure additional attestation data was supplied.
+    if (!verifyAdditionalAttestationData(additionalAttestationData)) {
+      logger.log(Level.WARNING, "Additional attestation data was invalid");
       return false;
     }
 
     // Check the hash of the public key and additional info
-    AttestationInfo attestationInfo = AttestationInfo.parseFrom(
```
tiziano88, line 199: I'm not sure why the ephemeral public key was previously not included. @ipetr0v do you remember if that was intended?

remote_attestation/rust/Cargo.toml
```diff
@@ -7,45 +7,25 @@ license = "Apache-2.0"
...
 aes-gcm = { version = "*", optional = true }
 anyhow = { version = "*", default-features = false }
 bytes = { version = "*", default-features = false }
-getrandom = { version = "*", optional = true }
 hkdf = { version = "*", optional = true }
-log = "*"
 p256 = { version = "*", default-features = false, optional = true, features = [
   "ecdsa"
 ] }
-prost = { version = "*", default-features = false, features = ["prost-derive"] }
```
tiziano88, line 33: No more proto deps.

remote_attestation/rust/src/handshaker.rs
```diff
@@ -505,33 +484,45 @@ impl ServerHandshaker {
         client_identity: ClientIdentity,
         key_negotiator: KeyNegotiator,
     ) -> anyhow::Result<()> {
-        if self.behavior.contains_peer_attestation() {
```
tiziano88, line 508: I think this logic was incorrect: if the behavior does not have the expected peer attestation, then nothing was being checked. Instead, I think only the attestation report check should be skipped, and in fact even that now is much stricter (if the binary does not expect a remote attestation from the other side, then it will fail if there is one, so that it cannot be accidentally ignored so easily).

remote_attestation/rust/src/handshaker.rs
```diff
@@ -589,89 +583,86 @@ impl Encryptor {
...
-    /// Expected value of the peer's TEE measurement.
-    expected_tee_measurement: Option<Vec<u8>>,
-    // /// Convenience struct for creating attestation info and signing data with TEE
-    // /// firmware key.
-    // tee_provider: Option<TeeProvider>,
-    /// PEM encoded X.509 certificate that signs TEE firmware key.
-    tee_certificate: Option<Vec<u8>>,
-    /// Signer containing a key which public part is signed by the TEE firmware key.
-    /// Used for signing protocol transcripts and preventing replay attacks.
-    signer: Option<Signer>,
```
tiziano88, line 602: This was moved one level up.

remote_attestation/rust/src/handshaker.rs
```diff
@@ -589,89 +583,86 @@ impl Encryptor {
...
 ///   - Represents an attestation process, where current machine and a remote peer remotely attest
 ///     each other.
-pub struct AttestationBehavior {
-    /// Expected value of the peer's TEE measurement.
-    expected_tee_measurement: Option<Vec<u8>>,
-    // /// Convenience struct for creating attestation info and signing data with TEE
-    // /// firmware key.
-    // tee_provider: Option<TeeProvider>,
-    /// PEM encoded X.509 certificate that signs TEE firmware key.
-    tee_certificate: Option<Vec<u8>>,
```
tiziano88, line 599: Removed.

remote_attestation/rust/src/lib.rs
```diff
@@ -18,14 +18,8 @@
 
 extern crate alloc;
 
-pub mod proto {
```
tiziano88, line 21: No more protos.

remote_attestation/rust/src/message.rs
```diff
@@ -114,49 +114,35 @@ pub struct ServerIdentity {
...
 }
 
 /// Client identity message containing remote attestation information and a public key for
 /// Diffie-Hellman key negotiation.
 #[derive(Clone, PartialEq)]
 pub struct ClientIdentity {
-    /// Public key needed to establish a session key.
+    /// See [`ServerIdentity::ephemeral_public_key`].
     pub ephemeral_public_key: [u8; KEY_AGREEMENT_ALGORITHM_KEY_LENGTH],
-    /// Signature of the SHA-256 hash of all previously sent and received messages.
```
tiziano88, line 136: No point repeating all the comments everywhere.

tiziano88 (commented):
Some comments to help the review process.

oak_remote_attestation_amd/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+            measurement: TEST_TEE_MEASUREMENT.to_string().as_bytes().to_vec(),
+            data: data.to_vec(),
+            ..Default::default()
+        }
+    }
+}
+
+impl PlaceholderAmdReport {
+    pub fn from_string(input: &str) -> anyhow::Result<Self> {
+        serde_json::from_str(input)
```
conradgrobler, line 110: This can be fixed as part of #2842, but json is probably not a useful way to send the report to the client. For clients to verify the signature they would need to rebuild the exact bytes of the original report from the json encoding, so we could just send the bytes of the original report to save space and complexity.

oak_remote_attestation_amd/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub platform_version: u8,
+    /// Arbitrary data to put into the TEE report.
+    pub data: Vec<u8>,
+    /// TEE measurement, i.e. VM hash.
+    pub measurement: Vec<u8>,
+    /// Signature of this report.
+    pub signature: Vec<u8>,
+}
+
+// TODO(#1867): Add remote attestation support.
```
conradgrobler, line 94: #1867 has been closed, so should probalby not exist in TODOs (multiple places throughout the PR). A new issue (#2842) has been created to cover most of it.

remote_attestation/rust/Cargo.toml
```diff
@@ -7,45 +7,25 @@ license = "Apache-2.0"
...
-# Restrict the rand_core to a version that is compatible with x25519-dalek.
-rand_core = { version = "*", optional = true, default-features = false, features = [
```
conradgrobler, line 35: I thought the getrandom and rand_core dependencies are needed for the bare-metal binary, so I am surprised that the bare-metal binaries build.

remote_attestation/rust/src/handshaker.rs
```diff
@@ -696,72 +687,37 @@ impl Transcript {
...
 }
 
-pub fn serialize_protobuf<M: prost::Message>(message: &M) -> anyhow::Result<Vec<u8>> {
-    let mut message_bytes = Vec::new();
-    message
-        .encode(&mut message_bytes)
-        .map_err(anyhow::Error::msg)
-        .context("Couldn't serialize Protobuf message to bytes")?;
-    Ok(message_bytes)
+/// Compute a hash over values of possibly different lenght.
```
conradgrobler, line 711: nit: "length"

conradgrobler (commented):
Nice. It is looking much better.

I noticed that the kokoro job is failing due to the EUFI app's remote attestation handshake. I guess there is some compatibility issue between that and the current loader implementation.

conradgrobler (approved):

remote_attestation/rust/Cargo.toml
tiziano88: Soo this was... complicated. It turns out that our remote attestation crate only actually includes the rust-crypto parts if rust-crypto is enabled *and* ring-crypto is not:

https://github.com/project-oak/oak/blob/241a59e4dd5f1e405822daff190a77f4333893ad/remote_attestation/rust/src/crypto/mod.rs#L27-L28

And there is a compile time check that at least one is enabled:

https://github.com/project-oak/oak/blob/241a59e4dd5f1e405822daff190a77f4333893ad/remote_attestation/rust/src/crypto/mod.rs#L22-L23

What is *not* there is a similar check to ensure that they are not both enabled at the same time (which I now added).

In practice they were both getting enabled, and therefore the ring-crypto would be chosen. Still not entirely sure how that actually worked though.

tiziano88: cc @andrisaar

remote_attestation/rust/src/handshaker.rs
tiziano88: Done.

oak_remote_attestation_amd/src/lib.rs
tiziano88: Yeah I think this JSON stuff is only here because of the x509 extension approach that @ipetr0v implemented a while ago. I agree we should not have JSON eventually, added a TODO.

oak_remote_attestation_amd/src/lib.rs
tiziano88: Done.

tiziano88 (commented):
Thanks for the review

tiziano88:
> Will wait with merging #2896 until after this PR is merged, as #2896 will need some (small) changes to adapt to this.

Thanks Juliette, I appreciate you facilitating this!

grpc_unary_attestation/src/client.rs
```diff
@@ -63,16 +66,23 @@ pub struct AttestationClient {
 }
 
 impl AttestationClient {
-    pub async fn create(
+    /// Create an [`AttestationClient`] with a [`PlaceholderAmdAttestationVerifier`].
+    pub async fn create(uri: &str) -> anyhow::Result<Self> {
+        Self::create_with_attestation_verifier(uri, PlaceholderAmdAttestationVerifier).await
```
ipetr0v, line 71: Will this placeholder later become an argument for `create` (once we have attestation)?
Or it will be hardcoded?

ipetr0v: ah, I see the next function now

oak_remote_attestation_amd/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Security version number of SNP firmware.
+    pub svn: u8,
+    /// The installed version of firmware.
+    pub platform_version: u8,
+    /// Arbitrary data to put into the TEE report.
+    pub data: Vec<u8>,
+    /// TEE measurement, i.e. VM hash.
+    pub measurement: Vec<u8>,
+    /// Signature of this report.
+    pub signature: Vec<u8>,
```
ipetr0v, line 90: IIUC this is a signature **by the** hardware key.
But there should also be the signature **of the** hardware key done by AMD. So that the client can use the root AMD key to verify all other signatures.

remote_attestation/rust/src/tests/handshaker.rs
```diff
@@ -15,54 +15,68 @@
...
+    }
+}
+
+#[derive(Clone)]
+struct TestAttestationVerifier;
+
+impl AttestationVerifier for TestAttestationVerifier {
+    fn verify_attestation(
+        &self,
+        attestation: &[u8],
```
ipetr0v, line 47: Could you please add a comment, what `attestation` means here?

ipetr0v (approved):

oak_remote_attestation_amd/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// TODO(#2842): Add remote attestation support.
+#[derive(Clone)]
+pub struct PlaceholderAmdAttestationVerifier;
+
+impl AttestationVerifier for PlaceholderAmdAttestationVerifier {
+    fn verify_attestation(
+        &self,
+        attestation: &[u8],
```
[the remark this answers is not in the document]
tiziano88: Added a bit of detail in the trait definition (I don't think we should duplicating documentation around, that's part of what this PR is trying to address).

tiziano88 (commented):
Thanks

remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java
tiziano88: Ok thanks, I think we should probably re-think this part anyways (adding or removing things that get signed). The good news is that now this is defined in a single place (well, two, one for Rust and one for Java).

PR #2921, opened 2022-06-02
Add useful extensions to the devcontainer

jul-sh:


tiziano88 (approved):
Nice, thanks

andrisaar:
Drive-by comment, while you're editing the file anyway: I think the `matklad.rust-analyzer` extension has been replaced by `rust-lang.rust-analyzer`. We should switch to the new one.

PR #2930, opened 2022-06-06
Build the Multiboot1 header in pure Rust

andrisaar:
The exercise here was three-fold:
a) I wanted to see if I could make the data structure more Rust-like (instead of just relying on `u32`s as you'd get from a bindgen-generated implementation), which I think looks much better (and now I'm thinking of doing this Rust-ish rewrite for other data structures as well instead of relying on bindgen);
b) getting rid of the raw assembler as it's not required;
c) I was bored waiting for my next plane and wanted something small to tinker with.


jul-sh (approved):
nice! thank you

experimental/uefi/baremetal/src/multiboot.rs
```diff
@@ -55,3 +56,59 @@ impl BootInfo<multiboot_mmap_entry> for multiboot_info {
...
+        const MULTIBOOT_VIDEO_MODE = 0x00000004;
+
+        /// This flag indicates the use of the address fields in the header.
+        const MULTIBOOT_AOUT_KLUDGE = 0x00010000;
+    }
+}
+
+/// Multiboot header.
+///
+/// The Multiboot header must be contained completely within the first 8192
```
conradgrobler, line 88: nit: this block seems narrower than the max width or 100 chars. also line 102-103.

conradgrobler (approved):
Looks much nicer than the assembly.

PR #2933, opened 2022-06-06
Add virtio console driver

conradgrobler:
This PR adds a simple virtio PCI console driver. It only support a single port and does not support any configuration, such as console size.

It is used as the communication channel when running the bare-metal runtime under QEMU.


experimental/uefi/kernel/src/lib.rs
```diff
@@ -72,8 +70,11 @@ fn main<E: boot::E820Entry, B: boot::BootInfo<E>>(info: &B) -> ! {
 
 // Use a serial port for the communications channel if we don't support virtio vsock.
```
mariaschett, line 71: Is this comment still up-to-date?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+///
+/// See <https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-39000010>.
+pub struct Console {
+    /// The base virtio-over-PCI device used for configuration and notification.
+    device: VirtioBaseDevice<VirtioPciTransport>,
+    /// The receive queue.
+    rx_queue: DeviceWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// The transmit queue.
+    tx_queue: DriverWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// A temporary buffer to store extra data from a buffer that was not fully read when using
```
mariaschett, line 58: Why is a buffer needed? Can we use it only for buffering data for reading? Is the buffer temporary (i.e., do we aim to remove it) or does it only temporarily hold data?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+const PCI_DEVICE_ID: u16 = 0x1040 + DEVICE_ID;
+
+/// Simple driver implementation a virtio serial/console device that only supports a single port and
+/// no configutration.
+///
+/// See <https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-39000010>.
+pub struct Console {
+    /// The base virtio-over-PCI device used for configuration and notification.
+    device: VirtioBaseDevice<VirtioPciTransport>,
+    /// The receive queue.
```
mariaschett, line 54: What does the receive/transmit queue hold?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+        let transport = VirtioPciTransport::new(pci_device);
+        let device = VirtioBaseDevice::new(transport);
+        let mut result = Self::new(device);
+        result.init()?;
+        // Let the device know there are available buffers in the receiver queue.
+        result.device.notify_queue(RX_QUEUE_ID);
+        Ok(result)
+    }
+
+    /// Reads the next available bytes from the receive queue, if any are available.
```
mariaschett, line 79: Should we also comment that it notifies the device that buffer is available too after reading?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+        // For now we just scan the first 32 devices on PCI bus 0 to find the first one that matches
+        // the vendor ID and device ID.
+        let pci_device = find_device(super::PCI_VENDOR_ID, PCI_DEVICE_ID)
+            .ok_or_else(|| anyhow::anyhow!("Couldn't find a virtio console device."))?;
+        let transport = VirtioPciTransport::new(pci_device);
+        let device = VirtioBaseDevice::new(transport);
+        let mut result = Self::new(device);
+        result.init()?;
+        // Let the device know there are available buffers in the receiver queue.
+        result.device.notify_queue(RX_QUEUE_ID);
```
mariaschett, line 75: nit: why is `result.device.notify_queue` not part of `init`? I am not 100% sure I understand what the state after init is.

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+    rx_queue: DeviceWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// The transmit queue.
+    tx_queue: DriverWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// A temporary buffer to store extra data from a buffer that was not fully read when using
+    /// `read_all`.
+    pending_data: Option<VecDeque<u8>>,
+}
+
+impl Console {
+    /// Finds the virtio console PCI device, initialises the device, and configures the queues.
```
mariaschett, line 64: Clarification about `find_and_configure_device`. When I first read the function name I thought I have to provide (a) a place to find and (b) a configuration---luckily I did need neither :) (I think). 

How about `start_device`? Is that what I do?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+            .map_err(|error| anyhow::anyhow!("Queue configuration error: {:?}", error))
+            .context("Couldn't configure the transmit queue.")?;
+        self.device
+            .complete_init()
+            .map_err(|error| anyhow::anyhow!("Device activation error: {:?}", error))
+            .context("Couldn't activate the PCI device.")?;
+        Ok(())
+    }
+
+    /// Tries once to fill the destination with as much data as is currently available, either in
```
mariaschett, line 158: Is there a reason we try the pending buffer first?

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+        let data_len = data.len();
+        while start < data_len {
+            let count = self.write_bytes(&data[start..]).unwrap_or(0);
+            start += count;
+        }
+        Ok(())
+    }
+
+    fn flush(&mut self) -> Result<(), Self::Error> {
+        // We always flush on write, so do nothing.
```
mariaschett, line 210: Do we flush on write?

mariaschett (approved):
Thank you, very well commented, I really enjoyed it!

Out of interest: Is it possible/worth it to add tests?

conradgrobler (commented):
Thanks for the review.

Adding tests requires a fake/mock device transport implementation (the current device assumes a valid PCI interface). I will add that as part of a future PR.

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,215 @@
...
+            self.pending_data.replace(source);
+        }
+        Some(position)
+    }
+}
+
+impl ciborium_io::Read for Console {
+    type Error = anyhow::Error;
+
+    fn read_exact(&mut self, data: &mut [u8]) -> Result<(), Self::Error> {
```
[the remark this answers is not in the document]
mariaschett: True, I think I misread the `unwrap_or`! Thanks for the clarification!

experimental/virtio/src/console.rs
mariaschett: Great, maybe we want to make it explicit that we first want to read the pending buffer, i.e., "first in the pending buffer" and "then from the next available buffer". 

Just a short clarification: If we cannot fully read `pending_data`, we move the remaining `pending_data` back to `pending_data`.

experimental/virtio/src/console.rs
```diff
@@ -0,0 +1,223 @@
...
+///
+/// See <https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-39000010>.
+pub struct Console {
+    /// The base virtio-over-PCI device used for configuration and notification.
+    device: VirtioBaseDevice<VirtioPciTransport>,
+    /// The receive queue, used for receiving bytes.
+    rx_queue: DeviceWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// The transmit queue, used for sending bytes.
+    tx_queue: DriverWriteOnlyQueue<QUEUE_SIZE, DATA_BUFFER_SIZE>,
+    /// A temporary buffer to store extra data from the device that was not fully read when using
```
mariaschett, line 58: nit: If we write "A buffer to temporarily store extra data" we cannot misread it as the buffer being temporary (I think).

mariaschett (approved):

experimental/virtio/src/console.rs
conradgrobler: done.

experimental/virtio/src/console.rs
conradgrobler: Added more clarification to the comments. I think "first in the pending buffer" and "then from the next available buffer" could be misleading. The "either/or" nature is important.

PR #2939, opened 2022-06-08
oak_idl: Improve error model

jul-sh:
Aligns the oak_idl error model a bit closer to what gRPC does. This means:
- Error codes represent a numeric ID, with a few missing ones added
- Added an error message for debugging

Ideally we imo should fully align the IDL error model with the gRPC status model: https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto

Specifically this would mean replacing our error codes, and returning a status object for both successful responses and errors.

Doing so would require larger changes to the IDL, and imo is not a priority right now.

### DIFF
--- a/oak_idl_tests/tests/test_schema.rs
+++ b/oak_idl_tests/tests/test_schema.rs
@@ -43,9 +43,12 @@ impl test_schema::TestService for TestServiceImpl {
             &mut b,
             &test_schema::LookupDataResponseArgs { value },
         );
-        let b = b
-            .finish(m)
-            .map_err(|_| oak_idl::Error::new(oak_idl::ErrorCode::InternalError))?;
+        let b = b.finish(m).map_err(|err| {
+            oak_idl::Error::new_with_message(
+                oak_idl::ErrorCode::InternalError,
+                format!("failed to build response: {:?}", err),
+            )
+        })?;
         Ok(b)
     }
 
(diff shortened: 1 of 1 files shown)


oak_idl/src/lib.rs
```diff
@@ -18,21 +18,61 @@
...
 
+// TODO(#2500): Align with gRPC Status.
+// Ref: https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto
+/// Logical error model.
 #[derive(Debug)]
-pub enum Error {
+pub struct Error {
+    pub code: ErrorCode,
+    /// English message that helps developers understand and resolve the error.
+    pub message: Option<String>,
```
conradgrobler, line 33: nit: I personally think having only a string is sufficient as it can be empty. That is unless there is a logical distinction between None and Some(""), in which case it would be worth documenting the difference.

oak_idl_gen_services/src/lib.rs
```diff
@@ -245,15 +245,15 @@ fn generate_client_method(rpc_call: &RPCCall) -> anyhow::Result<Vec<String>> {
             response_type(rpc_call)
         ),
         format!(
-            "        flatbuffers::root::<{}>(request_body).map_err(|_err| oak_idl::Error::InvalidRequest)?;",
+            "        flatbuffers::root::<{}>(request_body).map_err(|_err| oak_idl::Error {{ code: oak_idl::ErrorCode::InvalidRequest, message: None }})?;",
```
conradgrobler, line 248: I think implementing `Error::new` that takes just a code and sets an empty message, and `Error::new_with_message` will make this a bit cleaner.

Also, should the original error not be printed into the message, rather than ignoring it?

conradgrobler (approved):

oak_idl/src/lib.rs
jul-sh: Thanks, that's a good point. Let's go with just `String` then.

oak_idl_gen_services/src/lib.rs
jul-sh: thanks, addressed

PR #2941, opened 2022-06-08
Add tests for virtio devices

conradgrobler:
This PR also adds a test-specific VirtioTransport implementation to support testing of virtio devices without having to emulate a full PCI device,


jul-sh (approved):
nice! thank you

PR #2942, opened 2022-06-08
Add `wasm_hash` to `WasmHandler`

### LINKED ISSUE #2940: Store `wasm_hash` in WasmHandler
We don't keep the `wasm_module_bytes` around [after we instatiate the Wasmi module](https://github.com/project-oak/oak/blob/0ef1c81e687eef3ef10962d01ed0379a22c1218f/oak_functions/wasm/src/lib.rs#L430). 

I suggest we extend the [`WasmHandler`](https://github.com/project-oak/oak/blob/0ef1c81e687eef3ef10962d01ed0379a22c1218f/oak_functions/wasm/src/lib.rs#L413) to  store the `wasm_hash` permanently instead of only keeping it in the [config report](https://github.com/project-oak/oak/blob/0ef1c81e687eef3ef10962d01ed0379a22c1218f/oak_functions/loader/src/lib.rs#L154).

This also simplifies the interface to the grpc server, as [it wouldn't need to be given the `config_report` as an argument](https://github.com/project-oak/oak/blob/0ef1c81e687eef3ef10962d01ed0379a22c1218f/oak_functions/loader/src/lib.rs#L162).

mariaschett:
Fixes #2940 .

### DIFF
--- a/oak_functions/wasm/src/lib.rs
+++ b/oak_functions/wasm/src/lib.rs
@@ -30,6 +30,7 @@ use hashbrown::HashMap;
 use oak_functions_abi::proto::{ExtensionHandle, OakStatus, Request, Response, StatusCode};
 use oak_functions_extension::{ExtensionFactory, OakApiNativeExtension};
 use oak_logger::{Level, OakLogger};
+use oak_remote_attestation::crypto::get_sha256;
 use wasmi::ValueType;
 
 const MAIN_FUNCTION_NAME: &str = "main";
@@ -415,6 +416,8 @@ pub struct WasmHandler<L: OakLogger> {
     // cloneable.
     module: Arc<wasmi::Module>,
     extension_factories: Arc<Vec<Box<dyn ExtensionFactory<L>>>>,
+    // The hash of the bytes loaded in the Wasm module.
+    wasm_hash: Vec<u8>,
     logger: L,
 }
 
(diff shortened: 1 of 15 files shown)


oak_functions/wasm/src/lib.rs
```diff
@@ -30,6 +30,7 @@ use hashbrown::HashMap;
 use oak_functions_abi::proto::{ExtensionHandle, OakStatus, Request, Response, StatusCode};
 use oak_functions_extension::{ExtensionFactory, OakApiNativeExtension};
 use oak_logger::{Level, OakLogger};
+use oak_remote_attestation::crypto::get_sha256;
```
mariaschett, line 33: @conradgrobler I am not sure, that adding this dependency here is a good idea. Do you agree?

conradgrobler: I don't think a reference to the remote attestation create should be added to this crate. The runtime should generate the hash before creating the wasm instance rather than expecting the wasm crate to generate a hash.

conradgrobler: `ring` cannot be used in bare-metal environments, and rust crypto `sha2` cannot be used in uefi apps , which is why we have the two implementations.

I am not sure the reason for moving the hash into the wasm handler, as it does not seem like something it should be responsible for. Can you give a bit more context for this refactoring?

mariaschett: Ah, thank you, Conrad! 

re: refactoring: I am trying to simplify the way we configure/pass around the configuration around the Oak Functions Runtime. 

I thought we could easily compute the `wasm_hash` from the bytes within the WasmHandler because giving it as an argument is just an additional point for inconsistency.

But thanks for reminding me that the uefi/bare-metal make that impossible, so I'll give it as argument to WasmHandler.

mariaschett: I don't think adding the `wasm_hash` as argument to WasmHandler as actually improves the code, so I am abandoning this PR.

I still think it would be nice if the WasmHandler stores some introspective about the running Wasm Code even after it was initially created (hence storing the `wasm_hash`), but there is no clear reason for it at the moment.

PR #2943, opened 2022-06-08
Fix and run Bazel Test again

### LINKED ISSUE #2538: Cannot run bazel tests from Docker image
```console
$ bazel build //oak_functions/...:all

ERROR: While resolving toolchains for target //oak_functions/client/android:client_activity: no matching toolchains found for types @bazel_tools//tools/android:sdk_toolchain_type
ERROR: Analysis of target '//oak_functions/client/android:client_activity' failed; build aborted: no matching toolchains found for types @bazel_tools//tools/android:sdk_toolchain_type
```

Do I need to run it from the Android Docker image instead, or am I missing some flags? Where (in our CI) are these Bazel targets built / tested at the moment?

mariaschett:
Fixes #2538 .


xtask/src/main.rs
```diff
@@ -755,7 +748,13 @@ fn run_bazel_build() -> Step {
         name: "bazel build".to_string(),
         command: Cmd::new(
             "bazel",
-            &["build", "--", "//remote_attestation/java/...:all"],
+            // Exclude Android specific targets because the will not run in the main Docker Image.
```
rbehjati, line 751: Perhaps this comment should be moved to where `tag=["manual"]` is added?

rbehjati, line 751:
```suggestion
            // Exclude Android specific targets because they will not run in the main Docker Image.
```

rbehjati (approved):
Nice. Thanks :)

xtask/src/main.rs
mariaschett: Done, thank you :)

PR #2944, opened 2022-06-08
Use the Oak IDL to encode communication between the bare-metal runtime and loader

jul-sh:
- Updates frames to include method_id/status
- Replaces the use `SerializableRequest` to serialize user requests with the oak IDL& flatbuffer


experimental/uefi/loader/src/main.rs
```diff
@@ -155,19 +187,47 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Use a bmrng channel for serial communication. The good thing about spawning a separate
     // task for the serial communication is that it serializes (no pun intended) the communication,
     // as right now we don't have any mechanisms to track multiple requests in flight.
-    let (tx, mut rx) = bmrng::unbounded_channel::<Vec<u8>, anyhow::Result<Vec<u8>>>();
+    let (tx, mut rx) =
+        bmrng::unbounded_channel::<SessionRequest, Result<Vec<u8>, oak_idl::Error>>();
```
jul-sh, line 191: This type signature of the channel is essentially the same as that of `handle_user_request` client method generated by the ID. It basically just proxies this client method using non-flatbuffer/oak_idl types (while still using oak_idl Error types).

I'm not a huge fan of this. It'd be far nicer if we could use the actual IDL generated client where messages are sent into this channel. It would allow us to get rid of the `SessionRequest` struct altogether (it's contains the same information as flatbuffer `UserRequest` table).

The reason we cannot right now is that it would require the client code generated by the IDL to support asynchronous invocations. We also cannot just send the flatbuffer types over the channel, due to lifetime bounds imposed by flatbuffer.

oak_idl_gen_services/src/lib.rs
```diff
@@ -151,6 +151,8 @@ fn generate_service(service: &Service) -> anyhow::Result<String> {
         format!(""),
         format!("extern crate alloc;"),
         format!(""),
+        format!("use alloc::string::ToString;"),
+        format!(""),
```
jul-sh, line 155: Required when compiling to a no_std binary

experimental/uefi/loader/src/main.rs
jul-sh: Refactoring the IDL to generate async capable client invocation handlers is unfortunately non-trivial. It would require making the `Handler` [trait](https://github.com/project-oak/oak/blob/b2829838caabe3967f4a332528d084f1620e988f/oak_idl/src/lib.rs#L117) async. 

But rust does not natively support async traits. The common work-around, [async-trait](https://crates.io/crates/async-trait), does not work in no_std environments.

conradgrobler: I guess this is OK for now. The bmrng channel will go away soon, as it is based on a FIFO assumption that will not remain the valid when we start multiplexing other requests over the same channel (e.g. update lookup data). This code can be cleaned as part of that refactoring.

One way to remove the `SessionRequest` would be to just pass the underlying buffer rather than the flatbuffer struct over the channel and reconstruct it on the other side. This should not be much overhead, as the flatbuffer struct is just a wrapper around the underlying buffer.

conradgrobler (approved):

experimental/uefi/channel/src/lib.rs
```diff
@@ -19,18 +19,53 @@
...
+
+    include!(concat!(env!("OUT_DIR"), "/schema_generated.rs"));
+    include!(concat!(env!("OUT_DIR"), "/schema_services.rs"));
+}
+
 /// Length of the entire frame, including the header.
 pub type FrameLength = u32;
-const FRAME_BYTES_SIZE: usize = 4;
+const FRAME_LENGTH_SIZE: usize = 4;
+static_assertions::assert_eq_size!([u8; FRAME_LENGTH_SIZE], FrameLength);
```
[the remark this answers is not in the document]
jul-sh: I recall in #2873 @tiziano88 mentioned preferring explicitly hardcoded sizes for clarity :)

experimental/uefi/runtime/src/framing.rs
```diff
@@ -15,12 +15,72 @@
...
+    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> anyhow::Result<Vec<u8>>,
+    G: AttestationGenerator,
+    V: AttestationVerifier,
+{
+    fn handle_user_request(
+        &mut self,
+        request_message: &schema::UserRequest,
+    ) -> Result<oak_idl::utils::Message<schema::UserRequestResponse>, oak_idl::Error> {
+        let session_id: SessionId = {
+            let session_id_flatbuffers_array = request_message
```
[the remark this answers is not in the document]
conradgrobler: What does `.get()` on the array return? Is that not a slice?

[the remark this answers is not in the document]
conradgrobler: Actually, ignore that comment. I see `get` indexes into it.

experimental/uefi/loader/src/main.rs
jul-sh: > The bmrng channel will go away soon, as it is based on a FIFO assumption that will not remain the valid when we start multiplexing other requests over the same channel (e.g. update lookup data). This code can be cleaned as part of that refactoring.

In the currently planned case of handling multiplexing on the invocation level that would not that necessarily be the case though? In the initial implementation where all invocation's of the runtime method's would be sent as single frames (and lookup data updates sent over multiple invocations) the FIFO assumption would still hold true. Each incremental lookup data update would be answered by a Ok/Error response. 

Even when later implementing the ability to split messages into multiple frames (eg because we exceed the buffer size of vsock virtio implementation), we wouldn't necessarily need to interleave frames.

conradgrobler: I dont think it would work, unless the bmrng channel is wrapped in some global `Arc<Mutex<>>` so that everything can send the data over a single bmrng channel, including the lookup data. At that point, we could just as well implement it directly with something that is under our control.

Also, if we start splitting client requests into multiple frames then forcing it to send all the requests from one client before sending the request from the next client defeats the point of splitting it into frames in the first place.

experimental/uefi/channel/src/lib.rs
```diff
@@ -19,18 +19,53 @@
...
 }
 
+impl Frame {
+    /// Returns the size of the Frame encoded as bytes. On the wire each frame
+    /// is prefixed with its length. This size includes all of the frame,
+    /// including the length prefix. The length is intentionally not stored as a
+    /// value in the frame struct itself, as this may lead to inconsistent
+    /// information if other values are mutated.
+    fn len(&self) -> anyhow::Result<FrameLength> {
+        FrameLength::try_from(FRAME_HEADER_SIZE + self.body.len())
```
ipetr0v, line 63: Why doesn't `Frame::len` include the `METHOD_SIZE` ?

experimental/uefi/loader/src/main.rs
jul-sh: > One way to remove the SessionRequest would be to just pass the underlying buffer rather than the flatbuffer struct over the channel and reconstruct it on the other side. This should not be much overhead, as the flatbuffer struct is just a wrapper around the underlying buffer.

It doesn't actually work with flatbuffers structs directly, since their structs only ever expose references to the underlying slice. Creating an owned slice/Vector requires copying, and hence overhead.

Looked into it a bit more though, and we already do this overhead in the IDL message struct. It's a convience struct that does contain an owned Vector (by indeed copying from flatbuffer slice reference). We should be able use that. :) 

https://github.com/project-oak/oak/blob/67c0840be3223f07a54a881e2b23a6822e012ffc/oak_idl/src/utils.rs#L116

jul-sh: Although we actually still wouldn't be able to send it directly, since the message struct wraps flatbuffer struct type, which still expects a lifetime. 

I guess we may be able to update the IDL's Message struct to expose a method to get just the underlying Vec (without the flatbuffers data).

However, then we'd loose the ability to read underlying fields from the message, it would effectively become opaque, until parsed again by the IDL's Message struct with the correct flatbuffers type.

ipetr0v (approved):

experimental/uefi/loader/src/main.rs
conradgrobler: That is disappointing. Since a flatbuffer struct is basically just a wrapper around some byte buffer (IIUC) I would have expected it to be easy to consume the flatbuffer struct to get ownership of the underlying buffer, and then later recreate it by just wrapping it around the buffer.

jul-sh: Nope. Thankfully in this case our IDL's Message struct does provide that, which makes it a bit more ergonomic. But it's not possible for any nested flatbuffer structs contained within (without again copying). 

Anyways, I went ahead and implemented the approach of leveraging our Message struct by adding a method to get the underlying owned buffer. See the commit for how it turned out: https://github.com/project-oak/oak/pull/2944/commits/a724a42f245d7e2a66ab9725181e73267020c205

It's a bit nicer, but only works here because we use our own wrapper from the flatbuffer struct.

jul-sh: The reason flatbuffer doesn't let you take ownership of the slice btw is because in the underlying implementation it owns a large slice and a vtable that says where various structs are encoded on that slice. 

Returning an owned slice of just a struct would require splitting this larger into multiple owned slices. This isn't possible to do safely in rust without reallocating. For more context see the prev discussion when we ran into this: https://github.com/project-oak/oak/pull/2769#discussion_r858656495

jul-sh: Anyways, to get back to the previous discussion, even if we remove the bmrng channel, it would not solve the problem of being unable to invoke the sync oak IDL client directly in the gRPC logic of the loader. In fact atm the channel helps by providing an async surface to what would otherwise be a blocking call.

conradgrobler: > Anyways, to get back to the previous discussion, even if we remove the bmrng channel, it would not solve the problem of being unable to invoke the sync oak IDL client directly in the gRPC logic of the loader. In fact atm the channel helps by providing an async surface to what would otherwise be a blocking call.
> 
> So in any scenario we'd end up with this somewhat counter-intuitive proxy situation.

Yes, we would have to write an async wrapper around the channel, but that would give us more control over scheduling, ordering of messages etc.

jul-sh: > Yes, we would have to write an async wrapper around the channel, but that would give us more control over scheduling, ordering of messages etc.

🤔 I'm confused, the bmrng channel already is async? I suppose you mean async wrapper around the loader/runtime communication channel, which would indeed be the case. :) 

We could alternatively add extra control of scheduling and ordering by just adding extra channels too though (with eg lookup data updates using a seperate tokio channel). Might be attractive because it'd remove the need for a DIY async wrapper.

Either way, regardless of how this works out, it's somewhat orthogonal to my original point: The IDL client would still inside the logic handling the communication channel, and be exposed via proxy. It's not completely ideal, since these eg these proxy functions then continue to use IDL Error types. :)

experimental/uefi/runtime/src/framing.rs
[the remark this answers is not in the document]
jul-sh: Does so by copying also, but checks the length constraint via types is really nice.

experimental/uefi/loader/src/main.rs
conradgrobler: Yes, the communications channel, not the bmrng channel.

> We could alternatively add extra control of scheduling and ordering by just adding extra channels too though (with eg lookup data updates using a seperate tokio channel). Might be attractive because it'd remove the need for a DIY async wrapper.

I am not sure how that would work. What would synchronise the messages in the different channels so that we know which response should go to which requester?

experimental/uefi/channel/src/lib.rs
```diff
@@ -62,19 +102,66 @@ where
...
+                body: error.message.as_bytes().to_vec(),
+            },
+        }
+    }
+}
+
+impl From<Frame> for Result<Vec<u8>, oak_idl::Error> {
+    fn from(frame: Frame) -> Self {
+        match frame.method {
+            0 => Ok(frame.body),
```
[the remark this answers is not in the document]
jul-sh: Agreed this code reads a bit counterintuitive. Added a comment for extra context. 

What's going on here is that a response frame is being created from the response to an invocation in the oak_idl. On the framing layer the status code represents either succesful invocation `0`, or an error code of the oak_idl (any code > 0). See https://github.com/project-oak/oak/blob/67c0840be3223f07a54a881e2b23a6822e012ffc/oak_idl/src/lib.rs#L53

Ideally the oak_idl should probably handle detecting whether a status is an error or not. I'll issue a follow up PR to improve that. :)

experimental/uefi/channel/src/lib.rs
jul-sh: `FRAME_HEADER_SIZE` contains `METHOD_SIZE `, see L45 :)

experimental/uefi/loader/src/main.rs
```diff
@@ -155,19 +186,35 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Use a bmrng channel for serial communication. The good thing about spawning a separate
     // task for the serial communication is that it serializes (no pun intended) the communication,
     // as right now we don't have any mechanisms to track multiple requests in flight.
-    let (tx, mut rx) = bmrng::unbounded_channel::<Vec<u8>, anyhow::Result<Vec<u8>>>();
+    let (tx, mut rx) = bmrng::unbounded_channel::<Vec<u8>, Result<Vec<u8>, oak_idl::Error>>();
```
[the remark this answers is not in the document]
conradgrobler: Is there not a way to tell the wrapped struct that its lifetime should be the same as the owning message? Having to carry explicit lifetime annotations around everywhere is going to make using this quite annoying, I think.

experimental/uefi/loader/src/main.rs
jul-sh: For user requests we would still use the brmng channel. It's wrapper around a tokio channel, but includes a handle to a dedicated response channel with every message that is sent. 

For things like lookup data, it could be a simple [tokio](https://docs.rs/tokio/0.1.22/tokio/index.html)::[sync](https://docs.rs/tokio/0.1.22/tokio/sync/index.html)::[mpsc](https://docs.rs/tokio/0.1.22/tokio/sync/mpsc/index.html)::[channel](https://docs.rs/tokio/0.1.22/tokio/sync/mpsc/fn.channel.html).

Instead of immediately forwarding received messages into the loader/runtime channel the thread may collect a number of messages in its own cache (together with the relevant response handle), and prioritize which ones to send to the runtime first. 

Conceptually it's not that different from the thread that handles loader/runtime channel being invoked using a mutex. One passes messages, the other requires locking.

conradgrobler: Actually, I missed the fact that bmrng is MPSC, so the request sender can be cloned. This means that it would be usable for a while still, until we start interleaving frames or add multiprocessor support.

I guess it is fine for now. I just don't like the fact that the correctness of the system relies on a behaviour (FIFO) that is an accidental side effect of the current state of the runtime rather than an explicit supported property of the runtime.

experimental/uefi/loader/src/main.rs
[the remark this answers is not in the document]
jul-sh: > Having to carry explicit lifetime annotations around everywhere is going to make using this quite annoying, I think.
yes, I have the same concern. 

> Is there not a way to tell the wrapped struct that its lifetime should be the same as the owning message?
Possibly, I'm not sure. Would require a bit of a deep dive into Rust's type/lifetime annotations. Maybe @tiziano88 knows more.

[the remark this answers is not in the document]
conradgrobler: Yes, I don't think it is needed to resolve as part of this PR, but something we should look into if we plan on using flatbuffers more widely.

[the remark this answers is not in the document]
jul-sh: https://crates.io/crates/rental looks like it offers that.

It seems like at that point we're basically working around the flatbuffer approach of exclusively using references which seems a little dubious

experimental/uefi/loader/src/main.rs
jul-sh: Yeah that makes sense, thanks for explaining.

PR #2946, opened 2022-06-09
Use gRPC status codes in the oak IDL

jul-sh:
Tiziano proposed this earlier, and I agree it's probably a good idea. Crucially gRPC Status codes include a definition of a status code. This is nice, as it means we can use status codes as is in the framing implementation, doing away with what's a somewhat dodgy implementation rn. 

The downside is that technically an invocation could error with an Ok status, which wouldn't make sense. That's the same type / error signature as used by tonic though, so it's probably fine

### DIFF
--- a/oak_idl/src/lib.rs
+++ b/oak_idl/src/lib.rs
@@ -24,25 +24,22 @@ use core::fmt::Debug;
 
 pub mod utils;
 
-// TODO(#2500): Align with gRPC Status.
-// Ref: https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto
-/// Logical error model.
 #[derive(Debug)]
-pub struct Error {
-    pub code: ErrorCode,
+pub struct Status {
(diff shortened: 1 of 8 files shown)


oak_idl/src/lib.rs
```diff
@@ -24,70 +24,117 @@ use core::fmt::Debug;
...
             code,
             message: message.into(),
         }
     }
 }
 
-// TODO(#2500): Align these with gRPC status codes.
-/// Basic error code to categorize the error. Error codes are represented as an
-/// u32 id. The id `0` is reserved, as it may be used to indicate an "Ok" status
-/// in transport implementations.
```
jul-sh, line 56: This led to a somewhat messy implementation where the transport layers handles some status codes (but not others), which @conradgrobler rightly pointed out wasn't great: https://github.com/project-oak/oak/pull/2944#discussion_r893102044

conradgrobler (approved):
Thanks.

PR #2947, opened 2022-06-10
Send wasm module from the loader

jul-sh:


experimental/uefi/runtime/src/framing.rs
```diff
@@ -25,69 +28,109 @@ use oak_remote_attestation::handshaker::{
...
+            )),
+            AttestationState::Uninitialized(attestation_behavior) => {
+                let wasm_module_bytes: &[u8] = initialization
+                    .wasm_module()
+                    .ok_or_else(|| oak_idl::Status::new(oak_idl::StatusCode::InvalidArgument))?;
+                let wasm_handler = wasm::new_wasm_handler(wasm_module_bytes)
+                    .map_err(|_| oak_idl::Status::new(oak_idl::StatusCode::Internal))?;
+                let attestation_handler = Box::new(AttestationHandler::create(
+                    move |v| wasm_handler.handle_raw_invoke(v),
+                    attestation_behavior.clone(),
```
jul-sh, line 69: Rust question!

I'm not a huge fan of this `.clone` operation, since the `attestation_behavior` is no longer needed after it has been used here.

However, rust doesn't let me take it, since it's behind a mutable reference. 

Is there a way I could avoid this operation? I looked at storing `self.attestation_state` in a `RefCell` and calling `replace_with`, but it also doesn't allow me to consume the previous value.

jul-sh:
The Kokoro build is timing out, since the serial port is too slow :P

experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -194,6 +194,27 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                &mut builder,
+                &schema::InitializationArgs {
+                    wasm_module: Some(wasm_module),
+                },
+            );
+
+            match builder.finish(message) {
+                Ok(initialization_message) => initialization_message,
+                Err(err) => panic!("errored when creating initialization message: {:?}", err),
+            }
```
tiziano88, line 212: expect(..)

experimental/oak_baremetal_runtime/src/framing.rs
```diff
@@ -25,69 +28,112 @@ use oak_remote_attestation::handshaker::{
...
+}
+
+struct InvocationHandler<G, V>
 where
-    F: Send + Sync + Clone + FnOnce(Vec<u8>) -> anyhow::Result<Vec<u8>>,
     G: AttestationGenerator,
     V: AttestationVerifier,
 {
-    attestation_handler: AttestationHandler<F, G, V>,
+    attestation_state: AttestationState<G, V>,
```
tiziano88, line 45: I'm not sure this is the attestation state, shouldn't it be just initializationstate or just state? AFAICT the Wasm module is being initialized, not the attestation.

experimental/oak_baremetal_runtime/src/framing.rs
```diff
@@ -25,69 +28,112 @@ use oak_remote_attestation::handshaker::{
...
+        match &mut self.attestation_state {
+            AttestationState::Initialized(_attestation_handler) => Err(oak_idl::Status::new(
+                oak_idl::StatusCode::FailedPrecondition,
+            )),
+            AttestationState::Uninitialized(attestation_behavior) => {
+                let wasm_module_bytes: &[u8] = initialization
+                    .wasm_module()
+                    .ok_or_else(|| oak_idl::Status::new(oak_idl::StatusCode::InvalidArgument))?;
+                let wasm_handler = wasm::new_wasm_handler(wasm_module_bytes)
+                    .map_err(|_| oak_idl::Status::new(oak_idl::StatusCode::Internal))?;
```
tiziano88, line 69:
```suggestion
                    .map_err(|_err| oak_idl::Status::new(oak_idl::StatusCode::Internal))?;
```

(and below)

experimental/uefi/runtime/src/framing.rs
tiziano88: Since you already modified the bounds to include a 'static lifetime (which I think makes sense), then perhaps we can just a reference to the attestation behaviour from everywhere, so that it never needs to be cloned at all. Would that work?

experimental/uefi/runtime/src/remote_attestation.rs
```diff
@@ -31,6 +31,10 @@ use oak_remote_attestation_sessions::{SessionId, SessionState, SessionTracker};
 /// Number of sessions that will be kept in memory.
 const SESSIONS_CACHE_SIZE: usize = 10000;
 
+pub trait AttestationTrait {
```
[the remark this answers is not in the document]
jul-sh: Fair, renamed the trait to `AttestationHandler` and the struct to `AttestationSessionHandler`. The trait is actually technically needed here, since it allows us to use a dynamic implementation of the trait at runtime. L37 main.rs.

Given that the WASM module is passed at runtime, we cannot know the precise signature of the struct inn advance. :)

experimental/oak_baremetal_runtime/src/framing.rs
jul-sh: Renamed to InitializationState, probably a better description. It's only `AttestationState` in so far as that it governs whether the attestation logic can call into the Wasm module

experimental/oak_baremetal_runtime/src/framing.rs
```diff
@@ -25,69 +28,115 @@ use oak_remote_attestation::handshaker::{
...
 use oak_remote_attestation_sessions::SessionId;
 
-struct InvocationHandler<F, G, V>
+enum InitializationState<G, V>
+where
+    G: AttestationGenerator,
+    V: AttestationVerifier,
+{
+    Uninitialized(Option<AttestationBehavior<G, V>>),
+    Initialized(Box<dyn AttestationHandler>),
```
conradgrobler, line 37: Drive-by comment: can this not be replaced by a generic paramater? (e.g. type parameter `H` where `H: AttestationHandler` and then either `Initialized(Box<H>)` or even just `Initialized<H>`)

experimental/uefi/runtime/src/remote_attestation.rs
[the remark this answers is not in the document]
conradgrobler: I only noticed this after my other comment. I don't think I understand, as it does not seem that the type of the struct that will be used depends on the Wasm module bytes, only on the closure that is used to handle requests (which seems to be fixed). I would have assumed adding an additional trait bound to account for the closure should be sufficient.

experimental/oak_baremetal_runtime/src/framing.rs
jul-sh: It could be yes `Initialized(Box<H>)`, or `Initialized<H>`, if we require `H` to be Sized. :) We'd still have to implement scope the type down in the implementation of the `InvocationHandler`, since it invokes methods of the `AttestationHandler` trait. :)

jul-sh: Added a comment to explain why dyn is used. :) It might be possible to remove the need for it down the road, but likely requires refactoring the attestation logic.

conradgrobler (approved):

experimental/uefi/runtime/src/remote_attestation.rs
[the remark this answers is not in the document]
jul-sh: I'm not sure trait bounds for closures are possible, afaik rust closures have unique, anonymous types. But it might be possible to achieve this by refactoring the attestation logic to instead of a closure, accept the implementation of a trait.

experimental/oak_baremetal_app_crosvm/layout.ld
```diff
@@ -63,7 +63,7 @@ SECTIONS {
 
     /* Stack grows down, so stack_start is the upper address in memory. */
     .stack (NOLOAD) : ALIGN(4K) {
-        . += 128K;
+        . += 512K;
```
conradgrobler, line 66: You probably need to do the same in https://github.com/project-oak/oak/blob/main/third_party/rust-hypervisor-firmware-boot/layout.ld#L43 for the QEMU version.

jul-sh: yup, thanks!

tiziano88:
FYI there is some overlap with #2979 , should we still review them separately?

jul-sh:
@tiziano88 I think it's fine to review them separately. This PR basically just reduces the maximum frame size to work around the current vsock implementation issue. :) Will rebase the PRs as needed depending on which one is merged first.

PR #2951, opened 2022-06-10
Address review comments from #2948

conradgrobler:
- Updates doc comments for UEFI/baremetal client
- Uses random values for echo test


rbehjati (approved):
Thanks.

PR #2952, opened 2022-06-10
Add simple VSock echo runtime

ipetr0v:
This PR adds a simple runtime for running inside a sandbox which uses VSock for communications.

### DIFF
--- a/experimental/vsock/echo/src/main.rs
+++ b/experimental/vsock/echo/src/main.rs
@@ -1,0 +1,57 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use libc::c_int;
+use std::{
+    io::{Read, Write},
+    net::Shutdown,
+    os::unix::io::FromRawFd,
+};
+use vsock::VsockStream;
+
+const FILE_DESCRIPTOR: c_int = 1023;
+const BUFFER_SIZE: usize = 1024;
+
+fn main() {
+    let mut stream;
+    unsafe {
+        stream = VsockStream::from_raw_fd(FILE_DESCRIPTOR);
+    }
+    println!("Connected to the {}", stream.peer_addr().expect("Couldn't get peer address"));
+
+    let mut buffer = vec![];
+    buffer.resize(BUFFER_SIZE, 0);
+    loop {
+        let read_bytes = match stream.read(&mut buffer) {
+            Ok(0) => break,
+            Ok(read_bytes) => read_bytes,
+            Err(error) => panic!("Couldn't read bytes: {}", error),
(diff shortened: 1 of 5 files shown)


experimental/vsock/echo/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    unsafe {
+        stream = VsockStream::from_raw_fd(FILE_DESCRIPTOR);
+    }
+    println!(
+        "Connected to the {}",
+        stream.peer_addr().expect("Couldn't get peer address")
+    );
+
+    let mut buffer = vec![];
+    buffer.resize(BUFFER_SIZE, 0);
```
tiziano88, line 39:
```suggestion
    let mut buffer = vec![0; BUFFER_SIZE];
```

I think

experimental/vsock/echo/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    loop {
+        let read_bytes = match stream.read(&mut buffer) {
+            Ok(0) => break,
+            Ok(read_bytes) => read_bytes,
+            Err(error) => panic!("Couldn't read bytes: {}", error),
+        };
```
tiziano88, line 45: I would separate the error handling from the handling of the zero case, e.g.

```suggestion
        let read_bytes = stream.read(&mut buffer).expect("Couldn't read bytes");
        if read_bytes == 0 {
          break // add comment why
        }
```

experimental/vsock/echo/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+        println!("Received {:?} bytes", read_bytes);
+
+        let mut write_bytes_sum = 0;
+        while write_bytes_sum < read_bytes {
+            let write_bytes = match stream.write(&buffer[write_bytes_sum..read_bytes]) {
```
tiziano88, line 50: Not sure what this logic is, is it trying to emulate a ring buffer? If so, could you add a comment? But also, what's the point of a ring buffer, if reads and writes are always interleaved? It seems fine to always start from the start of the buffer. Or maybe I'm misunderstanding how it works.

conradgrobler: I think this can be simplified by using `write_all`. It will either write the entire buffer (which should always succeed for a buffer of 1024 bytes unless the stream is disconnected) or give an error.

conradgrobler (approved):

experimental/vsock/echo/src/main.rs
ipetr0v: Done

experimental/vsock/echo/src/main.rs
ipetr0v: Rewritten

experimental/vsock/echo/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    net::Shutdown,
+    os::unix::io::FromRawFd,
+};
+use vsock::VsockStream;
+
+const FILE_DESCRIPTOR: c_int = 1023;
+const BUFFER_SIZE: usize = 1024;
+
+fn main() {
+    let mut stream;
```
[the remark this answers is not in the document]
ipetr0v: Done

tiziano88 (approved):

tiziano88:
I think this should become a proper trusted runtime shortly, at the same level of https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_base and https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_unsafe, perhaps called `oak_functions_linux_vsock` or similar. Personally I would have started it there directly, but I think it's fine to move it there soon, I don't think it should keep existing under experimental for long though.

PR #2954, opened 2022-06-13
Implement the ability to send requests/responses over several frames

jul-sh:


experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
-static_assertions::assert_eq_size!([u8; METHOD_SIZE], Method);
+static_assertions::assert_eq_size!([u8; METHOD_SIZE], MethodOrStatus);
 
-pub const FRAME_HEADER_SIZE: usize = 8;
+// TODO(#2848): Use this space to send a stream identifier that identifies
+// frames as belonging to the same message. Currently the frames of every message
+// must be sent as an uninterrupted seqeuence. A stream identifier would allow us
+// to interleave messages.
+/// Reserved space to maintain 8 byte alignement.
+const SPACER_SIZE: usize = 4;
```
tiziano88, line 51: Perhaps padding is a more common term than spacer for this sort of thing, but up to you.

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+
+#[repr(u32)]
+#[derive(Copy, Clone)]
+enum Flag {
+    /// Atomic frame, a message sent as a single frame
+    AtomicFrame = 0,
+    /// Starts a stream of frames that constitute a message
+    StreamStart = 1,
+    /// Continues a stream of frames that constitute a message
+    StreamContinuance = 2,
```
tiziano88, line 61:
```suggestion
    StreamContinuation = 2,
```

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+impl TryFrom<[u8; FLAG_SIZE]> for Flag {
+    type Error = anyhow::Error;
+
+    fn try_from(value: [u8; FLAG_SIZE]) -> Result<Self, Self::Error> {
+        let flag = match u32::from_le_bytes(value) {
+            0 => Flag::AtomicFrame,
+            1 => Flag::StreamStart,
+            2 => Flag::StreamContinuance,
+            3 => Flag::StreamEnd,
+            _ => anyhow::bail!("invalid flag"),
```
tiziano88, line 77: There must be a way of performing this automatically? Otherwise what's the point of `repr(u32)`? I think a [numeric cast](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics) should work, or if not, perhaps see https://enodev.fr/posts/rusticity-convert-an-integer-to-an-enum.html .

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+enum MessageReconstructionErrors {
+    ExpectedStartFrame,
+    DoubleStartFrame,
+    MismatchingFrameHeader,
+}
+impl MessageReconstructionErrors {
+    fn as_str(&self) -> &'static str {
+        match self {
+            MessageReconstructionErrors::ExpectedStartFrame => "Invalid frame order. The first frame must be a start frame.",
+            MessageReconstructionErrors::DoubleStartFrame => "Invalid frame order. The frame method/status header does must be the same for all frames belonging to the same message.",
```
conradgrobler, line 191: nit: "... header must be ..."

Also, the description for this seems to be swapped with the next one.

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+            0 => Flag::AtomicFrame,
+            1 => Flag::StreamStart,
+            2 => Flag::StreamContinuance,
+            3 => Flag::StreamEnd,
+            _ => anyhow::bail!("invalid flag"),
+        };
+        Ok(flag)
+    }
+}
+impl From<Flag> for [u8; 4] {
```
tiziano88, line 82:
```suggestion
impl From<Flag> for [u8; FLAG_SIZE] {
```

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+    AtomicFrame = 0,
+    /// Starts a stream of frames that constitute a message
+    StreamStart = 1,
+    /// Continues a stream of frames that constitute a message
+    StreamContinuance = 2,
+    /// Ends a stream of frames that constitute a message, indicating that the recipient may
+    /// process the message.
+    StreamEnd = 3,
+}
+const FLAG_SIZE: usize = 4;
```
tiziano88, line 66: 4 bytes for a flag seems excessive? I guess this is mostly for alignment?

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+}
+
+const FRAME_HEADER_SIZE: usize = 16;
 static_assertions::assert_eq_size!(
     [u8; FRAME_HEADER_SIZE],
-    [u8; FRAME_LENGTH_SIZE + METHOD_SIZE]
+    [u8; FRAME_LENGTH_SIZE + METHOD_SIZE + SPACER_SIZE + FLAG_SIZE]
 );
+// Ensure that the frame header is 8 Byte aligned.
+static_assertions::const_assert!(FRAME_HEADER_SIZE % 8 == 0);
```
tiziano88, line 94: Nice!

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
+const MAX_FRAME_SIZE: usize = 64000;
+const MAX_FRAME_BODY_SIZE: usize = MAX_FRAME_SIZE - FRAME_HEADER_SIZE;
+
+/// A [`Frame`] is a small unit data that is sent over the communication
+/// channel. Usually several [`Frame`]'s are used to send a [`Message`].
+#[derive(Clone)]
+struct Frame {
+    method_or_status: MethodOrStatus,
+    flag: Flag,
+    body: Vec<u8>,
```
tiziano88, line 105: Perhaps we should distinguish Request and Response frames at some point, instead of overloading the semantics of method_or_status and flag fields?

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
     }
 }
 
-pub struct Framed<T: Read + Write> {
+/// A [`Message`] is conceptual unit of data that is sent over the communication
+/// channel. It can represent either a full request, or full response of an oak_idl
+/// invocation. Requests and responses can be converted to a [`Message`], and
+/// vice versa. Usually a single [`Message`] is sent over the communication
+/// channel over several [`Frame`]'s.
```
tiziano88, line 125:
```suggestion
/// channel over several [`Frame`]s.
```

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+                Flag::StreamEnd => {
+                    if frame.method_or_status != partial_message.method_or_status {
+                        Err(MessageReconstructionErrors::MismatchingFrameHeader)
+                    } else {
+                        partial_message.body.append(&mut frame.body);
+                        Ok(Ok(partial_message))
+                    }
+                }
+                Flag::StreamContinuance => {
+                    if frame.method_or_status != partial_message.method_or_status {
```
conradgrobler, line 235: nit: there is a lot of duplication here that could be lifted out of the match.

experimental/uefi/channel/src/lib.rs
```diff
@@ -113,59 +367,51 @@ where
         self.inner.write_all(&length_bytes)?;
         let method_or_status_bytes = frame.method_or_status.to_le_bytes();
         self.inner.write_all(&method_or_status_bytes)?;
+        let spacer_bytes = [0; SPACER_SIZE];
+        self.inner.write_all(&spacer_bytes)?;
+        let flag_bytes: [u8; FLAG_SIZE] = frame.flag.into();
+        self.inner.write_all(&flag_bytes)?;
```
conradgrobler, line 373: Can be done later, but we should consolidate all of this writing into a single write onto the inner channel relatively soon. At the moment this code will send 5 packets for a single frame.

conradgrobler (approved):
Thanks

tiziano88 (commented):
Thanks, I haven't finished reviewing yet but sending some comments in the meanwhile.

experimental/uefi/channel/src/lib.rs
conradgrobler: We already have `strum` as a dependency and use it elsewhere in the codebase for converting between enums and its numeric representations. https://github.com/project-oak/oak/blob/7596b65f92f0bd5f86b9e9d2a1804c41ef58129e/experimental/uefi/kernel/src/boot.rs#L17-L19

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+            })
+            .collect()
+    }
+}
+
+/// Struct to incrementally build a [`Message`] from a set of [`Frame`]s.
+#[derive(Default, Debug, PartialEq, Eq)]
+struct PartialMessage {
+    inner: Option<Message>,
+}
```
tiziano88, line 180: (nit) consider adding blank lines between items.

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+#[derive(Debug, PartialEq, Eq)]
+enum MessageReconstructionErrors {
+    ExpectedStartFrame,
+    DoubleStartFrame,
+    MismatchingFrameHeader,
+}
+impl MessageReconstructionErrors {
+    fn as_str(&self) -> &'static str {
```
tiziano88, line 188: Maybe call this `description`?

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+impl PartialMessage {
+    /// Attempt to complete the [`PartialMessage`] with a frame. Returns
+    /// an outer [`Result`] to handle errors resulting from invalid arguments.
+    /// Returns an inner [`Result`] that contains either the succesfully
+    /// completed message, or if more frames are needed to complete the message,
+    /// a new [`PartialMessage`] that includes the previously passed frames.
+    pub fn try_complete(
+        self,
+        mut frame: Frame,
+    ) -> Result<Result<Message, Self>, MessageReconstructionErrors> {
```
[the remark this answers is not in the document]
tiziano88: +1, I think the current return type is a bit unusual and not super ergonomic.

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+impl<'a> From<&'a Message> for oak_idl::Request<'a> {
+    fn from(frame: &'a Message) -> Self {
+        oak_idl::Request {
+            method_id: frame.method_or_status,
+            body: &frame.body,
+        }
+    }
+}
+
+/// Construct a [`Message`] from an the response to an [`oak_idl::Request`].
```
tiziano88, line 274: (nit) either `a` or `the`.

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+impl<'a> From<Result<Vec<u8>, oak_idl::Status>> for Message {
```
tiziano88, line 275: Unnecessary lifetime?

experimental/uefi/channel/src/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+
+const BODY_LEN_MULTIPLICATOR: usize = 5;
+const MOCK_LARGE_MESSAGE_LEN: usize = MAX_FRAME_BODY_SIZE * BODY_LEN_MULTIPLICATOR;
+
+#[test]
+fn test_message_fragmentation() {
+    let mock_body = {
+        let mut x: u8 = 0;
+        let filler = || {
+            x = x.overflowing_add(1).0;
```
tiziano88, line 27: Why overflowing? Shouldn't it be a checked add?

edit: I understand now that it is meant to just create some variation in the message body, but IMO it distracts from the point of this test. Perhaps consider filling it with some constant instead, unless you think the variation actually adds value to this test.

jul-sh: I'll add a comment for extra context. The TL;DR is that I think it makes sense to check that the message is reconstructed in the correct order. I actually used the variant of just filling it with a constant first, but then it means we basically just check that the length of the message and reconstructed message matches.

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +33,76 @@ pub mod schema {
...
-    pub method_or_status: Method,
-    pub body: Vec<u8>,
+const MAX_FRAME_SIZE: usize = 64000;
+const MAX_FRAME_BODY_SIZE: usize = MAX_FRAME_SIZE - FRAME_HEADER_SIZE;
+
+/// A [`Frame`] is a small unit data that is sent over the communication
+/// channel. Usually several [`Frame`]'s are used to send a [`Message`].
+#[derive(Clone)]
+struct Frame {
+    method_or_status: MethodOrStatus,
```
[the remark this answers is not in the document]
tiziano88: In fact I think in principle the framing layer should not even know what the message represents (and in particular whether it has a method or status or anything else). It should just get some bytes, and split them into fragments, and then reconstruct them on the other side.

experimental/uefi/channel/src/lib.rs
```diff
@@ -66,16 +118,205 @@ impl Frame {
...
+/// vice versa. Usually a single [`Message`] is sent over the communication
+/// channel over several [`Frame`]'s.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Message {
+    pub method_or_status: MethodOrStatus,
+    pub body: Vec<u8>,
+}
+
+/// Deconstruct a [`Message`] into a set of [`Frame`]s.
+impl<'a> From<Message> for Vec<Frame> {
```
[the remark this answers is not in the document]
jul-sh: not needed :)

experimental/uefi/channel/src/lib.rs
jul-sh: p much, yes. Originally we planned a smaller value for this. But with the oak_idl using u32 for the method_id I had to increase the mehood_id header value, and that point the easiest way to maintain alignment was to make header key 4 bytes long.

experimental/uefi/channel/src/lib.rs
jul-sh: thx, updated

experimental/uefi/channel/src/lib.rs
jul-sh: thanks for catching :)

experimental/uefi/channel/src/lib.rs
tiziano88: I still think we should document the format somewhere else, i.e. the reference for the framing format should not be this Rust code, but some markdown file somewhere, which defines the actual offsets and padding etc. within the header, in which case I think flags could be 1 byte, and we could have 3 bytes of padding / alignment (or reserved for future use). Or would we want the spec to say that this is really a 4 bytes flag field?

experimental/uefi/channel/src/lib.rs
jul-sh: thanks, updated to use strum

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: I don't disagree, logical layers seem nice. But IRRC we had this discussion at once point earlier, and decided that for now it would be nice to have the method recorded in each frame, in case it helps with routing / prioritization later on. 

Happy to revisit this, but for now it might be good to keep it around. :)

experimental/uefi/channel/src/lib.rs
jul-sh: Makes sense. Does that apply to both virtio console and vsock though? It seems it'd only be the case for vsock. The initial reason for incremental writes was to avoid reallocating to merge all the frame pieces, but agreed that the cost of that is out-weight by sending multiple packages.

experimental/uefi/channel/src/lib.rs
jul-sh: Maybe, imo it should be fine for now. Having a single struct makes it a lot more straightforward to do the logic for message encoding and message fragmentation, both of which is shared. Flags should also be the same across requests and responses

experimental/uefi/channel/src/lib.rs
```diff
@@ -32,25 +34,62 @@ pub mod schema {
...
+/// Reserved space to maintain 8 byte alignement.
+const PADDING_SIZE: usize = 4;
+
+#[repr(u32)]
+#[derive(Copy, Clone, Display, FromRepr)]
+enum Flag {
+    /// Atomic frame, a message sent as a single frame
+    AtomicFrame = 0,
+    /// Starts a stream of frames that constitute a message
+    StreamStart = 1,
```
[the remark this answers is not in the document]
jul-sh: Probably, yeah

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: RE the result return type: that's fair, I was going back and forth between whether to use an Enum (ala `Enum { Complete(Message), Incomplete(PartialMessage) }` or a result. :) 

RE recursion though, that actually comes from the fact that this approach is more functional, in so far as that it does not use mutation. 

Expressed as a trait, the current version looks something like:
```
trait FunctionalAssembler: Sized {
    fn try_complete(
        self,
        frame: Frame,
    ) -> Result<Result<Message, Self>, MessageReconstructionErrors>;
}
```

A mutable (and by extension iteration-friendly) struct would look something like this:

```
trait MessageBuilder {
    fn requires_more_frames(&self) -> bool;
    fn get_completed_message(self) -> Message;
    fn add_frame(&mut self, frame: Frame) -> Result<(), MessageReconstructionErrors>;
}
```

We could then iterate over it via:

```
while message_builder.requires_more_frames() {
    let frame = framed.read_frame()?;
    message_builder.add_frame(frame)?;
}
let reconstructed_message = message_builder.get_completed_message();
```

While friendly on the surface, I think it actually a bit of complexity, since it methods to align in behavior. (Eg if requires_more_frames returns true, add_frame should fail. If requires_more_frames is false, get_completed_message is still available but should throw). It feels more ergonomic, but in each state there's methods that while available will need to error if called.

We used the iterative approach before, and arguably it led to some unwelcome coupling: https://github.com/project-oak/oak/pull/2906/files/e613d85fbb2be9963a71caa075962e55609fd5e0#r884977491

Hence I'm personally partial to the function approach. But happy to go with iterative if people prefer. :)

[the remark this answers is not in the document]
jul-sh: Definitely happy to change the inner Result to an Enum though either way :)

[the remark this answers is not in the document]
jul-sh: Also should note that the need for recursion should go away once we implement message multiplexing :)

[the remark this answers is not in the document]
conradgrobler: I disagree that the recursive approach is stateless. It hides the state in the execution stack and does require mutation (`append_partial_message`).

I agree that the recursion will go away once interleaving of messages is implemented, which is part of the reason I preferred an approach that would be closer to the future state. Having said that, I don't have a strong objection to temporarily keeping this approach if it unblocks other work.

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
conradgrobler: Another reason to favour removing the method ID from the header is that (as far as I can tell) is the only reason for having the `PartialMessage` construct. But I agree that it is fine to keep as is for now and refactor as needed as part of future code changes (will probably happen naturally as part of supportring multiplexed/interleaves message).

conradgrobler (approved):

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: Agreed that it's about hiding inner mutability & state from the outside, it doesn't go away entirely. Not sure if mutability necessarily makes the result closer to what we'll have with message interleaving, but ok to go with it.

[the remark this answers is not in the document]
conradgrobler: I assume interleaving will be implmented using some cache (e.g. LRU map). When a frame is received the message ID will be the key for finding the partial buffer in the cache. If the in-progress buffer is found (and the flags are OK0 it will be mutated by appending the bytes from the latest frame. This seems much closer to the iterative approach than the recursive approach.

Perhaps there is a different implementation for interleaving that will look closer to the recursive approach, but I could not think of an abvious candidate.

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: Yes, since we're dealing with one message at a time we could move all logic into a single function that uses local state only. 

It seems that's not necessarily tied to the `method_id`, said function with local state could likewise check that this value is consistent.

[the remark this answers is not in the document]
jul-sh: I should note that if we do go with buffer approach we will require two IDs, one to link request messages with their response messages, and another one to link frames belonging to the same message. 

Imo fine and nice for having logical layers, but I recall that we earlier felt this was undesireable.

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: In the recursive approach, instead of appending the partial buffer in the LRU cache in place, it would be replaced with an appended version of the same buffer.

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
conradgrobler: To be fair, that comment was under the assumption that the initial implementation would not use the IDL, and that having a single header would simplify the code and the initial implementation. At this point it seems that having a single header is complicating the code rather than simplifying it.

For this PR, I am fine with either approach. We will probably go through at least one more design cycle around the IDL and framing to implement interleaving, so any refactoring can happen as part of that.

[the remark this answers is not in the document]
jul-sh: Makes sense, thanks for explaining :)

jul-sh:
Merging this as frame fragmentation unblocks https://github.com/project-oak/oak/pull/2947, will iterate on the logical layers more later. 

Thanks for the reviews @conradgrobler & @tiziano88

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -66,16 +105,233 @@ impl Frame {
...
+            })
+            .collect()
+    }
+}
+
+/// Struct to incrementally build a [`Message`] from a set of [`Frame`]s.
+#[derive(Default, Debug, PartialEq, Eq)]
+struct PartialMessage {
+    inner: Option<Message>,
+}
```
tiziano88, line 193: I'm not sure I see the value of building a partial message and having an inner message which is semantically not a message. How about just keeping track of the Vec<Vec<u8>> parts here (i.e. just append the next frame each time), and once all of them are collected, then just do a single new allocation of a new Vec to concat all of them into a single final body (which is the message body). This way there is never a malformed intermediate Message (which IMO currently makes things semantically unclear), and also there is a single allocation instead of potentially many intermediate reallocations.
Also the add_frame method may be simplified (though I'm not sure to what extent).

experimental/uefi/channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: Updated the design doc with the new approach of using clearly delineated layers (and by extension seperate IDs for frames and messages). https://docs.google.com/document/d/1Q4u5hYzOFF74_RfwZzpRWgSSwJwfQ-GXRvNllgCXOO8/edit?usp=sharing

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -66,16 +105,233 @@ impl Frame {
...
+            Err(error) => Message {
+                method_or_status: error.code.into(),
+                body: error.message.as_bytes().to_vec(),
+            },
+        }
+    }
+}
+
+/// Construct a the response to a [`oak_idl::Request`] from a [`Message`].
+impl From<Message> for Result<Vec<u8>, oak_idl::Status> {
```
tiziano88, line 306: Shouldn't this be TryFrom<Message> with an error of oak_idl::Status instead?

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -104,68 +371,56 @@ where
...
+        let frames: Vec<Frame> = message.into();
+        for frame in frames.into_iter() {
+            self.inner.write_frame(frame)?
         }
+        Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests;
```
tiziano88, line 426: This is usually at the top.

experimental/oak_baremetal_channel/src/tests.rs
```diff
@@ -0,0 +1,182 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use super::*;
+
+const BODY_LEN_MULTIPLICATOR: usize = 5;
```
tiziano88, line 19: multiplier?

experimental/uefi/channel/src/lib.rs
tiziano88: I think we should have the source of truth of the format in this repo though (as a markdown file), since it needs to be kept in sync with the actual implementation, and we need to be able to refer to it when implementing it in other languages (e.g. @ipetr0v will probably have to do that quite soon in the untrusted loader in C++).

PR #2960, opened 2022-06-14
Add Oak Baremetal Linux Wrapper

ipetr0v:
This PR adds a simple wrapper binary to run Oak Baremetal in a Linux VM.

This binary connects to the file descriptor created by Bedebox using vsock.


tiziano88:
This looks very promising so far, thanks Ivan!

I'm not sure about the terminology though -- baremetal and linux seems to contradict each other.

AFAICT this is a version of Oak Functions as a linux binary that expects to receive messages over a vsock connection, right? If so, perhaps the name should be something like `oak_functions_loader_linux_vsock` (parallel to https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_base)?

cc @conradgrobler @andrisaar

andrisaar:
> 
> I'm not sure about the terminology though -- baremetal and linux seems to contradict each other.
>

I agree. I think the "baremetal" term has tainted more code than it should. For example, I think the whole `oak_baremetal_runtime` should lose the baremetal moniker; as we see here, there's nothing really bare-metal about that crate.

experimental/oak_functions_loader_linux_vsock/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+name = "oak_functions_loader_linux_vsock"
+version = "0.1.0"
+authors = ["Ivan Petrov <ivanpetrov@google.com>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+anyhow = { version = "*", default-features = false }
+byteorder = "*"
+ciborium-io = { version = "*", default-features = false }
```
andrisaar, line 11: We're not using ciborium any longer.

experimental/oak_functions_loader_linux_vsock/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::channel::Channel;
+use libc::c_int;
+use log::info;
+use oak_remote_attestation::handshaker::{
+    AttestationBehavior, EmptyAttestationGenerator, EmptyAttestationVerifier,
+};
+use std::os::unix::io::FromRawFd;
+use vsock::VsockStream;
+
+const FILE_DESCRIPTOR: c_int = 1023;
```
andrisaar, line 42: Where does this constant come from? Is it guaranteed to be 1023? (Maybe it should be a flag?)

experimental/oak_functions_loader_linux_vsock/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// Connect to the file descriptor created by Bedebox using vsock.
+pub fn create_vsock_stream(file_descriptor: c_int) -> anyhow::Result<VsockStream> {
```
andrisaar, line 45: I'd prefer `RawFd` instead of `c_int` here.

experimental/oak_functions_loader_linux_vsock/src/channel.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+impl<T> oak_baremetal_communication_channel::Read for Channel<T>
+where
+    T: std::io::Read + std::io::Write
+{
+    fn read(&mut self, data: &mut [u8]) -> anyhow::Result<()> {
+        let mut buffer = vec![0; BUFFER_SIZE];
```
conradgrobler, line 61: I don't think this logic works. It tries to read 64KiB from the stream. If it has been a long time since the last read, it will fill the entire 64KiB buffer, but that might not represent a correct cut-off point for the proto serialisation.

experimental/oak_functions_loader_linux_vsock/src/channel.rs
```diff
@@ -0,0 +1,141 @@
...
+        let read_bytes = self
+            .stream
+            .read(&mut buffer)
+            .map_err(|error| anyhow!("Couldn't read from stream: {:?}", error))?;
+        if read_bytes > 0 {
+            let mut current_buffer = &buffer[..read_bytes];
+            let mut cursor = Cursor::new(&mut current_buffer);
+
+            // Read multiple size:value pairs from the buffer.
+            while !cursor.is_empty() {
```
conradgrobler, line 71: Why is the buffer/cursor needed? This copies data from the stream to a buffer and then reads from the buffer using a cursor so that is looks like a stream again. Why not just read direclty from the stream in the first place?

experimental/oak_functions_loader_linux_vsock/src/channel.rs
```diff
@@ -0,0 +1,141 @@
...
+                // Read message size.
+                let mut size_buffer: Vec<u8> = vec![0; size_of::<u64>()];
+                cursor.read_exact(&mut size_buffer).map_err(|error| {
+                    anyhow!(
```
conradgrobler, line 75: All of these remapping of errors should probably be replaced with anyhow's context (e.g. `cursor.read_exact(&mut size_buffer).context("Couldn't read ....")?` for consistecy with the rest of the code base.

experimental/oak_functions_loader_linux_vsock/src/channel.rs
```diff
@@ -0,0 +1,141 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::proto::oak::sandbox::runtime::{Request, Response};
+use anyhow::anyhow;
+use byteorder::{BigEndian, ByteOrder};
```
conradgrobler, line 19: Is this dependency really needed? Is `u64::from_be_bytes` not sufficient?

experimental/oak_functions_loader_linux_vsock/src/tests.rs
```diff
@@ -0,0 +1,45 @@
...
+    let mut write_buffer = vec![0; TEST_BUFFER_SIZE];
+
+    let write_stream = Cursor::new(&mut write_buffer);
+    let mut write_channel = Channel::new(write_stream);
+
+    let write_result = write_channel.write(&TEST_MESSAGE);
+    assert_matches!(write_result, Ok(_));
+
+    // Copy data to a new stream and remove trailing zeroes.
+    let mut read_buffer: Vec<u8> = write_buffer.into_iter().rev().skip_while(|&x| x == 0).collect();
```
conradgrobler, line 36: This seems potentially flaky. If you are using a Vec anyway, can you not just start with a mutable vec for the buffer? I would assume that it would then have the correct size once writing has finished and you can reuse it for the read cursor.

experimental/oak_functions_loader_linux_vsock/src/channel.rs
```diff
@@ -0,0 +1,141 @@
...
+use prost::Message;
+use ringbuf::{Consumer, Producer, RingBuffer};
+use std::{
+    io::{Cursor, Read, Write},
+    mem::size_of,
+};
+
+const BUFFER_SIZE: usize = 65536;
```
conradgrobler, line 27: This size is probably excessive. The current max size for a frame is 4000 bytes (not 64000 as it used to be).

conradgrobler:
I had a look through this again and there is an underlying concern about the implementation of `oak_baremetal_communication_channel::Read` that my previous review comments did not highlight clearly:

The runtime will call `read` twice for every frame that is processed. This implementation reads from the underlying stream every time that `read` is called. In case of heavy load (i.e. many user requests queued) this would mean that at least 32 frames are read from the stream and copied into the ringbuffer for every frame that is processed and removed from the ringbuffer. This would cause the ringbuffer to grow very quickly, and with no flow control or backpressure.

I think a safer implementation would look something like:

- Try to complete the read operation from the ringbuffer without reading from the underlying stream
- If there are not enough bytes to complete the read operation, read only one proto from the underlying stream and add the data to the ringbuffer
- Go back to the first step

This would automatically create backpressure on the underlying stream and avoid the unbounded growth of the ringbuffer.

For a version of read that is implemented using an `Option<Vecdeque>` to achieve a similar outcome see e.g. https://github.com/project-oak/oak/blob/main/experimental/oak_baremetal_kernel/src/simpleio.rs

experimental/oak_functions_loader_linux_vsock/src/tests.rs
ipetr0v: I don't think `Cursor` is going to increase the buffer size.

conradgrobler: Based on the comment in https://doc.rust-lang.org/std/io/struct.Cursor.html#method.new I think it will overwrite a mutable vector if it is not empty, but append if it is empty.

ipetr0v:
Updated the vsock stream reading logic. Now it only reads `while self.read_buffer_consumer.len() < data.len()`, i.e. only if there is not enough data in the ring buffer.

experimental/oak_functions_loader_linux_vsock/src/tests.rs
```diff
@@ -0,0 +1,43 @@
...
+fn test_channel() {
+    let mut write_buffer = vec![];
+
+    let write_stream = Cursor::new(&mut write_buffer);
+    let mut write_channel = Channel::new(write_stream);
+
+    let write_result = write_channel.write(&TEST_MESSAGE);
+    assert_matches!(write_result, Ok(_));
+
+    // Copy data to a new stream and remove trailing zeroes.
```
conradgrobler, line 34: I don't think this comment is accurate anymore.

conradgrobler (approved):
Thanks.

For consistency with the new crate name, perhaps also update the PR title.

experimental/oak_functions_loader_linux_vsock/proto/BUILD
```diff
@@ -0,0 +1,34 @@
...
+    licenses = ["notice"],
+)
+
+proto_library(
+    name = "runtime_proto",
+    srcs = ["runtime.proto"],
+    cc_api_version = 2,
+)
+
+cc_proto_library(
```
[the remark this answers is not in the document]
ipetr0v: Done

experimental/oak_functions_loader_linux_vsock/src/tests.rs
ipetr0v: Done

andrisaar (approved):

PR #2962, opened 2022-06-14
Baremetal: Add a test for sending a large message

jul-sh:
There seems to be a problem with sending and receiving large amounts of data over virtio vsock. Let's add a test for that. (Currently not passing).


conradgrobler (approved):
Thanks. This will only pass once #2968 is implemented, so can only be merged afterwards.

Issue #2961, opened 2022-06-14
Document frame / message wire format


tiziano88:
I think we should have the source of truth of the format in this repo though (as a markdown file), since it needs to be kept in sync with the actual implementation, and we need to be able to refer to it when implementing it in other languages (e.g. @ipetr0v will probably have to do that quite soon in the untrusted loader in C++).

_Originally posted by @tiziano88 in https://github.com/project-oak/oak/pull/2954#discussion_r896922244_

PR #2963, opened 2022-06-14
Implement a MessageId in preparation for multiplexing

jul-sh:
This PR adds a `MessageId` in preparation for implementing message multiplexing in the future. For now messages are still sent in sequence.

### DIFF
--- a/experimental/oak_baremetal_channel/src/client.rs
+++ b/experimental/oak_baremetal_channel/src/client.rs
@@ -1,0 +1,73 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::{InvocationChannel, Message, String, Vec};
+use ciborium_io::{Read, Write};
+
+pub struct ClientChannelHandle<T: Read + Write> {
+    inner: InvocationChannel<T>,
+}
+
+impl<T> ClientChannelHandle<T>
+where
+    T: Read<Error = anyhow::Error> + Write<Error = anyhow::Error>,
+{
+    pub fn new(socket: T) -> Self {
+        Self {
+            inner: InvocationChannel::new(socket),
+        }
+    }
+    pub fn write_request(&mut self, request: Message) -> anyhow::Result<()> {
+        self.inner.write_message(request.into())
+    }
+    pub fn read_response(&mut self) -> anyhow::Result<Message> {
+        let message = self.inner.read_message()?;
+        Ok(message.into())
+    }
+}
+
+/// Construct a [`Message`] from an [`oak_idl::Request`].
+impl<'a> From<oak_idl::Request<'a>> for Message {
+    fn from(request: oak_idl::Request) -> Message {
+        Message {
+            method_or_status: request.method_id,
+            // TODO(#2848): It'd be nice if we didn't have to reallocate here.
+            // We may be able to avoid doing so by making [`Message`] use
+            // slices and lifetimes. Or alternatively use the Bytes crate for
+            // reference counting.
+            body: request.body.to_vec(),
+        }
+    }
+}
+
+/// Construct the response to a [`oak_idl::Request`] from a [`Message`].
+impl From<Message> for Result<Vec<u8>, oak_idl::Status> {
+    fn from(frame: Message) -> Self {
+        if frame.method_or_status == oak_idl::StatusCode::Ok.into() {
+            Ok(frame.body)
+        } else {
+            Err(oak_idl::Status {
+                code: frame.method_or_status.into(),
+                message: String::from_utf8(frame.body.to_vec()).unwrap_or_else(|err| {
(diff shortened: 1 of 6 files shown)


experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -88,6 +96,7 @@ const MAX_FRAME_BODY_SIZE: usize = MAX_FRAME_SIZE - FRAME_HEADER_SIZE;
 #[derive(Clone)]
 struct Frame {
     method_or_status: MethodOrStatus,
+    message_id: MessageId,
     flag: Flag,
```
tiziano88, line 100: incidentally, I feel more and more that this should probably be turned into a bitmap of flags, instead of a single flag field.

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -88,6 +96,7 @@ const MAX_FRAME_BODY_SIZE: usize = MAX_FRAME_SIZE - FRAME_HEADER_SIZE;
 #[derive(Clone)]
 struct Frame {
     method_or_status: MethodOrStatus,
+    message_id: MessageId,
```
tiziano88, line 99: FYI (not specifically to this line), this may be an interesting read for you since you are getting deeper and deeper into these packing and alignment details: http://www.catb.org/esr/structure-packing/ (I haven't read it myself though -- if you do, let me know whether it is actually relevant for us)

tiziano88 (commented):
cc @ipetr0v too since this will also have to be implemented in the untrusted runtime in C++ in order to be able to communicate with the trusted runtime.

Though I would suggest prioritizing #2961 to check in the format first, and then get back to the two implementations based on that spec (of course it does not mean that we cannot change it later on, but at least it will be obvious when we are making backwards incompatible changes, which we will have to be careful about).

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -38,18 +44,20 @@ type FrameLength = u32;
...
 
-// TODO(#2848): Use this space to send a stream identifier that identifies
-// frames as belonging to the same message. Currently the frames of every message
-// must be sent as an uninterrupted seqeuence. A stream identifier would allow us
-// to interleave messages.
-/// Reserved space to maintain 8 byte alignement.
-const PADDING_SIZE: usize = 4;
+// TODO(#2848): Create a seperate invocation ID on the message layer to link
+// request and response messages, seperating the layers.
+/// Links the frames that constitute a message. Links request messages and
```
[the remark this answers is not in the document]
tiziano88: Also, is multiplexing required for any use case already? If not, can we get away in the C++ implementation to assume no interleaving (i.e. always send all the frames for a given message before switching to the next one?) 

This PR (and anything else necessary for multiplexing) can still go ahead, as I think it will be useful, but I just wonder if in the short term we can get away without implementing all of this logic in C++ in order to get the stack up and running.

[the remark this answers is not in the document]
conradgrobler: Regarding the flakiness: since #2967 I am no longer experiencing that. I was unable to get it to stall even when running 100 000 iterations of the test. It is possible that the stalling was caused by some wrapping issue when doing flow control credit calculations with a stream buffer size set at the maximum value that the datatype could handle.

experimental/oak_baremetal_channel/src/client.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Default)]
+struct MessageIdCounter {
+    next_message_id: MessageId,
+}
+
+impl MessageIdCounter {
+    fn next_message_id(&mut self) -> MessageId {
+        let next_message_id = self.next_message_id;
+        let _ = self.next_message_id.overflowing_add(1);
```
conradgrobler, line 66: I don't think this works. If I understand the documentation correctly `overflowing_add` returns a tuple which includes the sum, but does not increment the message id.

Also, if the code does not check whether it was overflowing or not it's probably better to use `wrapping_add`, or just use `Wrapping<u32>` as the type which would always wrap.

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -385,7 +342,7 @@ where
 }
 
 /// Public struct used to send and receive messages over an underlying channel.
```
conradgrobler, line 344: This comment seems out of date.

experimental/oak_baremetal_channel/src/lib.rs
conradgrobler: +1: either that, or rename the field. The expection is usually that something named `flag` is a boolean and something names `flags` represents a set of bit flags.

But that can be included as part of the next framing layer redesign iteration.

conradgrobler (approved):

experimental/oak_baremetal_channel/src/lib.rs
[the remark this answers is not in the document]
jul-sh: > I think the terminology message id is a bit misleading here: this is really the id of an invocation, which is exemplified by the fact that (AFAICT) it is shared between the request and response message, so it does not uniquely identify a message (in my original IDL design I called it transaction id).

> Also see https://fuchsia.dev/fuchsia-src/reference/fidl/language/wire-format#transactional-messages for some more (non-normative) ideas.

> I think in our case we can stick with invocation instead of transaction for the time being, though we can keep thinking about it longer term (e.g. if we decide to support streaming multiple requests or responses instead of just unary invocations).

I don't disagree, the use of shared ID was based on the version of the design doc we reviewed together in the on demand slot. I agree that ideally it would be be better to use two separate IDs for linking, and for linkning requests/responses. 

Imo sensible names would be `MessageId` for the ID to link the frames that constitute a message, and `InvocationId`, for linking requests with responses. 

We can discuss it more and decide in a committed spec.

PR #2967, opened 2022-06-15
Update credit calculation constants

conradgrobler:
The original choice of stream buffer length that we chose caused the comms to stall if we received a lot of data without sending any packets back, even though it indicated a very large capacity.

This updates value seems to avoid stalling without sending too many credit update packets.

### DIFF
--- a/experimental/virtio/src/vsock/socket/mod.rs
+++ b/experimental/virtio/src/vsock/socket/mod.rs
@@ -16,7 +16,7 @@
 
 use super::{
     packet::{Packet, VSockFlags, VSockOp, HEADER_SIZE},
-    VSock, DATA_BUFFER_SIZE, QUEUE_SIZE,
+    VSock, DATA_BUFFER_SIZE,
 };
 use alloc::collections::VecDeque;
 use core::num::Wrapping;
--- a/experimental/virtio/src/vsock/socket/mod.rs
+++ b/experimental/virtio/src/vsock/socket/mod.rs
@@ -24,16 +24,16 @@ use rust_hypervisor_firmware_virtio::virtio::VirtioTransport;
 
 /// The maximum buffer size used by the socket.
 ///
-/// This is used for flow-control calculations. For now we use the maximum size seeing that we don't
-/// have a fixed limit and don't want to send too many credit update packets.
-const STREAM_BUFFER_LENGTH: Wrapping<u32> = Wrapping(u32::MAX);
+/// This is used for flow-control calculations. Seeing that we don't use an actual streambuffer,
+/// this is currently an arbitrary value that seems to avoid stalling.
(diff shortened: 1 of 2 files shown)


experimental/virtio/src/vsock/socket/mod.rs
```diff
@@ -16,24 +16,24 @@
...
 use rust_hypervisor_firmware_virtio::virtio::VirtioTransport;
 
 /// The maximum buffer size used by the socket.
 ///
-/// This is used for flow-control calculations. For now we use the maximum size seeing that we don't
-/// have a fixed limit and don't want to send too many credit update packets.
-const STREAM_BUFFER_LENGTH: Wrapping<u32> = Wrapping(u32::MAX);
+/// This is used for flow-control calculations. Seeing that we don't use an actual streambuffer,
+/// this is currently an arbitrary value that seems to avoid stalling.
+const STREAM_BUFFER_LENGTH: Wrapping<u32> = Wrapping(1024 * 1024);
```
jul-sh, line 29: To rephrase, we currently don't use an actual stream buffer in the guest (since just use the available memory). This number therefore only concerns what we communicate to the host.

Correct?

experimental/virtio/src/vsock/socket/mod.rs
```diff
@@ -16,24 +16,24 @@
...
 
 /// The limit that triggers a voluntary credit update message to avoid stalling.
 ///
 /// If the peer's calculation of our free buffer space falls below this point (e.g when we receive a
 /// lot of data without sending any packets back) we send a credit update packet to make sure the
 /// peer knows we have more space available.
-const CREDIT_UPDATE_LIMIT: Wrapping<u32> = Wrapping((DATA_BUFFER_SIZE * QUEUE_SIZE) as u32);
+const CREDIT_UPDATE_LIMIT: Wrapping<u32> = Wrapping(512 * 1024);
```
jul-sh, line 36: What are those numbers based on? Experimentation? 

It seems 1024 is shared across both constants, would it make sense to extract it as a shared const?

jul-sh (approved):
nice, thanks

ipetr0v (approved):

experimental/virtio/src/vsock/socket/mod.rs
conradgrobler: Yes, that is correct.

experimental/virtio/src/vsock/socket/mod.rs
conradgrobler: Yes, it is experimentation. It will be based on an actual stream buffer when #2968 is implemented.

The 1024 is just there to give and indication of how many KiB this represents. I think it is more readable than the raw numbers, but apart from that the 1024 does not represent anything specific.

conradgrobler (commented):
Thanks for the reviews.

PR #2970, opened 2022-06-15
Baremetal Communication Protocol Specification

### LINKED ISSUE #2961: Document frame / message wire format

jul-sh:
Fixes #2961


experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,116 @@
...
+A message forms one half of an Invocation.
+
+### Request Message Definition
+
+A byte encoded request message MUST consist of the following seqeuential fields:
+
+- length, u32, little endian: length of the entire message, including this field
+- invocation_id, u32, little endian: Identifies the invocation the request
+  initiates
+- method_id, u32, little endian: identifies the method that is being invoked
```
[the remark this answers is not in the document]
jul-sh: See the invocation section: "Each invocation consists out of a Request Message
followed by a Response Message.". :) 

In essence it's a request/response pair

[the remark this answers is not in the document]
jul-sh: Whereas a method is function exposed by the trusted runtime that the runtime may invoke. See the methods currently supported by the runtime in the schema: https://github.com/project-oak/oak/blob/main/experimental/oak_baremetal_channel/schema.fbs

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,118 @@
...
+- padding, 4 bytes: reserved to maintain 8 byte alignment. MUST be unset and
+  ignored
+- body, variable length byte array: the method’s arguments, encoded with
+  flatbuffer
```
tiziano88, line 54: In the same spirit as my other comments, I would not even mention flatbuffer here, and just leave it as an opaque byte blob from the point of view of this spec. Your call.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,121 @@
...
+
+A byte encoded request message MUST consist of the following seqeuential fields:
+
+- length, u32, little endian: length of the entire message, including this field
+- invocation_id, u32, little endian: Identifies the invocation the request
+  initiates
+- method_id, u32, little endian: identifies the method that is being invoked
+- padding, 4 bytes: reserved to maintain 8 byte alignment. MUST be unset and
+  ignored
+- body, variable length byte array: the method’s arguments, encoded with
```
conradgrobler, line 53: nit: "encoded with flatbuffer" feels a bit weird. I am not 100% sure about the correct terminology, but would have expected something more like "encoded as FlatBuffers"

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,121 @@
...
+  [gRPC status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+- status_message_length, u32, little endian: length of the status message
+- status_message, utf-8 encoded string: developer-facing error message
+- body, variable length byte array: the method’s return value, encoded with
+  flatbuffer
+
+### Sending Messages
+
+To send a message, it MUST first be encoded into frames. The resulting frame set
+MUST then be send as an uninterrupted sequence.
```
conradgrobler, line 74: "... be sent ..."

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,121 @@
...
+### Encoding Messages into Frames
+
+The byte representation of a message MUST be chunked. The last chunk MAY be
+shorter than 63,984 bytes (the maximum size of a frame body). Any preceeding
+chunks MUST be exactly 63,984 bytes in length.
+
+A frame MUST then be formed for each resulting message chunk, with the message
+chunk forming the frame body.
+
+Messages shorter than 63,984 bytes MUST be encoded as a single frame.
```
conradgrobler: nit: or equal to?

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,121 @@
...
+  The length of a frame MUST NOT exceed 64000 bytes
+- flags, two bytes: Bitflags, see [Frame Flags](#3-frame-flags-definition)
+- padding, two bytes: reserved to maintain 8 byte alignment
+- body, variable length byte array: Byte chunk of an encoded message. In order
+  to not exceed the maximum frame length, this byte array MUST NOT exceed 63,984
+  bytes
+
+### Frame Flags Definition
+
+- 0x1: MUST be set for frames that are the first of the set that constiutes a
```
conradgrobler, line 117: nit: The case where no bits are set should probably be explicitly mentioned for clarity.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,121 @@
...
+## Frames
+
+Messages are sent and received as frames. Each message can be fragmented into
+one or more frames.
+
+### Frame Definition
+
+A byte encoded frame MUST consist of the following seqeuential fields:
+
+- padding, four bytes: for future use
```
jul-sh, line 107: The idea would be to use this for a message_id in the future, allowing us to support multiplexing

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,118 @@
...
+
+## Messages
+
+A message forms half of an Invocation.
+
+### Request Message Definition
+
+A byte encoded request message MUST consist of the following seqeuential fields:
+
+- length, u32, little endian: length of the entire message, including this field
```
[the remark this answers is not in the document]
conradgrobler: There are two main reasons for the length. One is that is serves as a simple consistency check to ensure all the frames were received. The second is that it would allow us to pre-allocate a buffer for collecting all the frames at the start rather than to keep reallocating as we receive more frames.

Another possible approach is to always specify the entire message length in the length field of the frame, seeing that all frames except the last frame of the messages has a fixed length anyway.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,118 @@
...
+- invocation_id, u32, little endian: MUST match the invocation_id of the
+  relevant Request Message
+- status_code, u32, little endian: response status as a
+  [gRPC status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+- status_message_length, u32, little endian: length of the status message
+- status_message, utf-8 encoded string: developer-facing error message
+- body, variable length byte array: the method’s return value, encoded with
+  flatbuffer
```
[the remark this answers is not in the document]
jul-sh: Good question! I was thinking about this too, and I don't lean too strongly either way. 

The reason I went with the current approach is that this is [what gRPC does](https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto), and we use gRPC status codes.

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Maybe, yeah. :) Updated

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Fair, moved the mention of the encoding protocol up to method section, and replaced this. :)

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
tiziano88: Doesn't gRPC convert this status message into the response body though?

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,118 @@
...
+## Method
+
+Methods are exposed by the Trusted Runtime. They MAY be remotely invoked by the
+Unrusted Loader. Each method has clearly defined parameters and return types.
+The Oak IDL is used to define the methods exposed by the Trusted Runtime.
+
+## Invocation
+
+An invocations of a Method. Each invocation consists out of a Request Message
+followed by a Response Message.
```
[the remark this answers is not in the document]
jul-sh: Updated to use client and service :)

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
tiziano88: I think we should have the invocation_id in the C API if we allow the runtime to have interleaved requests; anyways this is fine, it's also fine to have the C API repeat the length, not a big deal (though it's redundant since it will also have to be passed as an arg to the C API which will probably have a signature of `invoke(buf, len)` or similar).

ipetr0v:
Could you please add a diagram that shows all of the layers involved. Similar to how it's done for IP: https://en.wikipedia.org/wiki/IPv4#Header

IIUC right now we have 4 different layers involved:
- **Application layer**
  - Contains data meant for the WASM application (i.e. user data). Encoding is application dependent
- **Oak layer**
  - Contains structs meant for Oak. Defined in the `schema.fbs`. Encoding is Flat Buffers
- **Message Layer**
  - Defined in the `struct Message` (has 2 types of messages: `Request` and `Response`). Encoding is length prefixed.
- **Frame Layer**
  - Defines frames in which messages are split. Encoding is length prefixed.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+## Frames
+
+Messages are sent and received as frames. Each message can be fragmented into
+one or more frames.
+
+### Frame Definition
+
+A byte encoded frame MUST consist of the following seqeuential fields:
+
+- padding, four bytes: for future use
```
tiziano88, line 110:
```suggestion
- padding, 4 bytes: for future use
```

(etc.)

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
jul-sh: Fair. :) I'd be inclined to keep it then, I don't see a very good option for avoiding having it here. Only ones are remove it (bad for reasons above), move it to the framing layers (adds some coupling, but maybe ok?), or add another layer on top (don't love the complexity of that).

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+  to not exceed the maximum frame length, this byte array MUST NOT exceed 63,936
+  bytes
+
+### Frame Flags Definition
+
+- 0x0 (LSB): MUST be set for frames that are the first of the set that
+  constiutes a message
+- 0x0: MUST be set for frames that are the last of the set that constiutes a
+  message
+- 0x3-0x15: Reserved for future use. MUST be unset and ignored
```
tiziano88, line 124: The numbers are inconsistent.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+- body, variable length byte array: byte encoded method return value
+
+### Sending Messages
+
+To send a message, it MUST first be encoded into frames. The resulting frame set
+MUST be sent as an uninterrupted sequence.
+
+### Receiving Messages
+
+To receive a message the recipent incrementaly reads frames from the
```
tiziano88, line 81:
```suggestion
To receive a message the recipent incrementally reads frames from the
```

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+communication channel. The recipent MAY start parsing a message before all of
+its frames have been received.
+
+### Encoding Messages into Frames
+
+The byte representation of a message MUST be chunked. The last chunk MAY be
```
tiziano88, line 87: Can we avoid introducing the term chunk here, or define it?

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+shorter than 63,936 bytes (the maximum size of a frame body). Any preceeding
```
tiziano88, line 88:
```suggestion
shorter than 63,936 bytes (the maximum size of a frame body). Any preceding
```

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
jul-sh: Looked into a bit more, and while gRPC does not use the body to store the status message, it does seem to omit it for non error responses. :) https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors

Which means we probably could indeed follow your suggestion. The only downside would be that it adds some coupling between the gRPC status messages and our message format (since we depend on gRPCs defintion of "Ok").

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,124 @@
...
+
+### Decoding Frames into Messages
+
+Messages are parsed from Frames by appending the frame bodies in the order they
+are received.
+
+## Frames
+
+Messages are sent and received as frames. Each message can be fragmented into
+one or more frames.
```
tiziano88, line 104:
```suggestion
a set of one or more frames.
```

(optional, to use the terminology that you use below)

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
jul-sh: Updated to store it in the body, but happy to revert to using a dedicated field if you think it'd be better. :P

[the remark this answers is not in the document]
tiziano88: Either way, I would avoid having a len-delimited field inside an already len-delimited message.

Personally I think we can just say if the status is OK (whatever status code we decide to represent that), then the body contains the response; otherwise, the body contains the error message supporting the non-OK status.

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: yeah 😅, sorry

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,126 @@
...
+- length, u16, little endian: Length of the entire frame, including this field.
+  The length of a frame MUST NOT exceed 64000 bytes
+- flags, two bytes: Bitflags, see [Frame Flags](#3-frame-flags-definition)
+- body, variable length byte array: Byte chunk of an encoded message. In order
+  to not exceed the maximum frame length, this byte array MUST NOT exceed 63,936
+  bytes
+
+### Frame Flags Definition
+
+- 0x0 (LSB): MUST be set for frames that are the first of the set that
```
tiziano88, line 122: I always get this wrong (confusing little vs big endian, and LSB vs MSB); to confirm, is this expected to be the left-most bit in the message as serialized?

jul-sh: Yes.

jul-sh: Do you think we should use that language? MSB / LSB feels a bit cryptic :P

experimental/oak_baremetal_channel/SPEC.MD
tiziano88: Up to you, maybe it's fine as it is, if we assume it means "the frame body"?

tiziano88 (approved):
LG, thanks! Please consider @ipetr0v 's suggestion of a diagram that shows the layering in visual form too.

@ipetr0v please make sure that you have enough information from this spec only to reimplement the entire protocol in C++.

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: added examples for clarity. :)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Replaced chunks with frame bodies when used a noun, agreed that it's clearer. :-)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: AFAICT the numbering in the PR seems to be correct (or at least be matching fuchsia). Fuchsia labels 0 as the LSB (as do we), and appears to start using flags starting 0, with larger values unused (as do we). See https://fuchsia.dev/fuchsia-src/reference/fidl/language/wire-format#flags, Flags[0].

tiziano88: Perhaps we don't need to define endianness of the bytes field, and we could just talk about the two bytes independently (`flags[0]` and `flags[1]`). But I don't think we can assume the same endianness across the rest of the spec. I mean, at some point we need to parse a u32 from bytes, and we need to know which endianness it is in order to call the appropriate Rust / C++ method, right? I don't think we would want to just blindly treat those bytes as an int of whatever endianness the machine happens to be running in.

jul-sh: Fair, let's keep it then. I guess the downside is that if we do ever run on a processor that uses big endian natively we'd have to a lot of conversion. But seems unlikely. 

Interesting trivia fact (and what gave me the idea): Many common protocols used for actual networking between physical computers do not define endianness. UDP for one does not 😃

experimental/oak_baremetal_channel/SPEC.MD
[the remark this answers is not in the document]
jul-sh: implemented like that :)

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,296 @@
...
+
+Length of the entire frame, including this field. The length of a frame MUST NOT
+exceed 64000 bytes
+
+##### flags, two bytes, little endian
+
+###### flags[0]
+
+```
+- 7 (MSG): unused
```
tiziano88, line 202:
```suggestion
- 7 (MSB): unused
```

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,296 @@
...
+##### padding, 4 bytes
+
+for future use
+
+##### length, u16, little endian
+
+Length of the entire frame, including this field. The length of a frame MUST NOT
+exceed 64000 bytes
+
+##### flags, two bytes, little endian
```
tiziano88, line 197: If we treat them as separate bytes (which I think is a good idea), then endianness does not really apply, does it? They are just two consecutive bytes (and it's clear that [0] comes before [1])

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,296 @@
...
+
+```
+- 7 (MSG): unused
+- 6: unused
+- 5: unused
+- 4: unused
+- 3: unused
+- 2: unused
+- 1: end flag. MUST be set in the last frame of a message
+- 0 (LSB): start flag. MUST be set in the first frame of a message
```
tiziano88, line 209: In a single byte, the LSB is on the "right", so I think this is the wrong way around?

e.g. in `00000001` , the 1 is the LSB (and endianness is irrelevant for this, it's just how bytes and numbers are normally read)

So I think the fuchsia page does it correctly, i.e. the MSB (i.e. bit #7) is the "first" one.

Please others double check this though, I don't trust myself to always get this right :)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: classic juliette :P

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: I think we can remove it, since we indicate LSB and MSB in the flag section later. If represented as seperate bytes the order should be clear :)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Hehe you really shouldn't rely on me out of all people to get this right. :p 

But my two cents nonetheless: It seems to me that the convention is to start counting flags from the LSB on the right. 

I think because A) Fuchsia does this, where they list the bits in descending order from MSB to LSB (but in Flags[0] where two flags are defined and 6 undefined the defined flags start the LSB). 
B) Likewise the same appears to be the case here: https://blog.podkalicki.com/bit-level-operations-bit-flags-and-bit-masks/

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,296 @@
...
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         frame padding                         |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|          frame length         |          frame flags          |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++   frame body [ message body [ method encoded_parameters ] ]   +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
jul-sh, line 295: For future reference if we need to edit this: It was generated with [Protocol](https://www.luismg.com/protocol/), using the following schema:

`"frame padding:32,frame length:16,frame flags:16,frame body [ message body [ method encoded_parameters ] ]:64"`

thx @tiziano88 for recommending the tool

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,296 @@
...
+|                                                               |
++                                                               +
+|                                                               |
++   frame body [ message body [ method encoded_parameters ] ]   +
+|                                                               |
++                                                               +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ....
+```
```
jul-sh, line 280: Generated from the schema `frame padding:32,frame length:16,frame flags:16,frame body [ message length ]:32,frame body [ message invocation_id ]:32,frame body [ message method_id ]:32,frame body [ message padding ]:32,frame body [ message body [ method encoded_parameters ] ]:128`

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Layer Defintions
+
+### Method Layer
+
+Methods are exposed by the service. Each method has clearly defined parameters
+and return types. The Oak IDL is used to define which methods are available,
+their id, parameter and return types, as well the protcol used to encode
```
conradgrobler, line 22: "protcol" -> "protocol"

experimental/oak_baremetal_channel/SPEC.MD
tiziano88: Ah that's a good point, when implementing flag masks with bit shifts, we indeed usually start counting from the right. So yes, I think your current version makes sense. Anyways if we change our mind later on for whatever reason (e.g. if the implementation is more elegant one way or the other) we can always flip things around. Although now it looks a bit off that the first flag is in the right hand side of the first flag byte, which is like in between the two bytes. Anyways let's go ahead with this for now.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                            padding                            |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|             length            |             flags             |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                         body bytes...                         +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ...
+```
```
tiziano88, line 157: Is it possible to use hex for this? I think it would clarify where the byte boundaries are (but no worries if not easily possible). Something like

```
 0                               1                   
 0 1 2 3 4 5 6 7 8 9 A B C D E F 0 1 2 3 4 5 6 7 8 9 A B C D E F
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                            padding                            |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|             length            |             flags             |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                                                               |
+                         body bytes...                         +
|                                                               |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ...
```

Or maybe there are even better options that only count bytes instead of individual bits? Curious to see if the tool you used to generate it has any options for this.

tiziano88 (approved):

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
+##### Handling an invocation
+
+The service MUST respond to each request with a single response message. It
+SHOULD respond to requests in first-in-first-out order, but MAY prioritize some
+requests. The service MAY send a response before the request has been received
+in its entirety. Sending a response concludes the invocation.
+
+The client MUST NOT send response.
+
+### Message Layer
```
ipetr0v, line 45: Before describing each layer individually, could you please add a small overview of all of the layers (with their encodings) that will be described later?

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
+An invocation of a Method. Each invocation consists out of a request followed by
+a response.
+
+##### Initiating an invocation
+
+The client MAY initiate an invocation by sending a request.
+
+The service MUST NOT send requests.
+
+##### Handling an invocation
```
ipetr0v, line 36: Which layer does [`schema.fbs`](https://github.com/project-oak/oak/blob/main/experimental/oak_baremetal_channel/schema.fbs) belong to?

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
+The service MUST respond to each request with a single response message. It
+SHOULD respond to requests in first-in-first-out order, but MAY prioritize some
+requests. The service MAY send a response before the request has been received
+in its entirety. Sending a response concludes the invocation.
+
+The client MUST NOT send response.
+
+### Message Layer
+
+Messages are conceptual units to transport requests and responses.
```
ipetr0v, line 47: Could you please add a full list of message types here?
Also, is it possible to distinguish a message type by just looking at their bytes? There is no message type id in the diagrams below.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,290 @@
...
+Messages are sent and received as frames. Each message is fragmented into a set
+of one or more frames.
+
+#### Frame
+
+A byte encoded frame MUST consist of the following fields:
+
+```
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
```
ipetr0v, line 147: Why does it have
```
3
0 1
+-+
  |
```
at the end?
Is it two extra bits?

ipetr0v (approved):
Thanks!

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: yeah I know, it's not pretty with the two bytes. But it's what fuchsia does, so at least we're not alone :p

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Method layer :)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: No it's not possible. The reason being that service (trusted runtime) must only send response messages, whereas the client (untrusted loader) must only send request messages. Hence the service (trusted runtime) can safely assume that any message it receives is a request message. The client (untrusted loader) can safely assume that any message it receives is a response message. :)

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Not entirely familiar with all the design decisions behind the diagram format, but the draft RFC for it might have more context: https://datatracker.ietf.org/doc/html/draft-mcquistin-augmented-ascii-diagrams-10

The gist is that the width of the diagram is four bytes. The `0 1` you quoted are the last two bits of the fourth byte.

experimental/oak_baremetal_channel/SPEC.MD
jul-sh: Added a list. :)

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,293 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Layer Defintions
+
+### Method Layer
+
+Methods are exposed by the service. Each method has clearly defined parameters
+and return types. The Oak IDL is used to define which methods are available,
```
tiziano88, line 21: (nit) Strictly speaking, each method has one parameter and one return type.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,293 @@
+# Oak Baremetal Communication Protocol Specification
```
tiziano88, line 1: I think the spec is more general than Oak Baremetal, anyways we can move / rename later.

tiziano88 (approved):
More nits! :sweat_smile:

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,293 @@
...
+
+Identifies the invocation the request initiates.
+
+##### method_id, u32, little endian
+
+Identifies the method that is being invoked.
+
+##### padding, 4 bytes
+
+Reserved to maintain 8 byte alignment. MUST be unset and ignored.
```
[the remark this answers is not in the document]
conradgrobler: The exact alignment requirements depend on the operations being performed.

With the current implementation I agree that 8-byte alignment might not be needed. We are not accessing the bytes in the buffer directly via unsafe casts, and all the fixed-length fields are 32 bits or shorter anyway (for which 4 byte alignment is sufficient). I don't think we are planning on directly interpreting the body bytes through unsafe casts, so the alignment of the body bytes is also not important at the moment.

experimental/oak_baremetal_channel/SPEC.MD
```diff
@@ -0,0 +1,293 @@
...
+
+Messages are conceptual units to transport requests and responses.
+
+There's exactly two messages types: request messages and response messages.
+
+#### Request Message
+
+A byte encoded request message MUST consist of the following fields:
+
+```
```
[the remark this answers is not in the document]
jul-sh: done, thanks for recommending protocol for generating them btw

PR #2971, opened 2022-06-15
Remove references to (now broken) UEFI from the loader.

andrisaar:


andrisaar:
> I think all of this is covered by #2972 that I just created.

Yep, seems that way. I'll drop this.

PR #2972, opened 2022-06-15
Remove remaining UEFI-specific support

conradgrobler:


tiziano88 (approved):
LGTM but I may be missing something as I'm not the most familiar person with this code, so if you want an extra pair of eyes feel free to ask someone else to take a look.

andrisaar (approved):
I think we can drop OVMF in the Dockerfile as well now.

conradgrobler:
> I think we can drop OVMF in the Dockerfile as well now.

Good point, thanks. I have now removed that, and also some additional dependencies that were added to build `ring` for UEFI.

jul-sh (approved):
thank you!

PR #2976, opened 2022-06-17
Add lookup data support to the baremetal runtime

conradgrobler:
This PR makes the `oak_functions_lookup` crate fully compatible with `no_std` use and adds an initial implementation with fixed lookup data to the baremetal runtime.


oak_functions/sdk/oak_functions/tests/integration_test.rs
```diff
@@ -141,9 +142,8 @@ async fn test_write_log() {
 
 #[tokio::test]
 async fn test_storage_get_item() {
-    let entries = hashmap! {
-       b"StorageGet".to_vec() => b"StorageGetResponse".to_vec(),
-    };
+    let entries =
+        HashMap::from_iter([(b"StorageGet".to_vec(), b"StorageGetResponse".to_vec())].into_iter());
```
tiziano88, line 146: Not sure the from_iter / into_iter is actually doing anything here.

experimental/oak_baremetal_channel/schema.fbs
```diff
@@ -3,6 +3,7 @@ attribute "method_id";
 rpc_service TrustedRuntime {
     HandleUserRequest(UserRequest) : UserRequestResponse (method_id: 0);
     Initialize(Initialization) : Empty (method_id: 1);
+    UpdateLookupData(LookupData) : Empty (method_id: 2);
```
tiziano88, line 6: We should document these methods at some point (either here, or perhaps better, in a markdown file somewhere that we can link from here).

Could you add a comment here for now? Mainly to clarify whether this is a total or partial update (I guess it's total, right?).

oak_functions/abi/src/lib.rs
```diff
@@ -56,11 +57,57 @@ pub mod proto {
...
+        len_buffer.copy_from_slice(&buffer[..LENGTH_SIZE]);
+        let len = u64::from_le_bytes(len_buffer);
+
+        if buffer.len() != len as usize + LENGTH_SIZE {
+            anyhow::bail!(
+                "Invalid buffer: expected buffer size of {}, but found {}",
+                len as usize + LENGTH_SIZE,
+                buffer.len()
+            )
+        }
```
jul-sh, line 103: [optional] It seems like encoding length serves two purposes: a) differentiate `StorageGetItemResponse { value: None }` from `StorageGetItemResponse { value: Vec::new() }`, and b) validation. It might be nice comment that, so it doesn't get dropped if e.g. validation is no longer needed

experimental/oak_baremetal_runtime/src/framing.rs
```diff
@@ -144,6 +184,7 @@ where
 {
     let mut invocation_handler = InvocationHandler {
         initialization_state: InitializationState::Uninitialized(Some(attestation_behavior)),
+        lookup_data_manager: Arc::new(LookupDataManager::new_empty(StandaloneLogger::default())),
```
tiziano88, line 187: How is `Arc` implemented on no_std? Does it use some atomic increment x86 instruction?

experimental/oak_baremetal_channel/schema.fbs
```diff
@@ -22,4 +23,13 @@ table Initialization {
...
 }
 
+table LookupDataEntry {
+    key: [ubyte];
+    value: [ubyte];
+}
+
+table LookupData {
+    items: [LookupDataEntry];
+}
```
jul-sh, line 33: Later on we may want to add methods to build up a LookupData file incrementally, so parsing doesn't happen all at once

jul-sh (approved):
nice, thank you

tiziano88 (approved):
Nice, thanks!

experimental/oak_baremetal_runtime/src/framing.rs
```diff
@@ -132,6 +138,40 @@ where
...
+                entry
+                    .key()
+                    .ok_or_else(|| oak_idl::Status::new(oak_idl::StatusCode::InvalidArgument))?
+                    .to_vec(),
+                entry
+                    .value()
+                    .ok_or_else(|| oak_idl::Status::new(oak_idl::StatusCode::InvalidArgument))?
+                    .to_vec(),
+            );
+        }
```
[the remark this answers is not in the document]
conradgrobler: Yes, that works with the addition of an extra type parameter for Result and an Ok(...) in the closue. Thanks.

experimental/oak_baremetal_channel/schema.fbs
conradgrobler: Yes, I agree.

experimental/oak_baremetal_channel/schema.fbs
conradgrobler: done

oak_functions/sdk/oak_functions/tests/integration_test.rs
conradgrobler: This is the simplest way I found to convert an array of tupples into a hashmap. Even though I though it should work based on the documentation (https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.from) using `from` or `into` does not work. The compiler complains that "the trait bound `hashbrown::HashMap<Vec<u8>, Vec<u8>>: From<[(Vec<u8>, Vec<u8>); 1]>` is not satisfied".

oak_functions/abi/src/lib.rs
conradgrobler: Added a comment to clarify.

conradgrobler (commented):
Thanks for the reviews.

PR #2978, opened 2022-06-18
Add support for kernel boot arguments

andrisaar:
The idea is that in the future we may want to pass arguments to the kernel even before we establish our communication channel: for example, whether to use serial or vsock-virtio (or vsock-console) for said communication channel to begin with.

Part of this PR is also a request for ideas: I'm not happy about `args::args()` right now, but I couldn't figure out how to convince Rust that the memory is static so that a `&'static CStr` makes sense.

### DIFF
--- a/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
+++ b/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
@@ -61,8 +61,7 @@ impl BootInfo<boot_e820_entry> for boot_params {
     fn args(&self) -> &CStr {
         if self.hdr.cmdline_size == 0 {
             Default::default()
-        }
-        else {
+        } else {
             // Safety: Linux boot protocol expects the pointer to be valid, even if there are no
             // args.
             unsafe { CStr::from_ptr(self.hdr.cmd_line_ptr as *const c_char) }
--- a/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
+++ b/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
@@ -19,6 +19,7 @@
  *   --ctypes-prefix c_types arch/x86/include/uapi/asm/bootparam.h
  */
 
+use core::ffi::{c_char, CStr};
 use oak_baremetal_kernel::boot::{BootInfo, E820Entry, E820EntryType};
 
 #[allow(non_camel_case_types)]
@@ -55,6 +56,15 @@ impl BootInfo<boot_e820_entry> for boot_params {
 
     fn e820_table(&self) -> &[boot_e820_entry] {
         &self.e820_table[..self.e820_entries as usize]
+    }
+
+    fn args(&self) -> &CStr {
+        if self.hdr.cmdline_size == 0 {
+            // We know the constant ends with a \0, so the unwrap will always succeed.
+            return CStr::from_bytes_with_nul(b"\0").unwrap();
+        }
+        // Safety: Linux boot protocol expects the pointer to be valid, even if there are no args.
+        unsafe { CStr::from_ptr(self.hdr.cmd_line_ptr as *const c_char) }
     }
 }
 
--- a/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
+++ b/experimental/oak_baremetal_app_crosvm/src/bootparam.rs
@@ -60,11 +60,13 @@ impl BootInfo<boot_e820_entry> for boot_params {
 
     fn args(&self) -> &CStr {
         if self.hdr.cmdline_size == 0 {
-            // We know the constant ends with a \0, so the unwrap will always succeed.
-            return CStr::from_bytes_with_nul(b"\0").unwrap();
+            Default::default()
         }
-        // Safety: Linux boot protocol expects the pointer to be valid, even if there are no args.
-        unsafe { CStr::from_ptr(self.hdr.cmd_line_ptr as *const c_char) }
+        else {
+            // Safety: Linux boot protocol expects the pointer to be valid, even if there are no
+            // args.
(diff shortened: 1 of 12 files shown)


experimental/oak_baremetal_app_crosvm/src/bootparam.rs
```diff
@@ -56,6 +57,15 @@ impl BootInfo<boot_e820_entry> for boot_params {
     fn e820_table(&self) -> &[boot_e820_entry] {
         &self.e820_table[..self.e820_entries as usize]
     }
+
+    fn args(&self) -> &CStr {
+        if self.hdr.cmdline_size == 0 {
+            // We know the constant ends with a \0, so the unwrap will always succeed.
+            return CStr::from_bytes_with_nul(b"\0").unwrap();
```
tiziano88, line 64: (nit) you don't need an explicit return here if you also fill in the else branch of the if.

tiziano88, line 64: I think this could also be `CStr::default()`?

experimental/oak_baremetal_app_qemu/src/multiboot.rs
```diff
@@ -55,6 +56,16 @@ impl BootInfo<multiboot_mmap_entry> for multiboot_info {
             )
         }
     }
+
+    fn args(&self) -> &CStr {
+        if self.flags & (1 << 2) == 0 {
```
tiziano88, line 61: What are these flags? Could you give a name to the `2` constant, and point to some documentation of what it means?

experimental/oak_baremetal_kernel/src/args.rs
```diff
@@ -0,0 +1,74 @@
...
+// limitations under the License.
+//
+
+use atomic_refcell::AtomicRefCell;
+use core::ffi::CStr;
+use lazy_static::lazy_static;
+
+// This is a really crude analogue of CString, which we can't use, as CString requires allocation.
+// Instead of allocation, we rely on a static buffer.
+struct StaticCString {
```
tiziano88, line 23: I wonder if https://docs.rs/arrayvec/0.7.2/arrayvec/struct.ArrayString.html may help here?

experimental/oak_baremetal_kernel/src/args.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn as_c_str(&self) -> &CStr {
+        // Safety: this is safe as initally len will be zero, resulting in an empty CStr;
+        // otherwise, the fields are initialized from init_args() from a pre-existing valid CStr.
+        unsafe { CStr::from_bytes_with_nul_unchecked(&self.buf[..self.len]) }
+    }
+}
+
+lazy_static! {
+    static ref ARGS: AtomicRefCell<StaticCString> = Default::default();
+}
```
tiziano88, line 49: Would it work to just use plain [mut static](https://doc.rust-lang.org/reference/items/static-items.html#mutable-statics)?

```rust
static mut ARGS: [u8; 512];
```

Though that would require unsafe every time, so maybe not.

tiziano88 (approved):

experimental/oak_baremetal_app_crosvm/src/bootparam.rs
andrisaar: Well, `Default::default()`, but yes, I missed the obvious way how to do it cleanly; thanks.

experimental/oak_baremetal_app_qemu/src/multiboot.rs
andrisaar: I've added a comment for now explaining what the constant is. I want to Rust-ify the data structures used here in the future instead of using bindgen, so I think it's fine for now.

experimental/oak_baremetal_kernel/src/args.rs
andrisaar: Yep, that's much better!

experimental/oak_baremetal_kernel/src/args.rs
andrisaar: I went back and forth between `lazy_static` and plain `static mut` several times, trying to reduce the amount of unsafe code as much as possible. With the `StaticCString` I think `lazy_static` made sense.

Now that I've switched over to `ArrayString`, I think the tables have turned: just plain `static mut` is good enough and more readable.

conradgrobler (approved):
Nice, thanks.

experimental/oak_baremetal_kernel/src/args.rs
```diff
@@ -0,0 +1,40 @@
...
+/// This function is intended to be called fairly early in the boot process, when we don't even
+/// have memory allocation available. This also means that we can't use anyhow::Result as the
+/// return value, as anyhow relies on allocation.
+pub fn init_args(args: &CStr) -> core::result::Result<(), &str> {
+    let args = args
+        .to_str()
+        .map_err(|_| "Kernel arguments are not valid UTF-8")?;
+    // Safety: this is called once early in the initialization process from a single thread, so
+    // there will not be any concurrent writes.
+    unsafe { ARGS.try_push_str(args) }.map_err(|_| "Kernel arguments too long")
```
tiziano88, line 33: I would clarify the error so that it's obvious that nothing else is being ignored.

```suggestion
    unsafe { ARGS.try_push_str(args) }.map_err(|arraystring::error::OutOfBounds| "Kernel arguments too long")
```

tiziano88 (approved):

experimental/oak_baremetal_kernel/src/args.rs
andrisaar: Done.

PR #2979, opened 2022-06-20
Update bare metal channel implementation to match the specification

jul-sh:
Achieves full compliance of the runtime with specification defined in #2970.

The only missing piece for full spec compliance of the untrusted loader is adding support to the untrusted loader for receiving responses out of order. The spec says the runtime may send them out of order (but it currently does not). Will implement that in a separate PR.

The diff is relatively large, but a lot of that is down to renaming and moving the framing implementation to its own module.


experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -350,28 +50,54 @@ where
...
+                false => {
+                    encoded_message.append(&mut frame.body);
+                    if frame.flags.contains(frame::Flags::END) {
+                        break;
+                    }
+                }
+            };
         }
+
+        return Ok(M::decode(encoded_message));
```
jul-sh, line 93: No more recursion 😇

experimental/oak_baremetal_channel/src/message.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn encode(self) -> Vec<u8>;
+    fn decode(frames: Vec<u8>) -> Self;
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct RequestMessage {
+    pub invocation_id: InvocationId,
+    pub method_id: MethodId,
+    pub body: Vec<u8>,
+}
```
jul-sh, line 54: Introduce separate structs for request- and response messages, matching the specification. There's plenty of overlap in their encoding implementation, but imo that's ok (both are tested) and means we no longer have to deal with the somewhat ugly `method_id_or_status` of past days.

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -350,28 +50,70 @@ where
...
-        let frame = self.inner.read_frame()?;
-        match partial_message.add_frame(frame)? {
-            CompletionResult::Complete(message) => Ok(message),
-            CompletionResult::Incomplete(partial_message) => {
-                self.recursively_complete_partial_message(partial_message)
+    pub fn read_message<M: message::Message>(&mut self) -> anyhow::Result<M> {
+        let mut encoded_message: Vec<u8> = Vec::new();
+        let mut first_frame = self.inner.read_frame().context("failed to read frame.")?;
+
+        match (
```
conradgrobler, line 60: nit: this is just personal preference, but I think splitting this into two distinct if statements will make the code easier to read:

```
if !first_frame.flags.contains(frame::Flags::START)  {
    anyhow::bail!(...);
}
```

And then a separate if statemoer on whether the end bit is set or not.

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -350,28 +50,70 @@ where
...
+            }
+            _ => {
+                anyhow::bail!("expected a frame with the START flag set.")
             }
+        };
+
+        loop {
+            let mut frame = self.inner.read_frame().context("failed to read frame.")?;
+
+            match frame.flags.contains(frame::Flags::START) {
```
conradgrobler, line 96: nit: once again personal prefernce, but I find it easier to read if statements rather than a match on a bool.

experimental/oak_baremetal_channel/src/message.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn decode(mut encoded_message: Vec<u8>) -> Self {
+        let invocation_id = {
+            let mut invocation_id_bytes: [u8; INVOCATION_ID_SIZE] = [0; INVOCATION_ID_SIZE];
+            invocation_id_bytes.copy_from_slice(
+                &encoded_message[INVOCATION_ID_OFFSET..(INVOCATION_ID_OFFSET + INVOCATION_ID_SIZE)],
+            );
+            InvocationId::from_le_bytes(invocation_id_bytes)
+        };
+        let method_id = {
+            let mut invocation_id_bytes: [u8; METHOD_ID_SIZE] = [0; METHOD_ID_SIZE];
```
conradgrobler, line 85: nit: for consistency the variable should probably be method_id_bytes

experimental/oak_baremetal_channel/src/message.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        message_bytes.extend_from_slice(&length.to_le_bytes());
+        message_bytes.extend_from_slice(&self.invocation_id.to_le_bytes());
+        message_bytes.extend_from_slice(&self.method_id.to_le_bytes());
+        message_bytes.extend_from_slice(&[0; PADDING_SIZE]);
+        message_bytes.extend_from_slice(&self.body);
+
+        message_bytes
+    }
+    fn decode(mut encoded_message: Vec<u8>) -> Self {
```
conradgrobler, line 76: optional: perhaps add spacing between functions. The current case where there are spacing inside functions but no spacing between them makes it more difficult for me to quickly see the function boundaries.

experimental/oak_baremetal_channel/src/message.rs
```diff
@@ -0,0 +1,155 @@
...
+        let invocation_id = {
+            let mut invocation_id_bytes: [u8; INVOCATION_ID_SIZE] = [0; INVOCATION_ID_SIZE];
+            invocation_id_bytes.copy_from_slice(
+                &encoded_message[INVOCATION_ID_OFFSET..(INVOCATION_ID_OFFSET + INVOCATION_ID_SIZE)],
+            );
+            InvocationId::from_le_bytes(invocation_id_bytes)
+        };
+        let status_code = {
+            let mut invocation_id_bytes: [u8; STATUS_CODE_SIZE] = [0; STATUS_CODE_SIZE];
```
conradgrobler, line 139: nit: probably status_code_bytes

conradgrobler (approved):

experimental/oak_baremetal_channel/src/frame.rs
```diff
@@ -0,0 +1,161 @@
...
+
+type Length = u16;
+pub const LENGTH_SIZE: usize = 2;
+static_assertions::assert_eq_size!([u8; LENGTH_SIZE], Length);
+
+bitflags! {
+    #[derive(Default)]
+    pub struct Flags: u16 {
+        const START = 1;
+        const END = 2;
```
tiziano88, line 33: Could you add links to the spec for this and most other structs / fields / offsets?

experimental/oak_baremetal_channel/src/frame.rs
```diff
@@ -0,0 +1,161 @@
...
+        if length > MAX_SIZE {
+            return Err(anyhow::Error::msg("frame exceeds the maximum frame size"));
+        };
+
+        Length::try_from(length).map_err(|_error| {
+            anyhow::Error::msg(format!(
+                "could convert the frame length usize to {:?}",
+                core::any::type_name::<Length>()
+            ))
+        })
```
tiziano88, line 67: This seems like it is trying to second guess the lenght of the serialized message, but it is actually quite coupled with the serialization implementation. In fact, the serialization method seems to rely on this method, which seems backwards. My preference would be to inline this calculation in the serialization method, and not expose a len method at all; or, if we want to expose it, it should rely on the actual serialized format.
I don't feel too strongly about this, but at least some comments would help to make sure that this is well understood.

experimental/oak_baremetal_channel/src/frame.rs
```diff
@@ -0,0 +1,161 @@
...
+        .map(|frame_body| Frame {
+            flags: Flags::default(),
+            body: frame_body.to_vec(),
+        })
+        .collect();
+
+    frames.first_mut().map(|frame| {
+        frame.flags.set(Flags::START, true);
+        Some(())
+    });
```
tiziano88, line 153: Using map to perform side effects is not idiomatic. In fact, the compiler may just decide to optimize this away since the result is not actually used. Also, first and last are not expected to fail (i.e. they should never be None), so it seems fine to just unwrap / expect, instead of silently ignoring potential errors. Well, apart from empty data as input, but I would say that's in fact an error too (and maybe it could even be handled separately).

experimental/oak_baremetal_channel/src/lib.rs
tiziano88: +1

experimental/oak_baremetal_channel/src/tests.rs
```diff
@@ -14,177 +14,149 @@
...
+extern crate std;
+
+use crate::message::{Message, RequestMessage};
+use alloc::{collections::VecDeque, vec};
+
 use super::*;
 
 const BODY_LEN_MULTIPLICATOR: usize = 5;
-const MOCK_LARGE_MESSAGE_LEN: usize = MAX_FRAME_BODY_SIZE * BODY_LEN_MULTIPLICATOR;
+const MOCK_LARGE_PAYLOAD_LEN: usize = frame::MAX_BODY_SIZE * BODY_LEN_MULTIPLICATOR;
```
tiziano88, line 25: multiplier

tiziano88 (approved):

experimental/oak_baremetal_channel/src/frame.rs
jul-sh: > Using map to perform side effects is not idiomatic
haha I would be inclined to agree, I too feel that map should not be used for side effects. (though imo it's unlikely compilers would actually make that assumption)

Anyways, one part that does not agree with this sentiment is apparently "cargo clippy". Originally this used to be a "and_then", but it clippy flagged that as "error: using `Option.and_then(|x| Some(y))`, which is more succinctly expressed as `map(|x| y)`".

> Also, first and last are not expected to fail (i.e. they should never be None), so it seems fine to just unwrap / expect, instead of silently ignoring potential errors. Well, apart from empty data as input, but I would say that's in fact an error too (and maybe it could even be handled separately).

Yh, empty data would be the case where unwrap would lead to panic. I guess we can handle that separately, but originally decided not to, as the unwrap/expect here would be reliant behaviors a few lines up.

tiziano88: `and_then` should also not be used for side effects only. That would be done by an `if let` or a `match`.

experimental/oak_baremetal_channel/src/server.rs
```diff
@@ -30,38 +30,38 @@ where
...
+    pub fn write_response(&mut self, response: message::ResponseMessage) -> anyhow::Result<()> {
         self.inner.write_message(response)
     }
 }
 
-/// Construct a [`oak_idl::Request`] from a [`Message`].
-impl<'a> From<&'a Message> for oak_idl::Request<'a> {
-    fn from(frame: &'a Message) -> Self {
+/// Construct a [`oak_idl::Request`] from a [`message::RequestMessage`].
+impl<'a> From<&'a message::RequestMessage> for oak_idl::Request<'a> {
```
ipetr0v, line 43: I previously thought that `oak_idl::Request` was inside the `message::RequestMessage::data`, because it was a different layer in our protocol stack.

ipetr0v (approved):

experimental/oak_baremetal_channel/src/frame.rs
jul-sh: Added some references that link the structs to the spec. :)

I'm not sure if it makes sense to go too far into details and link to individual lines, I'd be worried it would cause a lot of needless overhead when updating the spec

PR #2981, opened 2022-06-20
Consume boot info when initializing memory

andrisaar:
As I found out in one of my other PRs, initializing the memory allocator can trash the contents of the boot metadata provided to us by the boot loader (as the metadata is stored _somewhere_ in valid memory, which may be overwritten at any point after we enable allocation).

Thus: let's make `memory::init_allocator()` take ownership of the reference, so that we know that it can't be used after we call the function.

### DIFF
--- a/experimental/oak_baremetal_kernel/src/memory.rs
+++ b/experimental/oak_baremetal_kernel/src/memory.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 //
 
-use crate::boot::{self, E820Entry};
+use crate::boot;
 use core::result::Result;
 use linked_list_allocator::LockedHeap;
 use log::info;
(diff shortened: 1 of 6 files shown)


experimental/oak_baremetal_kernel/src/memory.rs
```diff
@@ -26,7 +26,9 @@ static ALLOCATOR: LockedHeap = LockedHeap::empty();
 #[cfg(test)]
 static ALLOCATOR: LockedHeap = LockedHeap::empty();
 
-pub fn init_allocator<E: E820Entry>(e820_table: &[E]) -> Result<(), &str> {
+pub fn init_allocator<E: boot::E820Entry, B: boot::BootInfo<E>>(
```
tiziano88, line 29: Could you add a note about how / why you ended up with this version? I think it will be a good learning opportunity for future readers / maintainers.

tiziano88 (approved):

experimental/oak_baremetal_kernel/src/memory.rs
andrisaar: Good point, done.

PR #2987, opened 2022-06-20
Hand-roll Multiboot data structures instead of using bindgen.

andrisaar:
The main goal here was to make the data structures more Rust-y: for example, if otherwise we'd have a plain `u64` field that we know holds a pointer to a C string, then we can change the type of the field to `*const c_char`.

Same with flags: instead of relying on raw ints, we can use richer features provided by Rust that make things more legible.

The base code itself was copy-pasted from the bindgen-generated files.


experimental/oak_baremetal_app_qemu/src/hvm_start_info.rs
```diff
@@ -17,46 +17,65 @@
...
 use oak_baremetal_kernel::boot::{BootInfo, E820Entry, E820EntryType};
 
-include!(concat!(env!("OUT_DIR"), "/start_info.rs"));
+/// Magic value that needs to be present in the <StartInfo::magic> field.
+/// ("xEn3" with the 0x80 bit of the "E" set).
+#[cfg(not(feature = "multiboot"))]
+pub const BOOT_MAGIC: u32 = 0x336ec578;
 
-impl E820Entry for hvm_memmap_table_entry {
+#[repr(C)]
```
tiziano88, line 25: Could you mention here or somewhere else the command line that you used to generate the baseline of this file via bindgen? It will be useful to have that around if / when we need to update things again.

experimental/oak_baremetal_app_qemu/src/multiboot.rs
```diff
@@ -14,56 +14,140 @@
...
+
+        /// is the command-line defined?
+        const CMDLINE = 0x00000004;
+
+        /// are there modules to do something with?
+        const MODS = 0x00000008;
+
+        /// is there a symbol table loaded?
+        /// Mutually exclusive witht ELF_SHDR>.
+        const AOUT_SYMS = 0x00000010;
```
tiziano88, line 43: These may read better as binary constants, or bit-shifts (assuming that's a const operator, which I am not sure), but up to you. In particular, the bit-shift version seems easier to eyeball against the spec in https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#Boot-information-format .

tiziano88 (approved):

experimental/oak_baremetal_app_qemu/src/multiboot.rs
andrisaar: Those were straight from the C header file, but fair enough, yes, bit shifts make these nicer to read.

experimental/oak_baremetal_app_qemu/src/hvm_start_info.rs
andrisaar: There is no command line per se (as we were generating the files in `build.rs`), but I've added comments stating where the data structures are from.

PR #2992, opened 2022-06-21
Refactor virtio vsock channel creation in crosvm in preparation for simultaneous read/write

jul-sh:


experimental/oak_baremetal_loader/src/crosvm.rs
```diff
@@ -133,9 +85,25 @@ impl Vmm for Crosvm {
...
+                    Err(_error) => {
+                        sleep(Duration::from_millis(100)).await;
+                        continue;
+                    }
+                }
+            };
+            stream
+        })
+        .map_err(anyhow::Error::msg);
+        timeout(Duration::from_millis(1000), task).await?
```
jul-sh, line 107: For simultaneous Read/Write in the future, the plan is to update this function to return seperate Read and Write halves (both our virtio vsock & unix stream let us clone the handle). To do that though, the channel must be connected.

conradgrobler (approved):
Thanks

PR #2995, opened 2022-06-23
Share environment variables between the devcontainer and our docker scripts

### LINKED ISSUE #2717: Update devcontainer setup to use our docker scripts, instead of replicating their behavior

jul-sh:
Fixes #2717, means we can now use kvm in the dev container on system's where the docker group id is not 107.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -23,18 +23,14 @@
     // Crosvm wants access to syslog.
     "source=/dev/log,target=/dev/log,type=bind"
   ],
+  // Before creating the container, generate a .env file for the variables that
+  // should be set within the container.
+  "initializeCommand": "source ./scripts/docker_create_env_file",
   "runArgs": [
+    // Use the .env file generated prior to container creation.
+    "--env-file=.docker.env",
     // Required by crosvm.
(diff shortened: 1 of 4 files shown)


jul-sh:
Closing, this. It's not possible with the current devcontainer config, as we'd need to choose between setting the user to root (new VSCode terminals will start as root) or sticking with the docker user (which means we cannot run the scripts to set permissions).

Issue #2975, opened 2022-06-17
Replace protobufs and bincode serialisation with the flatbuffers-baesd IDL in the Oak Functions ABI


conradgrobler:
Manual serialisation in the ABI crate should also be repaced with the IDL generated code.

tiziano88:
This is about the Wasm ABI, right?

mariaschett:
+1. Do we want to replace, e.g., [StorageGetItemResponse](https://github.com/project-oak/oak/blob/65179ae5d477e9ee41564080fc411434e428f70e/oak_functions/abi/src/lib.rs#L60)?

conradgrobler:
Yes, the bincode serialisation is related to the ABI invoke requests and responses. So e.g. StorageGetItemResponse should be replaced.

I think the protobufs are not related to the Wasm ABI, so should probably be split into a separate crate as mentioned in #2495.

tiziano88:
Not planned. We are now using https://github.com/project-oak/oak/tree/main/micro_rpc for this.

PR #2999, opened 2022-06-24
Share a client between the linux and baremetal runtime

jul-sh:
Removes the bare-metal client and replaces it with the normal oak functions client.

Updates both runtimes to share status code encoding for responses


experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -254,7 +254,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     });
 
-    let server_future = server::server("127.0.0.1:8000".parse()?, tx);
+    let server_future = server::server("127.0.0.1:8080".parse()?, tx);
```
jul-sh, line 257: match the linux runtime default

oak_functions/abi/REQUEST_RESPONSE_ENCODING.MD
```diff
@@ -0,0 +1,90 @@
+When responding to user requests, the Trusted Runtime applies policies to
+prevent information egress.
+
+The policy enforcement layer defines an encoding format for requests and
+responses. This encoding takes place inside of the end-to-end encrypted channel
+established by remote attestation.
```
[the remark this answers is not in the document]
jul-sh: Brb, writing a small IDL using RLP encoding on the weekend 👀

experimental/oak_baremetal_loader/src/main.rs
andrisaar: This should really come from args. Hard-coding the address and port was a mistake to begin with.

oak_functions/abi/src/lib.rs
```diff
@@ -27,33 +27,110 @@ use core::mem::size_of;
 use serde::{Deserialize, Serialize};
 
 pub mod proto {
```
andrisaar, line 29: Is `proto::Response` still around? Do we need it? Or at the very least, does it need to be made public? (Do any of these need to be `pub`?)

oak_functions/abi/src/lib.rs
```diff
@@ -27,33 +27,110 @@ use core::mem::size_of;
...
+pub struct Request {
+    pub body: alloc::vec::Vec<u8>,
+}
+
+/// See REQUEST_RESPONSE_ENCODING.MD in the crate root.
+#[derive(Clone, PartialEq, Debug)]
+pub struct Response {
+    pub status: StatusCode,
+    pub body: alloc::vec::Vec<u8>,
+    pub length: u64,
```
andrisaar, line 46: Do we need this length field at all? In `Response::create`, below, you set it to `body.len()`, and I don't think it's ever set to anything other than that.

oak_functions/client/rust/src/main.rs
```diff
@@ -19,10 +19,13 @@
 
 use anyhow::Context;
 use clap::Parser;
-use oak_functions_abi::proto::Request;
+use oak_functions_abi::Request;
 use oak_functions_client::Client;
 use regex::Regex;
 
+const TWO_MIB: usize = 2097000;
```
andrisaar, line 26: That's a really weird constant. It's not 2\*1000^2 (2_000_000) nor 2\*1024^2 (2_097_152). What's going on here?

andrisaar (changes requested):
As a general comment, I think this PR is too large (36 files, ~1000 lines changed). Some of the work could definitely have been refactored to separate, smaller, PRs.

oak_functions/abi/REQUEST_RESPONSE_ENCODING.MD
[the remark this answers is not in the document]
tiziano88: I wonder if the low level IDL may just be a plain C struct? But it's interesting that the tool you used to generate the diagram is effectively doing that from a low level proprietary IDL, perhaps that could be used to generate objects in various languages too....

oak_functions/abi/src/lib.rs
jul-sh: Added a comment to clarify. This is outlined in `REQUEST_RESPONSE_ENCODING.MD` linked above the deck. In essence the body may includes padding to fit size requirements. The length then indicates which meaningful length (excluding padding).

jul-sh: Unfortunately yes, though this is a fair point. In essence, the ABI that the wasm module interacts still uses proto. We should be able to remove the invocation part at least though.

oak_functions/client/rust/src/main.rs
jul-sh: Pre-existing, but fixed, thanks.

oak_functions/abi/src/lib.rs
jul-sh: removed, thank you.

jul-sh:
@andrisaar 

Thanks for the review, fair points. In hindsight I agree that doing the larger refactor of the existing runtime ABI https://github.com/project-oak/oak/issues/2975 first, would've made for a clearer diff and smaller PRs. 

The issue here is that is in essence just replacing the existing proto generated structs with ones that use manual encoding. All the logic for what these structs contain, and how they are used is inherited from the proto generated code. It means it's not super obvious, and not 100% idiomatic. (Eg the the response containing both a body, and a length. Or the consumers of these structs [implementing additional traits](https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/server.rs) that are really core functionality).

That said, I think it makes sense to update it like this first (even if not pretty). It means we can share a client across runtimes right away, avoiding further fragmentation. Will follow this up with refactors if the larger ABI later however. :)

experimental/oak_baremetal_loader/src/main.rs
jul-sh: I agree, but also this PR is probably already big enough ;)

andrisaar (approved):
Thanks!

PR #3001, opened 2022-06-27
Add baremetal to provenance generation workflow

rbehjati:
* Adds a new xtask command for building the baremetal loader. Note: it builds the `debug` version, as `cargo build --release` fails with "permission denied" error.
* Adds `baremetal_build.toml`, a new static build config similar to `build.toml`, but for baremetal.
* Makes the provenance generation script more generic, to be able to use it for baremetal too.
* Adds a step to the `provenance.yaml` workflow to generate provenances for baremetal too.

Remaining: update reproducibility index to also upload baremetal binaries to Ent. 
(Note: The provenance generation and the reproducibility index workflows use different commands, and therefore result in binaries with different hashes. This should be fixed ASAP. The fix will require updating the build tool in transparent release to allow taking as input a list of options to pass to `docker run`.)


conradgrobler (commented):
I assume the intention of this PR is to build the bare-metal runtime that will run inside the VM, rather than the loader that launches the VM.

There are actually 2 such runtimes, one for running under CrosVM and one for running under QEMU.

https://github.com/project-oak/oak/tree/main/experimental/oak_baremetal_app_crosvm
and
https://github.com/project-oak/oak/tree/main/experimental/oak_baremetal_app_qemu

rbehjati:
Thanks @conradgrobler. I did not really look at the code, and the terminology was a bit confusing for me. 

> There are actually 2 such runtimes, one for running under CrosVM and one for running under QEMU.

Does this mean that `oak_baremetal_app_crosvm` and `oak_baremetal_app_qemu` are the binaries who's hashes will be included in the attestation quote? In that case I'll update the PR to generate the provenances for them instead.

conradgrobler:
> Does this mean that `oak_baremetal_app_crosvm` and `oak_baremetal_app_qemu` are the binaries who's hashes will be included in the attestation quote?

Yes, one of the two runtimes will be measured as part for the attestation reoprt.

For now generating a provenance for each one sounds good.

Note for future follow-up: the measurement will not be the exact hash of the binary, but a measurement of the memory of the VM after the binary has been loaded and laid out in memory ready for execution. This means that we will still need to implement a tool that will calculate the expected measurement from the binary and publish a mapping from the measurement to the binary hash in a public, discoverable location.

> the terminology was a bit confusing for me

Yes, we should find a more consistent set of terms so that it is always clear which parts we are talking about.

scripts/generate_provenance
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# Takes as input a TOML file, and builds a binary and its provenance from the
+# Git HEAD, using the build config in the TOML file. To run locally, pass the
```
tiziano88, line 4: AFAICT it does not build from HEAD, but from the currently checked out commit? Or maybe that's also called HEAD (perhaps local HEAD)?

scripts/generate_provenance
```diff
@@ -1,5 +1,12 @@
...
+# name of the TOML file as input. The generated provenance file will be named
+# `provenance.json` stored in the root of the repo. If an additional argument
+# is passed in, the `provenance.json`` file will be copied to
```
tiziano88, line 7:
```suggestion
# is passed in, the `provenance.json` file will be copied to
```

scripts/generate_provenance
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# Takes as input a TOML file, and builds a binary and its provenance from the
+# Git HEAD, using the build config in the TOML file. To run locally, pass the
+# name of the TOML file as input. The generated provenance file will be named
```
tiziano88, line 5: Maybe the provenance should be called similarly to the input file, so as not to overwrite itself each time?

scripts/generate_provenance
```diff
@@ -28,19 +35,22 @@ cp build.toml tmp/build.toml
...
 
 # cleanup
 rm -rf tmp
 
-# On GitHub we need to store the generated provenance files in a location that can easily be copied
-# to the provenance branch. We pass an optional argument to the script to trigger this behaviour.
-if [ $# -gt 0 ]; then
+# On GitHub we need to store the generated provenance files in a location that
+# can easily be copied to the provenance branch. We pass an optional argument
+# to the script to trigger this behaviour.
```
tiziano88, line 48: Not sure I understand this -- what is the optional argument?

tiziano88 (approved):

scripts/generate_provenance
rbehjati: Done.

scripts/generate_provenance
rbehjati: Fixed.

rbehjati (commented):
Thanks for the review :)

buildconfigs/oak_baremetal_app_crosvm.toml
```diff
@@ -0,0 +1,5 @@
+# This is the static build config that we use with the transparent release tool for building the provenance.
```
conradgrobler, line 1: This comment seems quite generic for a file that seems to be specific to the crosvm version of the baremetal binary.

buildconfigs/oak_baremetal_app_crosvm.toml
```diff
@@ -0,0 +1,5 @@
...
+# To be able to use it with the build tool from Transparent Release, additional information must be added.
+repo = "https://github.com/project-oak/oak"
+command = ["./scripts/xtask", "build-baremetal-variants"]
```
conradgrobler, line 4: Perhaps add a comment to clarify that we only generate proevenance for one variant even though this command builds both.

conradgrobler (approved):

buildconfigs/oak_baremetal_app_crosvm.toml
rbehjati: Thanks. Done. But I think it is probably better to change the command to take an input parameter to choose the variant. That would avoid building both variants twice when we add the build config for the `qemu` variant too. I can fix this in a future PR.

buildconfigs/oak_baremetal_app_crosvm.toml
rbehjati: Thanks. Done.

PR #3002, opened 2022-06-27
Get rid of ciborium_io dependency.

andrisaar:
We weren't using anything but the `Read` and `Write` traits from `ciborium_io` anyway, so let's get rid of that dependency.

I've added compatible `Read` and `Write` traits to `oak_baremetal_channel`. That seemed to be the best location for them.

I had to implement completely identical but separate traits in the `virtio` crate as well, as I didn't want to make `virtio` depend on the channel crate.

Apart from that nuance it's pretty straightforward... although I do have to say that all the bajillion conditional compilation statements in the kernel are _really_ getting out of hand. As a side effect, you never know what exactly will be built if you specify some combination of feature flags (or none at all, or --all-features). That needs to be cleaned up, but that's a separate undertaking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,12 +662,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ciborium-io"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "346de753af073cc87b52b2083a506b38ac176a44cfb05497b622e27be899b369"
-
-[[package]]
 name = "cipher"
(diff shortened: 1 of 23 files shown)


andrisaar:
> I think it makes sense to keep separate Read + Write traits (with- or without using ciborium_io for them). The reason is that in order to enable simultaneous reading/writing, we're likely to want to split the channel into two halves.

And I spent quite some time refactoring everything to have one trait, as we never use them separately :)

But fine, I'll go back to separate Read and Write traits.

PR #3009, opened 2022-06-29
Add buildconfig for oak_baremetal_app_qemu

rbehjati:
Follow up on #3001.
* Fixes the `scripts/generate_provenance` script
* Adds an option to `build_baremetal_variants` for specifying the variant
* Adds build config for `oak_baremetal_app_qemu`

### DIFF
--- a/xtask/src/internal.rs
+++ b/xtask/src/internal.rs
@@ -55,7 +55,7 @@ pub struct Opt {
 #[derive(Subcommand, Clone, Debug)]
 pub enum Command {
     RunVmTest,
-    BuildBaremetalVariants,
+    BuildBaremetalVariants(BuildBaremetalVariantsOpt),
     RunOakFunctionsExamples(RunOakExamplesOpt),
     BuildOakFunctionsExample(RunOakExamplesOpt),
     BuildOakFunctionsServerVariants(BuildServerOpt),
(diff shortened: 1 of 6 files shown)


xtask/src/internal.rs
```diff
@@ -263,6 +263,15 @@ pub struct RunCargoFuzz {
     pub args: Vec<String>,
 }
 
+#[derive(Parser, Clone, Debug)]
+pub struct BuildBaremetalVariantsOpt {
+    #[clap(
+        long,
+        help = "name of a specific baremetal variant (bios, or crosvm). If not specified, builds all variants."
```
conradgrobler, line 270: Based on the rest of the changes, I assume "bios" -> "qemu"

xtask/src/vm.rs
```diff
@@ -22,7 +22,7 @@ use strum_macros::{Display, EnumIter};
 use crate::internal::*;
 
 #[derive(Debug, Display, Clone, PartialEq, EnumIter)]
-enum Variant {
+pub enum Variant {
     Baremetal,
```
conradgrobler, line 26: pre-existing, but should probably be renamed to Qemu for consistency. I think this is left over from when QEMU was used for UEFI and baremetal.

conradgrobler (approved):
Thanks

xtask/src/internal.rs
rbehjati: Thanks. Fixed.

xtask/src/vm.rs
rbehjati: I agree.

rbehjati (commented):
Thanks for the review.
I changed Baremetal/Bios/bios to Qemu everywhere. Could you please take another look and let me know if looks correct? I could also do that replacement in a separate PR.

PR #3013, opened 2022-07-01
Update Remote Attestation Docs

jul-sh:
Making this PR as a follow up our discussion in the last on demand, where it seemed people were missing clarity around what the `AttestationReport` contains and when it is generated.

- It clarifies that (unlike what the docs previously stated) the Trusted Runtime does not request an Attestation Report for every handshake. Instead, the plan is to requests a report only once, and from there on out only uses a generated Signing key that was endorsed in the Attestation Report.
- Clarifies when the runtime is measured 
- It updates the diagram to use GitHub's native diagram renderer instead of the previous image. This means it's now readable on darkmode (😃) and by people using screen readers. Also makes it clear how to edit the diagram, and where the source of truth for the diagram is. You can see a rendered version of the diagram by visiting the PR's source branch [here](https://github.com/jul-sh/oak/blob/updated-diagram/docs/remote-attestation.md).
- Updated the diagram to include an "Untrusted Loader", to make clearer how this model translates to the new baremetal runtime. Worth noting that that does make the diagram a little bit specific to bare-metal, although deployments of the linux runtime are also likely to include untrusted intermediaries.

### DIFF
--- a/docs/remote-attestation.md
+++ b/docs/remote-attestation.md
@@ -62,17 +62,20 @@ Attestation from the start in order to create a new pair of Session keys.
 
 The workflow of the Remote Attestation protocol involves 4 interacting entities:
 
-1. Server
-   - Server application that processes requests from Clients
-   - Runs on a TEE Platform
 1. Client
    - Client application that connects to the Server
    - Client is provided with a TEE Provider's Root key
-1. TEE Platform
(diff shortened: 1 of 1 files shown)


conradgrobler (approved):
Great, thanks for clarifying the docs

PR #3014, opened 2022-07-03
Include all communication channels in one binary

andrisaar:
We support three communication channels -- virtio-console, virtio-vsock, and serial -- but right now choosing between them is with mutually incompatible features. That's not how features are supposed to work, for multiple reasons: features are supposed to be additive (otherwise `--all-features` turns into nonsense), and testing is difficult, and it's too easy to accidentally break code.

Now that we've added a facility to handle kernel command line arguments, let's make use of it to choose between the three comms channels, and build support for all three into the kernel binary.

Unfortunately this means that we won't know the size of the struct that implements said channel statically during build, so we have to `Box<>` things. This had other unfortunate repercussions in the channel code as well, as the easiest thing there was to change it to use a mutable borrow (as references always have a known size), but that made the code a bit more less readable. I guess the other option would be to bring the `Box` along for the ride, but I think semantically a mutable reference is more correct.

### DIFF
--- a/experimental/oak_baremetal_channel/src/server.rs
+++ b/experimental/oak_baremetal_channel/src/server.rs
@@ -14,17 +14,15 @@
 // limitations under the License.
 //
 
-use crate::{message, InvocationChannel, Read, Vec, Write};
+use crate::{message, Channel, InvocationChannel, Vec};
+use alloc::boxed::Box;
 
-pub struct ServerChannelHandle<'a, T: Read + Write + ?Sized> {
-    inner: InvocationChannel<'a, T>,
+pub struct ServerChannelHandle {
+    inner: InvocationChannel,
 }
 
-impl<'a, T> ServerChannelHandle<'a, T>
-where
-    T: Read + Write + ?Sized,
-{
-    pub fn new(socket: &'a mut T) -> Self {
+impl ServerChannelHandle {
+    pub fn new(socket: Box<dyn Channel>) -> Self {
         Self {
             inner: InvocationChannel::new(socket),
         }
--- a/experimental/oak_baremetal_channel/src/server.rs
+++ b/experimental/oak_baremetal_channel/src/server.rs
@@ -16,15 +16,15 @@
 
 use crate::{message, InvocationChannel, Read, Vec, Write};
 
-pub struct ServerChannelHandle<T: Read + Write> {
-    inner: InvocationChannel<T>,
+pub struct ServerChannelHandle<'a, T: Read + Write + ?Sized> {
+    inner: InvocationChannel<'a, T>,
 }
 
-impl<T> ServerChannelHandle<T>
+impl<'a, T> ServerChannelHandle<'a, T>
 where
-    T: Read + Write,
+    T: Read + Write + ?Sized,
 {
-    pub fn new(socket: T) -> Self {
+    pub fn new(socket: &'a mut T) -> Self {
         Self {
             inner: InvocationChannel::new(socket),
         }
--- a/experimental/oak_baremetal_kernel/src/args.rs
+++ b/experimental/oak_baremetal_kernel/src/args.rs
@@ -18,13 +18,30 @@ extern crate alloc;
 
 use alloc::collections::btree_map::BTreeMap;
 use arrayvec::ArrayString;
-use core::ffi::CStr;
-use lazy_static::lazy_static;
+use core::{ffi::CStr, lazy::Lazy};
 
-static mut ARGS_BUF: ArrayString<512> = ArrayString::new_const();
+static mut ARGS: ArrayString<512> = ArrayString::new_const();
 
-lazy_static! {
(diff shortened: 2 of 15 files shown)


experimental/oak_baremetal_kernel/src/args.rs
```diff
@@ -14,10 +14,18 @@
...
+use alloc::collections::btree_map::BTreeMap;
 use arrayvec::ArrayString;
 use core::ffi::CStr;
+use lazy_static::lazy_static;
+
+static mut ARGS_BUF: ArrayString<512> = ArrayString::new_const();
 
-static mut ARGS: ArrayString<512> = ArrayString::new_const();
+lazy_static! {
+    static ref ARGS: BTreeMap<&'static str, &'static str> = split_args(args());
```
conradgrobler, line 27: Is there something that guarantees that this will not be instantiated before the buffer has been populated?

conradgrobler (approved):
I am not 100% convinced of this approach, as it runs counter to the idea of minimising dependencies. I guess it is fine for now, but a better longer term solution would be to make the kernel agnostic to the channel being used and make it the responsibility of the app binary to provide the channel, seeing that the channel typically corresponds tho the platform where the binary is being run.

tiziano88:
I agree with @conradgrobler , but also with the sentiment that we should not be using feature flags to control this.

Ideally we would have a crate agnostic of the underlying channel, and then individual implementation of the channel parts; then the caller would depend on the core crate plus exactly one of the channel implementations, and would pass in the appropriate implementation in some way (ideally via a zero-cost generic type, but failing that just by constructing an instance of a struct that implements some trait, and passing that on to the core library).

Something like that is already happening for the remote attestation components:

- https://github.com/project-oak/oak/tree/main/remote_attestation
- https://github.com/project-oak/oak/tree/main/oak_remote_attestation_amd

experimental/oak_baremetal_channel/src/server.rs
```diff
@@ -16,15 +16,15 @@
...
 }
 
-impl<T> ServerChannelHandle<T>
+impl<'a, T> ServerChannelHandle<'a, T>
 where
-    T: Read + Write,
+    T: Read + Write + ?Sized,
 {
-    pub fn new(socket: T) -> Self {
+    pub fn new(socket: &'a mut T) -> Self {
```
jul-sh, line 27: I think conceptually it'd be nice for this struct to own the bytestream. We explicitly don't want any other reads/write of bytestream, since it risks making the struct out of sync wrt to the start & end points of messages.

andrisaar: Shouldn't the fact that we have a mutable borrow enforce that? You need a `&mut` to be able to call Read or Write, and Rust guarantees there can only be one mutable reference at a time. Therefore, you won't be able to pass one socket to multiple channels.

(But I may be wrong, please double-check my thinking here.)

jul-sh (approved):
+1 to @conradgrobler @tiziano88's point. An abstract kernel with a channel implementation in another crate seems conceptually nicest (but also most complex). 

If we do want to go this way with supporting multiple implementations chosen at runtime, it would be nice to still use feature flags to optionally strip out unused channel implementations.

jul-sh:
See also what the rust docs recommend in the case of non additive features: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

experimental/oak_baremetal_kernel/src/args.rs
andrisaar: Okay, I was able to get it to a state where it should be much safer to use, with less statics all around.

andrisaar:
It sounds like a better long-term strategy would be to extract all communication channel drivers into separate crates so that they could be included on demand.

I am not against that strategy -- it's probably a good idea, especially if we figure out a good driver model -- but that's a bigger change, and I'm going to say that the approach introduced in this PR is better than status quo, so unless anyone feels really strongly against it, I'm inclined to submit this as-is.

conradgrobler:
I agree that this is probably better than the status quo, but perhaps create an issue to track the future refactoring so we don't forget about it.

andrisaar:
Okay, PTAL. I think I've managed to strike a compromise with the comms channels:

- You can still enable and disable them with features as you wish. In fact, you can disable all three, if you want to.
- If you enable only one channel type, it will be used by default without the need to pass in command line args.
- If you enable multiple channel types, you can switch between them with a command line arg.
- The features are additive, so `--all-features` does the correct thing.

jul-sh:
Nice, that sounds reasonable to me. :)

jul-sh:
What does PTAL stand for?

conradgrobler (approved):
Nice, I think this is a good compromise for now.

I would still at some point want to move to the model where the binary can easily specify which boot protocol, channel, runtime, and kernel to use.

PR #3019, opened 2022-07-05
Revert "Bump linked_list_allocator from 0.9.1 to 0.10.0"

jul-sh:
Reverts project-oak/oak#3005. Post this commit the baremetal runtime panic when starting up, with "ERROR: PANIC: panicked at 'Freed node aliases existing hole! Bad free?', /home/docker/.cargo/registry/src/github.com-1ecc6299db9ec823/linked_list_allocator-0.10.0/src/hole.rs:439:17\n". Kokoro CI is likewise failing. 

Kokoro didn't run against the initial PR (it does not run against PRs from non project-oak users). However one can force it to run, by adding the tag `kokoro:force-run`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2287,9 +2287,9 @@ checksum = "7fb9b38af92608140b86b693604b9ffcc5824240a484d1ecd4795bacb2fe88f3"
 
 [[package]]
 name = "linked_list_allocator"
-version = "0.10.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


conradgrobler (approved):
Thanks

tiziano88:
Thanks for the quick rollback and for fixing the root issue @jul-sh !

tiziano88:
I'm still not sure exactly what went wrong with the Kokoro run though; what is the correct fix for applying the intended update? @conradgrobler @andrisaar

conradgrobler:
There are some bugs in that version of the linked list allocator around the allocation and freeing logic. We should wait for the next version with the bugs fixed before upgrading.

These were not caught by CI because PRs from dependabot did not trigger kokoro. #3020 should ensure that this does not happen in future.

PR #3020, opened 2022-07-05
Ensure Kokoro runs against dependabot PRs

jul-sh:
By default kokoro checks are only triggered for PRs from users that belong to the same org as the repo. Other PRs can trigger kokoro if they include a certain label. Adds this label for all dependabot PRs.

Related to https://github.com/project-oak/oak/pull/3019, hopefully this can prevent similar build breaks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,7 @@ updates:
       interval: 'daily'
     # We do not set an explicit reviewer, as that is already automatically assigned via CODEOWNERS.
     # reviewers
-    # Explicitly trigger kokoro checks. They do not automatically run for
-    # external users.
+    # Explicitly trigger kokoro checks. They do not automatically run for PRs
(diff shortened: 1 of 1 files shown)


conradgrobler (approved):
Once this is submitted we should test it checking that is ran for one of the rebased dependabot PRs

jul-sh:
Looks like it's working as expected :-) 
https://github.com/project-oak/oak/pull/3023

tiziano88:
Is it expected that kokoro removes that label once it runs?

tiziano88:
see https://github.com/project-oak/oak/pull/3023#event-6937422422

tiziano88:
Also, it sounds like we should make kokoro a required CI step for merging to main, should I do that? 
cc @conradgrobler @andrisaar

conradgrobler:
yes, I think it should be required

PR #3022, opened 2022-07-05
Update allocator

tiziano88:


conradgrobler:
I think this can be closed since #3218 updated it to a more recent version.

PR #3026, opened 2022-07-07
Move wasm path and http listen port to Config

### LINKED ISSUE #3008: Move wasm path and http listen port to Config
By moving the [Wasm file path and http listen port](https://github.com/project-oak/oak/blob/f105dd0c9d3bd9975d2c26ab41c4008f078da7d4/oak_functions/loader/src/lib.rs#L63-L73) from `Opt` to the `Config` we have all the configuration of the Oak Functions runtime in config.

mariaschett:
Fixes #3008 .

### DIFF
--- a/oak_functions/examples/echo/config.toml
+++ b/oak_functions/examples/echo/config.toml
@@ -1,3 +1,5 @@
+wasm_path = "bin/echo.wasm"
+
 [policy]
 constant_response_size_bytes = 90
 constant_processing_time = "20ms"
--- a/oak_functions/examples/echo/example.toml
+++ b/oak_functions/examples/echo/example.toml
@@ -4,7 +4,6 @@ name = "echo"
 
 [applications.rust]
 type = "OakFunctions"
-wasm_path = "bin/echo.wasm"
 target = { Cargo = { cargo_manifest = "oak_functions/examples/echo/module/Cargo.toml" } }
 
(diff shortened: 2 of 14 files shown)


jul-sh:
> Even more so, sometimes the actual location of the wasm file, [...], are not known until actually starting the job

Tbf, you could make pretty much the same argument for lookup data when it is loaded from a file. Yet the path of it is specified in the config. 

My two cents on this are that it's probably not worth it to worry too much about what goes where. :) 

The ideal solution would likely be something like https://lib.rs/crates/viperus (one schema, each key can be read from a file, or overwritten via a CLI flag)

jul-sh (approved):

mariaschett:
My motivation is to have everything that could change the runtime in one place: the config file.

Now, I agree, the trade-off is convience, but I don't see anything which would be impossible to do now. @tiziano88  Am I missing something?

tiziano88:
How would we use this on Cloud Run for instance, in which the port is in fact passed as $PORT? 

See https://cloud.google.com/run/docs/container-contract#port , though TBF the default is 8080 so I guess in most cases it will work to hardcode it, even if it's not great -- is that the plan?

mariaschett:
@andrisaar Would you say that seems reasonable?

andrisaar (approved):

PR #3029, opened 2022-07-07
Load Wasm module from the file system

jul-sh:
Previously the Wasm module was included in the untrusted loader binary


xtask/src/vm.rs
```diff
@@ -15,6 +15,7 @@
 //
 
 const CLIENT_PATH: &str = "./target/debug/oak_functions_client";
+const WASM_PATH: &str = "./experimental/oak_baremetal_loader/key_value_lookup.wasm";
```
mariaschett, line 18: Probably because I am missing a lot of context + pre-existing, but: why do we hard-code the "key_value_lookup.wasm"? Why not, e.g., "echo.wasm"?

Is the redirection to the `WASM_PATH` variable worth it or could we inline?

experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -58,8 +59,17 @@ struct Args {
     mode: Mode,
 
     /// Path to the kernel to execute.
-    #[clap(parse(from_os_str), validator = path_exists)]
+    #[clap(long, parse(from_os_str), validator = path_exists)]
     app: PathBuf,
+
+    /// Path to a Wasm file to be loaded into he trusted runtime and executed by it per invocation.
```
mariaschett, line 65: typo: he -> the trusted runtime. I think it must also export `alloc`, but maybe not? Does it help to link to [ABI](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md)?

xtask/src/vm.rs
mariaschett: If it helps: Here I use [WORKSPACE_ROOT](https://github.com/project-oak/oak/blob/228236fefa5c8279b42138d2452b7fd87f7ff9b1/oak_functions/sdk/oak_functions/tests/integration_test.rs#L32)

jul-sh: Mostly we do just that because this way we can test that the lookup logic works. :) I guess in the future we'd want to also have multiple examples for the bare-metal runtime, which test different functionality.

jul-sh: Re [WORKSPACE_ROOT](https://github.com/project-oak/oak/blob/228236fefa5c8279b42138d2452b7fd87f7ff9b1/oak_functions/sdk/oak_functions/tests/integration_test.rs#L32) I like the clarity of it, but realistically we'd have to then update all relative paths in this file, not sure if it's worth it. :)

jul-sh:
@mariaschett thanks for the review!

Issue #3007, opened 2022-06-28
xtask run-cargo-tests fails with `qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.07H:EDX.spec-ctrl`


mariaschett:
For me, a re-run of the failed Action helped. This is a bug to track how often this happens:

```
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║    Compiling oak_baremetal_runtime v0.1.0 (/workspace/experimental/oak_baremetal_runtime)
333
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║    Compiling oak_baremetal_kernel v0.1.0 (/workspace/experimental/oak_baremetal_kernel)
334
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║     Finished test [unoptimized + debuginfo] target(s) in 1m 04s
335
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║      Running unittests src/main.rs (target/target/debug/deps/oak_baremetal_app_qemu-9023609f94ef2d1f)
336
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║ qemu-system-x86_64: warning: host doesn't support requested feature: CPUID.07H:EDX.spec-ctrl [bit 26]
337
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║ qemu-system-x86_64: Host doesn't support requested features
338
[14:11:47 ✓:0,✗:0,⠇:2] │││ ║ error: test failed, to rerun pass '--bin oak_baremetal_app_qemu'
339
[14:11:47 ✓:0,✗:0,⠇:2] │││ ╚════════════════════════
```

rbehjati:
Is the following the command that fails?
```
$ cargo test --all-features --manifest-path=Cargo.toml (in directory ./experimental/oak_baremetal_app_qemu
```

PR #3034, opened 2022-07-11
Don't run tests on baremetal binaries.

andrisaar:
Running tests in these crates is difficult as they're built against the baremetal target, and thus require special handling -- such as running the tests inside an emulator.

However, we've moved all of real code into the `oak_baremetal_kernel` crate, where it can be tested normally; these crates effectively exist only to force their dependencies to be built against bare metal, and contain no meaningful code (nor tests).

In the future, once we standardize on one boot protocol, I foresee one of these crates going away completely and the other becoming even more vestigial.

Thus, to make our lives simpler, let's disable tests for the baremetal crates.


conradgrobler (approved):
Nice, thanks

andrisaar:
I think by accident this also fixes #3007.

PR #3038, opened 2022-07-12
Refactor bmrng channel in the untrusted loader to increase clarity

jul-sh:
Makes it so that the requests and responses sent over this channel consistently represent flatbuffer encoded messages. Previously this was the case only for requests but not responses. 

Improves comments/naming for clarity. 

This will make the upcoming code change (sending lookup data updates over the same channel) easier to review.

### DIFF
--- a/experimental/oak_baremetal_loader/src/main.rs
+++ b/experimental/oak_baremetal_loader/src/main.rs
@@ -176,11 +176,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let comms = vmm.create_comms_channel().await?;
 
-    // Use a bmrng channel for serial communication. The good thing about spawning a separate
-    // task for the serial communication is that it serializes (no pun intended) the communication,
-    // as right now we don't have any mechanisms to track multiple requests in flight.
-    let (tx, mut rx) = bmrng::unbounded_channel::<Vec<u8>, Result<Vec<u8>, oak_idl::Status>>();
+    // A message based communication channel that permits other parts of the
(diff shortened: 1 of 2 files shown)


experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -176,11 +176,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let (tx, mut rx) = bmrng::unbounded_channel::<Vec<u8>, Result<Vec<u8>, oak_idl::Status>>();
-
+    // A message based communication channel that permits other parts of the
+    // untrusted loader to send requests to the tasks that handles communicating
+    // with the runtime and receive responses. Requests and (ok) responses
+    // sent over this channel represent flatbuffer encoded messages. This is
+    // done as it is not possible to send the flatbuffer Message structs
+    // themselves over the channel.
+    let (request_dispatcher, mut request_receiver) =
+        bmrng::unbounded_channel::<Vec<u8>, Result<Vec<u8>, oak_idl::Status>>();
```
jul-sh, line 186: (Pre-existing) the conversion of flatbuffer Message structs to and from owned Vectors is arguably not the most ergonomic or efficient (reallocations). 

It will also become slightly less ergonmic in the future, when start sending lookup data messages over this channel (we'll need an enum to differentiate between them & user-requests). 

Imo not worth it to sweat this too much right now however, as this will change if a) change away from flatbuffer or b) away from the channel towards a RwLock with our own queue.

PR #3039, opened 2022-07-12
Untrusted Loader: Load lookup data from the file system

jul-sh:


experimental/oak_baremetal_loader/src/lookup.rs
```diff
@@ -0,0 +1,53 @@
...
+        .finish(message)
+        .expect("errored when creating lookup data update message")
+}
+
+pub fn load_lookup_data(
+    file_path: &std::path::PathBuf,
+) -> anyhow::Result<HashMap<Vec<u8>, Vec<u8>>> {
+    let bytes = fs::read(&file_path)
+        .unwrap_or_else(|| &format!("Couldn't read Wasm file {}", &file_path.display()));
+    oak_functions_loader::lookup_data::parse_lookup_entries(bytes.as_slice())
```
jul-sh, line 52: Re-uses the existing logic/format for encoding lookup data to a file.

experimental/oak_baremetal_loader/src/lookup.rs
```diff
@@ -0,0 +1,53 @@
...
+            schema::LookupDataEntry::create(
+                &mut builder,
+                &schema::LookupDataEntryArgs {
+                    key: Some(key),
+                    value: Some(value),
+                },
+            )
+        })
+        .collect();
+    let items = builder.create_vector(&entries);
```
jul-sh, line 39: There's a bit of reallocation happening here, which is not super efficient. The file is loaded in memory as a byte array, then parsed and reallocated into a hashmap of entries, before finally being reallocated again when entered into the flatbuffer struct. 

Ideally we would read the file as a stream, and on the fly directly assemble the flatbuffer message. That doesn't seem possible atm though, since the flatbuffer builder's rust API requires  knowledge of the number of elements when creating a new vector. 

Anyways, I'm not sure if it's a priority to fully optimize this at this stage, given that this logic is basically just used for testing at this stage.

experimental/oak_baremetal_loader/src/lookup.rs
```diff
@@ -0,0 +1,53 @@
...
+    let message =
+        schema::LookupData::create(&mut builder, &schema::LookupDataArgs { items: Some(items) });
+    builder
+        .finish(message)
+        .expect("errored when creating lookup data update message")
```
rbehjati, line 44: Could the function return Result instead? Perhaps returning an `oak_idl::Status` in case of error?

experimental/oak_baremetal_loader/src/lookup.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+pub fn load_lookup_data(
+    file_path: &std::path::PathBuf,
+) -> anyhow::Result<HashMap<Vec<u8>, Vec<u8>>> {
+    let bytes = fs::read(&file_path)
+        .unwrap_or_else(|_| panic!("Couldn't read Wasm file {}", &file_path.display()));
```
rbehjati, line 51:
```suggestion
        .unwrap_or_else(|_| panic!("Couldn't read lookup data file {}", &file_path.display()));
```
I think! :)

rbehjati: Why does not this line return a Result, particularly given that the function returns Result? If the panic is intentional might be worth adding a comment to describe why.

experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -188,29 +197,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            let items = builder.create_vector(&[entry]);
-            let message = schema::LookupData::create(
-                &mut builder,
-                &schema::LookupDataArgs { items: Some(items) },
-            );
-            builder
-                .finish(message)
-                .expect("errored when creating lookup data update message")
-        };
+        let lookup_data = lookup::load_lookup_data(&cli.lookup_data).unwrap();
```
rbehjati, line 200:
```suggestion
        let lookup_data = lookup::load_lookup_data(&cli.lookup_data)?;
```
If possible.

rbehjati (approved):
Nice!

experimental/oak_baremetal_loader/src/lookup.rs
conradgrobler: I think this is fine for now. Seeing that we are not fully committed to using flatbuffers as an IDL in the long run, optimising at this point seems like potentially wasted effort.

conradgrobler (approved):

experimental/oak_baremetal_loader/src/lookup.rs
jul-sh: Yeah you're right, thanks for catching. And this should indeed return an error instead of throwing.

experimental/oak_baremetal_loader/src/lookup.rs
jul-sh: Thanks, updated for it to return a result. The status code is a code idea but probably not quite right here, since this code is related to initiating a call to an IDL method, instead of handling one.

PR #3040, opened 2022-07-13
Use simple IO device as a comms channel

conradgrobler:


experimental/oak_baremetal_app_qemu/Cargo.toml
```diff
@@ -6,14 +6,15 @@ edition = "2021"
...
 pvh = ["rust-hypervisor-firmware-boot/pvh"]
 multiboot = []
-serial_channel = ["oak_baremetal_kernel/serial_channel"]
+simple_io_channel = ["oak_baremetal_kernel/simple_io_channel"]
+virtio_console_channel = ["oak_baremetal_kernel/virtio_console_channel"]
 
 [dependencies]
 bitflags = "*"
-oak_baremetal_kernel = { path = "../../experimental/oak_baremetal_kernel" }
+oak_baremetal_kernel = { path = "../../experimental/oak_baremetal_kernel", default-features = false }
```
andrisaar, line 17: Why disable the default features, if you're just enabling `virtio_console_channel` again? (Or maybe it'd make sense to remove the `virtio_console_channel` from default features in the kernel crate?)

conradgrobler: Removed the default feature from the kernel crate.

experimental/oak_baremetal_kernel/src/lib.rs
```diff
@@ -41,8 +41,8 @@ mod interrupts;
 mod libm;
 mod logging;
 mod memory;
-#[cfg(feature = "serial_channel")]
-mod serial;
```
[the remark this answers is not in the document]
conradgrobler: Added it back for now, even though I don't think we should ever use it again. It is unusable with the current runtime as it takes 7 minutes to upload the simplest wasm module.

conradgrobler (commented):
Thanks for the review

andrisaar (approved):

experimental/oak_baremetal_app_qemu/Cargo.toml
conradgrobler: I had to add a default channel back. Cargo udeps does not check with --al-features, and the kernel does not build if we do not specify at least one channel. For now I switched to using the vsock channel as the default.

PR #3048, opened 2022-07-15
Add more GHCB MSR protocol support

conradgrobler:
This is a follow-up to #3045. It adds support for more of the GHCB MSR protocol, specifically:

- Requesting the hypervisor's preferred location for the GHCB
- Registering the location of the GHCB

Additional protocol features will be implemented in subsequent PRs.

### DIFF
--- a/experimental/sev_guest/src/msr.rs
+++ b/experimental/sev_guest/src/msr.rs
@@ -31,7 +31,8 @@
 //!
 //! Note: This implementation does not include the AP Reset Hold operations seeing that it is not
 //! the preferred mechanism for starting a new application processor (AP) under AMD SEV-SNP. The
-//! SEV-SNP AP creation NAE should be used instead.
+//! SEV-SNP AP creation NAE should be used instead. See section 4.3.2 of
+//! <https://developer.amd.com/wp-content/resources/56421.pdf>.
 
 use crate::instructions::vmgexit;
 use anyhow::{bail, Result};
@@ -100,7 +101,7 @@ impl TryFrom<u64> for SevInfoResponse {
     fn try_from(msr_value: u64) -> Result<Self> {
         const SEV_INFO_RESPONSE_INFO: u64 = 0x001;
         if msr_value & GHCB_INFO_MASK != SEV_INFO_RESPONSE_INFO {
-            bail!("Value is not a valid SEV information response.");
+            bail!("Value is not a valid SEV information response");
         }
         let max_protocol_version = (msr_value >> 48) as u16;
         let min_protocol_version = (msr_value >> 32) as u16;
@@ -172,7 +173,7 @@ impl TryFrom<u64> for CpuidResponse {
         const SEV_INFO_RESPONSE_INFO: u64 = 0x005;
         const RESERVED_MASK: u64 = 0x3FFFF000;
         if msr_value & GHCB_INFO_MASK != SEV_INFO_RESPONSE_INFO || msr_value & RESERVED_MASK != 0 {
-            bail!("Value is not a valid CPUID response.");
+            bail!("Value is not a valid CPUID response");
         }
         let value = (msr_value >> 32) as u32;
         const REGISTER_MASK: u64 = 0xC0000000;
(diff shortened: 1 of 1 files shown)


experimental/sev_guest/src/msr.rs
```diff
@@ -174,6 +190,102 @@ pub fn get_cpuid(request: CpuidRequest) -> Result<CpuidResponse> {
     read_msr().try_into()
 }
 
+/// A request for the hypervisor's preferred location for the GHCB page.
+pub struct PreferredGhcbGpaRequest;
+
+impl From<PreferredGhcbGpaRequest> for u64 {
+    fn from(_request: PreferredGhcbGpaRequest) -> Self {
```
mariaschett, line 197: Why do we ignore the `_request` here and hard-coded return `0x010`?

experimental/sev_guest/src/msr.rs
```diff
@@ -174,6 +190,102 @@ pub fn get_cpuid(request: CpuidRequest) -> Result<CpuidResponse> {
...
+        if ghcb_gpa & GHCB_INFO_MASK != 0 {
+            bail!("GHCB must be 4KiB-aligned");
+        }
+        Ok(Self { ghcb_gpa })
+    }
+}
+
+impl From<RegisterGhcbGpaRequest> for u64 {
+    fn from(register_ghcb_gpa_request: RegisterGhcbGpaRequest) -> Self {
+        let mut result = 0x012;
```
mariaschett, line 250: Would that be `PREFERED_GHCB_GPA_REQUEST` from GHCB page 14/Table 2?

experimental/sev_guest/src/msr.rs
```diff
@@ -174,6 +190,102 @@ pub fn get_cpuid(request: CpuidRequest) -> Result<CpuidResponse> {
...
+/// The response containing the result of the GHCB registration.
+pub struct RegisterGhcbGpaResponse {
+    /// The registered guest-physical address for the GHCB.
+    ghcb_gpa: u64,
+}
+
+impl TryFrom<u64> for RegisterGhcbGpaResponse {
+    type Error = anyhow::Error;
+    fn try_from(msr_value: u64) -> Result<Self> {
+        const REGISTER_GHCB_GPA_RESPONSE_INFO: u64 = 0x013;
```
mariaschett, line 265: Why do we have a `_INFO` here and not simply `REGISTER_GHCB_GPA_RESPONSE`?

experimental/sev_guest/src/msr.rs
```diff
@@ -174,6 +190,102 @@ pub fn get_cpuid(request: CpuidRequest) -> Result<CpuidResponse> {
     read_msr().try_into()
 }
 
+/// A request for the hypervisor's preferred location for the GHCB page.
+pub struct PreferredGhcbGpaRequest;
+
+impl From<PreferredGhcbGpaRequest> for u64 {
+    fn from(_request: PreferredGhcbGpaRequest) -> Self {
+        0x010
```
mariaschett, line 198: Is this `PREFERRED_GHCB_GPA_REQUEST` from page 14/Table 2?

experimental/sev_guest/src/msr.rs
```diff
@@ -174,6 +190,102 @@ pub fn get_cpuid(request: CpuidRequest) -> Result<CpuidResponse> {
...
+    }
+}
+
+/// The response containing the preferred location of the GHCB.
+pub struct PreferredGhcbGpaResponse {
+    /// The preferred guest-physical address for the GHCB.
+    pub ghcb_gpa: usize,
```
mariaschett, line 205: Why do we use usize here?

mariaschett (approved):
Thank you! 

I am missing quite a lot of context for a good review, but I added a few comments, where I was not sure.

I did not check all the bit-operations against the standard. Would you want me to?

Also: would it make sense to add unit tests?

experimental/sev_guest/src/msr.rs
conradgrobler: There is no additional data as part of the request, just the request code. This could have been a constant rather than a struct that implements `From`, but I thought this approach keeps the way the requests are used consistent.

experimental/sev_guest/src/msr.rs
conradgrobler: This is the GHCBInfo value for "Register GHCB GPA Request" from that table.

conradgrobler (commented):
Thanks.

conradgrobler:
> I did not check all the bit-operations against the standard. Would you want me to?
>
> Also: would it make sense to add unit tests?

I think using unit tests to double-check the bitwise operation might be a good idea. I will add it some in a follow-up PR once all the protocol operations are implemented.

PR #3054, opened 2022-07-18
Add final GHBC MSR protocol support

conradgrobler:
This is a follow-up to #3045 and #3048. It adds support for the remaining operations of the GHCB MSR protocol, specifically:

- Requesting a change to the state of a memory page (private or shared)
- Requesting the hypervisor's supported features
- REquesting termination of the guest VM

Unit tests for the bit manipulation will be added in a follow-up PR.


mariaschett (approved):
Looks good to me! 

Again, I have not checked every bit manipulation in detail, but happy to do so, if you'd think it would be good!

conradgrobler (commented):
Thanks

PR #3056, opened 2022-07-18
Run dependabot against baremetal crates as well

andrisaar:
The downside is that we will get far more dependabot PRs for now.

In the future this will get better:
  - we expect one of the baremetal crates to go away (and thus one less `Cargo.lock` to maintain)
  - hopefully Rust will figure out how to include baremetal crates in the main workspace soon.


ipetr0v (approved):

jul-sh (approved):
thanks

PR #3061, opened 2022-07-19
Add GHCB MSR protocol unit tests

conradgrobler:
This PR adds unit tests for the conversion of the MSR protocol request objects to u64 values and u64 values to response objects.

### DIFF
--- a/experimental/sev_guest/src/msr.rs
+++ b/experimental/sev_guest/src/msr.rs
@@ -133,7 +133,7 @@ pub fn get_sev_info() -> Result<SevInfoResponse> {
 }
 
 /// The register of interest from the result of executing CPUID.
-#[derive(Debug, FromRepr)]
+#[derive(Debug, FromRepr, Clone, Copy, PartialEq, Eq)]
 #[repr(u8)]
 pub enum CpuidRegister {
     Eax = 0,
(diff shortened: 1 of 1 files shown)


experimental/sev_guest/src/msr.rs
```diff
@@ -462,3 +462,174 @@ pub fn read_msr() -> u64 {
...
+            (valid_page as u64) | ((assignment as u64) << 52) | 0x14,
+            msr_value
+        );
+    }
+
+    #[test]
+    fn test_snp_page_state_change_response() {
+        let error_code = 1u64;
+        let invalid = 0x15u64 | (error_code << 31);
+        let invalid2 = 0x16u64 | (error_code << 32);
```
rbehjati, line 595: Is it possible to give it a better name?

rbehjati (approved):
Nice. Thanks. I feel like generally some documentation would be nice. In particular, it is not clear to me where the different values come from, particularly for `page`. I assume the `page` values are not any arbitrary value. It also feels like the tests are mostly validating the conversion from numbers to various Request and Response types. If that is the intention, perhaps this could be described as a comment in `mod tests`.  

(Or maybe everything is clear to someone who is more familiar with SEV :))

experimental/sev_guest/src/msr.rs
conradgrobler: Done.

conradgrobler (commented):
Thanks for the review.

Yes, the tests only check the conversions. I added a comment to the `tests` module to clarify.

The values (e.g. page addresses) in the tests are arbitrary, they just need to fall in certain ranges to be considered valid for the protocol. Any testing of the actual protocol would require running inside a VM under AMD SEV-SNP.

PR #3070, opened 2022-07-20
Oak IDL: Add ability to generate Async Clients

jul-sh:
Now ready, after the preparation PRs were merged:

- [x] #3068
- [x] #3069

---- 

Support for async clients will unlock two things in the untrusted loader:
- sending requests / receiving responses out of sequence
- avoid the need for duplicate structs/enum for sending IDL messages through async code


conradgrobler (approved):
Nice, thanks.

PR #3071, opened 2022-07-20
Rename the oak IDL Message struct to something more descriptive

jul-sh:
Helps delineate it from Messages on the communication layer. The new name is `OwnedFlatbuffer`, as being an owned flatbuffer is in essence what this struct does. It's not that specific to our IDL logic. :)

### DIFF
--- a/experimental/oak_baremetal_loader/src/lookup.rs
+++ b/experimental/oak_baremetal_loader/src/lookup.rs
@@ -21,8 +21,8 @@ use std::fs;
 
 pub fn encode_lookup_data<'a>(
     mut data: HashMap<Vec<u8>, Vec<u8>>,
-) -> anyhow::Result<oak_idl::utils::Message<schema::LookupData<'a>>> {
-    let mut builder = oak_idl::utils::MessageBuilder::default();
+) -> anyhow::Result<oak_idl::utils::OwnedFlatbuffer<schema::LookupData<'a>>> {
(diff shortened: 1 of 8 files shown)


ipetr0v:
I think a good name could be `Invocation`.
Since we already have an `invocation_id` in the protocol (which refers to the upper IDL layer).

ipetr0v:
Also `OwnedFlatbuffer` doesn't provide information about the fact, that it's intended for Oak.

conradgrobler:
> I think a good name could be `Invocation`. Since we already have an `invocation_id` in the protocol (which refers to the upper IDL layer).

I think `Invocation` would be a good name for the `Message` trait defined in https://github.com/project-oak/oak/blob/main/experimental/oak_baremetal_channel/src/message.rs#L45-L49, rather than this helper struct. I think `OwnedFlatbuffer` is an accurate description of the intended use of this struct.

tiziano88:
This layer should not know about invocations, I think `OwnedFlatbuffer` is a good descriptive name for this.

tiziano88 (approved):

conradgrobler:
> Who owns the Flatbuffer since it's an `OwnedFlatbuffer`?

The `OwnedFlatbuffer` struct owns the buffer referenced by the flatbuffer object. Normally flatbuffer objects only have a reference to the underlying buffer that stores their data, but do not own the buffer. The OwnedFlatbuffer struct is only a helper so that we could move the buffer and its associated flatbuffer object around together.

PR #3074, opened 2022-07-21
Use Async IDL Client in the Untrusted Loader

jul-sh:
Updates the untrusted loader to use an async IDL client. Instead of there being a central client that is being invoked over a bmrng channel by different parts of the runtime, there will now be multiple async clients that issue invocations over the bmrng channel.

This means that we'll get the code to differentiate various messages sent over the channel for free, provided by the oak IDL. This will come in handy when implementing periodic lookup-data refreshing.

In the future async clients will also allow us to more easily support non-sequential invocations.

### DIFF
--- a/experimental/oak_baremetal_channel/src/lib.rs
+++ b/experimental/oak_baremetal_channel/src/lib.rs
@@ -14,11 +14,13 @@
 // limitations under the License.
 //
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
 #![feature(never_type)]
 #![allow(rustdoc::private_intra_doc_links)]
 
+#[cfg(feature = "client")]
 pub mod client;
+
 pub mod server;
 
 pub mod schema {
@@ -27,7 +29,11 @@ pub mod schema {
 
     include!(concat!(env!("OUT_DIR"), "/schema_generated.rs"));
     include!(concat!(env!("OUT_DIR"), "/schema_services_servers.rs"));
-    include!(concat!(env!("OUT_DIR"), "/schema_services_clients.rs"));
+    #[cfg(feature = "client")]
+    include!(concat!(
+        env!("OUT_DIR"),
+        "/schema_services_async_clients.rs"
+    ));
 }
 
 mod frame;
--- a/experimental/oak_baremetal_loader/src/main.rs
+++ b/experimental/oak_baremetal_loader/src/main.rs
@@ -113,21 +113,46 @@ impl oak_baremetal_communication_channel::Read for CommsChannel {
     }
 }
 
-pub struct ClientHandler {
+/// Client implementation that communicates with the communication
+/// task via a bmrng channel. Used by various parts of the loader to communicate
+/// with the trusted runtime.
+pub struct Client {
+    request_dispatcher: bmrng::unbounded::UnboundedRequestSender<
+        oak_idl::Request,
+        Result<Vec<u8>, oak_idl::Status>,
+    >,
+}
+
+#[async_trait::async_trait]
+impl oak_idl::AsyncHandler for Client {
+    async fn invoke(&mut self, request: oak_idl::Request) -> Result<Vec<u8>, oak_idl::Status> {
+        self.request_dispatcher
+            .send_receive(request)
+            .await
+            .map_err(|err| {
+                oak_idl::Status::new_with_message(
+                    oak_idl::StatusCode::Internal,
+                    format!("failed when invoking the request_dispatcher: {}", err),
+                )
+            })?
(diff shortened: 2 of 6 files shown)


experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -186,60 +211,64 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    // untrusted loader to send requests to the tasks that handles communicating
-    // with the runtime and receive responses. Requests and (ok) responses
-    // sent over this channel represent flatbuffer encoded messages. This is
-    // done as it is not possible to send the flatbuffer Message structs
-    // themselves over the channel.
+    // untrusted loader to send requests to the task that handles communicating
+    // with the runtime and receive responses.
     let (request_dispatcher, mut request_receiver) =
-        bmrng::unbounded_channel::<Vec<u8>, Result<Vec<u8>, oak_idl::Status>>();
+        bmrng::unbounded_channel::<oak_idl::Request, Result<Vec<u8>, oak_idl::Status>>();
```
jul-sh, line 217: Note that we could alternatively also replace this channel with a central client, that is shared to different parts of the untrusted loader via mutex. The result would be pretty much the same. 

The bmrng channel (and more generally sharing state via messages) has kind of grown on me though, so I'm happy to keep it.

experimental/oak_baremetal_channel/src/lib.rs
```diff
@@ -14,11 +14,13 @@
...
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
 #![feature(never_type)]
 #![allow(rustdoc::private_intra_doc_links)]
 
+#[cfg(feature = "client")]
 pub mod client;
+
 pub mod server;
```
andrisaar, line 24: You introduce the `server` feature in `Cargo.toml`, above; should this be gated on that feature being enabled, much like the client is?

experimental/oak_baremetal_loader/src/main.rs
```diff
@@ -186,60 +211,64 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            // At the moment requests are sent sequentially, and in FIFO order. The next request
+            // is sent only once a response to the previous message has been implemented.
+            // TODO(#2848): Implement message prioritization, and non sequential invocations.
+            let response = communication_channel.invoke(request);
             response_dispatcher.respond(response).unwrap();
         }
     });
 
+    let mut client = schema::TrustedRuntimeAsyncClient::new(Client {
+        request_dispatcher: request_dispatcher.clone(),
```
andrisaar, line 233: Okay, this may be my C++ background and allergies about constantly copying data around, but just cloning things makes me suspicious. Why can't it be a mutable reference or take ownership?

andrisaar (changes requested):
It looks cleaner, but (a) I may miss some context around IDL here, and (b) honestly, I think we could simplify all the code somewhat in the future. The layers of structs are starting to give me vertigo.

experimental/oak_baremetal_channel/src/lib.rs
jul-sh: Ah thanks for catching, overlooked the server feature. Removed that feature, imo we don't need it right now. The reason the client code needs feature gating is that it introduces a dependency on the standard library. The server code if not used can just be compiled out.

experimental/oak_baremetal_loader/src/main.rs
jul-sh: At the current stage of the code it could be a mutable reference (if we'd modify the client struct to take a lifetime param). However, sooner or later we'll need multiple senders, and might well want to use this client at the same time as others.

Cloning suspicions in general are imo not misplaced, but it's fine here. The ability to clone senders to create multiple of them is pretty much the point of mspc tokio channels, see: https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html (bmrng is a thin abstraction top of that)

jul-sh:
@andrisaar and yes, agreed, that we should further clean up the untrusted loader code at some point, albeit in a separate PR.

experimental/oak_baremetal_loader/src/main.rs
jul-sh: Under the hood the sender half uses atomic reference counting, so cloning should be extremely cheap in practice. 

https://github.com/tokio-rs/tokio/blob/4daeea8cad1ce8e67946bc0e17d499ab304b5ca2/tokio/src/sync/mpsc/chan.rs

andrisaar (approved):

PR #3077, opened 2022-07-22
Make Trusted Shuffler pass on gRPC requests

### LINKED ISSUE #3051: Make Trusted Shuffler work with gRPC requests
First, we extend our [experimental client](https://github.com/project-oak/oak/tree/main/experimental/trusted_shuffler/client) and [experimental backend](https://github.com/project-oak/oak/tree/main/experimental/trusted_shuffler/backend) to work with gRPC.

Next we try Trusted Shuffler to shuffle and pass on the gRPC requests using  HTTP2.

mariaschett:
Fixes #3051 .


experimental/trusted_shuffler/scripts/run_trusted_shuffler
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+RUST_LOG=info cargo run --manifest-path=experimental/trusted_shuffler/server/Cargo.toml -- \
+--k=1 --listen-address="[::]:8080" --backend-url="http://localhost:8888" &
+   
+# Convenient to know the PID to kill when the Trusted Shuffler crashes.
+echo "Trusted Shuffler PID: "
+echo $!
```
ipetr0v, line 8: nit: Looks like github wants you to add a blank line in the end of the file =)

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -63,30 +74,23 @@ impl hyper::service::Service<Request<Body>> for Service {
...
-                (&Method::POST, "/request") => {
-                    let body = hyper::body::to_bytes(request.into_body())
-                        .await
-                        .expect("Couldn't read request body");
-
+            match request.method() {
+                // TODO(mschett): Check if http still works.
+                // (&Method::POST, "/experimental.trusted_shuffler.echo.Echo/Echo")
+                // (&Method::POST, "/request")
+                &Method::POST => {
```
ipetr0v, line 84: I think we should accept any requests, not just `POST` ones

mariaschett: Done!

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+
+        // We want to keep the path of the orginal request from the client.
+        let path = request.uri().path();
+        // And extend the URL to the backend.
+        let new_uri = format!("{}{}", self.backend_url, path)
+            .parse::<Uri>()
+            .expect("Couldn't parse URI for backend.");
+        let uri = request.uri_mut();
+        *uri = new_uri;
+        log::info!("New request to the backend: {:?}\n", request);
```
ipetr0v, line 54: I think `log::info!` adds `\n` automatically.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+                error
+            )),
+            Ok(response) => Ok(hyper_to_trusted_response(response).await),
+        }
+    }
+}
+
+// We keep the body and the uri from the `hyper::Request`. We convert the body to vec, so it has to
+// be read fully.
+async fn hyper_to_trusted_request(hyper_request: Request<Body>) -> TrustedShufflerRequest {
```
ipetr0v, line 68: I think it should be `hyper_to_trusted_shuffler_request`

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+}
+
+// Apart from setting the body and the URI, for our test backend setting to HTTP/2 and POST
+// seems sufficient.
+fn trusted_to_hyper_request(trusted_shuffler_request: TrustedShufflerRequest) -> Request<Body> {
+    let body = Body::from(trusted_shuffler_request.body);
+    Request::builder()
+        .uri(trusted_shuffler_request.uri)
+        .method(Method::POST)
+        .version(http::Version::HTTP_2)
```
ipetr0v, line 86: I think this is here temporarily, because it assumes that we only support gRPC. But for now it's ok, since we'll need this for the experiment.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -63,40 +119,26 @@ impl hyper::service::Service<Request<Body>> for Service {
...
-                    }
+            // gRPC call: (&Method::POST, "/experimental.trusted_shuffler.echo.Echo/Echo")
+            // http call: (&Method::POST, "/request")
+            let request_start = Instant::now();
+            match trusted_shuffler
+                .invoke(hyper_to_trusted_request(request).await)
+                .await
+            {
+                Ok(response) => {
+                    let _response_time = request_start.elapsed();
```
ipetr0v, line 133: nit: why does it have `_` as a prefix?

experimental/trusted_shuffler/scripts/run_trusted_shuffler
tiziano88, line 8: (optional) Not sure I see the point of having a script that spawns the command in the background. Why not just make this script `exec` the command, and the caller of the script can decide to run it in the background if they want to?

experimental/trusted_shuffler/server/src/http.rs
tiziano88, line 54: I don't think you need an additional newline.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+
+// We generate a default `hyper::Response` and set only the body.
+fn trusted_to_hyper_response(trusted_shuffler_request: TrustedShufflerResponse) -> Response<Body> {
+    let body = Body::from(trusted_shuffler_request.body);
+    Response::new(body)
+}
+
+// We keep only the body from the `hyper::Response`. We convert the body to vec, so it has to be
+// read fully.
+async fn hyper_to_trusted_response(hyper_response: Response<Body>) -> TrustedShufflerResponse {
```
ipetr0v, line 99: You could implement [`std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html) trait (which will automatically implement the `Into` trait as well).

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -22,48 +22,69 @@ mod tests;
...
+    Response(TrustedShufflerResponse),
+}
+
+// We only compare the body, but we need revisit this and decide if this
+// is how we want to sort our requests and responses.
+fn cmp(w1: &Wrapper, w2: &Wrapper) -> Ordering {
+    match (w1, w2) {
+        (Wrapper::Request(r1), Wrapper::Request(r2)) => r1.body.cmp(&r2.body),
+        (Wrapper::Response(r1), Wrapper::Response(r2)) => r1.body.cmp(&r2.body),
+        (_, _) => panic!(
```
ipetr0v, line 65: Should it panic or is there a special constant in `Ordering` for this?

ipetr0v: You could also look into implementing the following traits: https://doc.rust-lang.org/std/cmp/index.html

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -138,34 +162,41 @@ impl TrustedShuffler {
     }
 
     // Lexicographically sorts requests and sends them to the backend using the
-    // [`TrustedShuffler::request_handler`].
+    // [`TrustedShuffler::request_handler`]. Blocks until all responses are received, except if a
+    // timeout is set, then it sends a default response.
```
ipetr0v, line 166: Does it mean that we can't process a new request batch until the first batch is done?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -193,11 +224,31 @@ impl TrustedShuffler {
...
             )
             .collect::<Result<Vec<_>, _>>()
             .map(|_: Vec<()>| ())
     }
 }
+
+fn get_request(w: Wrapper) -> TrustedShufflerRequest {
+    match w {
+        Wrapper::Request(r) => r,
+        _ => panic!(
```
ipetr0v, line 239: I think panics are generally discouraged because they stop the whole program. It may be better to return an error.
Or in this case implement them as compile time checks (via generics and etc.)

ipetr0v: But it would probably be a separate PR

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -27,40 +27,52 @@ impl RequestHandler for TestRequestHandler {
...
+fn generate_response(request: &TrustedShufflerRequest) -> TrustedShufflerResponse {
+    TrustedShufflerResponse {
+        body: request.body.clone(),
+    }
 }
 
 // Generates request which will be dropped by the test backend, because we indicate it has to be
 // dropped in the request itself.
-fn generate_dropped_request() -> Vec<u8> {
+fn generate_dropped_request() -> TrustedShufflerRequest {
```
ipetr0v, line 53: Why do we need these drop requests? Is it for testing timeouts?

ipetr0v (approved):

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
-            Ok,
-        )
+    async fn handle(
+        &self,
+        request: TrustedShufflerRequest,
+    ) -> anyhow::Result<TrustedShufflerResponse> {
+        let mut request = trusted_to_hyper_request(request);
+
+        // We want to keep the path of the orginal request from the client.
+        let path = request.uri().path();
```
tiziano88, line 47: Isn't this removing query params and potentially other parts of the URI? I think it should be possible to just clone the original URI, and then modify the "authority" part of that (which also avoids having to concatenate strings together). In particular, https://docs.rs/hyper/latest/hyper/struct.Uri.html#method.into_parts may be useful, together with https://docs.rs/http/0.2.8/http/uri/struct.Authority.html#method.from_static to create the new authority, and then replace it with https://docs.rs/hyper/latest/hyper/struct.Uri.html#method.from_parts .

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+        // And extend the URL to the backend.
+        let new_uri = format!("{}{}", self.backend_url, path)
+            .parse::<Uri>()
+            .expect("Couldn't parse URI for backend.");
+        let uri = request.uri_mut();
+        *uri = new_uri;
+        log::info!("New request to the backend: {:?}\n", request);
+
+        match send_with_request(request).await {
```
tiziano88, line 56: Doesn't the usual `.context(...)` work here for error handling?

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+            Err(error) => Err(anyhow!(
+                "Couldn't receive response from the backend: {:?}",
+                error
+            )),
+            Ok(response) => Ok(hyper_to_trusted_response(response).await),
+        }
+    }
+}
+
+// We keep the body and the uri from the `hyper::Request`. We convert the body to vec, so it has to
```
tiziano88, line 66:
```suggestion
// We keep the body and the URI from the `hyper::Request`. We convert the body to `Vec`, so it has to
```

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -26,26 +26,82 @@ use std::{
...
+    }
+}
+
+// Apart from setting the body and the URI, for our test backend setting to HTTP/2 and POST
+// seems sufficient.
+fn trusted_to_hyper_request(trusted_shuffler_request: TrustedShufflerRequest) -> Request<Body> {
+    let body = Body::from(trusted_shuffler_request.body);
+    Request::builder()
+        .uri(trusted_shuffler_request.uri)
+        .method(Method::POST)
```
tiziano88, line 85: (optional) Supporting also arbitrary methods may be an easy win to allow using this same exact binary as a proxy for arbitrary web servers, e.g. it can be pointed in front of wikipedia or some other server and it would make a cool demo from a browser (assuming the web server is HTTP2 compatible). But we can look into that separately.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -63,40 +119,26 @@ impl hyper::service::Service<Request<Body>> for Service {
...
-                        }
-                        Err(error) => {
-                            error!("Couldn't receive response: {:?}", error);
-                            let mut internal_error = Response::default();
-                            *internal_error.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
-                            Ok(internal_error)
-                        }
-                    }
+            // gRPC call: (&Method::POST, "/experimental.trusted_shuffler.echo.Echo/Echo")
+            // http call: (&Method::POST, "/request")
```
tiziano88, line 126:
```suggestion
            // HTTP call: (&Method::POST, "/request")
```

experimental/trusted_shuffler/server/src/http.rs
tiziano88, line 133: Did you mean to also log this?

tiziano88 (approved):
Great, looking forward to it running on Cloud :)

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Yes, I agree, I'd revisit that once we know how we want to sort our TrustedShufflerRequests/Responses.

experimental/trusted_shuffler/scripts/run_trusted_shuffler
mariaschett: Done, thanks, it was an oversight!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Done, thank you!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Done!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Done.

tiziano88:
@mariaschett also I'm curious, did you run this locally already and connected to an existing back-end gRPC service, or will we try that on cloud run after this is submitted?

conradgrobler (approved):
It seems like some of the recent fixes have not yet been pushed to the repo.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Actually, I had to roll back, because it requires a static lifetime for backend_url, which would be a bigger change. I think we should in general store not a string, but an Authority in `HttpRequestHandler`. But I'd leave that for a seperate PR.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Done.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: So for now the experimental/trusted_shuffler/run script still works :D

experimental/trusted_shuffler/server/src/http.rs
mariaschett: It is not so easy because we are having `async` here. I could for two of the four methods, but I thought to go with consistency.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Done.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Logging it now :)

mariaschett: Logging it now, thank you!

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Rephrased to clarify.

experimental/trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Yes!

mariaschett:
> @mariaschett also I'm curious, did you run this locally already and connected to an existing back-end gRPC service, or will we try that on cloud run after this is submitted?

Yes, I did run TrustedShuffler locally and connected to our [run_grpc_backend](https://github.com/project-oak/oak/blob/1c41703b541680b467eb45cb620cbbef4a047009/experimental/trusted_shuffler/scripts/run_grpc_backend) by using our [single gRPC client](https://github.com/project-oak/oak/blob/1c41703b541680b467eb45cb620cbbef4a047009/experimental/trusted_shuffler/scripts/run_single_grpc_client).

PR #3085, opened 2022-07-25
Remove remaining references to baremetal tests

jul-sh:
As of #3034 the binaries no longer have a test entrypoint, hence it seems these references can be removed :)


andrisaar (approved):
Good catch!

PR #3086, opened 2022-07-25
Kernel: Add simple breakpoint exception handler that logs

jul-sh:
After this change the kernel no longer double faults when encountering a breakpoint exception and instead continues afterwards. 

Tested this manually, and appears to work fine. This PR intentionally doesn't add an automated test since baremetal tests were removed in #3034, but would be happy to add a test back for this. 

Unfortunately this does not unlock debugging in crosvm. The runtime still crashes in crosvm when continuing from a breakpoint.


conradgrobler (approved):
Thanks.

mariaschett (approved):
LGTM! 

Is it worth it to add acomment that debugging Crosvm still fails?

Issue #3082, opened 2022-07-22
Remove "monolithic" Oak Functions binaries


tiziano88:
Once the rest of the tasks in #3075 are completed, we should remove the existing "monolithic" versions of the Oak Functions loader, namely

- [`oak_functions_loader_base`](https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_base)
- [`oak_functions_loader_unsafe`](https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_unsafe)

tiziano88:
Fixed by #3319.

Issue #3080, opened 2022-07-22
Support loading a linux binary as a child process in bare metal loader


tiziano88:
Currently the bare metal loader supports loading the trusted runtime in either QEMU or crosvm.

@conradgrobler suggested we should add support for spawning a trusted runtime instance (of the appropriate format) as a linux child process on the same host (i.e. without virtualization needed). This would allow us to keep developing a more fully featured version of the oak runtime with additional functionality like TF inference and differentially private metrics, which would otherwise be nontrivial to get to work in an actual bare metal environment.

As a starting point, this would probably require adding a new value to this enum:

https://github.com/project-oak/oak/blob/dd2dc6b1cae87e8210a4f532d040fb70d27f69ca/experimental/oak_baremetal_loader/src/main.rs#L43-L46

The loader and the child process should communicate with each other over a channel that follows the existing IDL interface, though it may be implemented via a Unix Domain Socket, a network socket, a named pipe, a pre-opened file descriptor, stdio, or some mechanism.

jul-sh:
I'd be curious of you had thoughts about including the trusted runtime as a library in the launcher, instead of spawning a separate child process? Communication could still take place over the IDL, but instead of a UDS the channel would simple buffer (eg `std::io::Cursor<std::vec::Vec<u8>>`. 

In principle that seems simpler and more straightforward?

conradgrobler:
It feels problematic to include a library that will mostly be unused (only used for non-production testing). I guess that could be solved via features, but features on a binary crate causes other problems (e.g. not easily being able to tell which features were enabled when the binary was built, as the name remains the same)

While it might simplify some things, it might complicate others because the model for starting and stopping the runtime would be different.

jul-sh:
Fair. :) I mean arguably the OSS untrusted loader as a whole will be mostly used for non-production testing at the moment, but I get the point about a separate binary creating a more obvious and clearer seperation. :)

jul-sh:
Fixed with #3106

PR #3093, opened 2022-07-26
Rename to Untrusted Loader to Untrusted Launcher

jul-sh:
We've been using both terms a little interchangeably, it'd be nice to stop doing so. Untrusted Launcher differentiates it from others loaders we have, is consistent with what AMD call the VM creation process.


ipetr0v:
There is also `experimental/oak_functions_loader_linux_vsock` but it's a linux wrapper for Oak runtime which runs in a TEE, so it's not a launcher.

We probably need to define how to call it at some point as well.

jul-sh:
@ipetr0v I actually think loader might be a fine term for that specific wrapper. It's seems rather similar to the "loader" that is part of the old oak functions linux binary.

tiziano88:
Perhaps we can call the trusted part just `oak_functions_runtime_$PLATFORM` (where `$PLATFORM` may be the VMM or something else like `linux`, even if it's not super accurate I think it will be ok), e.g.:

- `oak_functions_runtime_vanadium`
- `oak_functions_runtime_crosvm`
- `oak_functions_runtime_qemu`
- `oak_functions_runtime_linux`

And we can keep using `oak_functions_runtime` (without `$PLATFORM`) as the name of the library crate.

WDYT?

ipetr0v:
I think we could also remove "Untrusted" from the Launcher but keep "Trusted" for the Runtime (because this would imply that the Launcher is untrusted).

But I'm also fine with just having Launcher & Runtime without trusted/untrusted.

jul-sh:
I do think that removing untrusted from the launcher is probably a good idea. In the context of baremetal that adjective was helpful for clarity. However once #3080 and #3082 are done, the launcher it will be trusted in some deployment configurations, and untrusted in others.

jul-sh:
It's why I didn't include that term in the new crate name introduced in this PR (though the docs etc still use it)

PR #3097, opened 2022-07-27
Refactor launcher in preparation for supporting a runtime linux binary

jul-sh:
Soon the launcher will launch not just VMs, hence this removes references to VMs and VMMs where they're not needed. In preparation for #3080.

### DIFF
--- a/experimental/oak_baremetal_launcher/src/main.rs
+++ b/experimental/oak_baremetal_launcher/src/main.rs
@@ -18,12 +18,13 @@
 
 use anyhow::Context;
 use clap::Parser;
-use crosvm::Crosvm;
+use instance::LaunchedInstance;
+use instance_crosvm::CrosvmInstance;
+use instance_qemu::QemuInstance;
 use oak_baremetal_communication_channel::{
     client::{ClientChannelHandle, RequestEncoder},
     schema,
 };
-use qemu::Qemu;
 use std::{
     fs,
     io::{BufRead, BufReader},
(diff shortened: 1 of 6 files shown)


experimental/oak_baremetal_launcher/src/main.rs
```diff
@@ -18,51 +18,57 @@
...
+mod instance_qemu;
 mod lookup;
-mod qemu;
 mod server;
-mod vmm;
 
-#[derive(clap::ArgEnum, Clone, Debug, PartialEq)]
+/// Parameters used for launching VM instances
+#[derive(Parser, Clone, Debug, PartialEq)]
+pub struct VmParams {
```
andrisaar, line 44: Just from a code layout perspective I think the params should still be in (the moral equivalent of) `instance.rs`.

experimental/oak_baremetal_launcher/src/main.rs
```diff
@@ -18,51 +18,57 @@
...
-    #[clap(long, parse(from_os_str), validator = path_exists, default_value_os_t = PathBuf::from("/usr/bin/qemu-system-x86_64"))]
-    qemu: PathBuf,
-
-    /// Path to the `crosvm` binary.
-    #[clap(long, parse(from_os_str), validator = path_exists, default_value_os_t = PathBuf::from("/usr/local/cargo/bin/crosvm"))]
-    crosvm: PathBuf,
-
     /// Execution mode.
-    #[clap(arg_enum, long, default_value = "qemu")]
+    #[clap(subcommand)]
```
andrisaar, line 69: Ooh, TIL about subcommand.

experimental/oak_baremetal_launcher/src/main.rs
jul-sh: I typically find it helpful to keep the structs defining CLI args in the head of the main file, but fine, will change, don't feel strongly about this.

jul-sh: On a second thought, I think we should keep it here. Having the params be in one place is also what xtask does. Similarly they reuse the `path_exists` function in this file for validation. We could import it around, but I don't think it'd really make things clearer.

experimental/oak_baremetal_launcher/src/main.rs
jul-sh: It's pretty nice, imo we should use it more

jul-sh:
@andrisaar Thanks for the review :). I'm not personally convinced that we should start nesting directory structure. There are plenty arguments to be made in favor of a flat directory structure, and regardless of my personal opinion, the prefix_value pattern is I think by far most common in our repo. 

If anything we could chose to inline the contents of `instance.rs`, but I'm not convinced it makes a difference.

andrisaar:
Yeah, I agree, inlining `instance.rs` wouldn't make it better. In my mind `instance.rs` defines the API, and then `crosvm.rs` and `qemu.rs` implement that API -- ideally that's the only thing they should depend on.

I still think a nested structure would be an improvement (and maybe we should use it more often to enforce proper isolation), but that's not a hill to die on.

andrisaar (approved):

PR #3099, opened 2022-07-28
Remove unused xtask options

jul-sh:


andrisaar (approved):

rbehjati:
Why is this removed? This command is used in provenance generation:
https://github.com/project-oak/oak/blob/87a33746f3f512ec3ece204fa26704bdf9a08846/buildconfigs/oak_baremetal_app_crosvm.toml#L6

Could you please suggest a fix, or revert this change?

rbehjati:
@jul-sh Please revert this. See also #3152.

PR #3101, opened 2022-07-28
Remove Wrapper by splitting to Request/ResponseMessage

mariaschett:


experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -37,6 +37,13 @@ pub struct TrustedShufflerRequest {
     pub body: Vec<u8>,
     pub uri: Uri,
 }
+
+// We only compare the body, but we need revisit this and decide if this
+// is how we want to sort our requests.
+fn cmp_requests(r1: &TrustedShufflerRequest, r2: &TrustedShufflerRequest) -> Ordering {
```
ipetr0v, line 43: I think you can implement a [`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) trait on the `TrustedShufflerRequest` instead of defining a global compare function

mariaschett: Yeah, I hoped so too, but `PartialOrd` on `TrustedShufflerRequest` would require `Uri` to be `PartialOrd`, which it is not. Further: we'll need to revisit that when we decide how to sort our (decrypted) requests. However, I refactored to clarify---also that we only use the body to compare.

ipetr0v: But you are comparing bodies, which should have `PartialOrd`.
You can implement the `PartialOrd` trait and just compare bodies (I mean implement it manually and not use `derive`).

ipetr0v (approved):

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: True! However, I'd leave it as is for the moment, because it makes clear what we are comparing for when we revisit sorting of decrypted requests.

PR #3106, opened 2022-08-01
Add option to launch the baremetal runtime as a linux binary

jul-sh:


experimental/oak_baremetal_launcher/src/instance_binary.rs
```diff
@@ -0,0 +1,90 @@
...
+        info!("Cleaning up and shutting down.");
+        self.comms_host.shutdown(Shutdown::Both)?;
+        self.tempdir
+            .take()
+            .ok_or_else(|| anyhow::Error::msg("failed to take the tempdir"))?
+            .close()
+            .context("failed delete the socket file")?;
+        self.instance.start_kill()?;
+        self.wait().await
+    }
```
jul-sh, line 82: In a related comment, the cleanup logic required by the `LaunchedInstance` (used by all the VMs etc) seems like it could use updating fix subtle bugs. At the moment, cleanup is done only in the kill method, meaning that if the child instance dies unexpectedly, it is never performed.

experimental/oak_baremetal_launcher/src/instance_binary.rs
```diff
@@ -0,0 +1,90 @@
...
+/// An instance of the runtime running directly as a linux binary
+pub struct BinaryInstance {
+    tempdir: Option<tempfile::TempDir>,
+    comms_host: UnixStream,
+    instance: tokio::process::Child,
+}
+
+impl BinaryInstance {
+    pub fn start(params: BinaryParams) -> Result<Self> {
+        let tempdir = tempfile::tempdir().context("failed to create temp dir")?;
```
andrisaar, line 35: I don't think there's any point in giving the socket a name, just use raw file descriptors.

Use `UnixStream::pair()` to get a pair of anonymous connected sockets, pass one to the binary and use the other yourself, and you avoid having to deal with the temporary directories (and possibly leaving them around when you crash).

(As a side note, when implementing the original qemu/crosvm loaders, I thought about using tempdir as well, but tried really hard to avoid having to give a name to the sockets.)

experimental/oak_functions_loader_linux_uds/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::os::unix::net::UnixStream;
+
+#[derive(Parser, Clone, Debug)]
+#[clap(about = "Oak Functions Loader Linux UDS")]
+pub struct Opt {
+    #[clap(long, help = "UDS address to use for the communication channel")]
+    pub comms_address: std::path::PathBuf,
+}
+
+struct Logger {}
```
andrisaar, line 29: why not use some existing logger, such as `env_logger` we use for the launcher anyway?

xtask/src/internal.rs
```diff
@@ -54,7 +54,7 @@ pub struct Opt {
 
 #[derive(Subcommand, Clone, Debug)]
 pub enum Command {
-    RunVmTest,
+    RunLauncherTest,
```
andrisaar, line 57: Please move the rename into a separate PR.

xtask/src/launcher.rs
```diff
@@ -32,6 +32,7 @@ use crate::internal::*;
 pub enum LauncherMode {
     Qemu,
     Crosvm,
+    Binary,
```
andrisaar, line 35: Matter of taste, but I don't like the term 'binary' for this -- it's not like the other two aren't binaries. 'Linux' would be more descriptive, or maybe even 'native'.

experimental/oak_baremetal_launcher/src/instance_binary.rs
jul-sh: The downside to raw file descriptors is that it would require us to use unsafe rust in the loader to turn the file descriptor provided in the args into a UnixSocket. 

We can do it if you strongly feel that's the better way, but I thought all things considered temp files are probs fine

xtask/src/launcher.rs
jul-sh: I wouldn't go with linux, since one can also run linux inside one of the VMs (see experimental/oak_functions_loader_linux_vsock). I suppose native works, can go with that

experimental/oak_baremetal_launcher/src/instance_binary.rs
tiziano88: Which particular unsafe function would be used for the conversion? I don't think it's a problem to have unsafe code, and I also think it would look cleaner than the temp files.

experimental/oak_functions_loader_linux_uds/src/main.rs
jul-sh: It seems desirable to disable logs entirely for release builds, to avoid inadvertent logs. That's the same approach the old linux binary runtime used: https://github.com/project-oak/oak/blob/main/oak_functions/loader/src/logger.rs

experimental/oak_baremetal_launcher/src/instance_binary.rs
jul-sh: std::os::unix::io::FromRawFd
https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html

jul-sh: > This function is also unsafe as the primitives currently returned have the contract that they are the sole owner of the file descriptor they are wrapping. Usage of this function could accidentally allow violating this contract which can cause memory unsafety in code that relies on it being true.

andrisaar: Inside the loader you should be able to just use `/dev/fd/<x>` as a regular file, no need for unsafe code.

xtask/src/launcher.rs
andrisaar: Ah yes, we have linux overloaded already as well :(
In that case, yes, I think 'Native' would be a better name.

experimental/oak_baremetal_launcher/src/instance_binary.rs
jul-sh: updated to use a file descriptor. We did end up having to use unsafe code, but should be fine

experimental/oak_functions_loader_linux_native/README.MD
```diff
@@ -0,0 +1,3 @@
+# Oak Functions Loader Linux UDS
+
+A thin wrapper for the "baremetal" runtime to compile it as a linux binary. The wrapping loader provides communication with a launcher by connecting to a Unix Domain Socket.
```
andrisaar, line 3: Update this to say file descriptors as well.

andrisaar (approved):
Thanks.

PR #3109, opened 2022-08-01
Remove the qemu variant of our kernel

andrisaar:
It's quite clear by now that our long-term strategy will focus on the crosvm variant, as that's what we'll be using internally.
Thus -- even though the debugging story isn't quite there yet -- we can reduce our maintenance burden by getting rid of the qemu variant.

I have left the infrastructure that knows how to handle multiple variants intact, though: soon we'll have a native Linux variant, so we'll still need all of that.


jul-sh:
Is the ability to debug a concern?

jul-sh:
> b) If the bug is in the kernel code, we may need to attach a debugger to the VM -- but you're going to fix the debugging story with crosvm :)

haha no pressure :P

jul-sh (approved):

ipetr0v (approved):

PR #3117, opened 2022-08-02
Move instance implementations into a subdirectory.

andrisaar:
Let's do some bikeshedding :)

Case in point, I don't like using underscores as kind of namespace delineators when we have a perfectly fine `/` that natively works as one.

Thus, here's my proposal: move `instance_*.rs` to `instance/*.rs`. I think this code is cleaner[*], but as it's a matter of taste I'm willing to entertain other opinions.

[*] -- having to include `crate::path_exists` is ugly, but then again, I'm surprised something like that doesn't exist in some other crate we could include.

### DIFF
--- a/experimental/oak_baremetal_launcher/src/instance/crosvm.rs
+++ b/experimental/oak_baremetal_launcher/src/instance/crosvm.rs
@@ -14,15 +14,17 @@
 // limitations under the License.
 //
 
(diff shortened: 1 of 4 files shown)


tiziano88 (approved):
Thanks, I like the new file structure too

jul-sh (approved):
🤷‍♀️

Issue #3081, opened 2022-07-22
Make all integration tests use the split "bare metal" loader


tiziano88:
This is in order to align tests to exercise a version of Oak as close as possible to the one that will run in a real deployment, and in preparation for deleting the existing "monolithic" binaries.

tiziano88:
Fixed in #3317

PR #3123, opened 2022-08-04
Launcher: Support grpc-web with feature flags

jul-sh:
Will allow us to test the web client with the launcher. 

Tracking issue: #3081


rbehjati (approved):
Looks good to me, but I don't have much context!

PR #3124, opened 2022-08-04
Fix message size and socket blocks in Oak Functions Linux wrapper

ipetr0v:
This PR fixes:
- Vsock channel message sizes were read as BigEndian instead of NativeEndian
- Socket was set into the non-blocking mode by default causing the runtime to receive an error when reading an empty channel

### DIFF
--- a/experimental/oak_functions_loader_linux_vsock/src/main.rs
+++ b/experimental/oak_functions_loader_linux_vsock/src/main.rs
@@ -32,7 +32,6 @@ mod tests;
 
 use crate::channel::Channel;
 use clap::Parser;
-use log::info;
 use oak_remote_attestation::handshaker::{
     AttestationBehavior, EmptyAttestationGenerator, EmptyAttestationVerifier,
 };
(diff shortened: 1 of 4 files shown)


experimental/oak_functions_loader_linux_vsock/src/main.rs
```diff
@@ -53,14 +53,17 @@ pub struct Opt {
...
+        .map_err(|error| anyhow!("Couldn't set socket into blocking mode: {}", error))?;
     Ok(stream)
 }
 
 fn main() {
     let opt = Opt::parse();
 
     let stream = create_vsock_stream(opt.file_descriptor).expect("Couldn't create channel");
-    info!(
+    println!(
```
jul-sh, line 66: It seems fine to stick to the logger though?

jul-sh (approved):
noice

experimental/oak_functions_loader_linux_vsock/src/main.rs
ipetr0v: The internal sandbox currently couldn't set up the `RUST_LOG` environment variable

PR #3125, opened 2022-08-04
Refactor runtime to make entrypoint explicit

jul-sh:
The entrypoint was previously in a submodule and had a somewhat misleading name.


andrisaar:
I'd leave the rest of the code in `framing.rs` (or something equivalent), and move just `start` to `lib.rs`.

jul-sh:
What would you call the equivalent mod though, `implementation.rs`? :P

andrisaar:
> What would you call the equivalent mod though, `implementation.rs`? :P

Probably just `runtime.rs` and maybe even rename `RuntimeImplementation` to `Runtime`.

But this is well into bikeshedding territory, getting rid of the `framing::handle_frames` is a massive improvement on its own :)

andrisaar (approved):

PR #3126, opened 2022-08-04
Implement fixed response size

jul-sh:


ipetr0v (approved):

experimental/oak_baremetal_launcher/src/main.rs
```diff
@@ -58,6 +58,10 @@ struct Args {
     #[clap(subcommand)]
     mode: Mode,
 
+    /// Consistent response size that the runtime should apply
+    #[clap(long, default_value = "1024")]
+    constant_response_size: u32,
```
tiziano88, line 63: Does this mean we are moving away from the config file for the policy? Not that it matters too much at this point, since this is the only value that is really part of the policy anyways.

experimental/oak_baremetal_launcher/src/main.rs
```diff
@@ -58,6 +58,10 @@ struct Args {
     #[clap(subcommand)]
     mode: Mode,
 
+    /// Consistent response size that the runtime should apply
```
tiziano88, line 61: I assume this is in bytes (and maybe it would be good to clarify it)

experimental/oak_baremetal_launcher/src/main.rs
```diff
@@ -58,6 +58,10 @@ struct Args {
...
+    #[clap(long, default_value = "1024")]
```
tiziano88, line 62: Personally I would have left the default to 0, and also emit an error at startup if it's 0, effectively forcing the caller to specify it.

tiziano88 (commented):
Thanks, good stuff!

PR #3127, opened 2022-08-04
Fix incorrectly named feature check

jul-sh:
Fleeting mistake introduced in #3123


rbehjati (approved):
I wonder if this is something that would have been caught by a test. If so, is it possible to add a test?

Issue #3136, opened 2022-08-08
Add runtime implementation integration tests


jul-sh:
The (baremetal) runtime is presently only tested by a) [rudimentary end to end tests](https://github.com/project-oak/oak/blob/main/xtask/src/launcher.rs) and b) unit tests of components it imports from the legacy Oak Functions runtime. 

We should add more extensive integration tests to the runtime to prevent issues like #3135 from going uncaught.

PR #3137, opened 2022-08-08
Isolate runtime logic from communication channel logic

jul-sh:
This makes the runtime crate a pure implementation of flatbuffer service [schema](https://github.com/project-oak/oak/blob/main/experimental/oak_baremetal_runtime/schema.fbs/#L3). It does so by moving all logic for translating between flatbuffer and the communication protocol into the `oak_baremetal_channel` crate.

I think this makes most logical sense, and will make it far easier to add integration tests covering the runtime behavior (issue ref: https://github.com/project-oak/oak/issues/3136)

### DIFF
--- a/experimental/oak_baremetal_runtime/src/lib.rs
+++ b/experimental/oak_baremetal_runtime/src/lib.rs
@@ -51,10 +51,14 @@ use oak_remote_attestation_sessions::SessionId;
 
 /// Sole entrypoint to the runtime. Starts the runtime and listens for requests
 /// on the provided channel.
-pub fn start<G: 'static + AttestationGenerator, V: 'static + AttestationVerifier>(
+pub fn start<G, V>(
     channel: Box<dyn Channel>,
     attestation_behavior: AttestationBehavior<G, V>,
-) -> anyhow::Result<!> {
+) -> anyhow::Result<!>
+where
+    G: 'static + AttestationGenerator,
+    V: 'static + AttestationVerifier,
+{
     let mut invocation_handler = RuntimeImplementation {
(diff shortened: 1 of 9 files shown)


andrisaar (approved):

experimental/oak_baremetal_runtime/src/lib.rs
```diff
@@ -33,53 +33,20 @@ mod wasm;
...
-    let channel_handle = &mut ServerChannelHandle::new(channel);
-    loop {
-        let request_message = channel_handle
-            .read_request()
-            .context("couldn't receive message")?;
-        let request_message_invocation_id = request_message.invocation_id;
-        let response = invocation_handler.invoke(request_message.into());
-        // For now all messages are sent in sequence, hence the id of the next
-        // response always matches that of the preceeding request.
-        // TODO(#2848): Allow messages to be sent and received out of order.
```
conradgrobler, line 72: This TODO seems to have gone missing from the resulting code, even though I think it is still relevant.

conradgrobler (approved):

experimental/oak_baremetal_runtime/src/lib.rs
jul-sh: Good catch! I think it's fine to remove it from this spot, since it deals just with responding requests. We probably want the single-threaded runtime to respond to request in the order the come in. 

The todo still exists in the launcher, which deals the order in which requests are sent (and which atm waits for the response before sending a new requests). See https://github.com/project-oak/oak/blob/071dd8847569aab186dd269fe6da181aee872df1/experimental/oak_baremetal_launcher/src/channel.rs#L45

PR #3140, opened 2022-08-09
Add placeholder cpuid and secrets sections to the ELF kernel.

andrisaar:
For now, these do nothing, but will be used to denote locations in the memory where the SEV-SNP cpuid and secrets pages will be placed.

Although these are supposed to be page-sized (4K) I aligned the structures at 2M boundaries, as we've yet to figure out what page size do we want to use in the kernel (2M hugepages vs 4K pages), so that'll make it compatible with both at the cost of some wasted memory.

### DIFF
--- a/experimental/oak_baremetal_app_crosvm/layout.ld
+++ b/experimental/oak_baremetal_app_crosvm/layout.ld
@@ -18,7 +18,9 @@ ENTRY(_start)
 
 PHDRS
 {
-    ram  PT_LOAD FLAGS(5);
+    ram     PT_LOAD FLAGS(5);
+    cpuid   PT_LOAD FLAGS(1 << 23);
+    secrets PT_LOAD FLAGS(1 << 24);
 }
 
 SECTIONS {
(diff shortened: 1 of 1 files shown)


experimental/oak_baremetal_app_crosvm/layout.ld
```diff
@@ -50,11 +52,19 @@ SECTIONS {
...
+    } : ram
 
     /* Stack grows down, so stack_start is the upper address in memory. */
     .stack (NOLOAD) : ALIGN(4K) {
         . += 512K;
-    }
+    } : ram
     stack_start = .;
+
+    .cpuid (NOLOAD) : ALIGN(2M) {
```
conradgrobler, line 63: I don't think the 2M alignment adds value. Even if we decide to us 2M pages, the AMD Secure Processor that will populate these will always use 4K granularity for adding these special pages.

conradgrobler (approved):

experimental/oak_baremetal_app_crosvm/layout.ld
andrisaar: My idea was that if we're going for 2M page size, this will allow us to map the cpuid page and secrets page separately (otherwise the two 4K pages would be on the same 2M page). Not sure how much value it would add in the grand scheme, though.

conradgrobler: I can't think of a scenario where we would want different page table attributes or permissions for the two sections (e.g. we want both to be encrypted,  read-only and non-executable). But, I don't think it makes a big difference for now either way so it you think there might be value in future it is fine to leave as is.

PR #3141, opened 2022-08-09
Create separate sections in the ELF file for parts of the program that should have different permissions.

andrisaar:
None of the flags are actually respected right now, as the ELF loaders in the VMMs don't set up permissions (or page tables) for us.

But it's a first step toward setting up page tables with proper permissions.

### DIFF
--- a/experimental/oak_baremetal_app_crosvm/layout.ld
+++ b/experimental/oak_baremetal_app_crosvm/layout.ld
@@ -18,7 +18,20 @@ ENTRY(_start)
 
 PHDRS
 {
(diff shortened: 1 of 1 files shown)


conradgrobler (approved):

mariaschett (approved):
If I understand your issue description correctly, these flags are not yet respected.

Would it make sense to add a comment explaining that in the code?

andrisaar:
> If I understand your issue description correctly, these flags are not yet respected.
> 
> Would it make sense to add a comment explaining that in the code?

Yeah, probably a good idea. Done.

PR #3150, opened 2022-08-11
Update deploy script of Trusted Shuffler

mariaschett:
To allow for environment variables, which can be updated on Google Cloud, we don't use distroless docker any more. 

We also build the Rust binary with the builder Docker, so we don't rely on the host having Rust installed.

### DIFF
--- a/experimental/trusted_shuffler/scripts/deploy
+++ b/experimental/trusted_shuffler/scripts/deploy
@@ -10,4 +10,7 @@ gcloud run deploy "${DEPLOYMENT}" \
     --region=europe-west1 \
     --image="${DOCKER_IMAGE}" \
     --use-http2 \
-    --allow-unauthenticated
+    --allow-unauthenticated \
+    --set-env-vars "BACKEND_URL=http://localhost:8888" \
+    --set-env-vars "K=1" \
+    --set-env-vars "LISTEN_ADDRESS=[::]:8080" 
--- a/experimental/trusted_shuffler/scripts/deploy
+++ b/experimental/trusted_shuffler/scripts/deploy
@@ -11,6 +11,6 @@ gcloud run deploy "${DEPLOYMENT}" \
     --image="${DOCKER_IMAGE}" \
     --use-http2 \
     --allow-unauthenticated \
-    --set-env-vars "BACKEND_URL=http://localhost:8888" \
-    --set-env-vars "K=1" \
-    --set-env-vars "LISTEN_ADDRESS=[::]:8080" 
+    --set-env-vars BACKEND_URL=http://localhost:8888 \
+    --set-env-vars K=1 \
+    --set-env-vars LISTEN_ADDRESS=[::]:8080
--- a/experimental/trusted_shuffler/scripts/deploy
+++ b/experimental/trusted_shuffler/scripts/deploy
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2101 # False positive on [::] in LISTEN_ADDRESS.
 
 readonly SCRIPTS_DIR="$(dirname "$0")"
 # shellcheck source=scripts/common
--- a/experimental/trusted_shuffler/server/src/http.rs
+++ b/experimental/trusted_shuffler/server/src/http.rs
@@ -53,7 +53,10 @@ impl RequestHandler for HttpRequestHandler {
         // And extend the URL to the backend.
         let new_uri = format!("{}{}", self.backend_url, path)
             .parse::<Uri>()
-            .expect("Couldn't parse URI for backend.");
+            .expect(&format!(
+                "Couldn't parse URI for backend: {}{}.",
+                self.backend_url, path
+            ));
         let uri = request.uri_mut();
         *uri = new_uri;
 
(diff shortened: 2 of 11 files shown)


xtask/src/examples.rs
```diff
@@ -484,6 +484,55 @@ fn run_oak_functions_server(server: &Server) -> Box<dyn Runnable> {
...
+        name: "Trusted Shufflder and Backend".to_string(),
+        foreground: Box::new(client_step),
+        background: trusted_shuffler_cmd,
+    };
+
+    Step::WithBackground {
+        name: "Run echo gRPC Trusted Shuffler with k=1".to_string(),
+        foreground: Box::new(trusted_shuffler_step),
+        background: backend_cmd,
+    }
```
mariaschett, line 533: @tiziano88 I believe I am using the `WithBackground` somehow wrongly---it seems xtask seems to get stuck here. Do you know how to use it?

mariaschett: @tiziano88 xtask still gets stuck. Do you have an idea where I could start looking?

tiziano88: Is this workig now, after removing the bash script layer, or does it still get stuck?

experimental/trusted_shuffler/scripts/deploy
```diff
@@ -10,4 +10,7 @@ gcloud run deploy "${DEPLOYMENT}" \
     --region=europe-west1 \
     --image="${DOCKER_IMAGE}" \
     --use-http2 \
-    --allow-unauthenticated
+    --allow-unauthenticated \
+    --set-env-vars "BACKEND_URL=http://localhost:8888" \
```
tiziano88, line 14:
```suggestion
    --set-env-vars=BACKEND_URL=http://localhost:8888 \
```

(I don't think the quotes are necessary)

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -51,9 +51,10 @@ impl RequestHandler for HttpRequestHandler {
         // let new_uri = Uri::from_parts(parts).expect("Failed to create new URI");
         let path = request.uri().path();
         // And extend the URL to the backend.
-        let new_uri = format!("{}{}", self.backend_url, path)
-            .parse::<Uri>()
-            .expect("Couldn't parse URI for backend.");
+        let new_uri = match format!("{}{}", self.backend_url, path).parse::<Uri>() {
+            Ok(new_uri) => new_uri,
+            Err(error) => return Err(anyhow!("Couldn't parse URI for backend: {:?}", error)),
+        };
```
tiziano88, line 57: I think this can be accomplished more idiomatically with map_err and the `?` operator.

experimental/trusted_shuffler/scripts/deploy
mariaschett: Done.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: I didn't manage because of the return.

xtask/src/examples.rs
mariaschett: Removing the indirection helped it not getting stuck, but sadly the backend server still returns with an error: https://github.com/project-oak/oak/runs/7910075586?check_suite_focus=true#step:6:139

I am not sure exactly why. This does not happen if I control+c the backend server locally. I wasn't sure where/how xtask ends the processes if [it is not interrupted](https://github.com/project-oak/oak/blob/a4d9157a9104d6f212f3f0cc23bdcc3e7dbfe09c/xtask/src/main.rs#L91).

Do you have a pointer?

experimental/trusted_shuffler/server/src/http.rs
tiziano88: Is that because there are additional type conversions going on that are not performed by `Into`? Otherwise normally it should work for early returns

xtask/src/examples.rs
tiziano88: How did you confirm that it does not error out if you ctrl+c locally? AFAICT the backend server only handles ctrl+c correctly [in the HTTP case](https://github.com/project-oak/oak/blob/a4d9157a9104d6f212f3f0cc23bdcc3e7dbfe09c/experimental/trusted_shuffler/backend/src/main.rs#L84), not [in the gRPC](https://github.com/project-oak/oak/blob/a4d9157a9104d6f212f3f0cc23bdcc3e7dbfe09c/experimental/trusted_shuffler/backend/src/main.rs#L126-L129)

mariaschett: My bad, thank you, I just tried it by locally ctrl+c'ing a server and it didn't seem unusal. It seems to work now!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: Fixed now.

ipetr0v (approved):

xtask/src/examples.rs
tiziano88: For the record, what you need to look at is the exit code of the binary, not any particular log message.

experimental/trusted_shuffler/scripts/deploy
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2101 # False positive on [::] in LISTEN_ADDRESS.
```
tiziano88, line 2: Maybe the [::] actually needs to be single quoted? Not sure

tiziano88 (approved):
ship it :)

Issue #3066, opened 2022-07-19
Transport-agnostic client library for Oak 


rbehjati:
We need a generic and transport-agnostic Oak Client library with common functionality for interacting with a trusted Oak runtime by sending and receiving encrypted messages. This library should provide the functionality that is in common between different variants of the client, regardless of the remote attestation flavour that is used for establishing the attested connection. 

Here is a high-level description of the behaviour of an Oak client:
* Get encryption keys and create an encryptor object
* Get evidence about the server (endorsement evidence, and optionally some config) and verify that
* If happy, encrypt and send messages to the server

These three things are in general separate concerns, and our design should allow them to be done separately, by different types.

We need this library for both Rust and Java. The idea is to provide abstractions, and a public API based on that, that the client code (e.g., the Android client) can depend on without having to depend on specific implementations. 

The following is the architecture I propose, and have partly implemented in #3062 (in Java), for the generic client API.  

![Oak Client-library](https://user-images.githubusercontent.com/12942581/179790673-b7fc1eec-f5eb-4729-b210-aac3dd388db9.png)

The following diagram shows the general sequence of events on the client.

![Oak Client-Page-2 drawio](https://user-images.githubusercontent.com/12942581/181240315-2c035ec0-f711-4d0e-8803-d18c71e124b5.png)

1. Regardless of the remote attestation protocol, the client first asks the `EvidenceProvider` for evidence about the server. In a complete implementation, the evidence must contain the remote attestation quote, containing the server's signing public key, and endorsement evidence. 
2. The client verifies the evidence. If the provided evidence is valid (with respect to some _policy_ internal to the client), then the server's signing public key is returned. Otherwise, the connection is dropped. 
3. The client then requests the `EncryptorProvider` for an Encryptor. This requires passing the signing public key of the server to the `EncryptorProvider`. (In the current implementation of the interactive remote attestation protocol the `signingPublicKey` is part of the `ServerIdentity`. This is not a good approach, as this public key that is used for handshake is unverified. A better and more secure practice, as described here, is to first verify the signing public key and only proceed with the handshake if the public key is valid.)
4. The client requests the `RpcClientProvider` for an instance of an `RpcClient` that can be used for sending and receiving messages. 
5. The uses the `RpcClient` to send and receive encrypted messages, and uses the `Encryptor` to encrypt and decrypt messages. 

[...]

PR #3151, opened 2022-08-12
Add Placeholder implementation for an AMD-attested Oak client

rbehjati:
This PR:

- Adds placeholder implementation for `AmdAttestedOakClient`, and `AmdAttestationReport`.
- Moves classes related to `Evidence` into a separate package.
- Removes the abstract builder in `OakClient`, as a good implementation of it would require too many type parameters, and would not do much really.

Notable next steps:
- Implement Rekor logEntry verification and `verify` in `EndorsementEvidence`.
- Unifying the Rust and Java implementation. It is probably best to do this after updating the interactive remote attestation implementation. 

Ref #3066

### DIFF
--- a/java/src/main/java/com/google/oak/client/OakClient.java
+++ b/java/src/main/java/com/google/oak/client/OakClient.java
@@ -16,8 +16,8 @@
 
 package com.google.oak.client;
 
+import com.google.oak.evidence.Evidence;
 import com.google.oak.util.Result;
-import java.util.Optional;
 
 /**
  * Abstract client for sending and receiving encrypted messages to a server. We recommend concrete
@@ -31,7 +31,7 @@ import java.util.Optional;
  * @param R type of the requests that this client sends
  * @param T type of the responses that this client receives
  */
-public abstract class OakClient<R, T> implements AutoCloseable {
+public interface OakClient<R, T> extends AutoCloseable {
   /**
    * Sends a request to a remote server and receives the response.
    *
@@ -64,8 +64,8 @@ public abstract class OakClient<R, T> implements AutoCloseable {
   /**
    * An interface for providing instances of Evidence.
    *
-   * <p>A evidence normally includes the public key part of the server's signing key, an instance of
-   * {@code EndorsementEvidence}, and optionally some server configuration information. If the
+   * <p>An evidence normally includes the public key part of the server's signing key, an instance
+   * of {@code EndorsementEvidence}, and optionally some server configuration information. If the
    * client policy requires verification of the server configuration, then the client should be
    * built with a provider that does provide server configuration.
    */
--- a/java/src/main/java/com/google/oak/client/OakClient.java
+++ b/java/src/main/java/com/google/oak/client/OakClient.java
@@ -20,13 +20,15 @@ import com.google.oak.evidence.Evidence;
 import com.google.oak.util.Result;
 
 /**
- * Abstract client for sending and receiving encrypted messages to a server. We recommend concrete
- * subclasses of this class to provide builders for creating instances of the class. The builders
- * should use a {@code RpcClientProvider}, an {@code EncryptorProvider}, and optionally an {@code
- * EvidenceProvider} to build a concrete instance of {@code OakClient}. If an evidence supplier is
- * provided, the client has to verify the correctness of the evidence prior to getting an instance
- * of {@code Encryptor} and sending messages to the server. In a complete, and production-ready
- * implementation, an {@code EvidenceProvider} MUST be provided and verified.
+ * Generic client interface for sending and receiving encrypted messages to a server.
+ *
+ * <p>Three additional nested interfaces are provided to facilitate instantiating concrete
+ * implementations of this interface: {@code EvidenceProvider}, {@code EncryptorProvider}, and
+ * {@code RpcClientProvider}. Implementations of {@code OakClient} first verify the {@code
+ * Evidence} provided by the {@code EvidenceProvider}. Successful verification yields a public key
+ * that will be used by the {@code EncryptorProvider} to generate an {@code Encryptor}. The {@code
+ * RpcClient} provided by the {@code RpcClientProvider} will be used to send and receive encrypted
+ * messages.
  *
  * @param R type of the requests that this client sends
  * @param T type of the responses that this client receives
--- a/java/src/main/java/com/google/oak/client/OakClient.java
+++ b/java/src/main/java/com/google/oak/client/OakClient.java
@@ -20,9 +20,9 @@ import com.google.oak.util.Result;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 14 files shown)


java/src/main/java/com/google/oak/client/OakClient.java
```diff
@@ -16,13 +16,13 @@
...
 
+import com.google.oak.evidence.Evidence;
 import com.google.oak.util.Result;
-import java.util.Optional;
 
 /**
- * Abstract client for sending and receiving encrypted messages to a server. This class provides an
- * abstract builder that must be implemented by concrete subclasses of this class. The builder takes
- * a {@code RpcClientProvider}, an {@code EncryptorProvider}, and optionally an {@code
+ * Abstract client for sending and receiving encrypted messages to a server. We recommend concrete
```
conradgrobler, line 23: nit: perhaps this comment needs a bit more updating since it is now an interface, not an abstract class.

java/src/main/java/com/google/oak/client/amd/AmdAttestationReport.java
```diff
@@ -0,0 +1,67 @@
...
+// TODO(#2842): This is a placeholder implementation, and must be replaced with a valid and
+// complete implementation.
+/**
+ * Placeholder implementation of the remote attestation report from AMD SEV-SNP.
+ */
+public class AmdAttestationReport implements AttestationReport {
+  // TODO(#2842): in reality, it should be extracte from the report.
+  private final byte[] publicKey;
+
+  // TODO(#2842): in reality, it should be extracte from the report.
```
conradgrobler, line 31: nit: "extracted".

java/src/main/java/com/google/oak/client/amd/AmdAttestationReport.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.oak.util.Result;
+
+// TODO(#2842): This is a placeholder implementation, and must be replaced with a valid and
+// complete implementation.
+/**
+ * Placeholder implementation of the remote attestation report from AMD SEV-SNP.
+ */
+public class AmdAttestationReport implements AttestationReport {
+  // TODO(#2842): in reality, it should be extracte from the report.
+  private final byte[] publicKey;
```
conradgrobler, line 29: This would not be extractable from the report, as the report will probably only contain a hash of the public key.

java/src/main/java/com/google/oak/client/amd/AmdAttestedOakClient.java
```diff
@@ -0,0 +1,166 @@
...
+import com.google.oak.util.Result;
+import java.security.GeneralSecurityException;
+
+/**
+ * Placeholder implementation of an AmdAttestedOakClient. To create an instance of this class, an
+ * {@code AmdAttestationReport} must be received and verified.
+ *
+ * This implementation uses {@code byte[]} as the type of the request and response.
+ */
+public class AmdAttestedOakClient<C extends RpcClient> implements OakClient<byte[], byte[]> {
```
conradgrobler, line 38: I don't really have a better suggestion for a name, so I think this is fine for now, but to be pedantic: "attested client" is not really 100% accurate, as the server provides the attestation evidence and the client verifies it. It is the server state that is attested to.

java/src/main/java/com/google/oak/evidence/AttestationReport.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public interface AttestationReport extends Evidence {
+  /**
+   * Extracts and returns the public key part of the server's signing key included in this
+   * attestation report. The client code is responsible for verifying this report before trusting
+   * the public key.
+   *
+   * @return public key part of the remote server's signing key
+   */
+  Result<byte[], Exception> getServerSigningPublicKey();
```
conradgrobler, line 33: I don't think this will be included in the attestation report. A more accurate structure is probably PublicKeyInfo that contains the public key, the attestation report and other supporting evidence.

conradgrobler (approved):
Thanks

java/src/main/java/com/google/oak/client/OakClient.java
mariaschett, line 23: What do we mean by "recommend"? Is there another option? Similar, the builders "should use". Again, is there another option?

java/src/main/java/com/google/oak/client/amd/AmdAttestationReport.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.oak.evidence.AttestationReport;
+import com.google.oak.util.Result;
+
+// TODO(#2842): This is a placeholder implementation, and must be replaced with a valid and
+// complete implementation.
+/**
+ * Placeholder implementation of the remote attestation report from AMD SEV-SNP.
+ */
+public class AmdAttestationReport implements AttestationReport {
+  // TODO(#2842): in reality, it should be extracte from the report.
```
mariaschett, line 28: Maybe clarify what we mean by reality---"When placeholder implementation is replaced, .."?  Also below.

extracte -> extracteD

mariaschett (approved):
Great, thanks!

java/src/main/java/com/google/oak/client/OakClient.java
rbehjati: Thanks. Updated the description.

java/src/main/java/com/google/oak/client/amd/AmdAttestationReport.java
rbehjati: Reworded the sentence.

java/src/main/java/com/google/oak/client/amd/AmdAttestationReport.java
rbehjati: Thanks. Renamed it to `publicKeyHash`, keeping `byte[]` as the type for now. But I think it is better to use a custom defined type when we know more about the details of the attestation report.

java/src/main/java/com/google/oak/client/amd/AmdAttestedOakClient.java
rbehjati: I don't like this name either, but I could not come up with anything better. I'd be happy to iterate on the name as we fill out the implementation. Added a TODO.

java/src/main/java/com/google/oak/evidence/AttestationReport.java
rbehjati: Changed this to `getServerSigningPublicKeySha256Hash` and added `PublicKeyInfo` as a new field in `BasicEvidence`.

java/src/main/java/com/google/oak/client/OakClient.java
rbehjati: Updated the description.

rbehjati (commented):
Thanks for the reviews.

PR #3152, opened 2022-08-15
Merge provenance generation and reproducibility index workflows

### LINKED ISSUE #2990: Generate provenance and reproducibility index from TOML files
At the moment we have a TOML file that is used to drive the generation of provenance files:

https://github.com/project-oak/oak/blob/3112e7d5909faeed088ef16995b9c808632bdaad/build.toml#L1-L5

This only covers one binary, and is only used to generate provenance information, but the same approach should be used to also generate reproducibility index, and upload artifacts to Ent (which is currently done as part of https://github.com/project-oak/oak/blob/main/.github/workflows/reproducibility.yaml .

Basically, I think when the tooling is more mature, https://github.com/project-oak/oak/blob/main/.github/workflows/reproducibility.yaml should be merged into https://github.com/project-oak/oak/blob/main/.github/workflows/provenance.yaml (parametrized by the TOML files describing each of the binaries that we are interested in tracking provenance / reproducibilty about).

In particular, the current `build.toml` file should probably be renamed to `oak_functions_loader_base.toml`, and additional ones should be added for the other binaries (perhaps in a dedicated folder), and the workflow should perform the same steps for each of them.

cc @rbehjati

rbehjati:
Fixes #2990.

This PR merges the `Build Provenance` and `Build Reproducibility Index` into a single workflow:
- It changes the `Build Reproducibility Index` into a [reusable composite action](https://wallis.dev/blog/composite-github-actions) and calls it from the `Build Provenance` workflow,
- For simplicity, the "Upload to Ent" step is moved to the `Build Provenance` workflow,
- The `scripts/build_reproducibility_index` is updated to accept file paths as input arguments, instead of having them hard-coded,
- The `Build Provenance` is changed to fail if building the binary fails.

**Possible future improvement**: Currently the step for updating the reproducibility index only processes a single path. Ideally, we should be able to update the reproducibility index for any number of files. 

### Earlier attempt using a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)

With `Build Reproducibility Index` as a reusable workflow, it has to be called as a separate **job** after the `build_provenance` job. As a result `Build Reproducibility Index` won't immediately have access to artefacts generated by `build_provenance`. The most common solution is to use [artefact upload and download](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) to share artefacts (i.e., generated binaries) between the two jobs.

This approach works but has a number of drawbacks:
- Using artefact upload/download means that we have to keep the generated binaries around unnecessarily. There are a few solutions to this:
  - We can set very short [retention periods](https://docs.github.com/en/github-ae@latest/actions/managing-workflow-runs/removing-workflow-artifacts#setting-the-retention-period-for-an-artifact) for these artefacts, to make sure they won't use up too much space. 
  - We could use [caching](https://github.com/actions/cache) instead of artefact upload/download, but I was unable to make it work.
- Artefact upload/download does not support full paths (with `.` and `/`) as file names. So the file names that appear in the reproducibility index diff cannot be full paths!


rbehjati:
@tiziano88  What do you think about this? I guess it can be improved a lot with a composite action instead of a reusable workflow (see the PR description). Do you have any thoughts for or against that?

rbehjati:
@tiziano88  I have fixed the problem described above. 

With the change in this PR `build_reproducibility_index` is no longer valid. Can you remove it as a required step?

.github/workflows/provenance.yaml
```diff
@@ -73,3 +76,59 @@ jobs:
...
+
+      # Upload artifacts to Ent, where they will be retained and publicly accessible by their hash.
+      # This only applies to `push` events, since it needs access to a valid Ent API key (which is
+      # not available on forks / PRs).
+      - name: Upload to Ent
+        if: github.event_name == 'push'
+        run: |
+          ent put ${{inputs.binary_path}}
+
+      # Build reproducibility index on `push` events.
```
tiziano88, line 120: My expectations was that we could generate the provenance *and* upload to Ent and generate the repro index from the same build config file, since that has all the information necessary. Otherwise, like here, we have to repeat the paths of the built files. Something like another workflow that takes the same build config file as input, then it first runs the provenance generation with the workflow you just added, and then keeps going by uploading the artifact to ent and printing out a provenance diff. I think it should also be possible to control this with an additional input (e.g. `upload_artifacts` and `print_diff` could be added to the workload, and those steps only executed if they are set to true).

Also, my intention is to eventually actually delete the reproducibility index file and corresponding branch, since all the information can be reconstructed from the provenance branch, right? The provenance information is a strict superset of the reproducibiliy information, and anyone should be able to derive something like repro index just from the list of provenances. For instance, from a git commit, we can look up the parent commit id, and then use the provenance branch to find out what the hash of the generated code was at that commit. Then basically we can just show a diff from the previous provenance and the new provenance (perhaps removing noisy fields like timestamps). WDYT?

.github/actions/reproducibility_index/action.yml
```diff
@@ -0,0 +1,108 @@
+name: Build Reproducibility Index
+description: 'Updates the reproducibility index for the given path'
+inputs:
+  binary_path:
```
tiziano88, line 4: Is there a convention for these names? We have two here and they are in different capitalizations already

tiziano88 (approved):

.github/workflows/provenance.yaml
rbehjati: > My expectations was that we could generate the provenance and upload to Ent and generate the repro index from the same build config file, since that has all the information necessary. Otherwise, like here, we have to repeat the paths of the built files. Something like another workflow that takes the same build config file as input, then it first runs the provenance generation with the workflow you just added, and then keeps going by uploading the artifact to ent and printing out a provenance diff. I think it should also be possible to control this with an additional input (e.g. upload_artifacts and print_diff could be added to the workload, and those steps only executed if they are set to true).

I am not entirely sure I understand the approach you are suggesting. Here I am trying to avoid running the build step multiple times, because it takes time and consumes resources. Ideally, the upload to ent and reproducibility index steps could extract the `output_path` from all the files in the `buildconfigs` folder and do the operations for all those files, but currently there are two problems that prevent implementing this solution:
- The build commands generate files that are not included in the build config (e.g., the oak_functions_loader_unsafe)
- There are config files that cannot be built (i.e., the crossvm binary because the build command for it is removed in #3099).

If we can cleanup the builds and fix these issues these workflows will be a lot nicer. 
 
> Also, my intention is to eventually actually delete the reproducibility index file and corresponding branch, since all the information can be reconstructed from the provenance branch, right? The provenance information is a strict superset of the reproducibiliy information, and anyone should be able to derive something like repro index just from the list of provenances. For instance, from a git commit, we can look up the parent commit id, and then use the provenance branch to find out what the hash of the generated code was at that commit. Then basically we can just show a diff from the previous provenance and the new provenance (perhaps removing noisy fields like timestamps). WDYT?

Yes. This is what we eventually want. But do you still want the diff to be published as a comment on the PR? If so, that should be done in a separate PR. If you think the diff should be provided in a separate tool, then I can just remove the reproducibility index in this PR, and we'll build the tool later.

.github/workflows/provenance.yaml
```diff
@@ -73,3 +76,66 @@ jobs:
...
+        if: github.event_name != 'push'
+        run: |
+          sha256sum ${{inputs.binary_path}}
+
+      # Build reproducibility index on `push` events.
+      - name: Build reproducibility index
+        if: github.event_name == 'push'
+        uses: ./.github/actions/reproducibility_index
+        with:
+          binary_path: ${{ env.FUNCTIONS_BASE_PATH }}
```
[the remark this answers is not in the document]
rbehjati: IIUC, GitHub Actions does not support arrays as input, so we should either declare an input variable for each path, or merge all paths into one string on the caller, and then split that string into individual paths here. 

For now, I've added a new variable for the oak_functions_loader_unsafe.

rbehjati (commented):
Thanks for the review.

.github/workflows/provenance.yaml
tiziano88: > I am not entirely sure I understand the approach you are suggesting. Here I am trying to avoid running the build step multiple times, because it takes time and consumes resources. Ideally, the upload to ent and reproducibility index steps could extract the `output_path` from all the files in the `buildconfigs` folder and do the operations for all those files, but currently there are two problems that prevent implementing this solution:

I was thinking to leave that as an optimisation for later on; for now I would just create an action that can build and upload the result of a single build config file, and run that against all the build config files, one after the other, even if it means that it will repeat some work across them (if that's the case, it seems something that the build process should solve, e.g. by improving its own caching). I don't think we should have another way that uses the build config to build more stuff than actually specified in the target of that build config, and then expect to upload all of them together -- if we really need such a use case then we should design the build config to allow for that, e.g. by specifying multiple outputs, and creating provenance for each of them separately (although I personally prefer the simplicity of just one target per build file, if it's possible and practical). Otherwise this seems to show that we designed a format that does not cover our own needs and we need to work around it, which means other projects will also probably run into similar issues, and we should not require that they also have extra logic that repeats some of the information again in the GitHub actions file. 

Ideally the maintainer of a project that wants to opt in transparent release should just add all the build config files for all their targets, and mechanically add them to the GitHub action that we provide to them.

> * The build commands generate files that are not included in the build config (e.g., the oak_functions_loader_unsafe)

I think we should add a build config that builds that too, right? In fact we should add build configs for all the things we care about building as stand-alone binaries. I imagine build configs similar to top level makefile targets: they define everything necessary to build a given binary. Then, from the GitHub actions config, we just need to invoke the builder action for each of them. Anything else in the GitHub actions config should basically just be running tests / examples.

> * There are config files that cannot be built (i.e., the crossvm binary because the build command for it is removed in [Remove unused xtask options #3099](https://github.com/project-oak/oak/pull/3099)).

Yes that should be reverted, and tests added so that it is not possible to accidentally remove it again -- would adding the build config file and pointing to that from the GitHub actions file enough to make sure the build will fail in that case, so it cannot happen in the future?

> Yes. This is what we eventually want. But do you still want the diff to be published as a comment on the PR? If so, that should be done in a separate PR. If you think the diff should be provided in a separate tool, then I can just remove the reproducibility index in this PR, and we'll build the tool later.

Yes I think eventually we should have the tool (or some other way of looking up a diff between two provenances), and also run that tool to generate the diff from the previous commit, and post that as a comment. The tool may be super simple, for instance given a commit id, it may just download the correct file from the provenance branch, and print it out (then in the GH action we can diff the previous and the new one ourselves).

rbehjati: > Ideally the maintainer of a project that wants to opt in transparent release should just add all the build config files for all their targets, and mechanically add them to the GitHub action that we provide to them.

This is true. The problems discussed above are not due to the transparent release, but are due to suboptimal build commands on Oak (the maintainer of the project that has opt into transparent release), and should be fixed by Oak. 



> > * There are config files that cannot be built (i.e., the crossvm binary because the build command for it is removed in [Remove unused xtask options #3099](https://github.com/project-oak/oak/pull/3099)).
> 
> Yes that should be reverted, and tests added so that it is not possible to accidentally remove it again -- would adding the build config file and pointing to that from the GitHub actions file enough to make sure the build will fail in that case, so it cannot happen in the future?

Note that GitHub action already pointed to the config using the removed command. The action was not failing, intentionally, to avoid any potential flakiness in the provenance generation to block merges. The tooling seems stable now, and this PR makes the GitHub action fail. Once #3099 is reverted and the related TODO here is resolved, the workflow will fail and prevent this from happening in future.

PR #3153, opened 2022-08-16
Create a simple hello world kernel for future SEV-SNP enlightenment

andrisaar:
This is a basic stand-alone skeleton based on code from our normal code that just prints "Hello World" and shuts down the VM.

The idea is to use this as a basis for a test kernel that will verify that SEV-SNP has been enabled properly for use in tests.

Hence, I decided to _not_ add it to already overflowing `experimental` but rather create a new directory, following the general naming scheme what is in use internally. Feel free to bikeshed about the location :)


testdata/sev_snp_hello_world_kernel/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+    serial::init_logging();
+    log::info!("Hello World!");
+    panic!();
+}
+
+#[panic_handler]
+fn panic(_info: &PanicInfo) -> ! {
+    unsafe {
+        let mut port = PortWriteOnly::new(0x64);
+        port.write(0xFE_u8);
```
ipetr0v, line 37: What does this code do?

ipetr0v (approved):

testdata/sev_snp_hello_world_kernel/src/asm/boot.s
```diff
@@ -0,0 +1,27 @@
...
+
+.section .boot, "ax"
+.global _start
+.code64
+
+_start:
+    mov $stack_start, %rsp
+    # Push 8 bytes to fix stack alignment issue. Because we enter rust64_start
+    # with a jmp rather than a call the function prologue means that the stack
+    # is no loger 16-byte aligned.
```
conradgrobler, line 25: "no longer"

conradgrobler (approved):
Some bikeshedding about the location: I agree with moving out of experimental, but "testdata" as a root for code feels weird, and might cause someone browsing the repo to overlook this.

testdata/sev_snp_hello_world_kernel/src/main.rs
andrisaar: Shuts down the machine. I've added a comment about that.

testdata/sev_snp_hello_world_kernel/src/asm/boot.s
andrisaar: Done.

andrisaar:
> Some bikeshedding about the location: I agree with moving out of experimental, but "testdata" as a root for code feels weird, and might cause someone browsing the repo to overlook this.

I'm all ears if you have better suggestions :)

conradgrobler:
> I'm all ears if you have better suggestions :)

I am not sure about better suggestions, but perhaps "testing" rather than "testdata".

PR #3156, opened 2022-08-16
Excise `anyhow` from the `sev_guest` crate.

andrisaar:
`anyhow` uses the heap, so it requires a global allocator. We don't have one in the SEV-SNP hello world kernel.

This does the laziest thing of replacing the error with a `&'static str`; it worked for all but one location where unfortunately we will now lose the error code.

We should really introduce proper error types for this, but for now, this will let us move forward.


conradgrobler (approved):
Please create an issue and add a todo to the file.

PR #3157, opened 2022-08-17
Expose cpuid and secrets data structures in the helloworld kernel.

andrisaar:


testing/sev_snp_hello_world_kernel/src/main.rs
```diff
@@ -17,16 +17,30 @@
...
+#[link_section = ".secrets"]
+static SECRETS: MaybeUninit<SecretsPage> = MaybeUninit::uninit();
+
 #[no_mangle]
 pub extern "C" fn rust64_start() -> ! {
     serial::init_logging();
     log::info!("Hello World!");
+
+    #[allow(dead_code)]
+    let _cpuid: &CpuidPage = unsafe { CPUID.assume_init_ref() };
```
conradgrobler, line 39: Please add some comments on why this unsafe code is ok.

conradgrobler (approved):
Thanks

testing/sev_snp_hello_world_kernel/src/main.rs
tiziano88, line 39: Why does it need to be assigned to a local variable? Does it get optimized away otherwise?

tiziano88 (approved):

testing/sev_snp_hello_world_kernel/src/main.rs
andrisaar: Yep, Rust optimized away the whole static, much to my surprise -- even if I tagged it with `#[used]`.

However, I've added some statements logging the contents of these pages (as we're interested what's in them anyway), so these are no longer unused.

andrisaar: Done.

jul-sh (approved):

Issue #2986, opened 2022-06-20
Set up a stack guard page


andrisaar:
In our current memory layout stack is placed right after `.bss`, and as we haven't fixed #2983 yet the memory is writable.

Therefore, if you're not careful, you can overflow the stack and start overwriting other data structures (up to and including code in `.text`!)

The common solution is to set up a stack guard page just below the stack: a page which is marked as nonexistent, so that if anything tries to access virtual addresses on that page, we get a fault.

Relevant topic on the OSDev forum: https://forum.osdev.org/viewtopic.php?t=23661

andrisaar:
I think we fixed that a while ago.

Issue #2984, opened 2022-06-20
Move the kernel to the higher half of virtual address space


andrisaar:
Our kernel right now expects identity mapping between physical and virtual addresses, and thus resides in the lower half of the canonical address space (see https://en.wikipedia.org/wiki/X86-64#Virtual_address_space_details for more details).

Traditionally kernels live in the upper half, leaving the lower half for userspace. We should move our code to the upper half as well.

Estimate of things to do:
- [ ] change the linker script to set VirtAddrs to the upper half (note that PhysAddr will stay the same!)
- [ ] when setting up initial paging, map pages containing kernel code to their correct virtual addresses instead of using identity mapping

A good guide as to what the memory layout should be can be found in the Linux kernel documentation:
https://docs.kernel.org/x86/x86_64/mm.html#complete-virtual-memory-map-with-4-level-page-tables

and if you examine `vmlinux` using `readelf`, here's some interesting bits:
```
ELF Header:
 Entry point address:               0x1000000

Section Headers:
  [ 1] .text             PROGBITS        ffffffff81000000 200000 c01e40 00  AX  0   0 4096
  [ 2] .rodata           PROGBITS        ffffffff81e00000 1000000 404ff7 00  WA  0   0 4096

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x200000 0xffffffff81000000 0x0000000001000000 0x1644220 0x1644220 R E 0x200000
```
Note that code with virtual addresses in the higher half are still loaded to physaddr of zero.

Issue #2983, opened 2022-06-20
Set up proper page permissions (in particular, N^X bit)


andrisaar:
Our current page table setup code, inherited from `rust-hypervisor-firmware`, sets up identity paging for the lower 4G of memory and marks every page as `WRITABLE`:

https://github.com/project-oak/oak/blob/fd3ca98c97bb7b7129ecf7dd01f53e023fc76dd2/third_party/rust-hypervisor-firmware-boot/src/paging.rs#L26

This is obviously less than optimal, as this means that all the memory is both writable and executable (and thus fun things can happen when you overflow the stack, for example). We don't really have any use cases where we want pages to be both writable and executable.

We should ensure that:
- [ ] pages that contain executable code are _not_ marked as `WRITABLE`;
- [ ] the rest of the pages may be marked as `WRITABLE`, but must be marked as `NO_EXECUTE`.

PR #3159, opened 2022-08-18
Map last two gigabytes of virtual address space to physical 0x0

andrisaar:
Follow-up to #3158 -- we're rewriting page tables after the kernel starts, so we need to make sure we don't pull the rug from under our feet.

This code is temporary. In the not-too-distant-future I want to replace the whole `rust_hypervisor_firmware_boot::paging` code that just sets up identity mapping with more intelligent page table construction that pays attention to where the different ELF sections should go, and what permissions those regions should get (see #2983 and #2986). But for now, this hack allows us to progress on #2984.

Steps needed to load the kernel to the upper half:
- [x]  map physical 0x0 to virtual -2G during initial boot
- [x]  keep the -2G mapping when setting up page tables in `rust_hypervisor_firmware_boot::paging::init()`
- [ ]  change the kernel virtual address to -2G in `layout.ld`

### DIFF
--- a/third_party/rust-hypervisor-firmware-boot/src/paging.rs
+++ b/third_party/rust-hypervisor-firmware-boot/src/paging.rs
@@ -39,8 +39,18 @@ pub fn setup() {
         l3[i].set_addr(phys_addr(l2), pt_flags);
     }
 
+    // Upper half hack: we point the last two entries of the L3 table to the same L2 tables as the
+    // first two entries, and then we point to the last entry of the L4 table to the same L3 table
+    // as the first entry in the L4 table. This means we get an extra identity-mapped region at
+    // -2GB of virtual memory, where the kernel will live.
+    let addr_0 = l3[0].addr();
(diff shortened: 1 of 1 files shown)


third_party/rust-hypervisor-firmware-boot/src/paging.rs
```diff
@@ -38,9 +38,15 @@ pub fn setup() {
     for (i, l2) in l2s.iter().enumerate() {
         l3[i].set_addr(phys_addr(l2), pt_flags);
     }
+    // Upper half hack.
```
conradgrobler, line 41: Perhaps add a bit more a description for why this works for future reference.

conradgrobler (approved):

third_party/rust-hypervisor-firmware-boot/src/paging.rs
andrisaar: Done.

PR #3160, opened 2022-08-18
Move the kernel to the upper half of VMA

### LINKED ISSUE #2984: Move the kernel to the higher half of virtual address space

andrisaar:
This builds on #3158 and #3159 by actually setting the kernel virtual addresses to the kernel half of the address space.
Unfortunately Github doesn't have the concept of "this PR is dependent on those two", so this change contains changes from both of these. The core change here is the two-line change in `layout.ld`:

https://github.com/project-oak/oak/blob/fe8a7f0aecc3bf319ba7a3045155fb52a5d88783/experimental/oak_baremetal_app_crosvm/layout.ld#L40-41

Closes #2984.

Steps needed to load the kernel to the upper half:
- [x]  map physical 0x0 to virtual -2G during initial boot
- [x]  keep the -2G mapping when setting up page tables in `rust_hypervisor_firmware_boot::paging::init()`
- [x]  change the kernel virtual address to -2G in `layout.ld`

### DIFF
--- a/experimental/oak_baremetal_app_crosvm/layout.ld
+++ b/experimental/oak_baremetal_app_crosvm/layout.ld
@@ -37,8 +37,8 @@ PHDRS
     secrets PT_LOAD FLAGS(1 << 24);
 }
 
-/* To set kernel VMA to high memory, set this to 0xFFFFFFFF80000000. */
-KERNEL_MEM_START = 0x0;
+/* To set kernel VMA to low memory, set this to 0x0. */
+KERNEL_MEM_START = 0xFFFFFFFF80000000;
 
 SECTIONS {
     /*
--- a/experimental/oak_baremetal_kernel/src/lib.rs
+++ b/experimental/oak_baremetal_kernel/src/lib.rs
@@ -65,6 +65,7 @@ pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: B) -> ! {
     logging::init_logging();
     interrupts::init_idt();
     paging::setup();
+    log::info!("here");
     // We need to be done with the boot info struct before intializing memory. For example, the
     // multiboot protocol explicitly states data can be placed anywhere in memory; therefore, it's
     // highly likely we will overwrite some data after we initialize the heap. args::init_args()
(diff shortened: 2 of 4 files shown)


conradgrobler (approved):
I am surprised this works. Is the heap still in lower memory and identity mapped? I would have expected the io drivers to fail, as both assume an identity mapping when communicating the addresses of the shared memory pages with the hypervisor/virtual device.

andrisaar:
> I am surprised this works. Is the heap still in lower memory and identity mapped? I would have expected the io drivers to fail, as both assume an identity mapping when communicating the addresses of the shared memory pages with the hypervisor/virtual device.

Yes, the heap is still in the lower half and identity mapped. Eventually the kernel heap will have to move somewhere in the upper half as well, but I decided to postpone shaving that particular yak for now.

PR #3167, opened 2022-08-22
Remove unused methods.

qwandor:
Memory in a MemoryRegion should only be accessed via volatile methods, as it is used for memory shared with the host or used for MMIO. The other methods are not used, but it's better to remove them entirely to prevent them from mistakenly being used in future.


conradgrobler (approved):
Thanks.

PR #3168, opened 2022-08-22
Create a bitmap allocator for physical frames

andrisaar:
This is the first step toward better tracking of the physical memory that we have available in the kernel.

What is implemented here is a frame allocator, limited to fixed frame sizes and memory range(s). While the algorithm is... unsophisticated, the key feature of the frame allocator is that it doesn't do dynamic memory allocation itself, as the frame allocator is initialized before the heap (in fact, the heap allocator will have to ask for memory to manage from the frame allocator).

Rough outline of what needs to be done:
- [x] implement a base bitmap allocator
- [ ] implement a real frame allocator, capable of handling requests for 2 MiB and 4 KiB frames, for ~128G of memory
- [ ] initialize the frame allocator when the kernel starts up based on data in the E820 map (and the ELF header)
- [ ] change the heap initialization code to request memory from the frame allocator instead of relying on constants in the linker script

### DIFF
--- a/experimental/oak_baremetal_kernel/src/mm/bitmap_frame_allocator.rs
+++ b/experimental/oak_baremetal_kernel/src/mm/bitmap_frame_allocator.rs
@@ -1,0 +1,347 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use bitvec::{order::Lsb0, prelude::BitArray};
+use core::{
+    ops::{BitAnd, Not},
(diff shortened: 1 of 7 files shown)


experimental/oak_baremetal_kernel/src/mm/bitmap_frame_allocator.rs
```diff
@@ -0,0 +1,336 @@
...
+            panic!(
+                "Can't mark validity for frame range that's outside our range; our range: {:?}, frame range: {:?}",
+                self.range, range
+            );
+        }
+    }
+
+    /// Returns the largest contiguous section of unallocated memory.
+    #[allow(dead_code)]
+    pub fn largest_available(&self) -> Option<PhysFrameRangeInclusive<S>> {
```
conradgrobler, line 91: I am not sure whether this is a particularly useful function, as I think we would likely want to know the first location where a range or a specific size can fit, rather than overall the largest one.

conradgrobler (approved):

experimental/oak_baremetal_kernel/src/mm/bitmap_frame_allocator.rs
andrisaar: Longer-term, you are correct: we probably need something that will allow us to allocate a contiguous range of physical memory, and won't care about its exact location.

I implemented this method to be compatible with our existing heap allocation mechanism, where we just take the biggest slice of memory available and use it as the heap. In the future we won't want to give _all_ the memory to the kernel heap, at which point we can get rid of this function.

PR #3170, opened 2022-08-24
Generalize proxy of gRPC request

mariaschett:


experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -88,33 +131,63 @@ fn trusted_shuffler_to_hyper_request(
...
     let body = Body::from(trusted_shuffler_request.body);
-    Request::builder()
+
+    let mut request = Request::builder()
         .uri(trusted_shuffler_request.uri)
         .method(Method::POST)
         .version(http::Version::HTTP_2)
         .body(body)
-        .expect("Failed to convert Trusted to Hyper Request")
+        .expect("Failed to convert Trusted to Hyper Request");
```
ipetr0v, line 140: `expect` crashes the whole program if it fails
I think it's better to use `.context()` and return `anyhow::Result`

ipetr0v (approved):

experimental/trusted_shuffler/scripts/run_trusted_shuffler
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 
-RUST_LOG=info cargo run --manifest-path=experimental/trusted_shuffler/server/Cargo.toml -- \
---k=1 --listen-address="[::]:8080" --backend-url="http://localhost:8888" 
-   
-# Convenient to know the PID to kill when the Trusted Shuffler crashes.
-echo "Trusted Shuffler PID: "
-echo $!
+RUST_LOG=debug cargo run --manifest-path=experimental/trusted_shuffler/server/Cargo.toml -- \
+--k=1 --listen-address="[::]:8080" --backend-url="http://localhost:8888"
```
tiziano88, line 4: I think single quotes are sufficient, there is no interpolation going on here.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -59,7 +59,27 @@ impl RequestHandler for HttpRequestHandler {
...
+            Client::builder()
+                .http2_only(true)
+                .build(AlpnConnector::new())
+                .request(request)
+        } else {
+            Client::builder()
+                .http2_only(true)
+                .build_http()
+                .request(request)
+        };
```
tiziano88, line 80: I think this would look cleaner if it was conditionally building the client first and then sending the request.

```rust
let mut client_builder = Client::builder().http2_only(true);

let client = if is_https_url { client_builder.build(AlpnConnector::new()) } else { client_builder.build_http() };

let response = client.request(request);
```

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -76,8 +96,31 @@ async fn hyper_to_trusted_shuffler_request(hyper_request: Request<Body>) -> Trus
     let body = hyper::body::to_bytes(body)
         .await
         .expect("Couldn't read request body");
+    log::info!("Body {:?}", body);
+
+    let hyper_headers = parts.headers;
+
+    // Keep the minimal headers the backend requires to handle the request.
+    let mut minimal_headers = hyper::HeaderMap::new();
```
tiziano88, line 104: I still think we should just copy all the headers and the trailers eventually.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -88,33 +131,63 @@ fn trusted_shuffler_to_hyper_request(
...
+
+    let mut request = Request::builder()
         .uri(trusted_shuffler_request.uri)
         .method(Method::POST)
         .version(http::Version::HTTP_2)
         .body(body)
-        .expect("Failed to convert Trusted to Hyper Request")
+        .expect("Failed to convert Trusted to Hyper Request");
+
+    let headers = request.headers_mut();
```
tiziano88, line 142: I'm surprised this does not need to be mut, is that because it is used by value below?

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -48,12 +49,16 @@ impl TrustedShufflerRequest {
 
 #[derive(Debug, PartialEq)]
 pub struct TrustedShufflerResponse {
-    pub body: Vec<u8>,
+    pub data: hyper::body::Bytes,
+    pub trailers: hyper::HeaderMap,
 }
 
 impl TrustedShufflerResponse {
     fn empty() -> TrustedShufflerResponse {
```
tiziano88, line 57: Isn't this basically Default? If so, can it be derived automatically?

xtask/src/examples.rs
```diff
@@ -484,40 +484,22 @@ fn run_oak_functions_server(server: &Server) -> Box<dyn Runnable> {
...
     let backend_cmd = Cmd::new(
         "cargo",
-        vec![
+        spread![
             "run",
             "--manifest-path=./experimental/trusted_shuffler/backend/Cargo.toml",
             "--",
-            "--listen-address=[::]:8888",
-            "--use-grpc",
+            "--listen-address=[::]:8888", ...flags
```
tiziano88, line 495: `...flags` should be on the next line

xtask/src/examples.rs
```diff
@@ -528,19 +510,16 @@ pub fn run_trusted_shuffler() -> Step {
...
+        spread![
             "run",
             "--manifest-path=experimental/trusted_shuffler/client/Cargo.toml",
             "--",
             "--server-url=http://localhost:8080",
             "--qps=1",
-            "--seconds=1",
-            "--use-grpc",
+            "--seconds=1", ...
+            flags
```
tiziano88, line 522:
```suggestion
            "--seconds=1",
            ...flags
```

tiziano88 (approved):
Thanks!

experimental/trusted_shuffler/scripts/run_trusted_shuffler
mariaschett: done.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -59,7 +59,27 @@ impl RequestHandler for HttpRequestHandler {
 
         log::info!("New request to the backend: {:?}", request);
 
-        match send_with_request(request).await {
+        // If the backend_url is a https connection, we need to use the AlpnConnector for the
+        // client connecting to the backend.
+        let is_https_url: bool = request
+            .uri()
+            .scheme()
+            .expect("Could not get scheme from backend_url.")
```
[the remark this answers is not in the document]
mariaschett: Done.

experimental/trusted_shuffler/server/src/http.rs
mariaschett: I wish, yes, but because the `client_builder` is either `Client<AlpnConnector, {unknown}>` or `Client<HttpConnector, {unknown}>` and wasn't could not to convince the type checker it is fine. 

I also figured we'd want to revisit that anyway: It might be better to do that check and choice only when starting the Trusted Shuffler, not for every request.

experimental/trusted_shuffler/server/src/http.rs
```diff
@@ -76,8 +96,31 @@ async fn hyper_to_trusted_shuffler_request(hyper_request: Request<Body>) -> Trus
     let body = hyper::body::to_bytes(body)
         .await
         .expect("Couldn't read request body");
+    log::info!("Body {:?}", body);
```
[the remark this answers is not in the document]
mariaschett: haha, absolutely! I did enjoy these huge bodies though!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: I think we want to have control on what we copy to show we don't have identifying information in there. So I'd argue not even for a `HeaderMap` in `TrustedShufflerRequest` but even more fine grained control. But let's keep thinking on this!

experimental/trusted_shuffler/server/src/http.rs
mariaschett: I think it is fine: `headers: &mut HeaderMap`.

experimental/trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Ehm, good point, but I'd like to keep it that way. When we have fixed size responses, the empty response should have a fixed size, too.

xtask/src/examples.rs
mariaschett: done.

xtask/src/examples.rs
mariaschett: done.

mariaschett (commented):
Thank you for the reviews!

Issue #3096, opened 2022-07-27
Fetch the Signing Public Key before doing the remote attestation handshake


conradgrobler:
The public key is currently sent as part of the remote attestation handshake. We should fetch the key before the handhake starts (via a separate call to the untrusted launcher) and validate the public key before starting the handshake.

This means the we need to make the key available to the untrusted launcher before serving requests (e.g. add signing public key to the response from the call to `Initialize`).

The public key should be presented with all the required evidence to be able to validate it, including:

- Remote attestation report that includes the hash of the signing public key
- Certificate chain for validating the attestation report
- Appropriate artefacts from the transparent release process

All of the evidence must be independently verifiable by the client, as it will be provided by the untrusted launcher.

tiziano88:
cc @sidtelang

conradgrobler:
We are no longer doing the interactive attestation between client and server. Fetching the public key, attestation report and other info first will naturally happen as part of the interactive attestation implementation.

PR #3171, opened 2022-08-24
Make remote attestation public key and attestation information available outside the handshake

conradgrobler:
This change makes information about the attestation public key available to the client via a separate RPC method. This is the first step towards fixing #3096

### DIFF
--- a/experimental/oak_baremetal_runtime/schema.fbs
+++ b/experimental/oak_baremetal_runtime/schema.fbs
@@ -2,7 +2,7 @@ attribute "method_id";
 
 rpc_service TrustedRuntime {
     HandleUserRequest(UserRequest) : UserRequestResponse (method_id: 0);
-    Initialize(Initialization) : Empty (method_id: 1);
+    Initialize(Initialization) : InitializeResponse (method_id: 1);
     // Replaces the existing lookup data with the new value.
     UpdateLookupData(LookupData) : Empty (method_id: 2);
 }
@@ -23,6 +23,15 @@ table UserRequestResponse {
 table Initialization {
     wasm_module: [ubyte];
     constant_response_size: uint;
+}
+
+table InitializeResponse {
+    public_key_info: PublicKeyInfo;
+}
+
+table PublicKeyInfo {
+    public_key: [ubyte];
+    attestation: [ubyte];
 }
 
 table LookupDataEntry {
--- a/grpc_unary_attestation/src/server.rs
+++ b/grpc_unary_attestation/src/server.rs
@@ -17,12 +17,21 @@
 //! Server-side implementation of the bidirectional gRPC remote attestation handshake
 //! protocol.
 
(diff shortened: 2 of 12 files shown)


tiziano88:
I'm not sure adding a separate method would work for the gRPC streaming use case (i.e. with no stickiness(; I was thinking we would just piggy back on top of the single stream, adding a couple of initial messages that only terminate in the untrusted runtime. Perhaps a bit in the first message sent by the client can determine whether it is a message to request the pk info from the untrusted runtime, or the first message of the actual handshake with the trusted runtime. WDYT?

conradgrobler:
Good point. This only works for the unary request model with stickiness, so we would need to think about it more if we want to reimplement the streaming model.

I don't like the idea of the untrusted part intercepting some parts of a bidirectional stream and transparently proxying others, as it seems like it will be potentially errorprone. I don't have a good alternative to suggest though.

grpc_unary_attestation/src/server.rs
```diff
@@ -37,6 +46,8 @@ pub struct AttestationServer<F, L: LogError> {
     error_logger: L,
     session_tracker:
         Mutex<SessionTracker<PlaceholderAmdAttestationGenerator, EmptyAttestationVerifier>>,
+    signing_public_key: Vec<u8>,
```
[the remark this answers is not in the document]
conradgrobler: This key is used to sign the transcript and proves the server identity. The ephemeral DH key is independent from the signing key.

experimental/oak_baremetal_runtime/schema.fbs
```diff
@@ -25,6 +25,15 @@ table Initialization {
     constant_response_size: uint;
 }
 
+table InitializeResponse {
+    public_key_info: PublicKeyInfo;
+}
+
+table PublicKeyInfo {
+    public_key: [ubyte];
```
tiziano88, line 33: I think at some point we need to find a proper name for this key. Currently it's called public_key or signing_public_key in different places, neither of which is particularly descriptive. Perhaps root_public_key or instance_public_key or runtime_public_key or something sufficiently unambiguous would help (here, in code, and in documentation).

tiziano88 (approved):

experimental/oak_baremetal_runtime/schema.fbs
conradgrobler: The naming depends on the usage. For interactive remote attestation it will be a signing key, but for the offline attestation it will be an encryption (or DH) key. I created #3181 to track public key naming.

ipetr0v:
We can also continue sending the Signing key as part of the handshake.
And just update the handshake protocol itself (internally).

It would work with streaming, and also wouldn't require an additional roundtrip in the handshake.

conradgrobler:
There are three main reasons for moving this out of the handshake:

- Compatibility with the implementation by https://github.com/google/sealedcomputing
- More consistency between the clients for interactive vs offline attestation
- No need to include AMD certificate chain in every handshake

PR #3175, opened 2022-08-25
Implement `Default` for `Ghcb` and add a proper error type for `RegisterGhcbGpaRequest`.

andrisaar:


experimental/sev_guest/src/ghcb.rs
```diff
@@ -91,6 +91,13 @@ pub struct Ghcb {
     pub ghcb_usage: u32,
 }
 
+impl Default for Ghcb {
+    fn default() -> Self {
+        // This is safe because all-zeroes is a valid value of type Ghcb.
+        unsafe { core::mem::zeroed() }
```
conradgrobler, line 97: Optional: because FromBytes is already derived for Ghcb, you could also use `new_zeroed`, althout that is just a wrapper for the exact same code, so probably does not make a difference. (https://docs.rs/zerocopy/latest/src/zerocopy/lib.rs.html#228-237)

conradgrobler (approved):

experimental/sev_guest/src/msr.rs
```diff
@@ -257,30 +258,41 @@ pub struct RegisterGhcbGpaResponse {
     ghcb_gpa: u64,
 }
 
+#[derive(Debug, Snafu)]
+pub enum RegisterGhcbGpaError {
```
mariaschett, line 262: How did we decide on these errors and is there some documentation on when those can happen?

experimental/sev_guest/src/msr.rs
```diff
@@ -257,30 +258,41 @@ pub struct RegisterGhcbGpaResponse {
...
+    InvalidResponse,
+    GhcbLocationNotAccepted,
+    GhcbResponseLocationNotMatchingRequest { addr: u64 },
```
mariaschett, line 265: Out of curiosity: which address do we send back here?

experimental/sev_guest/src/msr.rs
```diff
@@ -257,30 +258,41 @@ pub struct RegisterGhcbGpaResponse {
...
 }
 
 /// Registers the location of the GHCB page for the current vCPU with the hypervisor.
-pub fn register_ghcb_location(request: RegisterGhcbGpaRequest) -> Result<(), &'static str> {
+pub fn register_ghcb_location(request: RegisterGhcbGpaRequest) -> Result<(), RegisterGhcbGpaError> {
     let request_ghcb_gpa: u64 = request.ghcb_gpa;
     write_msr_and_exit(request.into());
     let response: RegisterGhcbGpaResponse = read_msr().try_into()?;
     // Ensure that the registration was successful.
     if response.ghcb_gpa == GHCB_LOCATION_NOT_ACCEPTED {
```
mariaschett, line 287: Why don't we use `ensure` here too?

mariaschett (approved):

experimental/sev_guest/src/ghcb.rs
andrisaar: Ah, I missed that. I think `new_zeroed` is a good idea here as this means we don't have to resort to explicit unsafe code.

experimental/sev_guest/src/msr.rs
andrisaar: I'll let Conrad elaborate; I determined these errors by looking at code and giving enum entries for all different strings :)

experimental/sev_guest/src/msr.rs
andrisaar: It's the address from the response. I've updated the name of the variable to be more descriptive.

experimental/sev_guest/src/msr.rs
andrisaar: Done.

PR #3176, opened 2022-08-26
Zero-initialize BSS properly

andrisaar:
The BSS section contains static variables, and the C (and Rust's) memory model expects the memory there to be zero-initialized.

And it just happens by accident that the memory given to us by VMMs was always zero-initialized, meaning our code works. But we're well into undefined behaviour and it works by sheer luck only.

Once we tried enabling SNP, that property was no longer true (as zero doesn't necessarily decrypt to zero), and all kinds of interesting things started breaking as their internal invariants were horribly broken.

Thus: let's do what we should have always done, and manually reset the memory to zero. And x86 has a special magic instruction, `REP STOSB` to do just that, so we don't even need a loop.

### DIFF
--- a/experimental/oak_baremetal_app_crosvm/src/asm/boot.s
+++ b/experimental/oak_baremetal_app_crosvm/src/asm/boot.s
@@ -49,6 +49,12 @@ _start:
     # Finally, trigger a full TLB flush by overwriting CR3, even if it is the same value.
     movq %rbx, %cr3
 
+    # Clear BSS: base address goes to RDI, value goes to AX, count goes into CX.
+    mov $bss_start, %rdi
+    mov $bss_size, %rcx
+    xor %rax, %rax
+    rep stosb
+
     mov $stack_start, %rsp
     # Push 8 bytes to fix stack alignment issue. Because we enter rust64_start with a jmp rather
     # than a call the function prologue means that the stack is no longer 16-byte aligned.
--- a/testing/sev_snp_hello_world_kernel/src/asm/boot.s
+++ b/testing/sev_snp_hello_world_kernel/src/asm/boot.s
@@ -19,6 +19,12 @@
 .code64
 
 _start:
+    # Clear BSS: base address goes to RDI, value goes to AX, count goes into CX.
+    mov $bss_start, %rdi
+    mov $bss_size, %rcx
+    xor %rax, %rax
+    rep stosb
+
     mov $stack_start, %rsp
     # Push 8 bytes to fix stack alignment issue. Because we enter rust64_start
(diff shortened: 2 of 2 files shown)


conradgrobler (approved):

conradgrobler:
Thinking about this a bit more: how big is the BSS? If it is large enough to have its own page, this might be a problem under AMD SEV-SNP. If the page has not been initialised by the hypervisor before startup and the code tries to access any zero-initialised variables before we have run PVALIDATE on the page it will cause a fault.

ipetr0v:
> Thinking about this a bit more: how big is the BSS? If it is large enough to have its own page, this might be a problem under AMD SEV-SNP. If the page has not been initialised by the hypervisor before startup and the code tries to access any zero-initialised variables before we have run PVALIDATE on the page it will cause a fault.

What does `PVALIDATE` do?

ipetr0v (approved):

conradgrobler:
> What does `PVALIDATE` do?

PVALIDATE is a new instruction in SEV-SNP which allows the guest to mark guest-private memory pages as valid in the reverse map table (RMP). Only the guest and the secure processor (PSP) can change pages from invalid to valid state. If the hypervisor changes the attributes of a page in the RMP the state will change to invalid, which will allow the guest to detect tampering by the hypervisor. If the guest tries to access an invalid page it will generate a page fault.

andrisaar:
> Thinking about this a bit more: how big is the BSS? If it is large enough to have its own page, this might be a problem under AMD SEV-SNP. If the page has not been initialised by the hypervisor before startup and the code tries to access any zero-initialised variables before we have run PVALIDATE on the page it will cause a fault.

The size of the BSS depends on how many statics we have around; I assume it will ever be just a couple of 2 MB pages.

We could put the memory initialization on the VMM side (or the possible future stage0, which would need to do `PVALIDATE` for us), but that would mean we will ask the PSP to encrypt possibly several megabytes of zeroes, which it really doesn't need to.

I don't know what's the lesser evil here; my gut says sticking a `PVALIDATE` in here might be the lesser evil. But I'm not sure.

conradgrobler:
> I don't know what's the lesser evil here; my gut says sticking a `PVALIDATE` in here might be the lesser evil. But I'm not sure.

I don't think it makes a huge difference. The SEV-SNP spec includes support for passing a page type of PAGE_TYPE_ZERO to SNP_LAUNCH_UPDATE, which I guess is exactly for this purpose. Seeing that we will probably move to a stage 0 firware to do this setup work, any solution would only be temporary, so for now we should go with the easiest implementation option.

PR #3187, opened 2022-09-08
Remove Runtime dependency from the Restricted Kernel

ipetr0v:
This PR removes the Oak Runtime dependency from the Restricted Kernel and moves it to the App CrosVM.
This is a first step of allowing running different runtimes inside the Restricted Kernel.


experimental/oak_baremetal_channel/src/server.rs
```diff
@@ -22,15 +22,15 @@ use anyhow::Context;
 /// and responds to them using the provided [`oak_idl::Handler`].
 pub fn start_blocking_server<H: oak_idl::Handler>(
     channel: Box<dyn Channel>,
-    mut oak_idl_service_implementation: H,
+    mut idl_service: H,
```
conradgrobler, line 25: I guess this is fine as an initial step, but I am not sure that we will always use the same IDL for all workloads, so this might have to be based on a different trait in the future.

experimental/oak_baremetal_kernel/Cargo.toml
```diff
@@ -21,17 +21,9 @@ lazy_static = { version = "*", features = ["spin_no_std"] }
 linked_list_allocator = { version = "*" }
 log = "*"
 libm = "*"
-oak_remote_attestation = { path = "../../remote_attestation/rust", default-features = false, features = [
```
conradgrobler, line 24: I think the getrandom dependency above needs to also be removed. I can't remember exactly why it was required, but it might have to be moved to experimental/oak_baremetal_app_crosvm/Cargo.toml

conradgrobler (approved):

experimental/oak_baremetal_app_crosvm/Cargo.toml
```diff
@@ -12,6 +12,16 @@ simple_io_channel = ["oak_baremetal_kernel/simple_io_channel"]
...
 oak_baremetal_kernel = { path = "../../experimental/oak_baremetal_kernel", default-features = false }
+oak_baremetal_runtime = { path = "../oak_baremetal_runtime", default-features = false, features = [
+  "rust-crypto"
+] }
+oak_remote_attestation = { path = "../../remote_attestation/rust", default-features = false, features = [
+  "rust-crypto"
+] }
+oak_remote_attestation_amd = { path = "../../oak_remote_attestation_amd", default-features = false, features = [
+  "rust-crypto"
+] }
```
tiziano88, line 23: Is there any reason not to flip these defaults the other way round? @jul-sh IIRC you were looking into this or something related at some point, WDYT?

tiziano88 (approved):

experimental/oak_baremetal_kernel/src/lib.rs
```diff
@@ -91,18 +92,12 @@ enum ChannelType {
     SimpleIo,
 }
 
-fn main(protocol: &str, kernel_args: args::Args) -> ! {
+fn main<H: oak_idl::Handler>(protocol: &str, kernel_args: args::Args, idl_service: H) -> ! {
```
andrisaar, line 95: I'd go one step further refactoring all of this code, but as Conrad said above, it should likely be in a separate PR.

What I imagine is `start_kernel()` returning some kind of data structure -- for now, it might be just `Box<dyn Channel>`, and then in `oak_baremetal_app_crosvm`, instead of calling start_kernel with the idl_service as an argument, you call `oak_baremetal_communication_channel::server::start_blocking_server` in there. No refererences to IDL in the kernel required!

andrisaar (approved):

experimental/oak_baremetal_kernel/src/lib.rs
ipetr0v: Done
Also moved `main` to the `app_crosvm` instead of `kernel`

experimental/oak_baremetal_channel/src/server.rs
ipetr0v: Isn't `oak_idl::Handler` already a generic trait that could be used by all future IDLs?

conradgrobler: It is somewhat generic as it just passes bytes back and forth, but assumes a very specific structure for the IDL. Specifically it assumes that is will there will always be pairs of requests and responses that map one to one and that the request will identify the method by a numeric method ID. It also does not support any additional metadata or anything like that. This will not always be suitable for all workloads we might want to run in future.

PR #3190, opened 2022-09-09
Add GHCB data structure to the Hello World testing kernel

conradgrobler:
This is the first in a series of PRs to enable serial port based logging on AMD SEV-ES and SEV-SNP using the GHCB IOIO protocol to drive the serial port.

This PR adds the GHCB data structure to the Hello World testing kernel, initialises it and shares it with the hypervisor (only SEV-ES version at the moment which requires marking the page as unencrypted).

Actually using this data structure and the protocol to write to the serial port will be done in a follow-up PR.

### DIFF
--- a/testing/sev_snp_hello_world_kernel/src/ghcb.rs
+++ b/testing/sev_snp_hello_world_kernel/src/ghcb.rs
@@ -1,0 +1,43 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use core::mem::MaybeUninit;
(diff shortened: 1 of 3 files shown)


testing/sev_snp_hello_world_kernel/src/ghcb.rs
```diff
@@ -0,0 +1,120 @@
...
+        unimplemented!("SNP support is not yet implemented.");
+    }
+    // Mark the page as not encrypted.
+    set_encrypted_bit_for_page(&ghcb_address, false);
+}
+
+/// Sets the encrypted bit for the page that contains the address.
+///
+/// We assume for now that we will always use 2MiB huge pages.
+fn set_encrypted_bit_for_page(address: &VirtAddr, encrypted: bool) {
```
ipetr0v, line 59: IIUC this function is a page table walk for finding a specific phys address.
Is this page table walk used by other parts of the kernel? Could it be shared somewhere, so we won't have code duplication?

ipetr0v (approved):

testing/sev_snp_hello_world_kernel/src/ghcb.rs
conradgrobler: Yes, this does a manual page table walk because we need to find the right entry to modify. When we add more advanced page table manipulation features to the main restricted kernel we could change this code to reuse a subset of that logic, but for now I don't think there is really a suitable library where this could go.

andrisaar: There are page table wrappers in the `x86_64` library, but those don't know anything about the encrypted bit, so probably will not be useful here. We will likely need to implement our own page table abstractions at one point.

andrisaar (approved):

Issue #3191, opened 2022-09-09
Support Interactive Attestation over gRPC streams 


jul-sh:
# Context

To remotely attest an Oak instance a client performs a remote attestation handshake. This takes place using a sequence of messages, as shown below:
<img width="1183" alt="Screen Shot 2022-09-09 at 4 33 37 pm" src="https://user-images.githubusercontent.com/8875406/189387776-53767473-da50-4518-8e79-276638ef5e5d.png">

Messages belonging to the same sequence are linked using a `session_id`. The client generates a unique `session_id` at the start of the handshake, and includes it in every request. Communication between the client and untrusted launcher takes place over unary gRPC messages. 

# Objective

As an alternative to unary gRPC messages, the client and untrusted launcher should be able to communicate over a gRPC stream.

The contents and sequence of the messages exchanged in this scenario will remain the exact same. Communication between the untrusted launcher and trusted runtime will remain the same. The runtime logic will remain the same.

In this scenario the stream would only act as a proxy for stickyness, ensuring a consistent connection between the client and the same Oak instance.

# Tasks

- [ ] Define a streaming service
- [ ] Expose a streaming endpoint on the untrusted launcher
- [ ] Implement a client that connects to the streaming endpoint

jul-sh:
cc @tiziano88 @ipetr0v

conradgrobler:
We are moving away from using the interactive attestation between the client and server.

PR #3192, opened 2022-09-09
Rename the former UnaryClient trait to clarify its role of abstracting networking

jul-sh:
Useful as soon (see #3191) we'll have an implementation of this trait using streaming gRPC. Renaming this now avoids confusion later on code related to unary gRPC.


mariaschett (approved):
Thank you!

PR #3194, opened 2022-09-09
Add Echo Bare-Metal Runtime

ipetr0v:
This PR adds a simple Echo baremetal runtime that can be used for tests.


testing/echo_baremetal_app_crosvm/build.rs
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+fn main() {
+    println!("cargo:rerun-if-changed=target.json");
```
andrisaar, line 18: minor clean-up, but this line is not needed as there no longer is a `target.json` file.

andrisaar (approved):
In general: let's not call it `echo_baremetal_app_crosvm` but just `echo_baremetal_app` (if not even shorter than that, ditching baremetal as a moniker as well). The crosvm part made sense when we had two builds, one for qemu and one for crosvm, but we no longer do.

testing/echo_baremetal_runtime/README.md
```diff
@@ -0,0 +1,5 @@
+# Echo Baremetal Runtime
+
+No-std compatible implementation of the business logic of the Echo Runtime. The
+interface of the runtime is defined using the Oak IDL in `schema.rs`, the Rust
```
conradgrobler, line 4: "schema.fbs"?

conradgrobler (approved):

testing/echo_baremetal_app_crosvm/build.rs
ipetr0v: Done

ipetr0v:
Updated to `echo` and `echo_runtime`

testing/echo/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "echo"
```
tiziano88, line 2: This name is going to clash with existing crates. I think even though we are not going to publish this to crates.io, we should use crate names that are at least likely to be unique (e.g. oak_echo). Or do you think we should not care about this problem at all, and pretend none of the crates in this repo will ever be published? Or should we add a tag / comment on the ones that are internal only?
cc @conradgrobler

tiziano88 (approved):

testing/echo_baremetal_runtime/schema.fbs
```diff
@@ -0,0 +1,13 @@
+attribute "method_id";
+
+rpc_service TrustedRuntime {
```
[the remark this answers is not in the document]
ipetr0v: Done

testing/echo/Cargo.toml
ipetr0v: Done

PR #3195, opened 2022-09-12
Move Trusted Shuffler out of experimental

### LINKED ISSUE #3178: Move Trusted Shuffler out of experimental sub-folder
Move the code of Trusted Shuffler to a top level folder `trusted_shuffler` in workspace.

This also requires some updates of the paths
- in the [README](https://github.com/project-oak/oak/blob/592ade5c5b77c987fb54a8dd8a2dadfcc8ba5066/experimental/trusted_shuffler/README.md)
- in the internal documentation

I think we should keep the `scripts` for Trusted Shuffler in the Trusted Shuffler folder as opposed to moving them. 

As we are moving stuff we could move the [echo.proto](https://github.com/project-oak/oak/blob/592ade5c5b77c987fb54a8dd8a2dadfcc8ba5066/experimental/trusted_shuffler/common/src/echo.proto) to a more canonical place, too.

mariaschett:
Fixes #3178  .


ipetr0v (approved):
IIUC currently the plan is to have all of the Rust crates to be in the top-level directory.
But it's ok to leave it like this for now, because we have a lot of supporting crates here.

PR #3197, opened 2022-09-12
Set up data structures similar to crosvm and jump to kernel in our stage0 BIOS

andrisaar:
Things to note:
- This is a _massive_ pile of unsafe code.
- Linking Rust code in, especially in debug mode, required us to have a larger firmware image.
- There's a ton of clean-up to do in here, and missing features, but this is enough to get the machine to start executing from `0x200000`.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -15,7 +15,7 @@
 */
 
 HIDDEN(TOP = 4096M);
-HIDDEN(BIOS_SIZE = 256K);
+HIDDEN(BIOS_SIZE = 512K);
 
 MEMORY {
     bios : ORIGIN = TOP - BIOS_SIZE, LENGTH = BIOS_SIZE
@@ -61,6 +61,8 @@ SECTIONS {
         QUAD((TOP - 2M) | PAGE_PRESENT | PAGE_WRITABLE | PAGE_SIZE) /* (4GiB - 2MiB) .. 4 GiB */
     } > bios
 
+    pd_offset = ADDR(.rodata.pd);
+
     .rodata : {
         KEEP(*(.rodata .rodata.*))
     } > bios
@@ -80,7 +82,7 @@ SECTIONS {
     } > bios
 
     .stack ALIGN(4K) (NOLOAD) : {
-        . += 4K;
+        . += 16K;
         stack_start = .;
     } > bios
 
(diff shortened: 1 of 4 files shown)


stage0/src/main.rs
```diff
@@ -17,15 +17,139 @@
...
+    },
+    PhysAddr, VirtAddr,
+};
 mod asm;
 
+// Magic constants where various data structures will be stored.
+// Many of these are set to be same as with crosvm; see
+// https://google.github.io/crosvm/appendix/memory_layout.html for more details.
+const BOOT_GDT_ADDR: PhysAddr = PhysAddr::new(0x1500);
+// The Rust GlobalDescriptorTable has 8 entries; the one provided by crosvm has 4.
```
conradgrobler, line 43: nit: I assume this comment means that the addres of the IDT is at a different place compared to crosvm, but not 100% sure. Perhaps adding a bit of additional clarification to the comment will make the intention clearer.

stage0/layout.ld
```diff
@@ -15,7 +15,7 @@
 */
 
 HIDDEN(TOP = 4096M);
-HIDDEN(BIOS_SIZE = 256K);
+HIDDEN(BIOS_SIZE = 512K);
```
conradgrobler, line 18: I suspect this would once-again require updating the readme to keep in sync.

stage0/src/main.rs
```diff
@@ -17,15 +17,139 @@
...
+    // Safety: changing the page tables here is safe because (a) we've populated the new page tables
+    // at the well-known address of BOOT_PML4_ADDR and (b) we've ensured we keep the mapping for our
+    // own code.
+    unsafe {
+        Cr3::write(
+            PhysFrame::from_start_address(BOOT_PML4_ADDR).unwrap(),
+            Cr3Flags::empty(),
+        );
+    }
+    /* TODO: interrogate qemu for memory areas and set up the zero page. */
```
conradgrobler, line 147: I suspect some of the formatting code in CI will ask you to add issue numbers to the TODO statements.

conradgrobler (approved):

stage0/src/main.rs
andrisaar: Yes, crosvm places the IDT at 0x1520. I've expanded the comment.

stage0/layout.ld
andrisaar: I've just got rid of the explicit size in the README :)

ipetr0v (approved):

PR #3201, opened 2022-09-13
Use GHCB serial port for logging when running on AMD SEV-ES

conradgrobler:
This is a follow-up to #3190. It allows the Hello World test kernel to use the GHCB serial port implementation when running on AMD SEV-ES.

Enabling support for AMD SEV-SNP will be done in a later PR.


andrisaar (approved):
Let's hope we can actually test this in a day or two :)

PR #3205, opened 2022-09-13
Expand cargo-deny coverage and resolve warnings

conradgrobler:
We have not had full cargo-deny coverage over all of the workspaces, as not all of the appropriate `Cargo.toml` files contained `[workspace]` tags.

We also reused the same root `Deny.toml` file for all the cargo-deny checks which caused a lot of warnings by not being specific enough for each workspace.

This change adds `[workspace]` tags to ensure all workspaces are covered. It also creates a specific `Deny.toml` file for each workspace to remove the warnings.

Adding the `[workspace]` tag also expanded cargo-udeps coverage, which required a fix in xtask to work with the `no_std` binaries.


mariaschett (approved):
Thank you!

PR #3210, opened 2022-09-14
SEV enlightenment: set the encrypted bit if we're running under SEV-{,ES}.

andrisaar:
This automatically detects whether we're running under SEV or not, and if memory encryption is required, sets the correct bit in the page tables.

To do that, we first need to enter unreal mode -- real mode with a 32-bit data segment -- so that we can access the memory near the 4G boundary, where the page tables live.

### DIFF
--- a/stage0/deny.toml
+++ b/stage0/deny.toml
@@ -19,5 +19,5 @@ wildcards = "allow"
 # List of allowed licenses.
 # For more detailed information see http://go/thirdpartylicenses.
 [licenses]
-allow = ["Apache-2.0", "MIT"]
+allow = ["Apache-2.0", "BSD-2-Clause", "MIT", "Unicode-DFS-2016"]
 copyleft = "deny"
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -24,8 +24,10 @@ MEMORY {
 ENTRY(reset_vector)
 
 /* Segment descriptor flags.
- * See Section 4.8 in AMD64 Architecture Programmer's Manual, Volume 2 for more details.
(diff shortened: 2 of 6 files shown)


stage0/layout.ld
```diff
@@ -37,6 +39,7 @@ HIDDEN(SEGMENT_CONFORMING = 1 << (32 + 10));
 HIDDEN(PAGE_PRESENT = 1 << 0);
 HIDDEN(PAGE_WRITABLE = 1 << 1);
 HIDDEN(PAGE_SIZE = 1 << 7);
+HIDDEN(ENCRYPTED = 1 << 51);
```
conradgrobler, line 42: This doesn't seem to be used at the moment.

conradgrobler (approved):
Nice, thanks.

It seems like stage0/deny.toml needs some fixes.

ipetr0v (approved):

stage0/deny.toml
```diff
@@ -19,5 +19,17 @@ wildcards = "allow"
 # List of allowed licenses.
 # For more detailed information see http://go/thirdpartylicenses.
 [licenses]
-allow = ["Apache-2.0", "MIT"]
+allow = [
+  "Apache-2.0",
+  "Apache-2.0 WITH LLVM-exception",
```
conradgrobler, line 24: I don't think this license is required, and a few of the others. It probably only needed "Unicode-DFS-2016" and "BSD-3-Clause" on top of what was already there. Adding unused licenses to the allow list will cause noisy warnings in cargo deny.

andrisaar: Fair enough, I just copied everything from the root `deny.toml` file.

PR #3211, opened 2022-09-14
Add a raw channel to Oak Echo binary

ipetr0v:
This PR:
- Adds a `raw_server` and `idl_server` features to the Oak Echo baremetal binary
- Renames `oak_echo` to `oak_echo_bin`


testing/oak_echo_bin/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+pub extern "C" fn rust64_start(_rdi: u64, rsi: &bootparam::BootParams) -> ! {
+    let channel = oak_baremetal_kernel::start_kernel(rsi);
+    info!("In main!");
+    start_echo_server(channel)
+}
+
+fn start_echo_server(mut channel: Box<dyn Channel>) -> ! {
+    /// Starts an echo server that reads single bytes from the channel and writes
+    /// them back.
+    #[cfg(all(feature = "raw_server", not(feature = "idl_server")))]
```
conradgrobler, line 48: I think that one of these should not include the `not` part, otherwise this functions will be empty when clippy analyses it with all features enabled.

testing/oak_echo_bin/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+#![no_std]
+#![no_main]
+#![feature(alloc_error_handler)]
+#![feature(core_c_str)]
+#![feature(core_ffi_c)]
+
+extern crate alloc;
+
+use alloc::boxed::Box;
+#[cfg(all(feature = "raw_server", not(feature = "idl_server")))]
```
conradgrobler, line 26: It might be a bit cleaner to read if you move this and the constant in line 36 into the code block starting at line 49.

conradgrobler (approved):

testing/oak_echo_bin/src/main.rs
ipetr0v: Done

testing/oak_echo_bin/src/main.rs
ipetr0v: Done

PR #3223, opened 2022-09-20
Call CPUID to figure out whether memory encryption is supported at all

andrisaar:
We want the stage0 binary to be compatible with both machines that do support SEV, and machines that do not.

To determine whether memory encryption is enabled (and that we need to set the encrypted bit), we need to access the `SEV_STATUS` MSR.
Curveball 1: if we try to read the `SEV_STATUS` MSR on a machine that doesn't support it, we get killed with a `#GP` fault. Therefore, we need to determine via `CPUID` whether SEV is supported at all.
Curveball 2: if we're running under SEV-ES (and presumably SEV-SNP), calling `CPUID` triggers a `#VC` fault as you're supposed to use the GHCB protocol to get the CPUID values.

Thus: let's create a `#VC` handler that returns "yes, SEV is supported" (as if you're not under SEV-ES, how did you end up triggering that exception?), and then check the CPUID value before reading `SEV_STATUS`.

As reading the MSR is potentially crash-worthy, I've also removed references to that MSR from the Rust layer. Instead, let's pass in the bit number as a parameter from the assembly.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -24,7 +24,7 @@ MEMORY {
 ENTRY(reset_vector)
 
 /* Segment descriptor flags.
- * See Sections 4.7 and  4.8 in AMD64 Architecture Programmer's Manual, Volume 2 for more details.
+ * See Sections 4.7 and 4.8 in AMD64 Architecture Programmer's Manual, Volume 2 for more details.
  */
 HIDDEN(SEGMENT_4K = 1 << (32 + 23)); /* G */
 HIDDEN(SEGMENT_DEFAULT_32BIT_OP = 1 << (32 + 22)); /* D/B */
@@ -39,6 +39,10 @@ HIDDEN(SEGMENT_CONFORMING = 1 << (32 + 10));
 HIDDEN(PAGE_PRESENT = 1 << 0);
 HIDDEN(PAGE_WRITABLE = 1 << 1);
 HIDDEN(PAGE_SIZE = 1 << 7);
+
+/* Gate descriptor flags */
+HIDDEN(DESCRIPTOR_PRESENT = 1 << 47);
+HIDDEN(DESCRIPTOR_INTERRUPT_GATE = 0xE << 40);
 
 SECTIONS {
     . = ORIGIN(bios);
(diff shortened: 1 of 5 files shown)


stage0/layout.ld
```diff
@@ -149,8 +153,45 @@ SECTIONS {
              0xFFFF)              /* Segment Limit [15:0] */
     } > bios
 
+    /* 32-bit interrupt table */
     .rodata.idt ALIGN(8) : {
-        QUAD(0)
+        QUAD(0) /*  0 - #DE */
```
tiziano88, line 158: Any linkable reference for those values?

stage0/src/asm/boot.s
```diff
@@ -23,18 +23,46 @@ _start :
...
     mov %eax, %ds
+    mov %eax, %ss
+
+    # Set up a basic stack, as we may get interrupts.
+    mov $stack_start, %esp
 
-    # Determine if we're running under SEV
+    # Determine if we're running under SEV. Keep track of which bit is the encrypted bit in %rsi.
+    mov $0, %esi              # by default, no encryption
+    mov $0x8000001F, %eax
```
tiziano88, line 51: Any reference for these numbers? I'm thinking if you are OOO and somehow we get errors that seem related to these numbers, how would we (anyone other than you) double check them? :)

stage0/src/main.rs
```diff
@@ -112,7 +111,7 @@ pub unsafe fn jump_to_kernel(entry_point: VirtAddr) -> ! {
 }
 
 #[no_mangle]
-pub extern "C" fn rust64_start() -> ! {
+pub extern "C" fn rust64_start(encrypted: u64) -> ! {
```
tiziano88, line 114: Add documentation explaining the semantics of the param.

tiziano88 (approved):

stage0/src/asm/boot.s
andrisaar: You ask @conradgrobler of course :)

I've added some documentation.

stage0/src/main.rs
```diff
@@ -136,17 +140,12 @@ pub extern "C" fn rust64_start() -> ! {
     create_idt(idt);
     idt.load();
 
+    // We assume 0-th bit is never the encrypted bit.
```
tiziano88, line 143: Somehow I think this assumption will bite us back when we may want to use the 0-th bit as the encrypted bit :)

andrisaar: Bit 0 is the `PRESENT` bit, so I'd be impressed if they overload that one.

tiziano88: ah, ok then, probably not :)

conradgrobler (approved):
Nice

Issue #2602, opened 2022-03-09
Make scripts work with sudo Docker


tiziano88:
Currently we assume sudoless Docker for convenience, but that is actually discouraged. We should either support running scripts with explicit sudo, or install Docker in [rootless mode](https://docs.docker.com/engine/security/rootless/).

tiziano88:
I'm taking a closer look at this one, with the excuse that I just got a new machine.

PR #3224, opened 2022-09-21
Use rootless docker

tiziano88:
Required disabling buildkit support, which seems incompatible with
rootless for some reason (needs further investigation in https://github.com/project-oak/oak/issues/3225).

The next step (if no one complains) will be to remove the
`fix_docker_user_and_run` script, in another PR.

This should not affect CI, since CI usually already runs under root
anyways (Docker is not a security boundary, so it would be unlikely that
any CI would run the Docker daemon as root but the CI steps as a
different user).

Fix https://github.com/project-oak/oak/issues/2602

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -119,6 +119,46 @@ recommended, since it requires a lot of effort to manually install all the tool
 natively on your machine and keeping them up to date to the exact same version
 specified in Docker.
 
+## rootless Docker
+
+In order to run Docker without root privileges, follow the guide at
+https://docs.docker.com/engine/security/rootless/ .
+
+Below is a quick summary of the relevant steps.
+
+1. Install `uidmap`.
+
+   ```bash
+   sudo apt install uidmap
+   ```
+
+1. Add a range of subids for the current user.
+
+   ```bash
+   sudo usermod --add-subuids 500000-565535 --add-subgids 500000-565535 $USER
+   ```
+
+1. Download the install script for rootless docker, and run it as the current
+   user.
+
+   ```bash
+   curl -fSSL https://get.docker.com/rootless > $HOME/rootless
+   sh $HOME/rootless
+   ```
+
+1. Add the generated environment variables to your shell.
+
+   ```bash
+   export PATH=$HOME/bin:$PATH
+   export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
+   ```
+
+1. Test whether everything works correctly.
+
+   ```bash
+   docker run hello-world
+   ```
+
 ## xtask
 
 `xtask` is a utility binary to perform a number of common tasks within the Oak
--- a/docs/development.md
+++ b/docs/development.md
@@ -161,6 +161,27 @@ Below is a quick summary of the relevant steps.
    export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
    ```
 
+   **As an alternative** to setting the `DOCKER_HOST` environment variable, it
+   is possible to instead run the following command to set the Docker context.
+
+   ```bash
+   docker context use rootless
+   ```
+
+   In either case, running the following command should show the current status.
+
+   ```console
+   $ docker context ls
+   NAME        DESCRIPTION                               DOCKER ENDPOINT                       KUBERNETES ENDPOINT   ORCHESTRATOR
+   default *   Current DOCKER_HOST based configuration   unix:///run/user/152101/docker.sock                         swarm
+   rootless    Rootless mode                             unix:///run/user/152101/docker.sock
+   Warning: DOCKER_HOST environment variable overrides the active context. To use a context, either set the global --context flag, or unset DOCKER_HOST environment variable.
+   ```
+
+   This should show either that the default context is selected and is using the
+   user-local docker endpoint from the `DOCKER_HOST` variable, or that the
+   `rootless` context is selected.
+
 1. Test whether everything works correctly.
 
    ```bash
--- a/docs/development.md
+++ b/docs/development.md
@@ -126,6 +126,14 @@ https://docs.docker.com/engine/security/rootless/ .
 
 Below is a quick summary of the relevant steps.
 
+1. If you have an existing version of Docker running as root, uninstall that
+   first.
+
+   ```
+   sudo systemctl disable --now docker.service docker.socket
+   sudo apt remove docker-ce docker-engine docker-runc docker-containerd
+   ```
+
 1. Install `uidmap`.
 
    ```bash
--- a/docs/development.md
+++ b/docs/development.md
@@ -129,7 +129,7 @@ Below is a quick summary of the relevant steps.
 1. If you have an existing version of Docker running as root, uninstall that
    first.
 
-   ```
+   ```bash
    sudo systemctl disable --now docker.service docker.socket
    sudo apt remove docker-ce docker-engine docker-runc docker-containerd
    ```
@@ -166,6 +166,12 @@ Below is a quick summary of the relevant steps.
    ```bash
    docker run hello-world
    ```
+
+If you rely on VSCode remote / dev container support, that should also keep
+working, but you need to make sure that your VSCode instance sees the update
+environment variables so that it connects to the correct Docker host socket. The
+simplest way of doing this is to invoke `code` from a shell with the updated
+environment variables.
 
 ## xtask
 
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -4,7 +4,7 @@
 // - https://code.visualstudio.com/docs/remote/devcontainerjson-reference
 {
   // Do not modify manually. This value is automatically updated by ./scripts/docker_build .
-  "image": "sha256:2e64a98711d1957d1208bfb9d557552f337823adcc4b63ba1876a4bd567380f9",
+  "image": "sha256:215d9b11115a46f799341852ab6c43147f43414cb5e02be35d0abd63ed75cbb6",
   "extensions": [
     "13xforever.language-x86-64-assembly",
     "bazelbuild.vscode-bazel",
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -4,7 +4,7 @@
 // - https://code.visualstudio.com/docs/remote/devcontainerjson-reference
 {
   // Do not modify manually. This value is automatically updated by ./scripts/docker_build .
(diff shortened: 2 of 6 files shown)


tiziano88:
@conradgrobler I would appreciate if you could run these steps and perhaps also clone this branch to make sure I didn't break any workflow. In particular, I also fixed the devcontainer config, but it would be good if you could confirm that also still works for you.

@mariaschett @rbehjati I think this may help with the transparent release step of getting hold of the output file, since the UID should be transparently mapped between Docker and the host. But I'm not 100% sure that's the only issue you had.

conradgrobler:
I followed the install stesp, cloned this branch and did some testing:

- The initial install of rootless docker failed because I already had roofull docker installed
- I forced the updated by running `FORCE_ROOTLESS_INSTALL=1 sh $HOME/rootless`
- Running the scripts in docker seems find (e.g. `./script/docker_run ...`)
- I cannot open the dev container in Visual Studio anymore: it give the following logs:

```
[78248 ms] Remote-Containers 0.235.0 in VS Code 1.69.1 (b06ae3b2d2dbfe28bca3134cc6be65935cdfea6a).
[78247 ms] Start: Run: docker version --format {{.Server.APIVersion}}
[78280 ms] 
[78280 ms] Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

tiziano88:
Thanks Conrad, I should have mentioned that it's probably easier to first uninstall rootful Docker -- anyways glad to hear you managed to install them alongside.

I think the issue with VSCode is that it's still using the socket of the rootful daemon; have you tried running VSCode from a shell or some other environment in which the DOCKER_HOST var is correctly set as `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock` ?

docs/development.md
```diff
@@ -119,6 +119,54 @@ recommended, since it requires a lot of effort to manually install all the tool
...
+
+In order to run Docker without root privileges, follow the guide at
+https://docs.docker.com/engine/security/rootless/ .
+
+Below is a quick summary of the relevant steps.
+
+1. If you have an existing version of Docker running as root, uninstall that
+   first.
+
+   ```
```
conradgrobler, line 132: Looks like there should be a `bash` here as welll.

conradgrobler (approved):

mariaschett:
> @mariaschett perhaps you should test this with your remote container setup as well to make sure that still works for you as well.

Trying. Currently my VS Code on my local machine running with `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock` seems convinced that `Error: No such image: sha256:9e6f478527ad452a83ebfc56e3e0fe2550f164e43bb9810ab989bddf4923f163`, which is a lie!

Will continue experimenting tomorrow!

tiziano88:
> > @mariaschett perhaps you should test this with your remote container setup as well to make sure that still works for you as well.
> 
> Trying. Currently my VS Code on my local machine running with `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock` seems convinced that `Error: No such image: sha256:9e6f478527ad452a83ebfc56e3e0fe2550f164e43bb9810ab989bddf4923f163`, which is a lie!
> 
> Will continue experimenting tomorrow!

It may just be that you need to call the `docker_pull` script, since the image also changed. Let us know if that worked!

docs/development.md
tiziano88: Done

mariaschett:
Sadly, I `scripts/docker_pull`ed already on remote: 

```
latest: Pulling from oak-ci/oak
Digest: sha256:9e6f478527ad452a83ebfc56e3e0fe2550f164e43bb9810ab989bddf4923f163
Status: Image is up to date for gcr.io/oak-ci/oak:latest
```
Do you think I have to make sure `DOCKER_HOST` is set on remote too?

docs/development.md
```diff
@@ -119,6 +119,60 @@ recommended, since it requires a lot of effort to manually install all the tool
...
+
+1. If you have an existing version of Docker running as root, uninstall that
+   first.
+
+   ```bash
+   sudo systemctl disable --now docker.service docker.socket
+   sudo apt remove docker-ce docker-engine docker-runc docker-containerd
+   ```
+
+1. Install `uidmap`.
```
mariaschett, line 137: 1. -> 2. and below.

mariaschett: I just noticed, it shows that correctly in the preview! Please disregard!

conradgrobler:
> Do you think I have to make sure `DOCKER_HOST` is set on remote too?

Yes, I think it must be set up on the remote host. You could also try to run `docker context use rootless` on the remote host. For me that removed the need to set `DOCKER_HOST`.

mariaschett:
So I have just followed the instructions in `development.md` on rootless Docker on both: remote and local. I now get the same error as Conrad [before[(https://github.com/project-oak/oak/pull/3224#issuecomment-1255077554): "Cannot connect to the Docker daemon at http://docker.example.com. Is the docker daemon running?"

But sadly running VS Code with `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock code` does not fix it for me.

tiziano88:
That's odd, why would it expect the daemon to be running at http://docker.example.com/ ? Could you check that `$DOCKER_HOST` was expanded correctly?

mariaschett:
So after running `docker context use rootless` on remote, the [Docker daemon connecting to example.co](https://github.com/project-oak/oak/pull/3224#issuecomment-1255961725)  disappeared locally (NB: yes, `$DOCKER_HOST` was expanded correcty). 

But now I came full circle :)
```
Error: No such image: sha256:9e6f478527ad452a83ebfc56e3e0fe2550f164e43bb9810ab989bddf4923f163```

However, I think the reason is:

```Error response from daemon: pull access denied for sha256, repository does not exist or may require 'docker login': denied: requested access to the resource is denied```

**Comment by tiziano88** [comment]
For reference, this is what it looks like on mine:

```console
$ echo $DOCKER_HOST       
unix:///run/user/152101/docker.sock
```

tiziano88:
@mariaschett what is the command you are running that is giving you that error? or is it VSCode in its terminal when rebuilding the container?

mariaschett:
> @mariaschett what is the command you are running that is giving you that error? or is it VSCode in its terminal when rebuilding the container?

It is in VSCode when it is loading my local devcontainer.json.

mariaschett:
Maybe I haven't done a good enough job here:

> first uninstall rootful Docker

Would you have any pointers on how to check that?

Also, what makes me a bit suspicious on remote is:

```
Current context is now "rootless"
Warning: DOCKER_HOST environment variable overrides the active context. To use "rootless", either set the global --context flag, or unset DOCKER_HOST environment variable.
```

tiziano88:
I'm not super familiar with your setup (@conradgrobler also not sure if yours is the same or different).

AFAICT you have a local installation in which you only have the docker cli, and then a remote machine in which you have the actual docker daemon running, which is the one that you just replaced with the rootless daemon; is that correct? Or do you also have a docker daemon running locally?

tiziano88:
Leaving this here in case it's useful: https://code.visualstudio.com/remote/advancedcontainers/develop-remote-host#_connect-using-docker-contexts

conradgrobler:
> I'm not super familiar with your setup (@conradgrobler also not sure if yours is the same or different).

I am not using the remote container setup at the moment.

mariaschett:
> local installation in which you only have the docker cli

I am using the remote container extension in VS Code to [develop on a remote docker host](https://code.visualstudio.com/remote/advancedcontainers/develop-remote-host) using a "local devcontainer.json file as a way to configure, create, and connect to a remote dev container". Not sure what it is doing under the hood.

> remote machine in which is the one that you just replaced with the rootless daemon;
Yes. I also have the rootless daemon on my local machine, but I don't think that should make a difference.

tiziano88:
I think I finally managed to recreate your setup, and it seems to work here (without rootful docker installed on the server). I had a couple of complications related to env variables not being set up correctly on the server, specifically PATH and DOCKER_HOST, but in the end I solved the first one with a symlink (`sudo ln -s $HOME/bin/docker /usr/bin/docker`), and I used the docker context (`docker context use rootless` on the server) instead of the explicit socket env variable (I didn't think it would persist, but it did).

mariaschett:
Works for me too now with sudoless docker running on remote connecting with the remote container extension and VS Code.

docs/development.md
```diff
@@ -119,6 +119,60 @@ recommended, since it requires a lot of effort to manually install all the tool
...
+   ```
+
+1. Download the install script for rootless docker, and run it as the current
+   user.
+
+   ```bash
+   curl -fSSL https://get.docker.com/rootless > $HOME/rootless
+   sh $HOME/rootless
+   ```
+
```
mariaschett, line 156: Maybe add: You can check the context with `docker context ls`.

mariaschett (approved):
Thank you!

conradgrobler:
I found one other potential issue: with this approach it is no longer possible to run the crosvm/bare-metal examples in the VS Code integrated terminal in the dev container.

It is possible to get it to work when running the docker scripts on an external terminal by setting the appropriate permissions for the current user:

```
sudo setfacl -m u:${USER}:rw /dev/kvm
sudo setfacl -m u:${USER}:rw /dev/vsock
sudo setfacl -m u:${USER}:rw /dev/vhost-vsock

./scripts/docker_run ./scripts/xtask --logs run-launcher-test
```

But even with these permission changes it does not work with the VS Code terminal:

```
crosvm has exited with error: failed to create kvm: Permission denied (os error 13)
```

I don't think this is a blocker in terms of submitting the PR, as it is fine to test it using an external terminal. But it would be useful if we can figure out how to get this to run in the VS Code terminal as well, especially if we want to support interactive debugging with VS Code extensions.

tiziano88:
Thanks @conradgrobler , that seems WAI, in fact probably exactly the kind of issue that running the Docker daemon as root would introduce.
I am not sure why the fix you suggested does not work in VSCode though -- maybe we need to add more flags to the devcontainer config file?

conradgrobler:
I would guess that it is likely related to https://github.com/project-oak/oak/blob/main/.devcontainer.json#L27-L37, but I think that can be investigated as a separate PR.

docs/development.md
tiziano88: Good point, done.

PR #3226, opened 2022-09-26
SEV-SNP enlightenment: add necessary data structures to stage0

andrisaar:
This is enough to get the stage0 code up and running under SEV-SNP; verified by adding a `loop {}` just after `rust64_start`.

Mind you, the VM crashes as soon as we call `GlobalDescriptorTable::new()` inside the stage0 code proper; presumably that's because we haven't validated the memory we're trying to access.

The locations for the secrets/cpuid/etc pages can't be in the BIOS blob itself. I poked around EDK2 source code, and it looks like they place those pages in the lower 1M of memory, and thankfully it looks like the addresses don't clash with anything we're using (our L1 page table is at `0xB000`), so we can use the same addresses as they do.

We do need to document the memory layout properly, though, to ensure that we won't accidentally introduce clashes in the future.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -44,6 +44,13 @@ HIDDEN(PAGE_SIZE = 1 << 7);
 HIDDEN(DESCRIPTOR_PRESENT = 1 << 47);
 HIDDEN(DESCRIPTOR_INTERRUPT_GATE = 0xE << 40);
 
+/* SEV metadata section types; values defined in
+ * https://github.com/tianocore/edk2/blob/master/OvmfPkg/ResetVector/X64/OvmfSevMetadata.asm
+ */
+HIDDEN(SEV_SECTION_UNMEASURED = 0x1);
+HIDDEN(SEV_SECTION_SECRET = 0x2);
+HIDDEN(SEV_SECTION_CPUID = 0x3);
+
 SECTIONS {
     . = ORIGIN(bios);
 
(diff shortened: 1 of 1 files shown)


stage0/layout.ld
```diff
@@ -213,9 +220,59 @@ SECTIONS {
...
+        BYTE(0x41) BYTE(0x53) BYTE(0x45) BYTE(0x56)
+        LONG(sev_guided_structure_end - sev_guided_structure_start)
+        LONG(1)
+        LONG((sev_guided_structure_end - sev_guided_structure_start - 16) / 12)
+        /* Each section has the basic structure of uint32 address, uint32 length, uint32 type
+         * The locations of these pages have been chosen to be the same as EDK2:
+         * https://github.com/tianocore/edk2/blob/c05a218a9758225ddf94eedb365633f2154551da/OvmfPkg/OvmfPkgX64.fdf
+         * Thankfully none of these clashed with existing data structures that we're setting up.
+         */
+        /* Unmeasured page location */
```
conradgrobler, line 246: What will the unmeasured section used for? Will this be used for generating the boot parameter page in future?

conradgrobler (approved):

stage0/layout.ld
andrisaar: At this point it's unused. The main reason I added it was that our VMM refuses to boot if you don't have at least one unmeasured section, so there we are.

EDK2 sets up two or three unmeasured sections; I'll need to do some more digging to find out what they're used for in there.

PR #3230, opened 2022-09-26
Rename Markdown files to have correct extension

tiziano88:
The uppercase extension is non-standard, and those files were not being checked for correctness by the formatting tools.


mariaschett (approved):
Thanks!

PR #3238, opened 2022-09-28
Create a separate crate for Linux boot data structures.

andrisaar:
We need the data structures in `stage0` to set up the boot environment, and later in the kernel(s) to read the zero page. Hence, it makes sense to have a separate crate for these that we can share between the different crates.

There's nothing Oak-specific in there, either. It's just the Linux data structures, Rustified a bit by hand.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3404,6 +3404,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "oak_linux_boot_params"
+version = "0.1.0"
+dependencies = [
+ "bitflags",
+ "static_assertions",
+ "strum",
+]
+
+[[package]]
 name = "oak_logger"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
   "experimental/virtio",
   "experimental/vsock/echo",
   "experimental/web_client",
+  "linux_boot_params",
   "oak_functions/abi",
   "oak_functions/client/rust",
   "oak_functions/examples/benchmark/module",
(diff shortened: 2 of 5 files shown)


linux_boot_info/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "linux_boot_info"
```
tiziano88, line 2: I think this should be `oak_linux_boot_info`. Or in fact even `oak_boot_info_linux`? (so that if  there were other boot infos in the future, the prefix would be the same).

linux_boot_info/src/lib.rs
```diff
@@ -0,0 +1,626 @@
...
+
+#[repr(C, packed)]
+pub struct CCSetupData {
+    pub header: SetupData,
+    pub cc_blob_address: u32,
+}
+
+/// Real-mode Kernel Header.
+///
+/// For each field, some are information from the kernel to the bootloader (“read”), some are
```
tiziano88, line 113:
```suggestion
/// For each field, some are information from the kernel to the bootloader ("read"), some are
```

maybe use standard double quotes, unless this was auto generated by bindgen, in which case feel free to leave as is.

tiziano88 (approved):

conradgrobler (approved):

linux_boot_info/Cargo.toml
andrisaar: Actually this should be called `oak_linux_boot_params` as the main Linux C struct is called `boot_params`; that's where the name comes from, it's the Rust version of the Linux `boot_params`.

linux_boot_info/src/lib.rs
andrisaar: Nope, that's a manual copy-paste from the Linux kernel docs; I didn't even notice they were wrong. Fixed.

PR #3239, opened 2022-09-28
Move hard-coded addresses to linker script in stage0.

andrisaar:
The main goal here was to move magic memory addresses to the linker script rather than have them in the Rust code.

This means that the linker can do some sanity checks: for example, if we have overlapping data structures, the linker would throw an error when linking the code. That wouldn't be the case if we're using raw memory addresses in The Rust code.

So, in a way, this code is safer than it was before: first, the linker guarantees that the memory won't overlap; second, `MaybeUninit::write` guarantees that the data structure there will be initialized properly. (It's a safe function!).
Unfortunately as we're dealing with `static mut`-s, the increased safety is not immediately visible as we need to wrap every access to those in an `unsafe` block anyway. But it' slightly less unsafe!

As an added bonus, the linker script acts as kind of a documentation about the memory layout as well.

The only "magic" address left in code is `0x7000` for the zero page. I will handle that in a separate PR.

Fun fact: we now have more linker script than Rust code...

Fun fact 2: for the `SEV_*` static data structures, we have to convince the build system _twice_ that we know what we're doing: first, we need to have `#[used]` in the Rust code otherwise the Rust compiler will throw them away and second, we need to wrap the section in `KEEP()` in the linker script otherwise the linker will throw them away.

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -17,7 +17,7 @@
 #![no_std]
 #![no_main]
 
-use core::{arch::asm, ffi::c_void, panic::PanicInfo};
+use core::{arch::asm, ffi::c_void, mem::MaybeUninit, panic::PanicInfo};
 use sev_guest::instructions::{pvalidate, InstructionError, PageSize as SevPageSize, Validation};
 use x86_64::{
     instructions::{hlt, interrupts::int3, segmentation::Segment},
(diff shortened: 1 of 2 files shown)


stage0/src/main.rs
```diff
@@ -38,22 +38,32 @@ use x86_64::{
...
+#[link_section = ".boot.idt"]
+static mut BOOT_IDT: MaybeUninit<InterruptDescriptorTable> = MaybeUninit::uninit();
+#[link_section = ".boot.pml4"]
+static mut BOOT_PML4: MaybeUninit<PageTable> = MaybeUninit::uninit();
+#[link_section = ".boot.pdpt"]
+static mut BOOT_PDPT: MaybeUninit<PageTable> = MaybeUninit::uninit();
+#[link_section = ".boot.pd"]
+static mut BOOT_PD: MaybeUninit<PageTable> = MaybeUninit::uninit();
+#[link_section = ".boot.unmeasured"]
+#[used]
```
tiziano88, line 52: Does the used attribute actually do anything for these statics? I remember in your previous experiments it didn't seem to make a difference.

tiziano88 (approved):

stage0/src/main.rs
```diff
@@ -190,18 +198,15 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
...
      */
 
-    // Safety, for all these data structure dereferences: The addresses are to well-known locations,
-    // same as crosvm is using, in the lowest 1MB of memory which we know is valid.
-    let gdt = unsafe { &mut *(BOOT_GDT_ADDR.as_u64() as *mut GlobalDescriptorTable) };
-
     if snp {
         pvalidate(
-            x86_64::addr::align_down(gdt as *const _ as u64, 0x1000) as usize,
+            x86_64::addr::align_down(unsafe { BOOT_GDT.as_ptr() } as u64, 0x1000) as usize,
```
conradgrobler, line 203: Please add a comment somewhere in the code (e.g. in module-level doc commentss) to explain why these many unsafe blocks are actually ok, similar to what is in the PR description.

conradgrobler (approved):

stage0/src/main.rs
andrisaar: Yes, they do! See the PR description for details, but this means that the Rust compiler won't throw them away. I think why it looked like they didn't do anything in my previous code was because the linker threw them away, but here we force both of them to include the seemingly unused data structures.

stage0/src/main.rs
andrisaar: Done.

PR #3244, opened 2022-09-29
Always compile xtask with explicit target

tiziano88:
This prevents the cargo cache from being overwritten when some other build command is run. Currently, if you run xtask and it runs some other command, sometimes it invalidates its own binary in the cargo cache, which means the next time it is invoked it will have to be rebuilt again, which takes a few tens of extra seconds each time. Now it is instant instead.


rbehjati (approved):
Nice. Thanks.

PR #3245, opened 2022-09-29
Remove unsafe version of Oak Functions

### LINKED ISSUE #3242: Remove "unsafe" versions of monolithic Oak Functions binaries
We are not planning to use them anyways, so I would rather delete them to simplify the migration towards #3075 .

This would mean that we have only a single version of Oak Functions, packaged to run either on the Oak Restricted Kernel or as a Linux binary.

In particular, we would not have support to run TF models via tract, nor to export differentially private metrics.

Also, logging from the Wasm code would not be possible, at least until we re-introduce it via some other way that is compatible with the Oak Restricted Kernel (e.g. by sending logs back alongside the response message).

@conradgrobler WDYT? Is if fine to delete these?

tiziano88:
Move all tests to use the base version.

Enable support for grpc-web in the base version too; this will not be around long there anyways, as we transition to using the split launcher anyways, which will implement it on the untrusted side (therefore will not affect the trusted runtime at all).

Fix #3242


xtask/src/internal.rs
```diff
@@ -178,8 +178,6 @@ pub struct BuildClient {
 pub enum ServerVariant {
```
andrisaar, line 178: Are we expecting any other server variants in the future?

(Something something baremetal?)

tiziano88: Yeah I'm leaving it in for now, so that we can e.g. use it to select between multiple trusted runtimes, but it may well be that we won't actually need it, in which case I'll delete it later on.

andrisaar (approved):
I feel like this could have been split into smaller PRs.

conradgrobler (approved):

PR #3246, opened 2022-09-29
Populate locations of SEV secrets/cpuid pages in zero page.

andrisaar:
We were hoping to use magic ELF sections in the kernel itself to designate where the secrets and CPUID pages should go, but that won't work with the stage0 approach.

Instead of that, if stage0 detects we're running under SEV, let's add an extended setup_data arg to the zero page that points to the locations of those magic pages.

This is also how the location information is handed to the Linux kernel itself.

For now, we'll assume that the VMM has populated (the rest of) the zero page for us; in the hopefully not-too-distant future, the whole zero page will be created in stage0.

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -17,7 +17,13 @@
 #![no_std]
 #![no_main]
 
-use core::{arch::asm, ffi::c_void, mem::MaybeUninit, panic::PanicInfo};
+use core::{
+    arch::asm,
+    ffi::c_void,
+    mem::{size_of, zeroed, MaybeUninit},
+    panic::PanicInfo,
+};
+use oak_linux_boot_params::BootParams;
 use sev_guest::instructions::{pvalidate, InstructionError, PageSize as SevPageSize, Validation};
 use x86_64::{
     instructions::{hlt, interrupts::int3, segmentation::Segment},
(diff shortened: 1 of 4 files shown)


conradgrobler (approved):

stage0/src/main.rs
```diff
@@ -284,6 +294,32 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
     }
     /* TODO(#3198): interrogate qemu for memory areas and set up the zero page. */
 
+    // Safety: We assume the VMM has filled in the basic zero page for us. This is not true with
+    // qemu. In the future, construction of the full zero page should be in here, and instead of
+    // `assume_init_mut` we will use the safe `write` to zero it out and get a mutable reference.
+    let zero_page = unsafe { BOOT_ZERO_PAGE.assume_init_mut() };
+
+    if snp {
```
rbehjati, line 302: Why is this if-block not merged with the next? (I like the structure that it gives to the code, but I wonder if there is any other reason for it.)

stage0/src/main.rs
```diff
@@ -284,6 +294,32 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
     }
     /* TODO(#3198): interrogate qemu for memory areas and set up the zero page. */
 
+    // Safety: We assume the VMM has filled in the basic zero page for us. This is not true with
+    // qemu. In the future, construction of the full zero page should be in here, and instead of
+    // `assume_init_mut` we will use the safe `write` to zero it out and get a mutable reference.
+    let zero_page = unsafe { BOOT_ZERO_PAGE.assume_init_mut() };
```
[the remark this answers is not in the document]
andrisaar: It could be, but I left it outside as we'll need the variable for populating the zero page in the future anyway.

stage0/src/main.rs
andrisaar: No other reason but that they're semantically different operations - this deals with setting up the zero page, the other block deals with `PVALIDATE`ing the memory.

andrisaar: In particular, I expect the second block to go away in its current form in not too distant future.

PR #3247, opened 2022-09-29
Use the zero page to locate the cpuid/secrets structures in the hello world kernel

andrisaar:
This is complementary to https://github.com/project-oak/oak/pull/3246.

We were always printing out those data structures, even when we weren't running under SEV-SNP; as the pages weren't populated, they contained garbage anyway.

Instead, let's print out nice status messages about whether we've detected any of {SEV,SEV-ES,SNP} as enabled and print out the cpuid/zero pages only under SEV-SNP.

### DIFF
--- a/testing/sev_snp_hello_world_kernel/src/main.rs
+++ b/testing/sev_snp_hello_world_kernel/src/main.rs
@@ -17,7 +17,8 @@
 #![no_std]
 #![no_main]
 
-use core::{mem::MaybeUninit, panic::PanicInfo};
+use core::panic::PanicInfo;
+use oak_linux_boot_params::{BootParams, CCBlobSevInfo, CCSetupData, SetupDataType};
 use sev_guest::{
     cpuid::CpuidPage,
     msr::{get_sev_status, SevStatus},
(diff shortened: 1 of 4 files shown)


testing/sev_snp_hello_world_kernel/src/main.rs
```diff
@@ -29,25 +30,50 @@ mod asm;
...
+            setup_data = unsafe { (*setup_data).next };
+        }
 
-    log::info!("CPUID page: {:?}", cpuid);
-    log::info!("Secrets page: {:?}", secrets);
+        if !setup_data.is_null() {
+            let cc_setup_data = unsafe { &*(setup_data as *const CCSetupData) };
+            let cc_blob = unsafe { &*(cc_setup_data.cc_blob_address as *const CCBlobSevInfo) };
+            if cc_blob.magic != oak_linux_boot_params::CC_BLOB_SEV_INFO_MAGIC {
+                panic!("CCBlobSevInfo data structure has invalid magic!");
```
tiziano88, line 66: Perhaps also print it?

testing/sev_snp_hello_world_kernel/src/main.rs
```diff
@@ -29,25 +30,50 @@ mod asm;
...
+            }
+            let cpuid = unsafe { &*(cc_blob.cpuid_phys as *const CpuidPage) };
+            let secrets = unsafe { &*(cc_blob.secrets_phys as *const SecretsPage) };
+
+            log::info!("CPUID page: {:?}", cpuid);
+            log::info!("Secrets page: {:?}", secrets);
+        } else {
+            log::warn!("CCSetupData data structure not found in setup_data");
```
tiziano88, line 74: Should this be a panic directly?

tiziano88 (approved):

testing/sev_snp_hello_world_kernel/src/main.rs
andrisaar: Good idea, done.

testing/sev_snp_hello_world_kernel/src/main.rs
andrisaar: It doesn't matter in the hello world kernel itself; in the real Restricted Kernel, yes.

conradgrobler (approved):

PR #3248, opened 2022-09-30
Remove unneeded ignore statement

### LINKED ISSUE #3179: Stop ignoring RUSTSEC-2022-0053 once mapr is maintained of tract-tensorflow stops using it
The mapr crate is currently unmaintained and the tract-tensorflow crate depends on it.

conradgrobler:
Since we are no longer using tract tensorflow, we no longer need to ignore a cargo deny advisory that for one of its dependencies.

Fixes #3179


oak_functions/examples/mobilenet/client/rust/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
```
tiziano88, line 1: Was this re-added by mistake?

conradgrobler: I guess so. I am not sure how it ended up there, but I probably made a mistake when rebasing. Deleted now.

PR #3249, opened 2022-09-30
Update axum-core to resolve cargo-deny advisory

### LINKED ISSUE #3202: Stop ignoring "RUSTSEC-2022-0055" when Tonic relies on a newer version of Axum

conradgrobler:
Tonic has not been updated, but a fixed version of axum-core is compatible with the current version of axum.

Fixes #3202


mariaschett (approved):
Sorry for the delay!

PR #3250, opened 2022-09-30
Check if we have a ELF header at 0x200000 in stage0

### LINKED ISSUE #3199: Don't assume the kernel entry point is at `0x200000`
It probably won't be. The entry point is specified in the ELF header, but we don't have access to that in stage0 (yet). Figure out how to handle that properly.

andrisaar:
Our hello world kernel had its entry point at `0x200000`, so we blindly jumped to that address. However, our real Restricted Kernel has its entry point at `0x201000`, as it first loads its own ELF file header to address `0x200000`.

Thus, let's check if we have a valid ELF header at that address. If yes, look up the entry point address from the header. If we don't find an ELF header at that address, assume it's code and just jump there.

Strictly speaking there is no standard that requires there to be either code or ELF header at 0x200000, it's just a convention we follow. I suspect we need to do something more clever here if we ever want to make the Linux kernel boot with our stage0.

Fixes #3199

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -23,6 +23,7 @@ use core::{
     mem::{size_of, zeroed, MaybeUninit},
     panic::PanicInfo,
 };
+use goblin::elf::header;
 use oak_linux_boot_params::BootParams;
 use sev_guest::instructions::{pvalidate, InstructionError, PageSize as SevPageSize, Validation};
 use x86_64::{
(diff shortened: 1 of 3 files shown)


stage0/src/main.rs
```diff
@@ -332,10 +333,28 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
...
+    if header.e_ident[0] == header::ELFMAG[0]
+        && header.e_ident[1] == header::ELFMAG[1]
+        && header.e_ident[2] == header::ELFMAG[2]
+        && header.e_ident[3] == header::ELFMAG[3]
+        && header.e_ident[4] == header::ELFCLASS64
+        && header.e_ident[5] == header::ELFDATA2LSB
+        && header.e_ident[6] == header::EV_CURRENT
+        && header.e_ident[7] == header::ELFOSABI_SYSV
+    {
+        // Looks like we have a valid ELF header at 0x200000. Trust it's entry point.
```
conradgrobler, line 353: nit: "it's" -> "its"

conradgrobler (approved):

Issue #2918, opened 2022-06-01
Implement transcript signing logic in Java client


tiziano88:
It seems that the Java client does not actually sign the transcript as part of remote attestation (and the Rust logic happily skips over it).

https://github.com/project-oak/oak/blob/b43decc6069103a256c368d9c7cef85e6b9252fe/remote_attestation/java/src/com/google/oak/remote_attestation/ClientHandshaker.java#L146-L147

https://github.com/project-oak/oak/blob/b43decc6069103a256c368d9c7cef85e6b9252fe/remote_attestation/rust/src/handshaker.rs#L508-L533

conradgrobler:
We are no longer using the interactive remote attestation handshake, so this is no longer relevent.

PR #3253, opened 2022-10-03
Use Split launcher to run examples

### LINKED ISSUE #3241: Make Oak Functions examples use the split loader
Similar to #3081 , but for Oak Functions examples

tiziano88:
Change logic in `xtask` to parse the config values (e.g. wasm module
location, constant size) from the example.toml file directly, since the
split launcher accepts them as separate flags instead of a single config
file.

The web client does not seem to work with the split launcher, so I
temporarily disabled it and created #3256 to follow up.

The Java client was also initially not working, but I tracked that down
to the server trying to parse the transcript signing key, but the Java
client only providing zeros for it at the moment. That work is already
tracked in #2918.

Fix #3241

### DIFF
--- a/xtask/src/examples.rs
+++ b/xtask/src/examples.rs
@@ -78,8 +78,9 @@ struct WizerOpt {
 #[derive(serde::Deserialize, Debug, Default)]
 #[serde(deny_unknown_fields)]
 struct Server {
-    #[serde(default)]
-    additional_args: Vec<String>,
+    wasm_path: String,
+    constant_response_size_bytes: u64,
+    lookup_data: String,
     #[serde(default)]
     server_variant: ServerVariant,
 }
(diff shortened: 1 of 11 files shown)


xtask/src/examples.rs
```diff
@@ -239,6 +240,25 @@ pub fn build_oak_functions_server_variants(opt: &BuildServerOpt) -> Step {
     }
 }
 
+/// Build every variant of the function server.
```
conradgrobler, line 243: This comment does not seem to match the single-step nature of the implementation.

conradgrobler (approved):

xtask/src/examples.rs
tiziano88: Yeah I'm planning to clean this up in an upcoming PR (removing the enum entirely).

PR #3254, opened 2022-10-03
Initialize frame allocator from E820 and use it for finding memory for the heap

andrisaar:
Following up to #3168, this now makes use of the bitmap allocator.

First, we create a proper frame allocator for physical memory that is able to handle both 2M and 4K pages, as necessary.

Second, set up the frame allocator with the information from the E820 map, and use it to find the largest memory region for the heap, similar to what we do now.

This covers the basic functionality, but that's not all we need. It still relies on the magic linker variables to make sure we won't overwrite our other data structures; these reserved regions should be properly parsed from the ELF header (which we do load into memory).
And, we want the page allocator to use this frame allocator as well.

But we now do have an authoritative data structure to represent which memory regions are in use and which are not, and you can request a frame for your perusal during runtime.

### DIFF
--- a/experimental/oak_baremetal_kernel/src/mm/frame_allocator.rs
+++ b/experimental/oak_baremetal_kernel/src/mm/frame_allocator.rs
@@ -36,12 +36,12 @@ pub struct PhysicalMemoryAllocator<const N: usize> {
     /// When first asked for a 4K frame, we take one 2 MiB frame to hand out as 4 KiB frames; that
     /// gives us 512 4K frames to hand out. Thus, the allocator bitmap needs to be 512/64 = 8
     /// u64-s, or 64 B.
-    small_frames: Option<BitmapAllocator<Size4KiB, 1>>,
+    small_frames: Option<BitmapAllocator<Size4KiB, 8>>,
 }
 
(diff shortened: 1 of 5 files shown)


experimental/oak_baremetal_kernel/src/mm/frame_allocator.rs
```diff
@@ -0,0 +1,142 @@
...
+    Size2MiB, Size4KiB,
+};
+
+/// Allocator to track physical memory frames.
+///
+/// The basic unit we track is a 2 MiB frame. If necessary, we will take one 2 MiB frame and break
+/// it into smaller, 4 KiB frames to satisfy demand.
+///
+/// The parameter N needs to be set to the number of u64-s required to track the physical memory in
+/// a bitmap. For example, if we want to track 128 GiB of memory, that means 128G/2M = 65K frames,
```
tiziano88, line 29: (nit, not entirely sure)

```suggestion
/// a bitmap. For example, if we want to track 128 GiB of memory, that means 128Gi/2Mi = 65K frames,
```

tiziano88 (approved):

conradgrobler (approved):

Issue #3075, opened 2022-07-22
Consolidate trusted / untrusted binaries


tiziano88:
At the moment in the GitHub repository we have a few "monolithic" binaries, i.e. binaries that include both the trusted and untrusted parts of Oak; the main ones are [`oak_functions_loader_base`](https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_base) and [`oak_functions_loader_unsafe`](https://github.com/project-oak/oak/tree/main/oak_functions/oak_functions_loader_unsafe).

Separately, we have a "split" binary that allows running Oak Functions in a VMM (currently either QEMU or crosvm, selectable via command line flags): [`oak_baremetal_loader`](https://github.com/project-oak/oak/tree/main/experimental/oak_baremetal_loader).

The split binary is much closer to what we would run in a real deployment, so we should aim to remove the monolithic binaries and only keep the split ones, and use the split ones to run integration tests.

At the moment some functionality of `oak_functions_loader_unsafe` cannot be easily ported to the split binary, because TF inference and differentially private metrics are not `no_std` compatible, and therefore cannot be run as part of a bare metal binary inside a VM. @conradgrobler suggested that we could add a new mode to the untrusted launcher so that it is able to also run a regular linux binary as a child process, and communicate with it over some local communication channel (e.g. unix domain socket, pipe, file descriptor), without actually creating a VM instance. This would allow us to still maintain an architecture much closer to the real one, and we can then gradually "graduate" the functionality in this binary to be no_std compatible if / when needed.

The remaining tasks are:

- #3078 
- #3079
- #3080
- #3081
- #3241
- #3082

We should create separate issues for these, and keep this issue up to date with progress, so please let me know if I forgot anything so I can amend it directly.

tiziano88:
This can be considered as complete, even though #3079 is still missing, but that will require further changes to the Oak Restricted Kernel to support timers, so we can keep tracking it separately.

PR #3257, opened 2022-10-04
Generate provenance for crosvm Oak Functions

tiziano88:
The base and unsafe binaries are going away soon, and anyways we only care about provenance of the "trusted" binaries.

Ref #3075

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -20,9 +20,7 @@ jobs:
       hashes: ${{ steps.hash.outputs.hashes }}
     runs-on: ubuntu-20.04
     env:
-      OAK_FUNCTIONS_BASE_PATH: './target/x86_64-unknown-linux-musl/release/oak_functions_loader_base'
       OAK_BAREMETAL_CROSVM_PATH: './experimental/oak_baremetal_app_crosvm/target/x86_64-unknown-none/release/oak_baremetal_app_crosvm'
-      OAK_FUNCTIONS_RELEASE_PATH: './target/x86_64-unknown-linux-musl/release'
 
     permissions:
       # Allow the job to update the repo with the latest provenances and index.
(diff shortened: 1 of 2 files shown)


rbehjati:
@tiziano88 What is the plan for removing Oak functions? That is the only project that has a fuzz target, and we need that as an example for the internship project.

.github/workflows/provenance.yaml
```diff
@@ -133,21 +129,16 @@ jobs:
...
         run: |
-          # first cd to the release path. This allows us calling sha256 on oak_functions_loader_base
-          # instead of having to provide the full path. This is fine for this experimental version,
-          # but has to be cleaned-up once we switch completely to SLSA provenance generation.
-          cd ${{ env.OAK_FUNCTIONS_RELEASE_PATH }}
-          # sha256sum generates sha256 hash for oak_functions_loader_base.
+          # sha256sum generates sha256 hash for the target binary.
           # base64 --wrap=0 disables line wrapping, and encodes to base64 and outputs on a single line.
-          echo "::set-output name=hashes::$(sha256sum oak_functions_loader_base | base64 --wrap=0)"
+          echo "::set-output name=hashes::$(sha256sum ${{ env.OAK_BAREMETAL_CROSVM_PATH }} | base64 --wrap=0)"
```
rbehjati, line 141: I don't think this will work. The output will contain an artefact name that contains `.` and `/`. I am not sure that `slsa-github-generator` can use this kind of artefact name.

.github/workflows/provenance.yaml
```diff
@@ -133,21 +129,16 @@ jobs:
...
 
   # This job calls the generic workflow to generate provenance.
   # See https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md
```
rbehjati, line 144: In line 157, change 
`attestation-name: 'oak_functions_loader_base.intoto.jsonl'`
to 
`attestation-name: 'oak_baremetal_app_crosvm.intoto.jsonl'`

andrisaar (approved):
We really need to rename the `experimental/oak_baremetal_crosvm_app` to something more sensible sooner rather than later.

rbehjati (approved):
Thanks :)

Unfortunately I don't think there is a way to test the workflow locally. Also, currently it is only set up to be triggered manually.

PR #3261, opened 2022-10-04
Remove non-interference proofs

tiziano88:
Currently unused, and we don't have a plan to use them.


tiziano88:
Sure, here it is: https://github.com/project-oak/oak/releases/tag/coq_non_interference_proofs , PTAL.

aferr (approved):

aferr:
@tiziano88 as you may have already noticed, you missed a spot: https://github.com/project-oak/oak/blob/main/.github/workflows/coq.yaml :)

tiziano88:
Cc @googlebot

PR #3264, opened 2022-10-04
Add a template runtime for Tensorflow

ipetr0v:
This PR copies an existing Echo Baremetal binary and makes it a template for the future TensorFlow Baremetal binary.


tiziano88:
Please don't put more stuff under experimental, I'm trying to move stuff out of it

ipetr0v:
> Please don't put more stuff under experimental, I'm trying to move stuff out of it

Moved to `tensorflow/...`

tiziano88:
I would suggest just having the crates as top level, the nesting just makes it harder to find them, and I'm not sure it helps with anything in particular (or does it?).

ipetr0v:
> I would suggest just having the crates as top level, the nesting just makes it harder to find them, and I'm not sure it helps with anything in particular (or does it?).

Done

oak_tensorflow_bin/layout.ld
```diff
@@ -0,0 +1,111 @@
+/*
```
tiziano88, line 1: What's the plan to keep this in sync with the other binaries @andrisaar @conradgrobler  ?

tiziano88 (approved):

oak_tensorflow_bin/layout.ld
conradgrobler: We don't yet have a good plan for the linker scripts. I vaguely remember that one linker script can reference another, so we might be able to move a lot of it to a central script.

We should reuse the asm between the different binaries and do general deduplication where possible before creating any new binaries that are based on the restricted kernel.

tiziano88: Perhaps we could use some symlinks for the time being, instead of maintaining several copies in our repo?

ipetr0v:

PR #3265, opened 2022-10-04
Use the ELF header to mark memory as reserved

andrisaar:
We have lots of data structures in memory: first, stage0 populates several things in the lowest 2 MiB of memory, and the kernel/app itself is loaded somewhere in the physical memory.

Previously, we just used the end of the stack as a guide to reason where the memory we use ends, so that we could use the rest for the heap. This was at best an approximation.

One data structure we load into memory is the ELF file header itself, which contains program headers -- that's instructions for the loader as to what data will be loaded where from the ELF file.

We can use the same information to mark regions of memory where we know data has been loaded as unavailable, instead of relying on the end of the stack always being the marker.


experimental/oak_baremetal_kernel/src/mm/mod.rs
```diff
@@ -65,5 +68,44 @@ pub fn init<const N: usize, E: E820Entry>(
...
+            PhysFrame::from_start_address(PhysAddr::new(0x0)).unwrap(),
+            PhysFrame::from_start_address(PhysAddr::new(0x0)).unwrap(),
+        ),
+        false,
+    );
+
+    // Second, mark every `PT_LOAD` section from the phdrs as used.
+    program_headers
+        .iter()
+        .filter(|phdr| phdr.p_type == PT_LOAD)
```
conradgrobler, line 85: What about the stack and the rest of the BSS entries? I don't think they will be PT_LOAD headers?

experimental/oak_baremetal_kernel/src/elf.rs
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use goblin::elf64::program_header::ProgramHeader;
+use x86_64::VirtAddr;
+
+/// Interpret raw memory at the given address as an ELF header and return the program headers.
+///
+/// Safety: this virtual address must be valid and contain ELF headers and program headers.
```
mariaschett, line 22: The "Safety" reads more like a pre-condition on the virtual address than an argument why this is safe to do. Or do I misunderstand?

experimental/oak_baremetal_kernel/src/mm/mod.rs
```diff
@@ -35,11 +37,12 @@ pub fn init<const N: usize, E: E820Entry>(
         PhysFrame::from_start_address(PhysAddr::new((N as u64 * 64 - 1) * Size2MiB::SIZE)).unwrap(),
     ));
 
+    /* Step 1: mark all RAM as available (event though it may contain data!) */
     memory_map
         .iter()
         .inspect(|e| {
             info!(
-                "E820 entry: [{:#016x}..{:#016x}) ({}), type {}",
+                "E820 entry: [{:#018x}..{:#018x}) ({}), type {}",
```
mariaschett, line 45: Why did `016x` change?

experimental/oak_baremetal_kernel/src/mm/mod.rs
```diff
@@ -65,5 +68,44 @@ pub fn init<const N: usize, E: E820Entry>(
         })
         .for_each(|range| alloc.mark_valid(range, true));
 
+    // Step 2: mark known in-use regions as not available.
+
+    // First, leave out the first 2 MiB as there be dragons (and bootloader data structures)
```
mariaschett, line 73: Haha, dragons seem scary, but I'd think bootloader data structures seems scary enough! :)

experimental/oak_baremetal_kernel/src/mm/mod.rs
```diff
@@ -35,11 +37,12 @@ pub fn init<const N: usize, E: E820Entry>(
         PhysFrame::from_start_address(PhysAddr::new((N as u64 * 64 - 1) * Size2MiB::SIZE)).unwrap(),
     ));
 
+    /* Step 1: mark all RAM as available (event though it may contain data!) */
```
mariaschett, line 40: Why do we need to mark all RAM as available first?

mariaschett (approved):
I was mainly relying on your comments to understand what is going on, and that is what my comments are about, too!

Thank you!

experimental/oak_baremetal_kernel/src/mm/mod.rs
andrisaar: They do.

```
Program Headers:
  LOAD           0x000000 0x0000000000200000 0x0000000000200000 0x000190 0x000190 R   0x1000
  LOAD           0x001000 0x0000000000201000 0x0000000000201000 0x000048 0x000048 R E 0x1000
  LOAD           0x002000 0xffffffff80400000 0x0000000000400000 0x25e2ab 0x25e2ab R E 0x1000
  LOAD           0x261000 0xffffffff80800000 0x0000000000800000 0x049cb0 0x049cb0 R   0x1000
  LOAD           0x2ab000 0xffffffff80a00000 0x0000000000a00000 0x000068 0x000068 RW  0x1000
  LOAD           0x400000 0xffffffff80c00000 0x0000000000c00000 0x000000 0x280000 RW  0x200000

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .boot 
   02     .text 
   03     .rodata 
   04     .data .got 
   05     .bss .stack
```
Note that `.bss` and `.stack`, both marked `(NOLOAD)` in the linker script and placed in a segment with `FileSiz` of `0x000000` (indicating there's nothing to load from the file), the segment still has the type `LOAD`.

experimental/oak_baremetal_kernel/src/mm/mod.rs
andrisaar: 64-bit addresses are 16 hexadecimal digits long, but I forgot to account for the `0x` in front of them.

experimental/oak_baremetal_kernel/src/elf.rs
andrisaar: No, that's exactly what it is -- a precondition. The function is marked unsafe because we can't check that, so it's up to the caller to guarantee that.

experimental/oak_baremetal_kernel/src/lib.rs
```diff
@@ -57,15 +58,19 @@ use log::{error, info};
...
 use strum::{EnumIter, EnumString, IntoEnumIterator};
+use x86_64::VirtAddr;
 
 /// Main entry point for the kernel, to be called from bootloader.
 pub fn start_kernel<E: boot::E820Entry, B: boot::BootInfo<E>>(info: B) -> Box<dyn Channel> {
     avx::enable_avx();
     logging::init_logging();
     interrupts::init_idt();
 
+    // Safety: in the linker script we specify that the ELF header should be placed at 0x200000.
```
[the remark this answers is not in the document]
andrisaar: Good idea, done.

PR #3266, opened 2022-10-05
Create basic page tables using our frame allocator.

andrisaar:
We now have a frame allocator to manage physical memory, so we don't have to fly blind.

This creates basic page tables for us using the frame allocator that is compatible with our existing code: that is, it adds an identity-mapped region at [0, 2GiB) that's these days really only used for stack and another identity mapped region at [-2GiB, 2**64] for the kernel itself.

Next up, once we have the ELF header parsing in, we can make the page tables more clever in the upper region. And we should move the heap to the upper half as well, leaving the whole lower half of the memory available for user space in the future.


experimental/oak_baremetal_kernel/src/mm/mod.rs
```diff
@@ -65,5 +70,88 @@ pub fn init<const N: usize, E: E820Entry>(
...
+/// |---------------------|---------|---------------------|--------|-----------------------------|
+/// | 0000_0000_0000_0000 |    0    | 0000_7FFF_FFFF_FFFF | 128 TB | User space                  |
+/// | 0000_8000_0000_0000 | +128 TB | FFFF_7FFF_FFFF_FFFF |  16 EB | Non-canonical addresses, up |
+/// |                     |         |                     |        | to -128 TB                  |
+/// | FFFF_8000_0000_0000 | -128 TB | FFFF_FFFF_7FFF_FFFF |~128 TB | ... unused hole             |
+/// | FFFF_FFFF_8000_0000 |   -2 GB | FFFF_FFFF_FFFF_FFFF |   2 GB | Kernel code                 |
+pub fn init_paging<A: FrameAllocator<Size4KiB> + ?Sized>(
+    frame_allocator: &mut A,
+) -> Result<(), &'static str> {
+    // Safety: this expect the frame allocator to be initialized and the memory region it's hading
```
conradgrobler, line 106: nit: "expects", "handing"

conradgrobler (approved):

experimental/oak_baremetal_kernel/src/mm/page_tables.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use x86_64::{
+    structures::paging::{
+        frame::PhysFrameRange, mapper::MapToError, FrameAllocator, Mapper, Page, PageTableFlags,
+        Size2MiB, Size4KiB,
+    },
+    VirtAddr,
+};
+
+pub unsafe fn create_offset_map<A: FrameAllocator<Size4KiB> + ?Sized, M: Mapper<Size2MiB>>(
```
conradgrobler, line 25: Please add a doc comment with a `# Safety` heading.

conradgrobler (approved):

PR #3269, opened 2022-10-05
Clean-up: use oak_linux_boot_params everywhere.

andrisaar:
In our initial implementation we were rather careful about abstracting over possibly multiple boot protocols, because we didn't know what we'd end up using.

For now, I think it's pretty certain that we're going to use the Linux boot protocol. That's what crosvm uses, and that's what stage0 (will be) implementing.

And, for stage0, we created a separate crate for the Linux boot data structures. Therefore, let's refactor both our kernel and the app binaries to use that crate and the Linux data structures directly.


linux_boot_params/src/lib.rs
```diff
@@ -609,6 +626,26 @@ pub struct BootParams {
 }
 static_assertions::assert_eq_size!(BootParams, [u8; 4096usize]);
 
+impl BootParams {
+    pub fn protocol(&self) -> &'static str {
+        "Linux Boot Protocol"
+    }
+
+    pub fn e820_table(&self) -> &[BootE820Entry] {
```
ipetr0v, line 634: What's an E820 table?

ipetr0v (approved):

linux_boot_params/src/lib.rs
andrisaar: It refers to a BIOS function that provides the memory layout to the operating system:

https://en.wikipedia.org/wiki/E820
https://wiki.osdev.org/Detecting_Memory_(x86)#BIOS_Function:_INT_0x15.2C_EAX_.3D_0xE820

Originally invoked by setting `EAX` to `0xE820` (hence the name) and then invoking interrupt 15h, we don't really have a BIOS API available for us, but the name has stuck for this kind of "memory map" data structures. Hence the name here.

Otherwise, the table contains information about which physical memory address ranges are actually valid (not every physical address has a RAM chip behind it, after all), which ranges are reserved, and so on.

PR #3270, opened 2022-10-05
Move GHCB-capable Serial Port driver to a separate crate

conradgrobler:
Since the serial port driver is not really part of the SEV-SNP-specific library it is now moved to its own crate.

We now have an IO abstraction that can support either direct port-based IO or the GHCB protocol, so the serial device is updated to support this new abstraction. This removes the need for relying on a third-party serial port driver for the unencrypted case.


sev_serial/src/lib.rs
```diff
@@ -55,68 +61,116 @@ const DATA_TERMINAL_READY_AND_REQUEST_TO_SEND: u8 = 3;
 /// See <https://wiki.osdev.org/Serial_Ports#Line_status_register>.
 const OUTPUT_EMPTY: u8 = 1 << 5;
 
+/// A serial port implementation that uses direct port-based IO.
+pub type RawSerialPort<'a> = SerialPort<'a, PortReadOnly<u8>, PortWriteOnly<u8>, RawIoPortFactory>;
```
andrisaar, line 65: Why not just use `Port` instead of the ReadOnly and WriteOnly traits?

It looks like neither the RawSerialPort nor the StaticGhcbSerialPort do anything special that would require them to be separate.

andrisaar (approved):

sev_serial/src/lib.rs
conradgrobler: Not a very good reason, mostly because I did not implement the `PortReader` and `PortWriter` traits on `Port`. Also all of our device drivers use each IO port either for reading or writing, never both.

PR #3271, opened 2022-10-05
Create new fuzz target to replace existing one

tiziano88:
The oak_functions_loader crate will be refactored / removed as part of
#3082, so this maintains a fuzz target with similar semantics, in a
location where it is expected to stay untouched. The new fuzz logic is
somewhat simpler, but it boils down to testing the same policy,
although more directly.

This PR creates a new crate as the `fuzz` top-level directory,
so that it will be easier in the future to add new fuzz targets directly
there without having to synchronize the oss-fuzz configuration.

The previous fuzz target will be removed once the oss-fuzz configuration
is updated to point to the newly created one.

Ref #3082


rbehjati:
This change will break our oss-fuzz project: https://github.com/google/oss-fuzz/blob/40490c3c8d01ebf1fa9635145606939c3bcb9fcc/projects/oak/build.sh#L18. 

How about splitting this into two PRs: First adding `oak_functions/abi/fuzz`, and updating the `oss-fuzz` project to use it, and then removing the existing fuzz target in a separate PR after making sure that the new fuzz target works and provides the reports that we can use for the intern project?

tiziano88:
Thanks, that makes sense, I now added a new fuzz crate so that we can keep adding targets there without having to keep it in sync with the oss-fuzz setup.

rbehjati (approved):
Oh! This is very nice. Thanks.

PR #3278, opened 2022-10-06
Add `cargo-binutils` to our development image.

andrisaar:
`binutils` comes with various utilities to manage binaries, such as `objcopy` which can be used to turn ELF files into headerless binary blobs.

We need `objcopy` for `stage0`: namely, `cargo build` gives us an ELF file, but `qemu` and other VMMs expect the BIOS image to be a headerless blob.

Instead of doing `cargo build` + manually invoking objcopy on the target file, using `cargo-binutils` we can do it in one step:
```
cargo objcopy -- -O binary stage0.bin
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -220,9 +220,10 @@ RUN rustup target add wasm32-unknown-unknown
 # Install musl target for Rust (for statically linked binaries).
 RUN rustup target add x86_64-unknown-linux-musl
 
-# Install rustfmt and clippy.
+# Install various components we need.
 RUN rustup component add \
   clippy \
+  llvm-tools-preview \
   rust-src \
   rustfmt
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -238,6 +239,9 @@ RUN cargo install --version=${deadlinks_version} cargo-deadlinks
 # cf. https://github.com/rust-fuzz/cargo-fuzz/pull/277
 RUN cargo install --git https://github.com/rust-fuzz/cargo-fuzz/ --rev 8c964bf183c93cd49ad655eb2f3faecf543d0012
 
+# Install cargo-binutils.
+RUN cargo install --git https://github.com/rust-embedded/cargo-binutils --rev 36102732f7535b4730f7cd66c670ebe3959994ef
```
tiziano88, line 243: It looks like there are prebuilt release binaries for this: https://github.com/rust-embedded/cargo-binutils/releases/tag/v0.3.6

I think it is better to depend on those directly if possible, since eventually they will be part of the provenance generation, and, paradoxically, being able to say that we use a specific prebuilt binary with a given hash is a stronger story than saying we built it ourselves with our own Rust toolchain.

tiziano88 (approved):

PR #3279, opened 2022-10-07
Add Plain and Secret Request and Response

mariaschett:


trusted_shuffler/server/src/http.rs
```diff
@@ -31,10 +31,7 @@ struct HttpRequestHandler {
 
 #[async_trait]
 impl RequestHandler for HttpRequestHandler {
-    async fn handle(
-        &self,
-        request: TrustedShufflerRequest,
-    ) -> anyhow::Result<TrustedShufflerResponse> {
+    async fn handle(&self, request: PlainRequest) -> anyhow::Result<PlainResponse> {
```
tiziano88, line 34: Not sure about this terminology / naming. The correct word would be "plaintext", not "plain", so "PlainResponse" does not seem accurate to me. Also, that is usually contrasted by "encrypted" instead of "secret".

So personally I would suggest "PlaintextResponse" and "EncryptedResponse", and there would be "decrypt" / "encrypt" functions to convert between them.

WDYT?

trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -33,45 +33,76 @@ use std::{
...
+        PlainRequest {
+            body: self.body,
+            headers: self.headers,
+            uri: self.uri,
+        }
+    }
+}
+
+pub struct PlainRequest {
+    pub body: Vec<u8>,
```
tiziano88, line 54: Is it intentional that this struct has `body`, but `SecretResponse` has `data`? It seems they should use the same terminology (and also types, if possible?)

trusted_shuffler/server/src/http.rs
mariaschett: SG.

trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Yes, but I'd do that in a seperate #3282 .

mariaschett (commented):
Done, thanks!

trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -33,45 +33,77 @@ use std::{
...
+    fn empty() -> PlaintextResponse {
+        PlaintextResponse {
             headers: hyper::HeaderMap::new(),
             data: hyper::body::Bytes::new(),
             trailers: hyper::HeaderMap::new(),
         }
     }
+
+    // TODO(#3281): Add encryption.
+    fn encrypt(self) -> EncryptedResponse {
```
ipetr0v, line 90: Encrypting would require a key that will be stored by an encryptor object provided by the attestation crate. So not sure if the `encrypt` function could be a member of the `PlaintextResponse`

trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -27,23 +30,29 @@ impl RequestHandler for TestRequestHandler {
...
         }
-        Ok(generate_response(&request))
+        Ok(generate_plain_response(&request))
     }
 }
 
 // Non-async version of the `handle_request` function used to create expected responses.
-fn generate_response(request: &TrustedShufflerRequest) -> TrustedShufflerResponse {
-    TrustedShufflerResponse {
+fn generate_plain_response(request: &PlaintextRequest) -> PlaintextResponse {
```
ipetr0v, line 45: Generating would mean that the Trusted Shuffler generates something by its own. But it just proxies requests.
I think it could be renamed to `get_plaintext_response` or something similar

trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -27,23 +30,29 @@ impl RequestHandler for TestRequestHandler {
...
+    PlaintextResponse {
+        headers: hyper::HeaderMap::new(),
+        data: hyper::body::Bytes::from(request.body.clone()),
+        trailers: hyper::HeaderMap::new(),
+    }
+}
+
+// Non-async version of the `handle_request` function used to create expected responses.
+fn generate_secret_response(request: &EncryptedRequest) -> EncryptedResponse {
```
ipetr0v, line 54: Should it be an `encrypted_response` instead of a `secret_response`?

trusted_shuffler/trusted_shuffler/src/tests.rs
```diff
@@ -52,30 +61,30 @@ fn generate_response(request: &TrustedShufflerRequest) -> TrustedShufflerRespons
 
 // Generates request which will be dropped by the test backend, because we indicate it has to be
 // dropped in the request itself.
-fn generate_dropped_request() -> TrustedShufflerRequest {
+fn generate_dropped_request() -> EncryptedRequest {
```
ipetr0v, line 64: Why do we need requests that'll be dropped by the backend? Is it for testing?
I think testing code shouldn't be a part of the trusted part of the trusted shuffler

trusted_shuffler/trusted_shuffler/src/lib.rs
mariaschett: Yeah, I will need to revisit that once I implement #3281 .

trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Done. Thanks!

trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: Done. Thanks.

trusted_shuffler/trusted_shuffler/src/tests.rs
mariaschett: It is for testing and pre-existing. I'd leave it for this PR.

mariaschett (commented):
Thanks for the reviews!

The goal of this PR for me was to mark when we expect requests/responses to get/be en/decrypted. 

Do you think there is something missing?

trusted_shuffler/trusted_shuffler/src/lib.rs
```diff
@@ -33,45 +33,77 @@ use std::{
...
     pub uri: Uri,
 }
 
-impl TrustedShufflerRequest {
-    // We only compare the body, and we need revisit this and decide if this
-    // is how we want to sort our requests.
-    fn compare_body(&self, r2: &TrustedShufflerRequest) -> Ordering {
+// TODO(#3281): Add decryption.
+impl EncryptedRequest {
+    fn decrypt(self) -> PlaintextRequest {
```
tiziano88, line 45: I don't think an encrypted request will have a method to decrypt itself, anyways I guess this will be changed in the future.

tiziano88 (approved):

ipetr0v (approved):

PR #3280, opened 2022-10-07
Refactor the SimpleIO device to be GHCB-compatible

conradgrobler:
Similar to the serial port update in #3270, this change uses the IO abstraction to make the Simple IO device compatible with either direct port-based IO (when running without encryption or with SEV) and the GHCB IOIO protocol (when running with SEV-ES and SNP).

### DIFF
--- a/experimental/oak_baremetal_simple_io/src/lib.rs
+++ b/experimental/oak_baremetal_simple_io/src/lib.rs
@@ -117,7 +117,7 @@ where
     /// Reads the next available bytes from the input buffer, if any are available.
     pub fn read_bytes(&mut self) -> Option<VecDeque<u8>> {
         // Safety: we read the value as a u32 and validate it before using it.
-        let length = unsafe { self.input_length_port.try_read().unwrap() } as usize;
+        let length = unsafe { self.input_length_port.try_read().ok()? } as usize;
 
         // Use a memory fence to ensure the read from the device happens before the read from the
         // buffer.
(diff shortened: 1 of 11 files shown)


experimental/oak_baremetal_simple_io/src/lib.rs
```diff
@@ -45,51 +47,77 @@ pub const OUTPUT_BUFFER_LEGNTH: usize = 4096;
...
-pub struct SimpleIo {
+pub struct SimpleIo<'a, R: PortReader<u32> + 'a, W: PortWriter<u32> + 'a> {
     output_buffer: Vec<u8>,
     input_buffer: Vec<u8>,
-    output_length_port: PortWriteOnly<u32>,
-    input_length_port: PortReadOnly<u32>,
+    output_length_port: W,
+    input_length_port: R,
+    _phantom: PhantomData<&'a W>,
 }
```
tiziano88, line 60: The lifetime bound looks a bit odd in the way it's used. Would something like this work?

```rust
pub struct SimpleIo<'a, R: PortReader<u32>, W: PortWriter<u32>> {
  ...
  output_length_port: &'a W,
  input_length_port: &'a R,
  ...
}
```

Also I wonder if that removes the need of the phantom data.

tiziano88 (approved):

experimental/oak_baremetal_simple_io/src/lib.rs
conradgrobler: The SimpleIo struct owns the two ports. If I change these to be references I get an error that the values are dropped while they are being borrowed, as nothing owns them.

The reason for the lifetime paramter is because the ports have references to the GHCB that in turn needs a lifetime parameter.

I don't like how this turned out with the complications around lifetime parameters etc. I am considering making the lifetime static and using `lazy_static` for the GHCB, seeing that the GHCB for a specific vCPU will be fixed for its entire life. But if I do that I will do it as a separate PR.

PR #3283, opened 2022-10-07
Add a crude #GP handler to sev_snp_hello_world.

andrisaar:
We read the `SEV_STATUS` MSR (`0xC001_0131`) to determine whether we should use the GHCB-based I/O protocol and whether SEV is enabled at all.

If you try to read it on a machine that supports SEV, but SEV is not necessarily enabled, the read works just fine. However, if you try to read it on a machine that _doesn't_ support SEV at all, then it triggers a `#GP` exception as the MSR is unsupported.

Thus: let's implement a really basic `#GP` handler that, when triggered, figures out whether it fired because the next instruction was `rdmsr` and if yes, then just skip the instruction altogether by moving the instruction pointer forward.

### DIFF
--- a/testing/sev_snp_hello_world_kernel/src/main.rs
+++ b/testing/sev_snp_hello_world_kernel/src/main.rs
@@ -16,22 +16,55 @@
 
 #![no_std]
 #![no_main]
+#![feature(abi_x86_interrupt)]
 
-use core::panic::PanicInfo;
+use core::{arch::asm, ops::Deref, panic::PanicInfo};
+use lazy_static::lazy_static;
 use oak_linux_boot_params::{BootParams, CCBlobSevInfo, CCSetupData, SetupDataType};
 use sev_guest::{
     cpuid::CpuidPage,
     msr::{get_sev_status, SevStatus},
     secrets::SecretsPage,
 };
-use x86_64::instructions::{hlt, interrupts::int3};
+use x86_64::{
+    instructions::{hlt, interrupts::int3},
+    structures::idt::{InterruptDescriptorTable, InterruptStackFrame},
+};
 
 mod asm;
 mod ghcb;
 mod serial;
 
+lazy_static! {
+    static ref IDT: InterruptDescriptorTable = {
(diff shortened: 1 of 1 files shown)


testing/sev_snp_hello_world_kernel/src/main.rs
```diff
@@ -16,22 +16,51 @@
...
+        idt
+    };
+}
+
+extern "x86-interrupt" fn gp_handler(mut frame: InterruptStackFrame, _: u64) {
+    unsafe {
+        // RDMSR can cause a #GP if you try to read a non-existing MSR
+        if *frame.deref().instruction_pointer.as_ptr::<u16>() == 0x320Fu16 {
+            // just skip the instruction
+            frame.as_mut().update(|val| {
```
conradgrobler, line 51: Do you think it is perhaps also worth zeroing EAX and EDX to make sure it looks like the MSR returned 0? If we move the code around in future could it cuase stray values in the registers that could make the code think SEV-ES or SNP is enabled?

conradgrobler (approved):

testing/sev_snp_hello_world_kernel/src/main.rs
andrisaar: Yes, you're correct, we're not properly emulating the `rdmsr` instruction but rather relying on the side effects here.

However, for a proper implementation we need to drop into raw assembly; you can't just set the register in Rust code. The `x86-interrupt` calling convention pushes RAX and RDX onto stack, and restores them from the stack before calling `iretq`. Therefore, even if we use `asm!` to write a value to RAX, it'll get overwritten. I don't know if there are any hooks for us to say that we _want_ to clobber RAX and RDX.

andrisaar: I've added a comment stating that this handler doesn't emulate a `rdmsr` instruction, but rather just skips it.

Let's come back to it later, as we're likely going to need a better implementation for the Restricted Kernel anyway.

PR #3291, opened 2022-10-10
Distinguish between `VirtAddr` and `PhysAddr` in the simple_io crate.

andrisaar:
The purpose of this PR is the same as #3290: be explicit whether we're dealing with virtual or physical addresses.

Thankfully there's not much going on in this crate, so the code is trivial.


conradgrobler (approved):
Thanks

PR #3292, opened 2022-10-10
Refactor stripping and building hyper requests

mariaschett:
Refactor for readability.

### DIFF
--- a/trusted_shuffler/server/src/http.rs
+++ b/trusted_shuffler/server/src/http.rs
@@ -32,8 +32,11 @@ struct HttpRequestHandler {
 
 #[async_trait]
 impl RequestHandler for HttpRequestHandler {
-    async fn handle(&self, request: PlaintextRequest) -> anyhow::Result<PlaintextResponse> {
-        let mut request = trusted_shuffler_to_hyper_request(request)?;
+    async fn handle(
+        &self,
+        plaintext_request: PlaintextRequest,
+    ) -> anyhow::Result<PlaintextResponse> {
+        let mut request = HyperRequestWrapper::build(plaintext_request)?.0;
 
         // We want to keep the path of the orginal request from the client.
         //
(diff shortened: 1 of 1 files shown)


trusted_shuffler/server/src/http.rs
```diff
@@ -75,127 +78,130 @@ impl RequestHandler for HttpRequestHandler {
...
-        headers: minimal_headers,
-        uri,
-    };
-
-    Ok(trusted_shuffler_request)
-}
+struct HyperRequestWrapper(Request<Body>);
+
+impl HyperRequestWrapper {
+    // Apart from setting the body and the URI, for our test backend setting to HTTP/2 and POST
```
conradgrobler, line 89: nit: this comment should probably be inside the function as it does not seem to be a doc comment.

trusted_shuffler/server/src/http.rs
```diff
@@ -32,8 +32,11 @@ struct HttpRequestHandler {
 
 #[async_trait]
 impl RequestHandler for HttpRequestHandler {
-    async fn handle(&self, request: PlaintextRequest) -> anyhow::Result<PlaintextResponse> {
-        let mut request = trusted_shuffler_to_hyper_request(request)?;
+    async fn handle(
+        &self,
+        plaintext_request: PlaintextRequest,
+    ) -> anyhow::Result<PlaintextResponse> {
+        let mut request = HyperRequestWrapper::build(plaintext_request)?.0;
```
conradgrobler, line 39: This feeld a bit weird (always returning `wrapper.0` immediately after calling `build`). If the output of `build` is never used as a wrapper, perhaps the output type should change? Same applies to `HyperResponseWrapper::build`

conradgrobler (approved):

trusted_shuffler/server/src/http.rs
mariaschett: I used this Wrapper to `impl` methods for a hyper request, i.e., a `Request<Body>`, here. I think I cannot get around this, but let me know if I am missing something!

mariaschett (commented):
Thank you for the fast review!

trusted_shuffler/server/src/http.rs
conradgrobler: I think this is fine. Another way to  achieve it without the wrapper is to define a trait and implement the trait directly on the hyper request and response structs. A downside to traits is that you have different signatures (e.g. async vs not async) so would need two very similar traits (which might be a bit confusing).

I would probably have approached it somewhat differently by implementing these functions with different names on `PlaintextRequest` and `EncryptedRequest`. If there wasn't a need for async it could have been done by implementing something like `TryFrom` and `TryInto`. Having said that, I think it is more of a question of personal preference, so feel free to go ahead with this approach.

PR #3293, opened 2022-10-10
Update Docker image and Rust version

tiziano88:


rbehjati (approved):

tiziano88:
FYI I had to fix some minor code and clippy issues here and there to make CI happy, but nothing particularly unexpected.

Although I had to decrease the constant time policy in one of the tests, since for some reason the computation now happens faster, and the policy violation test started failing!

PR #3288, opened 2022-10-10
Add TFLM dependency

ipetr0v:
This PR adds:
- TensorFlow for Microcontrollers dependency in WORKSPACE
- Template C++ files that will be used by the Rust Runtime

PR #3295, opened 2022-10-10
Add third_party tflite for microcontroller

ianull:


ipetr0v:
We use Bazel to get dependencies
We have a PR that adds TFLM to our Bazel WORKSPACE:
https://github.com/project-oak/oak/pull/3288

PR #3299, opened 2022-10-11
Rename app and bin

tiziano88:


oak_functions_linux_fd_bin/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+log = "*"
+anyhow = { version = "*", default-features = false }
+clap = { version = "*", features = ["derive"] }
+oak_remote_attestation = { path = "../oak_remote_attestation", default-features = false, features = [
+  "rust-crypto"
+] }
+oak_baremetal_channel = { path = "../experimental/oak_baremetal_channel", features = [
+  "std"
+] }
+oak_functions_freestanding = { path = "../oak_functions_freestanding", default-features = false, features = [
```
conradgrobler, line 18: Since this is also used by the linux implementations, it seems a bit weird that this crate's name includes "freestanding". Perhaps just "oak_functions"? Or "oak_functions_runtime" or "oak_functions_core"?

conradgrobler (approved):

oak_functions_linux_fd_bin/Cargo.toml
tiziano88: Yes I agree, I'm planning to rename that to unify it with https://github.com/project-oak/oak/tree/main/oak_functions . In any case, I think freestanding is still appropriate, in the sense that it is no_std compatible, so it can be used in either context. Or do you think it should mean that it is exclusively for Oak Restricted Kernel?

conradgrobler: My initial thought was that it should only be for things built with the restricted kernel into a freestanding binary, but your point about no_std compatibility also makes sense.

Issue #3297, opened 2022-10-11
Add support for TensorFlow Lite for Microcontrollers


ipetr0v:
We need to create a new runtime that can run [TensorFlow Lite for Microcontrollers](https://github.com/tensorflow/tflite-micro) (TFLM).

This work will include:
- Compiling TFLM with `no-std`
- Adding C++ bindings to the Rust runtime
- Add memory management support to the Restricted Kernel

ipetr0v:
@ianull please reference this issue with when submitting code related to TFLM

tiziano88:
Not planned.

PR #3300, opened 2022-10-11
Link TFLite C++ code to Rust Runtime

ipetr0v:
This PR links together the TFLite C++ code and Rust TensorFlow Runtime.

### DIFF
--- a/cc/tflite_micro/tflite_micro.cc
+++ b/cc/tflite_micro/tflite_micro.cc
@@ -14,6 +14,16 @@
  * limitations under the License.
  */
 
-#include "model.h"
+#include "tflite_micro.h"
 
-namespace oak {}  // namespace oak
+#include <stddef.h>
+#include <stdint.h>
+
+// TODO(#3297): Implement TensorFlow Lite initialization logic.
+int tflite_init(const uint8_t* model_bytes, const uint8_t* tensor_arena_bytes) { return 0; }
+
+// TODO(#3297): Implement TensorFlow Lite inference logic.
+int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t* output_bytes,
+               size_t* output_bytes_len) {
+  return 0;
+}
--- a/cc/tflite_micro/tflite_micro.cc
+++ b/cc/tflite_micro/tflite_micro.cc
@@ -20,7 +20,10 @@
 #include <stdint.h>
 
 // TODO(#3297): Implement TensorFlow Lite initialization logic.
-int tflite_init(const uint8_t* model_bytes, const uint8_t* tensor_arena_bytes) { return 0; }
+int tflite_init(const uint8_t* model_bytes, size_t model_bytes_len,
+                const uint8_t* tensor_arena_bytes, size_t tensor_arena_len) {
+  return 0;
+}
 
 // TODO(#3297): Implement TensorFlow Lite inference logic.
 int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t* output_bytes,
--- a/cc/tflite_micro/tflite_micro.cc
+++ b/cc/tflite_micro/tflite_micro.cc
@@ -21,7 +21,7 @@
 
 // TODO(#3297): Implement TensorFlow Lite initialization logic.
 int tflite_init(const uint8_t* model_bytes, size_t model_bytes_len,
-                const uint8_t* tensor_arena_bytes, size_t tensor_arena_len) {
+                const uint8_t* tensor_arena_bytes, size_t tensor_arena_bytes_len) {
   return 0;
 }
 
--- a/oak_tensorflow_service/build.rs
+++ b/oak_tensorflow_service/build.rs
@@ -14,22 +14,11 @@
 // limitations under the License.
 //
 
-use lazy_static::lazy_static;
 use std::{path::PathBuf, process::Command};
 
 const SCHEMA: &str = "schema.fbs";
-const TFLITE_BUILD_TARGET: &str = "//cc/tflite_micro:tflite-micro";
-
-lazy_static! {
-    // WORKSPACE_ROOT is set in .cargo/config.toml.
-    static ref TFLITE_LIBRARY_SOURCE_DIR: PathBuf = {
-        [env!("WORKSPACE_ROOT"), "cc/tflite_micro/"].iter().collect()
-    };
-
-    static ref TFLITE_LIBRARY_DIR: PathBuf = {
-        [env!("WORKSPACE_ROOT"), "bazel-bin/cc/tflite_micro/"].iter().collect()
-    };
-}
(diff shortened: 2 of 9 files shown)


oak_tensorflow_service/build.rs
```diff
@@ -14,7 +14,13 @@
 // limitations under the License.
 //
 
+use std::{path::PathBuf, process::Command};
+
 const SCHEMA: &str = "schema.fbs";
+const TFLITE_LIBRARY_NAME: &str = "tflite-micro";
+const TFLITE_BUILD_TARGET: &str = "//cc/tflite_micro:tflite-micro";
+// TODO(#3297): Get workspace path from environment variables.
+const TFLITE_LIBRARY_DIR: &str = "/workspace/bazel-bin/cc/tflite_micro/";
```
tiziano88, line 23: You can use something like this

https://github.com/project-oak/oak/blob/ffb1cef36f28f58d568ce4a997dbf1c881c33804/oak_functions/sdk/oak_functions/tests/integration_test.rs#L28

oak_tensorflow_service/src/tflite.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// an error code.
+    fn tflite_run(
+        input_bytes: *const u8,
+        input_bytes_len: u64,
+        output_bytes: *mut u8,
+        output_bytes_len: *mut u64,
+    ) -> i32;
+}
+
+// TODO(#3297): Use 8Gb or 10Gb arena sizes.
```
tiziano88, line 39: I guess you meant this

```suggestion
// TODO(#3297): Use 8GiB or 10GiB arena sizes.
```

oak_tensorflow_service/src/tflite.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl TfliteModel {
+    pub fn new() -> Self {
+        let tensor_arena = vec![0; TENSOR_ARENA_SIZE];
+        Self { tensor_arena }
+    }
+
+    pub fn initialize(&self, model_bytes: &[u8]) -> anyhow::Result<()> {
+        unsafe {
+            let _ = tflite_init(model_bytes.as_ptr(), self.tensor_arena.as_ptr());
```
tiziano88, line 54: I don't think the let binding needs to be inside the unsafe block, does it? It can be `let _ = unsafe { ... }` (also below)

tiziano88 (approved):

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done

cc/tflite_micro/tflite_micro.cc
```diff
@@ -14,6 +14,16 @@
...
 
-#include "model.h"
+#include "tflite_micro.h"
 
-namespace oak {}  // namespace oak
+#include <stddef.h>
+#include <stdint.h>
+
+// TODO(#3297): Implement TensorFlow Lite initialization logic.
+int tflite_init(const uint8_t* model_bytes, const uint8_t* tensor_arena_bytes) { return 0; }
```
conradgrobler, line 23: I suspect you will have to specify the sizes of the model and the arena here as well.

oak_tensorflow_service/build.rs
```diff
@@ -26,4 +32,27 @@ fn main() {
...
+        panic!("Failed to run bazel build: exit status is {}", status);
+    }
+
+    // Add TensorFlow Lite build directory to the library search path.
+    println!("cargo:rustc-link-search={}", TFLITE_LIBRARY_DIR);
+
+    // Rerun build.rs if TensorFlow Lite library has been changed.
+    let library_path =
+        PathBuf::from(TFLITE_LIBRARY_DIR).join(format!("lib{}.a", TFLITE_LIBRARY_NAME));
+    println!("cargo:rerun-if-changed={}", library_path.display());
```
conradgrobler, line 57: This seems strange. AFAICT it will rerun build.rs if the output of the bazel build changed. I assume this should be rerun if any of the TFLite source code changed.

oak_tensorflow_service/src/tflite.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    pub fn run(&self, input_bytes: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let input_bytes_len = input_bytes
+            .len()
+            .try_into()
+            .map_err(|error| anyhow!("Failed to convert usize to u64: {}", error))?;
+        // TODO(#3297): Allocate enough bytes for the TensorFlow Lite model.
+        let mut output_bytes = vec![0; input_bytes.len()];
+        let mut output_bytes_len = vec![0; core::mem::size_of::<usize>()];
```
conradgrobler, line 66: I don't think this should be the size of a usize here, as `tflite_run` explicitly expects a mutable pointer to a u64. Also, instead of creating a vector here you could just create a u64 and pass a mutable pointer to it, e.g:

```
let mut output_bytes_len = 0u64;
let pointer = &mut output_bytes_len as *mut u64;
``` 

If you need it to be on the heap you can create a `Box<u64>` instead.

conradgrobler (approved):

cc/tflite_micro/tflite_micro.cc
ipetr0v: Done

oak_tensorflow_service/build.rs
ipetr0v: Done

oak_tensorflow_service/build.rs
ipetr0v: Done
Added `println!("cargo:rustc-link-search={}", TFLITE_LIBRARY_DIR.display());`

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done
I think it being on the stack is ok, because we won't remove the stack until the `tflite.run()` returns.
Unless it will be a buffer overflow =)

andrisaar (approved):

oak_tensorflow_service/build.rs
```diff
@@ -14,7 +14,22 @@
 // limitations under the License.
 //
 
+use lazy_static::lazy_static;
+use std::{path::PathBuf, process::Command};
+
 const SCHEMA: &str = "schema.fbs";
+const TFLITE_BUILD_TARGET: &str = "//cc/tflite_micro:tflite-micro";
+
+lazy_static! {
```
tiziano88, line 23: We don't use lazy_static any more, I think https://doc.rust-lang.org/nightly/std/cell/struct.LazyCell.html should work here?

ipetr0v: Removed `lazy_static`

PR #3302, opened 2022-10-11
Expose a memory translator function in rust-hypervisor-firmware-virtio.

andrisaar:
As with many other crates, the assumption that virtual addresses are the same as physical was baked deep into the virtio crate.

This PR cleans up the code by using `PhysAddr` and `VirtAddr` where appriopriate (instead of naked `u64`-s), and exposes a translator function to the API users in case we're not running under identity-mapped memory.

For now, the assumption (of running under identity mapping) is still true, so in the `virtio` crate we just pass in a dummy translator. But this gives us a hook to use later when we move away from identity mapping.

### DIFF
--- a/third_party/rust-hypervisor-firmware-virtio/src/lib.rs
+++ b/third_party/rust-hypervisor-firmware-virtio/src/lib.rs
@@ -8,7 +8,12 @@
 
 #![no_std]
 
+use x86_64::{PhysAddr, VirtAddr};
+
 pub mod device;
 pub mod mem;
 pub mod pci;
 pub mod virtio;
+
+pub trait Translator: Fn(PhysAddr) -> Option<VirtAddr> {}
+impl<X: Fn(PhysAddr) -> Option<VirtAddr>> Translator for X {}
--- a/experimental/virtio/src/console/mod.rs
+++ b/experimental/virtio/src/console/mod.rs
@@ -25,7 +25,7 @@ use rust_hypervisor_firmware_virtio::{
     pci::{find_device, VirtioPciTransport},
     virtio::VirtioTransport,
 };
-use x86_64::PhysAddr;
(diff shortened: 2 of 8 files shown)


conradgrobler (approved):
We have to be a bit careful here. I *think* the MMIO ranges will probably always be identity mapped, so we need to make sure we keep track of the appropriate translation functions when we implement ones that are not identity mapped.

andrisaar:
> We have to be a bit careful here. I _think_ the MMIO ranges will probably always be identity mapped, so we need to make sure we keep track of the appropriate translation functions when we implement ones that are not identity mapped.

Yes, the overarching assumption here is that _before_ we start setting up any of this stuff, we have loaded the correct page tables (with the directly mapped memory wherever it is), and that the mapping will not ever change from that point onward.

third_party/rust-hypervisor-firmware-virtio/src/lib.rs
```diff
@@ -8,7 +8,12 @@
...
 
+use x86_64::{PhysAddr, VirtAddr};
+
 pub mod device;
 pub mod mem;
 pub mod pci;
 pub mod virtio;
+
+pub trait Translator: Fn(PhysAddr) -> Option<VirtAddr> {}
+impl<X: Fn(PhysAddr) -> Option<VirtAddr>> Translator for X {}
```
tiziano88, line 19: See https://github.com/project-oak/oak/pull/3306#discussion_r992554251

tiziano88 (approved):

third_party/rust-hypervisor-firmware-virtio/src/lib.rs
andrisaar: As mentioned there simple type aliases didn't work, so I had to do this two-step workaround.

PR #3303, opened 2022-10-11
Add wrappers for IO ports and related factories

conradgrobler:
This change adds wrapper enums around the IO port implementations (raw and GHCB-based) and the related IO port factories. This will be used in follow-up PRs to simplify the drivers that depend on these port implementations.

### DIFF
--- a/experimental/sev_guest/src/io.rs
+++ b/experimental/sev_guest/src/io.rs
@@ -17,7 +17,8 @@
 use crate::ghcb::{Ghcb, GhcbProtocol};
 use core::marker::PhantomData;
 use lock_api::{Mutex, RawMutex};
-use x86_64::instructions::port::{PortReadOnly, PortWriteOnly};
+use spinning_top::{RawSpinlock, Spinlock};
+use x86_64::instructions::port::Port;
 
 /// Factory for instantiating IO port readers and writers.
 ///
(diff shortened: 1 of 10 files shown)


experimental/sev_guest/src/io.rs
```diff
@@ -165,43 +166,118 @@ where
 /// Factory for creating port reader and writers that perform direct raw IO operations.
```
mariaschett, line 166: Do we mean the same by "direct" and "raw"? If so, it we might stick to one of them. (also below)

experimental/sev_guest/src/io.rs
```diff
@@ -165,43 +166,118 @@ where
...
 
-impl PortWriter<u32> for PortWriteOnly<u32> {
+impl PortWriter<u32> for Port<u32> {
     unsafe fn try_write(&mut self, value: u32) -> Result<(), &'static str> {
         self.write(value);
         Ok(())
     }
 }
+
+/// An IO port reader and writer implementation that uses the GHCB protocol, static references and a
```
mariaschett, line 210: "GHCB protocol" - is this the GHCB IOIO protocol? Do we have a reference to the GHBC IOIO protocol?

mariaschett (approved):
Thank you!

experimental/sev_guest/src/io.rs
conradgrobler: Yes, this is the GHCB IOIO protocol. Updated the comment to clarify this. We have links to the GHCB IOIO protocol specification in the `ghcb` module where `Ghcb` and `GhcbProtocol` are defined. I don't think we need to add those links to every comment where the protocol is mentioned.

experimental/sev_guest/src/io.rs
conradgrobler: Direct and raw imply roughly the same thing. I updated the comment to be consistent with the usage throughout the rest of the file: "raw" refers to the port implementation that does "direct" IO.

conradgrobler (commented):
Thanks

PR #3306, opened 2022-10-11
Expose the memory translation function in simple_io crate.

andrisaar:
The main goal here was to expose the memory translation function to the users of the crate; for now, we still assume identity mapping in the kernel.

However, as I was in there, `cargo clippy` deemed a function with 9 arguments as too much, so I refactored how the buffer I/O port descriptors are handled into a separate struct. I think it looks cleaner this way, so can't say I disagree with `cargo clippy` in the end...

### DIFF
--- a/experimental/oak_baremetal_simple_io/src/lib.rs
+++ b/experimental/oak_baremetal_simple_io/src/lib.rs
@@ -27,23 +27,29 @@ use x86_64::{
     PhysAddr, VirtAddr,
 };
 
-/// The default I/O port to use for the most significant bytes of the output buffer guest-physical
-/// address.
-pub const DEFAULT_OUTPUT_BUFFER_MSB_PORT: u16 = 0x6421;
-/// The default I/O port to use for the least significant bytes of the output buffer guest-physical
-/// address.
-pub const DEFAULT_OUTPUT_BUFFER_LSB_PORT: u16 = 0x6422;
-/// The default I/O port to use for the length of output messages.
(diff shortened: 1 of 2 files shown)


experimental/oak_baremetal_simple_io/src/lib.rs
```diff
@@ -53,6 +59,10 @@ pub const INPUT_BUFFER_LEGNTH: usize = 4096;
 /// A Simple IO device implementation that uses direct port-based IO.
 pub type RawSimpleIo<'a> = SimpleIo<'a, PortReadOnly<u32>, PortWriteOnly<u32>>;
 
+/// Memory address translation function.
+pub trait Translator: Fn(VirtAddr) -> Option<PhysAddr> {}
```
tiziano88, line 63: Alternatively this may just be a type alias, then I think you can avoid the blanket impl below.

tiziano88 (approved):

experimental/oak_baremetal_simple_io/src/lib.rs
andrisaar: That was my first attempt, it didn't work.

andrisaar: The trick I used here is from this post: https://users.rust-lang.org/t/why-cant-type-aliases-be-used-for-traits/10002

conradgrobler (approved):

PR #3307, opened 2022-10-11
Expose virtio memory translation functions to the kernel

andrisaar:
This PR builds on #3302 and is similar in purpose to #3306: move the memory mapping functions to the kernel.

As in that PR, for now we continue to assume identity mapping. But at least after this the kernel is in control, and we can switch it out in one single place.

### DIFF
--- a/experimental/virtio/src/lib.rs
+++ b/experimental/virtio/src/lib.rs
@@ -22,6 +22,8 @@
 #![no_std]
 #![feature(let_chains)]
 
+use x86_64::{PhysAddr, VirtAddr};
+
 extern crate alloc;
 #[cfg(test)]
 extern crate std;
@@ -55,3 +57,9 @@ pub trait Write {
 
 /// The vendor ID for virtio PCI devices.
 const PCI_VENDOR_ID: u16 = 0x1AF4;
+
+/// Memory address translation functions.
+pub trait Translator: Fn(VirtAddr) -> Option<PhysAddr> {}
+impl<X: Fn(VirtAddr) -> Option<PhysAddr>> Translator for X {}
+pub trait InverseTranslator: Fn(PhysAddr) -> Option<VirtAddr> {}
+impl<X: Fn(PhysAddr) -> Option<VirtAddr>> InverseTranslator for X {}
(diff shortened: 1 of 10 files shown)


experimental/virtio/src/lib.rs
```diff
@@ -55,3 +57,9 @@ pub trait Write {
 
 /// The vendor ID for virtio PCI devices.
 const PCI_VENDOR_ID: u16 = 0x1AF4;
+
+/// Memory address translation functions.
+pub trait Translator: Fn(VirtAddr) -> Option<PhysAddr> {}
```
conradgrobler, line 62: nit: the P->V transform was called "Translator" in the base PCI transport implementation (https://github.com/project-oak/oak/blob/main/third_party/rust-hypervisor-firmware-virtio/src/lib.rs#L18). I think this way around is better, so perhaps rename the trait in https://github.com/project-oak/oak/blob/main/third_party/rust-hypervisor-firmware-virtio/ for consistency.

conradgrobler (approved):

experimental/virtio/src/queue/mod.rs
```diff
@@ -35,8 +37,8 @@ pub struct DriverWriteOnlyQueue<const QUEUE_SIZE: usize, const BUFFER_SIZE: usiz
 impl<const QUEUE_SIZE: usize, const BUFFER_SIZE: usize>
     DriverWriteOnlyQueue<QUEUE_SIZE, BUFFER_SIZE>
 {
-    pub fn new() -> Self {
-        let inner = Queue::new(DescFlags::empty());
+    pub fn new<VP: Translator>(translate: VP) -> Self {
```
tiziano88, line 40: I'm still unconvinced of this trait; I would either make it a proper trait with a `translate` method, and then call the object implementing it `translator`, or just define it as a function, in which case I would call the object `translate`. Up to you anyways.

tiziano88 (approved):

experimental/virtio/src/queue/mod.rs
andrisaar: Yeah, you're probably right, we should have a proper translator trait with methods that go both ways. But yet again I'm going to punt it to the pile of future clean-ups :)

PR #3308, opened 2022-10-11
Remove old fuzzing targets

tiziano88:
In #3271 a new fuzz target was added, and in
https://github.com/google/oss-fuzz/pull/8730 the OSS fuzz config was fixed.

Ref #3075


andrisaar (approved):
Deleted code is best code.

tiziano88:
AFAICT, the coverage is picking up the correct target now -- I am looking at https://storage.googleapis.com/oss-fuzz-coverage/oak/reports/20221011/linux/src/oak/fuzz/fuzz_targets/apply_policy.rs.html , which was added in #3271.

@rbehjati could you confirm? I will wait before submitting this

rbehjati:
> AFAICT, the coverage is picking up the correct target now -- I am looking at https://storage.googleapis.com/oss-fuzz-coverage/oak/reports/20221011/linux/src/oak/fuzz/fuzz_targets/apply_policy.rs.html , which was added in #3271.
> 
> @rbehjati could you confirm? I will wait before submitting this

Looks like the new target is being used now. The report is not yet shown in the [UI](https://oss-fuzz.com/fuzzer-stats?group_by=by-day&date_start=2022-10-04&date_end=2022-10-11&fuzzer=libFuzzer_oak_apply_policy&job=libfuzzer_asan_oak&project=oak), so I had not seen this latest version from 20221011.

PR #3311, opened 2022-10-11
Rename channel-related crates

tiziano88:
Rename

- `experimental/oak_baremetal_channel` -> `oak_channel`
- `experimental/oak_baremetal_simple_io` -> `oak_simple_io`

Ref #3294


andrisaar:
I think we can strip the `freestanding` from these crates, as it's as much of a misnomer as `baremetal`.

First, I'd rename `oak_baremetal_simple_io` to just `oak_simple_io` (or `oak_simple_io_driver`) -- it's a device driver, and I assume we're not renaming `experimental/virtio` to `oak_freestanding_virtio`, for example.

Second, the `oak_baremetal_channel` is used by the linux implementations as well, so the `freestanding` is not correct there either.

tiziano88:
Even better, thanks!

FTR, I was using "freestanding" essentially as "no_std-compatible". But I'm happy to drop it from the crate name (and I can remove it from some other crates too at some point).

andrisaar (approved):

PR #3313, opened 2022-10-12
Simplify drivers using the new port wrapper enum

conradgrobler:
The wrapper enums for IO ports and their related factories mean that we can now have simpler drivers that depend on these.


andrisaar (approved):
much cleaner, thanks!

PR #3314, opened 2022-10-12
Add how-to about downloading a binary and its provenance

rbehjati:
Adds documentation explaining how to download the Oak binary and its provenance once a PR is merged to main.


rbehjati:
Notes for review: 
- I am not sure if the `development.md` is the right place for this `how-to`. Let me know if you can think of a better place. 
- Would be nice if someone could follow the steps for a commit of their choice and see if everything works as expected.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
 ```bash
 xtask --logs run-cargo-fuzz --crate-name=loader --target-name=wasm_invoke -- -max_total_time=20
 ```
+
+## Retrieving a previously-built binary and its provenance
+
+Every time a pull-request is merged to `main` the
+`oak_functions_freestanding_bin` binary is built and uploaded to `ent`.
+Similarly, a
+[signed SLSA3+ provenance is generated](https://github.com/slsa-framework/slsa-github-generator/blob/04f1fe0c7b7902c9a95f4c7eef2dc04cf0f8e6a7/internal/builders/generic/README.md)
```
conradgrobler, line 353: super nit: I think "is generated" should be outside of the link.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+```
+
+#### Downloading provenance from GitHub
+
+The provenance and the signature over it can be downloaded directly from GitHub.
+For `oak_functions_freestanding_bin` it is stored under the name
+`oak_functions_freestanding_bin.intoto.jsonl`, as an attachment on the GitHub
```
conradgrobler, line 469: It might be useful to provide a bit more info, such as how to find the specific workflow action intand and its related attachment, or at least a link to the specific workflow action run for the specific commit that this example uses.

conradgrobler (approved):
Nice write-up, thanks.

I think this could be its own document (e.g. docs/provenances.md) with links from the main readme and devlopement.md

I went through the steps for a different PR and it worked fine, except for my unfamiliarity with `go install` and finding the resulting `rekor-cli` binary.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+```bash
+$ curl --output oak_functions_from_ent  https://ent-server-62sa4xcfia-ew.a.run.app/raw/sha256:d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+100 2327k    0 2327k    0     0  3637k      0 --:--:-- --:--:-- --:--:-- 3636k
+```
+
+### Step 2: Download the provenance
+
+The provenance can be downloaded from Rekor or the GitHub actions workflow that
```
tiziano88, line 384: I wonder if we should consider also automatically uploading the provenance to ent at some point, so that we don't even have to interact with rekor at the time when we are trying to download the binary and its provenance. Anyways, let's discuss separately

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+  }
+}
+```
+
+In addition to the provenance (in `Attestation`), the response contains
+additional fields, including the `Body.IntotoObj` object. The `payloadHash` in
+this object is the SHA256 hash of the provenance.
+
+Alternatively, the provenance can be retrieved, as part of a Rekor LogEntry,
+using the Rekor HTTP API:
```
tiziano88, line 447: Which of these two do we need for our workflows, e.g. checking in files into google3?

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"https://slsa.dev/provenance/v0.2","subject":[{"name":"oak_functions_freestanding_bin","digest":{"sha256":"d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc"}}],"predicate":{"builder":{"id":"https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.0"},"buildType":"https://github.com/slsa-framework/slsa-github-generator@v1","invocation":{"configSource":{"uri":"git+https://github.com/project-oak/oak@refs/heads/main","digest":{"sha1":"1b128fb2556e4bdcc4f92552654bfbca9d2fb8c6"},"entryPoint":".github/workflows/provenance.yaml"},"parameters":{},"environment":{ "/* GitHub context */"}},"metadata":{"buildInvocationID":"3230206088-1","completeness":{"parameters":true,"environment":false,"materials":false},"reproducible":false},"materials":[{"uri":"git+https://github.com/project-oak/oak@refs/heads/main","digest":{"sha1":"1b128fb2556e4bdcc4f92552654bfbca9d2fb8c6"}}]}}
+```
+
+#### Downloading provenance from GitHub
+
+The provenance and the signature over it can be downloaded directly from GitHub.
```
tiziano88, line 467: Is this also another alternative to the two above, or is it a different artifact? If these are all alternatives to get to the same data, I suggest clarifying that they are equivalent, and also split the first two; ultimately, a person would only need to follow a single one of these, right?

tiziano88 (approved):

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+provided by `rekor-cli`
+([installation instructions](https://docs.sigstore.dev/rekor/installation/)):
+
+```bash
+$ rekor-cli search --sha d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc
+Found matching entries (listed by UUID):
+24296fb24b8ad77a872690e11780e927d9eddd1bc3a598e5490ad1c75fe039289a5dccc5b4d71576
+```
+
+Usually there is only one entry, but it is possible to get multiple entires.
```
mariaschett, line 401: entires -> entries

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+Alternatively, one can use the SHA1 Git commit hash to find the provenances
+corresponding to that commit hash:
+
+```bash
+$ rekor-cli search --sha 1b128fb2556e4bdcc4f92552654bfbca9d2fb8c6
+Found matching entries (listed by UUID):
+24296fb24b8ad77a872690e11780e927d9eddd1bc3a598e5490ad1c75fe039289a5dccc5b4d71576
+```
+
+Note that `rekor-cli` by default uses https://rekor.sigstore.dev as the Rekor
```
mariaschett, line 411: Hmm... Why do I need to note this (at this point)?

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+    }
+  ]
+}
+```
+
+TODO: Describe the relation between the cert in DSSE format, and the public key
```
mariaschett, line 487: Is this a TODO(#) which should get an issue?

mariaschett (approved):
Great! Thanks!

As Conrad has already checked, I will run it through a bit later and let you know if I ran into any problems.

nit: I think you could sometimes use active instead of passive voice, e.g., "Every time a pull-request is merged" -> "Every time we merge a pull-request". But that might not be consistent with the remaining documentation.

docs/development.md
rbehjati: Thanks. Fixed.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+Similarly, a
+[signed SLSA3+ provenance is generated](https://github.com/slsa-framework/slsa-github-generator/blob/04f1fe0c7b7902c9a95f4c7eef2dc04cf0f8e6a7/internal/builders/generic/README.md)
+for it and uploaded to GitHub. The signed provenance is also uploaded to the
+public instance of Rekor hosted by Sigstore at https://rekor.sigstore.dev, and
+can be downloaded from it.
+
+This section describes how the binary and its provenance can be downloaded.
+
+Once a pull-request is merged to `main` and all CI steps are executed on branch
+`main`, an automated comment is added to the pull-request containing the hash of
```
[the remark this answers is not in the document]
rbehjati: Done.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+
+### Step 1: Download the binary
+
+The hash of the binary
+(`d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc`) can be used
+to download the binary from `ent`, with the following command using the
+[ent HTTP API](https://github.com/google/ent#raw-http-api):
+
+```bash
+$ curl --output oak_functions_from_ent  https://ent-server-62sa4xcfia-ew.a.run.app/raw/sha256:d059c38cea82047ad316a1c6c6fbd13ecf7a0abdcc375463920bd25bf5c142cc
```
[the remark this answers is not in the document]
rbehjati: I noticed we use both, but often without `$`. I personally prefer `$` as it makes the context clear.

docs/development.md
rbehjati: I think that is a good idea. We have to add a few steps to the GitHub workflow for this. Created #3339. I've also created #3333 to update the docs once the workflow is updated.

docs/development.md
```diff
@@ -344,3 +344,145 @@ respectively.
...
+
+### Step 2: Download the provenance
+
+The provenance can be downloaded from Rekor or the GitHub actions workflow that
+generated the binary and its provenance (i.e.,
+[Build Provenance](https://github.com/project-oak/oak/actions/workflows/provenance.yaml))
+
+#### Download from Rekor
+
+To fetch the provenance from Rekor, one has to first find the Rekor LogEntries
```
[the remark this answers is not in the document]
rbehjati: Good idea. Done. PTAL.

docs/development.md
rbehjati: Thanks. Fixed.

docs/development.md
rbehjati: Better clarified the point in the new version.

docs/development.md
rbehjati: Restructured the text so it should be clearer what is needed and what is optional. 

Once the workflow is updated, we should be able to download everything from Ent. 
In particular, the uploaded artifacts on GitHub expire after a few days and are removed. So we do need a more permanent storage.

docs/development.md
rbehjati: Thanks. Done.

rbehjati (commented):
Thanks for the reviews  and the helpful comments. Special thanks to @conradgrobler for verifying the instructions :)

PR #3315, opened 2022-10-12
Populate GDT in Restricted Kernel

andrisaar:
After we removed the identity-mapped memory, our interrupt handlers broke.

This was bit of a mystery, but boils down to the fact that when an interrupt happens the processor will try to switch segments to the one described in the IDT. However, we weren't setting up a proper new GDT in the kernel but rather relying on the one set up by the bootloader for us, which meant the GDT was somewhere in low memory that was no longer mapped.

All of our existing code worked just fine, because we don't switch between segments during normal execution. However, once an interrupt happened, the CPU triple-faulted as it couldn't find the GDT.

This PR sets up a new GDT, managed by our kernel. I've also used this opportunity to set up user code/data segments and a proper task state segment as well, although we're not using them (yet).

### DIFF
--- a/oak_restricted_kernel/src/descriptors.rs
+++ b/oak_restricted_kernel/src/descriptors.rs
@@ -1,0 +1,69 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


oak_restricted_kernel/src/descriptors.rs
```diff
@@ -0,0 +1,68 @@
...
+        let mut descriptors = Descriptors {
+            gdt: GlobalDescriptorTable::new(),
+            kernel_cs_selector: SegmentSelector::NULL,
+            kernel_ds_selector: SegmentSelector::NULL,
+            user_cs_selector: SegmentSelector::NULL,
+            user_ds_selector: SegmentSelector::NULL,
+            tss_selector: SegmentSelector::NULL,
+        };
+        descriptors.kernel_cs_selector =
+            descriptors.gdt.add_entry(Descriptor::kernel_code_segment());
```
ipetr0v, line 47: Where do segment values come from?

ipetr0v (approved):

conradgrobler (approved):

Issue #3079, opened 2022-07-22
Implement constant time policy support freestanding Oak Functions runtime


tiziano88:
Also allow configuring it via the init message in the IDL:

https://github.com/project-oak/oak/blob/dd2dc6b1cae87e8210a4f532d040fb70d27f69ca/experimental/oak_baremetal_channel/schema.fbs#L23-L25

Note this is qualitatively harder than #3078 because it requires figuring out how to deal with timers in the bare metal environment.

tiziano88:
Yes, I meant if there is a task tracking the multi CPU support at least for now.

Although it may also be possible to emulate something similar using interrupts maybe? But I'm not sure if it's even worth looking into.

conradgrobler:
I don't think using interrupts would be worth it at the moment. I created a task for supporting additional vCPUs.

tiziano88:
I'm thinking of an alternative approach that does not rely on timers in the TEE: instead of synchronously responding to the request from the untrusted launcher, we could separate the interaction for each client request into two parts (each modeled as an RPC from the untrusted launcher):

- initiate processing: send the contents of the (encrypted) user request, immediately receive a confirmation that processing started
- fetch result (and interrupt processing if still running): at a later point (controlled by the untrusted launcher), send a request asking to retrieve the current result of the computation (by session id, or some other id); this returns either the actual (encrypted) response, or an (encrypted) error message if the actual response is not ready yet

conradgrobler:
It should work, but at the moment the untrusted launcher treats messages from clients as opaque blobs that it just proxies through. This approach would mean that the untrusted launcher must understand the communications between the client and the trusted runtime to some extent. At the very least this would require that the untrusted launcher can distinguish a handshake message from an invocation message, which I don't think is always possible with our current protocol. The trusted runtime keeps state about each session (whether the handshake is in-progress or completed) to help with this, but the untrusted launcher does not have access to this state. The simplest way to achieve this would be to add a fixed byte to the start of every invocation message so that it can be easily detected (in the same way that each step of the handshake is identified by an initial byte).

tiziano88:
This issue is still relevant (though not super high priority).

Now that @mariaschett updated wasmi to a recent version, I think we could use the fuel mechanism to be able to get back execution from a running Wasm module (rather than letting it run forever, e.g. if it gets stuck in an infinite loop).

See https://docs.rs/wasmi/0.29.0/wasmi/struct.Config.html#method.consume_fuel

PR #3317, opened 2022-10-12
Use split Oak Functions server for tests

tiziano88:
The overall goal is to stop depending on the unified Oak Functions
loader library, so that it may be deleted.

Also remove some of the tests that rely on constant processing time,
since that's not supported in the split binary yet (https://github.com/project-oak/oak/issues/3079).

Incidentally, I made the integration tests more aligned with actually
testing meaningful integrations, and I am starting to think that they
can replace running the examples at some point in the future (so we
drive the examples from proper Rust tests, instead of doing that from xtask).

Ref https://github.com/project-oak/oak/issues/3081

### DIFF
--- a/oak_functions/examples/key_value_lookup/module/src/tests.rs
+++ b/oak_functions/examples/key_value_lookup/module/src/tests.rs
@@ -15,85 +15,40 @@
 
 extern crate test;
 
-use hashbrown::HashMap;
 use maplit::hashmap;
-use oak_functions_abi::{proto::ServerPolicy, Request, StatusCode};
-use oak_functions_loader::{
-    grpc::{create_and_start_grpc_server, create_wasm_handler},
-    logger::Logger,
-    lookup_data::{LookupDataAuth, LookupDataRefresher, LookupDataSource},
-    server::WasmHandler,
-};
-use oak_functions_lookup::{LookupDataManager, LookupFactory};
-use std::{
-    net::{Ipv6Addr, SocketAddr},
-    sync::Arc,
-    time::Duration,
-};
+use oak_functions_abi::StatusCode;
+use std::time::Duration;
 use test::Bencher;
 use test_utils::make_request;
 
 #[tokio::test]
 async fn test_server() {
+    let wasm_path = test_utils::build_rust_crate_wasm("key_value_lookup").unwrap();
+
+    let lookup_data_file =
+        test_utils::write_to_temp_file(&test_utils::serialize_entries(hashmap! {
+            b"key_0".to_vec() => b"value_0".to_vec(),
+            b"key_1".to_vec() => b"value_1".to_vec(),
+            b"key_2".to_vec() => b"value_2".to_vec(),
+            b"empty".to_vec() => vec![],
+        }));
+
     let server_port = test_utils::free_port();
-    let address = SocketAddr::from((Ipv6Addr::UNSPECIFIED, server_port));
+    let server_background = test_utils::create_and_start_oak_functions_server(
+        server_port,
+        &wasm_path,
+        lookup_data_file.path().to_str().unwrap(),
+    )
+    .unwrap();
 
-    let mut manifest_path = std::env::current_dir().unwrap();
-    manifest_path.push("Cargo.toml");
-
-    let wasm_module_bytes =
-        test_utils::compile_rust_wasm(manifest_path.to_str().expect("Invalid target dir"), false)
-            .expect("Couldn't read Wasm module");
-
-    let mock_static_server = Arc::new(test_utils::MockStaticServer::default());
-
-    let mock_static_server_clone = mock_static_server.clone();
-    let static_server_port = test_utils::free_port();
-    let mock_static_server_background = test_utils::background(|term| async move {
-        mock_static_server_clone
-            .serve(static_server_port, term)
-            .await
-    });
-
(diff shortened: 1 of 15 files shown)


oak_functions/examples/key_value_lookup/module/src/tests.rs
```diff
@@ -15,85 +15,40 @@
...
-
-    let wasm_handler =
-        create_wasm_handler(&wasm_module_bytes, vec![lookup_factory], logger.clone())
-            .expect("could not create wasm_handler");
-
-    let server_background = test_utils::background(|term| async move {
-        create_and_start_grpc_server(&address, wasm_handler, policy.clone(), term, logger).await
-    });
+    // Wait for the server to start up.
+    std::thread::sleep(Duration::from_secs(2));
```
mariaschett, line 45: Is there any way we could wait for `create_and_start_oak_functions_server` instead of `sleep`ing?

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -180,6 +189,75 @@ where
...
+    .dir(env!("WORKSPACE_ROOT"))
+    .run()
+    .context(format!("could not compile {crate_name}"))?;
+    Ok(format!(
+        "{}target/wasm32-unknown-unknown/release/{crate_name}.wasm",
+        env!("WORKSPACE_ROOT"),
+    ))
+}
+
+pub fn create_and_start_oak_functions_server(
```
mariaschett, line 230: Maybe some documentation would help here?

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -180,6 +189,75 @@ where
...
+        "native",
+        format!("--app-binary={oak_functions_linux_fd_bin_path}"),
+    )
+    .dir(env!("WORKSPACE_ROOT"))
+    .reader()
+    .context("could not run oak_functions_launcher")?;
+    Ok(handle)
+}
+
+pub fn kill_process(handle: duct::ReaderHandle) {
```
mariaschett, line 252: Can we add some documentation on what we expect to be killed afterwards?

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -180,6 +189,75 @@ where
...
+    .context(format!("could not compile {crate_name}"))?;
+    Ok(format!(
+        "{}target/x86_64-unknown-linux-musl/release/{crate_name}",
+        env!("WORKSPACE_ROOT")
+    ))
+}
+
+/// Builds the crate identified by the given name as a Wasm module, and returns the path of the
+/// resulting binary.
+pub fn build_rust_crate_wasm(crate_name: &str) -> anyhow::Result<String> {
```
mariaschett, line 213: Is there a difference between `build_rust_crate_wasm` and `build_rust_crate_binary` apart from the target? Should we maybe combine them if not with "build_rust_crate(name, target)"? That would get a simpler name too.

oak_functions/sdk/test_utils/src/lib.rs
```diff
@@ -180,6 +189,75 @@ where
     }
 }
 
+/// Builds the crate identified by the given name as a Linux binary, and returns the path of the
```
mariaschett, line 192: I find it always a bit hard to find out which "name" we to use and if there are any assumptions on it. I believe e.g., it has to be reachable from "WORKSPACE_ROOT" (maybe it is also dependent on our workspace setup?). 

Anyway: this would be an example for the name, or am I wrong?

https://github.com/project-oak/oak/blob/4228c6b87ce377cf8c595535dd5932c822170441/oak_functions_freestanding/Cargo.toml#L2 

So maybe "identified by "name" from Cargo.toml" would help? (Also below for `build_rust_crate_linux`)

mariaschett (approved):
Thanks!

I didn't in detail check how all the tests map and if something was lost. Let me know, if you want me to do that!

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: Done.

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: Done.

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: The output files generated in the two cases have different suffixes, so I think making this a single function that just branches based on the target in order to do one of two different things would not really buy us much. But maybe in the future if we need to compile to even more targets that may make sense?

oak_functions/sdk/test_utils/src/lib.rs
tiziano88: Done, though I think it should be clear what a crate name is. I also clarified the workspace relationship now, thanks!

oak_functions/examples/key_value_lookup/module/src/tests.rs
tiziano88: I don't think so, this is an opaque child process listening on the given port. I guess we could try and probe that port in a loop until it's open, then close that connection and allow the rest of the code to connect to it at that point, but it would still involve some polling / sleeping anyways.

tiziano88 (commented):
Thanks Maria

PR #3323, opened 2022-10-13
Replace the #GP handler in hello world test with one that emulates `rdmsr` properly.

andrisaar:
This trashes the RAX register, as we use it for our own purposes, but (a) if the instruction that triggered the #GP was indeed `rdmsr`, we want to overwrite RAX anyway and (b) if the instruction wasn't `rdmsr`, we'll crash as that shouldn't happen, so overwriting RAX doesn't matter.


conradgrobler (approved):

tiziano88 (approved):
Thanks, much nicer now.

PR #3324, opened 2022-10-13
Create a `#GP` handler that emulates `rdmsr` in Restricted Kernel.

andrisaar:
This is similar to #3323 where we introduced a similar handler in the hello world kernel, but a bit more clever.

As we can't change register values when using the `x86-interrupt` calling convention, the naked interrupt handler checks if we're dealing with a `rdmsr`. If yes, it emulates `rdmsr` by setting RAX and RDX to zero, skipping the `rdmsr`, and continuing execution.

If, however, the `#GP` as triggered by something else than `rdmsr`, we make sure to _not_ overwrite RAX and pass the control on to the interrupt handler written in Rust, so that it could do whatever it pleases.

All of this is necessary so that we could read the SEV_STATUS MSR on machines that may not support SEV, such as Github Actions or Kokoro, without crashing.

### DIFF
--- a/oak_restricted_kernel/src/interrupts.rs
+++ b/oak_restricted_kernel/src/interrupts.rs
@@ -15,7 +15,7 @@
 //
 
 use crate::i8042;
-use core::ops::Deref;
+use core::{arch::asm, ops::Deref};
 use lazy_static::lazy_static;
 use log::error;
 use x86_64::{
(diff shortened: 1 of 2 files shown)


conradgrobler (approved):
Nice

tiziano88 (approved):

PR #3330, opened 2022-10-14
Add macro to generate interrupt handlers

conradgrobler:
This change adds a (currently untested) macro for generating interrupt handlers that can modify the values of some general purpose registers. The x86-interrupt calling convention in Rust does not allow this, and we need to do it to implement the VMM Communications exception handler for intercepted CPUID calls on AMD SEV-SNP.

The actual use of this macro will be done in follow-up PRs.


experimental/sev_guest/src/interrupts.rs
```diff
@@ -0,0 +1,189 @@
...
+//! ```
+
+use x86_64::VirtAddr;
+
+/// A reference to the interrupt stack frame will be passed as the first argument of the inner
+/// handler function.
+///
+/// It will point to the backed-up values of registers on the stack. This makes it possible to
+/// modify the register values that will be restored when the outer handler returns.
+pub struct InterruptStackFrame {
```
andrisaar, line 60: should this be `repr(C)`?

experimental/sev_guest/src/interrupts.rs
```diff
@@ -0,0 +1,189 @@
...
+                "pop rbx",
+                "pop rcx",
+                "pop rdx",
+                "pop rdi",
+                "pop rsi",
+                // The stack should now be in the original state minus the error code, so return
+                // from the handler.
+                "iretq",
+                INNER_ADDRESS = sym inner_function,
+                options(noreturn)
```
andrisaar, line 183: minor nit, but I think we've used AT&T syntax everywhere else, so I think it would be good to maintain consistency and have this in AT&T and not Intel syntax as well. (Or translate everything else into Intel syntax, I'm also fine with that.)

andrisaar (approved):

ipetr0v (approved):

experimental/sev_guest/src/interrupts.rs
conradgrobler: Yes, it should indeed be. Done.

experimental/sev_guest/src/interrupts.rs
conradgrobler: The rest of this crate also uses Intel syntax, although it would mostly look the same for most of the other cases in this crate.

I personally really dislike AT&T syntaxt, especially since I think something like `[rsp + 13*32]` is must more readable than something like `0x1a0(%rsp)`.

But I agree with the sentiment that we should be consistent. I have created #3335 to track moving to a consistent syntax.

PR #3334, opened 2022-10-14
Remove fprintf from TFLM debug logging

ipetr0v:
This PR adds a patch to TFLM that removes the use of `fprintf`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -200,6 +200,7 @@ http_archive(
     name = "com_github_tensorflow_tflite_micro",
     patches = [
         # Replaces debug logging function with an empty stub.
+        # TODO(#3297): Replace TFLM logging with Restricted Kernel logging.
         "//third_party/tflite-micro:remove-debug-logging.patch",
     ],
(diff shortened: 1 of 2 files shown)


tiziano88 (approved):

andrisaar (commented):
Add some details to the PR description. Why are you patching this instead of defining `TF_LITE_STRIP_ERROR_STRINGS`?

andrisaar:
> > Add some details to the PR description. Why are you patching this instead of defining TF_LITE_STRIP_ERROR_STRINGS?
> 
> Can defining it in `copts` be propagated to all of the dependencies?

I don't now. If all of it is built using Bazel, I wouldn't be surprised if it did.

On the other hand: don't take this as a statement that you _must_ disable logging this way. For example, if your intent is to replace the `fprintf` with something else in the future (e.g. something that calls into our Rust code), then doing it by just removing the fprintf is fine.

ipetr0v:
Added a TODO for replacing TFLM logging with restricted kernel logging.

PR #3337, opened 2022-10-14
Move more Oak Functions crates up

tiziano88:
Ref #3075 #2533


mariaschett (approved):
Looks good to me!

I am not sure if anyone is working on any of these at the moment and how this move affects them, but I trust you have thought about that.

PR #3340, opened 2022-10-17
Add initial #VC exception handler

conradgrobler:
For now we just want to log if the VMM communication exception happens. Actually handling this when CPUID is called and returning a valid response will be done later once we are able to read the location of the CPUID page from the boot parameter in the kernel.


ipetr0v (approved):

andrisaar (approved):
It's unfortunate we need the roundtrip via `VirtAddr` and `set_handler_addr`, but cleaning that up would mean reimplementing `InterruptDescriptorTable`, so it probably isn't worth it (for now).

conradgrobler:
> It's unfortunate we need the roundtrip via `VirtAddr` and `set_handler_addr`, but cleaning that up would mean reimplementing `InterruptDescriptorTable`, so it probably isn't worth it (for now).

I agree. We can revisit this later if we have more reasons in future to reimplement it.

PR #3342, opened 2022-10-17
SEV enligthenment: implement encryption-aware page mapper

andrisaar:
Building page maps is complex enough, and this PR does its best to delegate things to the `x86_64` crate but I still needed to re-implement several bits of it.

The mapper created here still doesn't know how to add or remove the `ENCRYPTED` bit for now; that'll be addressed separately.

I also tried really hard to hoist the encryption status into the type system, but let's just say it got real unwieldy real quick.

The key here is that we want to lie to the mapper from `x86_64` crate: first, we create a `FrameAllocator` that asks the real frame allocator for a frame and sets the encrypted bit in there, if necessary. That encryption-aware FrameAllocator is passed to the `x86_64` OffsetMapper, which is completely oblivious to the fact that something weird is going on with the physical frames.
Conversely, the `OffsetMapper` itself is wrapped in an encryption-aware Mapper, that adds and/or strips the encrypted bit from the physical frame as required.

And the tests are a sight to behold...

### DIFF
--- a/oak_restricted_kernel/src/mm/encrypted_mapper.rs
+++ b/oak_restricted_kernel/src/mm/encrypted_mapper.rs
@@ -17,7 +17,7 @@
 use core::marker::PhantomData;
 use x86_64::{
     structures::paging::{
-        mapper::{MapToError, MapperFlush, PageTableFrameMapping},
+        mapper::{MapToError, MapperAllSizes, MapperFlush, PageTableFrameMapping},
         FrameAllocator, MappedPageTable, Mapper as BaseMapper, Page, PageSize, PageTable,
         PageTableFlags as BasePageTableFlags, PhysFrame, Size4KiB, Translate as BaseTranslate,
     },
@@ -34,7 +34,7 @@ pub enum MemoryEncryption {
     /// Memory encryption uses bit `C` in the page tables.
     ///
     /// The location of the C-bit is machine-dependent; it is reported by CPUID function 8000_001F
-    /// in EBX[5:0].
+    /// in EBX\[5:0\].
     /// See AMD64 Architecture Programmer's Manual, Volume 2, Section 15.34.1 and
     /// AMD64 Architecture Programmer's Manual, Volume 3, Section E.4.17 for more details.
     Encrypted(u8),
@@ -113,7 +113,7 @@ impl From<PageTableFlags> for BasePageTableFlags {
 /// Helper for mapper that assumes all memory is mapped to a given fixed offset.
 ///
 /// See <x86_64::structures::paging::mapper::PageTableFrameMapping> for more details.
-struct PhysOffset {
+pub struct PhysOffset {
     offset: VirtAddr,
     encryption: MemoryEncryption,
 }
@@ -184,13 +184,13 @@ pub trait Mapper<S: PageSize> {
         A: FrameAllocator<Size4KiB>;
 }
 
-pub struct EncryptedPageTable<'a> {
+pub struct EncryptedPageTable<N: MapperAllSizes> {
     encryption: MemoryEncryption,
     offset: VirtAddr,
-    inner: MappedPageTable<'a, PhysOffset>,
+    inner: N,
 }
 
-impl<'a> EncryptedPageTable<'a> {
+impl<'a> EncryptedPageTable<MappedPageTable<'a, PhysOffset>> {
     pub fn new(pml4: &'a mut PageTable, offset: VirtAddr, encryption: MemoryEncryption) -> Self {
         Self {
             encryption,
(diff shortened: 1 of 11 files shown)


conradgrobler (approved):

oak_restricted_kernel/src/mm/mod.rs
```diff
@@ -17,25 +17,42 @@
...
 
-use self::page_tables::DirectMap;
+use self::encrypted_mapper::{EncryptedPageTable, MemoryEncryption, PageTableFlags, PhysOffset};
 
 mod bitmap_frame_allocator;
+mod encrypted_mapper;
 pub mod frame_allocator;
 pub mod page_tables;
 
+pub trait Translate {
```
tiziano88, line 40: Should this trait be called `Translator` instead?

oak_restricted_kernel/src/mm/mod.rs
```diff
@@ -17,25 +17,42 @@
...
+    /// Translates the given virtual address to the physical address that it maps to.
+    ///
+    /// Returns `None` if there is no valid mapping for the given address.
+    fn translate(&self, addr: VirtAddr) -> Option<PhysAddr>;
+
+    /// Translate a physical address to a virtual address.
```
tiziano88, line 46:
```suggestion
    /// Translates a physical address to a virtual address.
```

(etc)

oak_restricted_kernel/src/mm/mod.rs
```diff
@@ -17,25 +17,42 @@
...
+    ///
+    /// Note that a physical address may be mapped multiple times. This function will always return
+    /// the address from the directly mapped region, ignoring ohter mappings if they exist.
+    fn translate_addr(&self, addr: PhysAddr) -> Option<VirtAddr>;
```
tiziano88, line 50: The method names do not seem particularly related to the actual functionality. e.g. it's unclear why `translate` and `translate_addr` (just looking at their names) operate in opposite directions. Maybe they could be called `virtual_to_physical` and `physical_to_virtual` or something like that?

oak_restricted_kernel/src/simpleio.rs
```diff
@@ -14,11 +14,12 @@
 // limitations under the License.
 //
 
-use crate::mm::page_tables::DirectMap;
 use alloc::collections::VecDeque;
 use oak_simple_io::SimpleIo;
 use sev_guest::io::PortFactoryWrapper;
-use x86_64::{structures::paging::Translate, VirtAddr};
+use x86_64::VirtAddr;
+
```
tiziano88, line 21: (nit) keep all the use statements together (etc)

oak_restricted_kernel/src/virtio.rs
```diff
@@ -63,11 +64,11 @@ pub fn get_console_channel(
 }
 
 #[cfg(feature = "vsock_channel")]
-pub fn get_vsock_channel(
-    mapper: &DirectMap,
+pub fn get_vsock_channel<A: Translate>(
+    mapper: &A,
```
tiziano88, line 68: (nit) consistent terminology: mapper vs translator

tiziano88 (approved):

oak_restricted_kernel/src/mm/mod.rs
```diff
@@ -150,9 +167,19 @@ pub fn init_paging<A: FrameAllocator<Size4KiB> + ?Sized>(
...
+    let encrypted = if get_sev_status()
+        .unwrap_or(SevStatus::empty())
+        .contains(SevStatus::SEV_ENABLED)
+    {
+        MemoryEncryption::Encrypted(51)
+    } else {
+        MemoryEncryption::NoEncryption
+    };
+
     // Safety: these page tables are unused (for now) and we have identity mapping for the lowest 2
```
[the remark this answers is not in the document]
andrisaar: I thought about whether `new` should be marked unsafe or not. I think it shouldn't, because it's not the _creation_ of the `EncryptedPageTable` that's dangerous, it's the _modification_ of active page tables that's dangerous.

That comment also holds for the page table structures in `x86_64`; I don't think `new` should be marked as `unsafe` in them.

oak_restricted_kernel/src/mm/encrypted_mapper.rs
```diff
@@ -0,0 +1,632 @@
...
+    unsafe fn map_to_with_table_flags<A>(
+        &mut self,
+        page: Page<S>,
+        frame: PhysFrame<S>,
+        flags: PageTableFlags,
+        parent_table_flags: PageTableFlags,
+        frame_allocator: &mut A,
+    ) -> Result<MapperFlush<S>, MapToError<S>>
+    where
+        A: FrameAllocator<Size4KiB>;
```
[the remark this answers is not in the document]
andrisaar: In here, yes. In general, no.

The mapper needs an allocator for 4K pages in case it needs to allocate new PageTable-s internally, as they are 4K in size.

oak_restricted_kernel/src/mm/mod.rs
andrisaar: It was mostly due to function names in the existing `x86_64` `Translate` trait. But yes, I agree, the names were confusing, so I've renamed them to something hopefully more readable.

PR #3352, opened 2022-10-18
Switch from inclusive ranges to exclusive ranges.

andrisaar:
Ranges are hard.

We're using a mix of inclusive ranges ([start...end], both ends included in range) and exclusive ranges ([start..limit), where the limit is not inside the range).

It's a mess and we should consolidate on one particular range style, and I think it makes sense to use exclusive ranges -- while we still need to do range arithmetic here and there, there's less of it when using exclusive ranges.

I've left the tests largely untouched on purpose: there's some clean-up to do there as well, but for now, I wanted the tests to pass with no (or at least minimal) changes.

### DIFF
--- a/oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
+++ b/oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
@@ -20,7 +20,7 @@ use core::{
     option::Option,
 };
 use x86_64::structures::paging::{
-    frame::PhysFrameRangeInclusive, page::PageSize, FrameAllocator, FrameDeallocator, PhysFrame,
+    frame::PhysFrameRange, page::PageSize, FrameAllocator, FrameDeallocator, PhysFrame,
 };
 
 /// Basic frame allocator implementation that keeps track of PageSize-sized chunks of contiguous
@@ -44,14 +44,14 @@ use x86_64::structures::paging::{
 pub(crate) struct BitmapAllocator<S: PageSize, const N: usize> {
     allocated: BitArray<[u64; N], Lsb0>,
     valid: BitArray<[u64; N], Lsb0>,
-    range: PhysFrameRangeInclusive<S>,
+    range: PhysFrameRange<S>,
 }
 
 impl<S: PageSize, const N: usize> BitmapAllocator<S, N> {
     /// Creates a new bitmap allocator for a physical frame range.
     /// Panics if N does not match the number of u64-s required to track all frames in that range.
     /// Initially, the allocator will mark the whole range as invalid.
-    pub fn new(range: PhysFrameRangeInclusive<S>) -> Self {
+    pub fn new(range: PhysFrameRange<S>) -> Self {
         // Unfortunately there doesn't seem to be a way to hoist this to the type system.
         let expected = bitvec::mem::elts::<u64>(range.count());
         if expected != N {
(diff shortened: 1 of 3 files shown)


oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
```diff
@@ -100,9 +100,9 @@ impl<S: PageSize, const N: usize> BitmapAllocator<S, N> {
             })
             .max_by_key(|(start_idx, end_idx)| end_idx - start_idx)
             .map(|(start_idx, end_idx)| {
-                PhysFrame::range_inclusive(
+                PhysFrame::range(
                     self.frame(start_idx).unwrap(),
-                    self.frame(end_idx).unwrap(),
+                    self.frame(end_idx).unwrap() + 1,
```
tiziano88, line 105: Isn't there a way to also treat `end_idx` itself as exclusive, elsewhere?

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
```diff
@@ -185,18 +182,18 @@ mod tests {
...
         ))
     }
 
     fn create_frame(start: u64) -> PhysFrame<Size4KiB> {
         PhysFrame::from_start_address(PhysAddr::new(start)).unwrap()
     }
 
-    fn create_frame_range(start: u64, end: u64) -> PhysFrameRangeInclusive<Size4KiB> {
-        PhysFrame::range_inclusive(create_frame(start), create_frame(end))
+    fn create_frame_range(start: u64, end: u64) -> PhysFrameRange<Size4KiB> {
```
tiziano88, line 195: Can this method also be modified to accept exclusive end? If not, could you rename it to make it more obvious, or at least add a comment?

tiziano88 (approved):

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
andrisaar: Yes, it can -- and I will do later. As I mentioned in the PR description, I wanted to do minimal changes to the tests in this PR.

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
```diff
@@ -20,7 +20,7 @@ use core::{
     option::Option,
 };
 use x86_64::structures::paging::{
-    frame::PhysFrameRangeInclusive, page::PageSize, FrameAllocator, FrameDeallocator, PhysFrame,
+    frame::PhysFrameRange, page::PageSize, FrameAllocator, FrameDeallocator, PhysFrame,
```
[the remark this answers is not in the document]
andrisaar: This sounds like a good idea, but `PhysFrameRange` comes from the `x86_64` crate, so the implementation would need to go there.

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
andrisaar: The particular algorithm I've implemented here comes up with an inclusive answer. In fact, this is why I originally used inclusive ranges to begin with.

However, (a) inclusive ranges have proved to be a pain elsewhere so it makes sense to concentrate it here, and (b) I suspect this whole `fn` will go away in not too distant future so I'm inclined to leave it as is.

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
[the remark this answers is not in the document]
tiziano88: Good point, maybe it's just made available as long as your underlying type implements Eq? Not sure

PR #3353, opened 2022-10-18
SEV enlightenment: create an allocator for unencrypted memory

andrisaar:
This adds a facility to change page table flags (by default, all memory is marked as `ENCRYPTED`) and then uses it to set up a 2M page for use with guest-host communication.

I've also shuffled some data structures around in the hope of keeping `encrypted_mapper` private.

The allocator is not used for anything (yet) -- that'll be in follow-up PRs. This PR just lays out the infrastructure.

### DIFF
--- a/oak_restricted_kernel/src/lib.rs
+++ b/oak_restricted_kernel/src/lib.rs
@@ -63,7 +63,7 @@ use oak_channel::Channel;
 use oak_linux_boot_params::BootParams;
 use strum::{EnumIter, EnumString, IntoEnumIterator};
 use x86_64::{
-    structures::paging::{Page, Size2MiB},
+    structures::paging::{FrameAllocator, Page, PhysFrame, Size2MiB},
     VirtAddr,
 };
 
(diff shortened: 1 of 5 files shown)


oak_restricted_kernel/src/lib.rs
```diff
@@ -89,11 +89,21 @@ pub fn start_kernel(info: &BootParams) -> Box<dyn Channel> {
     // Physical frame allocator: support up to 128 GiB of memory, for now.
     let mut frame_allocator = mm::init::<1024>(info.e820_table(), program_headers);
 
-    let mapper = mm::init_paging(&mut frame_allocator, program_headers).unwrap();
+    let mut mapper = mm::init_paging(&mut frame_allocator, program_headers).unwrap();
+
+    // Allocate a section for guest-host communication (without the `ENCRYPTED` bit set)
+    let guest_host_frame: PhysFrame<Size2MiB> = frame_allocator.allocate_frame().unwrap();
```
conradgrobler, line 95: For now 2MiB is fine, but we might need a bit more in future if we want to instantiate the virtio vsock in this memory as well, as it uses just over 2MiB for it buffers.

oak_restricted_kernel/src/memory.rs
```diff
@@ -31,14 +32,48 @@ static ALLOCATOR: LockedHeap = LockedHeap::empty();
...
-    info!("Using [{:#016x}..{:#016x}) for heap.", start, limit);
+    info!("Using [{:#018x}..{:#018x}) for kernel heap.", start, limit);
     // This is safe as we know the memory is available based on the e820 map.
     unsafe {
         ALLOCATOR.lock().init(start as *mut u8, limit - start);
     }
     Ok(())
 }
+
+/// Initializes an allocator for guest-heap communication on unencrypted memory.
```
conradgrobler, line 47: nit: "guest-host" or "guest-hypervisor"

conradgrobler (approved):

PR #3355, opened 2022-10-18
Change tests to use exclusive ranges as #3352 has been merged.

andrisaar:
I left the tests as-is in the previous PR to prove to myself that the behaviour of the code is the same; now we can fix the tests properly.

### DIFF
--- a/oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
+++ b/oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
@@ -184,7 +184,7 @@ mod tests {
     fn create_allocator<const N: usize>(start: u64, end: u64) -> BitmapAllocator<Size4KiB, N> {
         BitmapAllocator::<Size4KiB, N>::new(PhysFrame::range(
             PhysFrame::from_start_address(PhysAddr::new(start)).unwrap(),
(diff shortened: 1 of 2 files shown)


oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
```diff
@@ -193,33 +193,33 @@ mod tests {
...
 
     fn create_frame_range(start: u64, end: u64) -> PhysFrameRange<Size4KiB> {
-        PhysFrame::range(create_frame(start), create_frame(end) + 1)
+        PhysFrame::range(create_frame(start), create_frame(end))
     }
 
     #[test]
     fn silly_allocator_invalid() {
-        let mut alloc = create_allocator::<1>(0x0000, 0x0000);
+        let mut alloc = create_allocator::<1>(0x0000, 0x1000);
```
rbehjati, line 201: Not sure what the values mean, but does it make sense to declare some of them as constants, so that the value can be changed in only one place?

rbehjati (approved):

oak_restricted_kernel/src/mm/bitmap_frame_allocator.rs
andrisaar: Those are just arbitrary addresses chosen for the test, they have no particular meaning.

PR #3356, opened 2022-10-18
Update `ringbuf` and change code to match API changes

andrisaar:
Unfortunately dependabot doesn't know how to refactor code :(


tiziano88 (approved):
Thanks, I was in the process of updating this and most other deps (since no one has done in a while), but please go ahead with this first

tiziano88:
If only we had a way of using opaque identifiers instead of human readable symbols in code....

PR #3357, opened 2022-10-19
Get inference output bytes from TensorFlow Lite

ipetr0v:
This PR makes TensorFlow Lite C++ code return a pointer to the buffer containing the inference output bytes.

### DIFF
--- a/cc/tflite_micro/tflite_micro.h
+++ b/cc/tflite_micro/tflite_micro.h
@@ -27,7 +27,7 @@ extern "C" {
 int tflite_init(const uint8_t* model_bytes, size_t model_bytes_len,
                 const uint8_t* tensor_arena_bytes, size_t tensor_arena_bytes_len);
 
-int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t* output_bytes,
+int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t** output_bytes,
                size_t* output_bytes_len);
 
 #ifdef __cplusplus
--- a/oak_tensorflow_service/src/tflite.rs
+++ b/oak_tensorflow_service/src/tflite.rs
@@ -33,10 +33,11 @@ extern "C" {
 
     /// Performs inference on `input_bytes` and pass inference result to `output_bytes` and returns
     /// an error code.
+    /// `output_bytes` are allocated and freed by TensorFlow Lite.
     fn tflite_run(
         input_bytes: *const u8,
         input_bytes_len: u64,
-        output_bytes: *mut u8,
+        output_bytes: *mut *const u8,
         output_bytes_len: *mut u64,
     ) -> i32;
 }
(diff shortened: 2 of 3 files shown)


cc/tflite_micro/tflite_micro.h
```diff
@@ -27,7 +27,7 @@ extern "C" {
 int tflite_init(const uint8_t* model_bytes, size_t model_bytes_len,
                 const uint8_t* tensor_arena_bytes, size_t tensor_arena_bytes_len);
 
-int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t* output_bytes,
+int tflite_run(const uint8_t* input_bytes, size_t input_bytes_len, uint8_t** output_bytes,
```
tiziano88, line 30: I think the out param should be called `output_bytes_ptr` or something like that? Also it needs some documentation if nothing else to explain who owns that memory after calling this function.

oak_tensorflow_service/src/tflite.rs
```diff
@@ -33,10 +33,11 @@ extern "C" {
 
     /// Performs inference on `input_bytes` and pass inference result to `output_bytes` and returns
     /// an error code.
+    /// `output_bytes` are allocated and freed by TensorFlow Lite.
```
tiziano88, line 36: What does this mean? As the caller of this function, do I need to worry about allocation or not?

oak_tensorflow_service/src/tflite.rs
```diff
@@ -80,18 +80,27 @@ impl TfliteModel {
...
             )
         };
-        Ok(output_bytes)
+
+        let output_bytes_len = output_bytes_len_ptr
+            .try_into()
+            .map_err(|error| anyhow!("Failed to get output bytes length from bytes: {}", error))?;
+
+        let output_bytes =
+            unsafe { alloc::slice::from_raw_parts(*output_bytes_ptr, output_bytes_len) };
```
tiziano88, line 103: What happens to the slice after this? AFAICT the vector below is a new allocation in which the bytes are copied, but what about the existing pointer?

tiziano88 (approved):

oak_tensorflow_service/src/tflite.rs
```diff
@@ -80,18 +80,27 @@ impl TfliteModel {
         let input_bytes_len = input_bytes
             .len()
             .try_into()
-            .map_err(|error| anyhow!("Failed to convert usize to u64: {}", error))?;
-        // TODO(#3297): Allocate enough bytes for the TensorFlow Lite model.
-        let mut output_bytes = vec![0; input_bytes.len()];
-        let mut output_bytes_len = 0u64;
+            .map_err(|error| anyhow!("Failed to convert input bytes length to u64: {}", error))?;
+
+        // TensorFlow Lite is the owner of the corresponding `output_bytes` buffer.
```
conradgrobler, line 85: I don't think this is the best design. AFAICT the current implementation leaks the output bytes. I think the buffer should be owned by the Rust code and initialised to a maximum size. If less is used by TF lite it can return a smaller size so that the buffer can be shrunk.

ipetr0v: Current problem is that the Rust code doesn't know in advance what will be the output size for a certain model.

Another solution could be to have an output size as a parameter for the `Init(tf_model, output_size)` IDL call, that initializes the runtime.

oak_tensorflow_service/src/tflite.rs
ipetr0v: Current idea was to point this pointer to the Tensor Arena output in C++ (which is always the same buffer). Because without `malloc` TFLM cannot allocate new buffers at all.

But probably it's safer to allocate everything on the Rust side.

oak_tensorflow_service/src/tflite.rs
conradgrobler: Is it not possible to calculate the output size from the model in TF Lite? In that case the output size could be returned by the TF Lite wrapper either as a result from Init, or as a separate function.

ipetr0v:
Returning output buffer length from TFLite initialization (https://github.com/project-oak/oak/pull/3383)

PR #3360, opened 2022-10-19
Plumb an explicit allocator through oak_simple_io crate.

andrisaar:
We need a custom allocator to ensure that the vectors are allocated in the memory area reserved for guest-host communication; we can't just rely on the system allocator in the future.

Unfortunately this means we now have to track lifetimes throughout the crate to placate the borrow checker. I'm still not _entirely_ convinced it's worth it, but at least in this crate it wasn't much hassle.

### DIFF
--- a/oak_restricted_kernel/src/simpleio.rs
+++ b/oak_restricted_kernel/src/simpleio.rs
@@ -16,26 +16,29 @@
 
 use crate::mm::Translator;
 use alloc::collections::VecDeque;
+use core::alloc::Allocator;
 use oak_simple_io::SimpleIo;
 use sev_guest::io::PortFactoryWrapper;
 use x86_64::VirtAddr;
 
 /// A communications channel using a simple IO device.
-pub struct SimpleIoChannel {
+pub struct SimpleIoChannel<'a, A: Allocator> {
     /// The simple IO device.
-    device: SimpleIo,
+    device: SimpleIo<'a, A>,
     /// A buffer to temporarily store extra data from the device that was not fully read when using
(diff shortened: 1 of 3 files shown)


oak_simple_io/src/lib.rs
```diff
@@ -49,31 +50,34 @@ pub const DEFAULT_INPUT_BUFFER: BufferDescriptor = BufferDescriptor {
 };
 
 /// The length of the buffer that will be used for output messages.
-pub const OUTPUT_BUFFER_LEGNTH: usize = 4096;
+pub const OUTPUT_BUFFER_LENGTH: usize = 4096;
```
conradgrobler, line 53: Thanks for fixing my typos :)

conradgrobler (approved):
Nice, thanks,

oak_restricted_kernel/src/simpleio.rs
```diff
@@ -16,26 +16,29 @@
 
 use crate::mm::Translator;
 use alloc::collections::VecDeque;
+use core::alloc::Allocator;
 use oak_simple_io::SimpleIo;
 use sev_guest::io::PortFactoryWrapper;
 use x86_64::VirtAddr;
 
 /// A communications channel using a simple IO device.
```
mariaschett, line 24: Should we explain why we need the `Allocator` in a comment near the code, too?

IIUC: "can't just rely on the system allocator" and "custom allocator to ensure that the vectors are allocated in the memory area reserved for guest-host communication"

mariaschett (approved):

PR #3361, opened 2022-10-19
Plumb an explicit allocator through the virtio crate.

andrisaar:
We need a custom allocator to ensure that the vectors are allocated in the memory area reserved for guest-host communication; we can't just rely on the system allocator in the future.

Unfortunately this means we now have to track lifetimes throughout the crate to placate the borrow checker. I'm still not entirely convinced it's worth it, but at least in this crate it wasn't much hassle.

This PR mirrors #3360 for the simple_io crate.

### DIFF
--- a/experimental/virtio/src/console/mod.rs
+++ b/experimental/virtio/src/console/mod.rs
@@ -14,6 +14,8 @@
 // limitations under the License.
 //
 
+use core::alloc::Allocator;
+
 use crate::{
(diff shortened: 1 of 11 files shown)


conradgrobler (approved):
One minor thing to note: I think this will break when we try to use it with the current guest-host alloctor because the virtio vsock queues are too big to fit in the 2MiB available to that allocator at the moment.

PR #3362, opened 2022-10-20
Support building fully-static cc libraries

ianull:


cc/build_defs.bzl
```diff
@@ -0,0 +1,91 @@
...
+
+"""Provides a rule that outputs a monolithic static library."""
+
+load(
+    "@bazel_tools//tools/cpp:toolchain_utils.bzl",
+    "find_cpp_toolchain",
+    "use_cpp_toolchain",
+)
+
+# Bazel rule for collecting the header files that a target depends on.
```
conradgrobler, line 25: nit: I don't think that this comment accurately describes everything that is done here.

conradgrobler (approved):
I am not that familiar with build rules of the details of AR scripts, but the implementation looks good as far as I can tell. Thanks.

cc/build_defs.bzl
ipetr0v, line 25: Could you please add a link where this code comes from?

cc/build_defs.bzl
```diff
@@ -0,0 +1,91 @@
...
+def _cc_static_library_impl(ctx):
```
ipetr0v, line 26: How does this rule handle [copts](https://bazel.build/reference/be/c-cpp#cc_library.copts) and [defines](https://bazel.build/reference/be/c-cpp#cc_library.defines) ?

cc/build_defs.bzl
```diff
@@ -0,0 +1,91 @@
...
+            for flag in inp.user_link_flags
+        })
+    link_flags = unique_flags.keys()
+
+    # Collect static libraries
+    libs = []
+    for inp in input_depset.to_list():
+        for lib in inp.libraries:
+            if lib.pic_static_library:
+                libs.append(lib.pic_static_library)
```
ipetr0v, line 52: How does it handle duplicated dependencies?

cc/build_defs.bzl
```diff
@@ -0,0 +1,91 @@
...
+                libs.append(lib.static_library)
+
+    script_file = ctx.actions.declare_file("lib{}.mri".format(ctx.attr.name))
+    commands = ["create {}".format(output_lib.path)]
+    for lib in libs:
+        commands.append("addlib {}".format(lib.path))
+    commands.append("save")
+    commands.append("end")
+    ctx.actions.write(
+        output = script_file,
```
ipetr0v, line 63: In our CI we always check that we don't generate new files in the repo.
Does this rule create a new file that is not tracked by the repository?

tiziano88: These are bazel rules, I think their only output is within the bazel folders, they cannot possibly write to random parts of the host file system (I hope).

cc/build_defs.bzl
tiziano88: It should not matter if the libs are specified multiple times I think.

tiziano88 (approved):

ipetr0v (approved):

andrisaar (approved):

cc/build_defs.bzl
ianull: lol, should be **object files** instead of **header files**.

ianull: np

cc/build_defs.bzl
ianull: It doesn't need to handle copts and defines as it's working as a wrapper :)

Use current codebase as example:
```
cc_library(
    name = "tflite-micro",
    srcs = ["tflite_micro.cc"],
    hdrs = ["tflite_micro.h"],
    linkstatic = 1,
    deps = [
        "@com_github_tensorflow_tflite_micro//tensorflow/lite/micro:micro_framework",
    ],
)

cc_static_library(
    name = "tflite-micro-fully-static",
    deps = [":tflite-micro"],
)
```
The `tflite-micro-fully-static` target searches all required libraries for `tflite-micro`, then uses `ar` to archive all of them into one archive a.k.a. fully static library.
In this case, all *.a generated from tflite-micro as well as @com_github_tensorflow_tflite_micro//tensorflow/lite/micro:micro_framework are `ar` into a fully static `libtflite-micro-fully-static.a`.

cc/build_defs.bzl
ianull: The `unique` magic is done by `depset`: `input_depset = depset(transitive = lib_sets)`

cc/build_defs.bzl
ianull: Like what Tiziano said: the files (intermediates and final static .a lib) the rule generates are put in bazel-bin/**/* and not touching any of host files.

ianull:
How do I get pass the error that is blocking the merge:
`= note: /usr/bin/ld: final link failed: No space left on device`

@tiziano88 could you help clean up device space?

tiziano88:
It looks like that's from a previous run, so hopefully it will succeed now, I think @conradgrobler cleared the cache earlier today, so it should leave more space for the build. But let me know if still fails

PR #3374, opened 2022-10-24
SEV enlightenment: ignore the encrypted bit when writing to memory

andrisaar:
Literally a one-bit change that's required for our Restricted Kernel to boot under SEV.

The problem here is that the address in the page table potentially has the encrypted bit set, but we want to read from that physical address, which won't work with the encrypted bit set. Thus, let's ignore bit 51 for now.


conradgrobler (approved):

tiziano88:
Was this enough to make it boot on the machine where it was previously failing?

andrisaar:
Yes, and no -- it was required to get it to boot at all, but my fights with the page tables continue.

PR #3378, opened 2022-10-25
Upload DSSE document to Ent and print its hash

### LINKED ISSUE #3339: Upload the DSSE to ent in the Build Provenance workflow
I wonder if we should consider also automatically uploading the provenance to ent at some point, so that we don't even have to interact with rekor at the time when we are trying to download the binary and its provenance. Anyways, let's discuss separately

_Originally posted by @tiziano88 in https://github.com/project-oak/oak/pull/3314#discussion_r995624992_

rbehjati:
Towards #3339

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -158,3 +158,38 @@ jobs:
       upload-assets: true
       # Build the generator from source.
       compile-generator: true
+
+  dsse-upload:
+    if: github.event_name == 'push'
+    needs: [provenance]
+    steps:
+      - name: Download DSSE document
+        uses: actions/download-artifact@v3
+        with:
+          name: 'oak_functions_freestanding_bin.intoto.jsonl'
+
+      # See https://github.com/google/ent
+      - name: Download Ent CLI
+        if: steps.cache-ent.outputs.cache-hit != 'true'
+        env:
+          ENT_URL: https://ent-server-62sa4xcfia-ew.a.run.app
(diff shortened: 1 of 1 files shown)


.github/workflows/provenance.yaml
```diff
@@ -158,3 +158,38 @@ jobs:
...
+          ent
+          cat <<EOF > ~/.config/ent.toml
+          [[remotes]]
+          name = 'ent-store'
+          url = '${ENT_URL}'
+          write = true
+          api_key = '${{ secrets.ENT_API_KEY }}'
+          EOF
+
+      # TODO(#3339): publish as a comment
```
mariaschett, line 189: Mhm... I am not sure what you mean by publish as a comment. I might miss some context here.

mariaschett (approved):

.github/workflows/provenance.yaml
rbehjati: I mean the same way we publish the reproducibility index as a comment on the PR. That would make it easier in the future to fetch the DSSE document from Ent. 

I'll fix this immediately after merge :)

rbehjati (commented):
Thanks for the quick review.

.github/workflows/provenance.yaml
rbehjati: I want to publish the sha256sum as a comment. I think retrieving it from the comment would be easier than retrieving it from the logs.

Issue #3377, opened 2022-10-24
cargo deny reports deprecated crate: `parity-wasm`


tiziano88:
```
[20:59:26 ✓:0,✗:0,⠇:7] ││ ║ error[A003]: Crate `parity-wasm` deprecated by the author
[20:59:26 ✓:0,✗:0,⠇:7] ││ ║    ┌─ /workspace/oak_functions_freestanding_bin/Cargo.lock:87:1
[20:59:26 ✓:0,✗:0,⠇:7] ││ ║    │
[20:59:26 ✓:0,✗:0,⠇:7] ││ ║ 87 │ parity-wasm 0.42.2 registry+https://github.com/rust-lang/crates.io-index
[20:59:26 ✓:0,✗:0,⠇:7] ││ ║    │ ------------------------------------------------------------------------ unmaintained advisory detected
```

I looked into it briefly, and it looks like we are using `wasmi` v0.9.1, while the current version is v0.19.0 . Presumably we did not notice this because of the way cargo resolves features, and at the moment we specify a feature that has stopped existing for a while, therefore causing the dependency to effectively be stuck forever at the latest version that had it:

https://github.com/project-oak/oak/blob/dd68b8d9be38590c987b1fec9ffb50761617dd9d/oak_functions/wasm/Cargo.toml#L17

mariaschett:
I think you are right @tiziano88 , however, it seems we need the feature "core" to run wasmi in [no-std mode](https://github.com/project-oak/oak/blob/2a5825bebc8b579c3d56bf20114d41327a9c3c97/oak_functions/wasm/Cargo.toml#L16).

I am not sure we can run wasmi with no-std in v.0.19. 

I can try to fix the compile errors in v.0.19.0, but I am not sure how much that helps if we have no no-std.

Any ideas how to find that out?

cc: @conradgrobler

tiziano88:
AFAICT the issue is that that feature does not exist any more. It looks like just disabling default features should make it no_std compatible, or is that not the cas?

conradgrobler:
I agree with @tiziano88. Looking at the code it seems that it should work fine as long as the default features are not enabled.

tiziano88:
https://github.com/paritytech/wasmi/blob/52c8a0af44a83b065bf30bfc78dc40f300e4abac/crates/wasmi/Cargo.toml#L31-L33

PR #3381, opened 2022-10-25
Temporarily ignore deprecated parity-wasm

mariaschett:


tiziano88 (approved):
Thanks!

tiziano88:
Perhaps mention #3377 in the commit message.

andrisaar (approved):
I think there are other `deny.toml` files for other raw binaries in the repository; please double-check that the dependency is not in them.

PR #3382, opened 2022-10-25
Fix provenance workflow

rbehjati:
- Fixes the currently failing workflow job. 
- Fixes the set-output issue in GitHub workflow
- Adds a workflow step for publishing the output of uploading DSSE to Ent as a comment on the PR.

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -160,8 +160,16 @@ jobs:
       compile-generator: true
 
   dsse-upload:
-    if: github.event_name == 'push'
+    # if: github.event_name == 'push'
     needs: [provenance]
+    runs-on: ubuntu-20.04
+
+    permissions:
+      # Allow the job to update the repo with the latest provenances and index.
+      contents: write
+      # Allow the job to add a comment to the PR.
+      pull-requests: write
+
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/provenance.yaml
```diff
@@ -190,6 +198,40 @@ jobs:
...
 
+    # Also post a reply on the PR thread with the sha256sum of the DSSE artifact.
+    - name: Post SHA256 Hash of the DSSE Artifact (post-merge only)
+      uses: actions/github-script@0.9.0
+      # if: github.event_name == 'push'
+      with:
+        github-token: ${{inputs.GITHUB_TOKEN}}
+        script: |
+          const execa = require('execa')
+          const { stdout } = await execa('sha256sum', ['oak_functions_freestanding_bin.intoto.jsonl'])
```
tiziano88, line 209: How about `tee`ing the output of `ent` to a file, and then just read that file in the next step? Similar to what we do in the repro index case:

https://github.com/project-oak/oak/blob/e75740a22c5c3db3d25031cd8b660b50aa60eda2/.github/actions/reproducibility_index/action.yml#L43

rbehjati:
It is failing again :( 
Trying a fix in #3387.

PR #3383, opened 2022-10-25
Return output buffer length from TFLite initialization

ipetr0v:
This PR makes TFLite initialization function return an output buffer length, so that this buffer could be allocated in Rust before calling TFLite inference.

### DIFF
--- a/oak_tensorflow_service/src/tflite.rs
+++ b/oak_tensorflow_service/src/tflite.rs
@@ -44,8 +44,8 @@ extern "C" {
     /// Memory area that will be used by the TFLM memory manager is defined by `tensor_arena_bytes`:
     /// <https://github.com/tensorflow/tflite-micro/blob/main/tensorflow/lite/micro/docs/memory_management.md>
     ///
-    /// This function returns a buffer size value in the `output_buffer_len`. This buffer should be
-    /// allocated before calling `tflite_run` to store inference results.
+    /// This function returns a buffer size value in the `output_buffer_len_ptr`. This buffer should
+    /// be allocated before calling `tflite_run` to store inference results.
     fn tflite_init(
         model_bytes_ptr: *const u8,
         model_bytes_len: usize,
@@ -54,8 +54,11 @@ extern "C" {
         output_buffer_len_ptr: *mut usize,
     ) -> i32;
 
-    /// Performs inference on `input_bytes` and pass inference result to `output_bytes` and returns
-    /// an error code.
+    /// Performs inference on `input_bytes_ptr` and pass inference result to `output_bytes_ptr` and
+    /// returns an error code.
+    ///
+    /// The value in the `output_bytes_len_ptr` cannot be bigger than the value in
+    /// `output_buffer_len_ptr` from `tflite_init`.
     fn tflite_run(
         input_bytes_ptr: *const u8,
         input_bytes_len: usize,
@@ -88,7 +91,7 @@ impl TfliteModel {
 
     pub fn initialize(&mut self, model_bytes: &[u8]) -> anyhow::Result<()> {
         log!(Level::Info, "Initializing the TFLite model");
-        if let Some(_) = self.output_buffer_len {
+        if self.output_buffer_len.is_some() {
             return Err(anyhow!("TFLite model has already been initialized"));
         }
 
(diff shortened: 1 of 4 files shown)


oak_tensorflow_service/src/tflite.rs
```diff
@@ -66,51 +70,55 @@ const TENSOR_ARENA_SIZE: usize = 1024;
...
-            .len()
-            .try_into()
-            .map_err(|error| anyhow!("Failed to convert usize to u64: {}", error))?;
-        let tensor_arena_len = self
-            .tensor_arena
-            .len()
-            .try_into()
-            .map_err(|error| anyhow!("Failed to convert usize to u64: {}", error))?;
+    pub fn initialize(&mut self, model_bytes: &[u8]) -> anyhow::Result<()> {
+        let model_bytes_len = model_bytes.len();
```
tiziano88, line 86: Shouldn't this method check if the model was already initialized, and fail if so? In fact I think the TFmodel may be better represented as an enum of either Empty or Initialized (or something like that), like a state machine. Or does that make tracking ownership of the tensor arena harder?

oak_tensorflow_service/src/tflite.rs
```diff
@@ -66,51 +70,55 @@ const TENSOR_ARENA_SIZE: usize = 1024;
...
         let _ = unsafe {
             tflite_init(
                 model_bytes.as_ptr(),
                 model_bytes_len,
                 self.tensor_arena.as_ptr(),
                 tensor_arena_len,
+                &mut output_buffer_len,
             )
         };
+        self.output_buffer_len = Some(output_buffer_len);
```
tiziano88, line 98: Should we log something here? It may be useful for debugging

oak_tensorflow_service/src/tflite.rs
```diff
@@ -66,51 +70,55 @@ const TENSOR_ARENA_SIZE: usize = 1024;
...
-            .len()
-            .try_into()
-            .map_err(|error| anyhow!("Failed to convert usize to u64: {}", error))?;
-        // TODO(#3297): Allocate enough bytes for the TensorFlow Lite model.
-        let mut output_bytes = vec![0; input_bytes.len()];
-        let mut output_bytes_len = 0u64;
+    pub fn run(&mut self, input_bytes: &[u8]) -> anyhow::Result<Vec<u8>> {
+        let buffer_len = self
+            .output_buffer_len
+            .context("Running inference on a non-initialized TensorFlow model")?;
```
tiziano88, line 105: Naming seems inconsistent: buffer_len vs input_bytes vs output_bytes vs output_bytes_len vs "Response size" in the message.

tiziano88 (approved):

oak_tensorflow_service/src/tflite.rs
```diff
@@ -66,51 +70,55 @@ const TENSOR_ARENA_SIZE: usize = 1024;
...
                 output_bytes.as_mut_ptr(),
-                &mut output_bytes_len as *mut u64,
+                &mut output_bytes_len,
             )
         };
-        Ok(output_bytes)
+        if output_bytes_len <= buffer_len {
+            Ok(output_bytes[..output_bytes_len].to_vec())
+        } else {
+            Err(anyhow!("Response size is larger than the allocated buffer"))
```
conradgrobler, line 121: I would go as far as panicking in this case. If the c++ code wrote a response larger than the buffer it means there is a potential for memory corruption and we can't rely on any of the Rust memory safety assumptions any more.

conradgrobler (approved):

oak_tensorflow_service/src/tflite.rs
ipetr0v: Currently this state is represented by the `output_buffer_len: Option<usize>`
I think making the whole structure an enum may be not convenient, since all of our Runtimes are just plain structs and are called like so.

oak_tensorflow_service/src/tflite.rs
```diff
@@ -66,51 +70,55 @@ const TENSOR_ARENA_SIZE: usize = 1024;
 #[derive(Default)]
 pub struct TfliteModel {
     tensor_arena: Vec<u8>,
+    output_buffer_len: Option<usize>,
 }
```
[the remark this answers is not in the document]
ipetr0v: Added a comment

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done

oak_tensorflow_service/src/tflite.rs
tiziano88: Not sure I understand why it would be a problem.

Currently they are just pub structs with non-pub fields, so  nothing externally would care whether they are enums or structs. Having one field (called `output_buffer_len`, which seems quite arbitrary) represent the current state, and keeping the other fields in sync via manually enforced invariants seems problematic in the long term.

Perhaps in another PR though?

oak_tensorflow_service/src/tflite.rs
ipetr0v: Done
Now there are only the output buffer and the output bytes length inside this buffer.

PR #3384, opened 2022-10-25
SEV: don't set `GLOBAL` in parent page table entries

andrisaar:
I'm not entirely sure _why_, but Intel CPUs seem fine with the `GLOBAL` flag set but the AMD CPU we're trying things on refused to load page tables with the `GLOBAL` bit set for the parent entries.

Note that the _leaves_ of the page table still have `GLOBAL` set, and the AMD CPU seems fine with that.

Semantically, we do want `GLOBAL`: we're mapping memory in kernel space, which means it should be the same no matter the process. (Even though we don't have processes right now).

Stripping the bit off got the code to boot on the AMD CPU; now I'm off to read the specs as to _why_ exactly this was rejected...

### DIFF
--- a/oak_restricted_kernel/src/mm/page_tables.rs
+++ b/oak_restricted_kernel/src/mm/page_tables.rs
@@ -40,15 +40,16 @@ pub unsafe fn create_offset_map<S: PageSize, A: FrameAllocator<Size4KiB>, M: Map
     frame_allocator: &mut A,
 ) -> Result<(), MapToError<S>> {
     for (i, frame) in range.enumerate() {
+        // We don't set `PageTableFlags::GLOBAL` in `parent_table_flags` because Intel and AMD CPUs
+        // behave differently (Intel ignores the `G` bit in parent page table entries, AMD ignores
+        // it in lower entries _except_ PML4 and PML5); the `G` bit has semantic meaning only in the
+        // lowest level of page tables.
         mapper
             .map_to_with_table_flags(
                 Page::<S>::from_start_address(offset + i * (S::SIZE as usize)).unwrap(),
                 frame,
                 flags,
-                PageTableFlags::PRESENT
-                    | PageTableFlags::GLOBAL
-                    | PageTableFlags::WRITABLE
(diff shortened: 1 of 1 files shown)


oak_restricted_kernel/src/mm/page_tables.rs
```diff
@@ -45,10 +45,7 @@ pub unsafe fn create_offset_map<S: PageSize, A: FrameAllocator<Size4KiB>, M: Map
                 Page::<S>::from_start_address(offset + i * (S::SIZE as usize)).unwrap(),
                 frame,
                 flags,
-                PageTableFlags::PRESENT
-                    | PageTableFlags::GLOBAL
-                    | PageTableFlags::WRITABLE
-                    | PageTableFlags::ENCRYPTED,
+                PageTableFlags::PRESENT | PageTableFlags::WRITABLE | PageTableFlags::ENCRYPTED,
```
tiziano88, line 48: If I were you I would add almost the entire commit message as a comment around here. Our future selves (or our successors) will thank us.

tiziano88 (approved):

oak_restricted_kernel/src/mm/page_tables.rs
andrisaar: Better yet, I found out what the difference is.

- AMD64 Architecture Programmer's Manual, Volume 2, Section 5.3: in PML4E (and PML5E), the `G` bit must be zero. `G` is ignored in PDPE and PDE (assuming 4K pages).
- Intel 64 and IA-32 Architectures Software Developer’s Manual, Volume 3A, Section 4.5: in `G` bit is ignored in all of PML5E, PML4E, PDPE and PDE (again, assuming 4K pages).

So there's a genuine CPU manufacturer difference here in how they handle page tables that caught me out.

I'll add a comment about this warning that you shouldn't set `GLOBAL` in the parent page table entries, as at best it'll get ignored and at worst it will cause failures.

conradgrobler (approved):

Issue #3385, opened 2022-10-25
Use protobuf as IDL for Oak


tiziano88:
For trusted / untrusted communication, as well as for Wasm invocations.

We currently have a flatbuffer-based implementation, but we didn't get much out of it in terms of zero-copy, and using flatbuffers is quite inconvenient in Rust, compared to protobuf.

This will also make it simpler to interop with other systems within Google, which speak protobuf natively.

cc @sidtelang

tiziano88:
@sidtelang are you happy with this plan then? In particular, will you be able to switch to standard protobuf serialisation for your binaries, perhaps by leveraging nanopb?

PR #3386, opened 2022-10-25
Implement IDL based on protobuf

tiziano88:
Currently as a stand-alone crate, but in the future it may replace (and be renamed into) `oak_idl`.

Currently not used anywhere, just in its own tests.

The structure of the code generator is identical to the existing one from
https://github.com/project-oak/oak/blob/e75740a22c5c3db3d25031cd8b660b50aa60eda2/oak_idl_gen_services/src/lib.rs , which I copy-pasted and modified in order to be driven from prost instead of flatc.

The easiest way of reviewing this PR is probably to look at the generated file, which is also checked in for convenience and to catch regression or changes in the generator.

Ref #3385

### DIFF
--- a/oak_protobuf_idl_build/src/lib.rs
+++ b/oak_protobuf_idl_build/src/lib.rs
@@ -64,17 +64,15 @@ impl prost_build::ServiceGenerator for ServiceGenerator {
 /// Generate the Rust objects from the input [`Service`] instance, corresponding to a `service`
 /// entry.
 fn generate_service(service: &Service) -> anyhow::Result<String> {
+    let server_name = server_name(service);
+    let service_name = service_name(service);
     let mut lines = Vec::new();
     lines.extend(vec![
-        format!("pub struct {}<S> {{", server_name(service)),
+        format!("pub struct {server_name}<S> {{"),
         format!("    service: S"),
         format!("}}"),
         format!(""),
-        format!(
-            "impl <S: {}> ::oak_idl::Handler for {}<S> {{",
-            service_name(service),
-            server_name(service)
-        ),
+        format!("impl <S: {service_name}> ::oak_idl::Handler for {server_name}<S> {{"),
         format!("    fn invoke(&mut self, request: ::oak_idl::Request) -> Result<::prost::alloc::vec::Vec<u8>, ::oak_idl::Status> {{"),
         format!("        match request.method_id {{"),
     ]);
@@ -96,12 +94,12 @@ fn generate_service(service: &Service) -> anyhow::Result<String> {
         format!("    }}"),
         format!("}}"),
         format!(""),
-        format!("pub trait {}: Sized {{", service_name(service)),
+        format!("pub trait {service_name}: Sized {{"),
     ]);
     lines.extend(service.methods.iter().flat_map(generate_service_method));
     lines.extend(vec![
-        format!("    fn serve(self) -> {}<Self> {{", server_name(service)),
-        format!("        {} {{ service : self }}", server_name(service)),
+        format!("    fn serve(self) -> {server_name}<Self> {{"),
+        format!("        {server_name} {{ service : self }}"),
         format!("    }}"),
         format!("}}"),
         format!(""),
@@ -120,11 +118,11 @@ fn generate_service_client(service: &Service, asynchronous: bool) -> anyhow::Res
     };
     let mut lines = Vec::new();
     lines.extend(vec![
-        format!("pub struct {}<T: {}> {{", client_name, handler_trait),
+        format!("pub struct {client_name}<T: {handler_trait}> {{",),
         format!("    handler: T"),
         format!("}}"),
         format!(""),
-        format!("impl <T: {}>{}<T> {{", handler_trait, client_name),
+        format!("impl <T: {handler_trait}>{client_name}<T> {{"),
         format!("    pub fn new(handler: T) -> Self {{"),
         format!("        Self {{"),
         format!("            handler"),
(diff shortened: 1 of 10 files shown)


tiziano88:
cc @sidtelang

tiziano88:
Note for reviewers: you can focus the review on the generated file (https://github.com/project-oak/oak/pull/3386/files#diff-1af7722b767f7e900aab32051ffc5b2546bb111944c53bb95bebd810014575a4), and suggest changes there directly if you have any, then I can figure out how to modify the generator to apply those changes to the generated code. Other than that, the generation logic is basically identical to the existing one based on flatbuffers (https://github.com/project-oak/oak/blob/e75740a22c5c3db3d25031cd8b660b50aa60eda2/oak_idl_gen_services/src/lib.rs), and it even uses the same underlying types for the transport, namely:

https://github.com/project-oak/oak/blob/e75740a22c5c3db3d25031cd8b660b50aa60eda2/oak_idl/src/lib.rs#L164-L166

tiziano88:
My next step will then be to use this to complete the invocation-based Wasm ABI, so that we can express the interface as protobuf and generate client and server structs automatically from it. (cc @mariaschett ).

Then, we should discuss whether we also want to use this for trusted / untrusted communication, though that will require also updating the untrusted launcher in google3 at the same time as the trusted binary is checked in to google3 (cc @ipetr0v ).

oak_protobuf_idl_tests/oak.protobuf_idl.tests.rs.txt
```diff
@@ -0,0 +1,116 @@
...
+
+pub trait TestService: Sized {
+    fn lookup_data(&mut self, request: &LookupDataRequest) -> Result<LookupDataResponse, ::oak_idl::Status>;
+    fn log(&mut self, request: &LogRequest) -> Result<LogResponse, ::oak_idl::Status>;
+    fn serve(self) -> TestServiceServer<Self> {
+        TestServiceServer { service : self }
+    }
+}
+
+pub struct TestServiceClient<T: ::oak_idl::Handler> {
```
conradgrobler, line 54: Pre-existing, so should probably be addressed outside this PR, but "Handler" seems like a slightly weird name for the client side, as it is not really handling the work. I would have expected something more like Transport (or Channel).

conradgrobler (approved):
Nice, thanks

oak_protobuf_idl_build/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+    let server_name = server_name(service);
+    let service_name = service_name(service);
+    let mut lines = Vec::new();
+    lines.extend(vec![
+        format!("pub struct {server_name}<S> {{"),
+        format!("    service: S"),
+        format!("}}"),
+        format!(""),
+        format!("impl <S: {service_name}> ::oak_idl::Handler for {server_name}<S> {{"),
+        format!("    fn invoke(&mut self, request: ::oak_idl::Request) -> Result<::prost::alloc::vec::Vec<u8>, ::oak_idl::Status> {{"),
```
rbehjati, line 76: If there is a breaking change in `oak_idl` it will be detected only in tests, right? I wonder if there is a way to capture that those changes at compile time.

oak_protobuf_idl_build/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+fn generate_service_method(method: &Method) -> Vec<String> {
+    let method_name = method_name(method);
+    let request_type = request_type(method);
+    let response_type = response_type(method);
+    vec![format!("    fn {method_name}(&mut self, request: &{request_type}) -> Result<{response_type}, ::oak_idl::Status>;")]
+}
+
+/// Returns the value of the `method_id` commend on the method.
```
rbehjati, line 195:
```suggestion
/// Returns the value of the `method_id` comment on the method.
```

oak_protobuf_idl_tests/test_schema.proto
```diff
@@ -0,0 +1,38 @@
...
+  bytes value = 1;
+}
+
+message LogRequest {
+  string entry = 1;
+}
+message LogResponse {}
+
+service TestService {
+  // method_id: 15
```
rbehjati, line 34: I noticed that the generator uses this information in generating code. This is not required by the gRPC spec though. What happens if the comments are not present?

rbehjati (approved):
I don't have much context. Sorry if some of my comments are irrelevant.

oak_protobuf_idl_build/src/lib.rs
tiziano88: Yes it will only be in tests, these are just strings from the generator point of view; this is similar to macro expansion in Rust, they are only type checked after expansion.

oak_protobuf_idl_tests/oak.protobuf_idl.tests.rs.txt
tiziano88: Yeah I agree, I will revisit.

oak_protobuf_idl_tests/test_schema.proto
tiziano88: Correct, we use this instead of the method name to simplify things. Currently the generator emits an error if there is not exactly one such comment per rpc method, so hopefully users will figure out what they need to do to fix it (but I'm sure the UX can be improved).

tiziano88 (commented):
Thanks both

oak_protobuf_idl_build/src/lib.rs
tiziano88: Done.

PR #3390, opened 2022-10-26
Fix passing GITHUB_TOKEN

rbehjati:
Follow up fix for #3387.


tiziano88 (approved):

rbehjati:
The provenance workflow is still failing, but it seems to be relevant to a change in TUF. See with https://github.com/slsa-framework/slsa-github-generator/issues/1163.

PR #3396, opened 2022-10-27
Stop assuming identity mapping for GHCB

conradgrobler:
So far we have assumed an identity mapping between virtual and physical addresses of the GHCB. That works for the "hello world" test kernel, but will not work for the restricted kernel which is relocated to high memory.


andrisaar (approved):
Thanks for adding the note to other crates as well.

rbehjati (approved):
LGTM.

PR #3401, opened 2022-10-27
Serialize requests and responses via IDL

tiziano88:
This allows serializing the combination of method id and request body into an IDL message (so meta), which is then sent over the underlying transport, which now only needs to know about raw bytes and nothing about the serialization mechanism.

As a response, also the errors from the service are serialized back via the IDL to the client, so that we can have something much more similar to how RPC / IPC mechanisms usually work.

This adds an additional layer of encoding / decoding, but since it's the same as the messages in the inner layer, it seems acceptable after all.

Eventually the oak_protobuf_idl will replace the oak_idl crate, so the duplication will go away.

Ref #3385


oak_protobuf_idl/proto/messages.proto
```diff
@@ -0,0 +1,45 @@
...
+  bytes body = 2;
+}
+
+// A message representing an error status code with associated message.
+//
+// Similar to https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto.
+message Status {
+  // The status code, which should be an enum value of
+  // https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto.
+  int32 code = 1;
```
rbehjati, line 33: Is it not possible to import and use the enum as the type of this field?

oak_protobuf_idl/proto/messages.proto
```diff
@@ -0,0 +1,45 @@
...
+
+  // A developer-facing error message, which should be in English.
```
rbehjati, line 35: Why does it have to be in English? What happens if it is not?

conradgrobler (approved):

oak_protobuf_idl/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+/// A message-oriented transport that allows performing invocations.
+pub trait Transport {
+    /// Type representing any transport-specific errors. By default, the transport is assumed to be
+    /// infallible.
+    type Error = !;
+    fn invoke(&mut self, request_bytes: &[u8]) -> Result<Vec<u8>, Self::Error>;
+}
+
+#[async_trait::async_trait]
+pub trait AsyncTransport {
```
rbehjati, line 40: Will you add a 1-line doc comment here too?

rbehjati (approved):
Nice.

oak_protobuf_idl/proto/messages.proto
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak.protobuf_idl;
+
+// A message representing a request over a transport.
+message Request {
+  uint32 method_id = 1;
```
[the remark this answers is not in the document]
tiziano88: Done, also added more comments below.

oak_protobuf_idl/proto/messages.proto
tiziano88: I copied this comment from the gRPC Status. I guess it's there to ensure that developers don't expect this to be a localized message that is displayed to users. Anyways it's a *should*, so it's not a strict requirement.

oak_protobuf_idl/src/lib.rs
tiziano88: Sure, added on the methods too.

oak_protobuf_idl/proto/messages.proto
tiziano88: It seems to cause more problems than it solves, because of transitive imports etc. Perhaps that's also why gRPC does it like this, plus in principle the enum can evolve separately from this message, though I think that's unlikely to be useful in practice. Anyways, let's try it out as it is, but we can move things around later.

tiziano88 (commented):
Thanks both

oak_protobuf_idl/proto/messages.proto
rbehjati: I see. Thanks for clarifying :)

PR #3402, opened 2022-10-28
Initialize the GHCB in the restricted kernel

conradgrobler:
This PR only adds the code to initialise the GHCB in the Restricted Kernel and share it with the hypervisor. The code for actually using the GHCB will be added in a follow-up PR.

### DIFF
--- a/oak_restricted_kernel/src/ghcb.rs
+++ b/oak_restricted_kernel/src/ghcb.rs
@@ -1,0 +1,184 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use crate::mm::{ENCRYPTED_BIT_POSITION, KERNEL_OFFSET};
+use sev_guest::{
+    ghcb::{Ghcb, GhcbProtocol},
+    msr::{
+        change_snp_page_state, register_ghcb_location, PageAssignment, RegisterGhcbGpaRequest,
+        SnpPageStateChangeRequest,
+    },
+    Translator,
+};
+use x86_64::{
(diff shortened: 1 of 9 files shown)


oak_restricted_kernel/src/ghcb.rs
```diff
@@ -0,0 +1,187 @@
...
+    };
+    share_ghcb_with_hypervisor(ghcb, snp_enabled, translate);
+    ghcb.reset();
+
+    GhcbProtocol::new(ghcb, translate)
+}
+
+/// Marks the page containing the GHCB data structure to be shared with the hypervisor.
+fn share_ghcb_with_hypervisor<VP: Translator>(ghcb: &Ghcb, snp_enabled: bool, translate: VP) {
+    let ghcb_address = VirtAddr::new(ghcb as *const Ghcb as usize as u64);
```
andrisaar, line 76: Use `VirtAddr::from_ptr()`, that saves two casts here.

oak_restricted_kernel/src/ghcb.rs
```diff
@@ -0,0 +1,187 @@
...
+    // Mark the page as not encrypted in the page tables.
+    set_encrypted_bit_for_page(&ghcb_address, false);
+}
+
+/// Sets the encrypted bit for the page that contains the address.
+///
+/// Since this is called during early boot, before the frame allocator or page mapper is initialised
+/// we have to manually traverse the existing page tables. We use only 2MiB huge pages during early
+/// boot.
+fn set_encrypted_bit_for_page(address: &VirtAddr, encrypted: bool) {
```
andrisaar, line 98: Ah, good old manual page walk :)

I think you can use `EncryptedPageTable` from `mm/encrypted_mapper.rs` here. It has support for changing the encrypted bit of a page already.

oak_restricted_kernel/src/mm/mod.rs
```diff
@@ -38,6 +38,13 @@ mod encrypted_mapper;
 pub mod frame_allocator;
 pub mod page_tables;
 
+/// The start of kernel memory.
+pub const KERNEL_OFFSET: u64 = 0xFFFF_FFFF_8000_0000;
```
andrisaar, line 42: Move the `DIRECT_MAPPING_OFFSET` here as well so that the constants would be together.

oak_restricted_kernel/src/ghcb.rs
conradgrobler: Done

conradgrobler (commented):
Thanks

conradgrobler:
I agree. I will do some cleanup shortly.

PR #3406, opened 2022-10-28
Use the GHCB for drivers in the Restricted Kernel

conradgrobler:
We can now use the GHCB for serial port logging and the simple IO device to support running under SEV-ES and SEV-SNP.

The virtio drivers are not yet compatible as they require additional work to support the MMIO GHCB protocol. That will be added in future.

### DIFF
--- a/oak_restricted_kernel/src/ghcb.rs
+++ b/oak_restricted_kernel/src/ghcb.rs
@@ -70,7 +70,7 @@ lazy_static! {
 /// Shares the page containing the GHCB with the hypervisor again.
 ///
 /// This should be called as soon as the kernel memory has been initialised, as that would have
-/// cause the page to be marked as encrypted.
+/// caused the page to be marked as encrypted.
 pub fn reshare_ghcb<M: Mapper<Size2MiB>>(mapper: &mut M) {
     // Safety: we only use the reference to calculate the address and never dereference it.
     let ghcb_address =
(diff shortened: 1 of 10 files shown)


oak_restricted_kernel/src/ghcb.rs
```diff
@@ -53,8 +56,42 @@ static_assertions::assert_eq_size!(GhcbAlignmentWrapper, [u8; Size2MiB::SIZE as
...
+}
+
+/// Shares the page containing the GHCB with the hypervisor again.
+///
+/// This should be called as soon as the kernel memory has been initialised, as that would have
+/// caused the page to be marked as encrypted.
+pub fn reshare_ghcb<M: Mapper<Size2MiB>>(mapper: &mut M) {
+    // Safety: we only use the reference to calculate the address and never dereference it.
+    let ghcb_address =
+        unsafe { VirtAddr::new(&GHCB_WRAPPER as *const GhcbAlignmentWrapper as usize as u64) };
```
andrisaar, line 78: Use `VirtAddr::from_ptr` here.

andrisaar (approved):

oak_restricted_kernel/src/ghcb.rs
conradgrobler: Done. Missed that one.

PR #3408, opened 2022-10-28
Split echo binary

ipetr0v:
This PR splits Oak Echo binary into `oak_echo_bin` and `oak_echo_raw_bin` in order to remove feature gating.


testing/oak_echo_raw_bin/README.md
```diff
@@ -0,0 +1,4 @@
+# Echo Baremetal
```
tiziano88, line 1: What is the difference between the raw and non-raw binaries? Also we should not use baremetal, freestanding is the correct term.

What is the binary expected to do?

Also, perhaps the other binary should be renamed to clarify that it actually uses the IDL, while this one doesn't (is that correct)?

testing/oak_echo_raw_bin/build.rs
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+fn main() {
+    println!("cargo:rerun-if-changed=layout.ld");
+    println!("cargo:rustc-link-arg=--script=layout.ld");
```
tiziano88, line 19: @andrisaar @conradgrobler @ipetr0v for the time being, can we at least move this linker script to the root of the workspace, and use it from all the crates, via `env!("WORKSPACE_ROOT")`? We have too many copies already, it's getting out of hand.

testing/oak_echo_raw_bin/src/asm/boot.s
```diff
@@ -0,0 +1,62 @@
+/*
```
tiziano88, line 1: @andrisaar could this be encapsulated in an existing crate or a new one that will be shared across binaries?

tiziano88 (approved):

testing/oak_echo_raw_bin/build.rs
ipetr0v: This should be a separate PR, but I agree, we need to try creating a common crate that will be shared across different TEE binaries.

testing/oak_echo_raw_bin/README.md
ipetr0v: Submitted a PR, added comments in: https://github.com/project-oak/oak/pull/3410

PR #3410, opened 2022-10-28
Update Echo Readme

ipetr0v:


testing/oak_echo_bin/README.md
```diff
@@ -1,4 +1,7 @@
-# Echo Baremetal
+# Oak Echo
 
 Echo baremetal binary that can be run under
```
tiziano88, line 3: s/baremetal/freestanding/ ?

testing/oak_echo_raw_bin/README.md
```diff
@@ -1,4 +1,8 @@
-# Echo Baremetal
+# Oak Echo Raw
 
 Echo baremetal binary that can be run under
 [crosvm](https://chromium.googlesource.com/chromiumos/platform/crosvm/).
+
+This binary doesn't use IDL to communicate between the Untrusted Launcher and
+the guest VM, but instead uses a raw SimpleIO channel. The binary is used
+primarily for Vanadium tests.
```
tiziano88, line 8: But what does it actually do? I would expect something like: it receives a message as bytes, and sends back the same bytes, without interpreting them.

testing/oak_echo_bin/README.md
```diff
@@ -1,4 +1,7 @@
...
 [crosvm](https://chromium.googlesource.com/chromiumos/platform/crosvm/).
+
+This binary uses IDL to communicate between the Untrusted Launcher and the guest
+VM.
```
tiziano88, line 7: Could you add a link to the actual IDL definition for this binary?

tiziano88 (approved):

testing/oak_echo_bin/README.md
ipetr0v: Done

testing/oak_echo_bin/README.md
ipetr0v: Done

testing/oak_echo_raw_bin/README.md
ipetr0v: Done

testing/oak_echo_bin/README.md
```diff
@@ -1,4 +1,10 @@
...
 
-Echo baremetal binary that can be run under
+Echo freestanding binary that can be run under
 [crosvm](https://chromium.googlesource.com/chromiumos/platform/crosvm/).
+
+Echo binary receives a message as bytes, and sends back the same bytes, without interpreting them.
+
+This binary uses IDL to communicate between the Untrusted Launcher and the guest
+VM.
+The IDL is defined here: [`schema.fbs`](testing/oak_echo_runtime/schema.fbs)
```
tiziano88, line 10: IDL is a language: Interface Definition Language.

What is linked is the interface of the service, not the language itself.

testing/oak_echo_bin/README.md
```diff
@@ -1,4 +1,10 @@
-# Echo Baremetal
+# Oak Echo
 
-Echo baremetal binary that can be run under
+Echo freestanding binary that can be run under
 [crosvm](https://chromium.googlesource.com/chromiumos/platform/crosvm/).
+
+Echo binary receives a message as bytes, and sends back the same bytes, without interpreting them.
```
tiziano88, line 6: This is true for the raw one, right? This binary is interpreting the bytes via the IDL, which is the difference between them. Or did I misunderstand?

testing/oak_echo_bin/README.md
ipetr0v: Renamed to interface

testing/oak_echo_bin/README.md
ipetr0v: Yes, `oak_echo_bin` uses IDL but still receives bytes and sends them back.
Updated the comment.

PR #3413, opened 2022-10-31
Clean up GHCB sharing code

conradgrobler:
The GHCB sharing code duplicated code around address translation and page table management that was already written elsewhere in the restricted kernel. This PR removes the duplication and simplifies the implementation.


andrisaar (approved):
It'll be interesting to see if this changes the observed behaviour.

PR #3420, opened 2022-11-01
Add support for CPUID to the GHCB protocol

conradgrobler:
The GHCB protocol provides a richer and more efficient mechanism for calling CPUID under AMD SEV-ES than the MSR protocol, including support for sub-leafs.

With AMD SEV-SNP the CPUID page can be used to get the information, but with SEV-ES there is no CPUID page.

### DIFF
--- a/experimental/sev_guest/src/cpuid.rs
+++ b/experimental/sev_guest/src/cpuid.rs
@@ -30,15 +30,38 @@ pub const CPUID_PAGE_SIZE: usize = 4096;
 #[repr(C)]
 #[derive(Debug, FromBytes)]
 pub struct CpuidFunction {
-    /// The input value of the EAX register when CPUID was called. This represents the CPUID leaf.
-    pub eax_in: u32,
-    /// The input value of the ECX register when CPUID was called. This represents the CPUID
-    /// sub-leaf.
-    pub ecx_in: u32,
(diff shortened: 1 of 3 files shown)


experimental/sev_guest/src/cpuid.rs
```diff
@@ -30,15 +30,38 @@ pub const CPUID_PAGE_SIZE: usize = 4096;
...
+#[repr(C)]
+#[derive(Debug, FromBytes)]
+pub struct CpuidInput {
+    /// The input value of the EAX register, which represents the CPUID leaf.
+    pub eax: u32,
+    /// The input value of the ECX register, which represents the CPUID sub-leaf.
+    pub ecx: u32,
+    /// The input value of the XCR0 extended control register.
+    ///
+    /// Only required when a request for CPUID 0000_000D is made. Must be zero otherwise.
```
tiziano88, line 52:
```suggestion
    /// Only required when a request for CPUID 0x0000_000D is made. Must be zero otherwise.
```

tiziano88 (approved):

experimental/sev_guest/src/cpuid.rs
conradgrobler: Done, thanks.

PR #3422, opened 2022-11-01
Advance RIP in CPUID VC handler

conradgrobler:
I forgot to advance the instruction pointer past the CPUID instruction in the #VC exception handler, which meant that it kept on executing it after returning from the handler. This fixes the problem.

### DIFF
--- a/oak_restricted_kernel/src/interrupts.rs
+++ b/oak_restricted_kernel/src/interrupts.rs
@@ -144,7 +144,7 @@ mutable_interrupt_handler_with_error_code!(
                 let leaf = stack_frame.rax as u32;
                 get_cpuid_for_vc_exception(leaf, stack_frame).expect("Error reading CPUID");
                 // CPUID instruction is 2 bytes long, so we advance the instruction pointer by 2.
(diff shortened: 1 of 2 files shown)


oak_restricted_kernel/src/interrupts.rs
```diff
@@ -143,6 +143,8 @@ mutable_interrupt_handler_with_error_code!(
                 }
                 let leaf = stack_frame.rax as u32;
                 get_cpuid_for_vc_exception(leaf, stack_frame).expect("Error reading CPUID");
+                // CPUID instruction is 2 bytes long, so we advance the instruction pointer by 2.
+                stack_frame.rip = VirtAddr::new(stack_frame.rip.as_u64() + 2);
```
andrisaar, line 147: I think you can just do `stack_frame.rip += 2;` here.

andrisaar (approved):

oak_restricted_kernel/src/interrupts.rs
conradgrobler: I tried that originally but the compiler complained. Turns out it was just because it did not know what type "2" represented, so using `2u64` does work.

PR #3425, opened 2022-11-02
SEV-SNP: `pvalidate` all physical memory in stage0

andrisaar:
We need to run `pvalidate` on all physical memory before we use it; it's usually done in the bootloader, hence we put it in stage0.

For now, we ignore any errors that may occur. For some pages, we expect the validation status change to fail (everything populated by the hypervisor); in the future, we should double-check that the errors happen only where we expect them to, and crash if they occur anywhere else.

### DIFF
--- a/stage0/src/sev.rs
+++ b/stage0/src/sev.rs
@@ -1,0 +1,106 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


stage0/src/sev.rs
```diff
@@ -0,0 +1,106 @@
...
+    structures::paging::{PageSize, PageTable, PageTableFlags, PhysFrame, Size2MiB, Size4KiB},
+    PhysAddr,
+};
+
+/// Calls `PVALIDATE` on all memory ranges specified in the E820 table with type `RAM`.
+pub fn validate_memory(zero_page: &BootParams, encrypted: u64) {
+    let mut page_table = PageTable::new();
+
+    // Find a location for our (temporary) page table. The initial page tables map [0..2MiB), so it
+    // should be safe to put our page table to [2..4MiB).
```
conradgrobler, line 32: nit: I found this comment a bit confusing on the first pass through, as the page table itself is not located in this range - it just map values in this range to physical addresses.

conradgrobler (approved):
Nice.

stage0/src/sev.rs
andrisaar: Done.

PR #3430, opened 2022-11-03
Reduce the size of stage0 binary

andrisaar:
There are three major things in here:
  * Enabling `opt-level = "z"` reduces the code by ~50K
  * enabling LTO (which prunes a lot of dead code) reduces code size ~10x (!!!) -- from 70K to 7K.
 
This means we can fit the debug build into 256K instead of 512K, and in theory the release build would fit into 48K (!!!).

There are other things we can do to trim the size down further; for example, we could move the stack to low memory and we have 12K worth of page table structures in the BIOS data area. We could construct them during runtime in low memory with far less code than that -- but, to be blunt, I'm amazed at how much just enabling LTO for release builds saved us.

My main goal was to shave it down enough so that the release build could fit into 128K, and that goal is met, so I've left the rest of the yak unshaven.

### DIFF
--- a/stage0/build.rs
+++ b/stage0/build.rs
@@ -14,7 +14,15 @@
 // limitations under the License.
 //
 
+use std::env;
+
 fn main() {
     println!("cargo:rerun-if-changed=layout.ld");
     println!("cargo:rustc-link-arg=--script=layout.ld");
+
+    if env::var("PROFILE").unwrap() == "release" {
+        println!("cargo:rustc-link-arg=--defsym=BIOS_SIZE=64K");
+    } else {
+        println!("cargo:rustc-link-arg=--defsym=BIOS_SIZE=256K");
+    }
 }
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -15,7 +15,7 @@
 */
 
 HIDDEN(TOP = 4096M);
-HIDDEN(BIOS_SIZE = 512K);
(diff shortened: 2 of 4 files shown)


stage0/build.rs
```diff
@@ -14,7 +14,15 @@
 // limitations under the License.
 //
 
+use std::env;
+
 fn main() {
     println!("cargo:rerun-if-changed=layout.ld");
     println!("cargo:rustc-link-arg=--script=layout.ld");
+
+    if env::var("PROFILE").unwrap() == "release" {
```
tiziano88, line 23: could you use `env!("PROFILE")` here?

stage0/layout.ld
```diff
@@ -15,7 +15,6 @@
 */
 
 HIDDEN(TOP = 4096M);
-HIDDEN(BIOS_SIZE = 512K);
```
tiziano88, line 18: Is it worth adding a comment saying that BIOS_SIZE needs to be provided externally?

tiziano88 (commented):
Wow, 48K, impressive!

conradgrobler (approved):
Nice.

stage0/build.rs
andrisaar: ```error: environment variable `PROFILE` not defined```

stage0/layout.ld
andrisaar: Done.

PR #3431, opened 2022-11-03
Consolidate bootstrap assembly and linker scripts to one location

andrisaar:
Both the bootstrap assembly code and the linker script are identical for all of our binaries, so having multiple copies of them is a pointless maintenance nightmare.

Therefore, let's consolidate all of the code into one location under the kernel directory.


oak_functions_freestanding_bin/build.rs
```diff
@@ -15,7 +15,6 @@
 //
 
 fn main() {
-    println!("cargo:rerun-if-changed=target.json");
-    println!("cargo:rerun-if-changed=layout.ld");
-    println!("cargo:rustc-link-arg=--script=layout.ld");
+    println!("cargo:rerun-if-changed=../oak_restricted_kernel/layout.ld");
+    println!("cargo:rustc-link-arg=--script=../oak_restricted_kernel/layout.ld");
```
tiziano88, line 19: Could these be added to the kernel crate directly, and use something like the following to locate those files? Or even better, can it just be part of the build script of that crate, so that other crates don't even have to have build scripts themselves? Or does rustc-link-arg only apply to the crate that declares it?

https://github.com/project-oak/oak/blob/8783bce33212f755e70b39fc1513cb79b68c4377/oak_protobuf_idl_tests/build.rs#L19-L22

tiziano88 (approved):

oak_restricted_kernel/layout.ld
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
 */
 
-ENTRY(_start)
+ENTRY(_oak_start)
```
ipetr0v, line 17: Why was in changed form the standard `start` to `oak_start`?

ipetr0v (approved):

oak_restricted_kernel/layout.ld
andrisaar: `_start` is also used by libc, and tests are linked against libc. Thus, for test builds, these two functions would clash.

I guess another option would be to wrap the `global_asm` stanza in a `#[cfg(not(test))]` to ensure this file wouldn't be compiled in at all; but I think this is slightly cleaner.

oak_functions_freestanding_bin/build.rs
andrisaar: Moving the `build.rs` under the kernel doesn't work.

However, using `WORKSPACE_ROOT` works. I was rather surprised at that, as the freestanding binaries are not part of our global workspace. I guess that environment variable comes from elsewhere.

PR #3432, opened 2022-11-04
Update wasmi

### LINKED ISSUE #3377: cargo deny reports deprecated crate: `parity-wasm`

### LINKED ISSUE #3449: Update and evaluate `wasmi`
With the wasmi update we should revisit whether our original assumptions and reasons for choosing wasmi still hold.

Our reasons for wasmi:
- no JIT compilation, hence requires no executable memory
- we do not need wasi
- we need `no_std`

mariaschett:
Fixes #3449 . Fixes #3377 .

### DIFF
--- a/oak_functions/wasm/src/lib.rs
+++ b/oak_functions/wasm/src/lib.rs
@@ -142,12 +142,6 @@ where
 
         let mut linker: wasmi::Linker<UserState<L>> = wasmi::Linker::new(&module.engine());
 
-        // Although it would be nice to define the linker in WasmHandler to reuse it for
-        // different WasmStates, we can't because the externals Func and Memory defined in the
-        // linker depend on store, which needs to be new for every WasmState. In contrast,
-        // in wasmtime `func_wrap` does not depend on store (https://docs.rs/wasmtime/latest/wasmtime/struct.Linker.html#method.func_wrap).
-        // TODO(mschett): Update, we can use func_wrap.
-
         // Add memory to linker.
         // TODO(#3783): Find a sensible value for initial pages.
         let initial_pages = 16384; // corresponds to 1 GiB
@@ -155,6 +149,11 @@ where
             MemoryType::new(initial_pages, None).expect("failed to define Wasm memory type");
         let memory =
             wasmi::Memory::new(&mut store, memory_type).expect("failed to initialize Wasm memory");
+
+        // TODO(#3784): Use func_wrap. Maybe we can define the linker in WasmHandler to reuse it for
+        // different WasmStates, but so far we can't because the externals Func and Memory
+        // defined in the linker depend on store, which needs to be new for every WasmState.
+        // Related, wasmtime `func_wrap` does not depend on store (https://docs.rs/wasmtime/latest/wasmtime/struct.Linker.html#method.func_wrap).
 
         linker
             .define(HOST, MEMORY_NAME, wasmi::Extern::Memory(memory))
@@ -240,6 +239,7 @@ where
             .get_memory(&store, MEMORY_NAME)
             .ok_or(anyhow::anyhow!("couldn't find Wasm `memory` export"))?;
 
+        // TODO(#3785): Add logging and list exports.
         let wasm_state = Self { instance, store };
 
         Ok(wasm_state)
@@ -280,15 +280,16 @@ pub fn read_buffer(
 ) -> Result<Vec<u8>, OakStatus> {
     let mut target = alloc::vec![0; buf_len as usize];
 
-    let buf_ptr = usize::try_from(buf_ptr)
+    let buf_ptr = buf_ptr
+        .try_into()
         .expect("failed to convert AbiPointer to usize as required by wasmi API");
     memory.read(&ctx, buf_ptr, &mut target).map_err(|_err| {
-        // TODO(mschett): Add logging.
-        /*
-        self.logger.log_sensitive(
-            Level::Error,
-            &format!("Unable to read buffer from guest memory: {:?}", err),
-        ); */
+        // TODO(#3785): Add logging, which needs access to logger in the user state.
+        // We don't have access in ctx, so we either need to pass the logger as argument or
+        // think of a different refactoring.
+        // ctx.data().log_error(
+        //   &format!("Unable to read buffer from guest memory: {:?}", err),
+        // );
         OakStatus::ErrInvalidArgs
     })?;
     Ok(target)
@@ -304,13 +305,12 @@ pub fn write_u32(
     let value_bytes = &mut [0; 4];
     LittleEndian::write_u32(value_bytes, value);
     write_buffer(ctx, memory, value_bytes, address).map_err(|_err| {
-        // TODO(mschett): Fix logging
-        /*
-        self.logger.log_sensitive(
-            Level::Error,
-            &format!("Unable to write u32 value into guest memory: {:?}", err),
-        );
-        */
+        // TODO(#3785): Add logging, which needs access to logger in the user state.
+        // We don't have access in ctx, so we either need to pass the logger as argument or
+        // think of a different refactoring.
+        // ctx.data().log_error(
+        //    &format!("Unable to write u32 value into guest memory: {:?}", err),
+        // );
         OakStatus::ErrInvalidArgs
     })
 }
@@ -323,15 +323,16 @@ pub fn write_buffer(
     source: &[u8],
     dest: AbiPointer,
 ) -> Result<(), OakStatus> {
-    let dest = usize::try_from(dest)
+    let dest = dest
+        .try_into()
         .expect("failed to convert AbiPointer to usize as required by wasmi API");
     memory.write(ctx, dest, source).map_err(|_err| {
-        // TODO(mschett): Add logging.
-        /*
-        self.logger.log_sensitive(
-            Level::Error,
-            &format!("Unable to write buffer into guest memory: {:?}", err),
-        );  */
+        // TODO(#3785): Add logging, which needs access to logger in the user state.
+        // We don't have access in ctx, so we either need to pass the logger as argument or
+        // think of a different refactoring.
+        // ctx.data().log_error(
+        //    &format!("Unable to write buffer into guest memory: {:?}", err),
+        // );
         OakStatus::ErrInvalidArgs
     })
 }
@@ -350,9 +351,9 @@ pub fn call_alloc(
         .expect("`alloc` call failed");
     match address[0] {
         // Assumes 32-bit Wasm addresses.
-        wasmi::Value::I32(v) => {
-            u32::try_from(v).map_or_else(|_| Err(OakStatus::ErrInternal), |v| Ok(v))
-        }
+        wasmi::Value::I32(v) => v
+            .try_into()
+            .map_or_else(|_| Err(OakStatus::ErrInternal), |v| Ok(v)),
         _ => Err(OakStatus::ErrInternal),
     }
 }
--- a/oak_functions/wasm/src/lib.rs
+++ b/oak_functions/wasm/src/lib.rs
@@ -443,17 +443,17 @@ pub fn invoke_extension<L: OakLogger>(
 fn get_exported_memory<L: OakLogger>(
     caller: &mut wasmi::Caller<'_, UserState<L>>,
 ) -> Result<Memory, OakStatus> {
-    let ext = caller.get_export(MEMORY_NAME).ok_or({
+    let ext = caller.get_export(MEMORY_NAME).ok_or_else(|| {
         caller
             .data()
             .log_error(&format!("failed to get exported {}", MEMORY_NAME));
         OakStatus::ErrInternal
     })?;
 
-    ext.into_memory().ok_or({
+    ext.into_memory().ok_or_else(|| {
         caller
             .data()
-            .log_error(&format!("exported {} is not a memory", MEMORY_NAME));
+            .log_error(&format!("exported {} is not a memory: {}", MEMORY_NAME));
         OakStatus::ErrInternal
     })
 }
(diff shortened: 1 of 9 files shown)


mariaschett:
@tiziano88 This is quite a large PR with a couple of TODOs left, but I think it is mainly refactoring & adding logging now. 

If you have some time, could you maybe start looking at it, as I hope to merge as soon as possible.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+}
+
+/// Corresponds to the host ABI function [`read_request`](https://github.com/project-oak/oak/blob/main/docs/oak_functions_abi.md#read_request).
+pub fn read_request<L: OakLogger>(
+    caller: &mut wasmi::Caller<'_, UserState<L>>,
+    dest_ptr_ptr: AbiPointer,
+    dest_len_ptr: AbiPointer,
+) -> Result<(), OakStatus> {
+    // TODO(mschett): Fix unwraps.
+    let alloc = caller.get_export("alloc").unwrap().into_func().unwrap();
```
tiziano88, line 408: Use the const for `alloc`.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+pub fn read_buffer(
+    ctx: &mut impl AsContext,
+    memory: &mut wasmi::Memory,
+    buf_ptr: AbiPointer,
+    buf_len: AbiPointerOffset,
+) -> Result<Vec<u8>, OakStatus> {
+    let mut target = alloc::vec![0; buf_len as usize];
+
+    let buf_ptr = usize::try_from(buf_ptr)
+        .expect("failed to convert AbiPointer to usize as required by wasmi API");
```
tiziano88, line 308: I think `let buf_ptr: usize = buf_ptr.try_into()` or `let buf_ptr = buf_ptr.try_into::<usize>()` would be more idiomatic.

oak_functions/wasm/src/tests.rs
```diff
@@ -214,9 +386,23 @@ fn create_test_wasm_state() -> WasmState<TestingLogger> {
...
+
+// Read the u32 value at the `address` from the Wasm memory.
+// Only needed in tests.
+fn read_u32(
+    ctx: &mut impl wasmi::AsContext,
+    memory: &mut wasmi::Memory,
+    address: AbiPointer,
+) -> Result<u32, OakStatus> {
+    let address = read_buffer(ctx, memory, address, 4).map_err(|_err| OakStatus::ErrInvalidArgs)?;
+    let address = LittleEndian::read_u32(&address);
```
tiziano88, line 406: Does https://doc.rust-lang.org/std/primitive.i32.html#method.from_be_bytes not work?

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
-            Some(extension) => Ok(extension),
-            None => {
-                self.log_error(&format!("cannot find extension with handle {:?}", handle));
-                Err(OakStatus::ErrInvalidHandle)
-            }
-        }?;
-
-        let response = extension.invoke(request)?;
-        self.alloc_and_write_buffer_to_wasm_memory(response, response_ptr_ptr, response_len_ptr)
+        // Check that the instance exports "alloc".
```
tiziano88, line 231: As a future todo, it may be nice to list the [exports](https://docs.rs/wasmi/0.28.0/wasmi/struct.Instance.html#method.exports) from the instance, to help debugging.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+        OakStatus::ErrInvalidArgs
+    })
+}
+
+// Calls given alloc Func with ctx and length as parameters.
+// alloc Func has to belong to given ctx.
+pub fn call_alloc(ctx: &mut impl AsContextMut, alloc: Func, len: i32) -> AbiPointer {
+    let inputs = &[wasmi::Value::I32(len)];
+    // TODO(mschett): Check whether putting default value 0 is a good idea.
+    let mut outputs = [wasmi::Value::I32(0); 1];
```
tiziano88, line 371:
```suggestion
    let mut outputs = [wasmi::Value::I32(0)];
```

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+    alloc
+        .call(ctx, inputs, &mut outputs)
+        .expect("`alloc` call failed");
+
+    let result_value = outputs;
+
+    match result_value[0] {
+        wasmi::Value::I32(v) => v as u32,
+        _ => panic!("invalid value type returned from `alloc`"),
```
tiziano88, line 380: We should definitely not panic here, since a malformed / malicious module may trigger this, which may be observable by the host. (this may well be a pre-existing issue, I haven't checked)

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+pub fn invoke_extension<L: OakLogger>(
+    caller: &mut wasmi::Caller<'_, UserState<L>>,
+    handle: i32,
+    request_ptr: AbiPointer,
+    request_len: AbiPointerOffset,
+    response_ptr_ptr: AbiPointer,
+    response_len_ptr: AbiPointer,
+) -> Result<(), OakStatus> {
+    // TODO(mschett): Fix unwraps.
+    let alloc = caller.get_export("alloc").unwrap().into_func().unwrap();
```
tiziano88, line 468: Use const for alloc.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+
+    let mut memory = caller
+        .get_export("memory")
```
tiziano88, line 471: Use const for memory.

oak_functions/wasm/src/lib.rs
```diff
@@ -476,59 +565,13 @@ where
...
-    let oak_status_from_result = result.map_or_else(|x: OakStatus| x, |()| OakStatus::Ok);
-    let wasmi_value = wasmi::RuntimeValue::I32(oak_status_from_result as i32);
-    Ok(Some(wasmi_value))
+/// `wasmi` specific result type `Result<i32, wasmi::Trap>`.
+// TODO(mschett): Changed result type from Option<i32> to i32. Check implications.
+fn from_oak_status_result(result: Result<(), OakStatus>) -> Result<i32, wasmi::core::Trap> {
+    let oak_status: OakStatus = match result {
+        Ok(()) => OakStatus::Ok,
+        Err(oak_status) => oak_status,
+    };
```
tiziano88, line 575: Could this be `let oak_status = result.err().unwrap_or(OakStatus::Ok)` or something like that?

tiziano88 (commented):
Thanks Maria, looking great so far! Let me know if you would like me or someone else to take a closer look at some point.

oak_functions/wasm/src/lib.rs
mariaschett: Done.

oak_functions/wasm/src/lib.rs
mariaschett: Done.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +36,470 @@ use oak_functions_abi::{
...
+        let memory_type =
+            MemoryType::new(initial_pages, None).expect("failed to define Wasm memory type");
+        let memory =
+            wasmi::Memory::new(&mut store, memory_type).expect("failed to initialize Wasm memory");
+
+        linker
+            .define(HOST, "memory", wasmi::Extern::Memory(memory))
+            .expect("failed to define Wasm memory in linker");
+
+        let read_request = wasmi::Func::wrap(
```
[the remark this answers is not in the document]
mariaschett: Added TODO(#3784).

oak_functions/wasm/src/lib.rs
mariaschett: Added TODO(#3785)

oak_functions/wasm/src/lib.rs
mariaschett: Done.

oak_functions/wasm/src/tests.rs
mariaschett: Does it make a difference?

mariaschett (commented):
Thank you! I added some TODOs based on your comments.

oak_functions/wasm/src/tests.rs
tiziano88: It avoids an extra external dependency for something that the std library already does. But I'm not sure if it's exactly the same functionality that you need here.

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +35,452 @@ use oak_functions_abi::{
...
 const ALLOC_FUNCTION_NAME: &str = "alloc";
+const MEMORY_NAME: &str = "memory";
 
-/// Wasm host function index numbers for `wasmi` to map import names with. This numbering is not
-/// exposed to the Wasm client. See <https://docs.rs/wasmi/0.6.2/wasmi/trait.Externals.html>
-const READ_REQUEST: usize = 0;
-const WRITE_RESPONSE: usize = 1;
-const INVOKE: usize = 4;
+// Needs to be consistent with definition of the Wasm import module in the Oak Functions ABI.
+const HOST: &str = "oak_functions";
```
tiziano88, line 45: Still not sure why this is called HOST instead of, say, MODULE_NAME? Just for my understanding (and maybe worth a comment).

oak_functions/wasm/src/lib.rs
```diff
@@ -36,388 +35,452 @@ use oak_functions_abi::{
...
+        response_ptr_ptr,
+        response_len_ptr,
+    )
+}
+
+// Helper function to get exported MEMORY from caller.
+fn get_exported_memory<L: OakLogger>(
+    caller: &mut wasmi::Caller<'_, UserState<L>>,
+) -> Result<Memory, OakStatus> {
+    let ext = caller.get_export(MEMORY_NAME).ok_or({
```
tiziano88, line 446: May be more idiomatic to use https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err for the logging, and then do the conversion only in ok_or, but up to you.

oak_functions/wasm/src/tests.rs
```diff
@@ -24,183 +28,351 @@ fn test_invoke_extension_with_invalid_handle() {
...
 }
 
 #[test]
 fn test_read_write_u32_in_wasm_memory() {
-    let wasm_state = create_test_wasm_state();
+    let wasm_state = &mut create_test_wasm_state();
+
+    let mut memory = wasm_state
+        .instance
+        .get_export(&wasm_state.store, "memory")
```
tiziano88, line 53: Use const for memory.

oak_functions/wasm/src/tests.rs
tiziano88: Ah sorry this is a test anyways, so not really relevant after all.

oak_launcher_utils/src/launcher/virtualized.rs
```diff
@@ -100,7 +100,8 @@ impl Instance {
         // Needed to expose advanced CPU features. Specifically RDRAND which is required for remote
         // attestation.
         cmd.args(["-cpu", "IvyBridge-IBRS,enforce"]);
-
+        // TODO(#3786): Make memory configurable.
+        cmd.args(["-m", "4G"]);
```
tiziano88, line 104: This seems a lot, since each test instance will have to allocate all this memory I guess? What was the previous (default) value? It would be nicer to be more conservative with the increase if possible.

tiziano88 (approved):
Thanks Maria

oak_functions/wasm/src/lib.rs
mariaschett: I followed the example in https://docs.rs/wasmi/latest/wasmi/ where they called it host. I think it is a matter of perspective, but I changed to OAK_FUNCTIONS now.

oak_functions/wasm/src/tests.rs
mariaschett: Done.

mariaschett (commented):
Thank you! Let me know, if 1GiB is too much to ask too, or if it is fine if I merge and come back to this after.

oak_functions/wasm/src/lib.rs
conradgrobler: +1. I *think* the current implementation will always log, even if there are no errors, since the value would always be calculated but only returned if there was an error. The other option is to use ok_or_else and do the logging in the closure. In that case it will only call the closure if there is an error.

conradgrobler (approved):
Nice, thanks.

oak_functions/wasm/src/lib.rs
tiziano88: Ah yes I didn't even realise that, but it's true; here the value will always be evaluated, which looks like it will always print an error, regardless of the actual value? But that should be easy to falsify by looking at logs.

PR #3435, opened 2022-11-05
Enable LTO on Oak Functions binary

tiziano88:
For some yet unknown reason, the binary only works on SNP with this option or, or with some other random changes to the code. This is not a perfect solution, and will likely break again in the future, but at least it unblocks syncing previous changes to google3.


andrisaar (approved):
I take it that the binary fails consistently without this stanza? That hopefully gives us something to debug...

tiziano88:
Yes, though it's unclear whether it's actually the size of the code that matters, or something else.

PR #3437, opened 2022-11-07
Make it possible to derive FromBytes

conradgrobler:
Some of the structs could be generated outside the VM, so we cannot assume that the memory where the structs are read from will always be valid (e.g. enum fields might not have valid values). Assuming valid memory without checking can lead to undefined behaviour.

This change changes the structs so that we can derive `zerocpoy::FromBytes`, which does not make assumptions about any of the fields. It also adds additional functions to try to convert the values to enums or bitflag structs to make it a bit more usable.

### DIFF
--- a/experimental/sev_guest/src/guest.rs
+++ b/experimental/sev_guest/src/guest.rs
@@ -17,7 +17,9 @@
 //! Utilities for creating and processing SNP Guest Request protocol messages.
 
 use bitflags::bitflags;
+use core::mem::size_of;
 use strum::FromRepr;
+use zerocopy::{AsBytes, FromBytes};
 
 /// The size of a guest message, including the header and maximum payload size.
 pub const GUEST_MESSAGE_SIZE: usize = 4096;
@@ -37,7 +39,7 @@ pub const CURRENT_ATTESTATION_VERSION: u8 = 2;
 ///
 /// See section 8.26 in <https://www.amd.com/system/files/TechDocs/56860.pdf>.
 #[repr(C, align(4096))]
-#[derive(Debug)]
+#[derive(Debug, AsBytes, FromBytes)]
 pub struct GuestMessage {
     /// The message header.
     pub header: GuestMessageHeader,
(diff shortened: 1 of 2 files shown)


experimental/sev_guest/src/guest.rs
```diff
@@ -47,11 +49,20 @@ pub struct GuestMessage {
...
 static_assertions::assert_eq_size!(GuestMessage, [u8; GUEST_MESSAGE_SIZE]);
 
+impl GuestMessage {
+    pub const fn new() -> Self {
+        GuestMessage {
+            header: GuestMessageHeader::new(),
+            payload: [0; 4000],
+        }
+    }
+}
```
tiziano88, line 59: Should this be Default (and automatically derived)? Or does that not work because it's not const?

experimental/sev_guest/src/guest.rs
```diff
@@ -62,13 +73,17 @@ pub struct GuestMessageHeader {
     /// Reserved. Must be zero.
     _reserved_0: u64,
     /// The algorithm used to encrypt the payload.
-    pub algorithm: AeadAlgorithm,
+    ///
+    /// Use `get_algorithm` to try to convert this to an `AeadAlgorithm` enum.
```
tiziano88, line 77:
```suggestion
    /// Use [`GuestMessageHeader::get_algorithm`] to try to convert this to an [`AeadAlgorithm`] enum.
```

so that if names change it will be detected.

(also below)

experimental/sev_guest/src/secrets.rs
```diff
@@ -63,4 +66,21 @@ pub struct SecretsPage {
...
 static_assertions::assert_eq_size!(SecretsPage, [u8; SECRETS_PAGE_SIZE]);
+
+/// Whether an initial migration image is enabled.
+#[derive(Debug, FromRepr)]
+#[repr(u32)]
+pub enum Imi {
+    /// The initial migration image is not enabled.
+    No = 0,
+    /// The initial migration image is enabled.
+    Yes = 1,
```
tiziano88, line 85: I would have called them Enabled and Disabled, but up to you.

tiziano88 (approved):

experimental/sev_guest/src/guest.rs
conradgrobler: Default::default is not const, and also Default is only implemented for fixed sized arrays up to 32 elements. Arrays larger than that is not supported when deriving Default.

experimental/sev_guest/src/guest.rs
conradgrobler: Done throughout.

experimental/sev_guest/src/secrets.rs
conradgrobler: Done.

PR #3443, opened 2022-11-08
Turn off LTO on `oak_functions_freestanding_bin`

andrisaar:
@tiziano88 added this in #3435 as it seemed to help with weird errors we were seeing while running under SEV-SNP; however, it looks like @conradgrobler fixed the real underlying problem in #3441.

Thus, let's disable LTO, again. This is not to say we shouldn't enable LTO or LTO is bad; in fact, we probably should enable it, and will likely do so in the future. For now, I want to reduce the number of baseline changes until we're certain we caught the bug.

### DIFF
--- a/oak_functions_freestanding_bin/Cargo.toml
+++ b/oak_functions_freestanding_bin/Cargo.toml
@@ -36,6 +36,3 @@ static_assertions = "*"
 name = "oak_functions_freestanding_bin"
 test = false
 bench = false
-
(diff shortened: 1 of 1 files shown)


conradgrobler (approved):

tiziano88 (approved):

tiziano88:
Sorry I haven't actually verified the binary hash yet, will do shortly.

Issue #3403, opened 2022-10-28
Implement a no_std-compatible equivalent to `OnceLock`


conradgrobler:
We would like to move away from lazy_static so the we can fully control the executoin sequence when the restricted kernel boots. `OnceCell` cannot be used for shared static references, as it does not implement `Sync`. We need something similar to `std::sync::OnceLock`, but implemented with spinning.

PR #3444, opened 2022-11-08
Implement a synchronized no_std OnceCell

conradgrobler:
This PR provides an implementation of a synchronised OnceCell. We want this so that we can move away from using `lazy_static!` and have better control over the initialisation of static variables throughout.

See #3403 for more information.

The `oak_functions_utils` crate also contains a spinlock implementation, but this OnceCell implementation uses the spinlock from the `spinning_top` crate. We rely on the `spinning_top` crate for some other dependencies (such as the linked-list allocator), and having two equivalent but different spinlock implementations is not needed. I will remove our spinlock implementation and rename the crate in a follow-up PR.

### DIFF
--- a/oak_functions_utils/src/sync.rs
+++ b/oak_functions_utils/src/sync.rs
@@ -15,10 +15,70 @@
 //
 
 use core::{
+    cell::UnsafeCell,
     hint::spin_loop,
+    mem::MaybeUninit,
     sync::atomic::{AtomicBool, Ordering},
 };
 use lock_api::{GuardSend, RawMutex};
+use spinning_top::RawSpinlock;
+
+/// A synchronised implementation of a cell that can be initialized only once.
+///
+/// This achieves the same goal as `std::sync::OnceLock`, but uses a spinlock to be compatible with
+/// `no_std`.
+pub struct OnceCell<T> {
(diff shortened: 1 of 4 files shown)


oak_functions_utils/src/sync.rs
```diff
@@ -15,10 +15,70 @@
...
+        // reference and assume it is initialized.
+        Some(unsafe { (*self.value.get()).assume_init_ref() })
+    }
+
+    pub fn set(&self, value: T) -> Result<(), T> {
+        let mut result = Ok(());
+        // Do an initial check to see whether the value has been initialized.
+        if !self.initialized.load(Ordering::Acquire) {
+            // Lock to make sure we have exclusive access.
+            self.lock.lock();
```
andrisaar, line 60: Would tehere be some kind of RAII way of doing this? https://docs.rs/spinning_top/latest/spinning_top/type.SpinlockGuard.html maybe?

conradgrobler: Yes, I tried to avoid having a Spinlock<()>, but it is probalby cleaner to have that this the explicit unlock.

andrisaar (approved):

PR #3446, opened 2022-11-08
Add a 4K page table for the lowest 2M of memory in stage0

andrisaar:
Thus far we've relied on three levels of page tables (PML4, PDPT, PD) leaving the leaf covering [0, 2MiB) in the PD to be a 2M hugepage.

However, we want to designate one 4K page in there as the GHCB for early logging under SEV-SNP, which requires us to change flags for just that 4K page, and not the whole 2M page.

Therefore, let's break the 2M page into 512 4K pages, and introduce a fourth-level page table (PT) for that memory area.

Instead of copy-pasting something in the linker script 512 times, I've written a small loop in assembly to set up the page table.

As we can't rely on hard-coded entries, setting the encrypted bit is both simpler and more complicated: I thought it'd be easiest just to blanket set the encrypted bit in all 512 entries of all four page tables, even if the entries are unused. As those entries don't have the `PRESENT` bit set, the CPU ignores those entries.

### DIFF
--- a/stage0/src/asm/boot.s
+++ b/stage0/src/asm/boot.s
@@ -46,6 +46,20 @@ _protected_mode_start:
     # Set up a basic stack, as we may get interrupts.
     mov $stack_start, %esp
 
+    # Set up the 4K page table for the lowest 2 MiB of memory. We set up an individual page table
+    # instead of using a 2MiB hugepage because we may need to change the encrypted bit and the
+    # RMP state on individual 4K pages.
+    mov $pt_addr, %ecx      # base for page table
+    xor %eax, %eax          # loop counter
+    1:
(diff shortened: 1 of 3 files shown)


stage0/src/asm/boot.s
```diff
@@ -46,6 +46,18 @@ _protected_mode_start:
     # Set up a basic stack, as we may get interrupts.
     mov $stack_start, %esp
 
+    # Set up the 4K page table for the lowest 2 MiB of memory.
```
tiziano88, line 49: I think your commit message would also be excellent context here as a comment (especially since you already wrote it), but up to you.

tiziano88 (approved):
cool

conradgrobler (approved):
Nice

PR #3447, opened 2022-11-09
Remove unused code and rename oak_utils

conradgrobler:
The `oak_utils` crate is only used for GRPC and Proto building, so I renamed it to `oak_grpc_utils`. There was also some additional code that is no longer used since the monolithic oak functions binary has been removed that I also deleted.


rbehjati (approved):
Nice. Thanks :)

Is the `oak_grpc_unary_attestation` server covered by some other server? More specifically I wonder why we need the client code, but not the server.

conradgrobler:
> Is the `oak_grpc_unary_attestation` server covered by some other server? More specifically I wonder why we need the client code, but not the server.

Thanks for the review.

The server code was used in the monolithic Oak Functions that included a gRPC server, and that version was deleted a while ago. The version of Oak Functions that runs on the restricted kernel does not use gRPC, so it does not require this gRPC-specific server code. The clients still use gRPC, which is why the client code in this crate is currently still required.

PR #3450, opened 2022-11-09
Set up an early GHCB page in stage0.

andrisaar:
We need a GHCB page to use the IOIO protocol under SEV-ES/SNP for two reasons:
  - We want to do some early logging, which means we need to access the serial port.
  - We need to set up the E820 tables, for which we need to use the QEMU `fw_cfg` interface, which again implies using IO ports.

Both of these will be tackled separately in future PRs; this just sets up a GHCB page for us to use in the future.

In particular, we'd like the comms to be initialized as early as possible. Earlier than the point we run `PVALIDATE` on physical memory; partly because we need to query `fw_cfg` to figure out what the physical memory is at all. Thus, we ask the GHCB location to be pre-validated in the SEV metadata blob.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -107,6 +107,11 @@ SECTIONS {
     } > ram_low
     ASSERT(SIZEOF(.boot.cpuid) == 4K, "CPUID section has to be exactly one page in size")
 
+    .boot.ghcb ALIGN(4K) (NOLOAD) : {
+        KEEP(*(.boot.ghcb))
+    } > ram_low
+    ASSERT(SIZEOF(.boot.ghcb) == 4K, "GHCB has to be exactly one page in size")
+
     .boot ALIGN(4K) (NOLOAD) : {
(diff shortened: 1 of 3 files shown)


stage0/layout.ld
```diff
@@ -107,6 +107,11 @@ SECTIONS {
     } > ram_low
     ASSERT(SIZEOF(.boot.cpuid) == 4K, "CPUID section has to be exactly one page in size")
 
+    .boot.ghcb (NOLOAD) : {
```
conradgrobler, line 110: Not sure about the linker script format, but should this also specify an alignment to make sure it is page-aligned?

conradgrobler (approved):
Nice.

stage0/layout.ld
andrisaar: It shouldn't matter as the GHCB struct itself is `align(4096)`, but it won't hurt to add it here as well. (The resulting BIOS blob is unchanged.)

PR #3451, opened 2022-11-09
Initialize the SNP CPUID and secrets pages

conradgrobler:
When SEV-SNP is enabled the Stage 0 firmware provides the addresses of the SNP-specific pages (CPUID and secrets pages) as part of the boot parameters.

This PR reads the locations of the pages from the boot paramters and initialises static instances. Using the information from these pages will be done in follow-up PRs.

### DIFF
--- a/oak_restricted_kernel/src/snp.rs
+++ b/oak_restricted_kernel/src/snp.rs
@@ -47,9 +47,9 @@ pub struct SnpPageAddresses {
 /// Tries to extract the guest-physical addresses of the SEV-SNP secrets page and CPUID page.
 ///
 /// This function must only be used while the identity mapping is still in place, since the pointers
-/// in the boot paramter page was constructed by the Stage 0 firmware with an identity mapping.
+/// in the boot parameter page was constructed by the Stage 0 firmware with an identity mapping.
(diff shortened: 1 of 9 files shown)


oak_restricted_kernel/src/snp.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub secrets_page_address: PhysAddr,
+    pub cpuid_page_address: PhysAddr,
+}
+
+/// Tries to extract the guest-physical addresses of the SEV-SNP secrets page and CPUID page.
+///
+/// This function must only be used while the identity mapping is still in place, since the pointers
+/// in the boot paramter page was constructed by the Stage 0 firmware with an identity mapping.
+///
+/// This function will panic if the boot paramter page is not valid.
```
tiziano88, line 52:
```suggestion
/// This function will panic if the boot parameter page is not valid.
```

tiziano88 (approved):

oak_restricted_kernel/src/snp.rs
conradgrobler: Thanks, done.

Issue #3442, opened 2022-11-08
Remote Attestation roadmap


tiziano88:
In order to converge on our long-term remote attestation story (mostly non-interactive, based on a key provisioning service), there are a few steps that we need to take.

At the moment we have an implementation of an interactive remote attestation protocol between client and enclave. Instead what we want is something more similar to the [offline attestation](https://github.com/project-oak/oak/tree/117ba0cdb15d4f2afc33df0710e7a40e8f45a38e/experimental/offline_attestation) experiment.

Since that's quite different from the current version, I suggest we first remove any remote attestation implementation and simplify the interface and implementation of Oak (especially Oak Functions), which also would allow us to implement a naive client library that just sends data without actual remote attestation.

Once we have that, we can then build the offline attestation layer on top of that by separating the key generation from the transfer of encrypted data.

At high level, the interface of Oak Functions would change so that the `Initialize` method returns the public key at startup, and the `HandleUserRequest` method accepts data encrypted via tink hybrid encryption directly.

https://github.com/project-oak/oak/blob/117ba0cdb15d4f2afc33df0710e7a40e8f45a38e/oak_functions_freestanding/proto/oak_functions.proto#L21-L29

On the untrusted launcher side, we will still expose a single streaming gRPC method, over which the client will perform two round trips: one to fetch the key (from the untrusted launcher), and the second to send the encrypted data (to the enclave binary).

tiziano88:
cc @sidtelang

PR #3452, opened 2022-11-09
Add proto definition for offline attestation

tiziano88:
Ref #3442

### DIFF
--- a/oak_remote_attestation_noninteractive/proto/oak_session_noninteractive_v1.proto
+++ b/oak_remote_attestation_noninteractive/proto/oak_session_noninteractive_v1.proto
@@ -1,0 +1,84 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak.session.noninteractive.v1;
+
+option java_multiple_files = true;
+option java_package = "oak.session.noninteractive.v1";
+
+message PublicKeyInfo {
+  // The serialized enclave public key.
+  // TODO(#3442): Specify key format.
+  bytes public_key = 1;
+  // The serialized attestation report binding the public key.
+  bytes attestation = 2;
+  // TODO(#3442): Add evidence (endorsements and inclusion proofs) for the attested binary.
+}
+
+message GetPublicKeyRequest {}
+
+message GetPublicKeyResponse {
+  // The enclave public key.
+  PublicKeyInfo public_key_info = 1;
+}
+
+message InvokeRequest {
+  // Body of the request, encrypted with the enclave public key.
+  // TODO(#3442): Specify encryption format.
+  // TODO(#3442): Also pass client public or symmetric key to use for response encryption. Note that
+  // needs to be at least authenticated by the client, if not encrypted within this field.
+  bytes encrypted_body = 1;
+}
+
+message InvokeResponse {
+  // Body of the request, encrypted with the client public or symmetric key.
+  // TODO(#3442): Specify encryption format.
+  bytes encrypted_body = 1;
+}
+
+message RequestWrapper {
+  oneof request {
+    GetPublicKeyRequest get_public_key_request = 1;
+    InvokeRequest invoke_request = 2;
+  }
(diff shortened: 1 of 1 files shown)


tiziano88:
@sidtelang

oak_offline_attestation/proto/oak_session_stream_v2.proto
```diff
@@ -0,0 +1,83 @@
...
+
+option java_multiple_files = true;
+option java_package = "oak.session.stream.v2";
+
+message PublicKeyInfo {
+  // The serialized enclave public key.
+  // TODO(#3442): Specify key format.
+  bytes public_key = 1;
+  // The serialized attestation report binding the public key.
+  bytes attestation = 2;
```
conradgrobler, line 29: I think this proto should also include additional evidence needed to validate the attestation report, such as transparency log inclusion proofs and the AMD certificates to validate the report signature. But, since I don't think we have finalised the formats for those, it could just be another TODO for this PR.

oak_offline_attestation/proto/oak_session_stream_v2.proto
```diff
@@ -0,0 +1,83 @@
...
+message GetPublicKeyResponse {
+  // The enclave public key.
+  PublicKeyInfo public_key_info = 1;
+}
+
+message InvokeRequest {
+  // Body of the request, encrypted with the enclave public key.
+  // TODO(#3442): Specify encryption format.
+  // TODO(#3442): Also pass client public or symmetric key to use for response encryption. Note that
+  // needs to be at least authenticated by the client, if not encrypted within this field.
```
conradgrobler, line 43: Good point. I think I missed this point in the existing experimental offline attestation implementation.

conradgrobler (approved):

oak_offline_attestation/proto/oak_session_stream_v2.proto
```diff
@@ -0,0 +1,83 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package oak.session.stream.v2;
```
ipetr0v, line 19: While this is a v2, the v1 with interactive attestation will still be needed for the Attestation Verification Service.
Will it also be remaned to v2 ?
I think we should have the attestation type (`interactive`, `non-interactive` or `offline`) in the package name.

tiziano88: I think that is not going to be oak speicific, and it's also not going to be over gRPC. And when we have a key provisioning service we will not need the stream at all, so this will be renamed anyways. But I agree the package name could be more descriptive. Should I call it `oak.session.offline.stream.v1` or something like that? I don't particularly like it, but let me know if you think it clarifies things.

ipetr0v: Since we will have an `interactive` attestation, the counterpart could be called `noninteractive` or something else. Since the `offline` may not be descriptive enough.

tiziano88: Done.

oak_remote_attestation_noninteractive/proto/oak_session_noninteractive_v1.proto
```diff
@@ -0,0 +1,84 @@
...
+}
+
+message GetPublicKeyRequest {}
+
+message GetPublicKeyResponse {
+  // The enclave public key.
+  PublicKeyInfo public_key_info = 1;
+}
+
+message InvokeRequest {
```
sidtelang, line 40: eventually, when two enclaves do a handshake tunneled through this Stubby/gRPC interface,

would a handshake frame sent in response to an round-trip (e.g. clientId after clientHello and serverHello) be of type InvokeRequest? (name feels a bit off?)

sidtelang (commented):
LGTM, just a few comments

oak_remote_attestation_noninteractive/proto/oak_session_noninteractive_v1.proto
```diff
@@ -0,0 +1,84 @@
...
+  // Body of the request, encrypted with the enclave public key.
+  // TODO(#3442): Specify encryption format.
+  // TODO(#3442): Also pass client public or symmetric key to use for response encryption. Note that
+  // needs to be at least authenticated by the client, if not encrypted within this field.
+  bytes encrypted_body = 1;
```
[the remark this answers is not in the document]
tiziano88: I like that idea; so the app would define the proto in some other way, and the client selectively encrypts some of the fields (perhaps using some of the code you already implemented with proto annotations etc, as long as we make that available to clients), and then serializes that here.
And actually also works in the short term, since we don't have encryption yet, so we could stretch the semantics a bit and say that encoded_request does not have any secret fields yet (until we implement the client encryption).
Does that make sense? If so, I'll send another PR.

PR #3453, opened 2022-11-09
Update TFLM to support REDUCE_ANY OP

ianull:
REDUCE_ANY OP is still pending for review at TFLM github ([PR](https://github.com/tensorflow/tflite-micro/pull/1538)).

TFLM reviewers don't have time reviewing it until next week.
To support demo which requires the op, update Oak TFLM source tree first.

The OP has passed [tests](https://github.com/tensorflow/tflite-micro/blob/977c3f19786026912cd9df95f9f52b01e1c509a0/tensorflow/lite/micro/kernels/reduce_test.cc#L449-L481) that are ported from tensorflow lite.


ipetr0v (approved):
Thanks!

tiziano88 (approved):

PR #3454, opened 2022-11-10
Support multi tensors in generic model runner

ianull:
Generic model runner is extended to support multiple input/output tensors that demo model requires.
Demo model has: 1 input tensor and 3 output tensors.
input tensor[0]: int * 10
output tensor[0]: float * 3
output tensor[1]: float * 105
output tensor[2]: float * 4522

As we only accept one single input buffer and one single output buffer,
input buffer will be split properly into input tensors per tensor size and
output tensors will be consolidated and cancatenated to output buffer in order.

Next step: I'll create a testing freestanding binary for demo model to validate flowing data through all layers w/o crashes.
I'll be off for rest of the week but could still check and work intermittently.

### DIFF
--- a/cc/tflite_micro/tflite_micro.cc
+++ b/cc/tflite_micro/tflite_micro.cc
@@ -22,20 +22,24 @@
 #include "tensorflow/lite/micro/micro_interpreter.h"
 #include "tensorflow/lite/schema/schema_generated.h"
 
-TfLiteTensor* input = nullptr;
-TfLiteTensor* output = nullptr;
+// Max number of input/output tensors
+#define MAX_TENSORS 100
 
 namespace {
 const tflite::Model* model = nullptr;
 tflite::MicroInterpreter* interpreter = nullptr;
+TfLiteTensor* inputs[MAX_TENSORS] = {nullptr};
+TfLiteTensor* outputs[MAX_TENSORS] = {nullptr};
+size_t input_size = 0;
+size_t output_size = 0;
 }
 
-const TfLiteTensor* tflite_get_input_tensor() {
-    return input;
+const TfLiteTensor* tflite_get_input_tensor(int id) {
+    return inputs[id];
 }
 
-const TfLiteTensor* tflite_get_output_tensor() {
-    return output;
+const TfLiteTensor* tflite_get_output_tensor(int id) {
+    return outputs[id];
 }
 
 int tflite_init(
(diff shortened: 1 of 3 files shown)


cc/tflite_micro/tflite_micro.cc
```diff
@@ -22,20 +22,24 @@
 #include "tensorflow/lite/micro/micro_interpreter.h"
 #include "tensorflow/lite/schema/schema_generated.h"
 
-TfLiteTensor* input = nullptr;
-TfLiteTensor* output = nullptr;
+// Max number of input/output tensors
+#define MAX_TENSORS 100
```
ipetr0v, line 26: Is having multiple input tensors specific to this demo?
Or it's a general use-case that we need to support for other models?

cc/tflite_micro/tflite_micro.cc
```diff
@@ -93,15 +113,18 @@ int tflite_run(
...
-                input->bytes,
+                input_size,
                 input_bytes_len);
     return -1;
   }
 
   // Place input in the model's input tensors
-  memcpy(input->data.uint8, input_bytes_ptr, input_bytes_len);
+  for (size_t i = 0, offset = 0; i < interpreter->inputs_size(); i++) {
+    memcpy(inputs[i]->data.uint8, input_bytes_ptr + offset, inputs[i]->bytes);
```
ipetr0v, line 125: Why do we need multiple input tensors?
I thought that TFLite models only accept a single tensor as an input?
Or models may have multiple inputs?

cc/tflite_micro/tflite_micro.cc
```diff
@@ -71,13 +75,29 @@ int tflite_init(
...
+    || interpreter->outputs_size() > MAX_TENSORS) {
+    MicroPrintf(
+      "input_tensors: %d, output_tensors: %d, max input/output tensors: %d",
+      interpreter->inputs_size(), interpreter->outputs_size(), MAX_TENSORS);
+    return -1;
+  }
+
+  // Obtain pointers and total size (bytes) of the model's input tensors.
+  for (size_t i = 0; i < interpreter->inputs_size(); i++) {
+    inputs[i] = interpreter->input(i);
```
ipetr0v, line 88: Does this line run the same model multiple times on different inputs?

ipetr0v (approved):

cc/tflite_micro/tflite_micro.cc
ianull: I think it's generic and we should support as many models as possible in a generic way as we can.

cc/tflite_micro/tflite_micro.cc
ianull: Nope, it prepares all input tensors observed from the model in `inputs[]` which will be used to dispatch input buffer to corresponding input tensors later.

cc/tflite_micro/tflite_micro.cc
ianull: Right, models can have multiple input tensors.
Think of a text sequence: "dog is running in the yard":
`dog` can be put in one input tensor to realize who or what animal.
`running` can be put in another input tensor to realize behavior.
`yard` can be put in another input tensor to realize venue.

In short, multiple input tensors learn and realize multiple features from the input.

PR #3455, opened 2022-11-10
Add PortFactory and a serial log to stage0

andrisaar:
This sets up a PortFactory that we can use to get access to the raw I/O ports in the system, which we will need for communicating via the serial port and in the future to use the QEMU fw_cfg API.

As the first step, this PR exercises the PortFactory functionality to enable basic logging in stage0.

However, unfortunately formatting is really expensive for the stack. I blew the 16K stack with hilarious results as usual; bumping it to 32K gives it enough headroom. Unfortunately this means the BIOS image won't fit in 64K any longer, and I had to bump the size to 128K.

In the future we should move the stack to the low 1M of memory as well, as it doesn't really need to be in the BIOS image area.

### DIFF
--- a/oak_core/Cargo.toml
+++ b/oak_core/Cargo.toml
@@ -4,6 +4,10 @@ version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
 
+[features]
+default = ["alloc"]
+alloc = ["dep:lock_api"]
+
 [dependencies]
-lock_api = { version = "*", features = ["arc_lock"] }
+lock_api = { version = "*", features = ["arc_lock"], optional = true }
 spinning_top = "*"
--- a/oak_core/src/lib.rs
+++ b/oak_core/src/lib.rs
@@ -15,6 +15,7 @@
 //
 #![no_std]
 
+#[cfg(feature = "alloc")]
 extern crate alloc;
 
 pub mod sync;
--- a/stage0/Cargo.lock
+++ b/stage0/Cargo.lock
@@ -173,6 +173,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "log"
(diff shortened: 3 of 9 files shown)


oak_core/src/sync.rs
```diff
@@ -50,6 +50,16 @@ impl<T> OnceCell<T> {
         Some(unsafe { (*self.value.get()).assume_init_ref() })
     }
 
+    pub fn get_mut(&self) -> Option<&mut T> {
```
andrisaar, line 53: Breadcrumbs for myself for tomorrow: this fn is a copy-paste of `get()` above but it's dangerous (no matter the copy-pasted safety argument below) as it can leak mutable references.
I'll either have to flip it around (something like `mutate(fn: FnOnce(&mut T)` that calls `fn` while holding the lock) or just mark this as `unsafe` and let the callers deal with sorting out the guarantees.

conradgrobler: I don't think this will ever be safe, even it you use `mutate` or leave it up to the caller. One option is to require a `&mut self`, but I don't think that will work for your purposes. The other option is to have an atomic reference counter to be sure there are no other current references so taking a mutable reference becomes safe, but that basically just reimplements an atomic refcell.

For the case of the serial port, we will only ever need mutable references to it, so I would suggest using a `OnceCell<Spnlock<SerialPort>>` and lock it for every write to ensure safe mutable access.

conradgrobler: Actually, since the serial port will only ever be useful when a mutable reference is available the OnceCell does not really add value. You could just use `SpinLock<Option<SerialPort>>`.

conradgrobler (approved):

PR #3456, opened 2022-11-11
Add support for encrypting and decrypting messages

conradgrobler:
Communications with the AMD platform secure processor via the guest message protocol must be encrypted to ensure that the messages aren't leaked or tampered with by the hypervisor. This PR provides an implementation of the required crypto based on the RustCrypto crates.

### DIFF
--- a/experimental/sev_guest/src/crypto.rs
+++ b/experimental/sev_guest/src/crypto.rs
@@ -63,9 +63,9 @@ impl GuestMessageEncryptor {
 
     /// Creates a new instance with a specific sequence number.
     ///
-    /// This functions should only be used if the sequence number was incremented previously to a
-    /// known value and but can not be incremented further outside of this struct, e.g. use this in
-    /// the kernel if the firmware used the same key to encrypt and decrypt guest messages.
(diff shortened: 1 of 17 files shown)


experimental/sev_guest/src/crypto.rs
```diff
@@ -0,0 +1,155 @@
...
+pub struct GuestMessageEncryptor {
+    cipher: Aes256Gcm,
+    sequence_number: u64,
+}
+
+impl GuestMessageEncryptor {
+    /// Creates a new instance.
+    ///
+    /// For now we always assume that VMPCK_0 will be used as the key.
+    pub fn new(key: &[u8]) -> Result<Self, &'static str> {
```
ipetr0v, line 60: Where is this key taken from?

ipetr0v (approved):

experimental/sev_guest/src/crypto.rs
conradgrobler: This is normally provided by the platform secure processor (PSP) in the secrets page that is injected into the guest memory at startup when running with AMD SEV-SNP.

PR #3460, opened 2022-11-11
Add emojify freestanding binary and enlarge tensor arena size

ianull:
We'll use this binary running on Linux to investigate model behavior and examine model output.

Note:
- Model can execute through entire pipeline without crashes.
- Model output is not checked yet.


oak_tensorflow_service/src/tflite.rs
```diff
@@ -68,7 +68,7 @@ extern "C" {
 }
 
 // TODO(#3297): Use 8GiB or 10GiB arena sizes.
-const TENSOR_ARENA_SIZE: usize = 2048;
+const TENSOR_ARENA_SIZE: usize = 1024 * 1024 * 1024;
```
ipetr0v, line 71: At some point we'll need to think what size should be the maximum size (for all of the use-cases).
Since we are planning to use the exact same Restricted Kernel binary for all of the use-cases.

ipetr0v (approved):

oak_tensorflow_service/src/tflite.rs
ianull: +1, to advantage Oak and onboard more use-cases, as big arena size as possible is more welcome.
Alternatively and arguably, we set a default, and provide client api to optionally override it to a bigger size upon invocations.

ipetr0v: We already have an `initialize` API function that loads a TFLM model.
Can we somehow determine (based on the model size), what tensor arena we would need?

ianull: We cannot determine it without running it at least once I think.
Afaict, there are some tools which can look into model file, iterate all tensors and query/sum # of parameters, then to determine static size required for the model. That said, no static tools can foresee and count the size of dynamic allocations from tensor arena without actually running the model. The per-op auxiliary data and supporting data structures are determined by each op implementation at execution time, whose size cannot be estimate accurately ahead of time.

PR #3461, opened 2022-11-11
Use the QEMU fw_cfg device to read the memory layout in stage0.

andrisaar:
For now, neither of the VMMs we use exposes the QEMU `fw_cfg` device, but longer term we do want to switch to using it (which will mean that we'll got from crosvm back to QEMU).

Thus, the only thing this PR really does is just debug log the E820 entries. We likely need to do some post-processing here as well, because (a) the entries are not sorted and (b) I believe SeaBIOS does something extra in its code, like marking ACPI memory ranges as reserved, which we should duplicate as well.

As an example, here's the output when I run stage0 with `qemu` and give it 32 MB of memory:
```
stage0 INFO: starting...
stage0 DEBUG: early E820 entry: start 0x00000000feffc000, len 16384, type RESERVED
stage0 DEBUG: early E820 entry: start 0x0000000000000000, len 33554432, type RAM
```

and here's what it looks like when I give it 8 GB of memory:
```
stage0 INFO: starting...
stage0 DEBUG: early E820 entry: start 0x00000000feffc000, len 16384, type RESERVED
stage0 DEBUG: early E820 entry: start 0x0000000000000000, len 3221225472, type RAM
stage0 DEBUG: early E820 entry: start 0x0000000100000000, len 5368709120, type RAM
```

### DIFF
--- a/experimental/sev_guest/src/ghcb.rs
+++ b/experimental/sev_guest/src/ghcb.rs
@@ -60,6 +60,11 @@ const IOIO_READ: u64 = 1;
 /// See section 15.10.2 of <https://www.amd.com/system/files/TechDocs/24593.pdf> for more details.
 const IOIO_DATA_SIZE_8: u64 = 1 << 4;
 
+/// Indicator bit that the IO data is an 16 bit number.
+///
+/// See section 15.10.2 of <https://www.amd.com/system/files/TechDocs/24593.pdf> for more details.
+const IOIO_DATA_SIZE_16: u64 = 1 << 5;
+
 /// Indicator bit that the IO data is a 32 bit number.
 ///
 /// See section 15.10.2 of <https://www.amd.com/system/files/TechDocs/24593.pdf> for more details.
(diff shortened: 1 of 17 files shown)


experimental/sev_guest/src/ghcb.rs
```diff
@@ -317,6 +322,34 @@ where
         Ok(self.ghcb.as_mut().rax as u8)
     }
 
+    /// Writes an 16 bit number to an IO port via the IOIO protocol.
```
conradgrobler, line 325: nit: `Writes a 16 bit ...`. Below as well.

stage0/src/fw_cfg.rs
```diff
@@ -0,0 +1,203 @@
...
+                continue;
+            };
+            entry = Some(file);
+            break;
+        }
+        if let Some(file) = entry {
+            // Safety: we dereference a valid pointer (as we get it from a reference) and the
+            // resulting array will have the same size as the underlying type.
+            let buf = unsafe {
+                core::slice::from_raw_parts_mut(
```
conradgrobler, line 158: This bit of unsafe code (and below) can be made a bit safer by deriving AsBytes and FromBytes from the `zerocopy` crate. `AsBytes` basically does the same as this, but deriving `FromBytes` ensures that the object will not cause undefined behaviour when writing arbitrary bytes to the memory by e.g. checking alignment and ensuring that none of the fields make assumptions about the underlying bytes.

https://docs.rs/zerocopy/latest/zerocopy/trait.AsBytes.html#method.as_bytes_mut

conradgrobler (approved):
Nice!

stage0/src/zero_page.rs
```diff
@@ -0,0 +1,68 @@
...
+static mut BOOT_ZERO_PAGE: MaybeUninit<BootParams> = MaybeUninit::uninit();
+
+pub fn init_zero_page(fw_cfg: &mut FwCfg) -> &'static mut BootParams {
+    let mut zero_page = unsafe { BOOT_ZERO_PAGE.write(core::mem::zeroed()) };
+
+    // Magic constants.
+    zero_page.hdr.type_of_loader = 0xFF; // loader type undefined
+    zero_page.hdr.boot_flag = 0xAA55; // magic number
+    zero_page.hdr.header = 0x53726448; // Magic "HdrS" string
+    zero_page.hdr.kernel_alignment = 0x1000000; // Magic number from crosvm source.
```
tiziano88, line 34: If you have a permalink to somewhere that defines these values, I think it will be useful to add them here as comments.

tiziano88 (approved):

stage0/src/zero_page.rs
andrisaar: Done.

stage0/src/fw_cfg.rs
andrisaar: Cool, I wasn't familiar with the  `zerocopy` crate!

PR #3462, opened 2022-11-14
Implement attestation report request

conradgrobler:
For now we just get an example attestation report and log it to the output. Replacing the `PlaceholderAmdAttestationBehaviour` with a real implementation that uses this report will be done in follow-up PRs.

### DIFF
--- a/experimental/sev_guest/src/crypto.rs
+++ b/experimental/sev_guest/src/crypto.rs
@@ -18,15 +18,12 @@
 //! using the RustCrypto `aes-gcm` crate.
 
 use crate::guest::{GuestMessage, Message};
-use aes_gcm::{
-    aes::{
-        cipher::{BlockSizeUser, Unsigned},
-        Aes256,
-    },
-    AeadInPlace, Aes256Gcm, KeyInit, Nonce, Tag,
-};
+use aes_gcm::{AeadInPlace, Aes256Gcm, KeyInit, Nonce, Tag};
 use core::mem::size_of;
 use zerocopy::{AsBytes, FromBytes};
+
+/// The size of initialization vector/nonce for AES-GCM
+const IV_SIZE: usize = 12;
 
 /// Wrapper for encrypting and decrypting guest messages.
 ///
@@ -89,7 +86,7 @@ impl GuestMessageEncryptor {
         let message_size = buffer.len();
         destination.header.auth_header.message_size = message_size as u16;
         destination.header.sequence_number = self.sequence_number + 1;
-        let mut iv_bytes = [0u8; <Aes256 as BlockSizeUser>::BlockSize::USIZE];
+        let mut iv_bytes = [0u8; IV_SIZE];
         iv_bytes[0..size_of::<u64>()]
             .copy_from_slice(destination.header.sequence_number.as_bytes());
         let nonce = Nonce::from_slice(&iv_bytes[..]);
@@ -122,7 +119,7 @@ impl GuestMessageEncryptor {
         if sequence_number != self.sequence_number + 1 {
             return Err("Unexpected sequence numer");
         }
-        let mut iv_bytes = [0u8; <Aes256 as BlockSizeUser>::BlockSize::USIZE];
+        let mut iv_bytes = [0u8; IV_SIZE];
         iv_bytes[0..size_of::<u64>()].copy_from_slice(sequence_number.as_bytes());
         let nonce = Nonce::from_slice(&iv_bytes[..]);
         let associated_data = source.header.auth_header.as_bytes();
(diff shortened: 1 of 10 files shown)


experimental/sev_guest/src/crypto.rs
```diff
@@ -71,61 +68,70 @@ impl GuestMessageEncryptor {
...
+            .copy_from_slice(destination.header.sequence_number.as_bytes());
         let nonce = Nonce::from_slice(&iv_bytes[..]);
-        let associated_data = result.header.auth_header.as_bytes();
-        result.payload[0..message_size].copy_from_slice(bytes);
-        let buffer = &mut result.payload[0..message_size];
+        let associated_data = destination.header.auth_header.as_bytes();
         let auth_tag = self
             .cipher
             .encrypt_in_place_detached(nonce, associated_data, buffer)
             .map_err(|_| "Message encryption failed")?;
```
tiziano88, line 97: not this PR, but should this error be printed too? or may it contain sensitive data?

oak_restricted_kernel/src/attestation.rs
```diff
@@ -0,0 +1,93 @@
...
+// The maximum number of custom bytes that can be included in the attestation report.
+const REPORT_DATA_MAX_SIZE: usize = 64;
+
+/// Requests an attestation rerport.
+///
+/// # Arguments
+///
+/// * `report_data` - The custom data that must be included in the report. This is typically used to
+///   bind information (such as the hash of a public key) to the report. A maximum of 64 bytes is
+///   supported.
```
tiziano88, line 37: I thought it's exactly 64 bytes. I guess it will be padded with zeros if shorter, but I think this function should expect an array of the exact size if possible, WDYT? What does the spec say?

tiziano88 (approved):
very cool

andrisaar (approved):

experimental/sev_guest/src/crypto.rs
```diff
@@ -71,61 +68,70 @@ impl GuestMessageEncryptor {
         })
     }
 
-    /// Creates an encrypted `GuestMessage` from the provided message.
+    /// Creates an encrypted payload from the provided message and writes that to the target's
+    /// payload field. It also upates the target's header with the appropriate information:
```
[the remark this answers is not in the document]
conradgrobler: Done.

experimental/sev_guest/src/crypto.rs
conradgrobler: This error type is deliberately opaque, so printing it will not be useful.  See https://docs.rs/aead/latest/aead/struct.Error.html

oak_restricted_kernel/src/attestation.rs
conradgrobler: I thought a slice might be more flexible, but I agree that the fixed-sized array is clearer. Done.

oak_restricted_kernel/src/attestation.rs
```diff
@@ -0,0 +1,93 @@
...
+        .ok_or_else(|| anyhow::anyhow!("Guest message encryptor is not initialized."))?;
+    if report_data.len() > REPORT_DATA_MAX_SIZE {
+        anyhow::bail!(
+            "Report data must be at most {} bytes.",
+            REPORT_DATA_MAX_SIZE
+        );
+    }
+
+    let mut report_request = AttestationRequest::new();
+    report_request.report_data[0..report_data.len()].copy_from_slice(report_data);
```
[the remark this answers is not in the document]
conradgrobler: Done.

oak_restricted_kernel/src/attestation.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let alloc = GUEST_HOST_HEAP
+        .get()
+        .ok_or_else(|| anyhow::anyhow!("Guest-host heap is not initialized."))?;
```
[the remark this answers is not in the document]
conradgrobler: I agree. This is quite incocnsistent throughout the code base. I will do a separate pass to clean that up.

conradgrobler (commented):
Thanks

experimental/sev_guest/src/crypto.rs
tiziano88: It would be nice to show that by pattern matching on it with the struct type, which shows that there are no other fields in that error.

conradgrobler: Just to clarify: do you mean just changing it to `.map_err(|_: aead::Error| "Message encryption failed")?;`, or something else?

tiziano88: `map_err(|aead::Error| "Message encryption failed")?` would this work?

PR #3466, opened 2022-11-14
Implement noninteractive attestation

tiziano88:
Remove the existing interactive attestation, and put in its place the skeleton for the noninteractive version.

Note that there is actually no encryption and no attestation at this commit, but it opens the door to implementing them in the future over the current protobuf messages.

Ref #3442


tiziano88:
Note for reviewers: please focus mainly on the `oak_functions_freestanding` crate, which has the most interesting changes.

oak_functions_freestanding/src/remote_attestation.rs
```diff
@@ -44,95 +36,48 @@ pub struct PublicKeyInfo {
...
+    fn message(&mut self, body: &[u8]) -> anyhow::Result<Vec<u8>>;
     fn get_public_key_info(&self) -> PublicKeyInfo;
 }
 
-pub struct AttestationSessionHandler<F, G: AttestationGenerator, V: AttestationVerifier> {
-    session_tracker: SessionTracker<G, V>,
-    request_handler: F,
-    public_key_info: PublicKeyInfo,
+pub struct AttestationSessionHandler<H: Handler> {
+    // TODO(#3442): Use attestation generator to attest public key.
```
conradgrobler, line 49: super nit: I think "attest to the public key", or "provide attestation evidence for the public key" is more accurate.

conradgrobler (approved):
nice

oak_functions_freestanding/src/remote_attestation.rs
tiziano88: Done.

oak_functions_launcher/src/server.rs
```diff
@@ -15,59 +15,46 @@
...
-    attestation: Vec<u8>,
+    // TODO(#3442): Return cached key and attestation when clients ask for it.
+    _signing_public_key: Vec<u8>,
+    _attestation: Vec<u8>,
 }
 
 #[tonic::async_trait]
-impl UnarySession for SessionProxy {
-    async fn message(
+impl oak_remote_attestation_noninteractive::proto::streaming_session_server::StreamingSession
```
ipetr0v, line 34: Once we add both interactive and non-interactive attestation, will they be implemented as different RPC services?

ipetr0v (approved):

oak_functions_launcher/src/server.rs
tiziano88: yes, since the interactive one will never be exposed externally, right?

oak_functions_freestanding/src/lib.rs
```diff
@@ -154,40 +124,20 @@ where
...
         match &mut self.initialization_state {
-            InitializationState::Uninitialized(_attestation_behavior) => {
-                Err(oak_idl::Status::new_with_message(
-                    oak_idl::StatusCode::FailedPrecondition,
-                    "not initialized",
-                ))
-            }
+            InitializationState::Uninitialized => Err(oak_idl::Status::new_with_message(
+                oak_idl::StatusCode::FailedPrecondition,
+                "not initialized",
```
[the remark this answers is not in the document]
tiziano88: yes but the client is the untrusted host in this case, which happens to forward this to the actual client (but we can change that later if we want to).

PR #3467, opened 2022-11-15
Make errors more consistent

conradgrobler:
Error messages and additional information throughout the codebase have been somewhat inconsistent. This PR tries to make it a bit more consistent, mostly around the following:

- Don't capitalize the first letter of an error message
- Don't end error messages with a `.`
- Use consistent forms of words (mostly `couldn't` over `could not` and `failed to`)

### DIFF
--- a/experimental/sev_guest/src/msr.rs
+++ b/experimental/sev_guest/src/msr.rs
@@ -248,7 +248,7 @@ impl TryFrom<u64> for PreferredGhcbGpaResponse {
     fn try_from(msr_value: u64) -> Result<Self, &'static str> {
         const PREFERRED_GPA_RESPONSE_INFO: u64 = 0x011;
         if msr_value & GHCB_INFO_MASK != PREFERRED_GPA_RESPONSE_INFO {
-            return Err("Vvlue is not a valid preferred GHCP GPA response");
+            return Err("value is not a valid preferred GHCP GPA response");
         }
         let ghcb_gpa = (msr_value & GCHP_DATA_MASK) as usize;
         Ok(Self { ghcb_gpa })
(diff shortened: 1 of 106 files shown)


experimental/sev_guest/src/msr.rs
```diff
@@ -248,7 +248,7 @@ impl TryFrom<u64> for PreferredGhcbGpaResponse {
     fn try_from(msr_value: u64) -> Result<Self, &'static str> {
         const PREFERRED_GPA_RESPONSE_INFO: u64 = 0x011;
         if msr_value & GHCB_INFO_MASK != PREFERRED_GPA_RESPONSE_INFO {
-            return Err("Value is not a valid preferred GHCP GPA response");
+            return Err("Vvlue is not a valid preferred GHCP GPA response");
```
tiziano88, line 251:
```suggestion
            return Err("value is not a valid preferred GHCP GPA response");
```

experimental/virtio/src/console/mod.rs
```diff
@@ -148,38 +148,38 @@ where
...
-            .context("Couldn't initialize the PCI device.")?;
+            .map_err(|error| anyhow::anyhow!("virtio error: {:?}", error))
+            .context("couldn't initialize the PCI device")?;
         self.device
             .configure_queue(
                 RX_QUEUE_ID,
                 QUEUE_SIZE as u16,
                 translate(self.rx_queue.inner.get_desc_addr())
-                    .context("Failed to translate VirtAddr to PhysAddr")?,
+                    .context("failed to translate VirtAddr to PhysAddr")?,
```
tiziano88, line 158: Maybe use `couldn't` here too?

tiziano88 (approved):
Thanks for doing it! That was a lot of changes! I hope you managed to use search & replace for at least some of them

experimental/virtio/src/console/mod.rs
conradgrobler: I was undecided about that, but will go ahead with it

experimental/sev_guest/src/msr.rs
conradgrobler: Done, thanks

tiziano88 (approved):

PR #3469, opened 2022-11-15
Fix public key malleability in offline attestation

conradgrobler:
The public key provided by the client was not covered by any cryptographic authentication, which meant that someone in the middle could replace it with a different pulbic key and intercept the response.

This change binds the public key to the encrypted payload as context info, which means that the payload cannot be decrypted if the public key is modified.


tiziano88 (approved):
Thanks

PR #3472, opened 2022-11-15
Generalize Java attestation client

tiziano88:
Now it does not depend on the gRPC method definition, only on an abstraction over it that works with any streaming method with the appropriate request and response types.

Ref #3442

### DIFF
--- a/java/src/main/java/com/google/oak/functions/weather_lookup_client/Main.java
+++ b/java/src/main/java/com/google/oak/functions/weather_lookup_client/Main.java
@@ -27,6 +27,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import oak.session.noninteractive.v1.StreamingSessionGrpc;
 
 public class Main {
   private static Logger logger = Logger.getLogger(Main.class.getName());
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/google/oak/functions/weather_lookup_client/Main.java
```diff
@@ -43,12 +44,12 @@ public static void main(String[] args) throws Exception {
...
 
-    // Attest a gRPC channel.
-    AttestationClientNoninteractive client = new AttestationClientNoninteractive(channel);
+    // Create gRPC client stub.
+    StreamingSessionGrpc.StreamingSessionStub client = StreamingSessionGrpc.newStub(channel);
 
     // Test request coordinates are defined in `oak_functions/lookup_data_generator/src/data.rs`.
     byte[] requestBody = "{\"lat\":0,\"lng\":0}".getBytes(UTF_8);
-    byte[] response = client.send(requestBody);
+    byte[] response = AttestationClientNoninteractive.invoke(client::stream, requestBody);
```
ipetr0v, line 52: Does `::stream` return a `Function<StreamObserver<ResponseWrapper>, StreamObserver<RequestWrapper>>` out of the box?

ipetr0v (approved):

java/src/main/java/com/google/oak/functions/weather_lookup_client/Main.java
tiziano88: Yes, it's a Java function reference

conradgrobler (approved):

PR #3475, opened 2022-11-15
Rebuild Oak TensorFlow if third-party has been updated

ipetr0v:
This PR makes `oak_tensorflow_service` rebuild the `tflite_micro` if third-party dependencies have been updated

### DIFF
--- a/oak_tensorflow_service/build.rs
+++ b/oak_tensorflow_service/build.rs
@@ -14,15 +14,14 @@
 // limitations under the License.
 //
 
-use glob::glob;
 use std::{
     path::{Path, PathBuf},
     process::Command,
+    str::from_utf8,
 };
 
 const TFLITE_DIR: &str = "cc/tflite_micro";
 const TFLITE_LIBRARY_NAME: &str = "tflite_micro";
-const TFLITE_SOURCES_PATTERN: &str = "cc/tflite_micro/**/*";
 
 fn main() {
     oak_idl_build::compile(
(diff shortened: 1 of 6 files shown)


oak_tensorflow_service/build.rs
```diff
@@ -22,7 +22,17 @@ use std::{
...
+const TFLITE_SOURCES_PATTERNS: &[&str] = &[
+    "cc/libc/**/*",
+    "cc/libgcc/**/*",
+    "cc/libm/**/*",
+    "cc/tflite_micro/**/*",
+    "third_party/flatbuffers/**/*",
+    "third_party/gemmlowp/**/*",
+    "third_party/nanolibc/**/*",
+    "third_party/ruy/**/*",
+    "third_party/tflite-micro/**/*",
```
tiziano88, line 34: This seems pretty hard and flaky to maintain -- can we get the actual list of files from bazel?

From a quick try, it seems the following gives you the list of files:

```
bazel query 'kind("source file", deps(//cc/tflite_micro:tflite_micro))'
```

(you need to remove the leading `//` and also convert `:` into `/` I think)

For instance, your list missed `third_party/bionic`.

I think all of this could be encapsulated in a single function `build_bazel_target` or similar, that does both the build and generates the cargo-rerun lines, as I think we will need it in other cases too.

tiziano88 (approved):

oak_tensorflow_service/build.rs
ipetr0v: Done

PR #3479, opened 2022-11-15
Use consistent naming

tiziano88:
In particular enclave, service, server and corresponding definitions.

Also request and response messages.


conradgrobler (approved):
I think this is a lot better, thanks.

PR #3480, opened 2022-11-15
Run `format` step as presubmit

### LINKED ISSUE #3478: C++ code is not correctly formatted
It seems that `xtask check-format` is not actually enforcing this, but `xtask format` does correctly format it.

Maybe we should just run `xtask format` at the end of the format check step anyways and catch any diffs via the git check diff step.

tiziano88:
Remove run-clang-format script, since it's unnecessary (and wasn't working anyways).

Fix #3478


tiziano88:
cc @ipetr0v @ianull

conradgrobler (approved):

PR #3473, opened 2022-11-15
Parameterizing the provenance workflow

rbehjati:

Issue #3428, opened 2022-11-03
Parametrise the provenance generation workflow 


rbehjati:
Currently the provenance generation workflow works only for `oak_functions_freestanding_bin`. We want to do the same thing for other binaries as well. 

The best solution to this seems to be to use a [GitHub job matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs). The matrix will iterate over a list of buildconfig toml files, and for each run all the steps for building the binary, generating the provenance, and publishing the comments. 

Instead of publishing a separate pair of comments for each buildconfig, we want to combine the content for all the buildconfigs, and publish only two comments: one containing the hashes of all the binaries, and one containing the hashes of all the provenances. One way to do this is to use a file or a variable to combine the results of all the steps. Alternatively, we could use artifact upload/download to generate the combined comments. 

With this change, adding support for a new binary would only require adding a new item to the matrix. 

Note that new buildconfigs should be added separately.

PR #3481, opened 2022-11-15
Refactor the provenance workflow to make parameterization easier

rbehjati:
Ref #3428.
Prerequisite for #3473.

Refactors the provenances workflow, moving most of the steps to a reusable workflow, to make parameterization with a matrix easier.

Some steps only run on push... this might still fail after merge.

### DIFF
--- a/.github/workflows/reusable_provenance.yaml
+++ b/.github/workflows/reusable_provenance.yaml
@@ -1,0 +1,183 @@
+# Builds an artifact from the input build config, and uploads it to Ent. Then generates the SLSA provenance for it and uploads that to Ent too.
+# Returns (1) the name and digest of the generated artifact as a based64-encoded string, and (2) the digest of the provenance.
+name: Build Provenance
+
+on:
+  workflow_call:
+    inputs:
+      build-config-path:
+        required: true
+        type: string
+    # Map the workflow outputs to job outputs
+    outputs:
+      artifact_digest:
+        description:
+          'Base64-encoded digest of the artifact, including its name.'
+        value: ${{ jobs.build_binary.outputs.hashes }}
+      provenance_digest:
+        description: 'SHA256 digest of the provenance.'
+        value: ${{ jobs.upload_provenance.outputs.provenance_digest }}
+    # usage: secrets: inherit
+    secrets:
+      ENT_API_KEY:
+        required: true
+
+jobs:
+  build_binary:
+    outputs:
+      hashes: ${{ steps.hash.outputs.hashes }}
+      filename: ${{ steps.filename.outputs.filename }}
+    runs-on: ubuntu-20.04
+
+    permissions:
+      # Allow the job to update the repo with the latest provenance info and index.
+      contents: write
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable_provenance.yaml
```diff
@@ -0,0 +1,171 @@
...
+  workflow_call:
+    inputs:
+      build-config-path:
+        required: true
+        type: string
+    # Map the workflow outputs to job outputs
+    outputs:
+      artifact_digest:
+        description:
+          'Base64-encoded digest of the artifact, including its name.'
```
tiziano88, line 15: not sure what format this is exactly, do you mean the binary name plus a space plus the sha256 digest?

.github/workflows/reusable_provenance.yaml
```diff
@@ -0,0 +1,171 @@
...
+        id: hash
+        run: |
+          # first cd to the directory containing the target binary. This allows us calling sha256
+          # on the target binary instead of having to provide the full path. This is fine for this
+          # experimental version, but has to be cleaned-up once we switch completely to SLSA
+          # provenance generation.
+          cd $(dirname ${{ steps.artifact_path.outputs.artifact_path }})
+          # sha256sum generates sha256 hash for the target binary.
+          # base64 --wrap=0 disables line wrapping, and encodes to base64 and outputs on a single line.
+          echo "hashes=$(sha256sum $(basename ${{ steps.artifact_path.outputs.artifact_path }}) | base64 --wrap=0)" >> $GITHUB_OUTPUT
```
tiziano88, line 114: How many hashes are there? I thought it would be only one. Also maybe this could be renamed to digest at some point for consistency.

.github/workflows/reusable_provenance.yaml
```diff
@@ -0,0 +1,171 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ needs.filename.outputs.filename }}.intoto.jsonl
+
+      # See https://github.com/google/ent
+      - name: Download Ent CLI
+        if: steps.cache-ent.outputs.cache-hit != 'true'
+        env:
+          ENT_URL: https://ent-server-62sa4xcfia-ew.a.run.app
+          ENT_DIGEST: sha256:944a34854a2bf9c5d32f3bffa93885ee1c7ef8ab0f4fcc30898a981050ae4233
```
tiziano88, line 155: Can this be shared with the other workflow too by creating a top level variable? I should probably create a reusable action for ent too at some point to encapsulate this, I think that would make things a bit cleaner, right?

tiziano88 (approved):
looking forward to this :)

.github/workflows/reusable_provenance.yaml
rbehjati: I think it cannot be shared, unless I make it an input to the reusable workflow. Do you think that is a good idea? If so, I can fix it in a future PR.

rbehjati (commented):
Thanks for the review :)

ipetr0v (approved):

PR #3482, opened 2022-11-16
Move `sev_guest` and `virtio` ouf of experimental

conradgrobler:


tiziano88 (approved):
LG but I think they should be prefixed with oak_ like everything else (especially if we think we may release them at some point)

conradgrobler:
> LG but I think they should be prefixed with oak_ like everything else (especially if we think we may release them at some point)

Thanks. I have prefixed it with "oak_" for consistency. I am not 100% convinced about having so many folders at the root starting with "oak_", but whether that is the best way to go can be discussed separately.

PR #3483, opened 2022-11-16
Fix TFLM page fault

ipetr0v:


oak_tensorflow_service/src/tflite.rs
```diff
@@ -95,16 +99,16 @@ impl TfliteModel {
             return Err(anyhow!("TFLite model has already been initialized"));
         }
 
-        let model_bytes_len = model_bytes.len();
-        let tensor_arena_len = self.tensor_arena.len();
-        let mut output_buffer_len = 0usize;
+        // Save a copy of the TFLite model.
```
tiziano88, line 102: Is it not possible to get the model as `Vec<u8>` to this method already and keep ownership of that to avoid an extra allocation and copy?

tiziano88 (approved):

oak_tensorflow_service/src/tflite.rs
ipetr0v: It's a reference to the proto message field: https://github.com/project-oak/oak/blob/0bd982eb256979bd17b87a7736db01cb7a01cde8/oak_tensorflow_service/src/lib.rs#L48

Issue #3486, opened 2022-11-18
Implement a tool to predict the AMD SEV-SNP attestation report measurement


conradgrobler:
To make the measurement in the attestation report meaningful we must be able to show that a the measurement is the result of running a specific verison of the Stage 0 firmware and a specific version of an enclave binary. The measurement if made of of various components, and support will be implementing in the following order:

- Measure the Stage 0 firmware
- Measure the enclave binary
- Account for SNP-specific pages (secrets, CPUID, unmeasured and zero pages)
- Measure the initial VMSA for the vCPU

PR #3487, opened 2022-11-18
Add initial attestation measurement calculation

conradgrobler:
This is the first step towards calculating the expected measurement in the attestation report from the Stage 0 firmware image and the enclave binary.

See #3486 for more details.

### DIFF
--- a/snp_measurement/src/stage0.rs
+++ b/snp_measurement/src/stage0.rs
@@ -23,17 +23,67 @@ use x86_64::PhysAddr;
 /// The address of the first byte after the end of the firmware image.
 ///
 /// The firmware image gets loaded just below the 4GiB boundary.
-pub const FIRMWARE_TOP: PhysAddr = PhysAddr::new(0x1_0000_0000);
+const FIRMWARE_TOP: PhysAddr = PhysAddr::new(0x1_0000_0000);
 
-/// Loads the Stage0 firmware ROM image from the supplied path.
-pub fn load_stage0(stage0_rom_path: PathBuf) -> anyhow::Result<Vec<u8>> {
+/// The address of the first byte after the end of the legacy boot shadow firmware image.
+///
+/// To support legacy booting the last 128KiB of the firmware gets shadowed just below the end of
+/// 20-bit memory.
+const LEGACY_TOP: PhysAddr = PhysAddr::new(0x10_0000);
+
+/// The maximum size of the shadow firmware for legacy boot.
+const LEGACY_MAX_SIZE: usize = 128 * 1024;
+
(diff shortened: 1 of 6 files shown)


snp_measurement/src/stage0.rs
```diff
@@ -0,0 +1,39 @@
...
+use anyhow::Context;
+use log::{debug, info};
+use sha2::{Digest, Sha256};
+use std::path::PathBuf;
+use x86_64::PhysAddr;
+
+/// The address of the first byte after the end of the firmware image.
+///
+/// The firmware image gets loaded just below the 4GiB boundary.
+pub const FIRMWARE_TOP: PhysAddr = PhysAddr::new(0x1_0000_0000);
```
andrisaar, line 26: The BIOS gets shadowed in the low memory as well. I don't know exactly how QEMU handles this, but some other VMMs just copy the first 128K of memory to 1M-128K.

snp_measurement/src/stage0.rs
```diff
@@ -0,0 +1,39 @@
...
+pub fn load_stage0(stage0_rom_path: PathBuf) -> anyhow::Result<Vec<u8>> {
+    let stage0_bytes =
+        std::fs::read(&stage0_rom_path).context("couldn't load stage0 firmware ROM image")?;
+    debug!("Stage0 size: {}", stage0_bytes.len());
+
+    let mut stage0_hasher = Sha256::new();
+    stage0_hasher.update(&stage0_bytes);
+    let stage0_cecksum = stage0_hasher.finalize();
+    info!("Stage0 SHA256 hash: {}", hex::encode(stage0_cecksum));
+    Ok(stage0_bytes)
```
tiziano88, line 38: Personally I recommend the following:

```suggestion
    let mut stage0_hasher = Sha256::new();
    stage0_hasher.update(&stage0_bytes);
    let stage0_sha256_digest = stage0_hasher.finalize();
    info!("Stage0 digest: sha256:{}", hex::encode(stage0_sha256_digest));
    Ok(stage0_bytes)
```

In particular, use the term "hash" for the function, and "digest" for the value, and also prefix the digest with the algo when printing.

Especially here that there are other hashing algos floating around :)

tiziano88 (approved):

rbehjati (approved):

snp_measurement/src/stage0.rs
conradgrobler: Done.

snp_measurement/src/stage0.rs
conradgrobler: Done.

conradgrobler (commented):
Thanks

andrisaar (approved):

PR #3489, opened 2022-11-21
Include VMSA page in the attestation measurement

conradgrobler:
The initial state of the VM state save area for the boot vCPU must be included in the attestation measurement calculation.

See #3486 for more details.

### DIFF
--- a/snp_measurement/src/vmsa.rs
+++ b/snp_measurement/src/vmsa.rs
@@ -16,7 +16,10 @@
 
 use log::trace;
 use oak_sev_guest::vmsa::{calculate_rdx_from_fms, Vmsa, VmsaPage};
-use x86_64::PhysAddr;
+use x86_64::{
+    structures::paging::{PageSize, Size4KiB},
+    PhysAddr,
+};
 
 /// The CPU family of the vCPU we expect to be running on.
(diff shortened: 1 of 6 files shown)


snp_measurement/src/vmsa.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// The guest-physical address of the VMSA page.
+///
+/// The current implementation uses the same fixed address for all VMSA pages. It is calculated as
+/// the start-address of the last 4KiB page that can be addressed within the allowed physical bits.
+///
+/// For AMD "Milan" CPUs the maximum supported physical memory bit-width is 48 when SEV-SNP is
+/// enabled.
+#[allow(unused)]
+pub const VMSA_ADDRESS: PhysAddr = PhysAddr::new(0xFFFF_FFFF_F000);
```
andrisaar, line 38: I wonder if `(1 << 48) - 0x1000` would be worth using here, as it would make the address more explicit at the cost of slightly lower readability.

conradgrobler: Good idea. I ended up goind with `(1 << 48) - Size4KiB::SIZE` to try and make it as readable as possible.

PR #3490, opened 2022-11-22
Include SEV-SNP pages in attestation measurement

conradgrobler:
The metadata of the SEV-SNP-related pages (secret, CPUID and unmeasured pages) must also be included in the attestation measurement.

The addresses and types of these pages are included in the the Stage 0 firmware image as metadata tables in the same way as is done in [OVMF](https://github.com/tianocore/edk2/tree/master/OvmfPkg), so we parse these values from the firmware image to be able to include them in the calculation.

See #3486 for more details.

### DIFF
--- a/snp_measurement/src/stage0.rs
+++ b/snp_measurement/src/stage0.rs
@@ -14,11 +14,17 @@
 // limitations under the License.
 //
 
+use crate::page::PageType;
 use anyhow::Context;
-use log::{debug, info};
+use log::{debug, info, trace};
 use sha2::{Digest, Sha256};
-use std::path::PathBuf;
-use x86_64::PhysAddr;
+use std::{collections::HashMap, mem::size_of, path::PathBuf};
+use strum::FromRepr;
+use x86_64::{
+    structures::paging::{PageSize, Size4KiB},
+    PhysAddr,
+};
+use zerocopy::AsBytes;
 
 /// The address of the first byte after the end of the firmware image.
 ///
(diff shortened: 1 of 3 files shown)


snp_measurement/src/stage0.rs
```diff
@@ -34,6 +40,44 @@ const LEGACY_TOP: PhysAddr = PhysAddr::new(0x10_0000);
 /// The maximum size of the shadow firmware for legacy boot.
 const LEGACY_MAX_SIZE: usize = 128 * 1024;
 
+/// The reverse offset from the end of the firmware blob to the end of the GUID tables.
+const GUID_TABLE_END_OFFSET: usize = 0x20;
+
+/// The size of the header of an entry in the GUID table.
+///
+/// The header consists of a 16 byte GUID and a 16-bit length field.
+const GUID_TABLE_ENTRY_HEADER_SIZE: usize = size_of::<u128>() + size_of::<u16>();
```
andrisaar, line 49: Minor nit, but it is it really a header if it is at the end? :)

snp_measurement/src/stage0.rs
```diff
@@ -87,3 +220,141 @@ pub fn load_stage0(stage0_rom_path: PathBuf) -> anyhow::Result<Stage0Info> {
...
+            SevMetadataPageType::Secrets => PageType::Secrets,
+            SevMetadataPageType::Unmeasured => PageType::Unmeasured,
+        }
+    }
+}
+
+/// The header of a guid table entry.
+///
+/// The table footer also has the same structure.
+struct GuidTableEntryHeader {
```
andrisaar, line 300: Could this use zerocopy::FromBytes instead of manually parsing data?

snp_measurement/src/stage0.rs
```diff
@@ -57,6 +101,95 @@ impl Stage0Info {
...
+            assert!(entry_end >= entry_size as usize, "invalid entry size");
+            let content_end = entry_end - GUID_TABLE_ENTRY_HEADER_SIZE;
+            let content_start = entry_end - entry_size as usize;
+            if let Some(_existing) = result.insert(entry_guid, &content[content_start..content_end])
+            {
+                panic!("duplicate GUID detected in table");
+            }
+            entry_end = content_start;
+        }
+        result
```
andrisaar, line 163: You should check that there were no bytes left at this point, otherwise the table is malformed.

andrisaar (approved):
Parsing in reverse is fun.

snp_measurement/src/stage0.rs
conradgrobler: I don't think it is great, but couldn't think of a better name. I am open to suggestions.

snp_measurement/src/stage0.rs
conradgrobler: It complained about alignment mismatch with `repr(packed)` and a there was size mismatch due to the additional padding with `repr(C)`.

snp_measurement/src/stage0.rs
conradgrobler: Good point. I will add a check.

conradgrobler (commented):
Thanks.

PR #3491, opened 2022-11-22
stage0: create ACPI tables based on instructions from fw_cfg

andrisaar:
This is a fairly straightforward and, at times, naive implementation of the ACPI table builder. I spent some time trying to do it "properly" by implementing my own dedicated heap allocator, but it was too much of a hassle.

In the future, to clean this up, we should use the standard allocator from `alloc` and the bog-standard Rust `Box` etc data types, even if this means we need to instantiate a global heap allocator to placate the Rust runtime.

All of the `fw_cfg` related code should also be moved to a separate crate, so that we can do proper testing.

### DIFF
--- a/stage0/src/acpi.rs
+++ b/stage0/src/acpi.rs
@@ -1,0 +1,411 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// These data structures (and constants) are derived from qemu/hw/acpi/bios-linker-loader.c that
+// defines the interface.
+
+use crate::fw_cfg::FwCfg;
+use core::{
+    ffi::CStr,
+    fmt::{Debug, Formatter, Result as FmtResult},
+    mem::{zeroed, MaybeUninit},
+};
+use strum::FromRepr;
+use zerocopy::AsBytes;
+
+// RSDP has to be within the first 1 KiB of EBDA, so we treat it separately.
+// See Section 5.2.5 of the ACPI Specification, Release 6.5 for more details about RSDP, it's size,
+// structure, and where it should be placed.
+const RSDP_SIZE: usize = 36;
+// Rest of the EBDA. The full size of EBDA is 128 KiB, but let's reserve the whole 1 KiB for the
+// RSDP.
+const EBDA_SIZE: usize = 127 * 1024;
+#[link_section = ".ebda.rsdp"]
+static mut RSDP: MaybeUninit<[u8; RSDP_SIZE]> = MaybeUninit::uninit();
+#[link_section = ".ebda"]
+static mut EBDA: MaybeUninit<[u8; EBDA_SIZE]> = MaybeUninit::uninit();
+
+// Safety: we include a nul byte at the end of the string, and that is the only nul byte.
+const TABLE_LOADER_FILE_NAME: &CStr =
+    unsafe { CStr::from_bytes_with_nul_unchecked(b"etc/table-loader\0") };
+const RSDP_FILE_NAME: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"etc/acpi/rsdp\0") };
+const ACPI_TABLES_FILE_NAME: &CStr =
+    unsafe { CStr::from_bytes_with_nul_unchecked(b"etc/acpi/tables\0") };
+
+const ROMFILE_LOADER_FILESZ: usize = 56;
+
+type RomfileName = [u8; ROMFILE_LOADER_FILESZ];
+
+fn get_file(name: &CStr) -> Result<&'static mut [u8], &'static str> {
+    // Safety: we do not have concurrent threads so accessing the static is safe, and even if
+    // Allocate has not been called yet, all values are valid for an [u8].
+    if name == RSDP_FILE_NAME {
+        Ok(unsafe { RSDP.assume_init_mut() })
+    } else if name == ACPI_TABLES_FILE_NAME {
+        Ok(unsafe { EBDA.assume_init_mut() })
+    } else {
+        Err("Unsupported file in table-loader")
+    }
+}
+
+#[repr(u32)]
+#[derive(Debug, FromRepr)]
+enum CommandTag {
+    Allocate = 1,
+    AddPointer = 2,
+    AddChecksum = 3,
+    WritePointer = 4,
+}
+
+#[repr(u8)]
+#[derive(Debug, FromRepr)]
+enum Zone {
+    High = 1,
+    FSeg = 2,
+}
+
+/// COMMAND_ALLOCATE - allocate a table from `file` subject to `align` alignment (must be power of
+/// 2) and `zone` (can be HIGH or FSEG) requirements.
+///
+/// Must appear exactly once for each file, and before this file is referenced by any other command.
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct Allocate {
+    file: RomfileName,
+    align: u32,
+    zone: u8,
+    _padding: [u8; 60],
+}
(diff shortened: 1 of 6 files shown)


ipetr0v (approved):

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+
+impl RomfileCommand {
+    fn tag(&self) -> Option<CommandTag> {
+        CommandTag::from_repr(self.tag)
+    }
+
+    fn extract(&self) -> Command<'_> {
+        // Safety: we extract the value out of the union based on the tag value, which is safe to
+        // do.
+        match self.tag().unwrap() {
```
conradgrobler, line 366: Is crashing OK here? Since `extract` is only used in a `invoke` which returns a result, perhaps it is better for `extract` to also return a result and use `.ok_or(...)?` insead of `unwrap`.

conradgrobler (approved):
Nice

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+        let src_file = get_file(self.src_file())?;
+
+        if self.offset as usize + self.size as usize > dest_file.len() {
+            return Err("Write for COMMAND_ADD_POINTER would overflow destination file");
+        }
+        if self.size > 8 || self.size.count_ones() != 1 {
+            return Err("COMMAND_ADD_POINTER has invalid size");
+        }
+        let mut pointer = 0u64;
+        unsafe {
```
conradgrobler, line 188: Oh, I forgot to mention: I think these unsafe blocks can be removed by using something like:

`pointer.as_bytes_mut()[..self.size].copy_from_slice(&dest_file[self.offset..(self.offset+size)])`

and

`dest_file[self.offset..(self.offset+size)].copy_from_slice(pointer.as_bytes()[..self.size])`

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+use core::{
+    ffi::CStr,
+    fmt::{Debug, Formatter, Result as FmtResult},
+    mem::{zeroed, MaybeUninit},
+};
+use strum::FromRepr;
+
+// RSDP has to be within the first 1K of EBDA, so we treat it separately.
+// See Section 5.2.5 of the ACPI Specification, Release 6.5 for more details about RSDP, it's size,
+// structure, and where it should be placed.
```
ipetr0v, line 30: What for are we going to use ACPI ? IIUC it's a protocol to discover new devices?

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+const EBDA_SIZE: usize = 127 * 1024;
+#[link_section = ".ebda.rsdp"]
+static mut RSDP: MaybeUninit<[u8; RSDP_SIZE]> = MaybeUninit::uninit();
+#[link_section = ".ebda"]
+static mut EBDA: MaybeUninit<[u8; EBDA_SIZE]> = MaybeUninit::uninit();
+
+// Safety: we include a nul byte at the end of the string, and that is the only nul byte.
+const TABLE_LOADER_FILE_NAME: &CStr =
+    unsafe { CStr::from_bytes_with_nul_unchecked(b"etc/table-loader\0") };
+const RSDP_FILE_NAME: &CStr = unsafe { CStr::from_bytes_with_nul_unchecked(b"etc/acpi/rsdp\0") };
```
ipetr0v, line 42: Is this running inside Restricted Kernel ?
How will we access files without having a file system?

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+/// 2) and `zone` (can be HIGH or FSEG) requirements.
+///
+/// Must appear exactly once for each file, and before this file is referenced by any other command.
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct Allocate {
+    file: RomfileName,
+    align: u32,
+    zone: u8,
+    _padding: [u8; 60],
```
ipetr0v, line 88: What do we need padding for?

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+
+#[repr(C)]
+#[derive(Default)]
+struct RomfileCommand {
+    tag: u32,
+    body: Body,
+}
+
+#[derive(Debug)]
+enum Command<'a> {
```
ipetr0v, line 340: When are these commands executed?
Each time a new device is connected?

stage0/src/fw_cfg.rs
```diff
@@ -69,6 +69,16 @@ impl DirEntry {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("DirEntry")
+            .field("name", &self.name())
+            .field("size", &self.size())
+            .field("select", &self.selector())
+            .finish()
+    }
+}
+
 pub struct FwCfg {
```
ipetr0v, line 82: pre-existing: could you please add a comment what "Fw" stands for?

stage0/src/fw_cfg.rs
```diff
@@ -130,6 +140,11 @@ impl FwCfg {
         })
     }
 
+    pub fn find(&mut self, name: &CStr) -> Option<DirEntry> {
```
ipetr0v, line 143: Why do you need `CStr` ? Are you calling into C somewhere?

ipetr0v (approved):

stage0/src/acpi.rs
andrisaar: ACPI is many things; amongst others, yes, it's used to describe the hardware in the machine.

And that's exactly what we want to use it for: for example, if we use a MMIO (memory-mapped) virtio-vsock device in QEMU, we need to know what memory addresses were assigned to that device. We could go and just randomly start probing memory addresses, but a much better solution would be to just look at the device list in ACPI, find the descriptor for the device in the ACPI tables, and find the address for the device from the descriptor.

Further out in the future, ACPI contains details about CPUs in the system as well (which we need when we get to multiprocessing), and a ton of other stuff (like a more modern factility to trigger a shutdown).

stage0/src/acpi.rs
andrisaar: Worse, this is stage0, which is the code running _before_ we get to Restricted Kernel :)

The QEMU `fw_cfg` exposes a really crude file system-like API. It'd be better to think about it as a sort of really low-level key-value interface, and these are the keys. QEMU just calls them "files", so we use the same terminology throughout.

stage0/src/acpi.rs
ipetr0v: Who is populating the ACPI table with new devices?

andrisaar: QEMU (or crosvm) provides us with the ACPI tables that contains information about hardware in the VM.

Crosvm writes the tables straight into guest memory; all of this code is about using the QEMU `fw_cfg` interface to read the ACPI tables and write them into the guest memory where the Restricted Kernel expects them to be.

stage0/src/acpi.rs
andrisaar: I'd argue it is, because this means we got something we don't support from the `etc/table-loader`. However, I agree that it'd be more consistent to push the error handling up to one place if we're doing it everywhere anyway.

stage0/src/acpi.rs
```diff
@@ -0,0 +1,411 @@
...
+            .field("offset", &self.offset)
+            .field("size", &self.size)
+            .finish()
+    }
+}
+
+/// COMMAND_ADD_CHECKSUM - calculate checksum of the range specified by `start` and `length` fields,
+/// and then add the value at `offset`.
+///
+/// Checksum simply sums -X for each byte X in the range using 8-bit math.
```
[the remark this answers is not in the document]
andrisaar: -a - b = -a + -b = -(a + b), so mathematically this doesn't make a difference, but I've added a note to clarify.

stage0/src/fw_cfg.rs
andrisaar: Indirectly -- the `fw_cfg` file names are null-terminated, like C strings are. Thus, it makes sense to use the Rust `CStr` for them.

stage0/src/acpi.rs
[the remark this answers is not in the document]
conradgrobler: ah, I missed the `-= checksum` so it seemed that it was never negated.

stage0/src/acpi.rs
andrisaar: Much cleaner, thanks!

stage0/src/fw_cfg.rs
andrisaar: Done.

PR #3493, opened 2022-11-23
Add enclave binary to attestation measurement

### LINKED ISSUE #3486: Implement a tool to predict the AMD SEV-SNP attestation report measurement

conradgrobler:
The contents of the enclave binary is the final part that is needed to calculate the predicted attestation measurement.

Fixes #3486

Note: The result does not yet match the actual measurements we get, as the boot paramters (zero page) is not yet marked as unmeasured so the measurement depends on the memory layout of the VM. This will be addressed in a follow-up PR.

### DIFF
--- a/snp_measurement/src/elf.rs
+++ b/snp_measurement/src/elf.rs
@@ -1,0 +1,57 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use anyhow::Context;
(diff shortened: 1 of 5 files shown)


snp_measurement/src/elf.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Loads an ELF file and extracts the non-zero loadable memory segments.
+pub fn load_elf_segments(elf_path: PathBuf) -> anyhow::Result<Vec<MemorySegment>> {
+    let elf_bytes = load_elf_file(elf_path)?;
+    let elf = Elf::parse(&elf_bytes).context("invalid ELF file")?;
+    // For now we assume each segment's start address will be 4KiB aligned, so we do not pad the
+    // start. This assumption is validated when the segments are measured.
+    Ok(elf
+        .program_headers
+        .iter()
+        .filter(|header| header.p_type == PT_LOAD && header.p_filesz > 0)
```
andrisaar, line 41: For `(NOLOAD)` sections where `p_filesz == 0`, do you need to do anything special for them or will they not affect the signature?

andrisaar (approved):

PR #3495, opened 2022-11-23
Update the initial VMSA state for measurements

conradgrobler:
This is to match the expected state when running with AMD SEV-SNP in production.


rbehjati (approved):
I am not entirely sure what the values are, but LGTM :)

conradgrobler:
> I am not entirely sure what the values are, but LGTM :)

Thanks. These values are just from the initial state we observe on actual runs on SEV-SNP test servers.

PR #3501, opened 2022-11-25
Stop using `lazy_static` in the restricted kernel

### LINKED ISSUE #3403: Implement a no_std-compatible equivalent to `OnceLock`

conradgrobler:
Fixes #3403

### DIFF
--- a/oak_core/src/sync.rs
+++ b/oak_core/src/sync.rs
@@ -40,6 +40,14 @@ impl<T> OnceCell<T> {
         }
     }
 
+    /// Gets an unwrapped reference to the inner value.
+    ///
+    /// Will panic if the cell has not been initialized.
+    pub fn get_unwrapped(&self) -> &T {
+        self.get().expect("cell not initialized")
+    }
+
+    /// Gets a reference to the inner value if the cell has been initialized.
     pub fn get(&self) -> Option<&T> {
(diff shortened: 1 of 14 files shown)


oak_core/src/sync.rs
```diff
@@ -40,6 +40,14 @@ impl<T> OnceCell<T> {
         }
     }
 
+    /// Gets an unwrapped reference to the inner value.
+    ///
+    /// Will panic if the cell has not been initialized.
+    pub fn get_unwrapped(&self) -> &T {
+        self.get().expect("cell not initialized")
+    }
```
tiziano88, line 48: Not sure what the value of this is, may as well unwrap / expect manually at the calling site. Or does it mirror some existing API?

tiziano88 (approved):

oak_core/src/sync.rs
conradgrobler: I just thought it would be simler to do it in one place, but I agree that calling expect at the call site makes it clearer.

PR #3506, opened 2022-11-25
Add a bunch of placeholder interrupt handlers

andrisaar:
This is related to the panic @ipetr0v got: unfortunately, when we get a double fault, there is no way to know which exception actually happened.

This PR adds a bunch of placeholder interrupt handlers that hopefully log _which_ exception occurred.

### DIFF
--- a/oak_restricted_kernel/src/interrupts.rs
+++ b/oak_restricted_kernel/src/interrupts.rs
@@ -164,13 +164,167 @@ mutable_interrupt_handler_with_error_code!(
     }
 );
 
+extern "x86-interrupt" fn divide_error_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: DIVIDE BY ZERO!");
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
+extern "x86-interrupt" fn nmi_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: NON-MASKABLE INTERRUPT!");
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
+extern "x86-interrupt" fn overflow_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: OVERFLOW!");
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
+extern "x86-interrupt" fn bound_range_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: BOUND RANGE EXCEEDED!");
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
+extern "x86-interrupt" fn invalid_opcode_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: INVALID OPCODE!");
+    error!(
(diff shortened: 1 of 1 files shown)


oak_restricted_kernel/src/interrupts.rs
```diff
@@ -164,13 +164,164 @@ mutable_interrupt_handler_with_error_code!(
     }
 );
 
+extern "x86-interrupt" fn divide_error_handler(stack_frame: InterruptStackFrame) {
```
ipetr0v, line 167: Are these functions being added to the interrupt handler?
Do we have a custom one?

ipetr0v: ah, see it on the bottom of the file

oak_restricted_kernel/src/interrupts.rs
```diff
@@ -164,13 +164,164 @@ mutable_interrupt_handler_with_error_code!(
...
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
 pub fn init_idt() {
     let mut idt = InterruptDescriptorTable::new();
+    idt.divide_error.set_handler_fn(divide_error_handler); // vector 0
```
ipetr0v, line 299: Could you please add a link where this list of interrupts was taken from?

oak_restricted_kernel/src/interrupts.rs
```diff
@@ -164,13 +164,164 @@ mutable_interrupt_handler_with_error_code!(
...
     idt.general_protection_fault
         .set_handler_fn(general_protection_fault_handler); // vector 13
     idt.page_fault.set_handler_fn(page_fault_handler); // vector 14
+                                                       // there is no vector 15
+    idt.x87_floating_point
+        .set_handler_fn(x87_floating_point_handler); // vector 16
+    idt.alignment_check.set_handler_fn(alignment_check_handler); // vector 17
+    idt.machine_check.set_handler_fn(machine_check_handler); // vector 18
+    idt.simd_floating_point.set_handler_fn(simd_fp_handler); // vector 19
+                                                             // there is no vector 20
```
ipetr0v, line 324: Is this a complete list of interrupts? Or only a subset?

ipetr0v (approved):

oak_restricted_kernel/src/interrupts.rs
andrisaar: Only a subset. The Rust `IDT` struct exposes nice wrappers for only the first 20 exceptions; some we don't care about  (vector 21, which is not exposed by the IDT struct, is about shadow stacks which we do not have enabled, for example).

Even if we get an exception outside the range of the ones we're handling, we'd still get a double fault. And that'll tell us that the fault happened somewhere really weird.

oak_restricted_kernel/src/interrupts.rs
```diff
@@ -164,13 +164,164 @@ mutable_interrupt_handler_with_error_code!(
...
+extern "x86-interrupt" fn divide_error_handler(stack_frame: InterruptStackFrame) {
+    error!("KERNEL PANIC: DIVIDE BY ZERO!");
+    error!(
+        "Instruction pointer: {:#016x}",
+        stack_frame.deref().instruction_pointer.as_u64()
+    );
+    shutdown::shutdown();
+}
+
+extern "x86-interrupt" fn nmi_handler(stack_frame: InterruptStackFrame) {
```
tiziano88, line 176: If you think this list is something that we will keep long term, it may be interesting to use a macro to reduce boilerplate.

tiziano88 (approved):

oak_restricted_kernel/src/interrupts.rs
tiziano88: (in a future PR, not here)

oak_restricted_kernel/src/interrupts.rs
andrisaar: I added a note for the AMD manual as I had that open; Intel has its own list.

conradgrobler (approved):

PR #3509, opened 2022-11-25
Create or fix buildconfig files

### LINKED ISSUE #3507: Create buildconfig file for Oak Functions enclave binary
We currently have https://github.com/project-oak/oak/blob/main/buildconfigs/oak_functions_freestanding_bin.toml as a placeholder, but we should make it not depend on xtask. Also I suspect we would need to add support for env variables to the buildconfig file format, since for this binary we first need to change PWD, and I don't think we want to involve the shell in this.

@rbehjati if you think that adding support for env variables makes sense, perhaps we should track it as a separate issue for transparent release?

Currently these are the steps we manually run for this specific binary:

```sh
cd oak_functions_freestanding_bin
cargo build --no-default-features --features=simple_io_channel --release
```

We should also probably rename the buildconfig file itself to `oak_functions_enclave.toml`.

tiziano88:
For all the enclave binaries that we care about.

Also add them to the appropriate CI step.

Fix #3507

### DIFF
--- a/.github/workflows/provenance.yaml
+++ b/.github/workflows/provenance.yaml
@@ -18,7 +18,10 @@ jobs:
       fail-fast: false
       matrix:
         buildconfig:
+          - buildconfigs/oak_echo_bin.toml
+          - buildconfigs/oak_echo_raw_bin.toml
           - buildconfigs/oak_functions_freestanding_bin.toml
+          - buildconfigs/oak_tensorflow_freestanding_bin.toml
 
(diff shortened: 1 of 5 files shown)


rbehjati (approved):
Wow! Nice. Thanks. 
Looking forward to seeing the new post-merge comments on the PRs :)

tiziano88:
cc @ipetr0v @conradgrobler @andrisaar

PR #3519, opened 2022-11-28
Add missing prerequisite step

rbehjati:
Follow up on #3518.

Also 
- Adds a few debug steps that get executed before merge. This is enabled by wrapping the SLSA generator reusable workflow into internal reusable workflow with a dry-run job.
- Merges the `publish_comments` job into `upload_provenance` to make the workflow faster (by reducing the number of times the workflow has to wait for getting a runner assigned to it)


tiziano88 (approved):

.github/workflows/generate_provenance.yaml
```diff
@@ -0,0 +1,53 @@
...
+          'Base64-encoded string containing the name and sha256 digest of the
+          file to generate the provenance for.'
+        required: true
+        type: string
+
+jobs:
+  # This job calls the generic SLSA workflow to generate provenance.
+  # See https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md
+  generate_provenance:
+    # The `generator_generic_slsa3` workflow fails on `pull_request` events.
```
tiziano88, line 28: Why though? Is it intentional or a bug? Is it worth following up with them (or have you already)?

.github/workflows/reusable_provenance.yaml
```diff
@@ -114,44 +116,42 @@ jobs:
...
-  # See https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md
   generate_provenance:
-    if: github.event_name == 'push'
     needs: [build_binary]
     permissions:
       actions: read
       id-token: write
       contents: write
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.1
+    uses: ./.github/workflows/generate_provenance.yaml
```
tiziano88, line 125: Will we get the correct information in the generated provenance this way? Since now technically the workflow is different. In other words: what do we need to check on the generated provenance to make sure it was genuinely created by the real SLSA generator?

tiziano88 (approved):

.github/workflows/generate_provenance.yaml
rbehjati: This is by design so the secrets won't be revealed I think. But I have discussed the need for a dry-run option. Not sure what is the status of that.

.github/workflows/reusable_provenance.yaml
rbehjati: It is not very different from the previous situation. This only adds another nesting of reusable workflows. But I agree this is worth looking into.

rbehjati: I check, in the newly generated provenances the entrypoint is the top-level workflow:

```
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "predicateType": "https://slsa.dev/provenance/v0.2",
  "subject": [
    {
      "name": "oak_tensorflow_freestanding_bin",
      "digest": {
        "sha256": "7e96476a2a5f72e3070f79d37e97d99f127611768f816e8dd3e2be1491844107"
      }
    }
  ],
  "predicate": {
    "builder": {
      "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.1"
    },
    "buildType": "https://github.com/slsa-framework/slsa-github-generator/generic@v1",
    "invocation": {
      "configSource": {
        "uri": "git+https://github.com/project-oak/oak@refs/heads/main",
        "digest": {
          "sha1": "2337f5c47b1c9642ab9d5222ede49ea9d9825f03"
        },
        "entryPoint": ".github/workflows/provenance.yaml"
      },
...
}
```

And the `builder.id` still is the slsa-generator workflow, and this `builder.id` is what we have to check.

PR #3520, opened 2022-11-28
Include freestanding binaries in workspace

tiziano88:


andrisaar:
> @conradgrobler @andrisaar before I spend too much time into this, is there any obvious reason why it should not work?

Yes, it won't pick up .cargo/config.toml filed.

andrisaar:
Ah. I haven't had a closer look, but I documented some issues in #2654 back in the day.

PR #3522, opened 2022-11-28
Rename `crosvm` launch mode to `virtual`

andrisaar:
First, this is a minor refactoring of names: instead of calling the variant `crosvm`, call it `virtual`. (We used to have `qemu` and `crosvm` variants, then we removed `qemu`, and now we'll likely switch back to QEMU. I don't think renaming `crosvm` to `qemu` makes much sense, let's instead use something more generic.)

Second, although the BIOS image is not used for crosvm, I've also included that in the xtask build and the command-line flags to the launcher. We will use it once we've switched to QEMU, and even if it takes a while, it's still good to make sure that stage0 actually builds.


oak_functions_launcher/src/main.rs
```diff
@@ -19,7 +19,7 @@
 
 use anyhow::Context;
 use clap::Parser;
-use instance::{crosvm, native, LaunchedInstance};
+use instance::{native, virtualised, LaunchedInstance};
```
tiziano88, line 22:
```suggestion
use instance::{native, virtualized, LaunchedInstance};
```

Unfortunately I think we use american english everywhere else.

tiziano88 (approved):

oak_functions_launcher/src/main.rs
```diff
@@ -43,7 +43,7 @@ mod server;
 #[derive(clap::Subcommand, Clone, Debug, PartialEq)]
 enum Mode {
     /// Launch enclave in crosvm
-    Crosvm(crosvm::Params),
+    Virtual(virtualized::Params),
```
ipetr0v, line 46: What's the difference between Virtual and Virtualized?
Which one is the launch mode?

ipetr0v (approved):

oak_functions_launcher/src/main.rs
andrisaar: I really wanted to name everything `Virtual`/`virtual`, but `virtual` is a reserved keyword.

Launch mode is `Virtual`.

PR #3524, opened 2022-11-28
Refactor ACPI code to include a simple `AcpiDevice` wrapper

andrisaar:
We'll soon need to search the ACPI device tree for the virtio device: this sets up the infrastructure so that we can do that.

### DIFF
--- a/oak_restricted_kernel/src/acpi.rs
+++ b/oak_restricted_kernel/src/acpi.rs
@@ -15,11 +15,11 @@
 //
 
 use acpi::{AcpiHandler, AcpiTables, AmlTable, PhysicalMapping};
-use alloc::{boxed::Box, string::String};
+use alloc::{boxed::Box, string::String, vec::Vec};
 use aml::{
     resource::{resource_descriptor_list, MemoryRangeDescriptor, Resource},
     value::Args,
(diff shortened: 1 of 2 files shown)


oak_restricted_kernel/src/acpi.rs
```diff
@@ -249,6 +249,56 @@ fn name_from_eisa_id(eisa_id: u64) -> Result<String, AmlError> {
...
+impl AcpiDevice {
+    /// Return the value of the `_HID` object, if present.
+    ///
+    /// ACPI specifications require one of `_HID` or `_ADR` to be present.
+    /// See Section 6.1 of the ACPI spec for more details.
+    pub fn hid(&self, ctx: &mut AmlContext) -> Result<Option<String>> {
+        let hid = self.invoke("_HID", ctx);
+
+        if let Ok(hid) = hid {
+            // If the name is an integer, it's a compressed EISA identifier (yes, really)
```
conradgrobler, line 266: super nit:  comments should end with a `.`

conradgrobler (approved):

oak_restricted_kernel/src/acpi.rs
andrisaar: I'll address that in the next PR as I don't want to spend 40+ minutes waiting for checks :)

PR #3526, opened 2022-11-28
Change virtio-console to use MMIO, not PCI and use QEMU

andrisaar:
We currently use virtio-vsock drivers that only support PCI with crosvm. Unfortunately, QEMU microvm doesn't support PCI (only PCI Express), so we can't keep using that.

What we do here is that instead of relying on PCI(e), we just rely on a simple memory-mapped device, discovered from ACPI, for the communcation. As our own homegrown drivers only support PCI, this also meant switching over to the `virtio-drivers` crate (which we should have done at some point anyway). Unfortunately `virtio-drivers` don't support vsock; however, we we already using virtio-console for comms in the distant past so this is not that big of a change.

Longer term: we probably want to switch back to virtio-vsock, and either have to contribute some code to virtio-drivers ourselves or wait until support for that lands in that crate by someone else.

### DIFF
--- a/oak_restricted_kernel/src/virtio_console.rs
+++ b/oak_restricted_kernel/src/virtio_console.rs
@@ -20,7 +20,7 @@ use crate::{
     virtio::Channel,
     ADDRESS_TRANSLATOR, GUEST_HOST_HEAP,
 };
-use alloc::vec::Vec;
+use alloc::{string::String, vec::Vec};
 use aml::{
     resource::{MemoryRangeDescriptor, Resource},
     AmlContext,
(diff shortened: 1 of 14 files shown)


oak_restricted_kernel/src/virtio_console.rs
```diff
@@ -0,0 +1,203 @@
...
+                ))
+            }
+            _ => continue,
+        }
+    }
+    None
+}
+
+pub fn get_console_channel<'a, X: Translator>(
+    translator: &X,
```
tiziano88, line 167: Maybe `translator: &impl Translator` works here?

tiziano88 (approved):

oak_restricted_kernel/src/virtio_console.rs
```diff
@@ -0,0 +1,203 @@
...
+                    PhysAddr::new((base_address + range_length) as u64),
+                ))
+            }
+            _ => continue,
+        }
+    }
+    None
+}
+
+pub fn get_console_channel<'a, X: Translator>(
```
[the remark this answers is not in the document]
andrisaar: Yep, the other references confuse it if you don't use it.

oak_restricted_kernel/src/virtio_console.rs
andrisaar: Indeed it does, thanks.

oak_restricted_kernel/src/virtio_console.rs
```diff
@@ -0,0 +1,208 @@
...
+            .as_u64() as usize
+    }
+}
+
+#[repr(transparent)]
+pub struct MmioConsoleChannel<'a> {
+    inner: Spinlock<VirtIOConsole<'a, OakHal, MmioTransport>>,
+}
+
+// Safety: for now, this is safe as we don't have threads in our kernel.
```
conradgrobler, line 101: Please create an issue and add a TODO to improve this in future. I am worried that we forget about this if we add multi-vCPU support.

conradgrobler (approved):

oak_restricted_kernel/src/virtio_console.rs
andrisaar: Good point, done.

PR #3527, opened 2022-11-28
Build std for stage0

tiziano88:
There was a race between the following commits:
- dd3be6ab3e9c47844869ff5807aca0ee1d8c360a
- c7207686ecda0c6474714a34ea1d8c4c2e2766c8

Which introduced a bug that causes Kokoro to fail.


tiziano88:
FYI @rbehjati

andrisaar (approved):

PR #3530, opened 2022-11-29
Propose new channel communication protocol spec

conradgrobler:
The current communication protocol does not provide the features we need to support interleaving of frames from multiple messages.

This PR proposes a new format for the communications. Once this is approved and submitted the changes will be implemented in follow-up PRs.


oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
-|          frame length         |          frame flags          |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                                                               |
-+   frame body [ request_message body [ method parameters ] ]   +
-|                                                               |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-```
+To receive a message the recipent incrementally reads frames from the
+communication channel. Messages are parsed from frames by appending the frame
+bodies in the order they are received. The recipient SHOULD validate the
```
mariaschett, line 167: Why SHOULD the recipient validate the sequence?

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
-MUST be sent as an uninterrupted sequence.
-
-### Receiving Messages
-
-To receive a message the recipent incrementally reads frames from the
-communication channel. The recipent MAY start parsing a message before all of
-its frames have been received.
-
-### Encoding Messages into Frames
+To send a message, it MUST first be encoded into frames.
```
mariaschett, line 153: nit: Can we formulate to active? Is it the "Sender" who has to do this?

To send a message, the sender MUST encode a message into frames. The sender MUST chunk the byte representation of a message into a set of frame bodies. [..]

The sender must form a frame from each frame body and MUST send the resulting frame set in sequenctial order of the frame number over the channel.

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
-### Method Layer
+### Invocation Layer
 
-Methods define the functionality exposed by service. Each method has clearly
-defined parameters and return types. The Oak IDL is used to define which methods
-are available, their id, parameter and return types. The Oak IDL is also
-responsible for encoding method parameters and return values to bytes.
+Methods define the functionality exposed by service. Each method defines input
+parameters and return types. Protocol buffers (protobuf) is used to define which
+methods are available, the method ids, input parameters and return types. It
```
mariaschett, line 28: It uses ... -> The service converts the requests and responses into bytes using standard protobuf encoding.

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
+Conceptually, the Oak Restricted Kernel Communication Protocol is made up of
+three layers. The [Invocation Layer](#invocation-layer) uses protocol buffers
+(protobuf) to define the methods exposed by the service. It is responsible for
+encoding requests and responses that wrap input parameters and return values.
+The [Message Layer](#message-layer) provides a logical wrapper for the encoded
+requests and responses. The [Frame Layer](#frame-layer) converts messages into
+frames which are then sent over the transport channel.
 
-### Method Layer
+### Invocation Layer
```
mariaschett, line 24: How does this relate to `oak_idl`?

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
 ##### Handling an invocation
 
 The service MUST respond to each request message with a single response message.
-It SHOULD respond to request messages in first-in-first-out order, but MAY
-prioritize some request messages. The service MAY send a response message before
-the request message has been received in its entirety. Sending a response
-message concludes the invocation.
+The service MAY send a response message before the request message has been
+received in its entirety, but MUST NOT send start sending a response before any
```
mariaschett, line 47: IIUC "it its entirety" implies that messages be only partially sent/received, i.e., not atomic.

If so, what does: "Sending a response message" mean excatly below? The point where the service starts sending a message or the entire message is sent?

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
 
-Methods define the functionality exposed by service. Each method has clearly
-defined parameters and return types. The Oak IDL is used to define which methods
-are available, their id, parameter and return types. The Oak IDL is also
-responsible for encoding method parameters and return values to bytes.
+Methods define the functionality exposed by service. Each method defines input
+parameters and return types. Protocol buffers (protobuf) is used to define which
+methods are available, the method ids, input parameters and return types. It
+uses standard protobuf encoding to convert the requests and responses into
+bytes.
```
mariaschett, line 30: The "bytes" are the request messages/response message---true?

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
...
 The client MUST NOT send response messages.
 
 ### Message Layer
 
-Messages are conceptual units to transport requests and responses.
+Messages are logical units used to transport requests and responses.
 
 There are exactly two messages types: request messages and response messages.
+Since the invocation layer is responsible for encoding requests and reponses, at
+the communication protocol level the encoding for request and response messages
```
mariaschett, line 59: nit: encoding for request messages.

mariaschett (approved):

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
 ##### Handling an invocation
 
 The service MUST respond to each request message with a single response message.
```
tiziano88, line 45:
```suggestion
The service MUST respond to each request message with exactly one response message.
```

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
...
 
 ### Message Layer
 
-Messages are conceptual units to transport requests and responses.
+Messages are logical units used to transport requests and responses.
 
 There are exactly two messages types: request messages and response messages.
+Since the invocation layer is responsible for encoding requests and reponses, at
+the communication protocol level the encoding for request and response messages
+follow an identical format.
```
tiziano88, line 60: Worth mentioning that requests will always "flow" in one direction, and responses in the opposite direction? I.e. that each logical channel has fixed roles for client and server, and invocations are always initiated by the client.

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
...
 
-#### Request Message
+#### Message
 
-A byte encoded request message MUST consist of the following fields:
+A byte encoded message MUST consist only of the encoded bytes:
```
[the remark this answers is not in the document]
tiziano88: +1 this is a bit surprising (though I understand the point is to specify that there is no length prefix here)

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                            padding                            |
+|                         invocation_id                         |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                         message_length                        |
```
tiziano88, line 98: I'm confused -- is this really the higher-level message length, or just this frame length?

If the high-level message length, I'm not sure why it needs to be here.
If the frame length, it seems redundant with the body length; in fact I think it would be better to have the overall frame length instead of just the body length.

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
  0                   1                   2                   3
  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                            padding                            |
+|                         invocation_id                         |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
tiziano88, line 97: Maybe a nit, but I think it would be nice to keep length as the first field for consistency and to simplify the low level parsing logic, and also potentially allow stacking combinators to express this transformation (e.g. receive byte stream, cut it into frames by just looking at the length field and send the content to the next combinator, etc.). We don't have to actually implement it that way, but I think it's nice if it could be done from the outside in if desired.

oak_channel/SPEC.md
tiziano88: +1

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
-For small messages sent as a single frame, the start flag and the end flag would
-be set, all other flags would be unset.
-
-### Layers Diagram
-
-The diagram below shows a practial example of how the request half of an
-invocation would look like after being encoded as two frames.
+The last frame body of the resulting set MAY be shorter than 4,080 bytes (the
+maximum length of a frame body). Any preceding frame bodies MUST be exactly
+4,080 bytes in length.
```
tiziano88, line 158: Is this precluding the option of having smaller frames then? That may be problematic if in the future we allow larger frame sizes, since there would be no backwards compatibility. But perhaps we don't care about it to keep things simple.

oak_channel/SPEC.md
tiziano88: +1 I would expect this to be MUST, and in fact the client should abort if an out-of-order frame is received.

oak_channel/SPEC.md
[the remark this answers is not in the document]
andrisaar: As proposed, the message layer indeed sounds completely unnecessary. It just states that "requests and responses are sequences of bytes", which is pretty much always the case.

Why do we need this layer at all?

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ...
 ```
 
-<!-- Diagram generated with https://www.luismg.com/protocol/, using the schema
-"padding:32,length:16,flags:16,body bytes...:64" -->
+<!-- Diagram generated with https://www.luismg.com/protocol/, using the spec
+"invocation_id:32,message_length:32,frame_number:32,body_length:32,body bytes...:64" -->
 
-- `padding`, 4 bytes
+- `invocation_id`, u32, little endian
```
andrisaar, line 113: Instead of specifying that all these fields are "little endian", I'd propose "machine native endianness" -- these messages will be exchanged between processes on the same machine, so if we ever move to a BE machine having to do two swaps seems a bit silly.

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
+frames which are then sent over the transport channel.
 
-### Method Layer
+### Invocation Layer
 
-Methods define the functionality exposed by service. Each method has clearly
-defined parameters and return types. The Oak IDL is used to define which methods
-are available, their id, parameter and return types. The Oak IDL is also
-responsible for encoding method parameters and return values to bytes.
+Methods define the functionality exposed by service. Each method defines input
```
rbehjati, line 26:
```suggestion
Methods define the functionality exposed by the service. Each method defines input
```

oak_channel/SPEC.md
rbehjati, line 47:
```suggestion
received in its entirety, but MUST NOT start sending a response before any
```

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
...
+concludes the invocation.
 
 The client MUST NOT send response messages.
 
 ### Message Layer
 
-Messages are conceptual units to transport requests and responses.
+Messages are logical units used to transport requests and responses.
 
 There are exactly two messages types: request messages and response messages.
```
rbehjati, line 57:
```suggestion
There are exactly two message types: request messages and response messages.
```

oak_channel/SPEC.md
rbehjati, line 60:
```suggestion
follows an identical format.
```

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
-<!-- Diagram generated with https://www.luismg.com/protocol/, using the schema
-"padding:32,length:16,flags:16,body bytes...:64" -->
+<!-- Diagram generated with https://www.luismg.com/protocol/, using the spec
+"invocation_id:32,message_length:32,frame_number:32,body_length:32,body bytes...:64" -->
 
-- `padding`, 4 bytes
+- `invocation_id`, u32, little endian
 
-  For future use.
+  The unique ID of the invocation that the frame's forms a part of. All the
```
rbehjati, line 115: I don't understand this sentence.

oak_channel/SPEC.md
rbehjati: How could a response message be sent before any part of the request has been received? How would the server know where to send the response to? Or is a connection already made at this stage?

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
 
-Conceptually, the Oak Baremetal Communication Protocol is made up of three
-layers. The [Method Layer](#method-layer) uses the Oak IDL to define the methods
-exposed by the service. It is responsible for encoding method parameters and
-return values. The [Message](#message-layer) takes encoded parameters / return
-values and wraps them as distinct messages that include metadata. The
-[Frame](#frame-layer) creates individual frames from messages, which are then
-sent over the transport channel.
+Conceptually, the Oak Restricted Kernel Communication Protocol is made up of
+three layers. The [Invocation Layer](#invocation-layer) uses protocol buffers
```
rbehjati, line 17: It is not clear to me what is the lowest layer of the stack here. From the description I expect the message layer to be at the highest level of the stack but the fact that is stated in the middle makes me uncertain about that.

oak_channel/SPEC.md
```diff
@@ -1,39 +1,38 @@
...
+parameters and return types. Protocol buffers (protobuf) is used to define which
+methods are available, the method ids, input parameters and return types. It
+uses standard protobuf encoding to convert the requests and responses into
+bytes.
 
 #### Invocation
 
-An invocation of a method. Each invocation consists of two parts: a request
-message sent by the client, followed by a response message sent by the service.
+Each method invocation consists of two parts: a request message sent by the
```
ipetr0v, line 34:
```suggestion
Each method invocation consists of two messages: a request message sent by the
```

rbehjati (approved):

oak_channel/SPEC.md
```diff
@@ -44,91 +43,47 @@ The service MUST NOT send request messages.
 ##### Handling an invocation
 
 The service MUST respond to each request message with a single response message.
-It SHOULD respond to request messages in first-in-first-out order, but MAY
-prioritize some request messages. The service MAY send a response message before
-the request message has been received in its entirety. Sending a response
-message concludes the invocation.
+The service MAY send a response message before the request message has been
+received in its entirety, but MUST NOT send start sending a response before any
+part of the request message has been received. Sending a response message
```
ipetr0v, line 48: I'm not sure if the response should be called a message, since this is the invocation layer and not the message layer.
I think it's better to call it "Invocation request" and "Invocation response". And keep the "message" only for describing the "Message layer" entities.

oak_channel/SPEC.md
ipetr0v: I think the reason for putting the message length here was to preallocate a correct amount of bytes for the whole message once the first frame is received.

oak_channel/SPEC.md
ipetr0v: Does the restricted kernel's endiannes also depend on the architecture?

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
+  ID of a response message MUST match the value of its related request message.
+  The invocation ID MUST be incremented for each method invocation, naturally
+  wrapping when the maximum value for the data type is reached.
 
-- `length`, u16, little endian
+- `message_length`, unsigned 32-bit little-endian integer
 
-  Length of the entire frame, including this field. The length of a frame MUST
-  NOT exceed 4,000 bytes.
+  The length of the overall message. This value MUST be larger or equal to
```
ipetr0v, line 123: Is it about the upper level message length or the frame length?

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
 
-    - 7 (MSB): unused
-    - 6: unused
-    - 5: unused
-    - 4: unused
-    - 3: unused
-    - 2: unused
-    - 1: end flag. MUST be set in the last frame of a message
-    - 0 (LSB): start flag. MUST be set in the first frame of a message
+  The position of this frame within the overall message. The first frame MUST
```
ipetr0v, line 131: Does the channel implementation need to assume that all of the frames in the "stream" will be maximum 4K length (in order to calculate the abount of expected frames) ?

oak_channel/SPEC.md
andrisaar: Yes, on a big-endian CPU we'd run the kernel in big-endian mode as well.

oak_channel/SPEC.md
conradgrobler: This is really the message length. There are two reasons for storing the message length in the frame header. The first is as Ivan pointed out is to be able to reserve memory for the message so that we don't have to keep reallocating when we receive a new frame. The second is for frame validation.

Regarding the overall frame length vs the body length: we need both so we have to do a calculation anyway to determine the one that is not specified. The reason for choosing to specify the body length rather than the frame length is that it matches the approach in the virtio spec.

oak_channel/SPEC.md
conradgrobler: I disagree. Communication protocol specifications are usually explicit about endianness to avoid doubt. Also, at the moment we are running the untrusted launcher and the host on the same machine, but that might not always be the case so we cannot assume that we will always have the same endianness on both sides of the channel (even though it seems extremely unlikely that we will ever run either part on a BE machine).

oak_channel/SPEC.md
conradgrobler: A message is split into frames. All these frames that make up a message must have the same invocation ID for that message. The invocation ID is used to know which message a frame belongs to.

oak_channel/SPEC.md
conradgrobler: The "message length" is the length of the message, not the frame.

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
-  NOT exceed 4,000 bytes.
+  The length of the overall message. This value MUST be larger or equal to
+  `body_length`. The message length for all frames that make up the message MUST
+  be the same.
 
-- `flags`, 2 bytes
+  The size of a single message MUST NOT exceed 2GiB.
 
-  - `flags[0]`
+- `frame_number`, unsigned 32-bit little-endian integer
```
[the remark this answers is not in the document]
conradgrobler: This can be calculated from the message size.

oak_channel/SPEC.md
conradgrobler: Yes, this is specified lower down under "Sending messages"

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
 
-    - 7 (MSB): unused
-    - 6: unused
-    - 5: unused
-    - 4: unused
-    - 3: unused
-    - 2: unused
-    - 1: unused
-    - 0 (LSB): unused
+  The length of the body byte array of the rfram. This value must not exceed
```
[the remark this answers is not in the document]
conradgrobler: Done.

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
+  4,080. The sum of the body lengths of all the frames in a message must equal
```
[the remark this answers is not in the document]
conradgrobler: This is a number, so in my opinion the maximum value should be specified as a number, not a number of bytes.

We have a maximum frame size of 4096 bytes and the header is 16 bytes, so 4080 is what is left over.

oak_channel/SPEC.md
```diff
@@ -138,149 +93,80 @@ A byte encoded frame MUST consist of the following fields:
...
+  the message length.
 
 - `body`, variable length byte array
 
-  Byte fragment of an encoded message. In order to not exceed the maximum frame
-  length, this byte array MUST NOT exceed 3,936 bytes.
+  Byte fragment of an encoded message. This size of the byte array MUST NOT
+  exceed 4,080 bytes.
+
+The maximum total length of a single frame is 4,096 bytes.
```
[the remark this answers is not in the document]
conradgrobler: This is a choice. It happens to match the buffer size we use for some of our comms channels over shared memory and is the default size for memory pages, but is still an arbitrary choice.

oak_channel/SPEC.md
conradgrobler: I disagree. I interpret MUST as something that will always break the protocol if it is ignored. Not validating this will not detect corruption if it happened, but it will not actively break correctly formatted things.

oak_channel/SPEC.md
conradgrobler: Yes. The alternative is to also specify the expected number of frames. I would rather add a protocol version and fix the frame size per protocol version rather than allowing different frame sizes withing the same protocol.

oak_channel/SPEC.md
conradgrobler: The point when the complete message has been transmitted.

oak_channel/SPEC.md
conradgrobler: I think we will always parse the entire header for a layer before parsing the bytes, so the exact order doesn't matter. But I have swapped the body length and the invocation ID so the two length fields are adjacent.

oak_channel/SPEC.md
conradgrobler: Done.

conradgrobler (commented):
Thanks for the reviews. I have made significant changed. Please take another look.

oak_channel/SPEC.md
tiziano88: `MUST   This word, or the terms "REQUIRED" or "SHALL", mean that the
   definition is an absolute requirement of the specification.`

`SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.`

(From https://www.rfc-editor.org/rfc/rfc2119)

I think this should be an absolute requirement of the specification, so MUST is appropriate.

conradgrobler: I will change it to MUST, but I think SHOULD could also be appropriate. In the case where the communcations channel is infallible and the sender is trusted to accurately follow the protocol it could be a valid decision to skip this validation.

ipetr0v (approved):

oak_channel/SPEC.md
```diff
@@ -43,244 +42,157 @@ The service MUST NOT send request messages.
...
-+                         body bytes...                         +
-|                                                               |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ...
-```
+  The version of the communication protocol in use. This value MUST be 1 for
+  this version of the specification.
 
-<!-- Diagram generated with https://www.luismg.com/protocol/, using the schema
-"length:32,invocation_id:32,body bytes...:64" -->
+- `body_length`, unsigned 16-bit little-endian integer
```
tiziano88, line 98: Wouldn't it be better to have the total frame length here instead of just the body? That way, if we evolve the protocol to a future version, as long as we keep these two fields the same, a parser would be able to at least delimit the entire frame, even if it cannot parse it. At the moment, the parser would have to know the exact length of all the fields for the specific version. Also, even if we have only one version, it seems nice to logically separate the delimitation of frames via their length from their actual parsing. WDYT?

oak_channel/SPEC.md
```diff
@@ -43,244 +42,157 @@ The service MUST NOT send request messages.
...
 
-  Byte encoded method parameters.
+#### Frame
 
-#### Response Message
+MUST consist of a header and a body. The encoded header is 16 bytes long and
+MUST consist of the following fields:
 
-A byte encoded response message MUST consist of the following fields:
+- `protocol_version`, unsigned 16-bit little-endian integer
```
tiziano88, line 93: Great idea to start thinking about versioning here.

oak_channel/SPEC.md
```diff
@@ -43,244 +42,157 @@ The service MUST NOT send request messages.
...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ ...
-```
+To send a message, the sender MUST first encode the message into frames:
 
-<!-- Diagram generated with https://www.luismg.com/protocol/, using the schema
-"frame padding:32,frame length:16,frame flags:16,frame body [ request_message length ]:32,frame body [ request_message invocation_id ]:32,frame body [ request_message method_id ]:32,frame body [ request_message padding ]:32,frame body [ request_message body [ method parameters ] ]:128" -->
+The sender MUST chunk byte representation of the message into a set of frame
+bodies. The last frame body of the resulting set MAY be shorter than 4,080 bytes
+(the maximum length of a frame body). Any preceding frame bodies MUST be exactly
+4,080 bytes in length. The sender MUST then create a frame from each frame body.
```
tiziano88, line 164: I would like to leave it to the client to decide what framing size to use, and just set the upper bound. And we should handle frames of different sizes within the same message too IMO.

tiziano88 (approved):

oak_channel/SPEC.md
conradgrobler: If full frames don't have a fixed size the sender will need to put the number of frames and the message size in the header if we want to be able to reserve the appropriate amount or memory and know when we have received all the frames. I will add an additional field to the header to support that.

tiziano88: Isn't the message size (which is already there) enough?

conradgrobler: It is enough to reserve the space, but since we are not guaranteed that each full frame will have the same size we cannot calculate the expected number of frames from the message size.

tiziano88: But I thought in practice the logic on the receiving side is to allocate the space, then append each frame in order until it's full. We don't need to know the number of frames ahead of time, just check that the total size adds up correctly.

tiziano88: Let's step back for a moment. Is the issue that we risk missing a frame within a message and not notice? Or that individual bits may be corrupted? If the former, then a flag indicating the last frame of a message may be enough. For the latter, I agree some sort of checksum may be useful, though that may be also done at the application layer in principle.

conradgrobler: The important part is to have very high confidence that we will detect frame misalignment or corruption of the header, as this will indicate an unrecoverable situation for the channel and we should terminate the workload. Knowing an exact expected value (or small valid range) for each header field for a specific frame helps a lot with this.

If any bits in the body are corrupted we will detect it (at least for encrypted user requests) when we try to decrypt it, but that should only affect the single message that is corruped, not the stream as a whole, so we won't have to terminate the workload.

In my previous message I mentioned a checksum for the message, but actually meant a checksum for the frame.

tiziano88: Makes sense. So perhaps have a small checksum and also flags for each header? similar to TCP.

oak_channel/SPEC.md
andrisaar: I found this a bit hard to read as well; consider making it a bulleted list.

andrisaar (approved):

oak_channel/SPEC.md
andrisaar: +1, I'd like to see some kind of checksum. Memory corruption, even inside one machine, is all too common.

conradgrobler: I don't think out-of-order frames or entire frames going missing is likely to happen due to data corruption, so we probably don't need frame numbers if we have a checksum to validate the header of the frame and rely on message size to know when we have received everything.

If we leave message integrity up to the application (e.g. since we are using authenticated encryption for user requests) we only really need the checksum to cover the header:

- Version
- Frame size
- Message Size
- Invocation ID
- Header Checksum

The header checksum could be something simple like the result or xoring previous batches of 4 bytes from the header, or it could be a truncated hash.

WDYT?

PR #3538, opened 2022-11-30
Rename oak_idl to micro_rpc

### LINKED ISSUE #3463: Rename `oak_idl` crate to `micro_rpc`
It is not really an IDL, just a way of generating RPC / IPC stubs based on service definitions from protobuf files. And it's also not Oak-specific.

tiziano88:
Fix #3463


micro_rpc/proto/messages.proto
```diff
@@ -16,7 +16,7 @@
 
 syntax = "proto3";
 
-package oak.protobuf_idl;
+package micro_rpc;
```
ipetr0v, line 19: Do we need an `oak.` in front?

ipetr0v (approved):

micro_rpc/proto/messages.proto
tiziano88: I'm trying to make it agnostic, but we'll see.

PR #3540, opened 2022-11-30
Move server creation method

tiziano88:
The previous method was added to the list of methods on the trait, but that risks conflicting with a real method with the same name.


conradgrobler (approved):
Nice idea

PR #3543, opened 2022-12-01
Update remote attestation documentation

tiziano88:


tiziano88:
cc @sidtelang

ipetr0v (approved):

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+autonumber
+
+participant C as Client
+participant U as Untrusted Launcher
+participant T as Trusted Enclave
+participant S as Secure Processor
+
+Note over U,S: Boot Trusted Enclave
+
+U->>S: Trusted Enclave Binary + firmware
```
conradgrobler, line 29: This is not 100% accurate. This goes via the VMM/hypervisor, and most of the work is done there. The secure processor is only involved in encrypting and measuring the pages after the hypervisor loaded the firmware and enclave binary into the guest memory.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+The client then checks:
 
-  Client-->>Client: Generate a Signing key pair
+- that the attestation report is signed by the TEE manufacturer (e.g. AMD,
+  Intel)
+- that the attestation report additional data includes the hash of the public
+  key, to confirm that the key pair was in fact generated from inside the
+  enclave
+- that the attestation report measurement corresponds to a trusted version of
+  the enclave binary
```
mariaschett, line 88: Eventually, the hash of the binary too?

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
-
-  Untrusted Launcher->>TEE Platform: Trusted Runtime Memory Contents
-  TEE Platform -->> TEE Platform: Store a measurement of the Trusted Runtime
-  TEE Platform ->> Trusted Runtime: Start the Trusted Runtime in a secure VM
+autonumber
+
+participant C as Client
+participant U as Untrusted Launcher
+participant T as Trusted Enclave
+participant S as Secure Processor
```
conradgrobler, line 25: "Secure Processor" is specific to AMD SEV-SNP. Pehaps make it clear elsewhere in the doc that this flow is specific to AMD SEV-SNP.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+
+U->>T: Configuration
+activate T
+T-->>T: Init logic
+T-->>T: Generate Encryption key pair
+T-->>T: Put Encryption public key hash in AttestationReport request
+T->>S: AttestationReportRequest
+S-->>S: Generate AttestationReport
+S->>T: Signed AttestationReport
+T->>U: enclave Public Key
```
conradgrobler, line 42: Public Key + Attestation report is sent here.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+C->>U: GetPublicKeyRequest
+activate U
+U->>C: GetPublicKeyResponse
+deactivate U
+C-->>C: Verify Enclave Public Key attestation
+
+Note over C,S: Send Encrypted Data
+
+loop
+   C-->>C: encrypt request
```
conradgrobler, line 56: Perhaps add a step here for generating a client key-pair here.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
-  Untrusted Launcher->>Trusted Runtime: ClientIdentity
+For each invocation (consisting of a request followed by a response) the client
+concatenates its public key with the request payload, and encrypts the resulting
+blob with the enclave public key using
+[Hybrid Encryption](https://developers.google.com/tink/hybrid).
 
-  Client-->>Client: Derive Session Keys
-  Trusted Runtime-->>Trusted Runtime: Derive Session Keys
-  Note over Client,Trusted Runtime: Handshake Completed
+It then sends the encrypted message to the server, which ensures it gets
```
mariaschett, line 101: I am not sure why the encrypted message ensures that the trusted enclave which generated the key gets the message. I think this is the only enclave which can decrypt it, no? But it could be send to a different one.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
-  Trusted Runtime-->>Trusted Runtime: Generate a Signing key pair
-  Trusted Runtime-->>Trusted Runtime: Put Signing public key hash in the AttestationReport request
-  Trusted Runtime->>TEE Platform: AttestationReport request
-  TEE Platform-->>TEE Platform: Generate AttestationReport
-  TEE Platform->>Trusted Runtime: Signed AttestationReport
+### Fetch Enclave Public Key
 
-  Note over Untrusted Launcher,Trusted Runtime: Initialization
+Initially, the client connects to the untrusted launcher and requests the
+attested enclave public key. This key was generated in the enclave, but it is
```
conradgrobler, line 75: perhaps something like "enclave public key and supporting evidence for validating it".

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
 
-loop Data Exchange
-  Client->>Untrusted Launcher: Encrypted request
-  Untrusted Launcher->>Trusted Runtime: Encrypted request
-  Trusted Runtime-->>Trusted Runtime: Invoke business logic
-  Trusted Runtime->>Untrusted Launcher: Encrypted response
-  Untrusted Launcher->>Client: Encrypted response
-end
-```
+The client receives the encrypted response, and decrypts it with its private
```
mariaschett, line 110: private -> secret?

mariaschett (approved):
Maybe we can indicate how much of this is work-in-progress?

Thank you!

conradgrobler (approved):

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+participant S as Secure Processor
+
+Note over U,S: Boot Trusted Enclave
+
+U->>S: Trusted Enclave Binary + firmware
+S->>T: Boot Trusted Enclave
+
+Note over U,S: Initialize
+
+U->>T: Configuration
```
sidtelang, line 34: is this security-critical configuration?

if so, then it should be part of the AttestationReport data (together with hash of public key)?

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
-  Client->>Untrusted Launcher: ClientHello
-  Untrusted Launcher->>Trusted Runtime: ClientHello
-  Trusted Runtime-->>Trusted Runtime: Generate an Ephemeral key pair
+### Send Encrypted Data
 
-  Trusted Runtime->>Untrusted Launcher: ServerIdentity
-  Untrusted Launcher->>Client: ServerIdentity
-  Client-->>Client: Validate ServerIdentity
+The client also generates an encryption key pair ahead of time (it may be reused
+across invocations, or it may be generated for each invocation).
```
sidtelang, line 94: > reused across invocations

This could lead to subtle issues.
Say Client A uses the same pk_A in requests to enclaves T1 and T2, and each respond with messages m1 and m2 encrypted with pk_A
The attacker can make it appear that T1 responded with m2 and T2 with m1. 

This may not break certain kinds of applications, but it makes it easier to reason about security for **all** applications following this paradigm if clients require per-request response keys.

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
-  Client-->>Client: Generate an Ephemeral key pair
-  Client->>Untrusted Launcher: ClientHello
-  Untrusted Launcher->>Trusted Runtime: ClientHello
-  Trusted Runtime-->>Trusted Runtime: Generate an Ephemeral key pair
+### Send Encrypted Data
 
-  Trusted Runtime->>Untrusted Launcher: ServerIdentity
-  Untrusted Launcher->>Client: ServerIdentity
-  Client-->>Client: Validate ServerIdentity
+The client also generates an encryption key pair ahead of time (it may be reused
```
sidtelang, line 93: why not symmetric encryption key?

generating a key pair per request can be expensive (and see other comment for why generating response key per-request is desirable)

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+across invocations, or it may be generated for each invocation).
 
-  Client->>Untrusted Launcher: ClientIdentity
-  Untrusted Launcher->>Trusted Runtime: ClientIdentity
+For each invocation (consisting of a request followed by a response) the client
+concatenates its public key with the request payload, and encrypts the resulting
+blob with the enclave public key using
+[Hybrid Encryption](https://developers.google.com/tink/hybrid).
```
sidtelang, line 99: Wonder if it's worth mentioning the use of context_info as a way to bind public metadata about the request to the encrypted_request.

The framing here implies the whole request (as seen by the enclave) is private so context_info is moot

But more generally with the IDL, consider a public request message with some fields whose values are encrypted to the enclave public key. 
Other fields could be public and yet consumed by the enclave: having such fields be public helps debugging (e.g. a query of death)

docs/remote-attestation.md
sidtelang: +1  for including mention of "evidence" with every mention of "enclave public key"

docs/remote-attestation.md
```diff
@@ -1,212 +1,111 @@
...
+  the enclave binary
 
-  Note over Client,Trusted Runtime: Handshake Start
+If any of these checks fails, the client refuses to go ahead.
 
-  Client-->>Client: Generate an Ephemeral key pair
-  Client->>Untrusted Launcher: ClientHello
-  Untrusted Launcher->>Trusted Runtime: ClientHello
-  Trusted Runtime-->>Trusted Runtime: Generate an Ephemeral key pair
+### Send Encrypted Data
```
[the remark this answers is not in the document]
tiziano88: Done.

docs/remote-attestation.md
tiziano88: Good point, changed to use fresh symmetric keys for each invocation, PTAL.

docs/remote-attestation.md
tiziano88: Done.

docs/remote-attestation.md
tiziano88: I meant that the server ensures it, reworded it now.

docs/remote-attestation.md
tiziano88: Done.

docs/remote-attestation.md
tiziano88: The measurement **is** the hash of the binary -- well, of the memory at boot time. Or did I miss your point?

docs/remote-attestation.md
tiziano88: Done.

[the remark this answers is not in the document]
tiziano88: Done.

tiziano88 (commented):
Thanks all for the relevant comments, it's much better now with all of your suggestions.

tiziano88:
BTW you can view the rendered diagram from the diff view by clicking on the three dots in the top right corner and then "view file"

sidtelang (approved):

PR #3546, opened 2022-12-02
Use CMOS to populate E820 table if one is not available from QEMU

andrisaar:
Sometimes the E820 table is not available in the `fw_cfg` interface, and we have to resort to more arcane methods of determining what memory is available to us.

This code is largely inspired by what SeaBIOS is doing under QEMU in the exact same scenario.

### DIFF
--- a/stage0/src/cmos.rs
+++ b/stage0/src/cmos.rs
@@ -1,0 +1,86 @@
+//
+// Copyright 2022 The Project Oak Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use oak_sev_guest::io::{IoPortFactory, PortFactoryWrapper, PortReader, PortWrapper, PortWriter};
+
+const CMOS_INDEX_PORT: u16 = 0x0070;
+const CMOS_DATA_PORT: u16 = 0x0071;
+
+// QEMU-specific CMOS config fields.
+const CMOS_MEM_EXTMEM_LOW: u8 = 0x30;
+const CMOS_MEM_EXTMEM_HIGH: u8 = 0x31;
+const CMOS_MEM_EXTMEM2_LOW: u8 = 0x34;
+const CMOS_MEM_EXTMEM2_HIGH: u8 = 0x35;
+const CMOS_MEM_HIGHMEM_LOW: u8 = 0x5b;
+const CMOS_MEM_HIGHMEM_MID: u8 = 0x5c;
+const CMOS_MEM_HIGHMEM_HIGH: u8 = 0x5d;
+
+const NMI_DISABLE_BIT: u8 = 0x80;
+
+pub struct Cmos {
+    index_port: PortWrapper<u8>,
+    data_port: PortWrapper<u8>,
+}
+
+impl Cmos {
+    /// Creates a new CMOS reader wrapper.
+    ///
+    /// # Safety
+    ///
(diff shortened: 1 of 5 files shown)


stage0/src/cmos.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn high_ram_size(&mut self) -> Result<u64, &'static str> {
+        let mut high: u64 = (self.read(CMOS_MEM_HIGHMEM_LOW)? as u64) << 16;
+        high |= (self.read(CMOS_MEM_HIGHMEM_MID)? as u64) << 24;
+        high |= (self.read(CMOS_MEM_HIGHMEM_HIGH)? as u64) << 32;
+
+        Ok(high)
+    }
+
+    fn read(&mut self, index: u8) -> Result<u8, &'static str> {
+        // Safety: we've asked the caller to guarantee that these ports are available when calling
```
tiziano88, line 79: maybe "exclusively available" or "reserved"?

stage0/src/zero_page.rs
```diff
@@ -59,6 +79,39 @@ pub fn init_zero_page(fw_cfg: &mut FwCfg) -> &'static mut BootParams {
     zero_page
 }
 
+/// Builds an E820 table by reading the low and high memory amount from CMOS.
+///
+/// The code is largely based on what SeaBIOS is doing, but
```
tiziano88, line 84: If you have permalinks to the SeaBIOS code that inspired this, it would also be helpful for future reference.

stage0/src/zero_page.rs
```diff
@@ -59,6 +79,39 @@ pub fn init_zero_page(fw_cfg: &mut FwCfg) -> &'static mut BootParams {
...
+/// <https://wiki.osdev.org/Detecting_Memory_%28x86%29> is also a good read on the topic.
+fn build_e820_from_nvram(
+    zero_page: &mut BootParams,
+    port_factory: PortFactoryWrapper,
+) -> Result<(), &'static str> {
+    // Safety: fw_cfg is available, so we're running under QEMU(ish), and QEMU has CMOS available at
+    // the required I/O ports.
```
tiziano88, line 91: Is this always the case in QEMU? i.e. does it not depend on the machine type, for instance?

tiziano88 (approved):

stage0/src/zero_page.rs
andrisaar: Maybe. We're only using `microvm` machine type, and that has a pre-populated E820 table available, so we won't hit this code path under QEMU. With a certain other QEMU-compatible VMM, however, we do need to use this interface.

I've expanded a comment a bit to hedge some of the risks. All in all, like I said in the comment above, we're just playing the same game as SeaBIOS here.

stage0/src/zero_page.rs
andrisaar: Done.

stage0/src/zero_page.rs
```diff
@@ -59,6 +79,41 @@ pub fn init_zero_page(fw_cfg: &mut FwCfg) -> &'static mut BootParams {
...
+
+    // Time to put all that we know together.
+    // Leave the last 4 pages of low memory as reserved: this includes the BIOS area, and other
+    // things that according to SeaBIOS are reserved.
+    if rs > 0xFFFB_C000 {
+        rs = 0xFFFB_C000;
+    };
+    zero_page.e820_entries = 2;
+    zero_page.e820_table[0] = BootE820Entry::new(0, rs as usize, E820EntryType::RAM);
+    zero_page.e820_table[1] = BootE820Entry::new(0xFFFB_C000, 4 * 0x1000, E820EntryType::RESERVED);
```
conradgrobler, line 106: From looking at the SEABIOS implementation, it seems that it also adds the 256KiB below 4GiB as reserved for the BIOS image (everything above 0xFFFC_0000).

conradgrobler (approved):

stage0/src/zero_page.rs
andrisaar: Ah, good catch, I somehow lost it in the clean-ups!

My idea here was that as [0xFFFB_C000, 0xFFFC_0000) is just before [0xFFFC_0000, 0x1_0000_0000), we can just reserve all of it with one entry. But I somehow lost the reservation for the higher 256K.

PR #3547, opened 2022-12-02
Assign team to dependabot updates

tiziano88:
Now that we don't have auto assignment enabled any more on the repo.


.github/dependabot.yml
```diff
@@ -1,23 +1,28 @@
...
-
-    # Explicitly trigger kokoro checks. They do not automatically run for
-    # external users.
     labels:
       - 'kokoro:run'
-  # Keep the following two in sync with the main one, above, as the freestanding targets have their own `Cargo.lock` files.
+    assignees:
+      - 'project-oak/core'
+
+  # Keep the following in sync with the main one, above, as the freestanding targets have their own `Cargo.lock` files.
```
andrisaar, line 19: At the very least I think we should add stage0 here as well, but we can do that in a separate PR>

andrisaar (approved):

PR #3549, opened 2022-12-02
Mark the location where we'll write the zero page as unmeasured

andrisaar:
As we're no longer relying on the VMM to provide us with a zero page, we need to ensure that the memory location is validated when we run under SEV-SNP before we write to it.

However, we're in a bit of a chicken and egg scenario here: we need the zero page to know where the physical memory is in order to validate all of the memory, and we need the memory for the zero page to be valid before we write to it.

Thus, add the zero page location to the SEV metadata as an unmeasured page, which means the VMM will do the state change for us before we boot.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -311,7 +311,7 @@ SECTIONS {
      * Unfortunately it doesn't look like `SIZEOF()` works here, so we have to keep track of the data
      * structure size by hand to ensure that it ends at the correct location in memory.
      */
-    .guid_tables TOP - 0x20 - (16 + 3*12 + 2*22 + 18) : {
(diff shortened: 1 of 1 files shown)


stage0/layout.ld
```diff
@@ -334,6 +334,10 @@ SECTIONS {
         LONG(ADDR(.boot.ghcb))
         LONG(SIZEOF(.boot.ghcb))
         LONG(SEV_SECTION_UNMEASURED)
```
conradgrobler, line 336: nit: since the GHCB is no longer the only unmeasured page, perhaps update this items comment to be a bit clearer

conradgrobler (approved):

PR #3557, opened 2022-12-09
Use an allocator for boot data structures in stage0

andrisaar:
Thus far, we've had a lot of tricky code to ensure that the boot data structures end up at well-known addresses (where the addresses were mostly derived from crosvm).

Truth is, it doesn't really matter where in memory those data structures are and we were just making things much more difficult for us.

By using a simple bump allocator -- as we're never going to deallocate anything anyway -- we can make the code much more simpler, more understandable, and safe.

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -18,14 +18,10 @@
 #![no_main]
 #![feature(cstr_from_bytes_until_nul)]
 
-use core::{
-    arch::asm,
-    ffi::c_void,
-    mem::{size_of, zeroed, MaybeUninit},
-    panic::PanicInfo,
-};
+use core::{arch::asm, ffi::c_void, mem::MaybeUninit, panic::PanicInfo};
 use goblin::elf::header;
 use oak_sev_guest::io::PortFactoryWrapper;
+use static_alloc::bump::Bump;
 use x86_64::{
     instructions::{hlt, interrupts::int3, segmentation::Segment, tlb},
     registers::{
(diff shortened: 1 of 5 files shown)


stage0/src/main.rs
```diff
@@ -212,14 +202,22 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
...
+        .expect("Failed to allocate memory for IDT");
 
     create_idt(idt);
     idt.load();
 
-    let pml4 = unsafe { BOOT_PML4.write(PageTable::new()) };
-    let pdpt = unsafe { BOOT_PDPT.write(PageTable::new()) };
-    let pd = unsafe { BOOT_PD.write(PageTable::new()) };
+    let pml4 = alloc
+        .leak(PageTable::new())
```
ipetr0v, line 213: This is not a memory leak right? =)

ipetr0v (approved):

stage0/src/main.rs
```diff
@@ -194,11 +177,18 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
         sev::validate_memory(zero_page, encrypted);
     }
 
+    // Initialize the bump allocator.
+    // Safety: we're the only thread, and we need to explicitly zero out the internal state if we're
+    // running under memory encryption.
+    let alloc = unsafe { BOOT_ALLOC.write(Bump::uninit()) };
```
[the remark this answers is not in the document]
andrisaar: It's a (very simple) memory allocation algorithm: https://os.phil-opp.com/allocator-designs/#bump-allocator

TL;DR: it maintans an internal offset, and every time you allocate a new data structure, the offset is increased by the size of that data structure.

stage0/src/main.rs
andrisaar: In the strictest sense of the word, it is :)

Namely, once you call `leak`, the memory is never deallocated (as you'd expect when something goes out of scope). That'd make it a memory leak, but that's exactly what we want: we want the data structures to stick around in memory.

You can get the memory back if you call `BOOT_ALLOC.reset()`, which causes all allocations to become invalid, so in that sense it's not a proper memory leak as the memocy _can_ be recovered. But we don't want to; we explicitly want to leak the memory.

PR #3560, opened 2022-12-12
Set ACPI RSDP location and kernel command line in zero_page

andrisaar:
Some low-level changes for the future Linux kernel compatibility:

 - set the ACPI RSDP location in the zero page structures. The Linux kernel won't search EBDA for RSDP, it expects the value in the zero page to point to the correct location.
 - if there's a kernel command line available in fw_cfg, load that into memory and pass it to the kernel as well.

### DIFF
--- a/stage0/src/acpi.rs
+++ b/stage0/src/acpi.rs
@@ -384,7 +384,10 @@ impl RomfileCommand {
     }
 }
 
-pub fn build_acpi_tables(fwcfg: &mut FwCfg) -> Result<(), &'static str> {
+/// Populates the ACPI tables per linking instructions in `etc/table-loader`.
+///
+/// Returns the address of the RSDP table.
+pub fn build_acpi_tables(fwcfg: &mut FwCfg) -> Result<u64, &'static str> {
     let mut commands: [RomfileCommand; 32] = Default::default();
 
     let file = fwcfg
@@ -415,5 +418,5 @@ pub fn build_acpi_tables(fwcfg: &mut FwCfg) -> Result<(), &'static str> {
         command.invoke(fwcfg)?;
     }
 
-    Ok(())
+    Ok(unsafe { RSDP.as_ptr() } as u64)
 }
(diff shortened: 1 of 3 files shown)


stage0/src/main.rs
```diff
@@ -259,6 +259,23 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
         zero_page.hdr.setup_data = &setup_data.header as *const oak_linux_boot_params::SetupData;
     }
 
+    if let Ok(cmdline_size) = fwcfg.read_cmdline_size() {
+        if cmdline_size > 0 {
+            let mut buf = alloc
+                .alloc(Layout::from_size_align(cmdline_size as usize, 1).unwrap())
```
mariaschett, line 265: Do we want to change this `unwrap` to expect too?

stage0/src/acpi.rs
```diff
@@ -384,7 +384,7 @@ impl RomfileCommand {
     }
 }
 
-pub fn build_acpi_tables(fwcfg: &mut FwCfg) -> Result<(), &'static str> {
+pub fn build_acpi_tables(fwcfg: &mut FwCfg) -> Result<u64, &'static str> {
```
tiziano88, line 387: Could you add a comment explaining what is being returned?

tiziano88 (approved):

stage0/src/fw_cfg.rs
```diff
@@ -24,9 +24,23 @@ const FWCFG_PORT_DATA: u16 = 0x511;
 
 const SIGNATURE: &[u8] = b"QEMU";
 
+/// Selector keys for "well-known" fw_cfg entries.
+///
+/// See QEMU include/standard-headers/linux/qemu_fw_cfg.h for the authoritative list.
+#[allow(dead_code)]
```
mariaschett, line 30: Why did you add more than the one you need now?

mariaschett (approved):
LGTM.

stage0/src/main.rs
andrisaar: I don't think this can actually fail, because (a) cmdline_size is an `u32` and therefore by always smaller than `isize::MAX`, and we fix the alignment itself to be 2^0 = 1. But I've changed it to an `expect()`, just in case.

stage0/src/acpi.rs
andrisaar: Done.

PR #3563, opened 2022-12-13
Stop special-casing the zero page in stage0

andrisaar:
Yet another simplification, or at least what I think simplifies things a bit: remove the magic memory location for the zero page, and allocate it from the bump allocator like every other data structure we have.

The only curveball here is that as we previously had marked the zero page location as unmeasured, we could access it when running under SEV-SNP without doing `PVALIDATE` on the memory address; now that we allocate the location dynamically, we can't do it. I've just added a bit of assembly code to `PVALIDATE` all of the lowest 1 MiB of memory, where we store our data structures.

Otherwise, I think this code now looks a bit more Rust-y. Suggestions and opinions are always welcome, of course, because I'm not sure it's entirely idiomatic yet.


linux_boot_params/src/lib.rs
```diff
@@ -606,8 +606,8 @@ pub struct EDIDInfo {
     pub dummy: [u8; 128usize],
 }
 
-#[repr(C, packed)]
-#[derive(Copy, Clone)]
+#[repr(C, align(4096))]
```
mariaschett, line 609: Why did you choose 4096 here?

stage0/layout.ld
```diff
@@ -275,7 +272,7 @@ SECTIONS {
      * Unfortunately it doesn't look like `SIZEOF()` works here, so we have to keep track of the data
      * structure size by hand to ensure that it ends at the correct location in memory.
      */
-    .guid_tables TOP - 0x20 - (16 + 4*12 + 2*22 + 18) : {
+    .guid_tables TOP - 0x20 - (16 + 3*12 + 2*22 + 18) : {
```
mariaschett, line 275: Why did this change?

stage0/src/asm/boot.s
```diff
@@ -65,11 +65,27 @@ _protected_mode_start:
...
                               # Bit 0 - SEV enabled
                               # Bit 1 - SEV-ES enabled
+                              # Bit 2 - SEV-SNP active
     test %eax, %eax           # is it zero?
     je no_encryption          # if yes, jump to no_encryption
+    # See if we're under SEV-SNP, and if yes, pre-emptively PVALIDATE the first megabyte of memory, as that's
+    # where we'll be storing many data structures.
+    and $0b100, %eax          # eax &= 0b100; -- SEV-SNP active
+    test %eax, %eax           # is it zero?
+    je 2f                     # if yes, no SNP, skip ahead
```
mariaschett, line 78: skip ahead -> skip PVALIDATE of memory

stage0/src/asm/boot.s
```diff
@@ -65,11 +65,27 @@ _protected_mode_start:
...
+    and $0b111, %eax          # eax &= 0b111;
                               # Bit 0 - SEV enabled
                               # Bit 1 - SEV-ES enabled
+                              # Bit 2 - SEV-SNP active
     test %eax, %eax           # is it zero?
     je no_encryption          # if yes, jump to no_encryption
+    # See if we're under SEV-SNP, and if yes, pre-emptively PVALIDATE the first megabyte of memory, as that's
+    # where we'll be storing many data structures.
+    and $0b100, %eax          # eax &= 0b100; -- SEV-SNP active
+    test %eax, %eax           # is it zero?
```
mariaschett, line 77: What does it mean if "it is zero?"

stage0/src/zero_page.rs
```diff
@@ -15,68 +15,88 @@
...
-};
+use core::{ffi::CStr, mem::size_of};
 use oak_linux_boot_params::{BootE820Entry, BootParams, E820EntryType};
 use oak_sev_guest::io::PortFactoryWrapper;
 
-#[link_section = ".boot.zero_page"]
-static mut BOOT_ZERO_PAGE: MaybeUninit<BootParams> = MaybeUninit::uninit();
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct ZeroPage {
```
mariaschett, line 24: Can we add a comment here? And actually for all public functions below.

stage0/src/zero_page.rs
```diff
@@ -15,68 +15,88 @@
...
+    inner: BootParams,
```
mariaschett, line 25: Why do you call this `inner` and not `bootParams`?

mariaschett (approved):
LGTM as far as I can tell.

linux_boot_params/src/lib.rs
andrisaar: 4096 (4 KiB) is the size of a memory page on the x86 architecture, and the expectation is that this data structure is on a separate page of memory. (As a side note, the size of the data structure is also exactly 4 KiB, so it completely fills exactly one page).

In the past, this 4K alignment was enforced as a side effect by the fact that we set up a separate, 4K-aligned section, in the linker script. Now that we no longer do that in the linker script, we need to enforce it here.

We should have had the alignment requirement here from the get-go, the fact that it all worked was just a happy accident.

stage0/layout.ld
andrisaar: We used to have four entries in the GUIDed tables; this PR deletes the one for the zero page, thus we're down to three entries.

stage0/src/asm/boot.s
andrisaar: `if eax == 0`, this is the way how you check if a register is zero in x86 assembly.

stage0/src/zero_page.rs
andrisaar: No particular reason beyond that seems to be a common pattern I've seen elsewhere.

rbehjati (approved):
LGTM. Nice safety comments :)

PR #3564, opened 2022-12-13
Move the stage0 stack to low memory, out of BIOS ROM area

andrisaar:
Right now we have the stack somewhere just below the 4 GiB boundary, where the BIOS ROMs are mapped.

This has two unfortunate side effects:
  - Sometimes (like the QEMU q35 machine) the BIOS ROM areas are _really_ read-only. This means having the stack there wouldn't work.
  - Having the stack in there means that 32 KiB of the total 128 KiB image size are just zeroes for the stack. Moving it down will (a) allow us to shave 32K off the BIOS image size, if we want to, and (b) let us increase the stack size, if necessary, without bloating the BIOS image size.

I've also removed the special "boot_stack_pointer"; once we jump to the kernel, we might as well recycle our existing stack as we're not going to use it any longer.

### DIFF
--- a/stage0/layout.ld
+++ b/stage0/layout.ld
@@ -69,8 +69,6 @@ SECTIONS {
     } > ram_low
     ASSERT(SIZEOF(.boot.zero_page) == 4K, "Zero page has to be exactly one page in size")
 
-    boot_stack_pointer = 0x8000;
-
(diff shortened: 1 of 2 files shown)


mariaschett (approved):
LGTM.

PR #3565, opened 2022-12-14
Remove the fixed location for GHCB in stage0

andrisaar:
This is the last data structure which has a magic location pre-allocated in the linker script: the GHCB, or guest-host communcation block, that's used under SEV-ES and SEV-SNP to communicate with the hypervisor.

While I do think this code is _a bit_ cleaner than previously, it's still a mess and not particularly Rust-y. This is the minimum I was able to do without starting heavy refactoring of our existing GHCB APIs, so better take the easy win here and come back later for more clean-ups.

### DIFF
--- a/stage0/src/main.rs
+++ b/stage0/src/main.rs
@@ -150,9 +150,21 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
     // We assume 0-th bit is never the encrypted bit.
     let encrypted = if encrypted > 0 { 1 << encrypted } else { 0 };
 
+    // Initialize the bump allocator.
+    // Safety: we're the only thread, and we need to explicitly zero out the internal state if we're
+    // running under memory encryption.
+    let alloc = unsafe { BOOT_ALLOC.write(Bump::uninit()) };
+
     // If we're under SEV-ES or SNP, we need a GHCB block for communication.
     let ghcb_protocol = if es {
-        Some(sev::init_ghcb(snp, encrypted))
+        // No point in calling expect() here, the logging isn't set up yet.
+        // In any case, this allocation should not fail. This is the first thing we allocate, the
+        // GHCB is 4K in size, and the allocator should be far bigger than that (as we have more
+        // data structures we need to allocate in there).
+        // If the allocation does fail, something is horribly broken and we have no hope of
+        // continuing.
+        let ghcb = alloc.leak(sev::Ghcb::new()).unwrap();
+        Some(sev::init_ghcb(ghcb, snp, encrypted))
     } else {
         None
     };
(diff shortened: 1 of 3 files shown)


stage0/src/sev.rs
```diff
@@ -34,10 +33,21 @@ use x86_64::{
     PhysAddr, VirtAddr,
 };
 
-#[link_section = ".boot.ghcb"]
-static mut GHCB: MaybeUninit<Ghcb> = MaybeUninit::uninit();
+static GHCB_WRAPPER: OnceCell<Spinlock<GhcbProtocol<'static, SevGhcb>>> = OnceCell::new();
 
-static GHCB_WRAPPER: OnceCell<Spinlock<GhcbProtocol<'static, Ghcb>>> = OnceCell::new();
+#[repr(transparent)]
```
rbehjati, line 38: Perhaps add a `why` comment here, since `Ghcb` is wrapping another `Ghcb` from another crate. Why is this wrapping needed?

rbehjati (approved):

stage0/src/sev.rs
```diff
@@ -87,7 +100,7 @@ pub fn init_ghcb(snp: bool, encrypted: u64) -> &'static Spinlock<GhcbProtocol<'s
 }
 
 pub fn deinit_ghcb(snp: bool, encrypted: u64) {
-    let ghcb_addr = unsafe { GHCB.as_ptr() } as u64;
+    let ghcb_addr = GHCB_WRAPPER.get().unwrap().lock().get_gpa().as_u64();
```
mariaschett, line 103: Nice, getting rid of the unsafe / magic.

stage0/src/main.rs
```diff
@@ -150,9 +150,16 @@ pub extern "C" fn rust64_start(encrypted: u64) -> ! {
...
 
+    // Initialize the bump allocator.
+    // Safety: we're the only thread, and we need to explicitly zero out the internal state if we're
+    // running under memory encryption.
+    let alloc = unsafe { BOOT_ALLOC.write(Bump::uninit()) };
+
     // If we're under SEV-ES or SNP, we need a GHCB block for communication.
     let ghcb_protocol = if es {
-        Some(sev::init_ghcb(snp, encrypted))
+        // No point in calling expect() here, the logging isn't set up yet.
```
mariaschett, line 160: Can you add a comment on what you would write into the `expect` message, though?

mariaschett (approved):

stage0/src/sev.rs
andrisaar: I got rid of it; it was a remnant of me trying to invent something more Rust-y around the GHCB before giving up.

stage0/src/sev.rs
```diff
@@ -34,10 +33,21 @@ use x86_64::{
...
 };
 
-#[link_section = ".boot.ghcb"]
-static mut GHCB: MaybeUninit<Ghcb> = MaybeUninit::uninit();
+static GHCB_WRAPPER: OnceCell<Spinlock<GhcbProtocol<'static, SevGhcb>>> = OnceCell::new();
 
-static GHCB_WRAPPER: OnceCell<Spinlock<GhcbProtocol<'static, Ghcb>>> = OnceCell::new();
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct Ghcb {
```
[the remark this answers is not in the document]
andrisaar: I got rid of it; it was a remnant of me trying to invent something more Rust-y around the GHCB before giving up.

stage0/src/main.rs
andrisaar: Done.

